PR #120, opened 2022-04-19
Some improvements in TDS fault injection framework

Deepesh125:
### Description

This commit fixes the following issues as described below:

1. Freeing string info was not correctly accounted in a few fault injection callback functions like tamper_rpc_request and tamper_request_sequential. Which was causing the server to crash. So, this change fixes this issue. Also, We should not reset Message memory context while reading and processing the request. So, fixed that as well.

2. Certain types of fault injection should be called for certain type of requests. For example, tamper_rpc_request is supposed to be called for RPC request. This commit fixes this by checking request type inside fault injection callback before trying to tamper the request.

3. We were making use of list_delete() API incorrectly to disable any fault injection trigger  inside FaultInjectionDisableTest(). Which was causing error like “unrecognized node type: 1600483952” to occur since equal() function was not able to find the node type. So, this commit fixes that issue by using list_delete_cell() API to delete certain node from injected entries.

4. For "tds_comm_throw_error" trigger, we were throwing ERROR while processing the header (while reading the request). Any error raising from such a low level should be FATAL. So, introduced throw_error_comm() callback for "tds_comm_throw_error" trigger which will raise the FATAL error while processing the header if fault injection trigger is enabled.

5. Added TDS fault injection tests as well. 

Task: BABEL-3071
Signed-off-by: Dipesh Dhameliya (dddhamel@amazon.com)

### Check List


surendravishnoi (commented):
Please improve commit messages, there are some grammar issues and also make them more compact and targeted.

surendravishnoi (approved):

KushaalShroff (approved):

PR #124, opened 2022-04-20
Use logical schema name to write table name in column metadata instead of physical schema name.

Deepesh125:
### Description

Previously, Physical schema name i.e., master_dbo, temp_dbo, etc was being written as part of column metadata while sending the response.But logical schema should be written as far as Babelfish concerned.

In this commit, I introduced the rendezvous variable pltsql_get_logical_schema_name which would be initialised by babelfishpg_tsql extension to get_logical_schema_name_from_physical_schema_name().

get_logical_schema_name_from_physical_schema_name() would first check if given physical schema name needs to be translated to logical schema name or not. And it uses existing api get_logical_schema_name() to retrieve the logical schema name given physical schema name if required.

Task: BABEL-2890
Signed-off-by: Dipesh Dhameliya (dddhamel@amazon.com)

### Check List


contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
```diff
@@ -1719,7 +1720,16 @@ PrepareRowDescription(TupleDesc typeinfo, List *targetlist, int16 *formats,
 
 				/* fetch the relation name, schema name */
 				relMetaDataInfo->partName[0] = RelationGetRelationName(rel);
-				relMetaDataInfo->partName[1] = get_namespace_name(RelationGetNamespace(rel));
+				physical_schema_name = get_namespace_name(RelationGetNamespace(rel));
+
+				/* logical schema name should be sent as part of response as far as Babelfish is concerned. */
+				if (pltsql_plugin_handler_ptr && 
+					pltsql_plugin_handler_ptr->pltsql_get_logical_schema_name)
+					/* Its very unlikely that logical schema name could not be found for any other cases. */
```
surendravishnoi: is it possible that we do not find the logical schema name? what would be the behavior in that case?

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
```diff
@@ -1719,7 +1720,16 @@ PrepareRowDescription(TupleDesc typeinfo, List *targetlist, int16 *formats,
...
+					relMetaDataInfo->partName[1] = pltsql_plugin_handler_ptr->pltsql_get_logical_schema_name(physical_schema_name);
+
+				if (physical_schema_name)
```
surendravishnoi: Is it safe to delete this? it is possible that partName[1] is pointing to this schema name (for sys and information schema)

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
Deepesh125: We should always get logical schema name given physical schema because TDS receives information of physical schema from engine. Raising an error would be the best choice if we do not get logical schema for some physical schema which we are already doing.

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
Deepesh125: Yes, it shouldn't get freed. Updated this in next revision.

KushaalShroff (approved):

surendravishnoi:
comment incorrectly talks about two rendezvous variables, code introduces only one

surendravishnoi (approved):

PR #134, opened 2022-04-22
Modify sys.indexes view

ghost:
Signed-off-by: Avneet Toor <avneett@bitquilltech.com>

### Description

Because part of the work of `sys.indexes` was changing column data types, extensive work was done adding `sys.indexes` to the `babelfishpg_tsql--1.2.0--1.3.0.sql` upgrade script. However, PostgreSQL does not allow column data types to be changed in views once they exist, so any addition of updated views to the upgrade script where data types are changed returned an error. The steps I took to mitigate this was:
- in the upgrade script rename `sys.indexes` to `sys.indexes_deprecated`
- add the updated definition of `sys.indexes`
- add the definition of any views/functions that use `sys.indexes` as a dependency to rebuild them to use the updated `sys.indexes` definition. 
- if `sys.indexes_deprecated` has no other dependencies (by, say, a customer script) drop the view. Else, keep it for now. 

To aid in the above process I added 2 new helper functions:
- `sys.has_dependencies` : Returns True if an object is dependant upon by other objects. False otherwise.
- `sys.drop_view`: If a view is not depended upon by other objects, drop the view. Else, leave it. 
 
### Issues Resolved
BABELFISH-343
BABEL-2712

### Check List


contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--1.2.0--1.3.0.sql
```diff
@@ -898,21 +898,221 @@ LANGUAGE SQL IMMUTABLE PARALLEL RESTRICTED;
 -- with different input/output data types. The second argument will be ignored as its functionality is not supported
```
kuntalghosh: The changes should be moved to 2.0.0-2.1.0 upgrade script.

ghost: Thank you, done. Now I need a maintainer to approve running workflows & 2 people to approve the PR.

contrib/babelfishpg_tsql/sql/sys_function_helpers.sql
```diff
@@ -10126,4 +10126,33 @@ AS 'babelfishpg_tsql', 'pltsql_get_last_stmt_handle' LANGUAGE C;
 
 CREATE OR REPLACE FUNCTION sys.get_babel_server_collation_oid() RETURNS OID
 LANGUAGE C
-AS 'babelfishpg_tsql', 'get_server_collation_oid';
\ No newline at end of file
+AS 'babelfishpg_tsql', 'get_server_collation_oid';
+
+-- Returns True if an object is dependant upon by other objects. False otherwise.
```
nasbyj: s/by //

contrib/babelfishpg_tsql/sql/sys_function_helpers.sql
```diff
@@ -10126,4 +10126,33 @@ AS 'babelfishpg_tsql', 'pltsql_get_last_stmt_handle' LANGUAGE C;
 
 CREATE OR REPLACE FUNCTION sys.get_babel_server_collation_oid() RETURNS OID
 LANGUAGE C
-AS 'babelfishpg_tsql', 'get_server_collation_oid';
\ No newline at end of file
+AS 'babelfishpg_tsql', 'get_server_collation_oid';
+
+-- Returns True if an object is dependant upon by other objects. False otherwise. 
+CREATE OR REPLACE FUNCTION sys.has_dependencies(schema_name varchar, object_name varchar) RETURNS bool AS
```
nasbyj: This function claims to be generic, but it is not. It will only work for views.

contrib/babelfishpg_tsql/sql/sys_function_helpers.sql
```diff
@@ -10126,4 +10126,33 @@ AS 'babelfishpg_tsql', 'pltsql_get_last_stmt_handle' LANGUAGE C;
...
+    JOIN pg_class as source_table ON pg_depend.refobjid = source_table.oid 
+    JOIN pg_attribute ON pg_depend.refobjid = pg_attribute.attrelid AND pg_depend.refobjsubid = pg_attribute.attnum 
+    JOIN pg_namespace dependent_ns ON dependent_ns.oid = dependent_view.relnamespace 
+    JOIN pg_namespace source_ns ON source_ns.oid = source_table.relnamespace 
+    WHERE source_ns.nspname = schema_name AND source_table.relname = object_name);
+$BODY$
+LANGUAGE SQL VOLATILE STRICT;
+
+-- If a view is not depended upon by other objects, drop the view. Else, leave it. 
+CREATE OR REPLACE FUNCTION sys.drop_view(schema_name varchar, object_name varchar) RETURNS void AS
```
nasbyj: Function name sounds like it simply drops a view, which is not at all what it does.

nasbyj (changes requested):
This is creating permanent catalog functions (aka *new product features*) to deal with a single one-off case. IMO that's not an appropriate way to handle this scenario. We should handle this via a simple plpgsql DO block that traps the exception we get from DROP VIEW if there are dependent objects. There's no need to dig into pg_depends.

Additionally, if we have to leave behind the deprecated view then we should disassociate it from the extension, as well as do a `RAISE WARNING` explaining to the user what's happening, and that it is their responsibility to drop the view when no longer needed. We should possibly move the view out of the sys schema as well.

nasbyj:
One more comment: I absolutely do not think handling user objects that depend on catalog objects is a requirement. The Postgres community is very cautious about changing the definition of catalog objects, but when they feel it's warranted they simply change them and let the user deal with it.

contrib/babelfishpg_tsql/sql/sys_function_helpers.sql
ghost: As discussed offline, removed these permanent catalog functions in favour of creating a temporary procedure that serves the same purpose in the upgrade script which will then be dropped at the end.

contrib/babelfishpg_tsql/sql/sys_function_helpers.sql
ghost: As discussed offline, removed these permanent catalog functions in favour of creating a temporary procedure that serves the same purpose in the upgrade script which will then be dropped at the end. Hopefully the procedure name & description is more clear in the upgrade script.

contrib/babelfishpg_tsql/sql/sys_function_helpers.sql
ghost: As discussed offline, removed these permanent catalog functions in favour of creating a temporary procedure that serves the same purpose in the upgrade script which will then be dropped at the end. Hopefully the procedure name & description is more clear in the upgrade script.

contrib/babelfishpg_tsql/sql/sys_views.sql
```diff
@@ -748,32 +748,86 @@ and has_sequence_privilege(pg_get_serial_sequence(quote_ident(ext.nspname)||'.'|
...
+    , 0 as is_hypothetical
+    , 1 as allow_row_locks
+    , 1 as allow_page_locks
+    , 0 as has_filter
+    , null as filter_definition
+    , 0 as auto_created
+    , 0 as index_id
+  from pg_class t 
+  where t.relkind = 'r'
+  and (t.relnamespace in (select schema_id from sys.schemas) or t.relnamespace::regnamespace::text = 'sys')
```
nasbyj, line 826: We shouldn't be doing that casting dance unless we have a really good reason to do so (and if so it needs to be documented)

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.0.0--2.1.0.sql
```diff
@@ -3,6 +3,31 @@
...
+    query2 := format('drop view %s.%s', schema_name, view_name);
+    execute query1;
+    execute query2;
+EXCEPTION
+    when object_not_in_prerequisite_state then --if 'alter extension' statement fails
+        GET STACKED DIAGNOSTICS error_msg = MESSAGE_TEXT;
+        raise notice '%', error_msg;
+    when dependent_objects_still_exist then --if 'drop view' statement fails
+        GET STACKED DIAGNOSTICS error_msg = MESSAGE_TEXT;
+        raise notice '%', error_msg;
```
nasbyj: IMO this should be a WARNING

contrib/babelfishpg_tsql/sql/sys_views.sql
```diff
@@ -748,32 +748,86 @@ and has_sequence_privilege(pg_get_serial_sequence(quote_ident(ext.nspname)||'.'|
...
+    , 1 as allow_page_locks
+    , 0 as has_filter
+    , null as filter_definition
+    , 0 as auto_created
+    , case when i.indisclustered then 1 else c.oid end as index_id
+  from pg_class c
+  inner join pg_index i on i.indexrelid = c.oid
+  left join pg_constraint constr on constr.conindid = c.oid
+  where c.relkind = 'i' and i.indislive
+  and (c.relnamespace in (select schema_id from sys.schemas) or c.relnamespace::regnamespace::text = 'sys')
```
nasbyj, line 798: We shouldn't be doing that regnamespace casting dance unless we have a really good reason to do so (and if so it needs to be documented)

nasbyj (changes requested):

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.0.0--2.1.0.sql
ghost: Done.

contrib/babelfishpg_tsql/sql/sys_views.sql
ghost: I have created an issue to address this in the future here: https://github.com/babelfish-for-postgresql/babelfish_extensions/issues/190

This will take some time to change & properly test but I want to unblock other developers who require some of the additions in this PR ASAP. 

The view still works, and from the customer's perspective they won't see a difference if casting is in the implementation or not, but I do agree a change is in order and this casting stuff isn't great.

contrib/babelfishpg_tsql/sql/sys_views.sql
ghost: I have created an issue to address this in the future here: https://github.com/babelfish-for-postgresql/babelfish_extensions/issues/190

This will take some time to change & properly test but I want to unblock other developers who require some of the additions in this PR ASAP. 

The view still works, and from the customer's perspective they won't see a difference if casting is in the implementation or not, but I do agree a change is in order and this casting stuff isn't great.

nasbyj: Agreed, but please note the comment I just added to #190

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.0.0--2.1.0.sql
```diff
@@ -3,6 +3,31 @@
  
 SELECT set_config('search_path', 'sys, '||current_setting('search_path'), false);
 
+-- Drops a view if it does not have any dependent objects.
+-- Is a temporary procedure for use by the upgrade script. Will be dropped at the end of the upgrade.
+-- Please have this be one of the first statements executed in this upgrade script. 
+CREATE OR REPLACE PROCEDURE babelfish_drop_deprecated_view(schema_name varchar, view_name varchar) AS
```
nasbyj, line 9: Just for reference sake...

I'm very tempted to say we should define this function in pg_temp, so that we ensure it can't be permanent. However, there's one significant gotcha to doing that: the extension will be marked as depending on this function, so if the function is sitting in pg_temp when the session disconnects, the forcible removal will actually drop the extension. Ultimately I wish there was a better way to safeguard/ensure that this function will actually be dropped at the end of the script...

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.0.0--2.1.0.sql
```diff
@@ -3,6 +3,31 @@
...
+$$
+DECLARE
+    error_msg text;
+    query1 text;
```
nasbyj, line 13: FWIW there's really no need for `query1` or `query2`; we could simply do `EXECUTE format($$....$$, schema_name, view_name)`.

nasbyj (dismissed):

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.0.0--2.1.0.sql
ghost: Ack. Good to know.

contrib/babelfishpg_tsql/sql/sys_views.sql
ghost: Ack

thephantomthief (approved):

KushaalShroff (approved):

PR #138, opened 2022-04-25
Support storing TSQL view definition in Babelfish catalog

HarshLunagariya:
Previously, TSQL view definitions were not supported because internally
definitions didn't get stored in the catalog, they get de-parsed at runtime in
pg dialect from tuple and query tree data.

This commit adds a Babelfish catalog babelfish_object_def for storing object
definitions along with db, schema, object name, flags of settings like ansi nulls,
quoted identifier. It implements pltsql_store_view_definition_hook function to 
store TSQL view definition along with flags into catalog. pltsql_drop_view_definition
function is implemented to drop the TSQL definition from catalog when
object_access_hook is invoked on a drop object event for TSQL
view.

This commit blocks DDLs for TSQL view objects which has entry in catalog from
PG dialect by introducing enable_create_alter_view_from_pg GUC which will
block DDL statements from PG dialect. enable_create_alter_view_from_pg GUC's
value will be false by default i.e. DDL's will be blocked. Currently it is only blocking
DDLs for view objects from PG endpoint. It is required to set this GUC to true in
upgrade script and while creating any internal views.

It implements information_schema_tsql.views such that it returns data related to
every view which is accessible to current user.

Task: BABEL-3135,2889
Signed-off-by: Harsh Lunagariya [lunharsh@amazon.com](mailto:lunharsh@amazon.com)

### Check List


contrib/babelfishpg_tsql/sql/object_definition_tsql.sql
```diff
@@ -0,0 +1,8 @@
+CREATE TABLE sys.babelfish_view_def (
```
2jungkook, line 1: I would suggest this table to have only oid (pkey) and view definition columns. Otherwise, we need additional special handlings for `ALTER DATABASE` and `ALTER SCHEMA`.

contrib/babelfishpg_tsql/sql/object_definition_tsql.sql
```diff
@@ -0,0 +1,8 @@
...
+	db_name NAME NOT NULL,
+	schema_name NAME NOT NULL,
+	object_name NAME NOT NULL,
+	definition TEXT NOT NULL COLLATE "C",
```
2jungkook: Let's use `nvarchar(max)` to align with the definition of `sys.sql_modules` (https://github.com/MicrosoftDocs/sql-docs/blob/live/docs/relational-databases/system-catalog-views/sys-all-sql-modules-transact-sql.md). Btw, Is there any reason to use C collation here?

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--1.2.0--1.3.0.sql
```diff
@@ -914,5 +914,14 @@ AS $$
...
  
+CREATE TABLE sys.babelfish_view_def (
+	db_name NAME NOT NULL,
+	schema_name NAME NOT NULL,
+	object_name sys.NVARCHAR(128) NOT NULL,
+	definition TEXT NOT NULL COLLATE "C",
+	PRIMARY KEY(db_name, schema_name, object_name)
+);
+GRANT SELECT ON sys.babelfish_view_def TO PUBLIC;
+
```
2jungkook: What is our policy for views created before this change? Are we going to keep those entries with null view definition or just to ignore them?

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -1454,3 +1467,125 @@ modify_insert_stmt(InsertStmt *stmt, Oid relid)
 	table_close(pg_attribute, AccessShareLock);
 
 }
+
+/*
+ * Stores view object's TSQL definition to bbf_view_def catalog
+ */
+static void
+pltsql_store_view_definition(const char *queryString, ObjectAddress address)
```
2jungkook, line 1540: There can be two corner cases:
1. Creating a view in PG connection with sql_dialect = T-SQL
2. `ALTER VIEW ... ALTER COLUMN` in PG connection

contrib/babelfishpg_tsql/sql/object_definition_tsql.sql
HarshLunagariya: The problem with having oid as primary key is that oid might get changed during Major Version Upgrade/Pg dump-restore, in that case we've to figure out the oid and update the catalog which will be more complex. So it is better to have combined key of db, schema and object name and handle the cases for `ALTER DATABASE` and `ALTER SCHEMA`.

contrib/babelfishpg_tsql/src/hooks.c
HarshLunagariya: 1. First case is handled correctly by current implementation.
2. For `ALTER VIEW ... ALTER COLUMN`, We shouldn't allow user to alter TSQL views from pg endpoint. For that I'll add the checks in engine.

Will also add required test cases.

contrib/babelfishpg_tsql/sql/object_definition_tsql.sql
2jungkook: Okay, that makes sense. I'm not sure whether all those functionalities are in the current BBF. If not, let's add at least test cases to notify future developers :) FYI, object name may also be changed by `sp_rename`.

If we are storing those names, I'd suggest to avoid using pg-only types because it can cause different result for comparisons. I guess T-SQL has their own types for db/schema/object names.

contrib/babelfishpg_tsql/src/hooks.c
2jungkook: If we execute T-SQL in a PG connection, it won't go through ANTLR parser so the result can be different, e.g., `select 'value' = '1';`. Also I just realized that we modify query strings in both ANTLR and parse_analyze stages. Could you check whether we store the original query or modified one? For example,
- `select * from db..object` is changed to `select * from db.dbo.object`
- `select * from dbo.object` is changed to `select * from master_dbo.object`

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--1.2.0--1.3.0.sql
KushaalShroff: For now, there wont be any entires for views prior to this change. We can ask the User to either drop and re-create the view they are blocked on or we can provide some other way to insert their view definition into this table (The first workaround seems to be the better one)

contrib/babelfishpg_tsql/sql/object_definition_tsql.sql
KushaalShroff: Even if we use nvarchar(max) we might need to cast it to different string types of different lengths while using it, since sys.sql_modules is not the only system view which stores the definitions

2jungkook: I'm afraid of using pg-only types (NAME, TEXT) in BBF catalogs because they behave differently, e.g., ANSI_PADDINGS. And would you point me to some examples where different system views use different data types for the object definitions or names?

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--1.2.0--1.3.0.sql
```diff
@@ -914,5 +914,14 @@ AS $$
 $$
 LANGUAGE SQL IMMUTABLE PARALLEL RESTRICTED;
```
surendravishnoi: this file is deleted in git farm

contrib/babelfishpg_tsql/sql/object_definition_tsql.sql
```diff
@@ -0,0 +1,8 @@
+CREATE TABLE sys.babelfish_view_def (
+	dbid SMALLINT NOT NULL,
+	schema_name NAME NOT NULL,
+	object_name NAME NOT NULL,
+	definition TEXT NOT NULL COLLATE "C",
```
sangilsong: Don't we need to capture session property (i.e. ANSI_NULLS)? In my short experiment, select on SQL view follows the setting at when it is created.

```
1> create table tt1(a varchar(10));
2> go
1> insert into tt1 values (null);
2> go
1> set ansi_nulls off;
2> go
1> select * from tt1 where a = null;
2> go
a
----------
NULL

1> create view v11 as select * from tt1 where a = null;
2> go
1> select * from v11;
2> go
a
----------
NULL

-- v11 behaves like ansi_nulls=off
1> set ansi_nulls on;
2> go
1> select * from v11;
2> go
a
----------
NULL

(1 rows affected)
1> select * from tt1 where a = null;
2> go
a
----------

(0 rows affected)
```

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -1290,6 +1300,9 @@ bbf_object_access_hook(ObjectAccessType access, Oid classId, Oid objectId, int s
 	if (prev_object_access_hook)
 		(*prev_object_access_hook) (access, classId, objectId, subId, arg);
 
+	if (access == OAT_DROP && classId == RelationRelationId)
+		pltsql_drop_view_definition(objectId);
```
sangilsong, line 1367: is this hook called when base object which a T-SQL view depends on is dropped cascade?

HarshLunagariya: yes. it will be called.
Next question might be then why `clean_up_bbf_view_def()` is needed?
Reason: 
- When we drop logical db in babelfish, `drop_bbf_db()` function gets executed. 
- In that function we generate subcommands to drop child objects.
- At that time this hook is called but in that case we're unable to lookup babelfish_namespace catalog to get logical schema name. So need to explicitly clean the catalog in case of `drop_bbf_db()`

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -774,6 +830,109 @@ get_authid_user_ext_db_users(const char *db_name)
...
+							  bbf_view_def_idx_oid ,
+							  true, NULL, 3, scanKey);
+
+	while ((scantup = systable_getnext(scan)) != NULL)
+	{
+		/* Raise error if multiple matches */
+		if (oldtup)
+			elog(ERROR,
+				 "multiple babelfish_view_def entries for object %u",
+				 relid);
```
xhfanhe: Under what circumstances will there be multiple entries for single object?

.github/workflows/minor-version-upgrade.yml
```diff
@@ -10,9 +10,9 @@ on:
 jobs:
   extension-tests:
     env:
-      ENGINE_VER_FROM: BABEL_2_X_DEV__PG_14_2
+      ENGINE_VER_FROM: jira-babel-3135
       EXTENSION_VER_FROM: BABEL_2_0_STABLE
-      ENGINE_VER_TO: BABEL_2_X_DEV__PG_14_2
+      ENGINE_VER_TO: jira-babel-3135
```
xhfanhe: Please make sure to reset yml changes before final merge

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -774,6 +830,109 @@ get_authid_user_ext_db_users(const char *db_name)
...
+		oldtup = heap_copytuple(scantup);
+	}
+
+	if (HeapTupleIsValid(oldtup))
+		is_tsql_view = true;
+
+	heap_freetuple(oldtup);
```
xhfanhe: Do we need to free scantup here as well?

xhfanhe (changes requested):
Left some comments, please take a look.
Also, with this change, it would be safer to add more rules in metadata integrity check, in which way we can check if babelfish_view_def catalog is working correctly. Please consider filing a separate Jira for that.

contrib/babelfishpg_tsql/sql/object_definition_tsql.sql
HarshLunagariya: Yes, we should add session properties like ANSI_NULLS, QUOTED_IDENTIFIER. I will address this in upcoming commits.

contrib/babelfishpg_tsql/src/catalog.c
HarshLunagariya: Yes, It's redundant because we're defining primary key on dbid, schema name and object name. I will address it in next commits.

contrib/babelfishpg_tsql/src/catalog.c
HarshLunagariya: Documentation for `systable_getnext(SysScanDesc sysscan)` function goes like:
`Note that returned tuple is a reference to data in a disk buffer; it must not be modified, and should be presumed inaccessible after next getnext() or endscan() call.` 
So, We don't need to free scantup.

test/JDBC/input/bbf_view_def.mix
```diff
@@ -0,0 +1,168 @@
+-- tsql
+create table vd_t1 (c1 int primary key);
```
shalinilohia50, line 2: Please drop this table at the end. It leads to one failure in JDBC.

HarshLunagariya: Done.

contrib/babelfishpg_tsql/sql/information_schema_tsql.sql
```diff
@@ -401,4 +401,41 @@ CREATE VIEW information_schema_tsql.table_constraints AS
...
+
+CREATE OR REPLACE VIEW information_schema_tsql.views AS
+	SELECT CAST(nc.dbname AS sys.nvarchar(128)) AS "TABLE_CATALOG",
+			CAST(ext.orig_name AS sys.nvarchar(128)) AS  "TABLE_SCHEMA",
+			CAST(c.relname AS sys.nvarchar(128)) AS "TABLE_NAME",
+
+			CAST(
+				CASE WHEN LENGTH(vd.definition) <= 4000
+					THEN vd.definition
+					ELSE NULL END
```
surendravishnoi, line 480: do we want to return NUL or truncated value?

contrib/babelfishpg_tsql/sql/object_definition_tsql.sql
```diff
@@ -0,0 +1,14 @@
+CREATE TABLE sys.babelfish_view_def (
```
surendravishnoi, line 1: not for this change but I think we should device a way to create such tables in binary itself. That will save a lot of trouble for upgrades in future

shalinilohia50 (dismissed):

contrib/babelfishpg_tsql/src/catalog.h
```diff
@@ -89,6 +89,7 @@ extern const char *get_one_user_db_name(void);
 #define NAMESPACE_EXT_TABLE_NAME "babelfish_namespace_ext"
 #define NAMESAPCE_EXT_PK_NAME "babelfish_namespace_ext_pkey"
 #define Anum_namespace_ext_namespace 1
+#define Anum_namespace_ext_dbid 2
```
surendravishnoi, line 92: please make sure that we never used this id in previous releases

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -1454,3 +1468,145 @@ modify_insert_stmt(InsertStmt *stmt, Oid relid)
...
+	ReleaseSysCache(reltup);
+	heap_freetuple(tuple);
+	table_close(bbf_view_def_rel, RowExclusiveLock);
+}
+
+/*
+ * Drops view object's TSQL definition from bbf_view_def catalog
+ */
+static void
+pltsql_drop_view_definition(Oid objectId)
```
surendravishnoi, line 1636: it can share code with it_tsql view

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -2045,6 +2045,81 @@ static void bbf_ProcessUtility(PlannedStmt *pstmt,
...
+		{
+			case T_ViewStmt:
+			{
+				ViewStmt *vstmt = (ViewStmt *) parsetree;
+				Oid relid;
+				relid = RangeVarGetRelid(vstmt->view, NoLock, true);
+				if (vstmt->replace && check_is_tsql_view(relid))
+				{
+					ereport(ERROR,
+							(errcode(ERRCODE_INTERNAL_ERROR),
```
surendravishnoi, line 2069: if it is triggered by user action, it should not be an internal error. Rather not supported or something like that

contrib/babelfishpg_tsql/sql/information_schema_tsql.sql
HarshLunagariya: Yes, We want to return NULL and it is expected . See [Documentation](https://github.com/MicrosoftDocs/sql-docs/blob/live/docs/relational-databases/system-information-schema-views/views-transact-sql.md)

contrib/babelfishpg_tsql/src/catalog.h
HarshLunagariya: Double Checked. We haven't used it yet.

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -413,6 +424,51 @@ get_logical_schema_name(const char *physical_schema_name, bool missingOk)
...
+	SysScanDesc	scan;
+	Datum		datum;
+	int16		dbid;
+	TupleDesc	dsc;
+	bool		isnull;
+
+	if (get_namespace_oid(physical_schema_name, false) == InvalidOid)
+		return NULL;
+
+	rel = table_open(namespace_ext_oid, AccessShareLock);
```
Deepesh125, line 451: why access share lock here? shouldn't we use row exclusive lock?

shalinilohia50 (dismissed):

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -774,6 +830,118 @@ get_authid_user_ext_db_users(const char *db_name)
...
+							  true, NULL, 3, scanKey);
+
+	scantup = systable_getnext(scan);
+	oldtup = heap_copytuple(scantup);
+	systable_endscan(scan);
+	return oldtup;
+}
+
+bool
+check_is_tsql_view(Oid relid)
```
2jungkook, line 998: Could you add a comment explaining that this function will return false if a T-SQL view was created before v2.1?

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -1454,3 +1468,147 @@ modify_insert_stmt(InsertStmt *stmt, Oid relid)
...
+	 * Inserting false/0 as default value for is_schema_bound and
+	 * uses_database_collation until they're supported
+	 */
+	new_record[6] = BoolGetDatum(0);
+	new_record[7] = BoolGetDatum(0);
+
+	tuple = heap_form_tuple(bbf_view_def_rel_dsc,
+							new_record, new_record_nulls);
+
+	CatalogTupleInsert(bbf_view_def_rel, tuple);
```
2jungkook, line 1623: Can this support rollback? What if an exception happens after pltsql_store_view_definition()?

test/JDBC/expected/bbf_view_def.out
```diff
@@ -0,0 +1,308 @@
...
+
+
+
+-- TODO: Add support when alter view is supported
+alter view def_v1 as select 2;
+go
+~~ERROR (Code: 33557097)~~
+
+~~ERROR (Message: 'ALTER VIEW' is not currently supported in Babelfish)~~
+
```
2jungkook, line 373: Let's add a select statement here as well because once we support alter view then the result of the select statement should be changed.

contrib/babelfishpg_tsql/src/hooks.c
HarshLunagariya: Yes, it supports rollback, I will add test case for the same. And I checked using inserting an exception after pltsql_store_view_definition(), it is rolling back as expected.

contrib/babelfishpg_tsql/src/catalog.c
HarshLunagariya: No need for row exclusive lock as we use this function only for getting dbid using namespace_ext_oid relation. If you're considering DROP DB/SCHEMA case then we must be holding RowExclusiveLock while dropping.

contrib/babelfishpg_tsql/sql/information_schema_tsql.sql
```diff
@@ -410,6 +410,43 @@ CREATE VIEW information_schema_tsql.table_constraints AS
...
+					THEN 'CASCADE'
+					ELSE 'NONE' END
+				AS sys.varchar(7)) AS "CHECK_OPTION",
+
+			CAST('NO' AS sys.varchar(2)) AS "IS_UPDATABLE"
+
+	FROM sys.pg_namespace_ext nc JOIN pg_class c ON (nc.oid = c.relnamespace)
+		LEFT OUTER JOIN sys.babelfish_namespace_ext ext ON nc.nspname = ext.nspname
+		LEFT OUTER JOIN sys.babelfish_view_def vd
+			ON ext.dbid = vd.dbid AND ext.orig_name = vd.schema_name AND c.relname = vd.object_name
```
KushaalShroff: We might need to add explicit collation for string comparison, this will start failing after Dipeshs changes go in

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -124,6 +130,10 @@ Datum init_catalog(PG_FUNCTION_ARGS)
 	bbf_authid_user_ext_idx_oid = get_relname_relid(BBF_AUTHID_USER_EXT_IDX_NAME,
 													sys_schema_oid);
 
+	/* bbf_view_def */
+	bbf_view_def_oid = get_relname_relid(bbf_view_def_TABLE_NAME, sys_schema_oid);
+	bbf_view_def_idx_oid = get_relname_relid(bbf_view_def_IDX_NAME, sys_schema_oid);
```
KushaalShroff: Let's keep the naming consistent, i.e. ALL CAPS

contrib/babelfishpg_tsql/src/dbcmds.c
```diff
@@ -622,16 +624,21 @@ PG_FUNCTION_INFO_V1(create_builtin_dbs);
 Datum create_builtin_dbs(PG_FUNCTION_ARGS)
 {	
 	const char  *sql_dialect_value_old;
+	const char	*enable_ddl_from_pgendpoint_old;
 	const char  *tsql_dialect = "tsql";
 	const char  *sa_name = text_to_cstring(PG_GETARG_TEXT_PP(0));
 
 	sql_dialect_value_old = GetConfigOption("babelfishpg_tsql.sql_dialect", true, true);
+	enable_ddl_from_pgendpoint_old = GetConfigOption(
```
KushaalShroff: If we have an extern variable pltsql_enable_ddl_from_pgendpoint in TSQL
Why do we need to use get/set config option?
IIRC then this is performance draining.

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -1510,6 +1523,150 @@ modify_insert_stmt(InsertStmt *stmt, Oid relid)
...
+		elog(ERROR,
+				"Could not find physical schemaname for %u",
+				 form_reltup->relnamespace);
+	}
+
+	/*
+	 * Do not store definition/data in case of sys and information_schema_tsql
+	 * views
+	 */
+	if (strcmp(physical_schemaname, "sys") == 0 ||
```
KushaalShroff: We can re-use is_shared_schema function which is currently a static function

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -1510,6 +1523,150 @@ modify_insert_stmt(InsertStmt *stmt, Oid relid)
...
+{
+	/* Store TSQL definition */
+	Relation	bbf_view_def_rel;
+	TupleDesc	bbf_view_def_rel_dsc;
+	Datum		new_record[bbf_view_def_NUM_COLS];
+	bool		new_record_nulls[bbf_view_def_NUM_COLS];
+	HeapTuple	tuple, reltup;
+	Form_pg_class	form_reltup;
+	int16		dbid;
+	char		*physical_schemaname, *logical_schemaname;
```
KushaalShroff, line 1551: Free these pointers

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -1510,6 +1523,150 @@ modify_insert_stmt(InsertStmt *stmt, Oid relid)
...
+ * Drops view object's TSQL definition from bbf_view_def catalog
+ */
+static void
+pltsql_drop_view_definition(Oid objectId)
+{
+	Relation	bbf_view_def_rel;
+	HeapTuple	reltuple, scantup;
+	Form_pg_class	form;
+	int16		dbid;
+	char		*physical_schemaname, *logical_schemaname, *objectname;
```
KushaalShroff, line 1642: Free these pointers

test/JDBC/expected/13_4__preparation__bbf_view_def_upgrade-vu-prepare.out
```diff
@@ -0,0 +1,13 @@
...
+GO
+
+CREATE VIEW dep_view_3135 AS SELECT * FROM sys.babelfish_view_def;
+GO
+~~ERROR (Code: 33557097)~~
+
+~~ERROR (Message: relation "sys.babelfish_view_def" does not exist)~~
+
+
+CREATE VIEW dep_view_3135_2 AS SELECT * FROM information_schema.views;
```
KushaalShroff: Can we please add comments for this and what is expected (If fail, then why?)

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -1510,6 +1523,150 @@ modify_insert_stmt(InsertStmt *stmt, Oid relid)
 
 }
 
+/*
+ * Stores view object's TSQL definition to bbf_view_def catalog
+ */
+static void
+pltsql_store_view_definition(const char *queryString, ObjectAddress address)
```
KushaalShroff, line 1540: What happens when we run into error during this function? Will the view be created?

.github/workflows/minor-version-upgrade.yml
```diff
@@ -4,12 +4,12 @@ on: [push, pull_request]
 jobs:
   extension-tests:
     env:
-      ENGINE_VER_FROM: BABEL_2_X_DEV_DEFAULT_mVU_SOURCE
+      ENGINE_VER_FROM: jira-babel-3135
```
surendravishnoi: It is working for others without any change, why did you need this? May be earlier it was an issue but not for a week or so

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -875,6 +931,119 @@ get_user_for_database(const char *db_name)
...
+{
+	Oid		schema_oid;
+	Relation	bbf_view_def_rel;
+	HeapTuple	scantup;
+	char		*view_name, *schema_name;
+	int16		logical_dbid;
+	const char	*logical_schema_name;
+	bool		is_tsql_view = false;
+
+	view_name = get_rel_name(relid);
```
surendravishnoi, line 1008: who should free this memory?

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -875,6 +931,119 @@ get_user_for_database(const char *db_name)
...
+		return false;
+	}
+	logical_schema_name = get_logical_schema_name(schema_name, true);
+	logical_dbid = get_dbid_from_physical_schema_name(schema_name, true);
+	if (logical_schema_name == NULL || !DbidIsValid(logical_dbid))
+	{
+		return false;
+	}
+	/* Fetch the relation */
+	bbf_view_def_rel = table_open(bbf_view_def_oid, RowExclusiveLock);
```
surendravishnoi: why exclusive lock?

contrib/babelfishpg_tsql/sql/information_schema_tsql.sql
Deepesh125: Specifying explicit collation is no required here since we are joining sys objects only so collation conflict should not be occurred ideally. But it is always advisable and good practice to specify explicit collate clause.

contrib/babelfishpg_tsql/src/catalog.c
HarshLunagariya: Fixed.

contrib/babelfishpg_tsql/src/dbcmds.c
HarshLunagariya: Got fixed indirectly, Now we're setting up GUC while initialising Babelfish.

contrib/babelfishpg_tsql/src/hooks.c
HarshLunagariya: Done.

contrib/babelfishpg_tsql/src/pl_handler.c
HarshLunagariya: Using ERRCODE_FEATURE_NOT_SUPPORTED now. Fixed.

contrib/babelfishpg_tsql/sql/object_definition_tsql.sql
HarshLunagariya: Fixed. Used TSQL types in catalog definition.

test/JDBC/expected/13_4__preparation__bbf_view_def_upgrade-vu-prepare.out
HarshLunagariya: Added.

contrib/babelfishpg_tsql/src/hooks.c
HarshLunagariya: Done.

contrib/babelfishpg_tsql/src/hooks.c
HarshLunagariya: Done. No need to free objectname as it is referred from tuple.

contrib/babelfishpg_tsql/src/hooks.c
HarshLunagariya: Currently, We've blocked DDL from pg endpoint using GUC.

test/JDBC/expected/13_4__preparation__bbf_view_def_upgrade-vu-prepare.out
HarshLunagariya: Added.

contrib/babelfishpg_tsql/sql/object_definition_tsql.sql
```diff
@@ -0,0 +1,14 @@
...
+	dbid SMALLINT NOT NULL,
+	schema_name sys.SYSNAME NOT NULL,
+	object_name sys.SYSNAME NOT NULL,
+	definition sys.NTEXT,
+	is_ansi_nulls_on sys.BIT,
```
surendravishnoi: what if we need more flags in future? will be tricky to upgrade existing customer. Will it make more sense to have a 64 bit field so that we can cover up to 64 flags without an upgrade

test/JDBC/upgrade/13_4/preparation/bbf_view_def_upgrade-vu-prepare.sql
```diff
@@ -0,0 +1,10 @@
+CREATE VIEW view_3135_upd AS SELECT 1;
```
surendravishnoi: Use filename_xxx as object name at all places to allow unique object names

test/JDBC/upgrade/14_3/preparation/bbf_view_def_upgrade-vu-prepare.sql
```diff
@@ -0,0 +1,10 @@
+CREATE VIEW view_3135_upd AS SELECT 1;
```
surendravishnoi: It seems all prepare files are same. If that is the case, add single file to input directory and add corresponding entry to schedule file

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--1.0.0--1.1.0.sql
```diff
@@ -2,6 +2,8 @@
 \echo Use "ALTER EXTENSION ""babelfishpg_tsql"" UPDATE TO '1.1.0'" to load this file. \quit
 
 SELECT set_config('search_path', 'sys, '||current_setting('search_path'), false);
+-- enable DDL from pgendpoint
+SELECT set_config('babelfishpg_tsql.enable_ddl_from_pgendpoint', 'true', false);
```
surendravishnoi: can we set it for whole create/update extension duration to avoid set/reset at multiple places

test/JDBC/upgrade/13_5/preparation/bbf_view_def_upgrade-vu-prepare.sql
```diff
@@ -0,0 +1,10 @@
+CREATE VIEW view_3135_upd AS SELECT 1;
+GO
+
+CREATE VIEW dep_view_3135 AS SELECT * FROM sys.babelfish_view_def;
```
surendravishnoi: can we create a view on non-existent tables? this table is only available on 14 latest

contrib/babelfishpg_tsql/src/hooks.c
HarshLunagariya: View won't be created and error will be thrown.

test/JDBC/upgrade/14_3/preparation/bbf_view_def_upgrade-vu-prepare.sql
HarshLunagariya: But output are different for them that's why I had to add in version specific folder. If I'll add it to schedule file then it will compare it's output with common expected file (i.e. bbf_view_def_upgrade-vu-prepare.out) and fail.

test/JDBC/upgrade/13_5/preparation/bbf_view_def_upgrade-vu-prepare.sql
HarshLunagariya: No, it will fail. I will try to remove it in next iteration.

contrib/babelfishpg_tsql/sql/object_definition_tsql.sql
HarshLunagariya: Addressed introduced flag_validity(to track whether we're use corresponding bit in flag_values to represent any flag like ansi_nulls, quoted_identifier etc. related to view) and flag_values(to store actual values) int64 values in catalog.

contrib/babelfishpg_tsql/src/catalog.c
HarshLunagariya: It was unused function, missed to remove as part of refactoring. Removed now.

contrib/babelfishpg_tsql/src/catalog.c
HarshLunagariya: It was unused function, missed to remove as part of refactoring. Removed now.

contrib/babelfishpg_tsql/sql/information_schema_tsql.sql
HarshLunagariya: Fixed

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--1.0.0--1.1.0.sql
HarshLunagariya: For create extension, We are setting/resetting at babelfishpg_tsql.in . But for for upgrade we've to add it to every script.

.github/workflows/minor-version-upgrade.yml
HarshLunagariya: Fixed

test/JDBC/upgrade/13_4/preparation/bbf_view_def_upgrade-vu-prepare.sql
HarshLunagariya: Fixed

test/JDBC/upgrade/13_5/preparation/bbf_view_def_upgrade-vu-prepare.sql
HarshLunagariya: Fixed

surendravishnoi (dismissed):

surendravishnoi (dismissed):

surendravishnoi (approved):

PR #140, opened 2022-04-26
An error when executing select func leads to internal error "InstrStartNode called twice in a row"

surendravishnoi:
A select function call will run function inside a SELECT portal.
If the function throws an error, TSQL error handling will rollback
the internal PG transaction to support undo behavior. The rollback
will cleanup all active portals assuming PG semantic for transactions.

As portal is still active at higher level, this can result in
unpredictable behavior including above error and server crash.

We solve this issue by always running a TSQL function inside an internal
savepoint. We rely on the fact that TSQL functions cannot have side
effects like data write, transaction etc. In case of an error, rollback
to savepoint takes care of statement undo without effecting portals
active at higher level.

Task : BABEL-3101
Signed-off-by: Surendra Vishnoi <vishnosu@amazon.com>

### Description

### Issues Resolved

### Check List

### DIFF
--- a/contrib/babelfishpg_tsql/src/iterative_exec.c
+++ b/contrib/babelfishpg_tsql/src/iterative_exec.c
@@ -1104,6 +1104,9 @@ int dispatch_stmt_handle_error(PLtsql_execstate *estate,
 	SimpleEcontextStackEntry *volatile topEntry = simple_econtext_stack;
 	bool support_tsql_trans = pltsql_support_tsql_transactions();
 	uint32 before_tran_count = NestedTranCount;
+	bool ro_func = (estate->func->fn_prokind == PROKIND_FUNCTION) &&
+		(estate->func->fn_is_trigger == PLTSQL_NOT_TRIGGER) &&
+		(strcmp(estate->func->fn_signature, "inline_code_block") != 0);
 
 	PG_TRY();
 	{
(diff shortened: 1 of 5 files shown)


thephantomthief (approved):

contrib/babelfishpg_tsql/src/iterative_exec.c
```diff
@@ -1132,8 +1135,12 @@ int dispatch_stmt_handle_error(PLtsql_execstate *estate,
 		 * We do not start savepoint for batch commands as
 		 * error handling must be taken care of at statement
 		 * level
+		 * For RO functions start savepoint even when transaction
+		 * is not active to retain top level portals. A transaction
+		 * rollback will cleanup portal data which can lead to
+		 * problems when control returns back to portal level
 		 */
-		if ((!pltsql_disable_internal_savepoint && !is_batch_command(stmt) && IsTransactionBlockActive()))
+		if (!pltsql_disable_internal_savepoint && !is_batch_command(stmt) && (IsTransactionBlockActive() || ro_func))
```
2jungkook, line 1143: Just for curiosity, does it mean we start N internal save points if a select statement contains N functions?

2jungkook (approved):

contrib/babelfishpg_tsql/src/iterative_exec.c
surendravishnoi: if there are N function calls like N1, N2, N3 then it will become sp1 N1 relsp1 sp2 N2 relsp2 sp3 N3 relsp3. If function call are nested where N1 -> N2 -> N3, there will be 3 active savepoints

PR #149, opened 2022-04-28
Special argument handling in suser_ functions (#708)

thephantomthief:
This commit contains the following changes to handle special arguments in the suser_ functions so that their behavior matches SQL Server:

    Return NULL if NULL is passed as a parameter. The exception to this is the second argument passed to suser_sid, which is ignored if NULL in SQL Server.
    Ignore case for text arguments.
    Ignore trailing whitespace for text arguments.

Task: BABEL-661
Author: Aaron Congo aarcongo@amazon.com
Signed-off-by: Justin Jossick jusjosj@amazon.com

### Check List


contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.0.0--2.1.0.sql
```diff
@@ -835,6 +835,84 @@ END;
 $$
 LANGUAGE 'pltsql';
 GRANT ALL ON PROCEDURE sys.sp_sproc_columns_100 TO PUBLIC;
+
+CREATE OR REPLACE FUNCTION sys.suser_name_internal(IN server_user_id OID)
+RETURNS sys.NVARCHAR(128)
+AS 'babelfishpg_tsql', 'suser_name'
+LANGUAGE C IMMUTABLE PARALLEL RESTRICTED;
+
+CREATE OR REPLACE FUNCTION sys.suser_name(IN server_user_id OID)
```
kuntalghosh, line 844: This change will cause an upgrade failure. We can't remove the DEFAULT argument keyword without explicitly dropping the function. Here's an example,
```
CREATE OR REPLACE FUNCTION suser_name(IN server_user_id OID DEFAULT NULL)                                                                                                                                   RETURNS INT AS $$                                                                                                                                                                                                             SELECT 1                                                                                                                                                                                                                      END;                                                                                                                                                                                                                          $$                                                                                                                                                                                                                            LANGUAGE SQL IMMUTABLE PARALLEL RESTRICTED;
CREATE FUNCTION
Time: 1.050 ms
postgres[75672]=# CREATE OR REPLACE FUNCTION suser_name(IN server_user_id OID)                                                                                                                                                RETURNS INT AS $$                                                                                                                                                                                                             SELECT 1                                                                                                                                                                                                                      END;                                                                                                                                                                                                                          $$                                                                                                                                                                                                                            LANGUAGE SQL IMMUTABLE PARALLEL RESTRICTED;
ERROR:  42P13: cannot remove parameter defaults from existing function
HINT:  Use DROP FUNCTION suser_name(oid) first.
LOCATION:  ProcedureCreate, pg_proc.c:515
Time: 1.182 ms
```

PR #168, opened 2022-05-06
Handle transaction error handling issues for table variables and functions

surendravishnoi:
Avoid starting new transaction for insert into table variables

There are two issue which lead to the problem.
A select on a function internally creates a portal to hold the results.
The portal has its own resource owner which is used for all the
snapshots created during function execution. A transaction event inside
the function will reset resource owner to top transaction. Ideally, we
should reset back to portal resource owner at the earliest. This is not
happening right now as we do not want to do reset for cases where
transaction/savepoint was started before portal creation. We need to
distinguish between these two cases.
We start internal transactions for DDLs/DMLs but not for queries. We
were starting transactions for insert into table variables which we
should not. This internal transaction was resetting the resource owner
from portal to top transaction.
This change fixes the second problem. The first problem will be fixed as
a followup later.

---------

An error when executing select func leads to internal error "InstrStartNode called twice in a row"

A select function call will run function inside a SELECT portal.
If the function throws an error, TSQL error handling will rollback
the internal PG transaction to support undo behavior. The rollback
will cleanup all active portals assuming PG semantic for transactions.

As portal is still active at higher level, this can result in
unpredictable behavior including above error and server crash.

We solve this issue by always running a TSQL function inside an internal
savepoint. We rely on the fact that TSQL functions cannot have side
effects like data write, transaction etc. In case of an error, rollback
to savepoint takes care of statement undo without effecting portals
active at higher level.

--------

Handle missing SPI finish call for error cases

Every TSQL execution batch/proc/func call will create
a SPI connection and later close it using SPI finish.
However, SPI finish was missed for error cases and
can lead to in-correct SPI call stack for nested calls.
As a result, current call handler can call finish on
a leaked SPI connection of its children. This can cause
many issues including memory corruption.
As a fix, make sure that a call handler always does SPI
finish including error cases.
Also, did one small fix to make a copy of procedure name
info when unwinding a call stack. Otherwise, procedure name allocation
is no longer valid.

Task:Babel-3101,3108,3195
Signed-off-by: Surendra Vishnoi <vishnosu@amazon.com>

### Description

### Issues Resolved

### Check List

### DIFF
--- a/contrib/babelfishpg_tsql/src/iterative_exec.c
+++ b/contrib/babelfishpg_tsql/src/iterative_exec.c
@@ -1050,7 +1050,8 @@ void handle_error(PLtsql_execstate *estate,
 					PLtsql_stmt *stmt,
 					ErrorData *edata, 
 					SimpleEcontextStackEntry *volatile topEntry, 
-					bool *terminate_batch)
+					bool *terminate_batch,
+					bool ro_func)
 {
 	/* Determine if we want to override the transactional behaviour. */
 	uint8_t override_flag = override_txn_behaviour(stmt);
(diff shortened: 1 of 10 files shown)


contrib/babelfishpg_tsql/src/iterative_exec.c
```diff
@@ -1104,6 +1105,9 @@ int dispatch_stmt_handle_error(PLtsql_execstate *estate,
 	SimpleEcontextStackEntry *volatile topEntry = simple_econtext_stack;
 	bool support_tsql_trans = pltsql_support_tsql_transactions();
 	uint32 before_tran_count = NestedTranCount;
+	bool ro_func = (estate->func->fn_prokind == PROKIND_FUNCTION) &&
+		(estate->func->fn_is_trigger == PLTSQL_NOT_TRIGGER) &&
+		(strcmp(estate->func->fn_signature, "inline_code_block") != 0);
```
sangilsong, line 1110: nit: it would be better to check oid is InvalidOid rather comparing a fn_signature

sangilsong (approved):

xhfanhe (approved):

PR #175, opened 2022-05-11
sys.dm_exec_sessions view shows incorrect transaction isolation level boot value

thephantomthief:
### Description

Initially, we were storing the boot value of the
"default_transaction_isolation" GUC in shared memory by first fetching
the value using GetConfigOption() and later converting it into suitable
integer using atoi(). In this particular case, GetConfigOption() was
returning a string that cannot directly be converted as an integer by
atoi. As a result atoi() was always returning 0 for the boot GUC value.

This commit uses the DefaultXactIsoLevel to set the boot value in shared
memory.

### Check List

### DIFF
--- a/contrib/babelfishpg_tds/src/backend/tds/tds.c
+++ b/contrib/babelfishpg_tds/src/backend/tds/tds.c
@@ -137,14 +137,6 @@ typedef struct LocalTdsStatus
 	TransactionId backend_xmin;
 } LocalTdsStatus;
 
-static const struct config_enum_entry isolation_level_options[] = {
-	{"serializable", XACT_SERIALIZABLE, false},
-	{"repeatable read", XACT_REPEATABLE_READ, false},
-	{"read committed", XACT_READ_COMMITTED, false},
-	{"read uncommitted", XACT_READ_UNCOMMITTED, false},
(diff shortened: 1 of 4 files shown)


contrib/babelfishpg_tds/src/backend/tds/tds.c
```diff
@@ -395,7 +406,17 @@ tdsstat_bestart(void)
 	ltdsentry.textsize = atoi(GetConfigOption("babelfishpg_tsql.textsize", true, true));
 	ltdsentry.datefirst = atoi(GetConfigOption("babelfishpg_tsql.datefirst", true, true));
 	ltdsentry.lock_timeout = atoi(GetConfigOption("lock_timeout", true, true));
-	ltdsentry.transaction_isolation = atoi(GetConfigOption("default_transaction_isolation", true, true));
+
+	isolation_level_str = GetConfigOption("default_transaction_isolation", true, true);
```
surendravishnoi: can't we use DefaultXactIsoLevel here, that way we do not need to do all the conversion?

surendravishnoi (approved):

KushaalShroff (approved):

PR #178, opened 2022-05-11
NUMERIC overflow causes TDS error

bsatarupa:
### Description
NUMERIC overflow causes TDS error for aggregate functions

Before fix, expected resultant precision for aggregate function SUM() was always same as individual precision which caused numeric overflow error even when #digits in sum became 1 more than precision of individual operand.

The fix includes sending default numeric precision(i.e., 38) on wire whenever we encounter aggregate function SUM() and AVG(); and setting resultant scale = max(6, input scale) for aggregate function AVG()

Task: BABEL-3157
Signed-off-by: Satarupa Biswas <satarupb@amazon.com>

### Check List


surendravishnoi (dismissed):
Commit message can be more descriptive for both problem and solution part

kuntalghosh (approved):

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
```diff
@@ -660,11 +660,22 @@ resolve_numeric_typmod_from_exp(Node *expr)
 
 			/*
 			 * [BABEL-3074] NUMERIC overflow causes TDS error for aggregate
-			 * function sum(); resultant precision should be individual precision + 1
+			 * function sum(); resultant precision should be tds_default_numeric_precision 
 			 */
 			if (aggFuncName && strlen(aggFuncName) == 3 &&
 					(strncmp(aggFuncName, "sum", 3) == 0))
```
nasbyj, line 666: I realize this is pre-existing code, but why aren't we just using `strcmp` here instead of having to invite `strlen` to the party as well?

nasbyj (approved):
I'm not a fan of the use of strncmp, but not going to block because of that.

hong-yuan (approved):
:shipit:

PR #122, opened 2022-04-19
Smart Github workflows for downloading engine repo

nasbyj:
### Description
Instead of hard-coding where to download the engine repository from, create a script that checks for alternate places to download from. In order, it checks for:

- A user fork with the same branch name
- A matching branch name in the project engine repo

If neither of those exists and $DEFAULT_BRANCH is set (via clone_engine_repo.conf) it uses that branch from the project engine repo. Otherwise, it uses the default branch for the project engine repo (ie, git clone without the --branch option).

Signed-off-by: Jim C. Nasby <nasbyj@amazon.com>

### Check List
- [x] Commits are signed per the DCO using --signoff 

By submitting this pull request, I confirm that my contribution is under the terms of the Apache 2.0 and PostgreSQL licenses, and grant any person obtaining a copy of the contribution permission to relicense all or a portion of my contribution to the PostgreSQL License solely to contribute all or a portion of my contribution to the PostgreSQL open source project.

For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/babelfish-for-postgresql/babelfish_extensions/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #214, opened 2022-05-18
Smart download

nasbyj:
Cherry-pick ec731d16b into 1-x-dev

# Description
Smart Github workflows for downloading engine repo (#122)

Instead of hard-coding where to download the engine repository from, create a script that checks for alternate places to download from. In order, it checks for:

A user fork with the same branch name
A matching branch name in the project engine repo
If neither of those exists and $DEFAULT_BRANCH is set (via clone_engine_repo.conf) it uses that branch from the project engine repo. Otherwise, it uses the default branch for the project engine repo (ie, git clone without the --bra

Signed-off-by: Jim C. Nasby nasbyj@amazon.com


.github/scripts/clone_engine_repo
```diff
@@ -0,0 +1,93 @@
...
+fi
+
+if [ "$1" == -n ]; then
+    DRY_RUN=1
+    shift
+fi
+
+GH_USER=$1
+BRANCH=$2
+<<<<<<< Updated upstream
```
2jungkook: Please resolve the conflict

2jungkook (changes requested):

hong-yuan (dismissed):
:shipit:

PR #231, opened 2022-05-24
Column is_fixed_role for 'sysadmin' should return 1

shalinilohia50:
Column `is_fixed_role`  in the view sys.server_principals returns 1 if the principal is one of the built-in server roles with fixed permissions. `Sysadmin` is a built-in server role and for this role, `is_fixed_role` should return 1, but currently it returns 0. 
With this change, we have updated the column value `is_fixed_role` for sysadmin in the view sys.server_principals.

Task: BABEL-3225


surendravishnoi (changes requested):
Please make commit message more descriptive. Also, add upgrade tests if applicable.

shalinilohia50:
> Please make commit message more descriptive. Also, add upgrade tests if applicable.

Updated the commit message. Not sure if upgrade tests are applicable for this minor change.

surendravishnoi (approved):

macwoody (approved):

PR #233, opened 2022-05-24
Implement tsql_constraintdef for check constraint definition

shalinilohia50:
This PR contains the following changes:
1. Added **pltsql_ruleutils.c** and **pltsql_ruleutils.h** files: This file contains the rules to de-parse various objects. The functions and structures in this file are taken from PG's ruleutils.c
2. Add **tsql_constraintdef** function to get the check constraint definition
3. Code clean up and basic tests for check constraint definition 
4. Support TSQL typename and typmod de-parsing : Implemented tsql_format_type_extended hook function which returns string representation of type by de-parsing it from type tuples. It will be used
to de-parse data types while de-parsing other object definitions like
constant or variable. Implemented tsql_printTypmod() for printing
typmod in TSQL way.
5. Replace :: with CAST AS function, which is a valid format for TSQL

Task: [BABEL-3222](https://jira.rds.a2z.com/browse/BABEL-3222), BABEL-3146
Authored-by: Harsh Lunagariya <lunharsh@amazon.com>, Shalini Lohia <lshalini@amazon.com>


test/JDBC/expected/check_constraints_definition.out
```diff
@@ -0,0 +1,167 @@
...
+create table test_functioncall(
+	col1 sys.nvarchar,
+	check (isjson(col1) > 0),
+	check (right(col1,1) <> ','),
+	check (ltrim(col1) <> ''),
+	check (CONVERT([nvarchar](128),(getutcdate() AT TIME ZONE [col1]))<>'')
+);
+go
+
+create table test_null(a int, b int, check(a IS NOT NULL), CONSTRAINT constraint1 check (a>10));
```
rishabhtanwar29, line 92: Can we also add a test for `<> NULL` construct with both `ANSI_NULLS` on and off settings. Something like this:
```
set ansi_nulls to on;
create table test_null1(a int, int b, check(a <> NULL), check(a = NULL);
set ansi_nulls to off;
create table test_null2(a int, int b, check(a <> NULL), check(a = NULL);
```
It is just to verify if we are generating correct constraint def with ansi_nulls settings.

contrib/babelfishpg_tsql/src/pltsql_ruleutils.c
```diff
@@ -0,0 +1,2441 @@
...
+
+		typname = NameStr(typeform->typname);
+
+
+		buf = quote_qualified_identifier(nspname, typname);
+
+		/*
+		 * Replace bbf_binary with binary
+		 */
+		if (strcmp(buf, "bbf_binary") == 0)
```
rishabhtanwar29: Don't we also need to replace bbf_varbinary to varbinary?

contrib/babelfishpg_tsql/src/pltsql_ruleutils.c
```diff
@@ -0,0 +1,2441 @@
...
+			return pstrdup("???");
+		else
+			elog(ERROR, "cache lookup failed for type %u", type_oid);
+	}
+	typeform = (Form_pg_type) GETSTRUCT(tuple);
+
+	with_typemod = (flags & FORMAT_TYPE_TYPEMOD_GIVEN) != 0 && (typemod >= 0);
+
+	buf = NULL;
+	LOCAL_FCINFO(fcinfo, 1);
```
rishabhtanwar29: Should not mix variable declaration with the implementation. Put `LOCAL_FCINFO` at top with other variable declarations.

contrib/babelfishpg_tsql/src/pltsql_ruleutils.c
```diff
@@ -0,0 +1,2441 @@
...
+	/*
+	* In case of time, datetime2 or datetimeoffset print typmod
+	* info directly because it uses timestamp typmodout function
+	* which appends timezone data along with typmod which is not
+	* required. Directly print typename for smalldatetime as it
+	* doesn't support typmod.
+	*/
+	if (strcmp(typname, "time") == 0 ||
+	   strcmp(typname, "datetime2") == 0 ||
+	   strcmp(typname, "datetimeoffset") == 0)
```
rishabhtanwar29: We should also consider the case when typname is schema qualified, in that case the above check won't work.

contrib/babelfishpg_tsql/src/pltsql_ruleutils.c
```diff
@@ -0,0 +1,2441 @@
...
+									  argnames, argtypes,
+									  expr->funcvariadic,
+									  &use_variadic,
+									  context->special_exprkind);
+
+	/*
+	 * AT TIMEZONE from TSQL is parsed to timezone function internally.
+	 * While de-parsing, convert it to AT TIME ZONE explicitly.
+	 */
+	if (strcmp(funcname,"timezone") == 0)
```
rishabhtanwar29, line 1590: Need consider the case when funcname is schema qualified.

contrib/babelfishpg_tsql/src/pltsql_ruleutils.c
```diff
@@ -0,0 +1,2441 @@
...
+		default:
+			simple_quote_literal(valbuf, extval);
+			break;
+	}
+
+	pfree(extval);
+
+	if (showtype < 0)
+	{
+		appendStringInfoString(buf, valbuf->data);
```
rishabhtanwar29, line 1355: We need to free memory of `valbuf->data` before returning

shalinilohia50: It's addressed in the next commit.

contrib/babelfishpg_tsql/src/pltsql_ruleutils.c
shalinilohia50: I have added some tests where the type name was schema qualified.

contrib/babelfishpg_tsql/src/pltsql_ruleutils.c
shalinilohia50: Done

test/JDBC/expected/check_constraints_definition.out
shalinilohia50: Added a test case for the same.

contrib/babelfishpg_tsql/src/pltsql_ruleutils.c
shalinilohia50: Added a new column in the test for varbinary, we don't see such a case for varbinary.

contrib/babelfishpg_tsql/src/pltsql_ruleutils.c
shalinilohia50: Added a test where the function is schema qualified.

contrib/babelfishpg_tsql/sql/sys_functions.sql
```diff
@@ -17,6 +17,11 @@ RETURNS sys.NVARCHAR(128)
 AS 'babelfishpg_tsql', 'user_name'
 LANGUAGE C IMMUTABLE PARALLEL SAFE;
 
+CREATE OR REPLACE FUNCTION sys.tsql_get_constraintdef(IN login OID DEFAULT NULL)
```
Deepesh125: Nitpicks: param name could have been better here. "login" is a bit confusing.

contrib/babelfishpg_tsql/src/pltsql_ruleutils.c
```diff
@@ -0,0 +1,2443 @@
...
+		case T_CollateExpr:
+			{
+				CollateExpr *collate = (CollateExpr *) node;
+				Node	   *arg = (Node *) collate->arg;
+
+				if (!PRETTY_PAREN(context))
+					appendStringInfoChar(buf, '(');
+				get_rule_expr_paren(arg, context, showimplicit, node);
+				appendStringInfo(buf, " COLLATE %s",
+								 generate_collation_name(collate->collOid));
```
Deepesh125, line 968: We need to be careful here. What would happen if collate->collOid = DEFAULT_COLLATION_OID? would generate_collation_name return "DEFAULT"? if so then is it valid to show it to end user? Probably, it would be okay to show as "DEFAULT" if original constraint was created under PG dialect. But we might need to use Oid corresponding to server_collation_name GUC if original constraint was created under TSQL dialect?

contrib/babelfishpg_tsql/sql/sys_functions.sql
KushaalShroff: I think I had also added the same comments in the previous PR. Looks like they arent addressed yet!

contrib/babelfishpg_tsql/sql/information_schema_tsql.sql
```diff
@@ -401,4 +400,31 @@ CREATE VIEW information_schema_tsql.table_constraints AS
...
 
+/*
+ * CHECK_CONSTRAINTS view
+ */
+
+CREATE VIEW information_schema_tsql.check_constraints AS
+	SELECT CAST(nc.dbname AS sys.nvarchar(128)) AS "CONSTRAINT_CATALOG",
+		CAST(extc.orig_name AS sys.nvarchar(128)) AS "CONSTRAINT_SCHEMA",
+        	CAST(c.conname AS sys.sysname) AS "CONSTRAINT_NAME",
+		CAST(sys.tsql_get_constraintdef(c.oid) AS sys.nvarchar(4000)) AS "CHECK_CLAUSE"
```
KushaalShroff: Extra indentation

contrib/babelfishpg_tsql/sql/sys_functions.sql
shalinilohia50: It's been addressed in the next commit.

contrib/babelfishpg_tsql/sql/information_schema_tsql.sql
```diff
@@ -401,4 +400,31 @@ CREATE VIEW information_schema_tsql.table_constraints AS
...
+
+    FROM sys.pg_namespace_ext nc LEFT OUTER JOIN sys.babelfish_namespace_ext extc ON nc.nspname = extc.nspname,
+         sys.pg_namespace_ext nr LEFT OUTER JOIN sys.babelfish_namespace_ext extr ON nr.nspname = extr.nspname,
```
KushaalShroff: Please explain why we need nc and nr? Can't we make use of nc only?

contrib/babelfishpg_tsql/src/pltsql_ruleutils.c
KushaalShroff: > it would be okay to show as "DEFAULT" if original constraint was created under PG dialect
This is work is strictly for TSQL constraints

We should have tests for collate default

test/JDBC/expected/check_constraints_definition.out
```diff
@@ -0,0 +1,210 @@
...
+db1#!#dbo#!#test_functioncall_col1_check1#!#(("right"(col1, 1) <> ','))
+db1#!#dbo#!#test_functioncall_col1_check2#!#((ltrim(col1) <> ''))
+db1#!#dbo#!#test_functioncall_col1_check3#!#((CAST((getutcdate() AT TIME ZONE col1) AS nvarchar(128)) <> CAST('' AS varchar)))
+db1#!#dbo#!#test_null1_a_check#!#((a <> CAST(NULL AS int)))
+db1#!#dbo#!#test_null1_b_check#!#((b = CAST(NULL AS int)))
+db1#!#dbo#!#test_null2_a_check#!#((a IS NOT NULL))
+db1#!#dbo#!#test_null2_b_check#!#((b IS NULL))
+db1#!#dbo#!#test_null_a_check#!#((a IS NOT NULL))
+db1#!#dbo#!#test_tsql_cast_c_decimal_check#!#(((c_decimal < CAST((CAST((133.23) AS money)) AS decimal(18,0))) AND (c_decimal < CAST((133.230182309832423) AS decimal(5,3)))))
+db1#!#dbo#!#test_tsql_cast_c_float_check#!#(((c_float < CAST((133.230182309832423) AS float)) AND (c_float < CAST((133.230182309832423) AS real))))
```
KushaalShroff, line 134: We shall document such diffs like float(7) is being displayed as real.

contrib/babelfishpg_tsql/src/pltsql_ruleutils.h
```diff
@@ -0,0 +1,12 @@
...
+ * pltsql_ruleutils.h
+ *	  Declarations for pltsql_ruleutils.h
+ *
+ */
+
+#ifndef PLTSQL_RULEUTILS_H
+#define PLTSQL_RULEUTILS_H
+
+#endif							/* PLTSQL_RULEUTILS_H */
+
```
KushaalShroff: What is the use of this file?

shalinilohia50: Removed this file in the next commit. It's not needed.

test/JDBC/expected/check_constraints_definition.out
shalinilohia50: Added this as a comment in the test, will document it in a quip with more differences.

contrib/babelfishpg_tsql/sql/information_schema_tsql.sql
shalinilohia50: Fixed it.

contrib/babelfishpg_tsql/src/pltsql_ruleutils.c
shalinilohia50: Added tests in the next commit.

test/JDBC/expected/check_constraints_definition.out
```diff
@@ -0,0 +1,213 @@
...
+
+alter table test_null add constraint constraint2 check(a<=20 and b>a and (a between 40 and 50));
+go
+
+set ansi_nulls on;
+go
+create table test_null1(a int, b int, check(a <> NULL), check(b = NULL));
+go
+	
+set ansi_nulls off;
```
rishabhtanwar29, line 106: I think by default ansi_nulls is on so we should reset it to default after this test

rishabhtanwar29 (approved):
Looks good to me!

xhfanhe (approved):

contrib/babelfishpg_tsql/sql/information_schema_tsql.sql
shalinilohia50: Removed the extra join in the next commit.

kuntalghosh (approved):

sangilsong (approved):

PR #254, opened 2022-06-08
Create a Github action that can test pg_hint_plan

gattania:
### Description

Currently Babelfish does not support query hints. To add support for query hints, we will be using the third-part library pg_hint_plan. This pull request has  the following changes:

1. Creating a Github action that can build and install the specified release of pg_hint_plan
2. Add a separate Github workflow to test pg_hint_plan as part of Github CI
3. Unit tests to verify if the query hints are getting applied

### Issues Resolved

Task: BABEL-3291

### Check List


.github/scripts/clone_pg_hint_plan_repo
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
+
+set -e
+
+BASE_URL=https://github.com/ossc-db/pg_hint_plan
+GIT_URL=${BASE_URL}.git
+
+echo running git clone "$GIT_URL"
+git clone "$GIT_URL" && exit
```
nasbyj: Do we really want to pull their master (unreleased) branch? I think at minimum we'd also want to pull the latest release of pg_hint_plan.

.github/composite-actions/build-pg-hint-plan/action.yml
```diff
@@ -0,0 +1,14 @@
+name: 'Build pg-hint-plan'
+
+runs:
+  using: "composite"
+  steps:
+    - name: Checkout, Build and Install pg-hint-plan
```
nasbyj, line 6: Since use of pg_hint_plan is optional, I think we also need a test run that does NOT include it.

nasbyj:
Overall, there's some important edge conditions that are not being tested, but these bring up a potentially bigger issue: exactly what versions of pg_hint_plan will we be supporting? The answer to that will also change over time (ie, Babelfish 3.x will eventually support a newer version than 2.x ever will).

nasbyj (changes requested):
Overall, there's some important edge conditions that are not being tested, but these bring up a potentially bigger issue: exactly what versions of pg_hint_plan will we be supporting? The answer to that will also change over time (ie, Babelfish 3.x will eventually support a newer version than 2.x ever will).

.github/composite-actions/build-pg-hint-plan/action.yml
gattania: I have added a separate flow for pg_hint_plan instead of merging it with the JDBC tests.

.github/scripts/clone_pg_hint_plan_repo
gattania: I've added a new configuration file from which the release version of pg_hint_plan is picked up

.github/workflows/pg-hint-plan-tests.yml
```diff
@@ -0,0 +1,88 @@
...
+        id: build-extensions
+        if: always() && steps.compile-antlr.outcome == 'success'
+        uses: ./.github/composite-actions/build-extensions
+      
+      - name: Install Extensions
+        id: install-extensions
+        if: always() && steps.build-extensions.outcome == 'success'
+        uses: ./.github/composite-actions/install-extensions
+
+      - name: Run JDBC Tests
```
2jungkook: We don't need the normal JDBC test here.

test/JDBC/expected/BABEL-QUERY-HINT.out
```diff
@@ -0,0 +1,35 @@
+drop table if exists jira3291
```
2jungkook: Our convention is `babel_3291_t1`. You can try searching other test cases. Also, I would suggest to change the test name to BABEL-PG-HINT-PLAN or just BABEL-3291.

.github/workflows/pg-hint-plan-tests.yml
gattania: Removed the JDBC tests from the pg_hint_plan workflow

test/JDBC/expected/BABEL-QUERY-HINT.out
gattania: Renamed the test files and the table name as per the convention

.github/workflows/pg-hint-plan-tests.yml
```diff
@@ -0,0 +1,80 @@
...
+        uses: ./.github/composite-actions/compile-antlr
+
+      - name: Build pg-hint-plan
+        id: build-pg-hint-plan
+        if: always() && steps.compile-antlr.outcome == 'success'
+        uses: ./.github/composite-actions/build-pg-hint-plan
+      
+      - name: Build Extensions
+        id: build-extensions
+        if: always() && steps.compile-antlr.outcome == 'success'
```
2jungkook, line 28: Let's use sequential condition checks, e.g.,
```
always() && steps.build-pg-hint-plan.outcome == 'success'
```

.github/workflows/pg-hint-plan-tests.yml
```diff
@@ -0,0 +1,80 @@
...
+        uses: ./.github/composite-actions/build-extensions
+      
+      - name: Install Extensions
+        id: install-extensions
+        if: always() && steps.build-extensions.outcome == 'success'
+        uses: ./.github/composite-actions/install-extensions
+
+      - name: Run pg-hint-plan tests
+        id: run-pg-hint-plan-test
+        if: always() && steps.build-pg-hint-plan.outcome == 'success'
```
2jungkook, line 43: same here

gattania: Updated the conditions sequentially

2jungkook (dismissed):

sangilsong (dismissed):

test/JDBC/expected/BABEL-3291.out
```diff
@@ -0,0 +1,35 @@
...
+
+set babelfish_showplan_all on
+go
+
+select /*+SeqScan(babel_3291_t1)*/ * from babel_3291_t1 where a1 = 1
+go
+~~START~~
+text
+Query Text: select /*+SeqScan(babel_3291_t1)*/ * from babel_3291_t1 where a1 = 1
+Seq Scan on babel_3291_t1  (cost=0.00..38.25 rows=1 width=8)
```
nasbyj: How do we verify that hints are actually working? For example, an empty table will *always* use a SeqScan if it's not hinted. So for these tests, at a minimum, we should also be trying to force an index scan.

nasbyj (changes requested):

test/JDBC/expected/BABEL-3291.out
gattania: I have added the comments to clarify the expected output with and without hints

test/JDBC/pg_hint_plan/BABEL-3291.sql
```diff
@@ -0,0 +1,26 @@
...
+go
+
+create table babel_3291_t1(a1 int PRIMARY KEY, b1 int)
+go
+
+-- Switch the session variable to show the query plan to ON
+set babelfish_showplan_all on
+go
+
+-- Run a SELECT query without any hints. The query plan will use an index scan in this case
```
nasbyj: Not to nitpick, but the comment doesn't mention what we'd discussed: that it's critical that this query plan is not what's hinted at. I suggest something like:

```
/*
 * Run a query with no hints. Besides verifying that un-hinted queries still
 * work, this test has a VERY important side-effect: it ensures that when this
 * query is not hinted it produces a *different* plan than the sequential scan
 * that we're hinting below. This is crucial because it's the only way to
 * verify that the next test is actually valid. IE: if the planner was going to
 * choose a sequential scan anyway, the next test wouldn't actually prove that
 * hints were working.
 */
```

nasbyj (changes requested):

nasbyj (approved):

sangilsong (approved):

2jungkook (approved):

PR #277, opened 2022-06-15
Active login should not be allowed to be dropped

HarshLunagariya:
### Description

This commit implements is_active_login() function which returns whether
given login is active ( being used by other session) or not. To get count of
active user backend for given login it uses CountUserBackends() which
counts backends that are used by specified user. Added check for the
active login for DROP LOGIN statements in pl_handler. Updated the tests
which drops login such that active session with given login will be terminated
before dropping it.

Task: BABEL-3232
Signed-off-by: Harsh Lunagariya <lunharsh@amazon.com>
 
### Issues Resolved

BABEL-3232

### Check List


contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -468,6 +468,34 @@ is_login(Oid role_oid)
...
+	for (curr_backend = 1; curr_backend <= num_backends; curr_backend++)
+	{
+		/*
+		 * Fetch the userid for active backend from pg_statistics and check if
+		 * given role_oid matches with curr_backend's userid, if it matches then
+		 * it is active login.
+		 */
+		Oid	backend_userid;
+		backend_userid = DatumGetObjectId(
+									DirectFunctionCall1(pg_stat_get_backend_userid,
```
macwoody: I think this is incorrect. 
Not all of this curr_backend is valid one, when return value id NULL. 
Actually, you don't have to rely on function API, you could call the underlying implementation

test/JDBC/expected/BABEL-LOGIN-USER-EXT.out
```diff
@@ -218,36 +226,48 @@ go
...
-3
+14
 ~~END~~
 
 
 -- Test initialization with password
 CREATE LOGIN r2 WITH PASSWORD = '123'
 go
+~~ERROR (Code: 33557097)~~
+
```
macwoody: I suggest you correct the test to make it aligned with original expectation.

test/JDBC/expected/BABEL-LOGIN-USER-EXT.out
```diff
@@ -37,6 +37,10 @@ go
 -- Create Login through tsql
 CREATE LOGIN r1 WITH PASSWORD = 'abc';
 go
+~~ERROR (Code: 33557097)~~
```
macwoody: Please do not change original test intention.  It it was expecting r1 to be created, then it should be unless that intention is wrong.

contrib/babelfishpg_tsql/src/catalog.c
HarshLunagariya: 1. Now instead of API call, we're fetching the backend entry(PgBackendStatus) which maintain by pg statistics collector.
2. Instead of checking for all the active backends, now only client backends are being considered.[Note: We get valid(not null) userid for client-backends, wal-sender and bgworker processes, other processes gives NULL userid]

test/JDBC/expected/BABEL-LOGIN-USER-EXT.out
HarshLunagariya: Fixed

test/JDBC/expected/BABEL-LOGIN-USER-EXT.out
HarshLunagariya: Fixed

xhfanhe (commented):
Code looks good to me, but there is a strange time out failure in MVU test, please take a look

xhfanhe:
Please add upgrade tests for this change

HarshLunagariya:
> Please add upgrade tests for this change

It doesn't contain SQL upgrade change or code change is not affecting upgrade in any way. Even if I add upgrade tests for it it won't make sense because this test needs creating login in prepare part(we haven't done any changes in create login part) and verify the support for preventing active login(This is what normal test is already verifying).

xhfanhe:
Yes that makes sense.
I was thinking that upgrade tests can work as usual JDBC tests, but you're right it doesn't have to be that way

xhfanhe (dismissed):

shalinilohia50 (dismissed):

shalinilohia50:
please remove the merge conflicts.

shalinilohia50 (approved):

macwoody (approved):

priyansx (approved):
LGTM

PR #279, opened 2022-06-15
Add sys.dm_hadr_cluster and sys.xml_indexes views

faviansamatha:
### Description

Previously, Babelfish did not support sys.dm_hadr_cluster and sys.xml_indexes. Stub implementations for both these views are included in this commit in order to unblock SSMS functionalities.

Task: BABELFISH-381, BABELFISH-438

### Check List


contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.1.0--2.2.0.sql
```diff
@@ -326,6 +326,48 @@ FROM sys.indexes si
...
+  , CAST(idx.is_primary_key AS sys.bit) AS is_primary_key
+  , CAST(idx.is_unique_constraint AS sys.bit) AS is_unique_constraint
+  , CAST(idx.fill_factor AS sys.tinyint) AS fill_factor
+  , CAST(idx.is_padded AS sys.bit) AS is_padded
+  , CAST(idx.is_disabled AS sys.bit) AS is_disabled
+  , CAST(idx.is_hypothetical AS sys.bit) AS is_hypothetical
+  , CAST(idx.allow_row_locks AS sys.bit) AS allow_row_locks
+  , CAST(idx.allow_page_locks AS sys.bit) AS allow_page_locks
+  , CAST(idx.has_filter AS sys.bit) AS has_filter
+  , CAST(idx.filter_definition AS sys.nvarchar) AS filter_definition
```
Deepesh125, line 349: shouldn't we cast this to nvarchar(max)?

contrib/babelfishpg_tsql/sql/sys_views.sql
```diff
@@ -2151,3 +2151,45 @@ SELECT
...
+  , CAST(idx.is_primary_key AS sys.bit) AS is_primary_key
+  , CAST(idx.is_unique_constraint AS sys.bit) AS is_unique_constraint
+  , CAST(idx.fill_factor AS sys.tinyint) AS fill_factor
+  , CAST(idx.is_padded AS sys.bit) AS is_padded
+  , CAST(idx.is_disabled AS sys.bit) AS is_disabled
+  , CAST(idx.is_hypothetical AS sys.bit) AS is_hypothetical
+  , CAST(idx.allow_row_locks AS sys.bit) AS allow_row_locks
+  , CAST(idx.allow_page_locks AS sys.bit) AS allow_page_locks
+  , CAST(idx.has_filter AS sys.bit) AS has_filter
+  , CAST(idx.filter_definition AS sys.nvarchar) AS filter_definition
```
Deepesh125: shouldn't we cast this to nvarchar(max)?

faviansamatha: Since it's in `plpgsql` it will produce a syntax error. Should we do `sys.nvarchar(4000)` instead?

faviansamatha: Chatted offline, and looks like the atttypmod value for this column (from pg_attribute view) is -1 which should be okay to use

Deepesh125: typmod = -1 would be considered as typmod = 30 for sys.nvarchar. check https://github.com/babelfish-for-postgresql/babelfish_extensions/blob/BABEL_2_X_DEV/contrib/babelfishpg_tsql/src/pltsql_utils.c#L93. So let's use sys.nvarchar(4000) for the time being.

faviansamatha: Sounds good!

Deepesh125 (dismissed):

hsuamz (dismissed):

hsuamz (approved):

leszekamz (approved):

Imran-imtiaz48 (commented):
This script creates several SQL Server system views, using `CREATE OR REPLACE VIEW`, with the intention of granting read access (via `GRANT SELECT`) to the `PUBLIC` role. Here's a review based on functionality, structure, and potential improvements:

### Positives:
1. **Consistency**: The structure of each view follows a consistent pattern. This ensures easy readability and maintenance.
2. **Use of Casting**: Proper use of `CAST` ensures that the data types for each field are clearly defined, which avoids ambiguity in the view definitions.
3. **Security Consideration**: By explicitly granting `SELECT` permissions to `PUBLIC`, the script defines who can access these views, which simplifies role management.
4. **Optimization for Query Performance**: By using `WHERE FALSE`, the views return no data, effectively acting as placeholders or security objects that are lightweight and have minimal impact on the server.

### Concerns:
1. **Utility of `WHERE FALSE`**: Since every `SELECT` query contains a `WHERE FALSE` clause, no actual data will be returned from any of the views. This is likely intentional (to create "empty" system views), but it might cause confusion for users expecting data. If the intention is to create stub views, this behavior should be documented clearly.
   
2. **Sys Object Dependencies**: Since you are referencing system objects (e.g., `sys.indexes` in `sys.spatial_indexes`), there's a risk that future system updates could break or conflict with these view definitions if those system objects change. This should be monitored.

3. **Security**: Granting `SELECT` to `PUBLIC` may expose metadata to users who may not require it. While this is fine for a low-security environment, in a production setting, it would be better to restrict these permissions to specific roles.

4. **Lack of Documentation**: Without comments or documentation in the code, it’s unclear why these views are necessary. It’s good practice to include comments explaining the purpose of each view, especially when they return no results.

### Improvements:
1. **Clarification via Comments**: Adding comments before each view explaining why it exists, what it’s supposed to represent, and the reason for returning no data (if intentional) would improve the clarity and usability of the script.
   
2. **Potential Role-Based Permissions**: Instead of granting `SELECT` permissions to `PUBLIC`, consider creating a specific role (e.g., `metadata_reader`) to limit exposure of these views to only the users who need access.

3. **Future Maintenance**: To future-proof the script, include error handling or checks for system object dependencies to ensure that updates to the underlying system objects won’t disrupt these views.

### Overall Assessment:
The script is well-structured but appears to be creating placeholder views that might not be intended for regular data use. The key issue is the `WHERE FALSE` clause, which should be clearly documented as part of the intended behavior to avoid confusion. Also, granting permissions to `PUBLIC` can raise security concerns in sensitive environments.

PR #286, opened 2022-06-17
Add support for Index-related T-SQL hint conversion

gattania:
### Description
Description

Currently Babelfish does not support query hints. To add support for query hints, we are using the third-part library pg_hint_plan. This pull request has the following changes:

1. Add support for Index-related T-SQL hint (INDEX(index_name), INDEX=index_name) conversion
2. Add a test to verify the hinted queries work as expected

 
### Issues Resolved

Task: BABEL-3292

### Check List
- [] Commits are signed per the DCO using --signoff 

### DIFF
--- a/contrib/babelfishpg_tsql/src/tsqlIface.cpp
+++ b/contrib/babelfishpg_tsql/src/tsqlIface.cpp
@@ -1458,6 +1458,7 @@ public:
 			add_rewritten_query_fragment_to_mutator(&mutator);
 			mutator.run();
 		}
+		clear_query_hints();
 		clear_rewritten_query_fragment();
 	}
 
--- a/contrib/babelfishpg_tsql/src/tsqlIface.cpp
+++ b/contrib/babelfishpg_tsql/src/tsqlIface.cpp
@@ -3135,6 +3135,12 @@ void extractQueryHintsFromOptionClause(TSqlParser::Option_clauseContext *octx)
 		if (option->TABLE())
 		{
 			std::string table_name = ::getFullText(option->table_name());
+			size_t idx = table_name.find('.');
+			while (idx != std::string::npos)
+			{
+				table_name = table_name.substr(idx + 1);
+				idx = table_name.find('.');
+			}
 			if (!table_name.empty())
 			{
 				for(auto table_hint: option->table_hint())
@@ -3159,7 +3165,7 @@ std::string extractTableName(TSqlParser::Ddl_objectContext *ctx)
 {
 	std::string table_name;
 	if (ctx->full_object_name())
-		table_name = ::getFullText(ctx->full_object_name());
+		table_name = stripQuoteFromId(ctx->full_object_name()->object_name);
 	else if (ctx->local_id())
 		table_name = ::getFullText(ctx->local_id());
 	return table_name;
@@ -4764,7 +4770,7 @@ static void post_process_table_source(TSqlParser::Table_source_itemContext *ctx,
 		if (!wctx->sample_clause()) {
 			std::string table_name;
 			if (ctx->full_object_name())
-				table_name = ::getFullText(ctx->full_object_name());
+				table_name = stripQuoteFromId(ctx->full_object_name()->object_name);
 			else if (ctx->local_id())
 				table_name = ::getFullText(ctx->local_id());
 			extractTableHints(wctx, table_name);
--- a/contrib/babelfishpg_tsql/src/tsqlIface.cpp
+++ b/contrib/babelfishpg_tsql/src/tsqlIface.cpp
@@ -98,7 +98,9 @@ void replaceTokenStringFromQuery(PLtsql_expr* expr, TerminalNode* tokenNode, con
 void replaceCtxStringFromQuery(PLtsql_expr* expr, ParserRuleContext *ctx, const char *repl, ParserRuleContext *baseCtx);
 void removeTokenStringFromQuery(PLtsql_expr* expr, TerminalNode* tokenNode, ParserRuleContext *baseCtx);
 void removeCtxStringFromQuery(PLtsql_expr* expr, ParserRuleContext *ctx, ParserRuleContext *baseCtx);
-void extractTableHints(TSqlParser::With_table_hintsContext *ctx, std::string table_name);
+void extractTableHintsFromOptionClause(TSqlParser::Option_clauseContext *octx);
+void extractTableHints(TSqlParser::With_table_hintsContext *tctx, std::string table_name);
+void extractTableHint(TSqlParser::Table_hintContext *table_hint, std::string table_name);
 std::string extractIndexValues(std::vector<TSqlParser::Index_valueContext *> index_valuesCtx, char *table_name);
 
 static void *makeBatch(TSqlParser::Tsql_fileContext *ctx, tsqlBuilder &builder);
@@ -2159,8 +2161,10 @@ static void process_select_statement(
 	Assert(mutator);
 	PLtsql_expr *expr = mutator->expr;
 	ParserRuleContext* baseCtx = mutator->ctx;
-	for (auto octx : selectCtx->option_clause()) // query hint
+	for (auto octx : selectCtx->option_clause()) {// query hint
+		extractTableHintsFromOptionClause(octx);
 		removeCtxStringFromQuery(expr, octx, baseCtx);
+	}
 }
 
 ////////////////////////////////////////////////////////////////////////////////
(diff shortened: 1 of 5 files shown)


contrib/babelfishpg_tsql/src/tsqlIface.cpp
```diff
@@ -2093,6 +2121,12 @@ static void process_query_specification(
 			}
 		}
 	}
+	
+	if(query_hints.size()) {
+		add_query_hints(expr);
```
2jungkook: Not sure process_query_specification is the right place to add query hints. Please check the following cases:
- INSERT/DELETE/UPDATE statements
- Multiple select statements in a single query, e.g., UNION and CTE (Common Table Expressions)

contrib/babelfishpg_tsql/src/tsqlIface.cpp
```diff
@@ -3090,6 +3127,45 @@ void removeCtxStringFromQuery(PLtsql_expr* expr, ParserRuleContext *ctx, ParserR
...
+	}
+}
+
+void extractTableHints(TSqlParser::With_table_hintsContext *tctx, std::string table_name) {
+	for (auto table_hint: tctx->table_hint())
+		extractTableHint(table_hint, table_name);
+}
+
+
+void extractTableHint(TSqlParser::Table_hintContext *table_hint, std::string table_name) {
```
2jungkook: If there is no written coding convention, then it would be better to follow conventions in the same file. Most of the code in this file look like the following:
```
void extractTableHint(...)
{
    if(table_hint->INDEX())
    {
        ...
    }
}
```

contrib/babelfishpg_tsql/src/tsqlIface.cpp
```diff
@@ -4658,9 +4734,17 @@ static void post_process_table_source(TSqlParser::Table_source_itemContext *ctx,
 	for (auto cctx : ctx->table_source_item())
 		post_process_table_source(cctx, expr, baseCtx);
 
-	for (auto wctx : ctx->with_table_hints())
+	for (auto wctx : ctx->with_table_hints()) {
+		if(!wctx->sample_clause()) {
+			std::string table_name;
+			if(ctx->full_object_name())
+				table_name = ::getFullText(ctx->full_object_name());
+			else
```
2jungkook: There are cases other than `full_object_name` or `local_id`. To prevent unexpected results, let's call `extractTableHints()` only if we have a valid table_name.

contrib/babelfishpg_tsql/src/tsqlIface.cpp
```diff
@@ -4670,31 +4754,64 @@ void process_execsql_remove_unsupported_tokens(TSqlParser::Dml_statementContext
...
 			removeCtxStringFromQuery(stmt->sqlstmt, dctx->delete_statement_from()->table_alias()->with_table_hints(), ctx);
-		if (dctx->with_table_hints()) // table hints
+		if (dctx->with_table_hints()) { // table hints
+			if(!dctx->with_table_hints()->sample_clause() && dctx->delete_statement_from()->ddl_object()) {
+				std::string table_name;
+				if(dctx->delete_statement_from()->ddl_object()->full_object_name())
+					table_name = ::getFullText(dctx->delete_statement_from()->ddl_object()->full_object_name());
+				else
+					table_name = ::getFullText(dctx->delete_statement_from()->ddl_object()->local_id());
+				extractTableHints(dctx->with_table_hints(), table_name);
```
2jungkook: We can introduce a function to reduce repetitive code.

test/JDBC/expected/BABEL-3292.out
```diff
@@ -0,0 +1,173 @@
...
+ * This also ensures that when the SELECT query is not hinted it produces a different plan(bitmap heap scan and bitmap index scan)
+ * than the index scan that we're hinting in the query below. This verifies that the next test is actually valid.
+ * If the planner was going to choose a sequential scan anyway, the next test wouldn't actually prove that hints were working.
+ */
+select * from babel_3292_t1 where b1 = 1
+go
+~~START~~
+text
+Query Text: select * from babel_3292_t1 where b1 = 1
+Bitmap Heap Scan on babel_3292_t1  (cost=4.23..14.79 rows=10 width=12)
```
2jungkook: Let's remove cost part from the expected result using
```
select set_config('babelfishpg_tsql.explain_costs', 'off', false);
select set_config('babelfishpg_tsql.explain_costs', 'on', false);
```

The reasons are
- costs are not our objectives
- costs can be changed by other factors such as vacuum analyze or PG versions

gattania: Added the configuration so that costs are no longer part of the query plan explanation

contrib/babelfishpg_tsql/src/tsqlIface.cpp
gattania: Moved the change to insert the hints at a common place for all queries

contrib/babelfishpg_tsql/src/tsqlIface.cpp
gattania: Added a check to see if the table_name is valid

contrib/babelfishpg_tsql/src/tsqlIface.cpp
gattania: Added a function to extract table name from DDL object

contrib/babelfishpg_tsql/src/tsqlIface.cpp
```diff
@@ -3090,6 +3128,70 @@ void removeCtxStringFromQuery(PLtsql_expr* expr, ParserRuleContext *ctx, ParserR
...
+			extractTableHint(table_hint, table_name);
+	}
+}
+
+std::string extractTableName(TSqlParser::Ddl_objectContext *ctx)
+{
+	std::string table_name;
+	if(ctx->full_object_name())
+		table_name = ::getFullText(ctx->full_object_name());
+	else
```
2jungkook: What happens if both `ctx->full_object_name()` and `ctx->local_id()` are nulls?

gattania: One of them needs to be present in both  ddl_object as well as in table_source_item having with_table_hints

contrib/babelfishpg_tsql/src/tsqlIface.cpp
```diff
@@ -543,6 +552,27 @@ add_rewritten_query_fragment_to_mutator(PLtsql_expr_query_mutator *mutator)
...
+{
+	std::string hint =  "/*+ ";
+	for(auto q_hint: query_hints) {
+		hint += q_hint;
+		hint += " ";
+	}
+	hint += "*/";
+	StringInfoData new_query;
+	initStringInfo(&new_query);
+	appendStringInfo(&new_query, "%s %s", const_cast <char *>(hint.c_str()), expr->query);
```
sangilsong, line 566: nit: why do we need `const_cast`? We have to avoid using const_cast as much as possible. I think `construct_unique_index_name` still needs `char *` (not `const char*`)  so we have to keep it but other function such as `extractIndexValues()` should use const char*.

sangilsong (dismissed):

contrib/babelfishpg_tsql/src/tsqlIface.cpp
gattania: The appendStringInfo expects a char* as an argument and hence I had to use const_cast

contrib/babelfishpg_tsql/src/tsqlIface.cpp
```diff
@@ -3135,6 +3135,12 @@ void extractQueryHintsFromOptionClause(TSqlParser::Option_clauseContext *octx)
 		if (option->TABLE())
 		{
 			std::string table_name = ::getFullText(option->table_name());
+			size_t idx = table_name.find('.');
+			while (idx != std::string::npos)
+			{
+				table_name = table_name.substr(idx + 1);
+				idx = table_name.find('.');
+			}
```
sangilsong: please use `getFullText(option->table_name()->table())` to obtain a table name instead of manually finding `.`. table can be delimited like "schema.name.having.dot"."table.name.having.dot" so we should rely on the parser to split an identifier into schema name and table name. It would be good to have relevant tests too.

sangilsong (dismissed):

contrib/babelfishpg_tsql/src/tsqlIface.cpp
gattania: Using the `getFullText(option->table_name()->table)` to extract the table name now

macwoody:
Implementation is straight forward. I just want to ask one more question: 
Do we already have pg_hint_plan added as dependent extension before? If yes, which version?

macwoody (approved):

sangilsong (approved):

PR #290, opened 2022-06-20
Fix the JDBC timeout error in release mode

shalinilohia50:
We were assigning a non-volatile variable "old_db_name" inside TRY earlier. Now we have moved assignment out of TRY block. 
In release mode, it was leading to segmentation fault in the case of runtime error earlier.

Task: BABEL-3077


surendravishnoi (changes requested):
Please add more details to commit message.

shalinilohia50:
> 

Updated the commit message.

surendravishnoi (approved):

kuntalghosh (approved):

PR #314, opened 2022-06-22
Add catalog_default collation.

seneramz:
### Description

This PR adds the missing catalog_default collation which is used by
SSMS. It also adds the related unit tests.

Task: BABELFISH-401

Signed-off-by: Sertay Sener <seners@amazon.com>

### Check List

### DIFF
--- a/contrib/babelfishpg_tsql/sql/collation.sql
+++ b/contrib/babelfishpg_tsql/sql/collation.sql
@@ -140,6 +140,8 @@ CREATE COLLATION IF NOT EXISTS sys.Vietnamese_CS_AS (provider = icu, locale = 'v
 CREATE COLLATION sys.Vietnamese_CI_AI (provider = icu, locale = 'vi_VN@colStrength=primary', deterministic = false);
 CREATE COLLATION sys.Vietnamese_CI_AS (provider = icu, locale = 'vi_VN@colStrength=secondary', deterministic = false);
 
+CREATE COLLATION IF NOT EXISTS catalog_default FROM ucs_basic;
+
 -- collation catalog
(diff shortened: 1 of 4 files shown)


contrib/babelfishpg_tsql/sql/collation.sql
```diff
@@ -140,6 +140,8 @@ CREATE COLLATION IF NOT EXISTS sys.Vietnamese_CS_AS (provider = icu, locale = 'v
 CREATE COLLATION sys.Vietnamese_CI_AI (provider = icu, locale = 'vi_VN@colStrength=primary', deterministic = false);
 CREATE COLLATION sys.Vietnamese_CI_AS (provider = icu, locale = 'vi_VN@colStrength=secondary', deterministic = false);
 
+CREATE COLLATION IF NOT EXISTS catalog_default FROM ucs_basic;
```
Deepesh125, line 143: This is not how we want to fix this. This probably requires the discussion before we decide on solution.

JimFinnerty: +1, we don't want to fix it this way.  Since with this solution catalog_default is an actual collation that is different from the server collation, which might be, for example bbf_unicode_general_ci_as, PostgreSQL won't be able to pick which collation to use when columns with these collations are compared.  Perhaps under the covers they have the same semantics, but PostgreSQL doesn't know that.  catalog_default has to be a synonym for the real server collation, and similarly database_default has to be a synonym for the real database collation.  When these happen to be the same (as they always are in Babelfish at this time), then it must be possible to compare them to each other without causing a collation conflict.

JimFinnerty (changes requested):

PR #325, opened 2022-06-24
Add framework for sql validation

Dhruvi-0802:
### Description

* Drop statements in the upgrade scripts may lead to failure of upgrade due to the dependency of user defined objects. Also for newly created objects, it is necessary that the objects get tested.
* This commit adds sql validation framework which
    1. identifies unexpected drop statements present in the upgrade scripts.
    2. checks that for each create statement in the installation and upgrade scripts, there should be at least some test in the JDBC framework. (Currently supporting view, procedure, function and table.)
* The create object type(function, view etc.) that is searched for can be specified in the config file.
* Expected drop statements and objects that don’t have tests currently can be added in the respective expected files in order to pass the test.
* If unexpected drop statements or objects without tests are found then the framework fails and gives the modified expected files as output.
* More DDL and DML search patterns can be specified using the config parameter searchPatterns in the config file.
 
Signed-off-by: Dhruvi Sheth [sdhruvii@amazon.com](mailto:sdhruvii@amazon.com)

### Check List


test/python/sql_validation.py
```diff
@@ -0,0 +1,402 @@
...
+    patterns = cfg["searchPatterns"].split(',')
+    found = True
+    result = False
+
+    for pattern in patterns:
+        # generate outputfile name
+        filename = "expected_" + re.sub("[^a-zA-Z0-9 ]", "", pattern)
+        
+        # search for the pattern
+        find_pattern_drop(pattern, filename, logger)
```
rishabhtanwar29: I think we should rename this function `find_pattern` or similar since this function is being used for multiple patterns.

test/python/sql_validation.py
```diff
@@ -0,0 +1,402 @@
...
+                if len(re.findall(r"[$]{2}", line, re.I)) == 1:
+                    readflag = not readflag    
+                line = file.readline()
+                
+    logger.info("Found all create objects successfully!")
+    return object_names
+
+
+# find if given pattern exists in the files passed by a list
+def find_obj(files, pattern):
```
rishabhtanwar29: Second argument should be named `object_name` instead of pattern.

shalinilohia50:
Please add details in the description for better context.

test/python/expected/sql_validation_framework/expected_create.out
```diff
@@ -0,0 +1,229 @@
...
+Could not find tests for function babelfish_has_any_privilege
+Could not find tests for function babelfish_remove_delimiter_pair
+Could not find tests for function babelfish_split_object_name
+Could not find tests for function information_schema_tsql._pgtsql_char_max_length
+Could not find tests for function information_schema_tsql._pgtsql_char_octet_length
+Could not find tests for function information_schema_tsql._pgtsql_datetime_precision
+Could not find tests for function information_schema_tsql._pgtsql_numeric_precision
+Could not find tests for function information_schema_tsql._pgtsql_numeric_precision_radix
+Could not find tests for function information_schema_tsql._pgtsql_numeric_scale
+Could not find tests for function information_schema_tsql._pgtsql_truetypid
```
shalinilohia50: These are helper functions starting with `_` and these don't expect relevant tests, can we ignore these functions?

test/python/expected/sql_validation_framework/expected_drop.out
```diff
@@ -0,0 +1,36 @@
...
+Unexpected drop found for procedure sys.babelfish_drop_deprecated_function in file ../../contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--1.2.1--2.0.0.sql
+Unexpected drop found for procedure sys.babelfish_drop_deprecated_function in file ../../contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.1.0--2.2.0.sql
+Unexpected drop found for procedure sys.babelfish_drop_deprecated_object in file ../../contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.0.0--2.1.0.sql
+Unexpected drop found for procedure sys.babelfish_drop_deprecated_view in file ../../contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.1.0--2.2.0.sql
+Unexpected drop found for procedure sys.babelfish_remove_object_from_extension in file ../../contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.1.0--2.2.0.sql
+Unexpected drop found for procedure sys.create_xp_qv_in_master_dbo in file ../../contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--1.1.0--1.2.0.sql
+Unexpected drop found for procedure sys.sp_babelfish_grant_usage_to_all in file ../../contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--1.1.0--1.2.0.sql
```
shalinilohia50: > CREATE OR REPLACE PROCEDURE sys.sp_babelfish_grant_usage_to_all()
> AS $$
> DECLARE
> 	schema_name text;
> BEGIN
> 	FOR schema_name IN SELECT nspname FROM sys.babelfish_namespace_ext
> 	LOOP
> 		EXECUTE format('GRANT USAGE ON SCHEMA %I TO PUBLIC', schema_name);
> 	END LOOP;
> END;
> $$ LANGUAGE plpgsql;
> 
> CALL sys.sp_babelfish_grant_usage_to_all();
> DROP PROCEDURE sys.sp_babelfish_grant_usage_to_all;

I don't see any unexpected drop for the `PROCEDURE sys.sp_babelfish_grant_usage_to_all` here. Why does it say unexpected drop for other procedures too like `create_xp_qv_in_master_dbo`?

rishabhtanwar29: All the listed procedures above are being dropped in the upgrade scripts, i.e.:
```
DROP PROCEDURE sys.sp_babelfish_grant_usage_to_all;
```
So that's why we see an error saying unexpected drop statement found.
But since they all are expected drop statements, we have put them in expected output file.

test/python/expected/sql_validation_framework/expected_create.out
rishabhtanwar29: +1, objects whose name starts with `_` can be ignored.

test/python/expected/sql_validation_framework/expected_drop.out
Dhruvi-0802: The `expected/expected_drop` file contains the drop statements that are expected in the scripts for which the test should pass.
Workflow of framework:
The framework looks for all drop statements in the upgrade scripts and adds them to `output/expected_drop` file. By taking diff of both files, if any drop except those in `expected/expected_drop` is found, the test will fail.

I think it would be better to remove the word unexpected from the file.

rishabhtanwar29 (dismissed):

test/python/sql_validation.py
```diff
@@ -0,0 +1,409 @@
...
+                    if len(re.findall(r"[$]{2}", line, re.I)) == 1:
+                        readflag = not readflag
+                        
+                    line = file.readline()
+    logger.info("Searching {0} statements completed successfully!".format(re.sub("[^a-zA-Z0-9 ]", "", pattern)))
+
+
+
+# list of installation and upgrade scripts
+def list_scripts_create(inpPath):
```
shalinilohia50: Nitpick: Can we rename it to `list_scripts` since it has the list of scripts?

test/python/sql_validation.py
Dhruvi-0802: Updated

test/python/sql_validation.py
Dhruvi-0802: Updated

shalinilohia50 (dismissed):

rishabhtanwar29 (dismissed):

shalinilohia50 (dismissed):

surendravishnoi (commented):
does the framework look for create objects in all JDBC tests or does it specifically look in upgrade scripts for the one which has upgrade req?

Dhruvi-0802:
> does the framework look for create objects in all JDBC tests or does it specifically look in upgrade scripts for the one which has upgrade req?

It looks for tests for all created objects (view, procedure, function, table) in installation and upgrade scripts in the JDBC tests.

surendravishnoi (dismissed):

rishabhtanwar29 (dismissed):

test/python/sql_validation.py
```diff
@@ -0,0 +1,453 @@
...
+    search_tests = set()
+    for filename in sch_files:
+        with  open(filename, "r") as file:
+            line = file.readline()
+
+            while line:
+                line = line.strip()
+
+                # ignoring comments and empty lines in the schedule file
+                if line.startswith("#") or line == '':
```
rishabhtanwar29: We need to skip some more patterns as well which we are skipping here https://github.com/babelfish-for-postgresql/babelfish_extensions/blob/BABEL_2_X_DEV/.github/composite-actions/setup-base-version/action.yml#L81

rishabhtanwar29 (approved):

surendravishnoi (approved):

shalinilohia50 (approved):

kuntalghosh (approved):

PR #333, opened 2022-06-27
Instead of trigger behaves incorrectly when called recursively

forestkeeper:
Add a check when executing INSTEAD OF triggers to see if the trigger has already fired for a given row.

Without this check, an INSTEAD OF trigger that issues it's own DML against the same table would have it's DML also fire the trigger, creating undesired recursion. For example, if an INSTEAD OF trigger on table A did it's own INSERT into table A (instead of the original insert), that new INSERT would fire the INSTEAD OF trigger a second time, resulting in no insert actually taking place.

### Details
recursively calling the instead of trigger may lose the last DML data for the triggered table

Instead of trigger behaves incorrectly in recursive call. It happens when there is recursive call inside an instead of trigger like below:
When DML triggers instead of trigger A_trigger and A_trigger will:
- condition i : it'll execute a DML statement and trigger A_trigger again
- condition ii : it'll execute a DML statement and trigger another B_trigger and inside B_trigger, it meets the condition i or condition ii

The right behaviour is that second call of A_trigger should not be triggered and the second DML should execute as normal DML do. It should not execute like a DML with instead of trigger.

Task: BABEL-3115 & BABEL-3116 & BABEL-3118

### Description

Row inserted by an INSTEAD-OF INSERT trigger is missing in the example below: 

```
-- repro in Babelfish:
1>
2> CREATE TABLE t
3> (
4> c1 int IDENTITY(1,1) PRIMARY KEY,
5> c2 varchar(20) UNIQUE,
6> );
7>
8> INSERT INTO t(c2)
9> VALUES ('Joe'), ('Steve');
10>
11>
12> SELECT * FROM t
13> GO
c1          c2
----------- --------------------
          1 Joe
          2 Steve
1>
2> CREATE TRIGGER TR_t ON t
3> INSTEAD OF INSERT
4> AS
5>  DECLARE @c1 int, @c2 varchar(20)
6> SELECT @c1 = c1, @c2 = c2 FROM inserted
7> INSERT INTO t (c2) VALUES(@c2 + '2') --BUG: This INSERT disappears
8> GO
1>
2> INSERT INTO t(c2) VALUES ('Joe')
3> GO
1>
2> --There should be 3 rows, but there are only 2 rows. The INSERT from the trigger is lost.
3> SELECT * FROM t
4> GO
c1          c2
----------- --------------------
          1 Joe
          2 Steve
```
Explanation: 
- Insert inside instead of trigger causes a recursive trigger execution
- A recursive trigger execution is capped at nesting level of one and should be a no-op.
- We skip both DML and trigger execution but we should execute the original DML as trigger is a no-op.

The same problem will happen for Delete/Update as well.

### Issues Resolved

The fix identifies a recursive instead of trigger and do the DML execution without instead of trigger
 

### Check List

### DIFF
--- a/contrib/babelfishpg_tsql/src/hooks.c
+++ b/contrib/babelfishpg_tsql/src/hooks.c
@@ -92,6 +92,8 @@ static void pltsql_ExecutorRun(QueryDesc *queryDesc, ScanDirection direction, ui
 static void pltsql_ExecutorFinish(QueryDesc *queryDesc);
 static void pltsql_ExecutorEnd(QueryDesc *queryDesc);
 
+static bool plsql_TriggerRecursiveCheck(ResultRelInfo *resultRelInfo);
+
 /*****************************************
  * 			Replication Hooks
  *****************************************/
@@ -194,6 +196,7 @@ InstallExtendedHooks(void)
 
 	prev_inherit_view_constraints_from_table = inherit_view_constraints_from_table_hook;
 	inherit_view_constraints_from_table_hook = preserve_view_constraints_from_base_table;
+	TriggerRecuresiveCheck_hook = plsql_TriggerRecursiveCheck;
 }
 
 void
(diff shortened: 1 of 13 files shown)


nasbyj (changes requested):
Please improve the commit message

kuntalghosh (dismissed):

kuntalghosh (commented):
Please don't check-in with the logfile.

surendravishnoi (changes requested):
Please remove log file and improve commit message

kuntalghosh:
This PR also removes the unnecessary logfile. Once #337 is merged, we just need to resolve the conflict. But, we should improve the commit message.

forestkeeper:
I have updated my commit msg

surendravishnoi (dismissed):

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -327,6 +330,48 @@ pltsql_ExecutorEnd(QueryDesc *queryDesc)
 		standard_ExecutorEnd(queryDesc);
 }
 
+/**
+ * @brief 
+ *  the function will depend on PLtsql_execstate to find whether 
+ *  the trigger is called before on this query stack, so that's why we 
+ *  have to add a hook into Manfred code to callback into babel code,
```
nasbyj: Should not be refering to Manfred in public commits

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -327,6 +330,48 @@ pltsql_ExecutorEnd(QueryDesc *queryDesc)
...
+	if (pltsql_recursive_triggers)
+		return false;
+	int i;
+	PLExecStateCallStack * cur;
+	PLtsql_execstate *estate;
+	cur = exec_state_call_stack;
+	while (cur != NULL){
+		estate = cur->estate;
+		if (estate->trigdata != NULL && estate->trigdata->tg_trigger != NULL
+		&& resultRelInfo->ri_TrigDesc!=NULL
```
nasbyj: `!=` should be wrapped in spaces

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -327,6 +330,48 @@ pltsql_ExecutorEnd(QueryDesc *queryDesc)
...
+	PLtsql_execstate *estate;
+	cur = exec_state_call_stack;
+	while (cur != NULL){
+		estate = cur->estate;
+		if (estate->trigdata != NULL && estate->trigdata->tg_trigger != NULL
+		&& resultRelInfo->ri_TrigDesc!=NULL 
+		&& (resultRelInfo->ri_TrigDesc->trig_insert_instead_statement
+		|| resultRelInfo->ri_TrigDesc->trig_delete_instead_statement 
+		|| resultRelInfo->ri_TrigDesc->trig_update_instead_statement)){
+			for (i =0; i<resultRelInfo->ri_TrigDesc->numtriggers; ++i){
```
nasbyj: extra space before `=0`

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -327,6 +330,48 @@ pltsql_ExecutorEnd(QueryDesc *queryDesc)
 		standard_ExecutorEnd(queryDesc);
 }
 
+/**
+ * @brief 
+ *  the function will depend on PLtsql_execstate to find whether
```
nasbyj, line 335: Should also describe what the function is actually doing... something along the lines of "Returns true if we are about to execute an INSTEAD OF trigger that has already been executed."

nasbyj (dismissed):

nasbyj (commented):
LGTM, but not marking approved since I'm not an expert on the code.

macwoody (commented):
why does your change prints additional row count in the result ?

forestkeeper:
It's because this behavior is different from sql server previously, and I corrected the instead of trigger execution behave and it has done some real changes into the table

surendravishnoi (approved):

kuntalghosh (approved):

PR #336, opened 2022-06-28
Support Cross-DB Stored Procedure Call

shalinilohia50:
This commit uses ANTLR to confirm if the query is cross-database.
It will then set a flag in the statement's execution object and
check that flag during the executor stage. If true, the current user
will be temporarily set to the Login to enable permissions for the
statement execution. Otherwise, permission checks will prevent the
cross-database statement from executing.

Task: BABEL-3275


surendravishnoi (approved):
Please validate that there are no new test failures with this change

shalinilohia50:
> Please validate that there are no new test failures with this change

I have verified, there is no new failure due to my change.

macwoody (approved):

PR #347, opened 2022-06-29
Modified sys.databases' compatibility level value

faviansamatha:
### Description

Previously, sys.databases.compatibility_level had a value of NULL which caused issues when right-clicking on a user database in SSMS. This commit adds a hardcoded value of 120 for the compatibility_level column in sys.databases. Furthermore, it also casts the columns to the documented data type.

Task: BABELFISH-477

### Check List


shalinilohia50 (dismissed):

xhfanhe (dismissed):

test/JDBC/input/views/sys-database_mirroring-vu-prepare.sql
```diff
@@ -0,0 +1,15 @@
+CREATE VIEW sys_database_mirroring_view AS
```
surendravishnoi: please use name like sys_database_mirroring_vu_prepare_view, this is consistent as it uses file name as prefix and will not conflict with non-uprgrade tests. Applies to all object created in vu-prepare scripts.

Deepesh125 (dismissed):

test/JDBC/input/views/sys-database_mirroring-vu-prepare.sql
faviansamatha: Sounds good, will do!

test/JDBC/sql_expected/sys-databases.out
```diff
@@ -15,7 +15,7 @@ int
 SELECT name FROM sys.databases where name = 'my_test_database';
```
Deepesh125: Why is this file present under sql_expected? Is it intentional?

test/JDBC/expected/BABEL-SYS-DATABASES.out
```diff
@@ -18,19 +18,19 @@ GO
 SELECT name, compatibility_level, collation_name FROM sys.databases ORDER BY name;
```
Deepesh125, line 18: Is it possible to add these tests as prepare/verify/cleanup format? If so then we should do that instead of maintaining multiple files for possibly same test cases.

surendravishnoi (dismissed):
Looks good for upgrade tests

test/JDBC/sql_expected/sys-databases.out
faviansamatha: Not sure why, the original location for sys-databases.out was originally here (maybe from the first implementer?). 

I will  remove this in the next commit and the sys-database test since this scenario is already tested in the prepare/verify/cleanup format.

test/JDBC/expected/BABEL-SYS-DATABASES.out
faviansamatha: Yea, I'll try to do that

faviansamatha: Tried this a bit, but not sure if it's possible or a good idea since it enables multi-db and switching to single-db after. The issue with this test is that it switches back to single-db is causing an issue. I think we should keep the test as a separate file since it does test that sys.database displays multiple user databases in multi-db mode.

 Correct me if I'm wrong though with these following points on why prepare/verify/cleanup format doesn't work or if there's a workaround for this scenario:

- If we keep it as is, (switch back to single-db) it will turn off multi-db option which would affect other upgrade tests in the upgrade test framework causing them to fail as multiple dbs are used. 

- If we switch back to single-db, it might cause problems for other tests in the regular JDBC since they run under single-db mode. 

- If we choose to keep it separate by switching back to single-db in prepare/verify/cleanup format in the input folder, but leave it as multi-db for all the version upgrade path, then it would fail when 14_latest schedule runs for the next version since we would switch back to single-db mode when 14_latest -> 14_latest+1 (making the test only effective for a regular non upgrade paths which is why we should keep it as is).

Deepesh125: Okay, Let it be as it is for the time being.

Deepesh125 (approved):

hsuamz (approved):

rossd-bitquill (approved):

PR #365, opened 2022-07-06
Support HOST_NAME() T-SQL and show client hostname in sys.sysprocesses

thephantomthief:
### Description

This commit adds support for HOST_NAME() T-SQL function. Now any client
connected to Babelfish with a hostname specified ("-H" option in sqlcmd
and "workstationID" property in JDBC driver for example) can now see the
same through the HOST_NAME() T-SQL function.

Also, the same hostname will now be visible in the sys.sysprocesses
T-SQL view.

Added relevant test cases for both normal and upgrade scenarios.

Task: BABEL-1608/2887

### Check List


test/JDBC/input/functions/sys-host_name-vu-prepare.sql
```diff
@@ -0,0 +1,12 @@
...
+GO
+
+CREATE FUNCTION host_name_func() 
+RETURNS NCHAR(128)
+AS
+BEGIN
+        DECLARE @hostname NCHAR(128);
+        SELECT @hostname = hostname FROM sys.sysprocesses WHERE spid = @@SPID;
+        RETURN @hostname;
+END
```
surendravishnoi: we can add a procedure as well

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.1.0--2.2.0.sql
```diff
@@ -1400,6 +1400,193 @@ INSERT INTO sys.babelfish_helpcollation VALUES (N'mongolian_cs_as', N'Mongolian,
...
+    , null::bigint as unsuccessful_logons
+    , null::int as group_id
+    , d.database_id::smallint as database_id
+    , 0 as authenticating_database_id
+    , d.trancount as open_transaction_count
+  from pg_catalog.pg_stat_activity AS a
+  RIGHT JOIN sys.tsql_stat_get_activity('sessions') AS d ON (a.pid = d.procid);
+  GRANT SELECT ON sys.dm_exec_sessions TO PUBLIC;
+
+create or replace view sys.dm_exec_connections
```
surendravishnoi, line 2684: do we need to recreate this view if it is not selecting hostname?

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.1.0--2.2.0.sql
```diff
@@ -1400,6 +1400,193 @@ INSERT INTO sys.babelfish_helpcollation VALUES (N'mongolian_cs_as', N'Mongolian,
...
+ left join pg_catalog.pg_stat_activity blocking_activity ON blocking_activity.pid = blocking_locks.pid
+ where a.datname = current_database(); /* current physical database will always be babelfish database */
+GRANT SELECT ON sys.sysprocesses TO PUBLIC;
+
+-- recreate views dependent on tsql_stat_get_activity
+create or replace view sys.dm_exec_sessions
+  as
+  select a.pid as session_id
+    , a.backend_start::sys.datetime as login_time
+    , a.client_hostname::sys.nvarchar(128) as host_name
```
surendravishnoi: if we are not changing host_name selection, do we need to recreate the view

test/JDBC/expected/sys-host_name-vu-prepare.out
```diff
@@ -0,0 +1,12 @@
+CREATE VIEW host_name_view AS SELECT DISTINCT hostname FROM sys.sysprocesses WHERE spid = @@SPID
```
surendravishnoi: we should add tests for dm_exec_session and connection as well

test/JDBC/input/functions/sys-host_name-vu-prepare.sql
thephantomthief: Added procedure as well

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.1.0--2.2.0.sql
thephantomthief: Just to be safe, I am recreating this since underlying function tsql_stat_get_activity is modified.

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.1.0--2.2.0.sql
thephantomthief: I had forgotten make sys.dm_exec_sessions use new host_name. Made that change now.

test/JDBC/expected/sys-host_name-vu-prepare.out
thephantomthief: Added tests for the same.

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.1.0--2.2.0.sql
```diff
@@ -1400,6 +1400,193 @@ INSERT INTO sys.babelfish_helpcollation VALUES (N'mongolian_cs_as', N'Mongolian,
...
+
+CREATE OR REPLACE FUNCTION sys.host_name()
+RETURNS sys.NVARCHAR(128)  AS 'babelfishpg_tsql' LANGUAGE C IMMUTABLE PARALLEL SAFE;
+
+ALTER FUNCTION sys.tsql_stat_get_activity(text) RENAME TO tsql_stat_get_activity_deprecated_2_1;
+ALTER VIEW sys.sysprocesses RENAME TO sysprocesses_deprecated_2_1;
+
+-- deprecate views dependent on tsql_stat_get_activity
+ALTER VIEW sys.dm_exec_sessions RENAME TO dm_exec_sessions_deprecated_2_1;
+ALTER VIEW sys.dm_exec_connections RENAME TO dm_exec_connections_deprecated_2_1;
```
Deepesh125: I don't think we need to do this. We can simply re-create dependent objects.

test/JDBC/expected/sys-host_name-vu-verify.out
```diff
@@ -0,0 +1,83 @@
...
+nvarchar#!#nvarchar#!#int#!#int#!#nvarchar
+TCP#!#TSQL#!#1946157060#!#4#!#FALSE
+~~END~~
+
+
+SELECT * from dm_exec_sessions_func(@@SPID)
+GO
+~~START~~
+nvarchar#!#nvarchar#!#int#!#nvarchar#!#nvarchar#!#nvarchar#!#smallint
+us_english#!#a_nice_workstation_name#!#131592#!#Microsoft JDBC Driver 8.2#!#Microsoft JDBC Driver for SQL Server#!#ymd#!#7
```
Deepesh125: Version of JDBC might not be static value. So we need to restructure this test case.

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -41,7 +41,7 @@
 #include "../src/collation.h"
 #include "../src/rolecmds.h"
 
-#define TSQL_STAT_GET_ACTIVITY_COLS 24
+#define TSQL_STAT_GET_ACTIVITY_COLS 25
```
Deepesh125, line 44: We need to deprecate sysprocesses which is dependent on tsql_stat_get_activity. So if any cx object is dependent on sysprocesses then it would still point to deprecated version of sysprocesses after which is expecting tsql_stat_get_activity to return 24 columns and hence it would be failed. 

Alternate way to achieve this is to deprecate C function definition also which would be referenced by deprecated version of sysprocesses.

thephantomthief: done

test/JDBC/expected/sys-host_name-vu-verify.out
thephantomthief: done

test/JDBC/upgrade/13_6/verification_cleanup/13_4/sys-host_name-vu-verify.sql
```diff
@@ -0,0 +1,2 @@
+SELECT DISTINCT hostname FROM sys.sysprocesses WHERE spid = @@SPID
```
Deepesh125, line 1: Corresponding expected file in not checked in?

contrib/babelfishpg_tsql/sql/sys_views.sql
```diff
@@ -962,7 +962,7 @@ select
   , 0 as open_tran
   , a.state as status
   , null::bytea as sid
-  , a.client_hostname as hostname
+  , t.host_name::sys.nchar(128) as hostname
```
Deepesh125: Use tsql specific syntax.

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.1.0--2.2.0.sql
```diff
@@ -1569,6 +1569,361 @@ INSERT INTO sys.babelfish_helpcollation VALUES (N'mongolian_cs_as', N'Mongolian,
...
```
Deepesh125: why do we need this? Simply recreating these objects would be sufficient

Deepesh125:
why isn't new testcase file added under 14_latest/schedule ?

test/JDBC/upgrade/13_6/verification_cleanup/13_4/sys-host_name-vu-verify.sql
thephantomthief: Please check test/JDBC/expected/13_6__verification_cleanup__13_4__sys-host_name-vu-verify.out

thephantomthief:
@Deepesh125 it shouldn't be added to `14_latest/schedule`. **14_latest** is not the base version for any upgrade. If we want to run a test from `input/` in an upgrade path like **base_version -> ... -> final_version** then the file name prefix should be added to `<base_version>/schedule` and not `<final_version>/schedule`

I have added that entry in 14_latest/schedule since it will be needed once we test future upgrades from 14.latest

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.1.0--2.2.0.sql
thephantomthief: Won't objects already dependent on these views be affected if we simply re-create them?

Deepesh125 (dismissed):

kuntalghosh (dismissed):

test/JDBC/input/functions/sys-host_name-vu-prepare.sql
```diff
@@ -0,0 +1,42 @@
+CREATE VIEW host_name_view AS SELECT DISTINCT hostname FROM sys.sysprocesses WHERE spid = @@SPID
```
surendravishnoi: please use filename_xxx as object name. Also, for release where input prepare file can be used, add it to schedule file instead of creating a copy in release directory. If we need changes for specific release, copy is fine.

thephantomthief: Actually if any one of the prepare, verify and cleanup files' output is different across releases, then we have to add all the three files in specific release folder. So in my case, prepare file is common as you mentioned. If I add an entry in `13_6/schedule` for `sys-host_name`, the framework will assume that file `sys-host_name-vu-verify.sql` and its output will match with the common expected output i.e. framework will use output file `sys-host_name-vu-verify.out` instead of using a version specific expected output `14_latest__verification_cleanup__13_6__sys-host_name-vu-verify.out`.

surendravishnoi: that is fine, if we have difference in behaviuor/output across releases, we can add a copy

surendravishnoi (approved):

PR #366, opened 2022-07-07
BCP in - command fails when the row count is ~20/40 million records

KushaalShroff:
Issue: BCP sends all of the Bulk Load Data in a single request. With the Current Design of TDS protocol, we read the entire request first and only then process it. The entire request is read in one StringInfoData which persists till end of fetching phase and we also create value arrays for each row which persists till end of processing. One single batch with a huge data was leading to an Out Of Memory scenario.

To Fix this issue we bypass the tdscomm layer for BULK LOAD Request and read the next frame as and when required. We then moved the logic for parsing the rows to the TDS Process phase where we read only a certain number of bytes or number of rows (whichever is lower). This way we have optimised the memory consumption at the TDS Layer. NOTE: We might still run into Out Of Memory Issue at even higher ends of data transfer because ultimately all of that data will be buffered at the socket waiting to be read by TDS process.

We introduce 2 GUCS that the customer can tune:
a: babelfishpg_tds.insert_bulk_rows_per_batch which sets the number of rows per batch to be processed for a single batch
b: babelfishpg_tds.insert_bulk_kilobytes_per_batch which sets the number of bytes per batch to be processed for a single batch

Task: BABEL-3244
Signed-off-by: Kushaal Shroff ([kushaal@amazon.com](mailto:kushaal@amazon.com))

### Description

### Issues Resolved

### Check List

### DIFF
--- a/contrib/babelfishpg_tds/src/backend/tds/guc.c
+++ b/contrib/babelfishpg_tds/src/backend/tds/guc.c
@@ -25,6 +25,9 @@
 #include "src/include/faultinjection.h"
 #include "src/include/guc.h"
 
+#define DEFAULT_INSERT_BULK_ROWS_PER_BATCH 1000
+#define DEFAULT_INSERT_BULK_PACKET_SIZE 8
+
 /* Global variables */
 int	pe_port;
 char	*pe_listen_addrs = NULL;
@@ -39,8 +42,8 @@ bool	tds_ssl_encrypt = false;
 int 	tds_default_protocol_version = 0;
 int32_t tds_default_packet_size = 4096;
 int	tds_debug_log_level = 1;
-int insert_bulk_rows_per_batch = 1000;
-int insert_bulk_kilobytes_per_batch = 8;
+int insert_bulk_rows_per_batch = DEFAULT_INSERT_BULK_ROWS_PER_BATCH;
+int insert_bulk_kilobytes_per_batch = DEFAULT_INSERT_BULK_PACKET_SIZE;
 #ifdef FAULT_INJECTOR
 static bool TdsFaultInjectionEnabled = false;
 #endif
@@ -280,7 +283,7 @@ TdsDefineGucs(void)
 		gettext_noop("Sets the number of rows per batch to be processed for Insert Bulk"),
 		NULL,
 		&insert_bulk_rows_per_batch,
-		1000, 1, INT_MAX,
+		DEFAULT_INSERT_BULK_ROWS_PER_BATCH, 1, INT_MAX,
 		PGC_SIGHUP,
 		GUC_NOT_IN_SAMPLE,
 		NULL,
@@ -292,7 +295,7 @@ TdsDefineGucs(void)
 		gettext_noop("Sets the number of bytes per batch to be processed for Insert Bulk"),
 		NULL,
 		&insert_bulk_kilobytes_per_batch,
-		8, 1, INT_MAX,
+		DEFAULT_INSERT_BULK_PACKET_SIZE, 1, INT_MAX,
 		PGC_SIGHUP,
 		GUC_NOT_IN_SAMPLE,
 		TdsGucDefaultPacketSizeCheck,
--- a/contrib/babelfishpg_tds/src/backend/tds/guc.c
+++ b/contrib/babelfishpg_tds/src/backend/tds/guc.c
@@ -291,7 +291,7 @@ TdsDefineGucs(void)
 		"babelfishpg_tds.insert_bulk_kilobytes_per_batch",
 		gettext_noop("Sets the number of bytes per batch to be processed for Insert Bulk"),
 		NULL,
-		&tds_default_packet_size,
+		&insert_bulk_kilobytes_per_batch,
 		8, 1, INT_MAX,
 		PGC_SIGHUP,
 		GUC_NOT_IN_SAMPLE,
--- a/contrib/babelfishpg_tds/src/backend/tds/guc.c
+++ b/contrib/babelfishpg_tds/src/backend/tds/guc.c
@@ -39,6 +39,8 @@ bool	tds_ssl_encrypt = false;
 int 	tds_default_protocol_version = 0;
 int32_t tds_default_packet_size = 4096;
 int	tds_debug_log_level = 1;
+int insert_bulk_rows_per_batch = 1000;
+int insert_bulk_kilobytes_per_batch = 8;
 #ifdef FAULT_INJECTOR
 static bool TdsFaultInjectionEnabled = false;
 #endif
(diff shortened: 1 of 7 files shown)


contrib/babelfishpg_tds/src/backend/tds/guc.c
```diff
@@ -39,6 +39,8 @@ bool	tds_ssl_encrypt = false;
 int 	tds_default_protocol_version = 0;
 int32_t tds_default_packet_size = 4096;
 int	tds_debug_log_level = 1;
+int insert_bulk_rows_per_batch = 1000;
```
surendravishnoi: use macro for constants

contrib/babelfishpg_tds/src/backend/tds/tdsbulkload.c
```diff
@@ -28,8 +28,20 @@
 #include "src/include/tds_response.h"
 #include "src/include/tds_typeio.h"
 
-static void SetBulkLoadRowData(TDSRequestBulkLoad request, const StringInfo message, uint64_t offset);
+static StringInfo SetBulkLoadRowData(TDSRequestBulkLoad request, StringInfo message);
 void ProcessBCPRequest(TDSRequest request);
+static void FetchMoreBcpData(StringInfo *message);
+static int ReadBcpPlp(ParameterToken temp, StringInfo *message, TDSRequestBulkLoad request);
+uint64_t offset = 0;
```
surendravishnoi, line 35: instead of offset, use name with context (as a prefix to _offset) to avoid variable name clash

contrib/babelfishpg_tds/src/backend/tds/tdsbulkload.c
```diff
@@ -59,6 +70,61 @@ do \
...
+	HOLD_CANCEL_INTERRUPTS();
+	BulkCopyPacketStatus = TdsReadNextPendingBcpRequest(temp);
+	RESUME_CANCEL_INTERRUPTS();
+
+	if (BulkCopyPacketStatus < 0)
+	{
+		TdsErrorContext->reqType = 0;
+		TdsErrorContext->err_text = "EOF on TDS socket while fetching For Bulk Load Request";
+		pfree((*message)->data);
+		pfree(temp);
```
surendravishnoi, line 121: should free temp->data as well

contrib/babelfishpg_tds/src/backend/tds/tdsbulkload.c
```diff
@@ -59,6 +70,61 @@ do \
...
+		ereport(ERROR,
+				(errcode(ERRCODE_PROTOCOL_VIOLATION),
+					errmsg("EOF on TDS socket while fetching For Bulk Load Request")));
+		return;
+	}
+
+	if ((*message))
```
surendravishnoi: why are we checking for *message here but not at line 96, can it change  in between?

contrib/babelfishpg_tds/src/backend/tds/tdscomm.c
```diff
@@ -893,6 +937,11 @@ TdsReadNextRequest(StringInfo message, uint8_t *status, uint8_t *messageType)
 				TdsDoProcessHeader = true;
 			return 0;
 		}
+		if (TdsRecvMessageType == TDS_BULK_LOAD)
+		{
+			TdsDoProcessHeader = true;
```
surendravishnoi, line 948: why are we doing this and at this particular place? Can we do it in TdsReadNextPendingBcpRequest?

KushaalShroff: Because for Bulk Load Request we want to stop reading after the first packet.
Later we read the pending packets using TdsReadNextPendingBcpRequest whenever required

contrib/babelfishpg_tds/src/backend/tds/guc.c
KushaalShroff: This is a GUC

contrib/babelfishpg_tds/src/backend/tds/tdsbulkload.c
KushaalShroff: No it won't, will move it up

contrib/babelfishpg_tds/src/backend/tds/guc.c
surendravishnoi: I meant macro for values (1000)

contrib/babelfishpg_tds/src/backend/tds/tdscomm.c
surendravishnoi: but can the logic be driven from TdsReadNextPendingBcpRequest? why we need to do it here only?

contrib/babelfishpg_tds/src/backend/tds/tdsbulkload.c
KushaalShroff: There is no variable name clash with offset, is it necessary to rename it? (The diff will be throughout the file then)

contrib/babelfishpg_tds/src/backend/tds/tdscomm.c
KushaalShroff: From GetTdsRequest Function we will call this "TdsReadNextRequest" function. If we do not do it here then we will read the entire BCP Request which we do not want. We instead  we read only the first packet only and then process the headers for that

contrib/babelfishpg_tds/src/backend/tds/tdsbulkload.c
```diff
@@ -94,6 +94,12 @@ FetchMoreBcpData(StringInfo *message)
 
 	temp = makeStringInfo();
 	appendBinaryStringInfo(temp, (*message)->data + offset, (*message)->len - offset);
+	if ((*message))
```
surendravishnoi: still the same question, why are we checking *message here for NULL but not in line 96

contrib/babelfishpg_tds/src/backend/tds/tdscomm.c
```diff
@@ -886,13 +930,27 @@ TdsReadNextRequest(StringInfo message, uint8_t *status, uint8_t *messageType)
 		if (TdsGetbytes(message->data + message->len, readBytes))
 			return EOF;
 		message->len += readBytes;
+
+		if (TdsRecvMessageType == TDS_BULK_LOAD)
+		{
+			/* if this is the last packet then notify the caller. */
+			if (TdsRecvPacketStatus & TDS_PACKET_HEADER_STATUS_EOM)
+				BulkCopyPacketStatus = 1;
```
surendravishnoi: this is not the right approach to set a global variable here. Better would to have a new function(or use existing one) to enquire for header status from bulk copy code

contrib/babelfishpg_tds/src/backend/tds/tdscomm.c
```diff
@@ -886,13 +930,27 @@ TdsReadNextRequest(StringInfo message, uint8_t *status, uint8_t *messageType)
...
+			TdsDoProcessHeader = true;
```
surendravishnoi: if we remove this whole condition and enquire packet status separately, is this needed? wouldn't 946 will work for us

KushaalShroff: No we will still need to return after reading the first packet only and NOT the entire request. If you see for line 946 to be executed it needs to be the last packet of that request

contrib/babelfishpg_tds/src/backend/tds/tdsbulkload.c
KushaalShroff: It is unlikely that message is NULL, I have moved the check at the top. But pfree must happen after 96

surendravishnoi (approved):

sangilsong (approved):

PR #375, opened 2022-07-11
Throw error if typmod is -1 for non-FuncExpr in TdsGetGenericTypmod

thephantomthief:
### Description

Previously, TdsGetGenericTypmod assumed that it will always get an
FuncExpr as an argument. However, TdsGetGenericTypmod is called whenever
an Expr typmod is -1 for CHAR/NCHAR data type. Thus, if an Expr was not
of FuncExpr type, it was incorrectly casting it to FuncExpr causing the
backend to crash.

In this commit, we added a default case in TdsGetGenericTypmod for all
non FuncExpr type Expr to throw an error suggesting user to use an
explicit CAST or CONVERT to provide Babelfish a deterministic string
size for CHAR/NCHAR data type. Ideally, we should figure out the correct
typmod for these data types in the engine or babelfishpg_tsql extension
itself.

Task: BABEL-2998/3348

### Check List

### DIFF
--- a/contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
+++ b/contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
@@ -1267,24 +1267,46 @@ SendColInfoToken(int natts, bool sendRowStat)
 static
 int TdsGetGenericTypmod(Node *expr)
 {
-	FuncExpr *func;
-	Oid     func_oid = InvalidOid;
 	int rettypmod = -1;
 
 	if (!expr)
 		return rettypmod;
-	func = (FuncExpr *) expr;
 
-	/*
-	 * Look up the return type typmod from a persistent
-	 * store using function oid.
-	 */
-	func_oid = func->funcid;
-	Assert(func_oid != InvalidOid);
+	switch(nodeTag(expr))
+	{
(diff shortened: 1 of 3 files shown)


kuntalghosh (dismissed):

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
```diff
@@ -1267,24 +1267,46 @@ SendColInfoToken(int natts, bool sendRowStat)
...
-	func_oid = func->funcid;
-	Assert(func_oid != InvalidOid);
+	switch(nodeTag(expr))
+	{
+		case T_FuncExpr:
+			{
+				FuncExpr *func;
+				Oid     func_oid = InvalidOid;
+
+				if (!expr)
```
surendravishnoi: unnecessary, already checked above

surendravishnoi (dismissed):

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
thephantomthief: removed the check

surendravishnoi (approved):

kuntalghosh (approved):

PR #383, opened 2022-07-12
Fixed the DATEADD() function fails when called with the 'ms' or 'mcs' unit

ZhenyeLee:
Fixed the DATEADD() function fails when called with the 'ms' or 'mcs' unit

Task:BABEL-3360
Signed-off-by: Zhenye Li <zhenye@amazon.com>

### Description
[BABEL-3360] The DATEADD() function fails when called with the 'ms' , 'millisecond' , 'mcs' or 'microsecond' unit. Since the second in sys is ``integer`` and it  will throw an error when integer * numeric(which is 0.001 or 0.000001 in each case) for operations. To fix this bug, integer is casting into ``DECIMAL`` and then do the operations.

[MANFRED-13687]Fixed numeric field overflow error when casting int to numeric number. It was ``CAST(num AS DECIMAL(2,0))``, and it failed when int number is larger than 99. Now it is ``CAST(num AS DECIMAL(20,10))``, which can handle 10bits Integer.

### Issues Resolved
[BABEL-3360] https://jira.rds.a2z.com/browse/BABEL-3360
CR: https://code.amazon.com/reviews/CR-72495501 (merged)
[MANFRED-13687]https://jira.rds.a2z.com/browse/MANFRED-13687
CR:https://code.amazon.com/reviews/CR-72848357

### Check List


sangilsong (dismissed):

kuntalghosh (commented):
we should add test cases to show the difference of the behaviour between different versions. I.e., DATEADD() was failing for  'ms' , 'millisecond' , 'mcs' or 'microsecond' unit in previous version while it works fine with latest.

contrib/babelfishpg_tsql/sql/sys_functions.sql
```diff
@@ -1037,9 +1037,9 @@ BEGIN
 	WHEN 'second' THEN
 		RETURN startdate OPERATOR(sys.+) make_interval(secs => num);
 	WHEN 'millisecond' THEN
-		RETURN startdate OPERATOR(sys.+) make_interval(secs => num * 0.001);
+		RETURN startdate OPERATOR(sys.+) make_interval(secs => CAST(num AS DECIMAL(20,10)) * 0.001);
 	WHEN 'microsecond' THEN
-		RETURN startdate OPERATOR(sys.+) make_interval(secs => num * 0.000001);
```
2jungkook: Do we support microsecond even though T-SQL does not?

2jungkook:
> we should add test cases to show the difference of the behaviour between different versions. I.e., DATEADD() was failing for 'ms' , 'millisecond' , 'mcs' or 'microsecond' unit in previous version while it works fine with latest.

+1 
We are trying to add upgrade tests as much as possible. For example, in `upgrade/13_6` directory, we can define a view like
```
create view v1 as (select dateadd(millisecond, 56, cast('2016-12-26 23:29:29' as datetime2)));
```
In 13.6, `select * from v1` throws an error, but in 14_latest, it will work as expected.

2jungkook:
Please remove all internal links in the description such as Jira, CRs, etc., before merge.

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.1.0--2.2.0.sql
```diff
@@ -1397,6 +1397,92 @@ from sys.table_types tt
 ) ot;
 GRANT SELECT ON sys.all_objects TO PUBLIC;
 
+
+CREATE OR REPLACE FUNCTION sys.dateadd_internal_df(IN datepart PG_CATALOG.TEXT, IN num INTEGER, IN startdate datetimeoffset) RETURNS datetimeoffset AS $$
```
surendravishnoi: we should add upgrade tests which validate dependency for these functions as well

PR #410, opened 2022-07-19
Allow Babelfish to display the 7th digit for DATE, DATETIME2, and DATETIMEOFFSET datatypes

Sairakan:
PG only allows time-related datatypes to have a maximum scale/precision of 6. However,
SQL Server allows for TIME, DATETIME2, and DATETIMEOFFSET datatypes to be specified with
a scale of 7. Currently, we just truncate these datatypes to use a scale of 6 in order to
allow them to be stored using PG's native timestamp datatype.

Ideally, we would create new datatypes to support the required scale. However, this would
be quite a bit of work and would also introduce complications for upgrading existing
customers who already have tables that would need to be updated with the new datatypes.

Since that is a lot of work for a precision that does not seem to be vitally important
for customer use cases, for now we will opt for the much simpler solution of keeping
the current storage solution, and just modifying the output functions to return 7 decimal
places. We will just return a 0 in the 7th decimal place. For example, '00:00:00.0000123'
would instead be displayed as '00:00:00.0000120'

This particular implementation takes advantage of the fact that for TIME, DATETIME2, and
DATETIMEOFFSET, if no scale is specified, then the default will be 7. Therefore, we can
treat TIME(7) the same as TIME, which then sets the typmod for the column to -1. PG will
appropriately store the timestamp values with its default precision of 6, and then in the
TDS reponse code we can detect that the typmod is -1 and set the scale to 7 for the TDS
receiver.

### Description

Preliminary support for DATE(7), DATETIME(7), and DATETIMEOFFSET(7)
 
### Issues Resolved

BABEL-2934

### Check List


contrib/babelfishpg_tds/src/backend/tds/tdstimestamp.c
```diff
@@ -423,7 +423,9 @@ TdsGetDayTimeFromTimestamp(Timestamp value, uint32 *numDays, uint64 *numSec,
 			((double)fsec/ 1000000));
 	while (scale--)
 		res *= 10;
-
+	/* Round res to the nearest integer */
```
2jungkook, line 426: Is there any special reason to round up?

contrib/babelfishpg_tsql/src/tsqlIface.cpp
```diff
@@ -5086,6 +5086,18 @@ post_process_column_definition(TSqlParser::Column_definitionContext *ctx, PLtsql
 	if (ctx->TIMESTAMP())
 		rewritten_query_fragment.emplace(std::make_pair(ctx->TIMESTAMP()->getSymbol()->getStartIndex(), std::make_pair(::getFullText(ctx->TIMESTAMP()), "timestamp " + ::getFullText(ctx->TIMESTAMP()))));
 
+ 	/*
+	* PG doesn't allow for TIME/DATETIME2/DATETIMEOFFSET to be declared with precision 7, but this is permitted in TSQL.
+	* In order to get around this, remove the scale factor so that the typmod is set to -1 (default). Luckily,
+	* in TSQL the default scale is also 7, so we can re-add the decimal digits to meet the scale factor on the return side.
+	*/
+	if (pg_strncasecmp(::getFullText(ctx->data_type()).c_str(), "TIME(7)", 7) == 0)
```
2jungkook, line 5147: Can this cover the following cases? 
- `select cast('12:15:04.1234567' as TIME(7)`
- TIME(7) in function input/output parameters

If so, please add test cases.

test/JDBC/init.sh
```diff
@@ -1,7 +1,7 @@
  
 #create test user and database from psql terminal
 echo "============================== CREATING USER AND DATABASE =============================="
-psql -U "$USER" -d postgres -a << EOF
+sudo ~/postgres/bin/psql -U "$USER" -d postgres -a << EOF
```
2jungkook: Was there an issue here? If we change this, then it enforces every developers/github actions to have `~/postgres` directory.

2jungkook:
Please add MVU (major version upgrade) tests as well. There can be two different use-cases: `13.x -> 14.latest` and `14.latest -> 14.latest`. 

1) 13.x -> 14.latest
- Need to check how TIME datatype w/o typmod is migrated for tables, views and functions.

2) 14.latest -> 14.latest
- Need to check TIME(7) datatype is well-migrated. Note that MVU is done by postgres dialect.

contrib/babelfishpg_tds/src/backend/tds/tdstimestamp.c
Sairakan: There is an issue that was also covered in the JIRA where occasionally some dates would not get displayed correctly. I determined that this was due to errors in floating point arithmetic that would for example, change "0.000012" to "0.000011999998" internally. When this gets converted back to an int64 for display, it gets truncated to "0.000011". This is why I add 0.5 prior to casting to an int; this has the effect of having the double rounded to the nearest integer instead of just getting floored down all of the time.

test/JDBC/init.sh
Sairakan: Oh sorry, I didn't mean to include this in the commit. I will change it back.

contrib/babelfishpg_tsql/src/tsqlIface.cpp
Sairakan: I just tested it and it works. Will include in the next revision

test/JDBC/input/BABEL-2934-vu-prepare.sql
```diff
@@ -0,0 +1,38 @@
+CREATE TABLE dbo.babel2934_1 (c TIME(7) NULL, d TIME(7) NULL)
```
surendravishnoi: recommended object naming convention is filename_xxx, so it should be BABEL-2934-vu-prepare_1. This is easy to follow and consistent across the board.

test/JDBC/input/BABEL-2934.sql
```diff
@@ -0,0 +1,50 @@
+CREATE TABLE dbo.t7 (c TIME(7) NULL, d TIME(7) NULL)
```
surendravishnoi: same here, filename_xxx

test/JDBC/expected/BABEL-1000.out
```diff
@@ -123,15 +123,15 @@ SELECT * FROM babel_1000_test6(CAST('2030-05-06 13:59:29.123456 -8:00' AS dateti
 GO
 ~~START~~
 datetimeoffset#!#datetime2
-2030-05-06 13:59:29.120000 -08:00#!#1235-01-01 00:00:00.000000
+2030-05-06 13:59:29.1200000 -08:00#!#1235-01-01 00:00:00.0000000
```
surendravishnoi, line 126: For all the tests where expected output is changing, it is a good practice to validate the results against SQL server. Framework provides easy way to run these tests against SQL server. This is specifically imp for such +-1 issues. Please ignore if already done.

Sairakan: Yeah, I tested these changes against SQL Server already, and there is a small but intended difference between babelfish and SQL Server (last digit is always 0 for babelfish)

test/JDBC/input/BABEL-2934-vu-prepare.sql
Sairakan: thanks, I'll make this change on the next revision

test/JDBC/expected/BABEL-1000.out
surendravishnoi: I assume this difference is discussed, agreed upon and documented.

Sairakan: Yes, this is the solution that was agreed upon with Bill and Rob.

test/JDBC/expected/BABEL-2934.out
```diff
@@ -0,0 +1,91 @@
...
+
+
+drop table BABEL_2934
+go
+
+select cast('12:15:04.1234567' as TIME(7))
+go
+~~START~~
+time
+12:15:04.123457
```
2jungkook, line 89: Is this expected?

Also, please add the following cases to MVU tests:
1. `create view v1 as select cast('12:15:04.1234567' as TIME(7))` and `select * from v1`
2. `create function f1 (TIME(7), ...) returns (TIME(7)) ...` and `select f1(...)`
Those are to ensure how customer-created views/functions are impacted by this change during MVU.

Sairakan: This is not ideal, but for now it is expected as there is no simple way to override PG's method of converting time strings into datetime values. I will work on the two extra MVU tests

test/JDBC/expected/13_5__preparation__BABEL-2934-vu-prepare.out
```diff
@@ -0,0 +1,60 @@
+CREATE TABLE BABEL_2934_vu_1 (c TIME(7) NULL, d TIME(7) NULL)
```
2jungkook: I'd suggest to remove unused output files.

2jungkook (dismissed):

JimFinnerty (approved):

2jungkook (approved):

PR #433, opened 2022-07-22
Incorrect schema name resolution for SQL objects inside procedures

shalinilohia50:
For DDL (CREATE TABLE), an object created in a proc should be in the default schema for the caller of the proc.

For DML (SELECT), the table will be resolved to the schema that the proc is in. If the object doesn’t exist in that schema, it will resolve to the dbo schema.

When an object is referenced without a schema name inside a procedure, used the following rules to resolve the schema name.

1. Search for the object in the same schema as the SQL object in which the reference occurs.
2. If not found, search in the dbo schema

Task: BABEL-2992

### Description

### Issues Resolved

### Check List


xhfanhe (commented):
Please add the test to all versions' upgrade test schedule

xhfanhe (dismissed):

shalinilohia50:
> Please add the test to all versions' upgrade test schedule

Added in the new commit.

xhfanhe (dismissed):

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -673,6 +678,10 @@ exec_stmt_exec(PLtsql_execstate *estate, PLtsql_stmt_exec *stmt)
 				need_path_reset = true;
 			}
 	}
+	if (stmt->schema_name != '\0')
+	 	estate->schema_name = stmt->schema_name;
```
surendravishnoi, line 682: do we need to rest the db_name and schema_name once exec is done?

contrib/babelfishpg_tsql/src/pl_exec.c
```diff
@@ -4597,10 +4597,79 @@ exec_stmt_execsql(PLtsql_execstate *estate,
...
 
+ 	top_es_entry = exec_state_call_stack->next;
+	if (stmt->is_dml)
+	{
+		while(top_es_entry != NULL)
+		{
+			/* traverse through the estate stack. If the occurrence of
+			 * exec in the call stack, update the search path.
+			 */
+			if(top_es_entry->estate && top_es_entry->estate->err_stmt &&
```
surendravishnoi: why are we doing it this way rather than setting right search path when a procedure starts execution?

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -648,15 +648,20 @@ exec_stmt_exec(PLtsql_execstate *estate, PLtsql_stmt_exec *stmt)
 	List *path_oids = fetch_search_path(false);
 	char *old_search_path = flatten_search_path(path_oids);
 	char *new_search_path;
+	estate->db_name = NULL;
```
surendravishnoi, line 651: does this code path also cover sp_execute?

contrib/babelfishpg_tsql/src/pl_exec.c
```diff
@@ -4597,10 +4597,79 @@ exec_stmt_execsql(PLtsql_execstate *estate,
...
+				top_es_entry->estate->err_stmt->cmd_type == PLTSQL_STMT_EXEC &&
+					top_es_entry->estate->db_name != NULL)
+				{
+					dbo_schema = get_dbo_schema_name(top_es_entry->estate->db_name);
+					new_search_path = psprintf("%s, %s", dbo_schema, old_search_path);
```
surendravishnoi: why DBO for DDLs, PR description seems to imply otherwise?

contrib/babelfishpg_tsql/src/pl_exec-2.c
shalinilohia50: We don't need to reset the db_name and schema_name for an estate. These will remain unique for an estate.

contrib/babelfishpg_tsql/src/pl_exec.c
shalinilohia50: If we set the search path at the start of the execution, it will look for the procedure name based on that search path which is not required. Also, we need to set the search path differently based on the DDL/DML statements being referred inside the procedure.

contrib/babelfishpg_tsql/src/pl_exec-2.c
shalinilohia50: No, it does not. This behaviour is not expected for T-SQL batches and for EXEC() and sp_executesql.
The schema name resolution is only driven by the current user's default schema(which is already implemented).

contrib/babelfishpg_tsql/src/pl_exec-2.c
surendravishnoi: so if I have proc p1 { exec p2; insert; exec p3; } shouldn't it change for each statement?

contrib/babelfishpg_tsql/src/pl_exec.c
surendravishnoi: I do not understand the diff req for DDL and DML, what if we add both physical and DBO schema to search paths, will it create an issue for DDL?

shalinilohia50: If we do that for DDL(CREATE) and we have that inside sch1.p1, we would end up creating the table in sch1 but not default [which is not the required behaviour]. In case of DML(SELECT), we need to select from sch1.t1 and if not found we need to search for dbo.t1 but not default.  This is why we need separate handling for DDL and DML.

contrib/babelfishpg_tsql/src/pl_exec-2.c
shalinilohia50: Yes, it's changing for each statement. `estate->db_name` and `estate->schema_name` are reset based on stmt, and inside proc p1 every exec is another stmt and based on which `db_name` and `schema_name` change. We don't need to explicitly change these values.

surendravishnoi: this function will only be called for exec, so exec p1 will set these values but same will carry to insert (insert will not come to this code path), is that expected? exec p2 will reset these to correct values again but there could be no exec statements in-between

contrib/babelfishpg_tsql/src/pl_exec.c
surendravishnoi: OK, I would still suggest to put this logic in a separate function with single loop and have conditions inside to set the correct search path, rather than repeating the same loop twice.

contrib/babelfishpg_tsql/src/pl_exec-2.c
shalinilohia50: Yes this is expected. In case of "insert", the parent proc would be p1 and we maintain a different estate for that call. The execution depends on that parent estate->db_name and schema_name which remains unchanged.

contrib/babelfishpg_tsql/src/pl_exec.c
shalinilohia50: For DDL (CREATE TABLE), an object created in a proc will be in the default schema for the caller of the proc. Modified it in the next commit.

contrib/babelfishpg_tsql/src/pl_exec.c
shalinilohia50: Moved this logic to a function and used a single loop.

contrib/babelfishpg_tsql/src/pl_exec.c
```diff
@@ -4973,14 +4983,23 @@ exec_stmt_execsql(PLtsql_execstate *estate,
...
 	}
 	PG_END_TRY();
 
+	if (need_path_reset)
+	 	(void) set_config_option("search_path", old_search_path,
+	 				PGC_USERSET, PGC_S_SESSION,
+	 				GUC_ACTION_SAVE, true, 0, false);
 	if (stmt->is_cross_db)
 		SetCurrentRoleId(current_user_id, false);
+	list_free(path_oids);
```
surendravishnoi, line 5003: should be done in CATH as well

test/JDBC/input/schema_resolution_proc-vu-prepare.sql
```diff
@@ -0,0 +1,50 @@
+create database schema_res_proc
+go
+	 
+use schema_res_proc
+go
+	 
+create schema sch1;
```
surendravishnoi: please use filename as prefix for all objects create in upgrade tests

contrib/babelfishpg_tsql/src/pl_exec.c
```diff
@@ -10014,3 +10033,62 @@ pltsql_clone_estate_err(PLtsql_estate_err *err)
...
+				/* Add default schema to the new search path */
+				(void) set_config_option("search_path", new_search_path,
+								PGC_USERSET, PGC_S_SESSION,
+								GUC_ACTION_SAVE, true, 0, false);
+				return true;
+			}
+		}
+		/* if the stmt is inside an exec_batch, return false */
+		else if(top_es_entry->estate && top_es_entry->estate->err_stmt &&
+				top_es_entry->estate->err_stmt->cmd_type == PLTSQL_STMT_EXEC_BATCH)
```
surendravishnoi, line 10090: why do we return from here?

shalinilohia50: Because in case the ddl/dml stmt is found inside the exec_batch, we don't need to iterate and check if it is inside an EXEC stmt and go with the default behaviour.

test/JDBC/input/schema_resolution_proc-vu-prepare.sql
shalinilohia50: Updated in the next commit

xhfanhe (approved):

macwoody (approved):

kuntalghosh (approved):

PR #434, opened 2022-07-25
Fix MVU failure when a view contains sys.suser_sid()

2jungkook:
[Engine]
If a view from v13.6 contains sys.suser_sid() and its
parameter is not explicitly casted to sys.sysname, then Major Version
Upgrade (MVU) to v14.latest fails. This is because we had assigned typmod for
all domains in T-SQL dialect in v13.x, which was fixed in v14.latest. This is only
problematic when a base type of a sys domain has a fixed typmod, e.g., sys.sysname.

To prevent the MVU failure, this commit adds a new hook removing typmods
for sys.sysname and sys._ci_sysname in gram.y.

[Extension]
Implement fix_tsql_domain_typmods hook that can detect wrong typmods and remove it.

Task: BABEL-3402
Signed-off-by: Jungkook Lee <jungkook@amazon.com>

### Check List


test/JDBC/input/BABEL-3402-vu-prepare.sql
```diff
@@ -0,0 +1,11 @@
+use master
+go
+
+DROP VIEW IF EXISTS babel_3402_v1
+go
+
+CREATE VIEW babel_3402_v1 AS SELECT suser_sid(-10)
```
surendravishnoi: pleas use filename as prefix for object names, babel_3402_vu_prepare_v1

test/JDBC/input/BABEL-3402-vu-prepare.sql
```diff
@@ -0,0 +1,11 @@
...
+go
+
+SELECT suser_sid(-10)
```
surendravishnoi: we also add tests where suser_sid is called from function and procedures

surendravishnoi (dismissed):

contrib/babelfishpg_tsql/src/backend_parser/gram_hook.c
```diff
@@ -14,12 +14,33 @@ static List *rewrite_typmod_expr(List *expr_list);
...
+static void
+fix_tsql_domain_typmods(TypeName *typname)
+{
+	const char *dump_restore = GetConfigOption("babelfishpg_tsql.dump_restore", true, false);
+	if (!dump_restore || strcmp(dump_restore, "on") != 0)
+		return;
+
+	/* sys.sysname and sys._ci_sysname should not have typmods */
+	if (strcmp(strVal(linitial(typname->names)), "sys") == 0 &&
+		(strcmp(strVal(lsecond(typname->names)), "sysname") == 0 ||
```
rishabhtanwar29, line 42: We should also check if length of names is two otherwise we might run into segfault if schema is not specified.

surendravishnoi (approved):

priyansx (approved):
LGTM

PR #442, opened 2022-07-26
Support SET PARSEONLY ON/OFF

timchang514:
Adding support for PARSEONLY. We replace the plan with an empty one if PARSEONLY
is enabled. To be able to turn PARSEONLY off, we actually alter the setting
at parse-time, which appears to be the behavior in T-SQL as well.

Task: BABEL-3317
Signed-off-by: Tim Chang <changti@amazon.com>

### Description

### Issues Resolved

### Check List


contrib/babelfishpg_tsql/src/guc.c
```diff
@@ -977,6 +978,15 @@ define_custom_variables(void)
 				 GUC_NOT_IN_SAMPLE | GUC_DISALLOW_IN_FILE | GUC_DISALLOW_IN_AUTO_FILE,
 				 NULL, NULL, NULL);
 
+    DefineCustomBoolVariable("babelfishpg_tsql.parseonly",
```
2jungkook: Why do we need a GUC? Isn't it enough to have a global variable?

contrib/babelfishpg_tsql/src/tsqlIface.cpp
```diff
@@ -2302,6 +2302,11 @@ antlr_parser_cpp(const char *sourceText)
 			 */
 			auto ssm = std::make_unique<tsqlSelectStatementMutator>();
 			handleBatchLevelStatement(tsql_file->batch_level_statement(), ssm.get());
+
+            /* If PARSEONLY is enabled, replace with empty statement */
+            if (pltsql_parseonly)
```
2jungkook: Where do you set `pltsql_parseonly` to true?

contrib/babelfishpg_tsql/src/tsqlIface.cpp
```diff
@@ -2302,6 +2302,11 @@ antlr_parser_cpp(const char *sourceText)
...
+                pltsql_parse_result = makeEmptyBlockStmt(0);
```
2jungkook: According to T-SQL doc, it should examine the syntax and return any error message without compilation. Can it be possible from this PR? If so, please add test cases.

contrib/babelfishpg_tsql/src/tsqlIface.cpp
```diff
@@ -2321,6 +2326,10 @@ antlr_parser_cpp(const char *sourceText)
 		if (pltsql_dump_antlr_query_graph)
 			toDotRecursive(tree, parser.getRuleNames(), sourceText);
 
+        /* If PARSEONLY is enabled, replace with empty statement */
```
2jungkook: Seems like it is a duplicate. Let's try to remove this.

contrib/babelfishpg_tsql/src/tsqlIface.cpp
```diff
@@ -4037,6 +4046,9 @@ makeSetStatement(TSqlParser::Set_statementContext *ctx, tsqlBuilder &builder)
 				query += " ";
 				query += getFullText(set_special_ctx->on_off());
 				query += "; ";
+
+				if (option->PARSEONLY() && pg_strcasecmp("off", getFullText(set_special_ctx->on_off()).c_str()) == 0)
```
2jungkook: The above logic adds `SET PARSEONLY OFF` to the query. Do we still need this?

contrib/babelfishpg_tsql/src/tsqlIface.cpp
```diff
@@ -4060,6 +4072,10 @@ makeSetStatement(TSqlParser::Set_statementContext *ctx, tsqlBuilder &builder)
 			auto option = set_special_ctx->set_on_off_option().front();
 			if (option->BABELFISH_SHOWPLAN_ALL())
 				return makeSetExplainModeStatement(ctx, true);
+            // PARSEONLY needs to be able to be turned off once it is on, so it is handled at parse time.
```
2jungkook: According to the comment, we should check first whether PARSEONLY is already on.

test/JDBC/expected/BABEL-3317.out
```diff
@@ -0,0 +1,44 @@
...
+GO
+SELECT * FROM t1
+GO
+~~START~~
+int
+1
+~~END~~
+
+
+SET PARSEONLY ON;
```
2jungkook, line 34: What are you trying to test here? It would be nice to have some explanations, especially for corner cases.

test/JDBC/expected/BABEL-3317.out
```diff
@@ -0,0 +1,44 @@
...
+go
+CREATE TABLE t1(a int);
```
2jungkook: Let's use `babel_3317_t1`. General table names such as `t1` can cause a problem when we are doing parallel tests.

contrib/babelfishpg_tsql/src/tsqlIface.cpp
```diff
@@ -2302,6 +2302,11 @@ antlr_parser_cpp(const char *sourceText)
 			 */
 			auto ssm = std::make_unique<tsqlSelectStatementMutator>();
 			handleBatchLevelStatement(tsql_file->batch_level_statement(), ssm.get());
+
```
2jungkook, line 2306: You may need to make a change in `tsqlUnsupportedFeatureHandler.cpp` file as well.

contrib/babelfishpg_tsql/antlr/TSqlLexer.g4
```diff
@@ -32,6 +32,7 @@ lexer grammar TSqlLexer;
 @header
 {
     extern bool pltsql_quoted_identifier;
+    extern bool pltsql_parseonly;
```
2jungkook: Why do we need this in TSqlLexer file?

timchang514: Needed to include this here so that the variable is visible from the parser code.

contrib/babelfishpg_tsql/src/guc.c
timchang514: I think it's best to have it as a GUC, so we can treat it like all of the other settings - we can turn it on/off using SET, check the value, etc. It was already added to the grammar previously - https://github.com/babelfish-for-postgresql/babelfish_extensions/blob/BABEL_2_X_DEV/contrib/babelfishpg_tsql/antlr/TSqlParser.g4#L3359

contrib/babelfishpg_tsql/src/tsqlIface.cpp
timchang514: It is set like the other GUCs, via SET statements.

contrib/babelfishpg_tsql/src/tsqlIface.cpp
timchang514: I think it makes no difference, right? If it is off, then flipping the bool shouldn't have any effect

contrib/babelfishpg_tsql/src/guc.c
2jungkook: Ah, it seems like there is a magic that converts T-SQL's `SET PARSEONLY` into PG's `SET babelfishpg_tsql.parseonly`. If so, `parseonly` can only be set to `TRUE` in execution time. Would you check the following case against both SQL Server and BBF?
```
CREATE TABLE t1(a int);
SET PARSEONLY ON;
GO

SET PARSEONLY OFF;
SELECT * FROM t1;
GO
```

2jungkook (dismissed):

2jungkook (approved):

kuntalghosh (approved):

kuntalghosh (commented):
Should we add some upgrade related test for this feature? Is there any impact on the upgrade?

timchang514:
There should be no upgrade impact. This is just introducing a new feature at the parser level - It needs to be specifically turned on to be used.

surendravishnoi (approved):

PR #458, opened 2022-07-28
Incorrect default value of (var)binary column after upgrade

rishabhtanwar29:
Default value of (var)binary columns get changed to some
incorrect value after major version upgrade. It happens only
when the original default value was specified as T-SQL
hex constant (not as string). The reason for this is that output
function for (var)binary datatype returns hex-encoded string
but input function directly stores the input string without
decoding it if TSQLHexConstTypmod is not passed. During
upgrade pg_dump dumps the default value as string and pg_restore
directly passes that string to varbinary input function without
specifying typmod = TSQLHexConstTypmod.

This commit fixes the issue by allowing hex decoding during
restore.

Additionally, removed upgrade test files for sys-suser-sname
from 13_6 directory and used common upgrade tests instead.

Task: BABEL-3031

### Check List


contrib/babelfishpg_common/src/varbinary.c
```diff
@@ -164,10 +164,12 @@ varbinaryin(PG_FUNCTION_ARGS)
 	int			len;
 	bytea	   *result;
 	int32 typmod = PG_GETARG_INT32(2);
+	const char *dump_restore = GetConfigOption("babelfishpg_tsql.dump_restore", true, false);
 
 	len = strlen(inputText);
 
-	if (typmod == TSQLHexConstTypmod)
+	if (typmod == TSQLHexConstTypmod ||
+		(dump_restore && strcmp(dump_restore, "on") == 0)) /* Treat input string as T-SQL hex constant during restore */
```
surendravishnoi, line 172: is it safe to do this decoding unconditionally during restore? what if default is not hex encoded string?

2jungkook (dismissed):

contrib/babelfishpg_common/src/varbinary.c
rishabhtanwar29: Yeah, pg_dump always dumps binary/varbinary values as hex encoded string, so unconditional decoding is safe here. Secondly, no matter what the default value was, dump will always contain hex encoded string since varbinary output function is written like that.

surendravishnoi (dismissed):

surendravishnoi (approved):

PR #459, opened 2022-07-28
Incorrect schema resolution for the SQL objects inside a function

shalinilohia50:
Incorrect schema resolution for the SQL objects inside a function.

### Change
When the function name is schema qualified and the SQL objects being referenced inside that function is not schema qualified, currently those objects were resolved to the default schema of the user. But that's incorrect. 
With this change, the SQL objects being referenced inside the function should resolve to the function's schema.

Steps we follow:
1. Search for the object in the same schema as the SQL object in which the reference occurs.
2. If not found, search in the `dbo` schema.

### Note
This issue is specific for name resolution inside a SQL object. Babelfish resolves the schema incorrectly. The consequence of this bug is that we may return wrong results or update the wrong table.

Task: BABEL-2992


contrib/babelfishpg_tsql/src/pl_exec.c
```diff
@@ -10045,6 +10045,18 @@ bool reset_search_path(PLtsql_stmt_execsql *stmt, char *old_search_path)
 	const char	*user = NULL;
 	const char	*schema;
 	top_es_entry = exec_state_call_stack->next;
+
+	if (stmt->func_call && stmt->schema_name != NULL && strncmp(stmt->schema_name, "sys", 3) != 0)
```
surendravishnoi: this will fail if schema name is sysxyz

contrib/babelfishpg_tsql/src/pl_exec.c
```diff
@@ -10045,6 +10045,18 @@ bool reset_search_path(PLtsql_stmt_execsql *stmt, char *old_search_path)
...
+	{
+		physical_schema = get_physical_schema_name(cur_dbname, stmt->schema_name);
+		dbo_schema = get_dbo_schema_name(cur_dbname);
+		new_search_path = psprintf("%s, %s, %s", physical_schema, dbo_schema, old_search_path);
+		/* Add the schema where the object is referenced and dbo schema to the new search path */
+		(void) set_config_option("search_path", new_search_path,
```
surendravishnoi, line 10060: do we need to reset it later?

contrib/babelfishpg_tsql/src/pl_exec.c
```diff
@@ -10045,6 +10045,18 @@ bool reset_search_path(PLtsql_stmt_execsql *stmt, char *old_search_path)
 	const char	*user = NULL;
 	const char	*schema;
 	top_es_entry = exec_state_call_stack->next;
+
+	if (stmt->func_call && stmt->schema_name != NULL && strncmp(stmt->schema_name, "sys", 3) != 0)
+	{
+		physical_schema = get_physical_schema_name(cur_dbname, stmt->schema_name);
```
surendravishnoi, line 10056: who manages memory for these variables?

test/JDBC/input/schema_resolution_func-vu-prepare.sql
```diff
@@ -0,0 +1,41 @@
+create schema schema_resolution_func_s1;
```
surendravishnoi: Please use filename_xxx for object names.

test/JDBC/input/schema_resolution_func-vu-prepare.sql
```diff
@@ -0,0 +1,41 @@
...
+create table schema_resolution_func_t1(a int);
+go
+
+create table schema_resolution_func_s1.schema_resolution_func_t1(a int);
+go
+
+insert into schema_resolution_func_s1.schema_resolution_func_t1 values (1);
+go
+
+create function dbo.schema_resolution_func_f1()
```
surendravishnoi: we should have tests which have procedures/functions/views and triggers nested at 3-4 levels to validate correct search path is used. Also, we should test TVF, MSTVF etc. as well

xhfanhe:
Please fix the merge conflicts

contrib/babelfishpg_tsql/src/pl_exec.c
shalinilohia50: Yes, we should reset the search path if we make any changes to it otherwise it may affect the next statements.

test/JDBC/input/schema_resolution_func-vu-prepare.sql
shalinilohia50: Since the file name is `schema_resolution_func-vu-prepare` which is too long so I have prefixed `schema_resolution_func` for the object names.

surendravishnoi: you mean it is so long that it does not work or just looks very long? I would still prefer to use long name if the work

contrib/babelfishpg_tsql/src/pl_exec.c
shalinilohia50: Updated this check.

test/JDBC/input/schema_resolution_func-vu-prepare.sql
shalinilohia50: It works but the names look very long. I will go with the long names and update it.

test/JDBC/input/schema_resolution_func-vu-prepare.sql
shalinilohia50: Added more tests in the next commit.

contrib/babelfishpg_tsql/src/pl_exec.c
shalinilohia50: These are local variables and we are not dynamically allocating any memory, it gets stored in the stack while function execution and gets cleaned up as it goes out of scope. We don't need to explicitly free the memory.

xhfanhe (dismissed):

contrib/babelfishpg_tsql/src/pl_exec.c
shalinilohia50: These are local variables stored in the stack and gets cleaned up at the end of the function execution. Since we are not allocated the memory dynamically, we don't need to free it explicitly.

macwoody (dismissed):

PR #541, opened 2022-08-16
Fixed identifying IF and TF in sys.all_objects and OBJECTPROPERTY

faviansamatha:
### Description

Previously, sys.all_objects mistakenly identified inline table-valued functions (IF), and table-valued functions (TF) as scalar functions (FN). This commit fixes that issue. In addition, this commit fixes the issue for the following properties in OBJECTPROPERTY: IsScalarFunction, IsTableFunction, IsInlineFunction. This commit also fixes misidentifying table functions as scalar functions in SSMS.

Tasks: BABEL-3369, BABEL-3370, BABELFISH-511

### Check List


xhfanhe (dismissed):

xhfanhe:
Code lgtm, but since this is a quite large change, I would suggest getting another approval just to double check. Thank you!

contrib/babelfishpg_tsql/sql/sys_views.sql
```diff
@@ -1292,7 +1306,9 @@ select
   , 0 as is_schema_published
 from pg_proc p
 inner join pg_namespace s on s.oid = p.pronamespace
-left join pg_trigger tr on tr.tgfoid = p.oid
+left join pg_trigger tr on tr.tgfoid = p.oid,
+format_type(p.prorettype, null) format_rettype,
```
macwoody: I feel that we don't need to join these two here since they will be needed only in two cases.

contrib/babelfishpg_tsql/sql/sys_views.sql
```diff
@@ -1271,17 +1271,31 @@ select
       when 'p' then 'P'::varchar(2)
       when 'a' then 'AF'::varchar(2)
       else
-        case format_type(p.prorettype, null) when 'trigger'
-          then 'TR'::varchar(2)
+        case 
+          when format_rettype = 'trigger'
+            then 'TR'::varchar(2)
+          when func_results LIKE 'TABLE(%' then
```
macwoody: proretset could replace this string comparison, since that's how "TABLE(" is printed

contrib/babelfishpg_tsql/sql/sys_views.sql
```diff
@@ -1271,17 +1271,31 @@ select
...
+            case 
+              when format_rettype = 'record'
```
macwoody: I think we could call function here to check if it is 'record' or use 'record''s OID

macwoody:
I feel that the implemented approach seems could not distinguish the two cases while. 
See this example. 
```
1> CREATE FUNCTION f2 (@a INT)
2> RETURNS TABLE AS
3> RETURN (SELECT @a + 1 AS col);
4> GO
1> CREATE FUNCTION f3 (@a INT)
2> RETURNS TABLE AS
3> RETURN (SELECT 1 AS COL, @a + 1 as COL2);
4> GO
```
Both of them I thin are SQL inlined table function if I am not wrong, 

```
babelfish_db=# select format_type(prorettype, null), proretset, pg_get_function_result(oid) from pg_proc where proname in ( 'f3' , 'f2');
 format_type | proretset |        pg_get_function_result
-------------+-----------+--------------------------------------
 integer     | t         | TABLE(col integer)
 record      | t         | TABLE("COL" integer, "COL2" integer)

```

contrib/babelfishpg_tsql/sql/sys_views.sql
faviansamatha: Done!

contrib/babelfishpg_tsql/sql/sys_views.sql
faviansamatha: Done, removed the format_rettype and the func_result

contrib/babelfishpg_tsql/sql/sys_views.sql
faviansamatha: Decided to not use the 'record' field due to the indistinguishable case you mentioned.

contrib/babelfishpg_tsql/sql/sys_views.sql
```diff
@@ -1283,17 +1283,31 @@ select
...
       when 'a' then 'AF'::varchar(2)
       else
-        case format_type(p.prorettype, null) when 'trigger'
-          then 'TR'::varchar(2)
+        case 
+          when pg_catalog.format_type(p.prorettype, null) = 'trigger'
+            then 'TR'::varchar(2)
+          when p.proretset then
+            case 
+              when t.typtype = 'c'
```
faviansamatha, line 1307: From what I've observed, it seems like Table type function will always return a composite type in pg_types, but inline-table-valued functions will return record or another datatype (like int in the case that it returns a single column)

faviansamatha:
> I feel that the implemented approach seems could not distinguish the two cases while. See this example.
> 
> ```
> 1> CREATE FUNCTION f2 (@a INT)
> 2> RETURNS TABLE AS
> 3> RETURN (SELECT @a + 1 AS col);
> 4> GO
> 1> CREATE FUNCTION f3 (@a INT)
> 2> RETURNS TABLE AS
> 3> RETURN (SELECT 1 AS COL, @a + 1 as COL2);
> 4> GO
> ```
> 
> Both of them I thin are SQL inlined table function if I am not wrong,
> 
> ```
> babelfish_db=# select format_type(prorettype, null), proretset, pg_get_function_result(oid) from pg_proc where proname in ( 'f3' , 'f2');
>  format_type | proretset |        pg_get_function_result
> -------------+-----------+--------------------------------------
>  integer     | t         | TABLE(col integer)
>  record      | t         | TABLE("COL" integer, "COL2" integer)
> ```

@macwoody I modified the solution and added these set of tests in sys.all_objects-vu-verify. It should be able to identify these two cases now.

macwoody (dismissed):
Looks good to me from the view implementation POV

leszekamz (approved):

PR #551, opened 2022-08-19
Remove restriction, input parameters after one with a default value must also have defaults, for all PL/tsql procedures/functions

rishabhtanwar29:
#### Engine
* PG has a restriction that input parameters after one with a default
value must also have defaults. Previously we provided a workaround
for this by letting user provide default parameters anywhere but
filling NULL as default value for all parameters after first default
parameter.  
* This commit removes this restriction for all PL/tsql functions. We will
be storing the positions of default parameters in a separate babelfish
catalog table (babelfish_function_ext) so that we can use these
positions while looking for the parameter default expression. Introduced
following hooks to achieve the functionality:
  1. `match_pltsql_func_call_hook`: Will help in parsing a PL/tsql function call.
  2. `insert_pltsql_function_defaults_hook`: Will be used to insert
      actual default values in a PL/tsql function call during execution phase.
  3. `print_pltsql_function_arguments_hook`: Will be used to print
      arguments for PL/tsql function depending upon default positions in
      catalog table.

  Additionally, introduced a syscache id `PROCNSPSIGNATURE`
  for new babelfish_function_ext catalog which takes sysdatabase oid,
  logical schema name and function signature as input keys.

#### Extensions
* Implemented all the hooks introduced on engine side.
* Added a sql definition of babelfish_function_ext catalog table in which
all three columns dbid, schema_name and function_signature combined
will be used as a primary key and default_positions column will hold
the list of argument's default positions.

Task: BABEL-2877
signed-off-by: Rishabh Tanwar <ritanwar@amazon.com>

### Check List


contrib/babelfishpg_tsql/sql/ownership.sql
```diff
@@ -14,6 +14,18 @@ CREATE TABLE sys.babelfish_sysdatabases (
 
 GRANT SELECT on sys.babelfish_sysdatabases TO PUBLIC;
 
+-- BABELFISH_FUNC_DEFAULT
+CREATE TABLE sys.babelfish_function_ext (
+	nspname NAME NOT NULL,
+	funcname NAME NOT NULL,
+	funcsignature TEXT NOT NULL COLLATE "C",
+	default_positions TEXT COLLATE "C",
+	PRIMARY KEY(nspname, funcsignature)
```
macwoody, line 24: why funcname is not part of the primary key ?  
My understanding is that you could implement drop / alter procedure and functions only when PK is identical to base catalog table? How do you handle drop/ / alter case. 
And is there a reason to exclude stored procedure from the beginning ?

macwoody (commented):
I didn't check details on your implementation, instead, I focused more on the backend change and catalog definition. Since this is introducing catalog table, test on DDL behaviors are also needed. And please reachout to Xiaohui to incorporate integrity checker on this catalog too.

contrib/babelfishpg_tsql/sql/ownership.sql
macwoody: Since we are introducing catalog for pg_proc, there was a need to store users' original input name too. Please check with Rob and leave a column for it even if not filling it at the moment.

rishabhtanwar29: > why funcname is not part of the primary key ? My understanding is that you could implement drop / alter procedure and functions only when PK is identical to base catalog table? How do you handle drop/ / alter case. And is there a reason to exclude stored procedure from the beginning ?

`funcsignature` actually consists of function name + arguments (e.g.: `f1(int, varchar)`) so `funcname` is not needed in PK. `schema name + function signature` can actually uniquely identify a function. I added a `funcname` column just to for the future reference, in case we need it. I am using `schema name + function signature` only for `DROP` also but currently I am not handling `ALTER` currently since alter function is not supported for functions in babel.
Also, can you clarify what are `DDL behavior` tests or point me to some existing DDL behavior tests present in our framework?

xhfanhe: Yes you would need to add some rules in integrity checker (in catalog.c)
I can think of several rules like
For each entry in babelfish_function_ext, 
1. nspname must exist in babelfish_namespace_ext
2. funcname must exist in pg_proc

It's developer's responsibility to list out all possible rules, i.e. the connection between babelfish_function_ext and other catalogs, especially PG catalogs, and add them in the checker.
Feel free to ask me if you have any other questions.

rishabhtanwar29: Added above integrity checks

PR #554, opened 2022-08-22
Incorrect schema name resolution in SQL objects for triggers

shalinilohia50:
### Description
Incorrect schema name resolution in SQL objects for triggers

Cases:

- If the stmt is DDL/DML and it's schema specified, then there is no need to update the search path. It already works as expected.
- If the stmt is DDL and it's not schema specified, then there is no need to update the search path. 
-  If the stmt is DML and it's not schema specified,  then update the search path. The table should be searched first in the same schema where the object is being referenced. If not found, search the dbo schema.

NOTE: Added pg_sleep() after the change in migration mode to fix the inconsistent MVU test failures in empty databases, the tests were failing because the migration mode was not getting updated instantly leading to errors like below:
```
 CREATE DATABASE db1;
 GO
+~~ERROR (Code: 33557097)~~
+
+~~ERROR (Message: Database 'db1' already exists. Choose a different database name.)~~
+
```

Task: BABEL-3464


xhfanhe (dismissed):

macwoody (approved):

contrib/babelfishpg_tsql/src/pl_exec.c
```diff
@@ -10111,6 +10120,30 @@ bool reset_search_path(PLtsql_stmt_execsql *stmt, char *old_search_path, bool* r
 		else if(top_es_entry->estate && top_es_entry->estate->err_stmt &&
 				top_es_entry->estate->err_stmt->cmd_type == PLTSQL_STMT_EXEC_BATCH)
 			return false;
+
+		/* Traverse through the estate stack, if the stmt is inside trigger
+		 * we set the search path
+		 */
+		if(top_es_entry->estate && top_es_entry->estate->err_stmt &&
```
macwoody: I feel like it would be more straight forward if we check 
top_es_entry->estate && top_es_entry->estate->err_stmt  in the beginning of the loop and use a switch to redirect different branches. 
And if you see it is better to keep current form, would it be better if it is another "else if " instead of "if"?

macwoody (dismissed):

xhfanhe (dismissed):

contrib/babelfishpg_tsql/src/pl_exec.c
shalinilohia50: Moving ahead with the current form and replaced 'if' with `else if` to maintain the consistency. There were some cases which failed in switch so finding it better to have this approach.

xhfanhe (approved):

macwoody (approved):

shardgupta (approved):

PR #558, opened 2022-08-24
Handle Bulk Load Request for PLP data when the PLP chunk lies in different/across packets

KushaalShroff:
Incase of PLP data we should not discard the previous packet since we first store the offset of the PLP Chunks first and then read the data later.
Task: BABEL-3500

### Description

### Issues Resolved

### Check List


contrib/babelfishpg_tds/src/backend/tds/tdsbulkload.c
```diff
@@ -915,7 +974,7 @@ ReadBcpPlp(ParameterToken temp, StringInfo *message, TDSRequestBulkLoad request)
 	Plp plpTemp, plpPrev = NULL;
 	unsigned long lenCheck = 0;
 
-	CheckMessageHasEnoughBytesToRead(message, sizeof(plpTok));
+	CheckPlpMessageHasEnoughBytesToRead(message, sizeof(plpTok));
```
surendravishnoi, line 977: Looking at the code it seems that previous logic should work, can you please describe error case in more details to understand why this change is needed

KushaalShroff: In ReadBcpPlp function we create a linked list of type Plp which stores the offset of the actual data-value for the field in that structure:
```
		plpTemp = palloc0(sizeof(PlpData));
		plpTemp->next = NULL;
		plpTemp->offset = offset;
		plpTemp->len = tempLen;
		if (plpPrev == NULL)
		{
			plpPrev = plpTemp;
			temp->plp = plpTemp;
		}
		else
		{
			plpPrev->next = plpTemp;
			plpPrev = plpPrev->next;
		}
```
Now inside ReadBcpPlp if while reading the next “tempLen” we do not have sufficient data then we used to call FetchMoreBcpData which to discards the previous bytes in the message-data.
In one test case the PLP_TERMINATOR (4 bytes) was partially in the next packet, so we discarded the the current packet and reset the offset (The Plp LinkedList which could have n previous nodes still has stale offsets).
So later when we actually fetch the data by calling TdsGetPlpStringInfoBufferFromToken the data it reads is corresponding to the stale offset

We might have to inefficiently manipulate the offsets in a loop for all n previous nodes or we dont discard the previous packet which will keep the offset genuine

KushaalShroff: May be we can optimise it in a better way in the next release

surendravishnoi (approved):

priyansx (approved):
LGTM

PR #570, opened 2022-08-30
Check and initialise ht_oid2collid hash table before we access it

Deepesh125:
### Description

Check and initialise ht_oid2collid hash table before we access it

Previously, ht_oid2collid hash was made to initialise on demand. We missed to handle one case where we are trying to access ht_oid2collid without checking if it is initialised which is causing backend to crash. 

This commit fixes it by appropriately checking and initialising ht_oid2collid hash table before attempt to hash_search is made.

Task: BABEL-3514

### Check List

### DIFF
--- a/contrib/babelfishpg_common/src/collation.c
+++ b/contrib/babelfishpg_common/src/collation.c
@@ -1180,6 +1180,9 @@ int get_persist_collation_id(Oid coll_oid)
 	bool found_coll;
 	int collidx;
 
+	if (ht_oid2collid == NULL)
+		init_collid_trans_tab_internal();
+
 	entry = hash_search(ht_oid2collid, &coll_oid, HASH_FIND, &found_coll);
 
 	if (found_coll)
--- a/test/JDBC/jdbc_schedule
+++ b/test/JDBC/jdbc_schedule
@@ -8,6 +8,8 @@
 #    new line
 # 6. If you want the framework to not run certain files, use: ignore#!#<test name>
 
+# check babelfish version first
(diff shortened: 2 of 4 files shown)


test/JDBC/jdbc_schedule
```diff
@@ -8,6 +8,8 @@
 #    new line
 # 6. If you want the framework to not run certain files, use: ignore#!#<test name>
 
+# check babelfish version first
+cmd#!#postgresql#!#select cast(case when cast(sys.SERVERPROPERTY('BabelfishVersion') as varchar(20)) LIKE '_._._' THEN 'valid' else 'invalid' end as sys.varchar(20));
```
surendravishnoi, line 12: why would we do this here in schedule file?

contrib/babelfishpg_common/src/collation.c
```diff
@@ -1180,6 +1180,9 @@ int get_persist_collation_id(Oid coll_oid)
 	bool found_coll;
 	int collidx;
 
+	if (ht_oid2collid == NULL)
```
surendravishnoi, line 1183: did we make sure that there are no other places which require similar check?

Deepesh125: Yes, I've verified the same.

test/JDBC/jdbc_schedule
Deepesh125: To repro this issue, We need to invoke given function right after installing babelfish. So, I could not find any other better place to verify that SERVERPROPERTY('BabelfishVersion') works as intended (at least does not crash or raise an error)

PR #608, opened 2022-09-16
Fix precision error when sending numeric zero

rishabhtanwar29:
### Description
When TDS gets a numeric zero with typmod = -1 to sent to client then it calls **numeric_get_typmod** function to calculate typmod, precision and scale using actual expression but numeric zero has both precision and scale equal to zero so we send 0 to client directly which causes error on some clients as precision 0 is invalid.

Fix this by sending (precision,scale) equal to T-SQL default **(18,0)** instead for numeric zero.

### Test description
1. Added exact repro test scenario in dotnet Framework.
2. Manually tested with SSMS client where the error was reported.
3. Edge case is numeric zero, so added test cases for both numeric zero and non-zero values.

Task: BABEL-3344

### Check List


bsatarupa (dismissed):

test/dotnet/input/Datatypes/TestNumeric.txt
```diff
@@ -176,3 +176,6 @@ prepst#!#exec#!#numeric|-|a7|-|-123|-|9|-|2
 #SELECT * FROM numeric_table;
 
 DROP TABLE numeric_table;
+
+# Test BABEL-3344
+SELECT CASE WHEN 1=1 THEN 0 ELSE cast(98.7627 AS numeric(24,6)) END
```
surendravishnoi: why we have written test this way where else will never execute

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
```diff
@@ -890,7 +890,7 @@ MakeEmptyParameterToken(char *name, int atttypid, int32 atttypmod, int attcollat
 				 * Get the precision and scale out of the typmod value if typmod is valid
 				 * Otherwise tds_default_numeric_precision/scale will be used.
 				 */
-				if (atttypmod >= VARHDRSZ)
```
surendravishnoi, line 893: do we have test cases to cover all three case <, = and >?

test/dotnet/input/Datatypes/TestNumeric.txt
rishabhtanwar29: This is the original test scenario given in the jira but although the `else` never gets returned but it's value decides the return type of result. So, `0` will be returned as numeric (not int) since else condition evaluates to numeric.

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
rishabhtanwar29: Yeah there are already tests which are testing this but I have confirmed with Satarupa that we should only check for `>` in above condition otherwise if atttypmod = VARHDRSZ, we will incorrectly return both precision and scale as zeros.

surendravishnoi: is this only applicable to numeric zero? otherwise, we do not have test coverage for = condition which was broken so far

rishabhtanwar29: Yeah atttypmod = VARHDRSZ only happens to be in case of numeric zero. I have added few more test cases for non zero numeric values.

bsatarupa (approved):

surendravishnoi (approved):

PR #635, opened 2022-09-27
Add JSON_MODIFY()

anujsarda3:
This function is used to update the value of the property in JSON string. Test cases are also added for the implementation. 

Signed-off-by: Anuj Sarda <sardanuj@amazon.com>

### Issues Resolved

BABEL-937


contrib/babelfishpg_tsql/sql/sys_functions.sql
```diff
@@ -2540,6 +2540,143 @@ CREATE OR REPLACE FUNCTION sys.json_query(json_string text, path text default '$
...
+    SELECT regexp_replace(json_path, '\$\.|]|\$\[' , '' , 'ig') INTO json_path; -- To remove the "$." and "]" sign from the string 
+    SELECT regexp_replace(json_path, '\.|\[' , ',' , 'ig') INTO json_path; -- To replace the "." and "[" with the "," to change into required format
+    SELECT CONCAT('{',json_path,'}') INTO json_path; -- Final required format of path by jsonb_set
+    return json_path;
+END
+$BODY$
+LANGUAGE plpgsql;
+
+--Function to set the append_modifier and create_if_missing flags
+CREATE TYPE falgs_set AS (create_if_missing BOOL, append_modifier BOOL);
```
Sairakan: typo here; should be `flags_set`, not `falgs_set`

contrib/babelfishpg_tsql/sql/sys_functions.sql
```diff
@@ -2540,6 +2540,143 @@ CREATE OR REPLACE FUNCTION sys.json_query(json_string text, path text default '$
 RETURNS sys.NVARCHAR
 AS 'babelfishpg_tsql', 'tsql_json_query' LANGUAGE C IMMUTABLE PARALLEL SAFE;
 
+-- Function to convert the path format into expected jsonb_set path format using Regular Expression
+CREATE OR REPLACE FUNCTION sys.regex_replace(json_path TEXT)
```
Sairakan: maybe a more descriptive name than just `regex_replace()`? Something that describes what the function does instead of how it works, so something like `convert_jsonpath()` or `convert_jsonpath_to_jsonb_path()`

contrib/babelfishpg_tsql/sql/sys_functions.sql
Sairakan: Also, the name of the datatype is not very descriptive - what if someone else wants to create a flag type? It should be more specific to its use, like `sys.json_modify_flags`. Also, make sure to declare this in the sys schema.

contrib/babelfishpg_tsql/sql/sys_functions.sql
```diff
@@ -2540,6 +2540,143 @@ CREATE OR REPLACE FUNCTION sys.json_query(json_string text, path text default '$
...
+    SELECT CONCAT('{',json_path,'}') INTO json_path; -- Final required format of path by jsonb_set
+    return json_path;
+END
+$BODY$
+LANGUAGE plpgsql;
+
+--Function to set the append_modifier and create_if_missing flags
+CREATE TYPE falgs_set AS (create_if_missing BOOL, append_modifier BOOL);
+
+CREATE OR REPLACE FUNCTION sys.set_flags(path_json TEXT)
```
Sairakan: Similar to above, change to `sys.set_json_modify_flags` to be more descriptive and specific

contrib/babelfishpg_tsql/sql/sys_functions.sql
```diff
@@ -2540,6 +2540,143 @@ CREATE OR REPLACE FUNCTION sys.json_query(json_string text, path text default '$
...
+
+
+--JSON_MODIFY
+CREATE OR REPLACE FUNCTION sys.json_modify(in json_string JSONB,in path_json TEXT, in new_value TEXT)
+RETURNS sys.NVARCHAR
+AS
+$BODY$
+DECLARE
+    json_path TEXT;
+    result falgs_set;
```
Sairakan: result is not a good variable name for this - makes it seem like we're saving the result in this variable.

contrib/babelfishpg_tsql/sql/sys_functions.sql
anujsarda3: Have made the changes and named the datatype sys.json_modify_flags

contrib/babelfishpg_tsql/sql/sys_functions.sql
anujsarda3: Have changed the function name to make it more specific sys.convert_jsonpath_to_jsonb_path

contrib/babelfishpg_tsql/sql/sys_functions.sql
anujsarda3: Have changed the variable name to mode_flags

contrib/babelfishpg_tsql/sql/sys_functions.sql
```diff
@@ -2540,6 +2540,143 @@ CREATE OR REPLACE FUNCTION sys.json_query(json_string text, path text default '$
...
+    
+    
+    --This if else block is to call the jsonb_set function based on the create_if_missing and append_modifier flags
+    
+    IF mode_flags.append_modifier THEN --When append modifier is present
+        IF key_exist THEN
+            key_value = jsonb_path_query_first(json_string::JSONB,json_path::jsonpath); -- To get the value of the key
+            type_key_value = jsonb_typeof(key_value::JSONB);
+            IF type_key_value='array' THEN
+                len_array = jsonb_array_length(key_value::JSONB);
```
Sairakan: small nitpick: casting is expensive, so try to cast once and save the value in a variable for reuse instead of casting types multiple times

contrib/babelfishpg_tsql/sql/sys_functions.sql
```diff
@@ -2540,6 +2540,143 @@ CREATE OR REPLACE FUNCTION sys.json_query(json_string text, path text default '$
...
+                new_jsonb_path = CONCAT(TRIM('}' FROM new_jsonb_path),',',CAST(len_array AS TEXT),'}'); -- To change the path format into the required jsonb_insert path format
+                IF new_value IS NULL THEN
+                    RETURN jsonb_insert(json_string::JSONB,new_jsonb_path::TEXT[],'null');
```
Sairakan: Can you combine all of these return calls into one return call at the bottom of the main IF block? And instead set some variable to pass into the return statement

contrib/babelfishpg_tsql/sql/sys_functions.sql
```diff
@@ -2540,6 +2540,143 @@ CREATE OR REPLACE FUNCTION sys.json_query(json_string text, path text default '$
...
+            IF mode_flags.create_if_missing='FALSE' THEN
+                RAISE EXCEPTION 'Property cannot be found on the specified JSON path.';
+            ELSE
+                RETURN jsonb_insert(json_string::JSONB,new_jsonb_path::TEXT[],to_jsonb(array_agg(new_value))); -- array_agg is used to convert the new_value text into array format as we append functionality is being used
+            END IF;
+        END IF;
+    ELSE --When no append modifier is present
+        IF new_value IS NOT NULL THEN
+            IF key_exist OR mode_flags.create_if_missing='TRUE' THEN
+                RETURN jsonb_set_lax(json_string::JSONB,new_jsonb_path::TEXT[],to_jsonb(new_value),mode_flags.create_if_missing);
```
Sairakan: similar comment to above re: return statement

contrib/babelfishpg_tsql/sql/sys_functions.sql
anujsarda3: Have made changes in the code to reduce the number of cast done.

contrib/babelfishpg_tsql/sql/sys_functions.sql
```diff
@@ -2540,6 +2540,146 @@ CREATE OR REPLACE FUNCTION sys.json_query(json_string text, path text default '$
...
+DECLARE
+    result_mode_flags json_modify_flags;
+    word_count INTEGER;
+    create_if_missing BOOL:='TRUE';
+    mode TEXT;
+    modifier TEXT;
+    append_modifier BOOL:='FALSE';
+BEGIN
+    
+    word_count = LENGTH(path_json) - LENGTH(REPLACE(path_json, ' ', '')) + 1; --To calculate the number of words in the path_json
```
nasbyj: Why not just count the number of spaces? Also, what happens with JSON that has multiple spaces between words? What about CR/LF?

nasbyj: Also, since there's several code paths that need to split parts of this out, it might be better to just do `string_to_array()` or `regexp_split_to_array()` once and be done with it.

contrib/babelfishpg_tsql/sql/sys_functions.sql
```diff
@@ -2540,6 +2540,146 @@ CREATE OR REPLACE FUNCTION sys.json_query(json_string text, path text default '$
...
+        modifier = SPLIT_PART(path_json COLLATE sql_latin1_general_cp1_cs_as, ' ',1);
+        mode = SPLIT_PART(path_json COLLATE sql_latin1_general_cp1_cs_as, ' ',2);
+        IF modifier='append' THEN
+            append_modifier:='TRUE';
+        END IF;
+        IF mode='strict' THEN
+            create_if_missing:='FALSE';
+        END IF;
+    END IF;
+    SELECT create_if_missing,append_modifier INTO result_mode_flags.create_if_missing, result_mode_flags.append_modifier;
```
nasbyj: FWIW I think it's clearer to just use `:=` instead of `SELECT INTO`, especially since `SELECT INTO` can be confused with the engine syntax.

contrib/babelfishpg_tsql/sql/sys_functions.sql
```diff
@@ -2540,6 +2540,146 @@ CREATE OR REPLACE FUNCTION sys.json_query(json_string text, path text default '$
...
+$BODY$
+DECLARE
+    json_path TEXT;
+    mode_flags json_modify_flags;
+    new_jsonb_path TEXT[];
+    len_array INTEGER;
+    type_key_value TEXT;
+    key_exist BOOL;
+    key_value JSONB;
+    json_expression JSONB;
```
nasbyj: to_jsonb() is pointless after doing `expression::jsonb`. Just do 

`json_expression jsonb := expression::jsonb;`

contrib/babelfishpg_tsql/sql/sys_functions.sql
```diff
@@ -2540,6 +2540,146 @@ CREATE OR REPLACE FUNCTION sys.json_query(json_string text, path text default '$
...
+    create_if_missing BOOL:='TRUE';
+    mode TEXT;
+    modifier TEXT;
+    append_modifier BOOL:='FALSE';
+BEGIN
+    
+    word_count = LENGTH(path_json) - LENGTH(REPLACE(path_json, ' ', '')) + 1; --To calculate the number of words in the path_json
+    
+    -- This if else block is added to set the create_if_missing and append_modifier flags
+    IF word_count=1 THEN --The path_json has only 1 word
```
nasbyj: This comment is really just noise... it's clear from `IF word_count=1` that this is the one-word case.

contrib/babelfishpg_tsql/sql/sys_functions.sql
```diff
@@ -2540,6 +2540,119 @@ CREATE OR REPLACE FUNCTION sys.json_query(json_string text, path text default '$
...
+    type_key_value TEXT;
+    mode TEXT;
+    modifier TEXT;
+    len_array INTEGER;
+    word_count INTEGER;
+    create_if_missing BOOL:='TRUE';
+    append_modifier BOOL:='FALSE';
+    key_exist BOOL;
+    key_value JSONB;
+    json_expression JSONB;
```
nasbyj: The `to_jsonb()` below is pointless after you've casted to jsonb. That whole line should just be removed by doing the assignment in the DECLARE section. While you're at it, you could also declare the variable as a constant:

```
json_expression CONSTANT jsonb := expression::jsonb;
```

contrib/babelfishpg_tsql/sql/sys_functions.sql
```diff
@@ -2540,6 +2540,119 @@ CREATE OR REPLACE FUNCTION sys.json_query(json_string text, path text default '$
 RETURNS sys.NVARCHAR
 AS 'babelfishpg_tsql', 'tsql_json_query' LANGUAGE C IMMUTABLE PARALLEL SAFE;
 
+--JSON_MODIFY
+CREATE OR REPLACE FUNCTION sys.json_modify(in expression NVARCHAR,in path_json TEXT, in new_value TEXT)
```
nasbyj: This function needs a comment block explaining what the overall big picture here is. It's extremely difficult to follow any of this code without that.

contrib/babelfishpg_tsql/sql/sys_functions.sql
```diff
@@ -2540,6 +2540,119 @@ CREATE OR REPLACE FUNCTION sys.json_query(json_string text, path text default '$
...
+BEGIN
+    json_expression = to_jsonb(expression::JSONB);
+
+    json_path = regexp_replace(path_json, '^.* ', '','ig'); -- To select the last word of the json_path string. \\\\\\Instead of this one can use split_part if regexp is slow
+    json_path_convert = regexp_replace(json_path, '\$\.|]|\$\[' , '' , 'ig'); -- To remove the "$." and "]" sign from the string 
+    json_path_convert = regexp_replace(json_path_convert, '\.|\[' , ',' , 'ig'); -- To replace the "." and "[" with the "," to change into required format
+    new_jsonb_path = CONCAT('{',json_path_convert,'}'); -- Final required format of path by jsonb_set
+    
+    
+    key_exist = jsonb_path_exists(json_expression,json_path::jsonpath); -- TO check if key exist in the given path
```
nasbyj: *Why* do we care if the key exists?

contrib/babelfishpg_tsql/sql/sys_functions.sql
```diff
@@ -2540,6 +2540,119 @@ CREATE OR REPLACE FUNCTION sys.json_query(json_string text, path text default '$
...
+    word_count = LENGTH(path_json) - LENGTH(REPLACE(path_json, ' ', '')) + 1; --To calculate the number of words in the path_json
```
nasbyj: Given that several cases below care about the first 2 arguments, it might be better to just break the path apart using `string_to_array()` or `regexp_split_to_array()`. Also, there's a serious bug here. Consider this perfectly valid JSON:

```
{
    "key": "value",
    "other key":     NULL
)
```

contrib/babelfishpg_tsql/sql/sys_functions.sql
```diff
@@ -2540,6 +2540,119 @@ CREATE OR REPLACE FUNCTION sys.json_query(json_string text, path text default '$
...
+    
+
+    -- This if else block is added to set the create_if_missing and append_modifier flags
```
nasbyj: What the block is doing is already pretty obvious; what's absolutely not obvious is *why* it's doing it.

contrib/babelfishpg_tsql/sql/sys_functions.sql
```diff
@@ -2540,6 +2540,119 @@ CREATE OR REPLACE FUNCTION sys.json_query(json_string text, path text default '$
...
+    IF word_count=1 THEN --The path_json has only 1 word
+        create_if_missing='TRUE';
```
nasbyj: Please don't wrap boolean values being assigned to a boolean in quotes. It's more to read and more work for the engine as well.

contrib/babelfishpg_tsql/sql/sys_functions.sql
```diff
@@ -2540,6 +2540,119 @@ CREATE OR REPLACE FUNCTION sys.json_query(json_string text, path text default '$
...
+        append_modifier='FALSE';
+    ELSIF word_count=2 THEN --The path_json has 2 words
+        mode = SPLIT_PART(path_json COLLATE sql_latin1_general_cp1_cs_as, ' ',1);
+        IF mode='append' THEN
```
nasbyj: Do we *really* want this comparison to be case-sensitive?

contrib/babelfishpg_tsql/sql/sys_functions.sql
```diff
@@ -2540,6 +2540,119 @@ CREATE OR REPLACE FUNCTION sys.json_query(json_string text, path text default '$
...
+            append_modifier:='TRUE';
+        ELSIF mode='strict' THEN
+            create_if_missing:='FALSE';
+        END IF;
+    ELSIF word_count=3 THEN --The path_json has 3 words
+        modifier = SPLIT_PART(path_json COLLATE sql_latin1_general_cp1_cs_as, ' ',1);
```
nasbyj: In the stanza above we call the 1st word "mode", but here we call it "modifier" even though it's checking for append, which is very confusing.

contrib/babelfishpg_tsql/sql/sys_functions.sql
```diff
@@ -2540,6 +2540,119 @@ CREATE OR REPLACE FUNCTION sys.json_query(json_string text, path text default '$
...
+        mode = SPLIT_PART(path_json COLLATE sql_latin1_general_cp1_cs_as, ' ',2);
+        IF modifier='append' THEN
+            append_modifier:='TRUE';
+        END IF;
+        IF mode='strict' THEN
+            create_if_missing:='FALSE';
+        END IF;
```
nasbyj, line 2686: What about the case of "strict append"? Is that definitely considered to be invalid? If so, maybe we should throw an error? (What does MSSQL do here?) Obviously if it is valid then we need to support it.

contrib/babelfishpg_tsql/sql/sys_functions.sql
```diff
@@ -2540,6 +2540,119 @@ CREATE OR REPLACE FUNCTION sys.json_query(json_string text, path text default '$
...
+    END IF;
+
+
+    --This if else block is to call the jsonb_set function based on the create_if_missing and append_modifier flags
+    
+    IF append_modifier THEN --When append modifier is present
```
nasbyj: Comment isn't useful; just remove it.

contrib/babelfishpg_tsql/sql/sys_functions.sql
```diff
@@ -2540,6 +2540,119 @@ CREATE OR REPLACE FUNCTION sys.json_query(json_string text, path text default '$
...
+                END IF;
+            END IF;
+        ELSE
+            IF create_if_missing='FALSE' THEN
+                RAISE EXCEPTION 'Property cannot be found on the specified JSON path.';
+            ELSE
+                result_json = jsonb_insert(json_expression,new_jsonb_path,to_jsonb(array_agg(new_value))); -- array_agg is used to convert the new_value text into array format as we append functionality is being used
+            END IF;
+        END IF;
+    ELSE --When no append modifier is present
```
nasbyj, line 2729: *This* comment actually is useful since this ELSE is 26 lines away from it's IF.

contrib/babelfishpg_tsql/sql/sys_functions.sql
```diff
@@ -2540,6 +2540,119 @@ CREATE OR REPLACE FUNCTION sys.json_query(json_string text, path text default '$
...
+                END IF;
+            ELSE
+                IF create_if_missing='FALSE' THEN
+                    RAISE EXCEPTION 'Property cannot be found on the specified JSON path.';
+                ELSE
+                    result_json = jsonb_set_lax(json_expression,new_jsonb_path,to_jsonb(new_value),FALSE);
+                END IF;
+            END IF;
+        END IF;
+    END IF;
```
nasbyj: It'd be worth adding a comment to the effect of `-- IF append_modifier` here.

contrib/babelfishpg_tsql/sql/sys_functions.sql
nasbyj: Also, nitpick: should be `key_exists`.

contrib/babelfishpg_tsql/sql/sys_functions.sql
```diff
@@ -2540,6 +2540,119 @@ CREATE OR REPLACE FUNCTION sys.json_query(json_string text, path text default '$
...
+
+
+    --This if else block is to call the jsonb_set function based on the create_if_missing and append_modifier flags
+    
+    IF append_modifier THEN --When append modifier is present
+        IF key_exist THEN
+            key_value = jsonb_path_query_first(json_expression,json_path::jsonpath); -- To get the value of the key
+            type_key_value = jsonb_typeof(key_value);
```
nasbyj: Nit: `key_value_type`, or maybe `key_value__type`. Or maybe `type_of_key_value`.

contrib/babelfishpg_tsql/sql/sys_functions.sql
```diff
@@ -2540,6 +2540,119 @@ CREATE OR REPLACE FUNCTION sys.json_query(json_string text, path text default '$
...
+            IF type_key_value='array' THEN
+                len_array = jsonb_array_length(key_value);
+                new_jsonb_path = CONCAT(TRIM('}' FROM new_jsonb_path::TEXT),',',CAST(len_array AS TEXT),'}'); -- To change the path format into the required jsonb_insert path format
```
nasbyj: FWIW, `format()` might be clearer here than `concat()`. If you want to use concat I suggest splitting this call across several lines, ie:

```
new_jsonb_path = concat(
  '{'
  , trim(...)
  , '}'
)
```

nasbyj: Also, the comment would be MUCH more useful if it gave an example of the two formats.

contrib/babelfishpg_tsql/sql/sys_functions.sql
```diff
@@ -2540,6 +2540,119 @@ CREATE OR REPLACE FUNCTION sys.json_query(json_string text, path text default '$
...
+                IF new_value IS NULL THEN
+                    result_json = jsonb_insert(json_expression,new_jsonb_path,'null');
```
nasbyj: Please add a comment that we can't simply do "to_jsonb(coalesce(new_value, 'null'))" here, because that won't result in a JSON NULL".

test/JDBC/expected/BABEL-937.out
```diff
@@ -0,0 +1,210 @@
+-- Not Append
+--NOT NULL, LAX, KP  
+SELECT JSON_MODIFY('{"name":"John","skills":["C#","SQL"]}','$.name','Mike');
```
nasbyj: Please add some examples that use fully formatted JSON (such as what you'd get with `jsonb_pretty()`).

test/JDBC/expected/BABEL-937.out
```diff
@@ -0,0 +1,210 @@
...
+GO
+~~START~~
+nvarchar
+{"name": "Mike", "skills": ["C#", "SQL"]}
+~~END~~
+
+
+--NOT NULL, STRICT, KP
+SELECT JSON_MODIFY('{"name":"John","skills":["C#","SQL"]}','strict $.name','Mike');
```
nasbyj: Need to test upper/mixed case, as well as extra spaces, leading/trailing spaces etc. Even if those aren't supported we should know what error they're returning.

nasbyj (changes requested):

contrib/babelfishpg_tsql/sql/sys_functions.sql
anujsarda3: Yes, you are right. It will be better to use regexp_split_to_array() instead of this. I will make the changes and implement it using regexp_split_to_array().

contrib/babelfishpg_tsql/sql/sys_functions.sql
anujsarda3: Will add more explanation about the implementation in my next commit.

contrib/babelfishpg_tsql/sql/sys_functions.sql
anujsarda3: For append modifier:
If the key exists does not exists then we need to convert the new_value as array and then add it in the expression or else we do not need to convert the new_value to array_type and also we only check the value of the key only if the key_exists.

If append modifier is not present:
for strict mode the function works in different way if the key_exists or not. It returns an error if key exists and returns a new value in case key_exists.

Overall key_exists makes it easy to divide working of the function in different cases.

contrib/babelfishpg_tsql/sql/sys_functions.sql
anujsarda3: will add comment to explain the use of setting up the flags in the next commit.

contrib/babelfishpg_tsql/sql/sys_functions.sql
anujsarda3: So the syntax of the input path is defined as below:
[append] [ lax | strict ] $.<json path>

So even if the path is defined in the way "strict append", MSSQL throws an error 'JSON path is not properly formatted.' I have implemented this and can be seen in the next commit.

contrib/babelfishpg_tsql/sql/sys_functions.sql
anujsarda3: The path that I am splitting is the input json_path and not the input expression here. The syntax of the input json path is "`[append] [ lax | strict ] $.<json path>`".

contrib/babelfishpg_tsql/sql/sys_functions.sql
anujsarda3: have used format instead of concat

test/JDBC/expected/BABEL-937.out
anujsarda3: Added the test cases.

test/JDBC/expected/BABEL-937.out
anujsarda3: Have tested the upper/mixed case and also leading/trailing spaces and have handled them in the implementation. Also added the test cases in the BABLE-937.sql file.

Sairakan (dismissed):

test/JDBC/input/BABEL-937.sql
```diff
@@ -0,0 +1,124 @@
+-- NOT APPEND
+--NOT NULL, LAX, KP  
+SELECT JSON_MODIFY('{"name":"John","skills":["C#","SQL"]}','$.name','Mike');
```
2jungkook: BABEL-937-vu tests are executed in both normal and upgrade tests. So let's move all test cases in this file into BABEL-937-vu. In the upgrade tests, please also use views instead of raw select statements. You can use a view to test multiple select statements, e.g.,
```
create view babel_937_vu_v2 as
select JSON_MODIFY(...),
       JSON_MODIFY(...),
       ...
```

contrib/babelfishpg_tsql/sql/sys_functions.sql
```diff
@@ -2540,6 +2540,146 @@ CREATE OR REPLACE FUNCTION sys.json_query(json_string text, path text default '$
...
+/*
+This function is used to update the value of a property in a JSON string and returns the updated JSON string.
+
+It has been implemented in three parts:
+1) Set the append and create_if_missing flag as postgres functions do not directly take append and lax/strict mode in the jsonb_path.
+2) To convert the input path into the expected jsonb_path.
+3) To implement the main logic of the JSON_MODIFY function by dividing it into 8 different cases.
+
+*/
+CREATE OR REPLACE FUNCTION sys.json_modify(in expression sys.NVARCHAR,in path_json TEXT, in new_value TEXT)
```
2jungkook, line 2627: Are input (nvarchar, text, text) and output (nvarchar) data types consistent with T-SQL's?

contrib/babelfishpg_tsql/sql/sys_functions.sql
```diff
@@ -2540,6 +2540,146 @@ CREATE OR REPLACE FUNCTION sys.json_query(json_string text, path text default '$
...
+    It is necessary as postgres functions do not directly take append and lax/strict mode in the jsonb_path.    
+    */
+    IF word_count=1 THEN
+        json_path = path_split_array[1];
+        create_if_missing=TRUE;
+        append_modifier=FALSE;
+    ELSIF word_count=2 THEN 
+        json_path = path_split_array[2];
+        comparison_string = path_split_array[1]; -- append or lax/strict mode
+        IF comparison_string = 'append' THEN
```
nasbyj, line 2662: Are we intentionally making this a case-sensitive comparison?

contrib/babelfishpg_tsql/sql/sys_functions.sql
```diff
@@ -2540,6 +2540,146 @@ CREATE OR REPLACE FUNCTION sys.json_query(json_string text, path text default '$
...
+            append_modifier:=TRUE;
+        ELSIF comparison_string = 'strict' THEN
+            create_if_missing:=FALSE;
+        ELSIF comparison_string = 'lax' THEN
+            create_if_missing:=TRUE;
+        ELSE
+            RAISE EXCEPTION 'JSON path is not properly formatted.';
```
nasbyj: Please review https://www.postgresql.org/docs/current/error-style-guide.html; in particular, error messages are not sentences, so the trailing `.` should go.

Need to use an explicit SQL state here. It would also be good to provide a DETAIL message showing what the problem is.

contrib/babelfishpg_tsql/sql/sys_functions.sql
```diff
@@ -2540,6 +2540,146 @@ CREATE OR REPLACE FUNCTION sys.json_query(json_string text, path text default '$
...
+        END IF;
+    ELSE
+        RAISE EXCEPTION 'JSON path is not properly formatted.';
+    END IF;
+
+    -- To convert input jsonpath to the required jsonb_path format
+    json_path_convert = regexp_replace(json_path, '\$\.|]|\$\[' , '' , 'ig'); -- To remove the "$." and "]" sign from the string
```
nasbyj: nit: s/remove the/remove/

contrib/babelfishpg_tsql/sql/sys_functions.sql
```diff
@@ -2540,6 +2540,146 @@ CREATE OR REPLACE FUNCTION sys.json_query(json_string text, path text default '$
...
+        ELSE
+            RAISE EXCEPTION 'JSON path is not properly formatted.';
+        END IF;
+    ELSE
+        RAISE EXCEPTION 'JSON path is not properly formatted.';
+    END IF;
+
+    -- To convert input jsonpath to the required jsonb_path format
+    json_path_convert = regexp_replace(json_path, '\$\.|]|\$\[' , '' , 'ig'); -- To remove the "$." and "]" sign from the string 
+    json_path_convert = regexp_replace(json_path_convert, '\.|\[' , ',' , 'ig'); -- To replace the "." and "[" with the "," to change into required format
```
nasbyj: nit: s/replace the/replace/, s/with the/with/

contrib/babelfishpg_tsql/sql/sys_functions.sql
```diff
@@ -2540,6 +2540,146 @@ CREATE OR REPLACE FUNCTION sys.json_query(json_string text, path text default '$
...
+                */
+                new_jsonb_path = FORMAT('%s,%s}',TRIM('}' FROM new_jsonb_path::TEXT),len_array);
+                IF new_value IS NULL THEN
+                    result_json = jsonb_insert(json_expression,new_jsonb_path,'null'); -- This needs to be done because "to_jsonb(coalesce(new_value, 'null'))" does not result in a JSON NULL
+                ELSE
+                    result_json = jsonb_insert(json_expression,new_jsonb_path,to_jsonb(new_value));
+                END IF;
+            ELSE
+                IF NOT create_if_missing THEN
+                    RAISE EXCEPTION 'Array cannot be found in the specified JSON path.';
```
nasbyj: ERRMSG shouldn't be a sentence. Need SQLSTATE. Would be good to provide the user with a DETAIL message as well (which should be a sentence), as well as a HINT about how they could specify to create the key if it doesn't exist.

contrib/babelfishpg_tsql/sql/sys_functions.sql
```diff
@@ -2540,6 +2540,146 @@ CREATE OR REPLACE FUNCTION sys.json_query(json_string text, path text default '$
...
+    create_if_missing BOOL:=TRUE;
+    append_modifier BOOL:=FALSE;
+    key_exists BOOL;
+    key_value JSONB;
+    json_expression JSONB = expression::JSONB;
+    result_json sys.NVARCHAR;
+BEGIN
+    path_split_array = regexp_split_to_array(TRIM(path_json) COLLATE "C",'\s+');
+    word_count = array_length(path_split_array,1);
+    /*
```
nasbyj, line 2649: nit: It'd be better to follow community block-comment formatting standard:

```
/*
 * This if else block is added to set the create_if_missing and append_modifier flags.
 * These flags will be used to know the mode and if the optional modifier append is present in the input path_json.
 * It is necessary as postgres functions do not directly take append and lax/strict mode in the jsonb_path.  
 */

**Review by nasbyj — CHANGES_REQUESTED** [review]

File: contrib/babelfishpg_tsql/sql/sys_functions.sql (hunk 31 above)
Comment (anujsarda3) [reply contrib/babelfishpg_tsql/sql/sys_functions.sql:2662]: Yes, because it is a case-sensitive comparison. Otherwise it throws an error.

File: contrib/babelfishpg_tsql/sql/sys_functions.sql (hunk 30 above)
Comment (anujsarda3) [reply contrib/babelfishpg_tsql/sql/sys_functions.sql:2627]: Yes, they are consistent with T-SQL's datatypes.

File: contrib/babelfishpg_tsql/sql/sys_functions.sql (hunk 35 above)
Comment (anujsarda3) [reply contrib/babelfishpg_tsql/sql/sys_functions.sql]: Are you suggesting to display the SQL state error code instead of the default ‘33557097’?

File: test/JDBC/expected/BABEL-937-vu-prepare.out [hunk 37]
```diff
@@ -0,0 +1,107 @@
...
+
+create view babel_937_case_sensitive_1 as
+SELECT JSON_MODIFY('{"name":"John","skills":["C#","SQL"]}','strIct $.name','James'); 
+go
+
+create view babel_937_case_sensitive_2 as
+SELECT JSON_MODIFY('{"name":"John","skills":["C#","SQL"]}','  aPpend    Strict    $.skills   ',NULL);
+go
+
+create view babel_937_case_sensitive_3 as
```
2jungkook: Why this is related to case sensitive?

test/JDBC/expected/BABEL-937-vu-prepare.out
```diff
@@ -0,0 +1,107 @@
+create view babel_937_not_append_kp as
```
2jungkook, line 2: From reviewer's perspective, it is not easy to know what `not_append_kp` means and what's the purpose of each test cases. Not only for this specific case but also for other test cases, please add some comments explaining which functionality is tested.

contrib/babelfishpg_tsql/sql/sys_functions.sql
```diff
@@ -2515,6 +2515,154 @@ CREATE OR REPLACE FUNCTION sys.json_query(json_string text, path text default '$
...
+        create_if_missing = TRUE;
+        append_modifier = FALSE;
+    ELSIF word_count = 2 THEN 
+        json_path = path_split_array[2];
+        comparison_string = path_split_array[1]; -- append or lax/strict mode
+        IF comparison_string = 'append' THEN
+            append_modifier = TRUE;
+        ELSIF comparison_string = 'strict' THEN
+            create_if_missing = FALSE;
+        ELSIF comparison_string = 'lax' THEN
```
2jungkook, line 2666: Don't we need a test case for `lax`?

test/JDBC/expected/BABEL-937-vu-verify.out
```diff
@@ -0,0 +1,136 @@
...
+nvarchar#!#nvarchar#!#nvarchar#!#nvarchar
+{"name": "John", "skills": ["C#", "SQL", "Azure"]}#!#{"name": "John", "skills": ["C#", "SQL", "Azure"]}#!#{"name": "Mike", "skills": ["C#", "SQL", null], "surname": "Smith"}#!#{"name": "Mike", "skills": ["C#", "SQL", null], "surname": "Smith"}
+~~END~~
+
+
+select * from babel_937_append_knp_notarr_1
+go
+~~START~~
+nvarchar#!#nvarchar
+{"name": "Mike", "skills": ["C#", "SQL"], "surname": ["Smith"]}#!#{"k": [null], "name": "Mike", "skills": ["C#", "SQL"], "surname": "Smith"}
```
2jungkook, line 74: `surname` is added at the end, but `k` is added at the beginning. Is this expected?

contrib/babelfishpg_tsql/sql/sys_functions.sql
anujsarda3: As Lax mode is the default mode(when no mode is provided), so tests for lax mode are already covered in the test cases.

test/JDBC/expected/BABEL-937-vu-verify.out
anujsarda3: Yes, This is an expected and known behavior when using postgres functions.

test/JDBC/expected/BABEL-937-vu-prepare.out
anujsarda3: Resolved in latest commit

contrib/babelfishpg_tsql/sql/sys_functions.sql
2jungkook: Even if the behavior of lax and no mode is the same, there should be at least two test cases where lax is explicitly provided when word count is 2 or 3.

test/JDBC/expected/BABEL-937-vu-verify.out
2jungkook: In that case, please add some comments in the test file that the result can be different between Babelfish and T-SQL.

contrib/babelfishpg_tsql/sql/sys_functions.sql
anujsarda3: Resolved in latest commit.

test/JDBC/expected/BABEL-937-vu-verify.out
anujsarda3: Resolved in latest commit.

Sairakan (dismissed):

2jungkook (dismissed):

contrib/babelfishpg_tsql/sql/sys_functions.sql
nasbyj: It'd be nice to mention that in a comment then (and I hope there are test cases for this as well.

nasbyj (dismissed):
Approved, though see my one comment about explicitly stating that we want to be case sensitive.

contrib/babelfishpg_tsql/sql/sys_functions.sql
anujsarda3: Yes, there are test cases for case-sensitive comparison.

nasbyj (dismissed):

Sairakan (approved):

2jungkook (approved):

PR #640, opened 2022-09-28
Few improvements in SSL/TLS handling

Deepesh125:
### Description
This commit aims to add few enhancements in SSL/TLS handling code path. For example, 
1. We use SslRead (safe wrapper over secure_raw_read()) or SslWrite (safe wrapper over secure_raw_write())  inside SslHandShakeRead and SslHandShakeWrite to read and write data on socket during ssl handshake. We improved these SslHandShakeRead and SslHandShakeWrite APIs to just return to caller whenever error/EOF occurs during reading or writing on socket. 
2.  We have to write extra TDS header on the top of what we asked to write inside SslHandShakeWrite which caller is not aware of. So improved SslHandShakeWrite to return number of bytes written on socket except TDS header. 
 
### Testing
1. JDBC test framework
2. Tested using custom JDBC program with using different version of SSL version TLSv1, TLSv1.1 and TLSv1.2 (Java does not let us use older version of SSL for example, SSLv3)
4. Tested using Data direct client. Server emits correct error when we try to use SSL version outside the range of babelfishpg_tds.tds_ssl_max_protocol_version and babelfishpg_tds.tds_ssl_min_protocol_version.
5. Also configured GUC like babelfishpg_tds.tds_ssl_max_protocol_version and babelfishpg_tds.tds_ssl_min_protocol_version. And tested all these again.
6. Tested using SQLCMD connections:

```
my_babelfish_db=# select * from pg_stat_ssl;
pid  | ssl | version |           cipher            | bits | client_dn | client_serial | issuer_dn
-------+-----+---------+-----------------------------+------+-----------+---------------+-----------
61230 | t   | TLSv1.2 | ECDHE-RSA-AES128-GCM-SHA256 |  128 |           |               |
61922 | f   |         |                             |      |           |               |
(2 rows)
```

Issues Resolved: BABEL-3574

### Check List


contrib/babelfishpg_tds/src/backend/tds/tdssecure.c
```diff
@@ -76,7 +76,8 @@ SslHandShakeRead(BIO *h, char *buf, int size)
 {
 	int res = 0;
 
-	res = SslRead(h, buf, size);
+	if ((res = SslRead(h, buf, size)) <= 0)
+		return res;
```
surendravishnoi, line 80: what does negative res indicate?

contrib/babelfishpg_tds/src/backend/tds/tdssecure.c
```diff
@@ -91,18 +92,24 @@ SslHandShakeRead(BIO *h, char *buf, int size)
 					/* Read the complete remaining of the header and throw away the bytes */
 					while(res < remainingRead)
 					{
-						res += secure_raw_read(((Port *) BIO_get_data(h)), tempBuf,
-								remainingRead - res);
+						int tmp_res = 0;
+						if ((tmp_res = SslRead(h, tempBuf, remainingRead - res)) <= 0)
+						{
+							return tmp_res;
```
surendravishnoi, line 98: what will returning 0 will mean here? Even though res might be non-zero

contrib/babelfishpg_tds/src/backend/tds/tdssecure.c
```diff
@@ -91,18 +92,24 @@ SslHandShakeRead(BIO *h, char *buf, int size)
...
+						}
+						res += tmp_res;
 					}
 
 					/*
 					 * Read the actual data and return the res of the actual data read
 					 * Don't worry if complete read, Openssl library will take care
 					 */
-					res = secure_raw_read(((Port *) BIO_get_data(h)), buf, size);
+					if ((res = SslRead(h, buf, size)) <= 0)
```
surendravishnoi, line 107: why change from secure_raw_read to Ssloread?

surendravishnoi (commented):
Please use new PR template. Do we have any plan to add automated test for this.

contrib/babelfishpg_tds/src/backend/tds/tdssecure.c
Deepesh125: As per documentation of [recv()](https://man7.org/linux/man-pages/man2/recvmsg.2.html), Negative value means there was an error. So we should return to the caller instead of going ahead with any other operations.

contrib/babelfishpg_tds/src/backend/tds/tdssecure.c
Deepesh125: I checked one of the use case of BIO_read() from openssl code and also checked documentation for recv(), this should only occur if there is EOF on socket.

contrib/babelfishpg_tds/src/backend/tds/tdssecure.c
Deepesh125: Because SslRead is safe wrapper around secure_raw_read()

surendravishnoi (approved):

kuntalghosh (approved):

PR #650, opened 2022-09-29
Update upgrade test template

rishabhtanwar29:
### Description
This commit does the following:
1. Adds upgrade paths for version 13.7 and 13.8. Currently tests for 13_6 are copied
to 13_7 and 13_8 upgrade directories.
3. Fixes branch template to use stable branches of extensions instead of tags.
4. Fixes branches in major/minor upgrade workflows for empty databases.

Task: BABEL-3583

### Check List


.github/configuration/upgrade-test-configuration.yml
```diff
@@ -65,4 +65,24 @@ upgrade-version: [[
...
+    upgrade-type: major 
+  }
+],
+[
+  {
+    version: 13.8,
+    upgrade-type: null
+  },
+  { 
+    version: target.latest,
```
surendravishnoi, line 85: we should also add minor upgrade for 13.7/13.8 as well

rishabhtanwar29: included minor version upgrade step in each upgrade path

.github/configuration/upgrade-test-configuration.yml
```diff
@@ -65,4 +65,32 @@ upgrade-version: [[
...
+    upgrade-type: null
+  },
+  {
+    version: 13.8,
+    upgrade-type: minor
+  },
+  { 
+    version: target.latest,
+    upgrade-type: major 
+  }
```
2jungkook, line 87: Is there a reason to include only `13.4->13.7->latest` and `13.4->13.8->latest`? I think at least we should include `13.7->latest` and `13.8->latest`, otherwise `13_7/schedule` and `13_8/schedule` are not tested.

In the long run, we should have a rule which upgrade paths are included for every PR. Here is one suggestion:
1. For every PR, test only 1-level upgrade paths so that we can ensure every schedule files are used at least once, e.g.,
    - 13.4 -> 14.latest (due to the MVU limitation, it is actually 13.4 -> 13.latest -> 14.latest)
    - 13.5 -> 14.latest (due to the MVU limitation, it is actually 13.5 -> 13.latest -> 14.latest)
    - 13.x -> 14.latest where x >= 6
    - 14.x -> 14.latest where x >= 3
2. Regularly (once a week?), test multi-level upgrade paths, e.g.,
    - 13.4 -> 13.6 -> 14.latest
    - 13.5 -> 13.6 -> 14.latest
    - 13.6 -> 13.7 -> 14.latest
    - 13.7 -> 13.8 -> 14.5 -> 14.latest
    - more combinations
 
Any other suggestions are welcome!

rishabhtanwar29: The reason to include 2-level upgrade paths is that it will test both minor and major version upgrades in a single github workflow. It is not like schedule files will not be used as the schedule files for middle versions will still be used to identify the verification files for them. I agree with your suggestion that we should also test 1-level upgrade paths but in that case we would have to add a lot of workflows for 1-level and 2-level upgrade paths.
My idea is to only include 1-level upgrade path for one of 13.6, 13.7 or 13.8 version (since the extension version is same for all of them and we already have an upgrade path from 13.6 -> 14.latest) and test 2-level upgrade paths for the rest. Let me know if you think otherwise.

2jungkook: 1. For example, BABEL-1493 is in `13_7/schedule` but is not tested in `13.4 -> 13.7 -> 14.latest` workflow.
2. If you choose one of 13.6, 13.7 or 13.8 because of the same extension version, then why do we need the following three workflows ?
    - `13.4 -> 13.6 -> 14.latest`
    - `13.4 -> 13.7 -> 14.latest`
    - `13.4 -> 13.8 -> 14.latest`

surendravishnoi: I think we should not copy everything from 13.6 to 13.7/13.8. These should be covered as part of 13.6->13.7->latest. 13.7 should have only its specific tests which will run in 13.7 -> latest/13.8. Otherwise, 13.7/13.8 runs will take time without much added value.

rishabhtanwar29: Actually the number of version specific tests are quite less than the tests coming from schedule file so not copying specific tests will not make much difference.

2jungkook (dismissed):

surendravishnoi (dismissed):

surendravishnoi (approved):

PR #651, opened 2022-09-29
Performance optimisation for INSERT BULK

KushaalShroff:
### Description
Performance optimisation for INSERT BULK

Initially Insert Bulk was implemented using SPI_execute_with_args, which practically were single inserts in the executor level. This made database migration slow for our users, with this task we have moved on from SPI to use PG’s multi insert functionality. PG’s multi insert feature which is much faster since we can write just a single WAL for all of the inserts at the end and we need to lock/unlock page only once. Moreover, there is no need to parse/plan/optimise the query (When compared to initial implementation) thus bypassing directly to inserting multiple tuples into the table at once.
 
Task: BABEL-3503

### Functional Testing
Dotnet framework has elaborate tests that cover INSERT BULK AND BCP for each and every datatype, identity columns, keep nulls and other insert bulk options. One can check the BCP test plan we had created during the initial feature implementation.

### Performance testing

Time Taken for BCP with the initial implementation:

Dataset | -b not specified | -b 100 | -b 1000 | -b 5000
-- | -- | -- | -- | --
 10K | 469 | 859 | 516 | 500
100K | 4266 | 7500 | 4781 | 4406
1M | 41485 | 75422 | 47359 | 44172
10M | NA | 749142 | 474547 | 438672

Time Taken for BCP with changes in this PR:

Dataset | -b not specified | -b 100 | -b 1000 | -b 5000
-- | -- | -- | -- | --
10K | 93 | 313 | 125 | 109
100K | 422 | 2750 | 672 | 453
1M | 3656 | 26813 | 6031 | 4110
10M | NA | 282781 | 65234 | 42968

Thus proving a speedup of 8x on average when compared to the initial implementation

### Check List


contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -2716,54 +2716,49 @@ exec_stmt_insert_execute_select(PLtsql_execstate *estate, PLtsql_expr *query)
...
 	MemoryContext	oldContext;
 	Oid schema_oid = InvalidOid;
 
+	oldContext = MemoryContextSwitchTo(TopMemoryContext);
+
+	/*
+	 * We use a global variable so that we do not need to call BeginBulkCopy
+	 * in case of implicit batching, which saves time.
+	 */
+	cstmt = (BulkCopyStmt *) palloc0(sizeof(BulkCopyStmt));
```
kuntalghosh, line 2779: You should allocate memory in TopMemoryContext only if you want to keep the allocations throughout the backend lifetime. The memory doesn't get reclaimed during an error.
IIUC, you want to palloc this memory for each insert bulk statement you're executing. In this case, it's better to allocate in a separate memory context that you can free after each execution (even in case of error).

contrib/babelfishpg_tsql/src/pltsql_bulkcopy.c
```diff
@@ -0,0 +1,938 @@
+/*-------------------------------------------------------------------------
+ *
+ * pltsql_bulkcopy.c		- Bulk Copy for PL/tsql
+ *
```
kuntalghosh, line 4: >  * Portions Copyright (c) 2022, AWS
>  * Portions Copyright (c) 1996-2022, PostgreSQL Global Development Group
>  * Portions Copyright (c) 1994, Regents of the University of California

contrib/babelfishpg_tsql/src/pltsql_bulkcopy.c
```diff
@@ -0,0 +1,938 @@
...
+			uint64 *processed);
+
+void EndBulkCopy(BulkCopyState cstate);
+
+static uint64
+ExecuteBulkCopy(BulkCopyState cstate, int rowCount, int colCount,
+				Datum *Values, bool *Nulls);
+
+static List *
+BulkCopyGetAttnums(TupleDesc tupDesc, Relation rel, List *attnamelist);
```
kuntalghosh: extern functions shouldn't be declared again in the c file. You've anyway included the header file.

kuntalghosh (commented):
1. It's not clear how you've tested these changes.
2. What's the performance numbers you see with this change?

Please update the PR description with the above info.

contrib/babelfishpg_tds/src/backend/tds/tdsbulkload.c
```diff
@@ -920,6 +902,9 @@ ProcessBCPRequest(TDSRequest request)
 
 				RESUME_CANCEL_INTERRUPTS();
 
+				/* Using Same callback function to fo the clean-up. */
```
surendravishnoi: misspell fo -> do

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -2716,54 +2716,49 @@ exec_stmt_insert_execute_select(PLtsql_execstate *estate, PLtsql_expr *query)
 
 int exec_stmt_insert_bulk(PLtsql_execstate *estate, PLtsql_stmt_insert_bulk *stmt)
```
surendravishnoi, line 2768: this should be renamed to better reflect new logic

contrib/babelfishpg_tsql/src/pltsql.h
```diff
@@ -915,7 +915,7 @@ typedef struct PLtsql_stmt_insert_bulk
     char  *table_name;
     char  *schema_name;
     char  *db_name;
-    char  *column_refs;
+	List *column_refs;
```
surendravishnoi, line 919: indentation

contrib/babelfishpg_tsql/src/tsqlIface.cpp
```diff
@@ -4290,15 +4290,13 @@ makeInsertBulkStatement(TSqlParser::Dml_statementContext *ctx)
 		/* create a list of columns to insert into */
 		if (!column_list.empty())
 		{
-			for (size_t i = 0; i < column_list.size() - 1; i++)
+			for (size_t i = 0; i < column_list.size(); i++)
 			{
-				if (column_list[i]->simple_column_name())
```
surendravishnoi, line 4374: can you explain why simple_column_name check is removed and downcasing is done here

contrib/babelfishpg_tsql/src/pltsql_bulkcopy.c
```diff
@@ -0,0 +1,938 @@
...
+
+		foreach(l, attnamelist)
+		{
+			char	   *name = (char *)lfirst(l);
+			int			attnum;
+			int			i;
+
+			/* Lookup column name. */
+			attnum = InvalidAttrNumber;
+			for (i = 0; i < tupDesc->natts; i++)
```
surendravishnoi, line 190: It will be more efficient to keep this as outer loop as other list is local

contrib/babelfishpg_tsql/src/pltsql_bulkcopy.c
```diff
@@ -0,0 +1,938 @@
...
+					   ti_options,
+					   buffer->bistate);
+	MemoryContextSwitchTo(oldcontext);
+
+	for (i = 0; i < nused; i++)
+	{
+		/*
+		 * If there are any indexes, update them for all the inserted tuples.
+		 */
+		if (resultRelInfo->ri_NumIndices > 0)
```
surendravishnoi, line 374: can go outside the loop

contrib/babelfishpg_tsql/src/pltsql_bulkcopy.c
```diff
@@ -0,0 +1,938 @@
...
+		{
+			List	   *recheckIndexes;
+
+			cstate->cur_rowno = buffer->linenos[i];
```
surendravishnoi, line 378: where is this used?

contrib/babelfishpg_tsql/src/pltsql_bulkcopy.c
```diff
@@ -0,0 +1,938 @@
...
+ * CopyMultiInsertBuffer items stored in CopyMultiInsertInfo's
+ * multiInsertBuffers list.
+ */
+#define MAX_BUFFERED_TUPLES		1000
+
+/* Trim the list of buffers back down to this number after flushing */
+#define MAX_PARTITION_BUFFERS	32
+
+/* Stores multi-insert data related to a single relation. */
+typedef struct CopyMultiInsertBuffer
```
surendravishnoi, line 54: How much is this code different from PG copy? Did we consider reusing PG functions here?

contrib/babelfishpg_tsql/src/pltsql_bulkcopy.c
KushaalShroff: In BulkCopyErrorCallback:
```
errcontext("Bulk Copy for %s, row: %ld  (doesn't take implicit batching into consideration)",
				cstate->cur_relname,  cstate->cur_rowno);
```
This will get printed every time we run into an error

contrib/babelfishpg_tsql/src/pl_exec-2.c
KushaalShroff: This function still executes INSERT BULK query.
execute_bulk_load_insert function BULK LOAD Request.

contrib/babelfishpg_tsql/src/tsqlIface.cpp
KushaalShroff: 1. Based on the grammar, column_list[i] will always have simple_column_name
```
insert_bulk_column_definition
    : simple_column_name (data_type system_versioning_column? | AS expression PERSISTED? ) ( special_column_option | (COLLATE (id | DEFAULT)) | null_notnull )*
      ( column_constraint? IDENTITY (LR_BRACKET sign? seed=DECIMAL COMMA sign? increment=DECIMAL RR_BRACKET)? )? for_replication? ROWGUIDCOL?
      column_constraint* column_inline_index?
    ;
```
column_list NULL check has been taken care.
We only need to check if column_refs is NOT NULL

2. We downcase the column names here, since later we directly use the same list to match with the the table's columns (which by default will be lowercase in PG)

contrib/babelfishpg_tsql/src/pltsql.h
KushaalShroff: earlier indentation seems to be of 4 spaces, will correct them instead

contrib/babelfishpg_tsql/src/pltsql_bulkcopy.c
KushaalShroff: We have re-used the static functions which start with CopyMultiInsert (with few changes)
Things to note:
1. PG COPY functionality is different since it reads from a file.
3. These functions are static to PG COPY and also provide additional help in maintaining the state for reading from files. (In our design discussion we brought this up and it was agreed that for now the best would be to add comments for the functions re-used)

contrib/babelfishpg_tsql/src/pltsql_bulkcopy.c
KushaalShroff: This loop is looking up for the next attnum that matches with user provided column-name (for each column-name in column-list). Check line 204-205.
I believe It has to be done this way.

contrib/babelfishpg_tsql/src/pl_exec-2.c
KushaalShroff: We have a copycontext which is part of BulkCopyStateData which is responsible for holding data during the actually copy/insertion.

But for your comment, we are creating cstmt (BulkCopyStmt) which communicates information from insert bulk query (sql batch req1) to the bulk load request (req2).That will require TOP Memory context to persist information outside per-query context. Even if we create a new context here, it will not give us additional benefit.

Regarding your concern on the cleanup of these allocations, please check these links to the code where we have called cleanup from TDS for error and non error scenarios:
1. Cleanup from TDS CATCH block:
[https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/651/files#diff-f01df05dcab287463fcee6b45e4[…]ba61e5c0f4ebbdb848343bdd32832R906](https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/651/files#diff-f01df05dcab287463fcee6b45e49fa8f46bba61e5c0f4ebbdb848343bdd32832R906)
2. For non-error scenario we cleanup here:
[https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/651/files#diff-f01df05dcab287463fcee6b45e4[…]5c0f4ebbdb848343bdd32832R787-R789](https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/651/files#diff-f01df05dcab287463fcee6b45e49fa8f46bba61e5c0f4ebbdb848343bdd32832R787-R789)

Actual cleanup:
Cleanup inside callback function:
[https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/651/files#diff-3ca2886287e484395b98c3d6f90[…]ef1e975c1bb1c79b93e7abR2789-R2792](https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/651/files#diff-3ca2886287e484395b98c3d6f90cb766e4f75c5cd9ef1e975c1bb1c79b93e7abR2789-R2792)

contrib/babelfishpg_tsql/src/pltsql_bulkcopy.c
KushaalShroff: I did not understand the suggestion or concern.

If you are suggesting that we can refactor it outside the loop then it would not be possible since we need to ExecInsertIndexTuples for each buffer->slots[i]

contrib/babelfishpg_tsql/src/pl_exec-2.c
surendravishnoi: there are still transaction context and message context which goes beyond query. Top memory context should be last one to use

contrib/babelfishpg_tsql/src/pltsql_bulkcopy.c
surendravishnoi: you can go through each attribute and see if it matches one of provided column names. Not sure why it can't happen other way around

contrib/babelfishpg_tsql/src/pltsql_bulkcopy.c
surendravishnoi: I meant rather than skipping each slot when there are no indices, skip the loop altogether if there are no indices
if (resultRelInfo->ri_NumIndices > 0)
{
   for (i = 0; i < nused; i++)
}
But you need to call ExecClearTuple anyway, so you can ignore it

contrib/babelfishpg_tsql/src/tsqlIface.cpp
surendravishnoi: just make sure that we do not need original name in some other context otherwise are loosing it from here itself

contrib/babelfishpg_tsql/src/pltsql_bulkcopy.c
kuntalghosh: Looks like this function is copy of CopyGetAttnums.
1. We should add a comment regarding the same.
2. Since we've copied this function, we should have the coding style. 

I also don;t see the rowversion related hook in this function.

contrib/babelfishpg_tsql/src/pltsql_bulkcopy.c
kuntalghosh: Again this is the exact same function defined in copyfrom.c

kuntalghosh (commented):
Since we've so much code from copy.c and copyfrom.c, we need to a way to maintain these files in future. Otherwise, it'll have the same difficulties as SSL code. One idea could be:
1. Copy both copy.c and copyfrom.c file in the extension.
2. Comment out the functions that you don't need.
3. If you need to edit a function in that file, add a comment with BABEL prefix in that code.
4. Include the c file in the new file and add all your changes.
In this way, you'll know which are Babel related changes. 

Reply: 
> How will we face the same issue like SSL? SSL had lots of intermingling and dependence on the copied code. This is simple single responsibility utility functions. In the end we are responsible for calling multi-insert. Not someone else.

contrib/babelfishpg_tsql/src/pltsql_bulkcopy.c
KushaalShroff: As mentioned earlier and in my design doc, functions starting with CopyMultiInsertInfo are copied and some are pruned and made changes to cater to TSQL Bulk Copy context.

contrib/babelfishpg_tsql/src/pltsql_bulkcopy.c
KushaalShroff: It is similar, we required to have a separate function since the list attnamelist is list of "Name" structures which gets created by gram.y parser, whereas in ANTLR we get a List of strings. Moreover Rowversion datatype are allowed in BulkCopy.

kuntalghosh (approved):

PR #652, opened 2022-09-29
Login who owns database should always be dbo also without sysadmin role

wboettge:
### Description

In SQL Server, the login that creates a database is automatically the
database owner. This means that the login can always access the
databases they create.

However, previously in babelfish, if a login is stripped of the sysadmin
role after creating a database, and the login does not have a
user associated with the database, then the login can no longer access
the database. Now, when a login attempts to access and there
is not an associated user, the login will be considered the dbo and
allowed access.

Task: BABEL-3548

### Check List

### DIFF
--- a/contrib/babelfishpg_tsql/src/catalog.c
+++ b/contrib/babelfishpg_tsql/src/catalog.c
@@ -229,6 +229,24 @@ char *get_db_name(int16 dbid)
 	return name;
 }
 
+char *get_owner_of_db(const char *dbname)
+{
+	char				*owner = NULL;
+	HeapTuple 			tuple;
+	Form_sysdatabases 	sysdb;
+
+	  tuple = SearchSysCache1(SYSDATABASENAME, CStringGetTextDatum(dbname));
+
+	  if (!HeapTupleIsValid(tuple))
+		  return InvalidDbid;
+
+	  sysdb = ((Form_sysdatabases) GETSTRUCT(tuple));
+	  owner = NameStr(sysdb->owner);
+	  ReleaseSysCache(tuple);
+
+	return owner;
+}
+
 const char *get_one_user_db_name(void)
 {
 	HeapTuple 		tuple;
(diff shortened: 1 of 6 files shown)


contrib/babelfishpg_tsql/src/dbcmds.c
```diff
@@ -832,3 +834,22 @@ drop_related_bbf_namespace_entries(int16 dbid)
...
 }
+
+const char *
+get_owner_of_db(const char *dbname)
+{
+	char				*owner = NULL;
+	HeapTuple 			tuple;
+	Form_sysdatabases 	sysdb;
+
+	  tuple = SearchSysCache1(SYSDATABASENAME, CStringGetTextDatum(dbname));
```
xhfanhe: Please fix the indentation here

test/JDBC/input/BABEL-3548.mix
```diff
@@ -0,0 +1,62 @@
+-- tsql
+create login test_login1 with password='123';
```
xhfanhe: For all the created objects in this file, please add a `babel_3548` at the beginning of the object name, in this way if something went wrong, we can easily locate where the error happens

contrib/babelfishpg_tsql/src/dbcmds.c
```diff
@@ -832,3 +834,22 @@ drop_related_bbf_namespace_entries(int16 dbid)
...
+get_owner_of_db(const char *dbname)
+{
+	char				*owner = NULL;
+	HeapTuple 			tuple;
+	Form_sysdatabases 	sysdb;
+
+	tuple = SearchSysCache1(SYSDATABASENAME, CStringGetTextDatum(dbname));
+
+	if (!HeapTupleIsValid(tuple))
+		return InvalidDbid;
```
xhfanhe: I feel it doesn't make much sense to return InvalidDbid here. We can consider to throw an error if the given db name is invalid. And please add some comment ahead of this function saying that caller is responsible for validating db name before calling this function.

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -936,22 +937,25 @@ get_user_for_database(const char *db_name)
 {
 	const char		*user = NULL;
 	const char		*login;
+	bool			login_is_db_owner;
 
 	login = GetUserNameFromId(GetSessionUserId(), false);
 	user = get_authid_user_ext_physical_name(db_name, login);
+	login_is_db_owner = 0 == strncmp(login, get_owner_of_db(db_name), NAMEDATALEN);
```
xhfanhe, line 944: I tend to think it's better to use strcmp here, is there any specific considerations here?

xhfanhe (approved):

PR #671, opened 2022-10-04
Accessing a db without required permissions should not change the current db

wboettge:
### Description

Accessing a database with use without required permissions no longer
changes the current database. Previously, if a login attempted to access
a database they did not have permissions for, running SELECT DB_NAME()
after attepting to switch to the databaseb would return the database
they had been denied access from. Additionally, a shared lock would be
claimed, preventing other logins from deleting the database.

Now, the current_db session variables are unchanged and a lock is not
claimed following a failed attempt to access a database.
 
Task: BABEL-3548

### Check List


contrib/babelfishpg_tsql/src/session.c
```diff
@@ -60,56 +61,80 @@ bbf_set_current_user(const char *user_name)
...
 	/* set current user */
 	bbf_set_current_user(user);
 	current_user_id = GetUserId();
 
 	/* set search path */
+	set_search_path_for_user_schema(db_name, user);
+}
+
+static void
+set_search_path_for_user_schema(const char* db_name, const char* user)
```
wboettge, line 118: Does it make sense for this helper function to be static? Also is it declared in the right place at the start of the file?

xhfanhe: Yes I think it's good to keep this static to avoid exposure as much as possible.

xhfanhe (approved):

PR #675, opened 2022-10-05
Fix SELECT FOR XML/JSON behavior

Sairakan:
### Description

SELECT FOR XML/JSON were both originally implemented by converting the select query into a function call that then called SPI_execute() against the original SELECT statement. While that implementation works for simple SELECT statements, it fails to execute properly against queries where the SELECT statement is a subquery which references CTEs or correlated aliases outside of the scope of the sub-SELECT. This is due to the fact that the SELECT statement is passed as a string argument to the internal function call for SPI. Thus, any external references are lost when SPI_execute() is called.

We resolve this issue by changing the implementation of SELECT FOR XML/JSON to use aggregate functions. This way, there is no need to store any query strings, and the executor will fully execute the sub-SELECT prior to passing the result rows to the aggregate function, allowing us to retain the context of any outer CTEs or correlations.

Matching engine PR: https://github.com/babelfish-for-postgresql/postgresql_modified_for_babelfish/pull/65
 
### Issues Resolved

BABEL-1876

### Testing

- [x] Regression testing against older tests
- [x] Test cases with correlated CTEs/subqueries

Performance testing results:
|# of rows | SQL Server | Old Implementation | New Implemenation
--- | --- | --- | ---
|10,000 | 10 ms | 180 ms | 176 ms|
|100,000 | 435 ms | 1816 ms | 1774 ms|
|1,000,000 | 8047 ms | 18568 ms | 18066 ms|

### Check List


test/JDBC/upgrade/latest/schedule
```diff
@@ -102,6 +102,7 @@ Test-Computed-Columns
 BABEL-1189
 BABEL-1243
 BABEL-1493
+BABEL-1876
```
surendravishnoi: Latest is only for self MVU. This change has minor upgrade implications as well. Add in 14.5 schedule as well.

Sairakan: Added

test/python/expected/upgrade_validation/expected_dependency.out
```diff
@@ -684,7 +684,6 @@ Function sys.tsql_get_constraintdef(oid)
 Function sys.tsql_get_functiondef(oid)
 Function sys.tsql_get_returntypmodvalue(oid)
 Function sys.tsql_query_to_xml(text,integer,text,boolean,text)
```
2jungkook: Could you add a test for this function as well?

test/JDBC/expected/BABEL-1876-vu-verify.out
```diff
@@ -0,0 +1,14 @@
+SELECT x,
+(SELECT y FROM babel_1876_vu_t2 WHERE b = t.b FOR XML PATH ('')) AS mycol
+FROM babel_1876_vu_t1 t
+go
+~~START~~
+int#!#ntext
+~~END~~
+
+select * from babel_1876_vu_v1
```
2jungkook: Could you also check the cases where the correlated FOR XML subquery is in `where` or `from` clause?

test/JDBC/upgrade/latest/schedule
```diff
@@ -138,8 +138,10 @@ BABEL-3646
 BABEL-383
 BABEL-405
 BABEL-937
-BABEL-3407
 forjson
+forjson-datatypes
```
surendravishnoi, line 143: this should move to 14.6 schedule to test Narwal to Ocelot

contrib/babelfishpg_tsql/sql/sys_functions.sql
```diff
@@ -1,22 +1,123 @@
...
+CREATE OR REPLACE FUNCTION sys.tsql_query_to_xml_sfunc(
+    state TEXT,
+    rec ANYELEMENT,
+    mode int,
+    element_name text,
+    binary_base64 boolean,
+    root_name text
+) RETURNS TEXT
+AS 'babelfishpg_tsql', 'tsql_query_to_xml_sfunc'
+LANGUAGE C COST 100;
```
2jungkook: Just wondering where did `COST 100` come from?

contrib/babelfishpg_tsql/sql/sys_functions.sql
```diff
@@ -1,22 +1,123 @@
...
+THEN
+    -- '{' indicates that root was specified, so add the corresponding end tag
+    rootname = (regexp_match(state, '<([^\/>]+)[\/]*>' COLLATE C))[1];
+    RETURN (substr(state, 2) || '</' || rootname || '>')::XML;
+ELSE 
+    RETURN state::XML;
+END IF;
+END;
+$$
+LANGUAGE PLPGSQL STRICT;
```
2jungkook: 1. Do we have a test for `STRICT` keyword?
2. Can we use `IMMUTABLE` here?

contrib/babelfishpg_tsql/sql/sys_functions.sql
```diff
@@ -1,22 +1,123 @@
...
+    rec ANYELEMENT,
+    mode int,
+    element_name text,
+    binary_base64 boolean,
+    root_name text
+) RETURNS TEXT
+AS 'babelfishpg_tsql', 'tsql_query_to_xml_sfunc'
+LANGUAGE C COST 100;
+
+CREATE OR REPLACE FUNCTION sys.tsql_query_to_xml_ffunc(
```
2jungkook, line 13: `tsql_query_to_xml_ffunc` and `tsql_query_to_xml_text_ffunc` are almost duplicate. Can we use `return tsql_query_to_xml_text_ffunc(...)::XML` in this function body?

contrib/babelfishpg_tsql/src/backend_parser/gram-tsql-epilogue.y.c
```diff
@@ -1369,40 +1219,85 @@ returning_list_has_column_name(List *existing_colnames, char *current_colname)
...
+	 * Finally make function call to tsql_select_for_xml_agg or tsql_select_for_xml_text_agg
+	 * depending on the return_xml_type flag (TYPE option in the FOR XML clause).
+	 * The only difference of the two functions is the return type. tsql_select_for_xml_agg
+	 * returns XML type, tsql_select_for_xml_text_agg returns text type.
+	 */
+	if (return_xml_type)
+		func_name= list_make2(makeString("sys"), makeString("tsql_select_for_xml_agg"));
+	else
+		func_name= list_make2(makeString("sys"), makeString("tsql_select_for_xml_text_agg"));
+	func_args = list_make5(makeColumnRef(construct_unique_index_name("rows", "forjson"), NIL, -1, NULL),
```
2jungkook: 1. `forjson` seems not correct
2. I understand that you are trying to make a unique column name, but using `construct_unique_index_name()` could give a wrong impression to both developers and users. `construct_unique_index_name()` is doing many additional things and may throw an error like "constructing unique index name failed".
3. I would suggest to create a separate function for this purpose.

contrib/babelfishpg_tsql/src/backend_parser/gram-tsql-rule.y
```diff
@@ -1361,155 +1361,119 @@ tsql_UpdateStmt: opt_with_clause UPDATE relation_expr_opt_alias
...
+			| select_clause sort_clause tsql_for_xml_clause
 				{
+					insertSelectOptions((SelectStmt *) $1, $2, NIL,
+										NULL, NULL,
+										yyscanner);
 					if ($3 == NULL)
-						insertSelectOptions((SelectStmt *) $1, $2, NIL,
-											NULL, NULL,
-											yyscanner);
+						$$ = $1;
```
2jungkook, line 1380: In which case, this line could happen?

contrib/babelfishpg_tsql/src/tsql_for/forjson.c
```diff
@@ -0,0 +1,222 @@
...
+	PG_RETURN_TEXT_P(cstring_to_text_with_len(state->data, state->len));
+}
+
+static void 
+tsql_row_to_json(StringInfo state, Datum record, bool include_null_values)
+{
+	HeapTupleHeader td;
+	Oid				tupType;
+	int32			tupTypmod;
+	TupleDesc		tupdesc;
```
2jungkook, line 130: We may need `ReleaseTupleDesc()`.

contrib/babelfishpg_tsql/src/tsql_for/forxml.c
```diff
@@ -0,0 +1,329 @@
...
+		nspoid = get_namespace_oid("sys", true);
+		Assert(nspoid != InvalidOid);
+
+		tsql_datatype_oid = GetSysCacheOid2(TYPENAMENSP, Anum_pg_type_oid, CStringGetDatum(typename), ObjectIdGetDatum(nspoid));
+	
+		/*
+		 * tsql_datatype_oid can be different from datatype_oid when there are datatypes in different namespaces
+		 * but with the same name. Examples: bigint, int, etc.
+		 */
+		if (tsql_datatype_oid == datatype_oid)
```
2jungkook: What about `bit` and `money` types? Do they behave differently than `for json`?

contrib/babelfishpg_tsql/src/tsql_for/forxml.c
```diff
@@ -0,0 +1,329 @@
...
```
2jungkook: This check logic can be shared at least between `tsql_row_to_xml_path()` and `tsql_row_to_xml_raw()`.

contrib/babelfishpg_tsql/src/tsql_for/forxml_old.c
```diff
@@ -45,6 +52,10 @@ tsql_query_to_xml(PG_FUNCTION_ARGS)
 	StringInfo result = tsql_query_to_xml_internal(query, mode,
 											element_name, binary_base64, root_name);
 
+	ereport(WARNING,
+			(errcode(ERRCODE_WARNING_DEPRECATED_FEATURE),
+			 errmsg("This version of FOR XML has been deprecated. We recommend recreating the view for this query.")));
```
2jungkook, line 57: There is no test for this. You can either
1. modify our JDBC test framework to test the warning message, or
2. use installcheck

contrib/babelfishpg_tsql/src/tsql_for/tsql_for.c
```diff
@@ -0,0 +1,73 @@
...
+
+#include "tsql_for.h"
+
+/*
+ * This function handles the format for datetime datatypes by converting the output
+ * into required format for SELECT FOR JSON/XML. For example:
+ * "2022-11-11 20:56:22.41" -> "2022-11-11T20:56:22.41" for datetime, datetime2 & smalldatetime
+ */
+void
+tsql_for_datetime_format(StringInfo format_output, char *outputstr)
```
2jungkook: Let's try to use `const char *` if we don't expect the input parameter be changed.

test/JDBC/expected/forjson-after-14-6-vu-prepare.out
```diff
@@ -0,0 +1,139 @@
+
+-- 14.6 (aka extension version 2.4.0) has a major change to how FOR JSON
```
2jungkook: 1. BBF 2.4 corresponds to PG 14.7
2. Can we use `before-14-x` like other tests?

contrib/babelfishpg_tsql/src/backend_parser/gram-tsql-rule.y
Sairakan: to be honest, I am not sure. This is that pattern that was previously used, which is why I kept it here, but based on the grammar rules there doesn't seem to be a way to actually match tsql_for_xml_clause while generating a null value. I'll try removing it and see what happens.

contrib/babelfishpg_tsql/src/tsql_for/forxml.c
Sairakan: Yeah, bit and money have a different behavior compared with `for json` - they are not coerced to boolean/numeric types and are left as-is with `for xml`

contrib/babelfishpg_tsql/sql/sys_functions.sql
Sairakan: This is what was used in the original function for `for xml` so I kept it here. I don't think it is actually necessary though

contrib/babelfishpg_tsql/sql/sys_functions.sql
Sairakan: Any tests that return 0 rows will trigger the STRICT keyword. I believe there are a few such test cases.
It's probably okay to use IMMUTABLE, let me think about it a bit more.

contrib/babelfishpg_tsql/sql/sys_functions.sql
Sairakan: Good point, that should work and reduce code duplication.

contrib/babelfishpg_tsql/src/backend_parser/gram-tsql-epilogue.y.c
Sairakan: 1. I can change the name to something else, but it does need to be the same for both `for json` and `for xml` since they share the same `TsqlForClauseSubselect` usage. I'll probably use something like `tsql_for`.
2. `construct-unique-index-name()` is overall a pretty simple function, it just does an md5 hash of the first argument and then appends the second argument to the end - the only case when the error would be raised is if the md5 itself fails, which is extremely unlikely and would basically only happen if the system hits OOM per the comments in the md5 function. I don't think it is worth creating a new function for this case, but I'll add some comments explaining why this is used in the first place.

contrib/babelfishpg_tsql/src/tsql_for/tsql_for.c
Sairakan: Will do

contrib/babelfishpg_tsql/src/tsql_for/forjson.c
Sairakan: Good point, added

test/JDBC/expected/forjson-after-14-6-vu-prepare.out
Sairakan: fixed the comment to refer to 14.7, and renamed the tests to before-14-7 and normal tests

contrib/babelfishpg_tsql/src/tsql_for/forxml.c
Sairakan: Extracted the shared code into a separate function.

contrib/babelfishpg_tsql/sql/sys_functions.sql
Sairakan: I've updated the sfuncs to STABLE and ffuncs to IMMUTABLE

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.3.0--2.4.0.sql
```diff
@@ -157,6 +157,97 @@ CREATE OR REPLACE FUNCTION sys.degrees(IN arg1 TINYINT)
...
 GRANT EXECUTE ON FUNCTION sys.degrees(TINYINT) TO PUBLIC;
 
+-- deprecate old FOR XML/JSON functions
+ALTER FUNCTION sys.tsql_query_to_xml(text, int, text, boolean, text) RENAME TO tsql_query_to_xml_deprecated_in_2_4_0;
+CALL sys.babelfish_drop_deprecated_object('function', 'sys', 'tsql_query_to_xml_deprecated_in_2_4_0');
+
+ALTER FUNCTION sys.tsql_query_to_xml_text(text, int, text, boolean, text) RENAME TO tsql_query_to_xml_text_deprecated_in_2_4_0;
+CALL sys.babelfish_drop_deprecated_object('function', 'sys', 'tsql_query_to_xml_text_deprecated_in_2_4_0');
+
+ALTER FUNCTION sys.tsql_query_to_json_text(text, int, boolean, boolean, text) RENAME TO tsql_query_to_json_text_deprecated_in_2_4_0;
```
2jungkook, line 167: Please be careful when you make the same change for `3_X_DEV`. In the `3.0.0--3.1.0` upgrade script, `sys.tsql_query_to_json_text` may or may not exist because of two different upgrade paths:
1. 2.3 --> 3.0 --> 3.1: `sys.tsql_query_to_json_text` exists
2. 2.4 --> 3.0 --> 3.1: `sys.tsql_query_to_json_text` does not exists

test/JDBC/upgrade/14_5/schedule
```diff
@@ -36,6 +36,7 @@ TestUDD
 TestUniqueIdentifier
 TestVarChar
 TestXML
+forxml-before-14-7
```
2jungkook: Let's rename this to `forxml-before-14-7-or-15-2` to have the same test names between 2_X_DEV and 3_X_DEV.

test/JDBC/upgrade/latest/schedule
```diff
@@ -139,8 +139,9 @@ BABEL-3646
 BABEL-383
 BABEL-405
 BABEL-937
-BABEL-3407
-forjson
+forjson-before-14-7
```
2jungkook: Why do we have before-14-7 in here?

Sairakan: hm, rethinking this, the name is not properly descriptive of what the tests are - these are supposed to be the regression tests that apply to both previous versions as well as any future versions. The normal forjson and forxml test suites then are only testing the new features added in 14.7. I think it would be best to go back to the original naming scheme, where these are the normal tests (aka forjson/forxml), and the other test files are named `after-14-7`

Sairakan: after discussion, renaming this to forjson/forxml, and the other tests to forjson-subquery/forxml-subquery

2jungkook (dismissed):
Please add the engine PR link to the PR description

2jungkook (approved):

PR #677, opened 2022-10-07
Support GRANT/REVOKE CONNECT TO/FROM user

shalinilohia50:
1. Create guest user on any user created database by default and disable the guest user.
2. Support enabling/disabling of any user on user created databases.
3. Added a column to the catalog table "sys.babelfish_authid_user_ext".
4. The column "user_can_connect" will track if any user has the connect permission to a database.
5. The value of "user_can_connect" will be 1 if the user has permission and on a database otherwise 0.
6. Upgrade Scenario: When the database is created on an old instance and after the instance is upgraded to the Narwhal, it should create guest user on all the database and disable it by default.
7. User 'guest' should not be allowed to be dropped in any database.

Task: BABEL-2678, BABEL-3522


contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.2.0--2.3.0.sql
```diff
@@ -1859,6 +1859,17 @@ LANGUAGE plpgsql;
 CALL sys.babelfish_drop_deprecated_view('sys', 'check_constraints_deprecated_in_2_3_0');
 CALL sys.babelfish_drop_deprecated_view('sys', 'default_constraints_deprecated_in_2_3_0');
 
+-- USER extension
+ALTER TABLE sys.babelfish_authid_user_ext add COLUMN IF NOT EXISTS user_can_connect INT NOT NULL DEFAULT 1;
+
+GRANT SELECT ON sys.babelfish_authid_user_ext TO PUBLIC;
+
+CREATE OR REPLACE PROCEDURE sys.babel_update_user_catalog_for_guest()
```
macwoody: for internal functions, let's use babelfish_ prefix, (please also double check with others as well)

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -2651,6 +2663,37 @@ exec_stmt_usedb_explain(PLtsql_execstate *estate, PLtsql_stmt_usedb *stmt, bool
...
 }
 
+static int
+exec_stmt_grantdb(PLtsql_execstate *estate, PLtsql_stmt_grantdb *stmt)
+{
+	char 	*dbname = get_cur_db_name();
+	char	*dbowner = get_owner_of_db(dbname);
+	char	*login = GetUserNameFromId(GetSessionUserId(), false);	
+	ListCell *lc;
+	if (strcmp(login, dbowner) != 0)
```
xhfanhe: Please add a check for sysadmin and sysadmin members

xhfanhe:
Please add metadata integrity check for this catalog change. You can file it as another Jira but it will need to be merged within this cycle

contrib/babelfishpg_tsql/src/pl_exec-2.c
shalinilohia50: Added a check for the condition where the login is a member of "sysadmin" and relevant tests.

shalinilohia50:
> Please add metadata integrity check for this catalog change. You can file it as another Jira but it will need to be merged within this cycle

Sure. Filed a JIRA for the same and marked it for Narwhal.

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -1995,3 +2009,275 @@ get_catalog_info(Rule *rule)
...
+	table_close(bbf_authid_user_ext_rel, RowExclusiveLock);
+
+	return role_exists;
+}
+
+static char
+*get_db_owner_role_name(char *dbname)
+{
+	char *name = palloc0(MAX_BBF_NAMEDATALEND);
+	snprintf(name, MAX_BBF_NAMEDATALEND, "%s_db_owner", dbname);
```
macwoody: I think this is wrong, under single db mode, there is not prefix.

macwoody: This behavior is slightly misaligned with users (I think all users were prefixed) for historical reason (introduced before user )

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -668,8 +670,18 @@ exec_stmt_exec(PLtsql_execstate *estate, PLtsql_stmt_exec *stmt)
 
 	if (stmt->is_cross_db)
 	{
+		char *login = GetUserNameFromId(GetSessionUserId(), false);
+		char *user = get_user_for_database(stmt->db_name);
+
 	 	estate->db_name = stmt->db_name;
```
macwoody, line 676: I wonder if there is only a single value here. 
What if we cross multiple DBs?

test/JDBC/input/BABEL_GRANT_CONNECT-vu-prepare.sql
```diff
@@ -0,0 +1,12 @@
...
+
+use grant_connect_db1;
+go
+
+create table grant_connect_t1(a int);
+go
+
+create login grant_connect_abc with password = 'Babel123'
+go
+
```
xhfanhe, line 12: Do we need to switch back to master db at the end of each test? I think the question is more about, will JDBC start a new session for each test script, both for normal JDBC and during upgrade?

shalinilohia50: No, we don't need to switch back to master at the end of the test. 
Yes, JDBC starts a new session for each test script.

contrib/babelfishpg_tsql/src/catalog.c
shalinilohia50: Yes, that was causing some issue with the upgrade. 
Fixed it and used the function 'get_db_owner_name' which handles both single db and multi db.

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.2.0--2.3.0.sql
shalinilohia50: Fixed it in the latest commit.

contrib/babelfishpg_tsql/src/pl_exec-2.c
shalinilohia50: If we have a cross db procedure call like "exec db1.dbo.p1" which has a cross db statement like "select * from db2.dbo.t2", the cross db procedure call will be executed in "exec_stmt_exec" and the cross db select statement will be taken care of in "exec_stmt_execsql". Each statement will be executed separately with its db name.

.github/composite-actions/run-pg-upgrade/action.yml
```diff
@@ -32,11 +32,13 @@ runs:
         echo 'Updating babelfish extensions...'
         cd ~/work/babelfish_extensions/babelfish_extensions/
         ~/${{ inputs.pg_new_dir }}/bin/pg_ctl -D ~/${{ inputs.pg_new_dir }}/data -l ~/${{ inputs.pg_new_dir }}/data/logfile14 start
+        echo 'Set the migration_mode before running the upgrade scripts..'
+        sudo ~/${{ inputs.pg_new_dir }}/bin/psql -d jdbc_testdb -U runner -c "ALTER SYSTEM SET babelfishpg_tsql.migration_mode = '${{inputs.migration_mode}}';"
+        sudo ~/${{ inputs.pg_new_dir }}/bin/psql -d jdbc_testdb -U runner -c "SELECT pg_reload_conf();"
```
xhfanhe, line 37: Why we need to do this?

shalinilohia50: In case of major version upgrade, after pg_upgrade the migration mode was set to default which is hard-coded as single-db and later we update the extensions and set the migration mode as it was before the MVU. 
The upgrade script change in this PR expects the original migration mode at the time of executing "update_catalog_for_guest" function, which becomes untrue with the previous implementation. 
That's the reason I have set the migration mode before running the upgrade scripts.

xhfanhe (approved):

PR #270, opened 2022-06-14
Computed column ISNULL(CONVERT()) crashes PG server with segfault

thephantomthief:
### Description

In PG, the typmod of a CoalesceExpr is -1 if all the arguments of the
expression are unable to agree on type and typmod. This is not the case
for T-SQL ISNULL() (which is implemented using PG COALESCE() in
Babelfish and so uses CoalesceExpr internally). In case of T-SQL
ISNULL(), the type/typmod is determined by the first not null argument.

This commit introduces a hook namely "coalesce_typmod_hook" which is
called if we encounter a T-SQL ISNULL() function. This hook returns the
typmod of the first argument in a CoalesceExpr. If the coalesce type is
numeric, then instead of sending typmod of the first argument (which is
not entirely correct since we may not be able to fit other arguments in
that scale/precision), we use the "resolve_numeric_typmod_from_exp"
method to figure out the scale/precision for us.

Signed-off-by: Sharu Goel <goelshar@amazon.com>

### Upgrade tests added

For any upgrades with base version prior to latest, to any version afterwards, will not have these changes reflected. So such upgrade tests will be having an error returned recommending workarounds. For upgrades with base version latest, these changes will be reflected.

### Issues Resolved

#117 

### Check List
- [x] Commits are signed per the DCO using --signoff 

By submitting this pull request, I confirm that my contribution is under the terms of the Apache 2.0 and PostgreSQL licenses, and grant any person obtaining a copy of the contribution permission to relicense all or a portion of my contribution to the PostgreSQL License solely to contribute all or a portion of my contribution to the PostgreSQL open source project.

For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/babelfish-for-postgresql/babelfish_extensions/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #684, opened 2022-10-11
Support nondeterministic collations for LIKE in CHECK constraint clause

AvantikaDG:
Utility statements such as ALTER TABLE are not planned. Due to collation like to ilike transformation logic being called in a planner hook, CHECK constraint in utlity commands do not support LIKE for nondeterministic collations.

To solve this problem, we add a hook to invoke the like to ilike transformation function after the raw parsetree is transformed into an executable expression.

Task: BABEL-3319

Engine PR: https://github.com/babelfish-for-postgresql/postgresql_modified_for_babelfish/pull/51


test/JDBC/expected/BABEL-CHECK-CONSTRAINT-vu-prepare.out
```diff
@@ -0,0 +1,9 @@
+-- BABEL-3319: Checking that alter table check constraint does not raise collation error
```
Deepesh125, line 2: should we also add upgrade tests to show the behaviour difference before latest version (2.3.0) and after the latest version? 

Also, should we extend existing verify script to show expected behaviour as well? For example,
```
create table foo (a varchar(100) collate sys.database_default);

insert few entries

alter table.. add constraint..

insert few entries which would violate this new constraint 
```

Deepesh125 (changes requested):

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -240,6 +244,9 @@ InstallExtendedHooks(void)
 
 	prev_print_pltsql_function_arguments_hook = print_pltsql_function_arguments_hook;
 	print_pltsql_function_arguments_hook = print_pltsql_function_arguments;
+
+	prev_transform_check_constraint_expr_hook = transform_check_constraint_expr_hook;
+	transform_check_constraint_expr_hook = transform_like_in_add_constraint;
```
2jungkook, line 277: For parity, please add a similar logic for `UninstallExtendedHooks()`.

test/JDBC/expected/BABEL-CHECK-CONSTRAINT-vu-prepare.out
2jungkook: \+ creation-time CHECK constraint and case-insensitive insertions.

2jungkook: To test different behavior before the latest, you can refer https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/270/files

.github/workflows/jdbc-tests.yml
```diff
@@ -48,7 +48,7 @@ jobs:
           sudo ~/postgres/bin/psql -d postgres -U runner -v user="jdbc_user" -v db="jdbc_testdb" -f .github/scripts/cleanup_babelfish_database.sql
 
       - name: Upload Log
-        if: always() && steps.jdbc.outcome == 'failure'
+        if: always() && (steps.jdbc.outcome == 'failure' || steps.install-extensions.outcome == 'failure')
```
2jungkook: Let's remove this once debugging is done.

contrib/babelfishpg_common/src/collation.c
```diff
@@ -1124,6 +1123,29 @@ bool collation_is_CI_AS(Oid colloid)
 	return false;
 }
 
+bool expr_has_ilike_node_and_ci_as_coll(Node *expr, bool check_for_ci_as_collation)
+{
+	int i = 0;
```
2jungkook: nit; we can move this into the for loop.

contrib/babelfishpg_common/src/collation.c
```diff
@@ -1124,6 +1123,29 @@ bool collation_is_CI_AS(Oid colloid)
...
+				sizeof(like_ilike_table[i].ilike_op_name)) == 0)
+		{
+			ilike_node_found = true;
+			break;
+		}
+	}
+	if (!check_for_ci_as_collation)
+		return ilike_node_found;
+	
+	return (ilike_node_found & collation_is_CI_AS(op->inputcollid));
```
2jungkook: nit; is there a reason to use a bitwise operator?

contrib/babelfishpg_common/src/collation.c
```diff
@@ -1124,6 +1123,29 @@ bool collation_is_CI_AS(Oid colloid)
 	return false;
 }
 
+bool expr_has_ilike_node_and_ci_as_coll(Node *expr, bool check_for_ci_as_collation)
```
2jungkook: function name could be misleading; `expr_has_ilike_node_and_ci_as_coll()` returns true even if expr doesn't have ci_as_coll. I don't have a clear suggestion, but can think of some alternatives:
- use two different functions, `has_ilike()` and `is_ci_as_coll()`
- use a more generic name such as `expr_needs_ilike_transform()`

contrib/babelfishpg_tsql/src/collation.c
```diff
@@ -123,6 +124,23 @@ make_or_qual(Node *qual1, Node *qual2)
 	return (Node *) make_orclause(list_make2(qual1, qual2));
 }
 
+static void
+init_and_check_collation_callbacks(void)
```
2jungkook: Let's use a declaration instead of moving it.

contrib/babelfishpg_tsql/src/collation.c
```diff
@@ -249,14 +267,46 @@ transform_funcexpr(Node* node)
  *		 col LIKE PATTERN -> col ILIKE PATTERN
  */
 static Node*
-transform_likenode(Node* node)
+transform_likenode(Node* node, Oid opno)
```
2jungkook: `opno` is not needed

contrib/babelfishpg_tsql/src/collation.c
```diff
@@ -675,3 +708,36 @@ tsql_find_collation_internal(const char *collation_name)
 	return (*collation_callbacks_ptr->find_collation_internal)(collation_name);
 }
 
+bool
+expr_contains_ilike_and_ci_collation_wrapper(Node *expr, bool check_for_ci_as_collation)
```
2jungkook: It seems like `check_for_ci_as_collation` is always true.

contrib/babelfishpg_tsql/src/collation.c
```diff
@@ -675,3 +708,36 @@ tsql_find_collation_internal(const char *collation_name)
...
+{
+	List 		*queue;
+	ListCell 	*lc = NULL;
+	bool 		expr_contains_ilike_and_ci_as = false;
+	
+	if(expr == NULL)
+		return false;
+	queue = NIL;
```
2jungkook: not needed?

contrib/babelfishpg_tsql/src/collation.c
```diff
@@ -249,14 +267,46 @@ transform_funcexpr(Node* node)
...
+						tsql_find_collation_internal(coll_info_of_inputcollid.collname));
+				op->inputcollid = tsql_get_oid_from_collidx(collidx_of_cs_as);
+			}
+			else
+			{
+				/* If a collation is not specified, use the default one */
+				op->inputcollid = DEFAULT_COLLATION_OID;
+			}
+
+			if (NOT_FOUND == collidx_of_cs_as)
```
2jungkook: You can move this into the above `if` block.

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -2987,3 +2994,22 @@ pltsql_planner_hook(Query *parse, const char *query_string, int cursorOptions, P
...
 	return plan;
 }
+
+static Node* 
+transform_like_in_add_constraint (Node* node)
+{
+	bool create_tbl_stmt_is_invalid = false;
+
+	if (current_query_is_create_tbl_check_constraint)
+		create_tbl_stmt_is_invalid = expr_contains_ilike_and_ci_collation_wrapper(node, true);
```
2jungkook: Why do we need a wrapper? Can't we just call the callback function?

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -2987,3 +2994,22 @@ pltsql_planner_hook(Query *parse, const char *query_string, int cursorOptions, P
...
+	if (!babelfish_dump_restore && current_query_is_create_tbl_check_constraint && create_tbl_stmt_is_invalid)
+	{
+		current_query_is_create_tbl_check_constraint = false;
+		ereport(ERROR,
+					(errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
+					errmsg("nondeterministic collations are not supported for ILIKE")));
+	}
+	current_query_is_create_tbl_check_constraint = false;
+	return pltsql_predicate_transformer(node);
+}
```
2jungkook, line 2988: suggestion
```
PG_TRY();
{
    if (!babelfish_dump_restore &&  current_query_is_create_tbl_check_constraint && expr_contains_ilike_and_ci_collation_wrapper(node, true))
    {
        throw an error
    }
}
PG_FINALLY();
{
    current_query_is_create_tbl_check_constraint = false;
}
return pltsql_predicate_transformer(node);
```

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -4796,3 +4802,33 @@ pltsql_revert_guc(int nest_level)
        PltsqlGUCNestLevel = nest_level - 1;
 
 }
+
+void
+check_current_query_is_create_tbl_check_constraint(Node *expr)
```
2jungkook: This is more like `set_current_query_is_create_tbl_check_constraint`.

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -4796,3 +4802,33 @@ pltsql_revert_guc(int nest_level)
...
+				break;
+			default:
+				break;
+		}
+		if (contains_check_constraint)
+		{
+			current_query_is_create_tbl_check_constraint = true;
+			break;
+		}
+	}
```
2jungkook, line 4841: The code is functional, but I'd suggest to polish a little bit so that it could be more succinct.

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -2987,3 +2994,22 @@ pltsql_planner_hook(Query *parse, const char *query_string, int cursorOptions, P
...
+	if (current_query_is_create_tbl_check_constraint)
+		create_tbl_stmt_is_invalid = expr_contains_ilike_and_ci_collation_wrapper(node, true);
+
+	if (!babelfish_dump_restore && current_query_is_create_tbl_check_constraint && create_tbl_stmt_is_invalid)
+	{
+		current_query_is_create_tbl_check_constraint = false;
+		ereport(ERROR,
+					(errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
+					errmsg("nondeterministic collations are not supported for ILIKE")));
+	}
```
2jungkook, line 2980: Do we have a test case for this?

contrib/babelfishpg_tsql/src/collation.c
```diff
@@ -675,3 +708,36 @@ tsql_find_collation_internal(const char *collation_name)
...
+			init_and_check_collation_callbacks();
+			expr_contains_ilike_and_ci_as = (*collation_callbacks_ptr->expr_contains_ilike_and_ci_as_coll)(predicate, true);
+			if (expr_contains_ilike_and_ci_as)
+				break;
+		}
+		else if (IsA(predicate, BoolExpr))
+		{
+			BoolExpr   *boolexpr = (BoolExpr *) predicate;			
+			queue = list_concat(queue, boolexpr->args);
+		}
```
2jungkook, line 741: Do we have a test case for this?

contrib/babelfishpg_tsql/src/collation.c
```diff
@@ -249,14 +267,46 @@ transform_funcexpr(Node* node)
...
+				collidx_of_cs_as =
+					tsql_find_cs_as_collation_internal(
+						tsql_find_collation_internal(coll_info_of_inputcollid.collname));
+				op->inputcollid = tsql_get_oid_from_collidx(collidx_of_cs_as);
+			}
+			else
+			{
+				/* If a collation is not specified, use the default one */
+				op->inputcollid = DEFAULT_COLLATION_OID;
+			}
```
2jungkook, line 291: Do we have a test case for `else` block?

contrib/babelfishpg_common/src/collation.c
AvantikaDG: We can use && and take advantage of short-circuiting here. Changing.

contrib/babelfishpg_tsql/src/hooks.c
AvantikaDG: I made a wrapper so we don't have to expose `init_and_check_collation_callbacks()`

contrib/babelfishpg_tsql/src/collation.c
AvantikaDG: Removed

contrib/babelfishpg_tsql/src/hooks.c
AvantikaDG: It is also being done for other collation callbacks (refer to [collation.h](https://github.com/babelfish-for-postgresql/babelfish_extensions/blob/BABEL_2_X_DEV/contrib/babelfishpg_tsql/src/collation.h#L76))

contrib/babelfishpg_tsql/src/hooks.c
AvantikaDG: We don't have a test case for this because this check is for psql syntax. ILIKE is not supported in tsql syntax. I have done manual testing for this in psql.

contrib/babelfishpg_tsql/src/collation.c
AvantikaDG: We don't have a test case for this because this check is for psql syntax (similar to the test case that you asked about in hooks.c). TSQL syntax will not trigger this code block at all.

I have done manual testing for this in psql.

contrib/babelfishpg_tsql/src/collation.c
AvantikaDG: I can write a test case to exercise this code, but the function will return false. It will return true only for psql syntax. 

However, I tried a simple test and found a bug in the LIKE to ILIKE transformation logic. The following table definition:
```
CREATE TABLE tbl3(
     a char(20) COLLATE bbf_unicode_cp1_ci_as,
     CHECK (NOT a LIKE '11%' AND a LIKE 'ab%')
)

```
is pg dumped as:
```
CREATE TABLE master_dbo.tbl3 (
    a sys.bpchar(20) COLLATE sys.bbf_unicode_cp1_ci_as,
    CONSTRAINT tbl3_a_check CHECK (((NOT ((a)::text ~~ '11%'::text)) AND (((a)::text ~~* 'ab%'::text) AND (((a)::text >= 'ab'::text) AND ((a)::text < ('ab'::text || '￿'::text COLLATE sys.bbf_unicode_cp1_ci_as))))))
)
```
(note that the `a LIKE '11%'` expression has **not** been converted into an ILIKE expression.

If it is okay with you, then I would like to file a separate bug for this and work on it separately from this PR since it has not come up yet. Adding the test will be pending this new fix.

2jungkook: okay, makes sense.

contrib/babelfishpg_tsql/src/collation.c
AvantikaDG: Added tests in .mix files

contrib/babelfishpg_tsql/src/hooks.c
AvantikaDG: Added tests in .mix files

contrib/babelfishpg_common/src/collation.c
```diff
@@ -1124,6 +1123,28 @@ bool collation_is_CI_AS(Oid colloid)
 	return false;
 }
 
+bool has_valid_coll(Node *expr, bool check_for_ci_as_collation)
```
Deepesh125: It's worth adding comment on what this function is doing. Also, I would suggest to rename it to reflect that this function checks for ilike operator.

Deepesh125 (dismissed):

.github/workflows/major-version-upgrade.yml
```diff
@@ -165,3 +165,13 @@ jobs:
...
+
+      - name: CI
+      on: [push]
+      jobs:
+        build:
+          runs-on: ubuntu-latest
+          steps:
+          - uses: actions/checkout@v2
+          - name: Setup upterm session
+            uses: lhotari/action-upterm@v1
```
faviansamatha: Not 100% sure if it would work since we're creating another job and not a step (might create a new environment or may cause an error, not too sure). We would also need an if condition to always run this even if the above failed `if: always()` Try replacing this with this: 

```
- name: Setup upterm session
        if: always()
        uses: lhotari/action-upterm@v1
```
(from lines 51-53 https://github.com/Bit-Quill/babelfish_extensions/blob/da22b8df7c62e18a536aba48c897d053f5345034/.github/workflows/jdbc-tests.yml which is from this commit when I was experimenting with it: https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/811/files/da22b8df7c62e18a536aba48c897d053f5345034)

AvantikaDG: Thank you! Updating it.

2jungkook (approved):

PR #685, opened 2022-10-11
Handle the error while sending error message to client in tds

sumitj824:

### Description

Earlier It was possible that we fail while trying to send error to client (for example, because of encoding conversion failure), since the error has occurred during error processing the code flow will not return to the original error's reporter or handler, in that case tds_disable_error_log_hook will never get reset to false and error of current/further query will not get send to client. So, we place a PG_TRY block so that we can log the internal error and tds_disable_error_log_hook can be reset to false so that further errors can be sent to client.
 
### Issues Resolved

BABEL-3501

### Tests scenarios covered
Since the query for which the issue has been found is get fixed with BABEL-2636. So I have done the manual testing by disabling BABEL-2636.

### Check List


contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
```diff
@@ -2196,11 +2196,30 @@ TdsSendError(int number, int state, int class,
...
+		TdsSendInfoOrError(TDS_TOKEN_ERROR, number, state, class,
 						  message,
 						  "BABELFISH",	/* TODO: where to get this? */
 						  "",		/* TODO: where to get this? */
 						  lineNo);
+	}
+	PG_CATCH();
+	{
+		/* Send message to client that internal error occurred */
+		TdsSendInfoOrError(TDS_TOKEN_ERROR, number, state, class,
```
KushaalShroff: Use ERRCODE_INTERNAL_ERROR rather than same error-number(error code)

sumitj824: addressed

contrib/babelfishpg_tds/src/backend/tds/err_handler.c
```diff
@@ -288,38 +288,60 @@ emit_tds_log(ErrorData *edata)
...
+			TdsSendDone(TDS_TOKEN_DONE, TDS_DONE_ERROR, 0, 0);
+			TdsFlush();
+		}
+	}
+	PG_CATCH();
+	{
+		/* Log the internal error message */
+		ErrorData *next_edata;
+		char *message;
+		next_edata = CopyErrorData();
```
surendravishnoi, line 338: why to copy? we are not keeping it?

sumitj824: From inner try block when we rethrow the error it will get catch by outer try block till then the second error cycle is not started

Deepesh125 (approved):

surendravishnoi (approved):

PR #705, opened 2022-10-17
Add test scenarios to PR template

AvantikaDG:
Add test scenarios to PR template to include different test cases. This will enforce high quality commits.

Task: BABEL-3626


nasbyj (dismissed):

jsudrik (commented):
Thanks for adding this. 

Can we please leave one line gap under every title so that it becomes more obvious that these are the sections to be filled?

Will the github actions fail the attempt to merge if the template is not present?

PR #708, opened 2022-10-18
Improve sp_sproc_columns performance

faviansamatha:
### Description

Previously, sp_sproc_columns had major performance issue due to some convoluted joins and it did not display table-valued parameters. This commit focuses on fixing those issues by reworking the query and allowing sp_sproc_columns to detect Table-valued parameters (TVP). This commit also adds some minor improvements such as:
- Showing correct COLUMN_TYPE value for output parameters in stored procedures, 
- Corrected the data type fields for table-valued-functions
- Displays some missing return parameters from procedures
- Fix nullable field to correspond with official documentaiton 
- Fix datatypes being returned (some were using `int` instead of `sys.int` and `varchar` instead of `sys.varchar` in the internal view)

NOTE: Any previous limitations from this proc is still present, the main focus is on performance and adding TVP support. 

Task: BABEL-3555, BABEL-3595

### Issues Resolved

- BABEL-3555
- BABEL-3595

### Test Scenarios Covered

**Use case based** - Yes
**Boundary conditions** - Yes
**Arbitrary inputs** - Yes
**Negative test cases** - Yes
**Minor version upgrade tests** - Yes
**Major version upgrade tests** - Yes
**Performance tests** - Yes manually tested performance. Running the previous version of sp_sproc_columns with the current test locally took ~2 minutes and 36 seconds where as the current implementation  runs between 3.3-5 seconds. Also added a test where 1300 similar procedures were added (wiith 25 parameters and various system types).  Previous implementation took 20 minutes and 39 seconds, and the new implementation took about 12 seconds to run `exec sp_sproc_columns`. 

**Tooling impact** - n/a
**Client tests** - n/a

### Check List


test/JDBC/input/BABEL-SP_SPROC_COLUMNS.sql
```diff
@@ -1,255 +0,0 @@
-DROP FUNCTION IF EXISTS net
```
faviansamatha, line 1: Removed this file (along with the `.out` version in the `expected` folder) because the three version upgrade tests:
- BABEL-SP_SPROC_COLUMNS-vu-prepare.sql
- BABEL-SP_SPROC_COLUMNS-vu-verify.sql
- BABEL-SP_SPROC_COLUMNS-vu-cleanup.sql

covers all these tests (plus the newly added one). Keeping this file in would cause the JDBC tests to run this test twice.

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.2.0--2.3.0.sql
```diff
@@ -2577,6 +2577,259 @@ AS 'babelfishpg_tsql', 'update_user_catalog_for_guest';
...
+    AS sys.varchar(254)) AS IS_NULLABLE
+  , CAST(
+  	  CASE
+  	    WHEN st.is_table_type = 1 THEN 0
+  	    ELSE sdit.ss_data_type
+  	  END
+  	AS sys.tinyint) AS SS_DATA_TYPE
+  , CAST(ss.proname AS sys.sysname) AS original_procedure_name
+  FROM pg_type t
+  INNER JOIN sys.types st ON st.user_type_id = t.oid
```
macwoody: sys.types is based on pg_type, why do we join pg_type again?

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.2.0--2.3.0.sql
```diff
@@ -2577,6 +2577,259 @@ AS 'babelfishpg_tsql', 'update_user_catalog_for_guest';
...
+  LEFT JOIN sys.spt_datatype_info_table sdit ON sdit.type_name = (SELECT name FROM sys.types WHERE user_type_id = st.system_type_id)
```
macwoody: Could you please elaborate more on the subquery? 
why join condition is not sdit.type_name = st.name

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.2.0--2.3.0.sql
```diff
@@ -2577,6 +2577,259 @@ AS 'babelfishpg_tsql', 'update_user_catalog_for_guest';
...
+  INNER JOIN
+  (
+    SELECT
+      p.oid AS p_oid,
+      p.proname,
```
macwoody: this subquery project more columns than needed

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.2.0--2.3.0.sql
```diff
@@ -2577,6 +2577,259 @@ AS 'babelfishpg_tsql', 'update_user_catalog_for_guest';
...
+    WHERE (
+      p.pronamespace in (select schema_id from sys.schemas union all select oid from pg_namespace where nspname = 'sys')
+      AND (pg_has_role(p.proowner, 'USAGE') OR has_function_privilege(p.oid, 'EXECUTE'))
+      AND ( p.probin like '{%typmod_array%}'-- typmod array exists for those that are user-defined
+      		OR p.proname like 'sp\_%' -- filter out internal babelfish-specific procs
+      		OR p.proname like 'xp\_%'
+      		OR p.proname like 'dm\_%'
+      		OR p.proname like 'fn\_%')) 
+  ) ss ON t.oid = (ss.x).x
+WHERE ( ss.proargmodes[(ss.x).n] in ('i', 'o', 'b')  OR ss.proargmodes[(ss.x).n] is NULL)
```
macwoody: why not put this where clause inside of the above subquery?

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.2.0--2.3.0.sql
```diff
@@ -2577,6 +2577,259 @@ AS 'babelfishpg_tsql', 'update_user_catalog_for_guest';
...
+
+UNION
+-- Select return types. This needs to be unionized because inline-table functions may
+-- end up with 2 't' outputs which causes previous query to miss these items
```
macwoody: I wonder if we could UNION input and output types in one subquery and avoiding copying similar joins and case expressions

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.2.0--2.3.0.sql
faviansamatha: Yes, so we need to join that to get the ODBC data type information for user-defined types. Currently, spt_datatype_info_table only have the primitive/system-defined types and are all hardcoded. Doing `sdit.type_name = st.name` would not work since spt_datatype_info_table doesn't track user-defined types since it isn't recorded so I did this condition so we can find what type the user type was based on and use that information.

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.2.0--2.3.0.sql
faviansamatha: Ack, don't think I need pg_type at all. Will remove that

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.2.0--2.3.0.sql
faviansamatha: I'll give that a shot and try it out.

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.2.0--2.3.0.sql
faviansamatha: My bad, will remove.

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.2.0--2.3.0.sql
faviansamatha: Not sure if that's possible (correct me if I'm wrong though) since proargmodes is an array and there would be no way of indexing it until after the subquery. After the the subquery, we can use the `.n` in the expanded array to index it which is why it's after.

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.2.0--2.3.0.sql
faviansamatha: Was able to do it, please see the newest revision.

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.2.0--2.3.0.sql
faviansamatha: Added comments to the code to explain it as well

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.2.0--2.3.0.sql
```diff
@@ -2577,6 +2577,254 @@ AS 'babelfishpg_tsql', 'update_user_catalog_for_guest';
...
+    FROM bbf_proc p
+  ) ss
+  INNER JOIN pg_catalog.pg_namespace ns ON ns.oid = ss.pronamespace
+  LEFT JOIN sys.babelfish_namespace_ext bne ON ns.nspname = bne.nspname 
+  LEFT JOIN sys.databases d ON d.database_id = bne.dbid
+  LEFT JOIN sys.types st ON ((ss.x IS NOT NULL AND st.user_type_id = (ss.x).x) OR (ss.x IS NULL AND st.user_type_id = ss.prorettype))
+  -- Because spt_datatype_info_table does contain user-defind types and their names,
+  -- the join below allows us to retrieve the system type name on what the user-defined type was based on.
+  LEFT JOIN sys.spt_datatype_info_table sdit ON sdit.type_name = (SELECT name FROM sys.types WHERE user_type_id = st.system_type_id)
+WHERE (ss.proargmodes[(ss.x).n] in ('i', 'o', 'b')  OR ss.proargmodes[(ss.x).n] is NULL);
```
macwoody: I feel that the first disjunction term could be moved to line 2802 and the second one could be removed completely

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.2.0--2.3.0.sql
```diff
@@ -2577,6 +2577,254 @@ AS 'babelfishpg_tsql', 'update_user_catalog_for_guest';
...
+      p.proargmodes as proargmodes,
+      p.prokind as prokind,
+      p.proretset as proretset,
+      p.prorettype as prorettype,
+      p.pronamespace as pronamespace,
+      NULL AS x -- null value indicates that we are retrieving the return values of the proc/func
+    FROM bbf_proc p
+  ) ss
+  INNER JOIN pg_catalog.pg_namespace ns ON ns.oid = ss.pronamespace
+  LEFT JOIN sys.babelfish_namespace_ext bne ON ns.nspname = bne.nspname
```
macwoody: Unless we want to show objects out side of BBF, I think LEFT JOIN in 2017/2018 could be changed into INNER JOIN

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.2.0--2.3.0.sql
```diff
@@ -2577,6 +2577,254 @@ AS 'babelfishpg_tsql', 'update_user_catalog_for_guest';
...
+      p.proretset as proretset,
+      p.prorettype as prorettype,
+      p.pronamespace as pronamespace,
+      NULL AS x -- null value indicates that we are retrieving the return values of the proc/func
+    FROM bbf_proc p
+  ) ss
+  INNER JOIN pg_catalog.pg_namespace ns ON ns.oid = ss.pronamespace
+  LEFT JOIN sys.babelfish_namespace_ext bne ON ns.nspname = bne.nspname 
+  LEFT JOIN sys.databases d ON d.database_id = bne.dbid
+  LEFT JOIN sys.types st ON ((ss.x IS NOT NULL AND st.user_type_id = (ss.x).x) OR (ss.x IS NULL AND st.user_type_id = ss.prorettype))
```
macwoody: in which case do we expect a type doesn't exist in sys.types ? 
In addition, we don't have to use this complex join condition, we could simply project x column as (ss.x).x or ss.prorettype. complext join condition will result in nested loop join.

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.2.0--2.3.0.sql
```diff
@@ -2577,6 +2577,254 @@ AS 'babelfishpg_tsql', 'update_user_catalog_for_guest';
...
+  -- Because spt_datatype_info_table does contain user-defind types and their names,
+  -- the join below allows us to retrieve the system type name on what the user-defined type was based on.
+  LEFT JOIN sys.spt_datatype_info_table sdit ON sdit.type_name = (SELECT name FROM sys.types WHERE user_type_id = st.system_type_id)
```
macwoody: Sorry that I am still confused with this join and the use of subquery in join condition. Could you please give me an example of this?

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.2.0--2.3.0.sql
faviansamatha: 1. `ss.proargmodes[(ss.x).n] in ('i', 'o', 'b')`
I don't think this one can be moved inside 2802 because we can only get the `.n` value from `(ss.x)` after the projection. Before the projection from the inner query, we only have one row per function/procedures since the inner queries only search for procs. 

2. `ss.proargmodes[(ss.x).n] is NULL` 
This was a mistake and was actually supposed to be two cases (`ss.proargmodes is NULL OR ss.x is NULL`). I accidentally left this in and it seemed to have worked out equivalently to those (sheer luck) but will change that in the next revision. We do need this though because
- `ss.proargmodes is NULL` is needed since in a procedure, if all parameters are inputs, it will be NULL (defined in the pg_proc documentation). 
- `ss.x is NULL` we need this because of the return_values query in the inner query (2805-2813). 

For 2, I added comments for each condition. Please check them out when you have the time.

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.2.0--2.3.0.sql
faviansamatha: All objects that are displayed should be from BBF. This is handled in the WHERE clause of the CTE (2776 - 2783). The CTE (bbf_proc) should only show BBF objects because it joins `sys.schema` unioned with `sys` database. 

For 2817, and 2818, we need these LEFT JOINs because `sys.babelfish_namespace_ext` don't include the `sys` schema and since the `sys` schema isn't attached to a BBF database, we need to also left join sys.databases. Doing the INNER joins would remove all the system-defined functions and procedures in the `sys` schema. Any NULL values for PROCEDURE_QUALIFIER and PROCEDURE_OWNER should be coalesced with `sys.db_name()`, and `sys` respectively (which was what was done in the old code).

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.2.0--2.3.0.sql
faviansamatha: the return types of table-valued-functions will not be in sys.types AND procedures (procedures are voids, and table-valued functions would be `record` or a cusstom table type depending if it's inline or not). 

I'm not sure if I can't project column x as (ss.x).x since I do need the `n` value that comes with it, but I'll try to see what i can do with `ss.prorettype`.

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.2.0--2.3.0.sql
faviansamatha: So I need sys.spt_datatype_info_table because it contains the ODBC information about the system datatypes. However, this table is hardcoded so any user-defined types would have this information. Furthermore, this table only contains the data-type names (no OID). 

So if I create a type like this: ` CREATE TYPE my_udt FROM INT`   I get a newly generated user_type_id (unique to this type) and a system_type_id (which corresponds to INT) in `sys.types`. 

If I try the suggested join condition `sdit.type_name = st.name`. This wouldn't work because `my_udt` does not exist in `sys.spt_datatype_info_table.type_name` (since we only have the names of the system types in that table). 

The only way to get the ODBC information about the user-defined type is to find the name of the datatype in `sys.types` that has `my_udt`'s `system_type_id` as their `user_type_id`.  In order to find it, I need to look for the `name` in `sys.types` in which the `user_type_id` matches the `system_type_id` of `my_udt` and then join it with the `name` in `spt_datatype_info_table` (in other words, trying to get the base type name of `my_udt` which is `int`)

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.2.0--2.3.0.sql
faviansamatha: I updated the code to project x and n in different columns so the complex join is removed. (marking `n` as NULL to signify that we are retrieving the OUTPUT of the proc/function). Please check the latest code

shalinilohia50:
Could you please add the below test template in the PR description?
* **Use case based -**
* **Boundary conditions -**
* **Arbitrary inputs -**
* **Negative test cases -**
* **Performance tests -**
* **Tooling impact -**
* **Client tests -**

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.2.0--2.3.0.sql
```diff
@@ -2577,6 +2577,266 @@ AS 'babelfishpg_tsql', 'update_user_catalog_for_guest';
...
+      p.proallargtypes as proallargtypes,
+      p.proargtypes as proargtypes,
+      s.name as schema_name
+    FROM 
+      pg_proc p
+    INNER JOIN (
+      SELECT name as name, schema_id as id  FROM sys.schemas 
+      UNION ALL 
+      SELECT CAST(nspname as sys.sysname) as name, CAST(oid as int) as id 
+        from pg_namespace WHERE nspname in ('sys', 'information_schema_tsql')
```
macwoody: I guess in this context, we still want to show "information_schema" instead of "information_schema_tsql"

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.2.0--2.3.0.sql
```diff
@@ -2577,6 +2577,266 @@ AS 'babelfishpg_tsql', 'update_user_catalog_for_guest';
...
+    ) as s ON p.pronamespace = s.id
+    WHERE (
+    p.pronamespace in (select schema_id from sys.schemas union all select oid from pg_namespace where nspname in ('sys', 'information_schema'))
```
macwoody: you won't need this filter condition any more , since you are joining them

faviansamatha: Ack, removed!

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.2.0--2.3.0.sql
faviansamatha: Sounds good, replaced it with `information_schema`

macwoody (approved):

PR #713, opened 2022-10-20
Fix T-SQL coercion hash table

rishabhtanwar29:
### Description
Commit fe0a0a7 implemented a fix to load datatype coercion hash table
on demand along with a logic to keep initializing the hash table on every
lookup until all the hash table entries are initialized completely.
But later it was discovered that some of the entries of static array (which we
use to fill the hash table with) are not correctly written so they never get
filled into the hash table and as a result we keep initializing the hash table 
on every lookup.

This commit addresses the issue by fixing all incorrect entries in the static
array as well as adds additional logic for searching a cast function.

### Testing
* Added an internal function `babelfish_integrity_checker`
  which can be used to perform integrity checks.
* Use case based: 
* Boundary conditions: N/A
* Arbitrary inputs: N/A
* Negative test scenarios: N/A
* Minor/Major version upgrade tests: added a test case in BABEL-637 to
  check if hash tables are fully initialized, this test runs after both major
  and minor version upgrade.
* Performance tests: None
* Tooling impact: N/A
* Client tests: N/A

Task: BABEL-3632

### Check List


contrib/babelfishpg_tsql/src/pltsql_coerce.c
```diff
@@ -538,6 +532,10 @@ init_tsql_coerce_hash_tab(PG_FUNCTION_ARGS)
 	/* mark the hash table initialised */
 	inited_ht_tsql_cast_info = true;
 
+    argTypes = (Oid *) palloc(3 * sizeof(Oid));
```
surendravishnoi, line 540: can you please explain why is this done?

contrib/babelfishpg_tsql/sql/sys_function_helpers.sql
```diff
@@ -10310,3 +10310,7 @@ AS 'babelfishpg_tsql', 'get_server_collation_oid';
 CREATE OR REPLACE FUNCTION sys.babelfish_get_pltsql_function_signature(IN funcoid OID)
 RETURNS text
 AS 'babelfishpg_tsql', 'get_pltsql_function_signature' LANGUAGE C;
+
+CREATE OR REPLACE FUNCTION sys.babelfish_datatype_hash_tables_initialised()
```
surendravishnoi: creating a SQL function with upgrade implications just for testing is not good, we should try to look for other alternatives

contrib/babelfishpg_tsql/src/pltsql_coerce.c
rishabhtanwar29: I just added a comment above explaining the reason

contrib/babelfishpg_tsql/sql/sys_function_helpers.sql
rishabhtanwar29: I think we already have a couple of functions for internal debugging/testing purpose, [ref](https://github.com/babelfish-for-postgresql/babelfish_extensions/blob/BABEL_2_X_DEV/contrib/babelfishpg_tsql/sql/sys_function_helpers.sql#L10293), moreover why do we think there are upgrade implications of this?

test/python/expected/upgrade_validation/expected_dependency.out
```diff
@@ -123,6 +123,7 @@ Function sys.babelfish_conv_time_to_string(text,text,time without time zone,nume
 Function sys.babelfish_conv_to_varchar(text,anyelement,numeric)
 Function sys.babelfish_conv_to_varchar(text,text,numeric)
 Function sys.babelfish_cursor_list(integer)
+Function sys.babelfish_datatype_hash_tables_initialised()
```
2jungkook: Instead of adding a test in BABEL-637, please consider creating an upgrade test with some dependencies, e.g., views.

contrib/babelfishpg_tsql/sql/sys_function_helpers.sql
2jungkook: I agree with Surendra. Once we expose SQL and C functions, it is almost impossible to remove them. One suggestion would be the following:

First of all, we can remove the SQL function in the installation script and create this on-demand in the test file and remove it.

For the C function, I'd suggest to expose a more generic function, e.g., babelfish_integrity_checker, and then inside that we can add some specific functions freely.

contrib/babelfishpg_tsql/src/pltsql_coerce.c
```diff
@@ -1087,6 +1100,17 @@ init_tsql_datatype_precedence_hash_tab(PG_FUNCTION_ARGS)
...
 }
 
+/*
+ * Returns true if both the datatype related hash tables
+ * are fully initialised, false otherwise.
+ */
+Datum
+datatype_hash_tables_initialised(PG_FUNCTION_ARGS)
+{
+    bool inited = inited_ht_tsql_cast_info & inited_ht_tsql_datatype_precedence_info;
```
2jungkook: Given that this is a test-purpose, I'd prefer to throw an error with a specific error message, e.g., "tsql cast info hash table is not initialized".

contrib/babelfishpg_tsql/src/pltsql_coerce.c
```diff
@@ -587,6 +591,15 @@ init_tsql_coerce_hash_tab(PG_FUNCTION_ARGS)
...
                     if (!OidIsValid(entry->castfunc))
+                    {
+                        /* also search cast function with 3 input arguments */
+                        argTypes[0] = castsource;
+                        entry->castfunc = GetSysCacheOid3(PROCNAMEARGSNSP, Anum_pg_proc_oid,
+                                    CStringGetDatum(tsql_cast_raw_infos[i].castfunc),
+                                    PointerGetDatum(buildoidvector(argTypes, 3)),
+                                    ObjectIdGetDatum(sys_nspoid));
+                    }
+                    if (!OidIsValid(entry->castfunc))
```
2jungkook: You can move this if-block into the above if-block because when `entry->castfunc` is valid this if-check is redundant.

contrib/babelfishpg_tsql/sql/sys_function_helpers.sql
rishabhtanwar29: Yeah it also makes sense to create the function on demand, addressed it in new commit!

test/python/expected/upgrade_validation/expected_dependency.out
rishabhtanwar29: Yup, added an upgrade test but it can only run in latest since the underlying function is not supported in previous releases.

contrib/babelfishpg_tsql/src/pltsql_coerce.c
rishabhtanwar29: Done!

contrib/babelfishpg_tsql/src/pltsql_coerce.c
rishabhtanwar29: Done!

2jungkook (approved):

PR #728, opened 2022-10-25
Improve identity value insert performance with IDENTITY_INSERT=on

kuntalghosh:
### Description
With IDENTITY_INSERT=on in a table, users are allowed to insert values in identity columns. We also need to update the last used identity for the table(seed) so that subsequent insertion with default identity can start from the same seed. As per the current approach, after inserting one row, we're doing a full scan of the table to find out the min/max identity value and updating these variables. This creates a huge performance bottleneck for data insertion and the performance degrades significantly as the number of rows increases in the table.

In T-SQL, the last used identity of a table is non-transactional. That means, it always increases (or decreases) irrespective of the underlying transaction commits/rollbacks. To implement this, we don't have to scan the entire table. We can just calculate the max(current seed, value to be inserted) and update the seed under the same lock.

Task: BABEL-3506
Signed-off-by: Kuntal Ghosh [kuntalgh@amazon.com](mailto:kuntalgh@amazon.com)
Engine PR: https://github.com/babelfish-for-postgresql/postgresql_modified_for_babelfish/pull/55

### Functional Testing
We already have a significant amount of tests for identity.
https://github.com/babelfish-for-postgresql/babelfish_extensions/blob/BABEL_2_X_DEV/test/JDBC/input/BABEL-IDENTITY.sql
https://github.com/babelfish-for-postgresql/babelfish_extensions/blob/BABEL_2_X_DEV/test/JDBC/input/BABEL-IDENTITY-COLUMN.sql
https://github.com/babelfish-for-postgresql/babelfish_extensions/blob/BABEL_2_X_DEV/test/JDBC/input/BABEL-IDENTITY-BIFS.sql
https://github.com/babelfish-for-postgresql/babelfish_extensions/blob/BABEL_2_X_DEV/test/JDBC/input/BABEL-SEQUENCE.sql
We just modified the transactional aspect of identity, i.e., the seed value changes will not be rolled back. I've added the test for the same.

### Upgrade
This commit removes two internal SQL functions that is no longer needed. Ideally, these functions should not be used by the end-users directly. As a precaution, we're not going to drop the functions as part of the upgrade process, instead we'll rename those as deprecated.
These changes only affect whether the sequence seed update should be transactional or not in TSQL dialect. As per the existing implementation,
From TDS connection, if you insert an identity column, the seed update is transactional. That means if the insert rollbacks, the seed will not be updated with the new value. This is not the expected T-SQL behaviour.
Internally, this is implemented as 'GENERATED ALWAYS AS IDENTITY'. From PG connection, you're not allowed to insert directly in an identity column. It'll throw an error.
But, there are ways to insert identity value from PG connection as well - a) using OVERRIDING SYSTEM VALUE with INSERT b) using COPY. But, both these options don't update the seed by default. User has to manually update the seed using setval function. In fact, pg_dump uses the same approach for dumping & restoring identity columns followed by an explicit setval function call and set the max value from that identity column.
In this patch, there will be no behavioural changes in PG connection (as a result pg_dump as well). Only from TDS connection, the seed update will become non-transactional similar to T-SQL and PG.

### Performance Testing
CREATE TABLE t1 (a int identity, b int);
Scenario: Test insert with IDENTITY_INSERT ON
Number of inserts: 100000
System: Local desktop, Default PG settings

Without these changes, it takes 300+ seconds time.
With these changes, it takes 27 seconds.


contrib/babelfishpg_tsql/sql/sys_function_helpers.sql
```diff
@@ -23,30 +23,6 @@ $BODY$
 $BODY$
 LANGUAGE SQL STABLE;
 
-CREATE OR REPLACE FUNCTION sys.get_min_id_from_table(IN id_colname TEXT,
```
2jungkook, line 26: Let's add upgrade script changes as well, e.g.,
- rename it to deprecated ones
- try to drop functions

contrib/babelfishpg_tsql/src/pl_exec.c
```diff
@@ -5458,71 +5445,67 @@ pltsql_update_identity_insert_sequence(PLtsql_expr *expr)
...
 					/* Find by name since other attributes not defined */
 					if (strcmp(NameStr(attr->attname), id_attname) == 0)
 					{
-						int64 last_identity = DatumGetInt64(SPI_getbinval(tuple,
-																		  tupdesc_ret,
-																		  attnum+1,
-																		  &isnull));
+						int tup_idx;
+						int64 last_identity;
+						int64 min_identity = INT_MAX;
```
2jungkook: INT_MAX is for int32?

contrib/babelfishpg_tsql/src/pl_exec.c
```diff
@@ -5458,71 +5445,67 @@ pltsql_update_identity_insert_sequence(PLtsql_expr *expr)
...
+
+						for (tup_idx = 0; tup_idx < n_processed; tup_idx++)
+						{
+							bool isnull;
+							HeapTuple tuple = tuptable->vals[tup_idx];
+
+							last_identity = DatumGetInt64(SPI_getbinval(tuple,
+																		tupdesc_ret,
+																		attnum+1,
+																		&isnull));
```
2jungkook, line 5474: nit; can we add an assertion for `isnull`?

contrib/babelfishpg_tsql/src/pl_exec.c
```diff
@@ -5458,71 +5445,67 @@ pltsql_update_identity_insert_sequence(PLtsql_expr *expr)
...
-			* case by default
-			*/
-			if (seq_incr < 0)
-				setval_query = "SELECT setval(pg_get_serial_sequence($1, $2), sys.get_min_id_from_table($2, $3, $4));";
-			else
-				setval_query = "SELECT setval(pg_get_serial_sequence($1, $2), sys.get_max_id_from_table($2, $3, $4));";
+							if (strcmp(defel->defname, "increment") == 0)
+								do_setval_tsql(seqid, max_identity, true);
+							else
+								do_setval_tsql(seqid, min_identity, true);
```
2jungkook: What would be other `defname` values? Is it correct to handle all other cases with `min_identity`?

test/JDBC/sql_expected/BABEL-IDENTITY.out
```diff
@@ -423,24 +423,115 @@ int#!#int
...
+~~START~~
+numeric
+300
+~~END~~
 
 
-SELECT sys.get_max_id_from_table('test_id', ';drop table master_dbo', 'test_table1;');
-GO
+-- Check Statement error shouldn't increase/decrease identity
+BEGIN TRAN t1; INSERT INTO dbo.t1_identity_1 (a,b) VALUES (400,2); ROLLBACK TRAN t1;
```
2jungkook: 1. "should" increase/decrease identity?
2. It would be better to have an error case here instead of a rollback.

test/JDBC/sql_expected/BABEL-IDENTITY.out
```diff
@@ -423,24 +423,115 @@ int#!#int
...
+CREATE TABLE dbo.t1_identity_1(a int identity primary key, b int);
+SET IDENTITY_INSERT dbo.t1_identity_1 ON;
+INSERT INTO dbo.t1_identity_1 (a,b) VALUES (1,1);
+go
+~~ROW COUNT: 1~~
+
+
+-- Check transaction rollback should increase/decrease identity
+BEGIN TRAN t1; INSERT INTO dbo.t1_identity_1 (a,b) VALUES (300,2); ROLLBACK TRAN t1;
+SELECT @@IDENTITY; SELECT IDENT_CURRENT('dbo.t1_identity_1'); SELECT SCOPE_IDENTITY();
```
2jungkook, line 437: Can we have an upgrade test for this? Even after mVU/MVU, do we expect the same value, e.g., 300?

contrib/babelfishpg_tsql/src/pl_exec.c
kuntalghosh: Good catch

contrib/babelfishpg_tsql/src/pl_exec.c
kuntalghosh: This is actually wrong. I should check the increment.

contrib/babelfishpg_tsql/sql/sys_function_helpers.sql
kuntalghosh: I'll rename it to deprecated ones.

test/JDBC/sql_expected/BABEL-IDENTITY.out
kuntalghosh: For IDENT_CURRENT(), we expect the same value after mVU/MVU. It fetches the last used value for the sequence from the PG catalog. 
Rest of the two functions are session specific. We already have existing upgrade tests for IDENT_CURRENT(). Do you think we need to test some other scenario?
One scenario could be all previous releases rollbacks the sequence value when transaction rollbacks, whereas after an upgrade to this version, the value is not rolled back.

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.2.0--2.3.0.sql
```diff
@@ -3170,3 +3170,7 @@ BEGIN
   RETURN 0;
 END;
 $$;
+
+-- Identity related helper functions are no longer needed
+ALTER FUNCTION sys.get_min_id_from_table RENAME TO get_min_id_from_table_deprecated_in_2_3_0;
+ALTER FUNCTION sys.get_max_id_from_table RENAME TO get_max_id_from_table_deprecated_in_2_3_0;
```
2jungkook, line 3536: Can we also call `babelfish_drop_deprecated_object()`?

contrib/babelfishpg_tsql/src/pltsql_identity.c
```diff
@@ -308,3 +308,39 @@ void pltsql_resetcache_identity()
...
+{
+	if (sql_dialect == SQL_DIALECT_TSQL)
+	{
+		ListCell *seq_lc;
+		List *seq_options;
+		int64 seq_incr = 0;
+
+		seq_options = sequence_options(seqid);
+
+		foreach (seq_lc, seq_options)
```
2jungkook, line 330: Can we pass `pgsform->seqincrement` so that we don't need to iterate seq_options?

contrib/babelfishpg_tsql/src/pltsql_identity.c
```diff
@@ -308,3 +308,39 @@ void pltsql_resetcache_identity()
...
+		{
+			DefElem *defel = (DefElem *) lfirst(seq_lc);
+
+			if (strcmp(defel->defname, "increment") == 0)
+				seq_incr = defGetInt64(defel);
+
+		}
+
+		if (seq_incr > 0)
+			val = val > last_val ? val : last_val;
```
2jungkook, line 340: Do we have a test case where last_val is used?

test/JDBC/sql_expected/BABEL-IDENTITY.out
2jungkook: Thanks for checking. In that case, I don't think we need further upgrade tests.

contrib/babelfishpg_tsql/src/pltsql_identity.c
kuntalghosh: If we want to pass this value from the engine, we need to add two more lines in the engine side. We need to store the value in a variable before releasing the catalog tuple. That makes the engine changes a bit clumsy. So, I removed that code and decided to retrieve the value in the hook function itself.

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.2.0--2.3.0.sql
kuntalghosh: Ooh, I wasn't aware about this procedure. Nice!!

contrib/babelfishpg_tsql/src/pltsql_identity.c
2jungkook: Yeah, that makes sense.

2jungkook (approved):

PR #784, opened 2022-11-02
Bulk Load Request Optimisation in TDS Layer

KushaalShroff:
For every row and every column we were creating a StringInfo and calling InitStringInfo which does palloc internally, this showed up as one of the bottlenecks in the perf analysis.
This StringInfo held the binary data of the messageData received from client which was already present as a complete stream of data in another StringInfo object. We can skip this InitStringInfo(palloc) by directly processing the binary data - offset the message stream and create respective datums upfront.

Task: BABEL-3621

Signed-off-by: Kushaal Shroff [kushaal@amazon.com](mailto:kushaal@amazon.com)

## Functional Testing
Dotnet framework has elaborate tests that cover INSERT BULK AND BCP for each and every datatype, identity columns, keep nulls and other insert bulk options. One can check the BCP test plan we had created during the initial feature implementation.

## Performance testing
We achieved around 1.5x speedup with this change during our local testing. (We have not collected elaborate numbers since we will be clubbing few fixes together and do the perf analysis)
One can also check the flame graphs for before and after the change (Internal only)

### Check List

### DIFF
--- a/contrib/babelfishpg_tds/src/backend/tds/tdsbulkload.c
+++ b/contrib/babelfishpg_tds/src/backend/tds/tdsbulkload.c
@@ -754,7 +754,7 @@ SetBulkLoadRowData(TDSRequestBulkLoad request, StringInfo message)
 				case TDS_TYPE_XML:
 				{
 					ParameterToken token = palloc0(sizeof(ParameterTokenData));
-					// offset and batchsize is increased via ReadBcpPlp
+
 					retStatus = ReadBcpPlp(token, &message, request);
 					CheckPLPStatusNotOK(request, retStatus, i);
 					if (token->isNull) /* null */
--- a/contrib/babelfishpg_tds/src/backend/tds/tdsbulkload.c
+++ b/contrib/babelfishpg_tds/src/backend/tds/tdsbulkload.c
@@ -59,10 +59,10 @@ do \
 } while(0)
 
 /* For checking the invalid length in the request. */
-#define CheckForInvalidLength(rowData, temp, colNum) \
+#define CheckForInvalidLength(len, temp, colNum) \
 do \
 { \
-	if ((uint32_t)rowData->columnValues[i].len > (uint32_t)temp->colMetaData[i].maxLen) \
+	if ((uint32_t)len > (uint32_t)temp->colMetaData[i].maxLen) \
 		ereport(ERROR, \
 				(errcode(ERRCODE_PROTOCOL_VIOLATION), \
 				errmsg("The incoming tabular data stream (TDS) Bulk Load Request (BulkLoadBCP) protocol stream is incorrect. " \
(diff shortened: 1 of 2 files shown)


contrib/babelfishpg_tds/src/backend/tds/tdsbulkload.c
```diff
@@ -465,6 +465,8 @@ SetBulkLoadRowData(TDSRequestBulkLoad request, StringInfo message)
 {
 	BulkLoadColMetaData *colmetadata = request->colMetaData;
 	int retStatus = 0;
+	uint32_t len;
+	StringInfo temp = palloc0(sizeof(StringInfoData));
```
kuntalghosh, line 469: Where are we freeing this memory?

contrib/babelfishpg_tds/src/backend/tds/tdsbulkload.c
```diff
@@ -599,34 +651,53 @@ SetBulkLoadRowData(TDSRequestBulkLoad request, StringInfo message)
...
-							temp->isNull = false;
+							token->isNull = false;
 							continue;
 						}
 
-						plpStr = TdsGetPlpStringInfoBufferFromToken(message->data, temp);
-						rowData->columnValues[i] = *plpStr;
-						pfree(plpStr);
-						pfree(temp);
+						temp = TdsGetPlpStringInfoBufferFromToken(message->data, token);
```
kuntalghosh, line 669: Shouldn't you use a different pointer for this? Before this line, temp was pointing to another set of StringInfo memory that you allocated at the beginning of the function.

contrib/babelfishpg_tds/src/backend/tds/tdsbulkload.c
```diff
@@ -642,73 +713,95 @@ SetBulkLoadRowData(TDSRequestBulkLoad request, StringInfo message)
...
 						i++;
-						temp->isNull = false;
+						token->isNull = false;
 						continue;
 					}
 
-					plpStr = TdsGetPlpStringInfoBufferFromToken(message->data, temp);
-					rowData->columnValues[i] = *plpStr;
-					pfree(plpStr);
+					temp = TdsGetPlpStringInfoBufferFromToken(message->data, token);
```
kuntalghosh, line 774: Here again we're overwriting the pointer. It's better to define temp variable in each switch case where you need the same.

surendravishnoi (dismissed):

contrib/babelfishpg_tds/src/backend/tds/tdsbulkload.c
KushaalShroff: I have free'd the temp pointer just before reassigning it

contrib/babelfishpg_tds/src/backend/tds/tdsbulkload.c
KushaalShroff: If we allocate stringInfo inside the switch case which happens in a loop for all rows' columns', it would defeat the purpose of the optimisation done in this PR. Rather just freed temp just before it

kuntalghosh (approved):

PR #787, opened 2022-11-02
Fix psqlodbc test framework `double` equality comparison

faviansamatha:
### Description
Fix psqlodbc test framework `double` equality comparison

Previously, the template psqlodbc_common functions utilized the regular `=` operator to check for two values of type double. This can cause issues due to floating point inaccuarcy. This commit mitigates this issue by creating a helper function that compares two doubles by taking the difference of the two values and comparing it to the machine's epsilon value.

Task: BABELFISH-560

### Check List

### DIFF
--- a/test/odbc/psqlodbc/psqlodbc_tests_common.cpp
+++ b/test/odbc/psqlodbc/psqlodbc_tests_common.cpp
@@ -761,3 +761,11 @@ void formatNumericExpected(vector<string> &vec, const int &scale, const bool &is
     vec[i] = formatNumericWithScale(vec[i], scale, is_bbf);
   }
 }
+
+void compareDoubleEquality(double actual, double expected) {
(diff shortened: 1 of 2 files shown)


ColinKYuen:
Would this be a problem for `float` and `numeric` as well? If so, we would need to add them as well.

faviansamatha:
> Would this be a problem for `float` and `numeric` as well? If so, we would need to add them as well.

No, it's recommended that `float` and `numeric` map to char (https://github.com/MicrosoftDocs/sql-docs/blob/live/docs/odbc/reference/appendixes/converting-data-from-c-to-sql-data-types.md) so it wouldn't affect these datatypes.

ColinKYuen (approved):

leszekamz (approved):

PR #788, opened 2022-11-03
Support datetime2 cast and rounding in datetime2 TRY_CONVERT

Jakeowen1:
### Description

Currently try_convert to datetime2 cannot handle certain cases. This commit changes the implementation to use cast, and also extends the datetime2 implementation to cover try_cast. It reverts the changes of the previous commit to rely on cast rather than the datetime_to_string function
 
Task: Babel-3614

### Check List


Sairakan (approved):

contrib/babelfishpg_tsql/sql/sys_function_helpers.sql
```diff
@@ -1639,10 +1639,6 @@ BEGIN
             v_day := substr(v_datestring, 7, 2);
             v_month := substr(v_datestring, 5, 2);
             v_year := substr(v_datestring, 1, 4);
-        ELSE
```
timchang514, line 1642: Do we need to replace this ELSE statement with something? Like a default value?

test/JDBC/expected/BABEL-3614-vu-verify.out
```diff
@@ -202,8 +204,9 @@ SELECT BABEL_3614_vu_prepare_f7()
 GO
 ~~START~~
 datetime2
-<NULL>
-~~END~~
```
timchang514, line 206: Just want to make sure, is this the same behavior as SQL Server - do they return NULL or throw an error in this case?

timchang514: For this case and all the other ones like it in this file*

contrib/babelfishpg_tsql/sql/sys_function_helpers.sql
Jakeowen1: This is how the function had it before I made the changes in my last pr so im just restoring it to the original version. If it makes sense to keep this change I can put it back

test/JDBC/expected/BABEL-3614-vu-verify.out
Jakeowen1: Yes this is how sql server handles it. If conversion to that type is prohibited it throws an error. If conversion to that type is allowed but the conversion cannot be completed it returns null. Will run through all of the test cases to confirm

Jakeowen1: Yeah so everything behaves the same aside from some of the rounding which just has to do with our difference in representation. Errors are thrown in the same cases though

AvantikaDG (approved):

AvantikaDG:
Let's change the PR title to something like "Support datetime2 cast and rounding in datetime2 TRY_CONVERT". You can mention in the description that you reverted the previous commit Also does this cover BABEL-3646?

Jakeowen1:
> Let's change the PR title to something like "Support datetime2 cast and rounding in datetime2 TRY_CONVERT". You can mention in the description that you reverted the previous commit Also does this cover BABEL-3646?

This commit does not cover babel-3646 because that involves more datatypes than just datetime2. This commit just fixes the datetime2 issues but also enables support for try_cast datetime2

PR #791, opened 2022-11-03
MVU fails in multi-db migration mode when there is one or more user databases

shalinilohia50:
Remove the upgrade dependency on the migration-mode GUC and get the db_owner rolename from the catalog.

Task: BABEL-3665

### DIFF
--- a/contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.2.0--2.3.0.sql
+++ b/contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.2.0--2.3.0.sql
@@ -3048,6 +3048,9 @@ AS 'babelfishpg_tsql', 'update_user_catalog_for_guest';
  
 CALL sys.babelfish_update_user_catalog_for_guest();
 
+-- Drop this procedure after it is executed once during upgrade.
+DROP PROCEDURE sys.babelfish_update_user_catalog_for_guest();
+
 ALTER VIEW sys.sp_sproc_columns_view RENAME TO sp_sproc_columns_view_deprecated_in_2_3_0;
 
 CREATE OR REPLACE VIEW sys.sp_sproc_columns_view
--- a/contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.2.0--2.3.0.sql
+++ b/contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.2.0--2.3.0.sql
@@ -3050,9 +3050,6 @@ AS 'babelfishpg_tsql', 'update_user_catalog_for_guest';
  
 CALL sys.babelfish_update_user_catalog_for_guest();
 
--- Drop this procedure after it gets executed once.
-DROP PROCEDURE sys.babelfish_update_user_catalog_for_guest();
-
 ALTER VIEW sys.sp_sproc_columns_view RENAME TO sp_sproc_columns_view_deprecated_in_2_3_0;
 
 CREATE OR REPLACE VIEW sys.sp_sproc_columns_view
@@ -3831,6 +3828,9 @@ CALL sys.babelfish_drop_deprecated_object('function', 'sys', 'get_max_id_from_ta
 -- Drops the temporary procedure used by the upgrade script.
 -- Please have this be one of the last statements executed in this upgrade script.
 DROP PROCEDURE sys.babelfish_drop_deprecated_object(varchar, varchar, varchar);
+
+-- Drop this procedure after it gets executed once.
+DROP PROCEDURE sys.babelfish_update_user_catalog_for_guest();
 
 -- Reset search_path to not affect any subsequent scripts
 SELECT set_config('search_path', trim(leading 'sys, ' from current_setting('search_path')), false);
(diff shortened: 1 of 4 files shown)


contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -2181,7 +2182,7 @@ static void
 create_guest_role_for_db(char *dbname)
 {
 	const char		*guest = get_guest_role_name(dbname);
```
rishabhtanwar29, line 2184: Does the guest name also depends on migration_mode? If so then we should also figure this out using catalog.

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -2265,3 +2266,42 @@ create_guest_role_for_db(char *dbname)
...
+				BTEqualStrategyNumber, F_TEXTEQ,
+				CStringGetTextDatum(dbname));
+
+	scan = table_beginscan_catalog(bbf_authid_user_ext_rel, 2, key);
+
+	tuple_user_ext = heap_getnext(scan, ForwardScanDirection);
+	if (HeapTupleIsValid(tuple_user_ext))
+		{
+			Form_authid_user_ext userform = (Form_authid_user_ext) GETSTRUCT(tuple_user_ext);
+			db_owner_role = NameStr(userform->rolname);
```
rishabhtanwar29: We should duplicate the rolname string here as the original pointer will become invalid after table close, something like:
```
db_owner_role = pstrdup(NameStr(userform->rolname));
```
FYI, don't forget to free it in `create_guest_role_for_db` function.

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.2.0--2.3.0.sql
```diff
@@ -3048,6 +3048,9 @@ AS 'babelfishpg_tsql', 'update_user_catalog_for_guest';
  
 CALL sys.babelfish_update_user_catalog_for_guest();
 
+-- Drop this procedure after it is executed once during upgrade.
```
rishabhtanwar29: Instead of this comment here we should rather add a comment before line no. 3045 to specify that it is a temporary procedure which serves XYZ functionality.

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -2265,3 +2266,42 @@ create_guest_role_for_db(char *dbname)
...
+ * migration mode GUC.
+ */
+static char *
+get_db_owner_role_name(char *dbname)
+{
+	Relation	bbf_authid_user_ext_rel;
+	HeapTuple	tuple_user_ext;
+	ScanKeyData		key[2];
+	TableScanDesc		scan;
+	char		*db_owner_role;
```
rishabhtanwar29: Let's initialize `db_owner_role` with NULL so that we don't return an uninitialized pointer if tuple_user_ext is somehow invalid.

contrib/babelfishpg_tsql/src/catalog.c
shalinilohia50: Talked to Di about it, the current implementation is fine.

contrib/babelfishpg_tsql/src/catalog.c
shalinilohia50: Resolved it in the next commit.

rishabhtanwar29 (dismissed):

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.2.0--2.3.0.sql
```diff
@@ -3042,12 +3042,17 @@ ALTER TABLE sys.babelfish_authid_user_ext add COLUMN IF NOT EXISTS user_can_conn
...
+-- This is a temporary procedure which is called during upgrade to create guest users
+-- for the user created databases if it doesn't have guest user already.
 CREATE OR REPLACE PROCEDURE sys.babelfish_update_user_catalog_for_guest()
 LANGUAGE C
 AS 'babelfishpg_tsql', 'update_user_catalog_for_guest';
  
 CALL sys.babelfish_update_user_catalog_for_guest();
 
+-- Drop this procedure after it gets executed once.
+DROP PROCEDURE sys.babelfish_update_user_catalog_for_guest();
```
xhfanhe: Please check with Jungkook/Surendra on whether we want to call drop stmt directly in upgrade script.

shalinilohia50: As discussed with Rishabh, it's fine since this proc is being created temporarily so there can't be any dependencies on it. Drop_deprecated proc is needed for the objects which might have dependent objects

xhfanhe: Then let's put this line at the end of the upgrade script, together with other DROP stmts.
Btw I'm seeing some function is added after resetting the search path, which I think is not right. It's not related to your change, but would you please check with Jungkook if we want to fix that as well within this commit?

xhfanhe: And please get approval from him as well, in the meantime please file a CR as well

xhfanhe (dismissed):

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.2.0--2.3.0.sql
shalinilohia50: Ack.

xhfanhe (approved):

rishabhtanwar29 (approved):

PR #803, opened 2022-11-07
Fix setval() behaviour in T-SQL mode

kuntalghosh:
We should not change the setval() behaviour in T-SQL dialect as some of the users may use this function as an alternative to DBCC CHECKIDENT for resetting the identity.

Task: BABEL-3506
Signed-off-by: Kuntal Ghosh <kuntalgh@amazon.com>

### Test plans
1. Test nextval(), setval() and currentval() behavior across different versions and upgrade path. It should be same irrespective of the dialect used.
2. Test with reseting the seed using setval and insert with identity_insert=off
3. Test if there's an error while using setval
4. PG regression tests already covers different possible inputs with setval. 

### Upgrade
The setval() behaviour should behave same way irrespective of upgrade.

### Performance
Doesn't affect the current performance


test/JDBC/expected/BABEL-IDENTITY-BIFS.out
```diff
@@ -869,6 +869,88 @@ numeric
 ~~END~~
 
 
+-- For DMS, we've suggested the following PG functions related to identity
+-- feature that can be called from TDS endpoint and they should get the exact
+-- same behaviour as PG endpoint.  So, let's add some tests.
+-- basic sequence operations for setval, nextval, currentval (tests are taken
+-- src/test/regress/sql/sequence.sql PG regression test suite)
+CREATE SEQUENCE sequence_test;
```
2jungkook: Just for future test coverage, could you make this as a part of upgrade test?

contrib/babelfishpg_tsql/src/pl_exec.c
```diff
@@ -5528,18 +5528,35 @@ pltsql_update_identity_insert_sequence(PLtsql_expr *expr)
...
+								Assert(0);
+							}
+							/* reset the value */
+							pltsql_setval_identity_mode = false;
+						}
+						PG_CATCH();
+						{
+							/* reset the value */
+							pltsql_setval_identity_mode = false;
+							PG_RE_THROW();
```
2jungkook: (minor) you can use `PG_FINALLY` here (https://github.com/babelfish-for-postgresql/postgresql_modified_for_babelfish/blob/BABEL_2_X_DEV__PG_14_X/src/include/utils/elog.h#L273-L287)

contrib/babelfishpg_tsql/src/pl_exec.c
```diff
@@ -5528,18 +5528,35 @@ pltsql_update_identity_insert_sequence(PLtsql_expr *expr)
...
+													ObjectIdGetDatum(seqid),
+													Int64GetDatum(min_identity));
+							else {
+								/* increment can't be zero */
+								Assert(0);
+							}
+							/* reset the value */
+							pltsql_setval_identity_mode = false;
+						}
+						PG_CATCH();
```
2jungkook: Just wondering whether we can have a test case for this catch block.

contrib/babelfishpg_tsql/src/pl_exec.c
```diff
@@ -5528,18 +5525,37 @@ pltsql_update_identity_insert_sequence(PLtsql_expr *expr)
...
+							else if (seq_incr < 0)
+								DirectFunctionCall2(setval_oid,
+													ObjectIdGetDatum(seqid),
+													Int64GetDatum(min_identity));
+							else {
+								/* increment can't be zero */
+								Assert(0);
+							}
+							/* reset the value */
+							pltsql_setval_identity_mode = false;
```
2jungkook: This is not needed because the cleanup code in `PG_FINALLY` will be run in either case.

test/JDBC/expected/BABEL-PG-SYSTEM-FUNCTIONS-vu-prepare.out
```diff
@@ -0,0 +1,78 @@
+-- For DMS, we've suggested the following PG functions related to identity
+-- feature that can be called from TDS endpoint and they should get the exact
+-- same behaviour as PG endpoint.  So, let's add some tests.
+-- basic sequence operations for setval, nextval, currentval (tests are taken
+-- src/test/regress/sql/sequence.sql PG regression test suite)
+CREATE SEQUENCE sequence_test;
```
2jungkook: Our object name convention is test_file_name + object, e.g., `BABEL_PG_SYSTEM_FUNCTIONS_vu_prepare_seq1`.

test/JDBC/expected/BABEL-PG-SYSTEM-FUNCTIONS.out
```diff
@@ -0,0 +1,81 @@
+-- For DMS, we've suggested the following PG functions related to identity
```
2jungkook: As the above MVU tests are executed in both normal and MVU scenario, we can remove this file.

test/JDBC/upgrade/13_4/schedule
```diff
@@ -187,3 +187,4 @@ BABEL-3010-before-14_6
 BABEL_GRANT_CONNECT
 BABEL-sp_helpdb
 BABEL-2795
+BABEL-PG-SYSTEM-FUNCTIONS
```
2jungkook, line 190: Please put this in other schedule files as well.

test/JDBC/expected/BABEL-PG-SYSTEM-FUNCTIONS-vu-prepare.out
kuntalghosh: Actually, I would like to maintain a single file for all such PG functions. Creating different files for different PG functions could be an overkill. That's why I kept the name as BABEL-PG-SYSTEM-FUNCTIONS-vu-<prepare/verify/cleanup>.

2jungkook (approved):

PR #804, opened 2022-11-08
Fix IS_MEMBER() to return correct result

raydhkim:
Previously, is_member() returned wrong result on ceratin cases, due to the design of Babelfish. As a part of the design, there is a indirect ownership relationship from db_owner and dbo_role, and this was not taken into account for implementation of is_member(), which cause the function to return wrong results in certain cases. Now, an additional check is included, so that the indirect ownership relationship is also taken into account and the function behaves correctly in those specific cases too.

Task: BABEL-3625

Signed-off-by: Ray Kim <raydhkim@amazon.com>

### Check List

### DIFF
--- a/test/JDBC/expected/BABEL-ROLE-MEMBER-vu-verify.out
+++ b/test/JDBC/expected/BABEL-ROLE-MEMBER-vu-verify.out
@@ -154,6 +154,69 @@ int
 0
 ~~END~~
 
+SELECT IS_ROLEMEMBER('BABEL_ROLE_MEMBER_vu_prepare_role1', 'db_owner')
+GO
+~~START~~
+int
+0
+~~END~~
+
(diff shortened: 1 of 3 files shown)


test/JDBC/expected/BABEL-ROLE-MEMBER-vu-verify.out
```diff
@@ -154,6 +154,27 @@ int
...
+0
+~~END~~
+
+SELECT IS_MEMBER('BABEL_ROLE_MEMBER_vu_prepare_role3')
+GO
+~~START~~
+int
+0
+~~END~~
+
```
xhfanhe, line 219: Please add several tests here for `IS_ROLEMEMBER` as well, where we specify the second argument to be `db_owner` or `dbo`

xhfanhe (approved):

PR #764, opened 2022-10-31
Fix the github issue : jdbc driver getTables return empty result

forestkeeper:
fix https://github.com/babelfish-for-postgresql/babelfish_extensions/issues/535
resolve the uncorrect result when execution sp_tables

Task: BABEL-3640

Signed-off-by: Zhibai Song  szh@amazon.com

### Description

previously  sp_tables returned wrong result when sp_type = '''TABLES''' with 3 single quota
fix this problem by modifying the sql script

### Issues Resolved

sp_tables returned wrong result when sp_type = '''TABLES''' with 3 single quota

### Check List
- [ ] Commits are signed per the DCO using --signoff 

By submitting this pull request, I confirm that my contribution is under the terms of the Apache 2.0 and PostgreSQL licenses, and grant any person obtaining a copy of the contribution permission to relicense all or a portion of my contribution to the PostgreSQL License solely to contribute all or a portion of my contribution to the PostgreSQL open source project.

For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/babelfish-for-postgresql/babelfish_extensions/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #818, opened 2022-11-09
Fix Error: "compressed lz4 data is corrupt" when dropping database

raydhkim:
Previously, when dropping a database, the error is raised with a message: "compressed lz4 data is corrupt", due to inaccurate comparison method used for Name type. Now, with correct comparison method for the specific type, the error is not thrown when database is being dropped.

Task: BABEL-3655

Signed-off-by: Ray Kim <raydhkim@amazon.com>

### Check List


xhfanhe (dismissed):

test/JDBC/expected/BABEL-3655.out
```diff
@@ -0,0 +1,46 @@
+-- tsql
```
2jungkook: Let's write an MVU test instead of a normal test (Ref: https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/764/files)

xhfanhe: Is there any particular reason that we want to have MVU test here? We are not altering any pre-defined BBF objects and thus don't need to test dependency. 
For functionality test I think normal JDBC would be enough?

2jungkook: (correction)
Usually, an MVU test covers both normal and MVU scenario, but in here, we need both a normal and an MVU test because of single/multi-db settings. More precisely,
- normal test: need to set `migration_mode` into `multi-db` at the beginning and reset it to `single-db` at the end
- MVU test: our MVU test framework is using `multi-db` migration mode, so we don't need to change the `migration_mode`

In other words, we should have `BABEL-3655.mix` and `BABEL-3655-vu-{prepare/verify/cleanup}.sql`

xhfanhe: Yes I agree, I've asked Ray to do the same.

test/JDBC/expected/BABEL-3655-vu-cleanup.out
```diff
@@ -0,0 +1,6 @@
+DROP DATABASE babel_3655_test_db
+GO
+~~ERROR (Code: 911)~~
+
+~~ERROR (Message: database "babel_3655_test_db" does not exist)~~
```
2jungkook: We don't need a cleanup file?

raydhkim: DROP DATABASE is the core of the test, so we decided to include it in the verify file and have additional DROP command in the cleanup file for additional check. Do you think it is redundant?

test/JDBC/expected/BABEL-3655.out
2jungkook: FYI, adding an MVU test is almost always beneficial in terms of testing coverage. A broader testing coverage can help future development from various regressions as well. Actually, we have a plan to migrate all normal tests into MVU tests (not sure when we do it though).

test/JDBC/expected/BABEL-3655-vu-cleanup.out
xhfanhe: Cleanup file is optional. We can remove this file

xhfanhe (approved):

PR #809, opened 2022-11-08
Replace unsafe functions with the safer options

basasairohan:
### Description

Currently in `check_server_collation_name` function in `guc.c` file we are using `strcpy` which may cause buffer overflow. This change replaces it with `strncpy` to avoid buffer overflow

Task: BABEL-3581
Signed-off-by: Sai Rohan Basa [bsrohan@amazon.com](mailto:bsrohan@amazon.com)
 
### Issues Resolved

[List any issues this PR will resolve]

### Test Scenarios Covered ###
* **Use case based -** We have tested the changes manually. But it is currently not possible to add automated tests

* **Boundary conditions -** N/A

* **Arbitrary inputs -** N/A

* **Negative test cases -** N/A

* **Minor version upgrade tests -** N/A

* **Major version upgrade tests -** N/A

* **Performance tests -** N/A

* **Tooling impact -** N/A

* **Client tests -** N/A

### Check List
- [ ] Commits are signed per the DCO using --signoff 

By submitting this pull request, I confirm that my contribution is under the terms of the Apache 2.0 and PostgreSQL licenses, and grant any person obtaining a copy of the contribution permission to relicense all or a portion of my contribution to the PostgreSQL License solely to contribute all or a portion of my contribution to the PostgreSQL open source project.

For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/babelfish-for-postgresql/babelfish_extensions/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #543, opened 2022-08-16
Use pstrdup for server_collation_name and default_locale

2jungkook:
### Description

Previously, we cached server_collation_name and default_locale from
GetConfigOption(). However, the returned pointers can become invalid if some
configuration related functions are called, e.g., ProcessConfigFile().

This commit makes those returned string pstrdup'd and adds assertions to
check server_collation_name and default_locale are not changed after the
first assignments.

Also, it adds --enable-cassert for all dev/test-related build scripts and 
remove an unstable test.

Signed-off-by: Jungkook Lee <jungkook@amazon.com>
 
### Issues Resolved

BABEL-3451

### Check List
- [x] Commits are signed per the DCO using --signoff 

By submitting this pull request, I confirm that my contribution is under the terms of the Apache 2.0 and PostgreSQL licenses, and grant any person obtaining a copy of the contribution permission to relicense all or a portion of my contribution to the PostgreSQL License solely to contribute all or a portion of my contribution to the PostgreSQL open source project.

For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/babelfish-for-postgresql/babelfish_extensions/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #492, opened 2022-08-03
Implemented sys.all_parameters

faviansamatha:
### Description
Previously, sys.all_parameters was not implemented in Babelfish. This caused issues in SSMS' Generate Scripts Wizard functionality. This commit adds this system view.

Task: BABELFISH-519
Signed-off-by: Favian (Ian) Samatha <ians@bitquilltech.com>

### Check List
- [x] Commits are signed per the DCO using --signoff 

By submitting this pull request, I confirm that my contribution is under the terms of the Apache 2.0 and PostgreSQL licenses, and grant any person obtaining a copy of the contribution permission to relicense all or a portion of my contribution to the PostgreSQL License solely to contribute all or a portion of my contribution to the PostgreSQL open source project.

For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/babelfish-for-postgresql/babelfish_extensions/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #822, opened 2022-11-10
Correct the default collation of Babelfish data types

Deepesh125:
### Description

Correct the default collation of Babelfish data types.

This commit contains following changes:
1. Move collation creation script from babelfishpg_tsql extension to babelfishpg_common extension.
2. Implements hook define_type_default_collation_hook to override the default collation of Babelfish data types being defined.
3. Move collation related GUC, babelfishpg_tsql.server_collation_name and babelfishpg_tsql.default_locale to babelfishpg_common. Initialisation of these GUCs required to load collations info correctly which would not happen if they are part of babelfishpg_tsql because we load babelfishpg_common before babelfishpg_tsql.
4. Changes in sys objects to avoid collation conflict error or to have correct collation
5. Introduced new function sys.babelfish_update_collation_to_default(schema_name varchar, table_name varchar, column_name varchar) in upgrade script, babelfishpg_tsql--2.2.0--2.3.0.sql, which could be used to set the collation of any column of any view to server's default collation.
6. Introduced new function sys.babelfish_update_collation_to_default(typname) in upgrade script babelfishpg_common--2.2.0--2.3.0.sql which could be used to update the default collation of Babelfish data types.
7. Introduced two new GUCs namely babelfishpg_tsql.restored_server_collation_name and babelfishpg_tsql.restored_default_locale to restore the user defined value of babelfishpg_tsql.server_collation_name and babelfishpg_tsql.default_locale respectively during major version upgrade.
8. Upgrade script, babelfishpg_tsql--2.2.0--2.3.0.sql, Fixed default collation of all sys views/catalog using above procedure

### Testing

- Use case based: Added few new tests cases
- Boundary conditions: Not required
- Arbitrary inputs: Not required
- Negative test scenarios: Not required
- Minor version upgrade tests: Relying on github actions 
- Major version upgrade tests: Relying on github actions. 
- Performance tests: Have not seen big impact on sla test cases except **few tests failure which will be fixed**
- Tooling impact: **To be done**
- Client tests

### MVU Testing

MVU works fine. But just need to ensure that we persist customer defined GUCs like `babelfishpg_tsql.server_collation_name` and `babelfishpg_tsql.default_locale` GUC during upgrade.

Engine PR: https://github.com/babelfish-for-postgresql/postgresql_modified_for_babelfish/pull/44
 
### Issues Resolved

BABEL-3315

Authored-by: Dipesh Dhameliya <dddhamel@amazon.com>
Co-Authored-by: Nirmit Shah <nirmisha@amazon.com>

### Check List


contrib/babelfishpg_common/sql/collation.sql
```diff
@@ -0,0 +1,146 @@
...
+CREATE COLLATION IF NOT EXISTS sys.Ukrainian_CS_AS (provider = icu, locale = 'uk_UA');
+CREATE COLLATION sys.Ukrainian_CI_AI (provider = icu, locale = 'uk_UA@colStrength=primary', deterministic = false);
+CREATE COLLATION sys.Ukrainian_CI_AS (provider = icu, locale = 'uk_UA@colStrength=secondary', deterministic = false);
+
+CREATE COLLATION IF NOT EXISTS sys.Vietnamese_CS_AS (provider = icu, locale = 'vi_VN');
+CREATE COLLATION sys.Vietnamese_CI_AI (provider = icu, locale = 'vi_VN@colStrength=primary', deterministic = false);
+CREATE COLLATION sys.Vietnamese_CI_AS (provider = icu, locale = 'vi_VN@colStrength=secondary', deterministic = false);
+
+DROP FUNCTION IF EXISTS sys.babelfishpg_common_get_babel_server_collation_oid;
+CREATE OR REPLACE FUNCTION sys.babelfishpg_common_get_babel_server_collation_oid() RETURNS OID
```
2jungkook: If this is only used in the upgrade script, then let's create this in the upgrade script and then drop it at the end of the script.

contrib/babelfishpg_common/src/babelfishpg_common.c
```diff
@@ -30,6 +35,29 @@ PreCreateCollation_hook_type prev_PreCreateCollation_hook = NULL;
...
 void	_PG_fini(void);
 
+static bool check_server_collation_name(char **newval, void **extra, GucSource source)
+{
+	if (is_valid_server_collation_name(*newval))
+	{
+		/*
+			* We are storing value in lower case since
+			* Collation names are stored in lowercase into pg catalog (pg_collation).
+			*/
```
2jungkook: indentation

contrib/babelfishpg_common/src/babelfishpg_common.c
```diff
@@ -42,6 +70,25 @@ _PG_init(void)
 	coll_cb_ptr = (collation_callbacks **) find_rendezvous_variable("collation_callbacks");
 	*coll_cb_ptr = get_collation_callbacks();
 
+	DefineCustomStringVariable("babelfishpg_tsql.server_collation_name",
```
2jungkook, line 85: Since this is in the `common` library, can we change the name to `babelfishpg_common.server_collation_name`? If so, we may also need to think about `parameter group` changes.

contrib/babelfishpg_tsql/sql/babelfishpg_tsql.sql
```diff
@@ -711,8 +711,8 @@ CREATE OR REPLACE VIEW sys.spt_tablecollations_view AS
         o.object_id         AS object_id,
         o.schema_id         AS schema_id,
         c.column_id         AS colid,
-        CASE WHEN p.attoptions[1] LIKE 'bbf_original_name=%' THEN split_part(p.attoptions[1], '=', 2)
-            ELSE c.name END AS name,
+        CASE WHEN p.attoptions[1] collate "C" LIKE 'bbf_original_name=%' THEN split_part(p.attoptions[1] collate "C", '=', 2)
```
2jungkook: Just for my knowledge, how can we know when to use `COLLATE sys.database_default` or `collate "C"`?

contrib/babelfishpg_tsql/sql/sys_views.sql
```diff
@@ -1204,32 +1204,32 @@ select t.name,t.type, ns.oid  as schemaid from
...
 create or replace view sys.all_objects as
 select 
-    cast (name as sys.sysname) 
+    cast (name as sys.sysname) collate sys.database_default
   , cast (object_id as integer) 
   , cast ( principal_id as integer)
   , cast (schema_id as integer)
   , cast (parent_object_id as integer)
-  , cast (type as char(2))
+  , cast (type as char(2)) collate sys.database_default
```
2jungkook, line 1218: sys.bpchar(2)?

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.2.0--2.3.0.sql
```diff
@@ -1,5 +1,5 @@
 -- complain if script is sourced in psql, rather than via ALTER EXTENSION
-\echo Use "ALTER EXTENSION ""babelfishpg_tsql"" UPDATE TO '2.3.0'" to load this file. \quit
+-- \echo Use "ALTER EXTENSION ""babelfishpg_tsql"" UPDATE TO '2.3.0'" to load this file. \quit
```
2jungkook: This is a temporary change, right?

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.2.0--2.3.0.sql
```diff
@@ -31,68 +31,62 @@ end
...
-			WHERE ((SELECT coalesce(in_table_name,'')) = '' OR table_name = in_table_name collate sys.bbf_unicode_general_ci_as)
-			AND ((SELECT coalesce(in_table_owner,'')) = '' OR table_owner = in_table_owner collate sys.bbf_unicode_general_ci_as)
-			AND ((SELECT coalesce(in_table_qualifier,'')) = '' OR table_qualifier = in_table_qualifier collate sys.bbf_unicode_general_ci_as)
-			AND ((SELECT coalesce(cs_as_in_table_type,'')) = ''
-			    OR table_type = opt_table
-			    OR table_type = opt_view)
-			ORDER BY table_qualifier, table_owner, table_name;
-		END IF;
-	END;
+CREATE OR REPLACE PROCEDURE babelfish_drop_deprecated_view(schema_name varchar, view_name varchar) AS
```
2jungkook: Why do we need this in addition to `babelfish_drop_deprecated_object`?

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.2.0--2.3.0.sql
```diff
@@ -203,785 +197,200 @@ SELECT
...
-GRANT EXECUTE ON PROCEDURE sys.sp_helpsrvrolemember TO PUBLIC;
-
--- Need to add parameter for tsql_type_max_length_helper 
-ALTER FUNCTION sys.tsql_type_max_length_helper RENAME TO tsql_type_max_length_helper_deprecated_in_2_3_0;
+LANGUAGE plpgsql;
 
-CREATE OR REPLACE FUNCTION sys.tsql_type_max_length_helper(IN type TEXT, IN typelen INT, IN typemod INT, IN for_sys_types boolean DEFAULT false, IN used_typmod_array boolean DEFAULT false)
-RETURNS SMALLINT
-AS $$
+CREATE OR REPLACE FUNCTION sys.babelfish_conv_date_to_string(IN p_datatype TEXT,
```
2jungkook, line 191: It would be really hard to review the upgrade script changes by 1-2 reviewers. If you have a spreadsheet containing a list of objects, then we may ask many others to review a reasonable number of objects.

contrib/babelfishpg_tsql/src/collation.c
```diff
@@ -79,10 +80,15 @@ collation_list(PG_FUNCTION_ARGS)
...
+
+/*
+ * get_server_collation_oid - this is being used by get_babel_server_collation_oid_deprecated_in_2_3_0
+ * which is depcrecated and is not recommended to be used.
+ */
 Datum
 get_server_collation_oid(PG_FUNCTION_ARGS)
 {
-	PG_RETURN_OID(tsql_get_server_collation_oid_internal(false));
+	PG_RETURN_OID(InvalidOid);
```
2jungkook: Let's add a warning message (reference: https://github.com/babelfish-for-postgresql/babelfish_extensions/blob/BABEL_2_X_DEV/contrib/babelfishpg_tsql/src/rolecmds.c#L1364)

contrib/babelfishpg_tsql/src/properties.c
```diff
@@ -168,8 +170,14 @@ Datum serverproperty(PG_FUNCTION_ARGS) {
 	}
     else if (strcasecmp(property, "Collation") == 0)
     {
+		#if 0
 		char *ret = pstrdup(pltsql_server_collation_name);
 		vch = tsql_varchar_input(ret, strlen(ret), -1);
+		#endif
+		if (!server_collation_name)
+			server_collation_name = GetConfigOption("babelfishpg_tsql.server_collation_name", true, false);
```
2jungkook: We should not cache the value from `GetConfigOption` (Ref: https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/543).

contrib/babelfishpg_common/sql/collation.sql
```diff
@@ -0,0 +1,146 @@
+-- create babelfish collations
+
+CREATE COLLATION IF NOT EXISTS sys.Arabic_CS_AS (provider = icu, locale = 'ar_SA');
```
surendravishnoi, line 3: why IF NOT EXISTS for few commands?

contrib/babelfishpg_common/sql/upgrades/babelfishpg_common--2.2.0--2.3.0.sql
```diff
@@ -3,6 +3,13 @@
 
 SELECT set_config('search_path', 'sys, '||current_setting('search_path'), false);
 
+CREATE OR REPLACE FUNCTION sys.babelfishpg_common_get_babel_server_collation_oid() RETURNS OID
```
surendravishnoi, line 6: another alternative is do everything using a C function

contrib/babelfishpg_common/sql/collation.sql
Deepesh125: This function is being used by upgrade scripts of both extensions? Do we want to re-create it separately in both extensions?

contrib/babelfishpg_common/src/babelfishpg_common.c
```diff
@@ -30,6 +35,29 @@ PreCreateCollation_hook_type prev_PreCreateCollation_hook = NULL;
...
+static bool check_server_collation_name(char **newval, void **extra, GucSource source)
+{
+	if (is_valid_server_collation_name(*newval))
+	{
+		/*
+			* We are storing value in lower case since
+			* Collation names are stored in lowercase into pg catalog (pg_collation).
+			*/
+		char *dupval = pstrdup(*newval);
+		strcpy(*newval, downcase_identifier(dupval, strlen(dupval), false, false));
```
surendravishnoi: use strncpy, also no need to duplicate string

contrib/babelfishpg_common/src/babelfishpg_common.c
Deepesh125: Yes, This would be major change and IMHO, should be taken up separately. I also need to evaluate complication around such changes. SO better to take it separate.

contrib/babelfishpg_tsql/sql/babelfishpg_tsql.sql
```diff
@@ -369,8 +369,8 @@ CREATE OR REPLACE VIEW sys.sp_columns_100_view AS
      JOIN sys.pg_namespace_ext t2 ON t1.relnamespace = t2.oid
      JOIN pg_catalog.pg_roles t3 ON t1.relowner = t3.oid
      LEFT OUTER JOIN sys.babelfish_namespace_ext ext on t2.nspname = ext.nspname
-     JOIN information_schema_tsql.columns t4 ON (t1.relname = t4."TABLE_NAME" AND ext.orig_name = t4."TABLE_SCHEMA")
-     LEFT JOIN pg_attribute a on a.attrelid = t1.oid AND a.attname = t4."COLUMN_NAME"
+     JOIN information_schema_tsql.columns t4 ON (t1.relname = t4."TABLE_NAME" COLLATE sys.database_default AND ext.orig_name = t4."TABLE_SCHEMA" COLLATE sys.database_default)
```
surendravishnoi: if we do not specific database default, which collation is used?

contrib/babelfishpg_tsql/sql/sys_views.sql
Deepesh125: We are not going to address such data type diff issue with this work.

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.2.0--2.3.0.sql
```diff
@@ -1,5 +1,5 @@
 -- complain if script is sourced in psql, rather than via ALTER EXTENSION
```
surendravishnoi: where is 2.1 to 2.2?

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.2.0--2.3.0.sql
```diff
@@ -31,68 +31,62 @@ end
...
+    update pg_attribute set attcollation = sys.babelfishpg_common_get_babel_server_collation_oid() where attname = column_name collate sys.database_default and attrelid = table_oid;
+END
 $$
 LANGUAGE plpgsql;
 
--- please add your SQL here
+
+
+-- deprecate get_babel_server_collation_oid as corresponding functionality is moved to babelfishpg_common extension as sys.babelfishpg_common_get_babel_server_collation_oid
+ALTER FUNCTION sys.get_babel_server_collation_oid() RENAME TO get_babel_server_collation_oid_deprecated_in_2_3_0;
```
surendravishnoi, line 64: how does ordering happens here? Can common upgrade script execute first

contrib/babelfishpg_tsql/sql/babelfishpg_tsql.sql
Deepesh125: In this case, split_part() works with deterministic collations only. So we need to ensure that we do not pass any column/var which is having non-deterministic collations. In this case, collation of p.attoptions[1] is already "C" so we don't need to specify collate "C" clause (I will remove this in next revision). 

But for other cases, we might need to specify the collation to ensure that there won't be any collation conflict. For example, equal operator, it will throw an error if both operands are having **non-default** collation and both are **different**. Same for **LIKE**. We also need to take care of such conflict in **UNION ALL** operation.

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.2.0--2.3.0.sql
Deepesh125: It is temporary. Will ask @shah-nirmit to remove this.

contrib/babelfishpg_tsql/src/collation.c
Deepesh125: Done!

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.2.0--2.3.0.sql
```diff
@@ -203,283 +197,3034 @@ SELECT
 $BODY$
 LANGUAGE SQL VOLATILE STRICT;
 
-CREATE OR REPLACE PROCEDURE sys.sp_helpsrvrolemember("@srvrolename" sys.SYSNAME = NULL) AS
+-- Drops a function if it does not have any dependent objects.
+-- Is a temporary procedure for use by the upgrade script. Will be dropped at the end of the upgrade.
+-- Please have this be one of the first statements executed in this upgrade script. 
+CREATE OR REPLACE PROCEDURE babelfish_drop_deprecated_function(schema_name varchar, func_name varchar) AS
```
surendravishnoi: drop object should work here?

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.2.0--2.3.0.sql
surendravishnoi: may be original writers review their procedures/functions

test/JDBC/input/BABEL-SP_COLUMNS_MANAGED-dep-vu-prepare.sql
```diff
@@ -0,0 +1,6 @@
+CREATE PROC babel_sp_columns_managed_dep_vu_prepare_p1 @Catalog AS sys.SYSNAME = NULL, @Owner AS sys.SYSNAME = NULL, @Table AS sys.SYSNAME = NULL, @Column AS sys.SYSNAME = NULL, @SchemaType AS sys.SYSNAME = 0
```
surendravishnoi, line 1: how did we get the list of tests to be added? All the functions/procedures modified with missing tests?

contrib/babelfishpg_common/sql/collation.sql
2jungkook: Yes, same as we do for `sys.babelfish_drop_deprecated_object`.

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.2.0--2.3.0.sql
```diff
@@ -1284,1602 +671,5951 @@ LANGUAGE plpgsql
...
+RETURNS text
+AS 'babelfishpg_tsql', 'get_pltsql_function_signature' LANGUAGE C;
+
+-- All the available views
+
+CALL sys.babelfish_update_collation_to_default('sys', 'sysdatabases', 'filename');
+
+CALL sys.babelfish_update_collation_to_default('sys', 'schemas', 'name');
+
+CALL sys.babelfish_update_collation_to_default('sys', 'server_principals', 'name');
```
2jungkook, line 2667: Don't we need to re-create dependent objects after updating collations?

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.2.0--2.3.0.sql
```diff
@@ -1284,1602 +671,5951 @@ LANGUAGE plpgsql
...
+  , CAST(0 AS SYS.BIT) as is_tempdb_spill_to_remote_store
+  , CAST(0 AS SYS.BIT) as is_stale_page_detection_on
+  , CAST(0 AS SYS.BIT) as is_memory_optimized_enabled
+  , CAST(0 AS SYS.BIT) as is_ledger_on
+ from sys.babelfish_sysdatabases d 
+ INNER JOIN sys.sysdatabases s on d.dbid = s.dbid
+ LEFT OUTER JOIN pg_catalog.pg_collation c ON d.default_collation = c.collname;
+GRANT SELECT ON sys.databases TO PUBLIC;
+
+CALL sys.babelfish_update_collation_to_default('sys', 'databases', 'name');
```
2jungkook, line 2777: Do we really need to update collations right after `create or replace view sys.databases`? Note that `2.2.0--2.3.0` upgrade script is executed in `2.3.0` binary.

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.0.0--2.1.0.sql
```diff
@@ -397,9 +397,9 @@ GRANT EXECUTE ON PROCEDURE sys.sp_column_privileges TO PUBLIC;
 CREATE OR REPLACE VIEW sys.sp_table_privileges_view AS
 -- Will use sp_column_priivleges_view to get information from SELECT, INSERT and REFERENCES (only need permission from 1 column in table)
 SELECT DISTINCT
-CAST(TABLE_QUALIFIER AS sys.sysname) AS TABLE_QUALIFIER,
+CAST(TABLE_QUALIFIER AS sys.sysname) COLLATE sys.database_default AS TABLE_QUALIFIER,
```
2jungkook: We should try to avoid changing the previous upgrade scripts. Why do we need to change `2.0.0--2.1.0` upgrade script here?

2jungkook:
Please remove `contrib/babelfishpg_tsql--2.3.0.sql` file.

contrib/babelfishpg_common/sql/collation.sql
Deepesh125: Yes, @shah-nirmit will make this change. Thanks

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.2.0--2.3.0.sql
Deepesh125: Yes, It is temp change. @shah-nirmit will fix this.

contrib/babelfishpg_common/sql/collation.sql
Deepesh125: It should be optional. But we just copy/paste this script from Babelfishpg_tsql to common extension.

contrib/babelfishpg_common/sql/upgrades/babelfishpg_common--2.2.0--2.3.0.sql
Deepesh125: it is C function. can you please elaborate what are you suggesting here? Thanks..

contrib/babelfishpg_common/src/babelfishpg_common.c
Deepesh125: Yes, we should use strncpy only. @shah-nirmit will make the change https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/809

contrib/babelfishpg_tsql/sql/babelfishpg_tsql.sql
Deepesh125: Default collation of Data type on both sides would be bubbled up to use as collation for comparison. Now there could be various cases,
1. If both side has default collation then either PG default or TSQL default collation would be used based on dialect
2. if either side has non-default collation then it will be used for comparison

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.2.0--2.3.0.sql
Deepesh125: Did not get you. Please add more info

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.2.0--2.3.0.sql
Deepesh125: Yes, common extension will be upgraded first and then tsql

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.2.0--2.3.0.sql
Deepesh125: It is removed in latest revision

test/JDBC/input/BABEL-SP_COLUMNS_MANAGED-dep-vu-prepare.sql
Deepesh125: We used upgrade_validation and sql_validation framework to list down all the objects with missing test cases. We then went through whole list manually to understand whether test cases are required. More details can be found at https://quip-amazon.com/PwbHAEdd9CYJ/Additional-Test-cases-required-for-BABEL-3315

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.0.0--2.1.0.sql
Deepesh125: Upgrade path 2.2.0--2.3.0 works fine with above approach. But other upgrade paths, for example, 2.0.0 → 2.1.0, is not working and failing with collation conflict error (during MVU, When we try to upgrade from 1.x to 2.3.0). 
    1. For example, `CREATE OR REPLACE VIEW sys.sp_table_privileges_view throws following error during 2.0.0 --> 2.1.0 upgrade
```
ERROR:  collation mismatch between implicit collations "C" and "bbf_unicode_cp1_ci_as"
            LINE 15: CAST(t2.dbname AS sys.sysname) AS TABLE_QUALIFIER,
                     ^
            HINT:  You can choose the collation by applying the COLLATE clause to one or both expressions.
```

This is happening probably because we upgrade babelfishpg_common extension first and then babelfishpg_tsql which involves re-creation of sys objects. So To resolve this, we updated previous upgrade script only for this branch.

contrib/babelfishpg_tsql/src/properties.c
Deepesh125: Updated in latest revision

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.2.0--2.3.0.sql
surendravishnoi: I meant where are changes in upgrade script for 2.1.0--2.2.0?

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.2.0--2.3.0.sql
surendravishnoi: isn't that a problem? you create this function in common first and later renaming it to deprecated here? it should be other way around

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.2.0--2.3.0.sql
```diff
@@ -203,1339 +175,5787 @@ SELECT
...
+
+CALL sys.babelfish_update_collation_to_default('information_schema_tsql', 'columns', 'TABLE_CATALOG');
+CALL sys.babelfish_update_collation_to_default('information_schema_tsql', 'columns', 'TABLE_SCHEMA');
+CALL sys.babelfish_update_collation_to_default('information_schema_tsql', 'columns', 'TABLE_NAME');
+CALL sys.babelfish_update_collation_to_default('information_schema_tsql', 'columns', 'COLUMN_NAME');
+CALL sys.babelfish_update_collation_to_default('information_schema_tsql', 'columns', 'COLUMN_DEFAULT');
+CALL sys.babelfish_update_collation_to_default('information_schema_tsql', 'columns', 'DATA_TYPE');
+CALL sys.babelfish_update_collation_to_default('information_schema_tsql', 'columns', 'CHARACTER_SET_CATALOG');
+CALL sys.babelfish_update_collation_to_default('information_schema_tsql', 'columns', 'CHARACTER_SET_SCHEMA');
+CALL sys.babelfish_update_collation
```
2jungkook: Please be careful as it has been changed recently (https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/850).

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.0.0--2.1.0.sql
2jungkook: For example, `sys.all_objects` is also in `2.1.0--2.2.0`. What happens if a user does mVU from 2.1 to 2.3?

test/JDBC/upgrade/13_4/schedule
```diff
@@ -76,7 +78,7 @@ schema_resolution_proc
 SYSTEM_USER
 Test-Default-Columns
 BABEL-1189
-BABEL-1963
+#BABEL-1963
```
2jungkook: reason?

2jungkook:
I saw many duplicate test input/output files. Can we avoid those by using a trick like `before-14_6` (ref: https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/270/files)?

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.2.0--2.3.0.sql
```diff
@@ -203,1339 +175,5787 @@ SELECT
...
-LEFT OUTER JOIN sys.babelfish_view_def bvd 
- on (
-      ext.orig_name = bvd.schema_name AND 
-      ext.dbid = bvd.dbid AND
-      ao.name = bvd.object_name 
-   )
-LEFT JOIN pg_proc p ON ao.object_id = CAST(p.oid AS INT)
-WHERE ao.type in ('P', 'RF', 'V', 'TR', 'FN', 'IF', 'TF', 'R');
-GRANT SELECT ON sys.all_sql_modules_internal TO PUBLIC;
+create or replace view sys.shipped_objects_not_in_sys AS
```
2jungkook: There is no meaningful change in the original sql file. Why do we need this?

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.2.0--2.3.0.sql
Deepesh125: We are not creating separate function in common upgrade script as well in tsql script and drop them at the end of script. Please check latest revision.

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.2.0--2.3.0.sql
Deepesh125: Yes, we need to remove this. @shah-nirmit please make this change. Thanks.

contrib/babelfishpg_common/src/babelfishpg_common.c
shah-nirmit: Resolved in latest revision

contrib/babelfishpg_common/sql/collation.sql
shah-nirmit: Resolved in latest revision

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.2.0--2.3.0.sql
shah-nirmit: Resolved in latest revision

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.2.0--2.3.0.sql
shah-nirmit: Resolved in latest revision

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.2.0--2.3.0.sql
Deepesh125: It is not necessary. Primary reason of re-creating dependent object would be because its column also pick the non-deterministic collations. But postgres, won't pick change in collation by merely re-creating the object. So we have to update catalog.

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.2.0--2.3.0.sql
Deepesh125: Yes, It is required. Because PG does not account change in collation while recreating views.

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.0.0--2.1.0.sql
Deepesh125: Only `sys.sp_table_privileges_view` requires this change. And is not present in `2.1.0--2.2.0` path

test/JDBC/upgrade/13_4/schedule
Deepesh125: Resolved by Nirmit in latest revision.

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.2.0--2.3.0.sql
timchang514: Why was babelfish_conv_date_to_string altered in the upgrade script but not in the original, sys_function_helpers.sql?

timchang514 (changes requested):
There seem to be several functions which are only altered in the upgrade script and not in their original definitions. For ex. function sys.babelfish_conv_date_to_string, babelfish_conv_datetime_to_string, and so on. Should the changes be reflected in the original definition? Or is the diff only caused by refactoring of the upgrade script.

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.2.0--2.3.0.sql
```diff
@@ -203,1339 +180,5759 @@ SELECT
...
-
-        WHEN arg_type IN ('integer'::regtype) THEN
-            RETURN sys.format_numeric(arg, p_format_pattern, p_culture, 'integer');
-
-         WHEN arg_type IN ('bigint'::regtype) THEN
-            RETURN sys.format_numeric(arg, p_format_pattern, p_culture, 'bigint');
-
-        WHEN arg_type IN ('numeric'::regtype) THEN
-            RETURN sys.format_numeric(arg, p_format_pattern, p_culture, 'numeric');
+select CAST(('DF_' || tab.name || '_' || d.oid) as sys.sysname) as name
```
rossd-bitquill, line 3632: Does this need to have a COLLATE added to the tab.name?

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.2.0--2.3.0.sql
```diff
@@ -203,1339 +180,5759 @@ SELECT
...
+    implied_dbo_permissions boolean;
+    fully_supported boolean;
+    object_name text COLLATE sys.database_default;
+    database_id smallint;
+    namespace_id oid;
+    object_type text;
+    function_signature text;
+    qualified_name text;
+    return_value integer;
+   	cs_as_securable text COLLATE "C" := securable;
```
rossd-bitquill: Inconsistent spacing.

contrib/babelfishpg_tsql/sql/babelfishpg_tsql.sql
```diff
@@ -1519,7 +1519,7 @@ UNION
 SELECT
 CAST(t2.dbname AS sys.sysname) AS TABLE_QUALIFIER,
 CAST(s1.name AS sys.sysname) AS TABLE_OWNER,
-CAST(t1.relname AS sys.sysname) AS TABLE_NAME,
+CAST(t1.relname AS sys.sysname) COLLATE sys.database_default AS TABLE_NAME,
```
faviansamatha: Curious, why does this one have COLLATE sys.database_default but not the other fields? I would assume because it's casted to sys.sysname the collation would also apply since it's a sys datatype?

contrib/babelfishpg_tsql/sql/sys_views.sql
```diff
@@ -1159,7 +1159,7 @@ SELECT CAST(c.conname as sys.sysname) as name
   , CAST(NULL as integer) as principal_id 
   , CAST(c.connamespace as integer) as schema_id
   , CAST(conrelid as integer) as parent_object_id
-  , CAST('C' as char(2)) as type
+  , CAST('C' as sys.bpchar(2)) as type
```
rossd-bitquill: Change to bpchar without COLLATE is inconsistent with other instances of the same type.  I.e. another comment said we would not be changing types here, but just added collation, this one changes type and does not add collation.

contrib/babelfishpg_tsql/sql/sys_views.sql
```diff
@@ -1195,27 +1195,27 @@ select t.name,t.type, ns.oid as schemaid from
...
 create or replace view sys.all_objects as
 select 
-    cast (name as sys.sysname) 
+    cast (name as sys.sysname) collate sys.database_default
   , cast (object_id as integer) 
   , cast ( principal_id as integer)
   , cast (schema_id as integer)
   , cast (parent_object_id as integer)
-  , cast (type as char(2))
+  , cast (type as char(2)) collate sys.database_default
```
leszekamz, line 1218: Sorry, commenting here, because further down, I'm not able to comment, but it is related.

Should similar change be done in the sys.data_spaces view for this column?
CAST('FG' as CHAR(2)) AS type,  --> CAST('FG' as CHAR(2)) collate sys.database_default AS type,

Then subsequently the sys.filegroups view is using this column in where statement
FROM sys.data_spaces ds WHERE type = 'FG';

Does the above where statement need the collate as well?
FROM sys.data_spaces ds WHERE type = 'FG' collate sys.database_default;

The sys.filegroups view is listed as one of the views to review in the related quip document.

contrib/babelfishpg_tsql/sql/sys_views.sql
Deepesh125: Thanks for pointing this out. Yes, we are not going to change any data types. @shah-nirmit please make this change.

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.2.0--2.3.0.sql
shah-nirmit: Removed in latest revision

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.2.0--2.3.0.sql
shah-nirmit: The change has been taken account into in latest revision

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.0.0--2.1.0.sql
2jungkook: Then why did you change other objects in `2.0.0--2.1.0`?

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.2.0--2.3.0.sql
Deepesh125: babelfish_conv_date_to_string is already present in upgrade script updated by you. So you can skip it.

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.0.0--2.1.0.sql
Deepesh125: This is happening probably because we upgrade babelfishpg_common extension first which will update default collation of sys data types and then babelfishpg_tsql extension which involves execution of multiple minor upgrade path. During `2.0.0--2.1.0`, we are running into issue only for `sys.sp_table_privileges_view` so we have made this change in the definition. 


One more thing to note is that, this object is being recreated during `2.2.0--2.3.0`. So any cx coming to 2.3.0 irrespective of upgrade or fresh install, they will get the same definition.

contrib/babelfishpg_tsql/sql/babelfishpg_tsql.sql
Deepesh125: Seems redundant, @shah-nirmit can you check once and remove if not required?

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.2.0--2.3.0.sql
Deepesh125: It does not seems to be required as tab.name is _CI_SYSNAME which would have correct tsql collation

contrib/babelfishpg_tsql/sql/sys_views.sql
shah-nirmit: The datatype for type column for `sys.data_spaces` view is char ,which would not have `sys.database_default` collation by default (suggestion:we want to use sys.bpchar instead to inherit the default collation) because of which `sys.filegroups` need explicit collate clause for type column

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.2.0--2.3.0.sql
shah-nirmit: This was due to direct copy paste of definition from fresh install script , fixed the spacing there as well in the latest revision

contrib/babelfishpg_tsql/sql/sys_views.sql
shah-nirmit: Fixed in latest revision

contrib/babelfishpg_tsql/sql/babelfishpg_tsql.sql
shah-nirmit: Yes it was a redundant collate clause fixed in the latest revision

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.2.0--2.3.0.sql
```diff
@@ -3243,6 +6807,8 @@ AS 'babelfishpg_tsql', 'update_user_catalog_for_guest';
  
 CALL sys.babelfish_update_user_catalog_for_guest();
 
+-- comment for the review: Check https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/708 to see if depcrecated
+-- view can cause any collation conflict to cx.
 ALTER VIEW sys.sp_sproc_columns_view RENAME TO sp_sproc_columns_view_deprecated_in_2_3_0;
```
faviansamatha, line 6819: This view hasn't been adjusted for collation and should be (both the original object and the upgrade script) . The previous version is of this object is indeed deprecated as the new version has correct datatypes. Are there any specific questions regarding this view? 

NOTE: There are some comparisons between the `ss` table from the inner query. This `ss` table consist of postgres tables with the pg collation.  I can't seem to comment  below this line here as well so can't give areas of suggestions to collate but the view should be similar to sys.all_parameters.

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.2.0--2.3.0.sql
```diff
@@ -203,1339 +180,5672 @@ SELECT
...
+CALL sys.babelfish_update_collation_to_default('sys', 'plan_guides', 'parameters');
+CALL sys.babelfish_update_collation_to_default('sys', 'plan_guides', 'hints');
+
+CALL sys.babelfish_update_collation_to_default('sys', 'spatial_index_tessellations', 'tessellation_scheme');
+CALL sys.babelfish_update_collation_to_default('sys', 'spatial_index_tessellations', 'level_1_grid_desc');
+CALL sys.babelfish_update_collation_to_default('sys', 'spatial_index_tessellations', 'level_2_grid_desc');
+CALL sys.babelfish_update_collation_to_default('sys', 'spatial_index_tessellations', 'level_3_grid_desc');
+CALL sys.babelfish_update_collation_to_default('sys', 'spatial_index_tessellations', 'level_4_grid_desc');
+
+CREATE OR REPLACE VIEW sys.all_parameters
```
faviansamatha, line 4489: I don't think this was modified to adjust for the new collations.

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.2.0--2.3.0.sql
```diff
@@ -203,1339 +180,5672 @@ SELECT
...
+      CASE 
+        WHEN st.is_table_type = 1 THEN 0 -- TVP case
+        WHEN st.is_user_defined = 1  THEN st.scale
+        ELSE sys.tsql_type_scale_helper(st.name, typmod,false)
+      END
+    AS sys.tinyint) AS scale
+  , CAST(
+      CASE
+        WHEN is_out_scalar = 1 THEN 1 -- Output of a scalar function
+        WHEN ss.proargmodes[(ss.x).n] in ('o', 'b', 't') THEN 1
```
faviansamatha, line 4529: Does this need to be collated? Goes for all the text/string items in ss. Everything in ss should be from pg tables and have the pg collation

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.2.0--2.3.0.sql
```diff
@@ -203,1339 +180,5672 @@ SELECT
...
+    SELECT
+      p.oid AS p_oid,
+      p.proargnames,
+      p.proargmodes,
+      p.prokind,
+      json_extract_path(CAST(p.probin as json), 'typmod_array') AS typmod_array,
+      information_schema._pg_expandarray(
+      COALESCE(p.proallargtypes,
+        CASE 
+          WHEN p.prokind = 'f' THEN (CAST( p.proargtypes AS oid[]) || p.prorettype) -- Adds return type if not present on proallargtypes
```
faviansamatha, line 4558: Do we need to collate here?

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.2.0--2.3.0.sql
```diff
@@ -203,1339 +180,5672 @@ SELECT
...
+  COALESCE(pg_get_function_result(ss.p_oid), '') AS return_type,
+  CAST(ss.typmod_array->>(ss.x).n-1 AS INT) AS typmod, 
+  CAST(
+    CASE
+      WHEN ss.prokind = 'f' AND ss.proargnames[(ss.x).n] IS NULL THEN 1 -- checks if param is output of scalar function
+      ELSE 0
+    END 
+  AS INT) AS is_out_scalar
+WHERE ( -- If it's a Table function, we only want the inputs
+      return_type NOT LIKE 'TABLE(%' OR
```
faviansamatha, line 4577: Do we need to collate here for return_type? It's part of ss which is pg

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.2.0--2.3.0.sql
```diff
@@ -203,1339 +180,5672 @@ SELECT
...
+    WHERE (
+      p.pronamespace in (select schema_id from sys.schemas union all select oid from pg_namespace where nspname = 'sys')
+      AND (pg_has_role(p.proowner, 'USAGE') OR has_function_privilege(p.oid, 'EXECUTE'))
+      AND p.probin like '{%typmod_array%}') -- Needs to have a typmod array in JSON format
+  ) ss ON t.oid = (ss.x).x,
+  COALESCE(pg_get_function_result(ss.p_oid), '') AS return_type,
+  CAST(ss.typmod_array->>(ss.x).n-1 AS INT) AS typmod, 
+  CAST(
+    CASE
+      WHEN ss.prokind = 'f' AND ss.proargnames[(ss.x).n] IS NULL THEN 1 -- checks if param is output of scalar function
```
faviansamatha, line 4572: Do we need to collate here as well?

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.2.0--2.3.0.sql
```diff
@@ -203,1339 +180,5672 @@ SELECT
...
+    object_name text)
+RETURNS INTEGER
+AS
+$BODY$
+DECLARE
+    relevant_permissions text[];
+    namespace_id oid;
+    function_signature text;
+    qualified_name text;
+    permission text;
```
rossd-bitquill, line 2041: Do these declarations need COLLATE?

shah-nirmit: No we dont need to specify collate here explicitly , we have handled string comparison in the function definition

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.2.0--2.3.0.sql
shah-nirmit: Yes you are correct ,but this view was updated to use new definition of `sys.tsql_type_max_length_helper` and to implement to view itself reference PR(https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/492). So this definition was already present in the sql upgrade script  on 2_3_STABLE Branch as well

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.2.0--2.3.0.sql
shah-nirmit: No we dont need explicit collate clause for collatable datatype in `ss` , as the comparsion happens with string literal which will have **default** collation , and the ss column's will have pg_collation which will not lead to any collation conflict

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.2.0--2.3.0.sql
shah-nirmit: No, explained in the above comment

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.2.0--2.3.0.sql
shah-nirmit: No, explained in the above comment

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.2.0--2.3.0.sql
shah-nirmit: No, explained in the above comment

contrib/babelfishpg_common/src/collation.c
```diff
@@ -1084,9 +1112,15 @@ get_server_collation_oid_internal(bool missingOk)
...
 {
-
 	if (sql_dialect == SQL_DIALECT_TSQL)
 	{
+		/* 
+		 * If upgrade is going on then we should use oid corresponding to 
+		 * babelfishpg_tsql.restored_server_collation_name.
+		 */
+		if (babelfish_restored_server_collation_name != NULL)
+			return get_collation_oid_internal(babelfish_restored_server_collation_name);
```
2jungkook: Previously, `BABELFISH_CLUSTER_COLLATION_OID` always returned valid oid. Is there any possibility of returning invalid oid here?

contrib/babelfishpg_common/src/collation.h
```diff
@@ -5,6 +5,14 @@
 #include "nodes/nodeFuncs.h"
 #include "nodes/pathnodes.h"
 
+
+extern char *pltsql_default_locale;
+extern char *pltsql_server_collation_name;
```
2jungkook, line 10: I may miss something because of large diff. Would you let me know where these extern variables are used?

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.0.0--2.1.0.sql
2jungkook: I'm also talking about other objects in this upgrade scripts: `sys.all_objects`, `sys.spt_tablecollations_view`, `sys.objects`, and `sys.spt_tablecollations_view`. Please avoid unnecessary changes.

contrib/babelfishpg_common/sql/upgrades/babelfishpg_common--2.2.0--2.3.0.sql
```diff
@@ -3,6 +3,43 @@
...
+
+
+-- Set the default collation of given sys."typname" to oid corresponding to babelfishpg_tsql.server_collation_name
+CREATE OR REPLACE PROCEDURE sys.babelfish_update_collation_to_default(type_name varchar) AS
+$$
+DECLARE
+    coll_oid oid;
+BEGIN
+    SELECT typcollation INTO coll_oid FROM pg_catalog.pg_type WHERE typname = type_name AND typnamespace = (SELECT oid FROM pg_namespace WHERE nspname = 'sys');
+    IF coll_oid = 100 THEN
```
surendravishnoi: what is 100 here? is it safe to assume that this value is reliable in all cases? It would be better to identify collation by name

contrib/babelfishpg_common/src/babelfishpg_common.c
```diff
@@ -30,6 +39,46 @@ PreCreateCollation_hook_type prev_PreCreateCollation_hook = NULL;
...
+}
+
+static bool check_default_locale (char **newval, void **extra, GucSource source)
+{
+	if (find_locale(*newval) >= 0)
+		return true;
+	return false;
+}
+
+static bool check_restored_server_collation_name(char **newval, void **extra, GucSource source)
```
surendravishnoi, line 64: why do we need check functions for restored collation/locale, they are already validated and set internally

contrib/babelfishpg_common/src/collation.c
```diff
@@ -1084,9 +1121,15 @@ get_server_collation_oid_internal(bool missingOk)
 
 Oid BABELFISH_CLUSTER_COLLATION_OID()
 {
-
 	if (sql_dialect == SQL_DIALECT_TSQL)
```
surendravishnoi, line 1099: during upgrade, most of restore run in PG dialect. How are we making sure that we are not missing a valid case due to that

contrib/babelfishpg_common/src/collation.c
Deepesh125: It should not return invalid OID ideally. There are reasons for it, for example, babelfish_restored_server_collation_name would always be valid babelfish collation name or NULL because we uses the same validity function as server_collation_name to validate this GUC also.

contrib/babelfishpg_common/src/collation.h
Deepesh125: These are global variable being used to define server_collation_name and default_locale GUCs.

contrib/babelfishpg_common/src/collation.c
```diff
@@ -1393,4 +1437,46 @@ get_collation_callbacks(void)
...
+
+/*
+ * get_babel_server_collation_oid - corresponding to sys.babelfishpg_tsql_get_babel_server_collation_oid() function
+ * which would be available and useful during upgrade only.
+ */
+Datum
+get_babel_server_collation_oid(PG_FUNCTION_ARGS)
+{
+	/* If babelfish_restored_server_collation_name is set then use it. */
+	if (babelfish_restored_server_collation_name != NULL)
```
surendravishnoi: I thought approach was :

- set restore GUC using dump
- Use this GUC when creating types in MVU to have correct collation
- Later as part of minor version upgrade, skip collation correction if current value is not PG default

It seems the approach taken is:
- set GUC at database level during dump
- I think types will not get the correct collation as alter database needs restart
- Later during minor upgrade, use restore GUC to set correct value wherever needed

It seems first solution is simpler

contrib/babelfishpg_common/src/collation.c
Deepesh125: We are mostly trying to fix the default collation of Babelfish data types which would sub sequently fix sys objects dependent on those data types because collation would bubbled up from default collation of data types. 

So given that we have handled data types differently, all other sys objects would be fixed automatically.

contrib/babelfishpg_common/sql/upgrades/babelfishpg_common--2.2.0--2.3.0.sql
Deepesh125: Yes, it is safe to assume that. check https://github.com/babelfish-for-postgresql/postgresql_modified_for_babelfish/blob/BABEL_2_X_DEV__PG_14_X/src/include/catalog/pg_collation.dat#L15
But I made it to fetch it from pg_collation catalog instead of hard coded value.

contrib/babelfishpg_common/src/babelfishpg_common.c
Deepesh125: Yes, it would be validated already. But only difference being is, NULL should be treated as valid value for these new GUCs.

surendravishnoi: point being that these are internal GUCs and no need to validate the values

contrib/babelfishpg_common/src/collation.c
```diff
@@ -1393,4 +1416,69 @@ get_collation_callbacks(void)
...
+{
+	/* We should only override the default collation for Babelfish data types. */
+	if (strcmp(get_namespace_name(typeNamespace), "sys") != 0)
+		return DEFAULT_COLLATION_OID;
+
+	/* 
+	 * If upgrade is going on then we should use oid corresponding to 
+	 * babelfishpg_tsql.restored_server_collation_name.
+	 */
+	if (babelfish_restored_server_collation_name != NULL)
```
surendravishnoi: we should also add check that this is restore mode

Deepesh125: Addressed in latest rev.

contrib/babelfishpg_common/src/collation.c
```diff
@@ -1393,4 +1416,72 @@ get_collation_callbacks(void)
...
+
+	if (!is_valid_server_collation_name(babelfish_restored_server_collation_name))
+	{
+		ereport(ERROR,
+				(errcode(ERRCODE_INTERNAL_ERROR),
+				 errmsg("Invalid value of babelfishpg_tsql.restored_server_collation_name GUC is detected.")));
+	}
+
+	oldContext = MemoryContextSwitchTo(TopMemoryContext);
+	server_collation_name = pstrdup(babelfish_restored_server_collation_name);
```
surendravishnoi, line 1474: we should free existing memory for server collation name

surendravishnoi (dismissed):

contrib/babelfishpg_common/src/collation.c
Deepesh125: server_collation_name is static variable.

contrib/babelfishpg_common/sql/upgrades/babelfishpg_common--2.0.0--2.1.0.sql
```diff
@@ -46,5 +46,14 @@ LANGUAGE C IMMUTABLE STRICT PARALLEL SAFE;
 CREATE CAST (sys.VARCHAR AS sys.BIT)
 WITH FUNCTION sys.varchar2bit(sys.VARCHAR) AS IMPLICIT;
 
+/* This helper function would only be useful and strickly be used during 1.x to 2.3 upgrade. */
```
2jungkook: strictly?

contrib/babelfishpg_common/src/collation.c
Deepesh125: Addressed in latest code.

surendravishnoi (dismissed):

contrib/babelfishpg_common/src/collation.c
```diff
@@ -1393,4 +1416,74 @@ get_collation_callbacks(void)
...
+	 * babelfishpg_tsql.restored_server_collation_name.
+	 */
+	if ((babelfish_dump_restore && 
+		 strncmp(babelfish_dump_restore, "on", 2) == 0) &&
+		 babelfish_restored_server_collation_name)
+		return get_collation_oid_internal(babelfish_restored_server_collation_name);
+
+	get_server_collation_oid_internal(false); /* set and cache server_collation_oid */
+
+	if (OidIsValid(server_collation_oid))
```
2jungkook: I think this is unnecessary because we call `get_server_collation_oid_internal` with `missingOk = false`. Can we change this into an assertion?

contrib/babelfishpg_common/src/collation.c
```diff
@@ -1393,4 +1416,74 @@ get_collation_callbacks(void)
...
+}
+
+PG_FUNCTION_INFO_V1(babelfish_update_server_collation_name);
+
+/*
+ * babelfish_update_server_collation_name - corresponding to sys.babelfish_update_server_collation_name() function
+ * which would be available and strickly be used during 1.x to 2.3 upgrade.
+ */
+Datum
+babelfish_update_server_collation_name(PG_FUNCTION_ARGS)
```
2jungkook, line 1453: Can we also clear `babelfish_restored_server_collation_name` here?

contrib/babelfishpg_common/src/collation.h
```diff
@@ -5,6 +5,14 @@
...
+
+/* Dump and Restore */
+extern char *babelfish_restored_server_collation_name;
+extern char *babelfish_restored_default_locale;
```
2jungkook: Let's try to clear the restored values when those are not used anymore.

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.2.0--2.3.0.sql
```diff
@@ -4359,10 +8015,26 @@ AS $BODY$ BEGIN
...
+-- And reset babelfishpg_tsql.restored_server_collation_name and babelfishpg_tsql.restored_default_locale GUC
+do
+language plpgsql
+$$
+    declare
+        query text;
+    begin
+        query := pg_catalog.format('alter database %s reset babelfishpg_tsql.restored_server_collation_name', CURRENT_DATABASE());
+        execute query;
+        query := pg_catalog.format('alter database %s reset babelfishpg_tsql.restored_default_locale', CURRENT_DATABASE());
```
2jungkook: Let's avoid setting/resetting restored GUCs other than `2.0.0--2.1.0` upgrade scripts.

contrib/babelfishpg_tsql/sql/sys_function_helpers.sql
```diff
@@ -10261,9 +10263,11 @@ CREATE OR REPLACE FUNCTION sys.babelfish_pltsql_get_last_stmt_handle()
 RETURNS INT
 AS 'babelfishpg_tsql', 'pltsql_get_last_stmt_handle' LANGUAGE C;
 
+/*
 CREATE OR REPLACE FUNCTION sys.get_babel_server_collation_oid() RETURNS OID
 LANGUAGE C
 AS 'babelfishpg_tsql', 'get_server_collation_oid';
+*/
```
2jungkook: Let's remove this.

Deepesh125: Will address in new rev.

contrib/babelfishpg_common/src/collation.c
Deepesh125: will fix in new rev.

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.2.0--2.3.0.sql
Deepesh125: will fix in new rev

contrib/babelfishpg_common/src/collation.h
Deepesh125: It will be reset at the end of 2.0--2.1 script, right?

contrib/babelfishpg_common/src/collation.c
Deepesh125: we are anyway going to clear this at the end of 2.0--2.1 upgrade script so let's not clear this global variable directly which is there to store value of GUC

2jungkook (dismissed):

surendravishnoi (approved):

PR #841, opened 2022-11-16
Add script for Github actions to detect engine branches

AvantikaDG:
The clone_engine_repo script makes sure that github actions that run on extension PRs pick up the correct 
engine branch (branch with the same name if found, else the default dev branch). This enables changes that 
include engine modifications to be tested by the github actions seamlessly. These scripts already exist in BABEL_2_X_DEV. This commit simply adds the scripts to BABEL_1_X_DEV. 

Task: BABEL-3709


.github/scripts/clone_engine_repo.conf
```diff
@@ -0,0 +1 @@
+DEFAULT_BRANCH=
```
nasbyj: You need to set this to something sane, otherwise it'll default to whatever the github default branch is (currently 2_X_DEV). Take a look in one of the older 2.x branches, like 2.1, to see what it's set to there.

nasbyj (changes requested):

nasbyj (approved):

PR #855, opened 2022-11-17
Function sys.tsql_get_functiondef should return NULL for Table Valued Function

shalinilohia50:
### Description

Function `sys.tsql_get_functiondef` should return NULL for Table Valued Function

### Issues Resolved

Function `sys.tsql_get_functiondef` was giving an error `invalid parameter mode 't'` for Table Valued Function. 

The following views:

1. information_schema.routines (information_schema_tsql.routines on PG side)
2. sys.sql_modules
3. sys.all_sql_modules

are impacted by this issue as they all use the function and would result in the "invalid parameter mode 't'" error when it queries a table-valued function. 

* **Use case based - When there is table valued function, sys.tsql_get_functiondef returns NULL.** 
```
1> CREATE FUNCTION routines_fc7()
2> RETURNS @myRetTable table (a int PRIMARY KEY)
3> as
4> begin
5> INSERT INTO @myRetTable VALUES (1)
6> return
7> end
8> go
1> SELECT sys.tsql_get_functiondef(OBJECT_ID('routines_fc7'));
2> go
tsql_get_functiondef                                                                                                                                                                                                                                            
-----------------------------
NULL   

-- It shows PG definition because there is a COALESCE function used and TSQL Definition is NULL for TVF.
1> select object_id, definition from all_sql_modules where object_id=OBJECT_ID('routines_fc7');
2> go
object_id definition 
----------------------- -----------------------
19308 CREATE OR REPLACE FUNCTION master_dbo.routines_fc7()
 RETURNS TABLE("@myrettable" "@myrettable_routines_fc7")
 LANGUAGE pltsql
AS '{"version_num": "1", "typmod_array": ["-1", "-1"], "original_probin": ""}', $function$begin
INSERT INTO @myRetTable VALUES (1)
return
end$function$

(1 rows affected)

```
* **Boundary conditions - N/A**
* **Arbitrary inputs - N/A**
* **Negative test cases - N/A**
* **Minor version upgrade tests - N/A**
* **Major version upgrade tests - N/A**
* **Performance tests - N/A**
* **Tooling impact - N/A**
* **Client tests - N/A**


contrib/babelfishpg_tsql/src/pltsql_ruleutils.c
```diff
@@ -800,6 +805,9 @@ tsql_print_function_arguments(StringInfo buf, HeapTuple proctup,
 				modename = "OUT ";
 				isinput = false;
 				break;
+			case PROARGMODE_TABLE:
+				*has_tvp = true;
+				return;
```
rishabhtanwar29: Make sure we release `bbffunctuple` (at line 882) before returning. I would recommend to `break` here and then come out of for loop with another `break` if `has_tvp` is set.

test/JDBC/expected/routines_definition-vu-prepare.out
```diff
@@ -38,3 +38,12 @@ BEGIN
 END;
 go
 
+-- Table valued Function returns NULL
```
rishabhtanwar29: Do we have a test for MSTVF? Better to add it now otherwise.

shalinilohia50: Added tests for MSTVF and Inline TVF.

rishabhtanwar29 (approved):

shardgupta (approved):

PR #859, opened 2022-11-17
Fix pipeline failure for sp_addrole-dep test

Suthapalli-Ramya-satya-vasavi-srija:
### Description
Previously the sp_addrole-dep test was passing locally but failing on remote instances.The reason for this is that the [pg_stat_statements](https://www.postgresql.org/docs/current/pgstatstatements.html) module is not enabled locally, but is enabled on remote instances.
Problem only happens for error cases where error is raised inside a function while doing insert exec on a table variable.
This is not a valid SQL server use case as SQL server does not allow insert exec inside functions but allows only inside procedures. Fixed by removing failing test case.
Signed-off-by: vasavi suthapalli <svasusri@amazon.com>
### Check List

### DIFF
--- a/test/JDBC/expected/Test-sp_addrole-dep-vu-verify.out
+++ b/test/JDBC/expected/Test-sp_addrole-dep-vu-verify.out
@@ -26,12 +26,3 @@ GO
 
 ~~ERROR (Message: The @ownername argument is not yet supported in Babelfish.)~~
 
-
-SELECT dbo.test_sp_addrole_func('sp_addrole_role2', '')
-GO
-~~START~~
-int
-~~ERROR (Code: 33557097)~~
-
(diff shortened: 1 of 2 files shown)


surendravishnoi (commented):
Please add the explanation in description part. Also, remove the guideline desc itself. Add Jira as well.

Suthapalli-Ramya-satya-vasavi-srija:
> Please add the explanation in description part. Also, remove the guideline desc itself. Add Jira as well.

There is no jira for this failure, can I proceed with procedure jira - 3087

priyansx (approved):
LGTM

shalinilohia50:
Please update the JIRA description, clean up unnecessary lines and `Signed-off-by` should be after the description.

surendravishnoi (approved):

PR #896, opened 2022-11-29
Ignore EXECUTE AS CALLER clause in CREATE PROC/FUNCTION/TRIGGER

Suthapalli-Ramya-satya-vasavi-srija:
Signed-off-by: vasavi suthapalli <svasusri@amazon.com>

### Description

1. *What* is the change? This is best described in terms of “Currently, Babelfish does X. With this change it now does Y.” Think of “What *did* it *used* to do?” and “What *does* it do *now*?”
> With the current implementation EXECUTE AS CALLER clause is ignored for PROC/FUNCTION/TRIGGER. The other variations of the WITH EXECUTE AS will continue to raise an error. For TRIGGER with EXECUTE AS CLAUSE syntax is being supported and an error will continue to raise until the time we support it.
> Previously there are many cases where proc/function/trigger use EXECUTE AS clause and all of these currently fail since bbf rejects the WITH EXECUTE AS clause and in case of triggers there was syntax error is being thrown for EXECUTE AS clause. Since CREATE... WITH EXECUTE AS CALLER is the same as CREATE ... without the WITH EXECUTE AS clause, there is no reason to reject WITH EXECUTE AS CALLER.

### Issues Resolved
Task:  BABEL-3661

* **Use case based -**
Created a test file for test execute_as_caller clause for function/procedure/trigger

* **Boundary conditions -**
Tested whether the CREATE...WITH EXECUTE AS CALLER is successfully able to create PROC/FUNC/TRIGGER and able to throw error for other variations of EXECUTE AS clause

* **Arbitrary inputs -**
Created function with different return types

* **Negative test cases -**
Tests whether the bbf is raising error for other variations of execute_As_clause

NA

NA

* **Performance tests -**
NA

* **Tooling impact -**
NA

* **Client tests -**
NA

NOTE: 
1. Couldn't add tests related to table_clr as CLR is not yet supported in bbf.
2. Couldn't add test cases related to DDL triggers as they were currently not supported in BBF.
```
1> CREATE TRIGGER trg_index_changes ON DATABASE FOR  CREATE_INDEX, DROP_INDEX as begin insert into index_logs(a,b) values (1,2); end
2> go
'DDL trigger' is not currently supported in Babelfish
```
### Check List


contrib/babelfishpg_tsql/src/tsqlIface.cpp
```diff
@@ -2679,6 +2679,8 @@ rewriteBatchLevelStatement(
 				auto commas = cctx->COMMA();
 				GetTokenFunc<TSqlParser::Function_optionContext*> getToken = [](TSqlParser::Function_optionContext* o) { return o->SCHEMABINDING(); };
 				bool all_removed = removeTokenFromOptionList(expr, options, commas, ctx, getToken);
+				getToken = [](TSqlParser::Function_optionContext* o) { return o->execute_as_clause()->CALLER(); };
```
shalinilohia50: This condition will crash the server in few cases. 
First add a check to verify if `execute_as_clause()` exists, then `{ return o->execute_as_clause()->CALLER(); }` makes sense.

contrib/babelfishpg_tsql/src/tsqlIface.cpp
```diff
@@ -2679,6 +2679,8 @@ rewriteBatchLevelStatement(
...
+				all_removed = removeTokenFromOptionList(expr, options, commas, ctx, getToken);
```
shalinilohia50: You would also need to add the appropriate checks inside the following conditions 
1. `ctx->create_or_alter_function()->func_body_returns_table()`
2. `if (ctx->create_or_alter_function()->func_body_returns_scalar())`

contrib/babelfishpg_tsql/src/tsqlIface.cpp
Suthapalli-Ramya-satya-vasavi-srija: Fixed in next revision by adding a pre-check for execute_as_clause().

test/JDBC/expected/BABEL-EXECUTE_AS_CALLER.out
```diff
@@ -0,0 +1,121 @@
+-- EXECUTE AS CALLER
+-- For created functions/procedure/trigger the EXECUTE AS CALLER should be work without raising an error
```
shalinilohia50: Modify the comment:
```
Create function/procedure/trigger with `EXECUTE AS CALLER` clause should not raise an error.
```

Suthapalli-Ramya-satya-vasavi-srija: Modified the comment in next revision

contrib/babelfishpg_tsql/src/tsqlUnsupportedFeatureHandler.cpp
```diff
@@ -301,7 +301,11 @@ antlrcpp::Any TsqlUnsupportedFeatureHandlerImpl::visitCreate_or_alter_function(T
 		else if (option->NATIVE_COMPILATION())
 			handle(INSTR_UNSUPPORTED_TSQL_ALTER_FUNCTION_NATIVE_COMPILATION_OPTION, option->NATIVE_COMPILATION());
 		else if (option->execute_as_clause())
-			handle(INSTR_UNSUPPORTED_TSQL_EXECUTE_AS_STMT, "EXECUTE AS", getLineAndPos(option->execute_as_clause()));
+		{
+			auto exec_as = option->execute_as_clause();
+			if (!exec_as->CALLER())
+				handle(INSTR_UNSUPPORTED_TSQL_EXECUTE_AS_STMT, "EXECUTE AS", getLineAndPos(option->execute_as_clause()));
```
kuntalghosh: We should improve the error message by saying which EXECUTE AS is not supported. For example,
`ERROR (Message: 'EXECUTE AS SELF|OWNER|<user>|<login>' is not currently supported in Babelfish)`

Suthapalli-Ramya-satya-vasavi-srija: Fixed in next revision by throwing better error message for execute_as_clause statements in pFUNC/PROC/TRIGGER.

shalinilohia50 (dismissed):

contrib/babelfishpg_tsql/src/tsqlIface.cpp
```diff
@@ -2681,6 +2681,19 @@ rewriteBatchLevelStatement(
...
+				if(execute_as_clause_check)
+				{
+					getToken = [](TSqlParser::Function_optionContext* o) { return o->execute_as_clause()->CALLER(); };
+					all_removed = removeTokenFromOptionList(expr, options, commas, ctx, getToken);
+					if (all_removed)
+						removeTokenStringFromQuery(expr, cctx->WITH(), ctx);
+				}
 			}
 		}
 		else if (ctx->create_or_alter_function()->func_body_returns_table()) /* CREATE FUNCTION ... RETURNS TABLE RETURN SELECT ... */
```
HarshLunagariya: Why are we not discarding the EXECUTE AS CALLER clause for following conditions?
- CREATE FUNCTION ... RETURNS TABLE RETURN SELECT
- CREATE FUNCTON ... RETURNS INT RETURN
- CREATE OR ALTER PROCEDURE

contrib/babelfishpg_tsql/src/tsqlIface.cpp
```diff
@@ -2681,6 +2681,19 @@ rewriteBatchLevelStatement(
...
+				bool execute_as_clause_check = false;
+				for (auto option : options)
+					if (option->execute_as_clause())
+						execute_as_clause_check = true;
+				if(execute_as_clause_check)
+				{
+					getToken = [](TSqlParser::Function_optionContext* o) { return o->execute_as_clause()->CALLER(); };
+					all_removed = removeTokenFromOptionList(expr, options, commas, ctx, getToken);
+					if (all_removed)
+						removeTokenStringFromQuery(expr, cctx->WITH(), ctx);
```
HarshLunagariya: We can remove following redundant code(after removing SCHEMABINDING and EXECUTE AS options, we can remove WITH() after removing both options at the end.):
```
if (all_removed)
					removeTokenStringFromQuery(expr, cctx->WITH(), ctx);
```
instead of that after 
```
if(execute_as_clause_check)
{
...
}
<we can remove WITH() clause here>
```

HarshLunagariya (changes requested):

contrib/babelfishpg_tsql/src/tsqlIface.cpp
shalinilohia50: Please check the corresponding tests added for the conditions mentioned above. 
Without making these explicit changes for the conditions, we are getting the desired output [EXECUTE AS CALLER is ignored, the other EXECUTE AS <> are throwing errors].

test/JDBC/input/BABEL-EXECUTE_AS_CALLER.sql
```diff
@@ -0,0 +1,97 @@
+-- EXECUTE AS CALLER
```
xhfanhe: Better as well to change the test into upgrade test format. 
Team is considering deprecating all normal JDBC tests and only keep upgrade tests. Even if there's no dependency needs to be tested, we can still run the test in upgrade scenario (because it would use different settings/configs than normal JDBC env)

contrib/babelfishpg_tsql/src/tsqlIface.cpp
Suthapalli-Ramya-satya-vasavi-srija: Fixed in next revision by replacing remove WITH() after removing both schemabinding and execute_As_clause.

contrib/babelfishpg_tsql/src/tsqlIface.cpp
HarshLunagariya: Vasavi mentioned that if we are adding this logic there then we are observing crash that means we are hitting that code that means it is not stale and we should see why above behaviour is observed.

test/JDBC/input/BABEL-EXECUTE_AS_CALLER.sql
Suthapalli-Ramya-satya-vasavi-srija: Fixed in next revision by changing the test to upgrade tests and added them in all version's upgrade schedule.

contrib/babelfishpg_tsql/src/tsqlIface.cpp
```diff
@@ -2679,6 +2679,16 @@ rewriteBatchLevelStatement(
...
 				bool all_removed = removeTokenFromOptionList(expr, options, commas, ctx, getToken);
+				/* Check if execute_as_clause is present, and if is present check for caller clause */
+				bool execute_as_clause_check = false;
+				for (auto option : options)
+					if (option->execute_as_clause())
+						execute_as_clause_check = true;
+				if(execute_as_clause_check)
+				{
+					getToken = [](TSqlParser::Function_optionContext* o) { return o->execute_as_clause()->CALLER(); };
+					all_removed = removeTokenFromOptionList(expr, options, commas, ctx, getToken);
```
KushaalShroff: in "removeTokenFromOptionList"
```
bool
removeTokenFromOptionList(PLtsql_expr *expr, std::vector<T>& options, std::vector<antlr4::tree::TerminalNode *>& commas, antlr4::ParserRuleContext *ctx, GetTokenFunc<T> getTokenFunc)
{
	Assert(options.size() -1 == commas.size());

	bool all_removed = true;
	bool comma_carry_over = false;

	for (size_t i=0; i<options.size(); ++i)
	{
		if (getTokenFunc(options[i]))
```

getTokenFunc(options[i]) is called for all of the options. Now we have passed o->execute_as_clause()->CALLER() as getTokenFunc but not all options (o) will have a NOT NULL execute_as_clause variable. Hence, any option other than "execute as clause" option will lead to seg fault (o->NULL->CALLER())

Suthapalli-Ramya-satya-vasavi-srija: Fixed in next revision by adding a pre check for execute_as_clause to avoid seg fault for other options.

xhfanhe (dismissed):

shalinilohia50 (dismissed):

HarshLunagariya (dismissed):

test/JDBC/expected/BABEL-EXECUTE_AS_CALLER-vu-verify.out
```diff
@@ -0,0 +1,260 @@
+-- EXECUTE AS CALLER
```
surendravishnoi: if we are creating all object in verify script only, what is the benefit of adding this test to all upgrade paths. Please add it to latest only.

contrib/babelfishpg_tsql/src/tsqlIface.cpp
```diff
@@ -2677,7 +2677,11 @@ rewriteBatchLevelStatement(
 			{
 				auto options = cctx->function_option();
 				auto commas = cctx->COMMA();
-				GetTokenFunc<TSqlParser::Function_optionContext*> getToken = [](TSqlParser::Function_optionContext* o) { return o->SCHEMABINDING(); };
+				GetTokenFunc<TSqlParser::Function_optionContext*> getToken = [](TSqlParser::Function_optionContext* o) {
+					if (o->execute_as_clause())
+						return o->execute_as_clause()->CALLER();
+					return o->SCHEMABINDING();
+				};
```
2jungkook, line 2684: Many code blocks under `if (cctx->WITH())` are duplicate. Can we consider using a template function?

contrib/babelfishpg_tsql/src/tsqlIface.cpp
```diff
@@ -2690,7 +2694,11 @@ rewriteBatchLevelStatement(
 			{
 				auto options = cctx->function_option();
 				auto commas = cctx->COMMA();
-				GetTokenFunc<TSqlParser::Function_optionContext*> getToken = [](TSqlParser::Function_optionContext* o) { return o->SCHEMABINDING(); };
+				GetTokenFunc<TSqlParser::Function_optionContext*> getToken = [](TSqlParser::Function_optionContext* o) {
```
2jungkook, line 2697: The above comment `CREATE FUNCTION ... RETURNS TABLE RETURN SELECT` looks incorrect. Please fix that as well for other's convenience.

contrib/babelfishpg_tsql/src/tsqlIface.cpp
```diff
@@ -2716,7 +2728,11 @@ rewriteBatchLevelStatement(
...
```
2jungkook, line 2731: Do we have a test case for this?

contrib/babelfishpg_tsql/src/tsqlIface.cpp
```diff
@@ -2767,7 +2791,11 @@ rewriteBatchLevelStatement(
 		{
 			auto options = cctx->trigger_option();
 			auto commas = cctx->COMMA();
-			GetTokenFunc<TSqlParser::Trigger_optionContext*> getToken = [](TSqlParser::Trigger_optionContext* o) { return o->SCHEMABINDING(); };
+			GetTokenFunc<TSqlParser::Trigger_optionContext*> getToken = [](TSqlParser::Trigger_optionContext* o) {
```
2jungkook, line 2794: Do we have test cases for both dml and ddl triggers?

Suthapalli-Ramya-satya-vasavi-srija: Test case for dml triggers were included, but for ddl isn't included as DDL is not currently supported in BBF

contrib/babelfishpg_tsql/src/tsqlIface.cpp
Suthapalli-Ramya-satya-vasavi-srija: Modified in next revision by changing the comment to `CREATE FUNCTION ... RETURNS TABLE(COLUMN_DEFINITION) AS BEGIN ... END` as its more suited for functions of return table which was declared as
```
func_body_returns_table
    : RETURNS LOCAL_ID table_type_definition
        (WITH function_option (COMMA function_option)*)?
        AS?
        BEGIN sql_clauses* SEMI? END
    ;
```

xhfanhe (approved):

PR #899, opened 2022-11-29
Support APP_NAME() Transact-SQL function

LareinaWei:
Return the application_name GUC from current_settings().

Task: BABEL-797
Signed-off-by: Yuhao Wei <yuhaowei@amazon.com>

### Description

Currently, babelfish does not support APP_NAME() Transact-SQL function. Since `application_name` has already been set as a pg guc when client connects, we return the guc as the result.

### Issues Resolved

BABEL-797

* **Use case based -**
Added a new test `functions/app_name`

* **Boundary conditions -**
N/A

* **Arbitrary inputs -**
N/A

* **Negative test cases -**
N/A

included

included

* **Performance tests -**
N/A

* **Tooling impact -**
N/A

* **Client tests -**
Manually tested with sqlcmd:
```
1> select app_name();
2> go
app_name
SQLCMD
```

### Check List


contrib/babelfishpg_tsql/sql/sys_functions.sql
```diff
@@ -3297,3 +3297,10 @@ END;
 $BODY$
 LANGUAGE plpgsql;
 GRANT EXECUTE ON FUNCTION sys.INDEXPROPERTY(IN object_id INT, IN index_or_statistics_name sys.nvarchar(128),  IN property sys.varchar(128)) TO PUBLIC;
+
+CREATE OR REPLACE FUNCTION sys.APP_NAME() RETURNS SYS.NVARCHAR(128)
+AS
+$$
+    SELECT current_setting('application_name');
+$$
+LANGUAGE sql PARALLEL SAFE IMMUTABLE RETURNS NULL ON NULL INPUT;
```
2jungkook: I feel like `IMMUTABLE` is not correct. Please refer https://www.postgresql.org/docs/current/xfunc-volatility.html

test/JDBC/expected/app_name-vu-verify.out
```diff
@@ -0,0 +1,15 @@
...
+~~END~~
+
+
+EXEC app_name_vu_prepare_p1
+GO
+~~START~~
+nvarchar
+Microsoft JDBC Driver for SQL Server
+~~END~~
+
```
2jungkook, line 15: In the verify file, let's add one more case:
```
-- change app_name value
-- execute app_name_vu_prepare_v1 and app_name_vu_prepare_p1 again
```

contrib/babelfishpg_tsql/sql/sys_functions.sql
kuntalghosh: Yes, it should be STABLE.
In SQL Server, application name can't be changed in a session. However, it can change across session. If you declare the function as IMMUTABLE, it'll cache the value in some cases and produce incorrect result. In fact, it can persist the value in some cases as well.

For T-SQL functions, the easiest way to check whether a function can be immutable or not is by creating a persisted computed column using the function in the expression.

Consider the following scenario in SQL Server:
```
1> SET QUOTED_IDENTIFIER ON
2> GO
1> CREATE TABLE t1(a int, b as APP_NAME() PERSISTED)
2> GO
Msg 4936, Level 16, State 1, Server EC2AMAZ-R3QO7RM, Line 1
Computed column 'b' in table 't1' cannot be persisted because the column is non-deterministic.
```

It doesn't allow persisted column with APP_NAME() because the value will change across different connections. So, it will not be immutable or deterministic and the output differs based on the state of the database. Hence, in this case the function should be declared as STABLE.

LareinaWei: > Yes, it should be STABLE. In SQL Server, application name can't be changed in a session. However, it can change across session. If you declare the function as IMMUTABLE, it'll cache the value in some cases and produce incorrect result. In fact, it can persist the value in some cases as well.
> 
> For T-SQL functions, the easiest way to check whether a function can be immutable or not is by creating a persisted computed column using the function in the expression.
> 
> Consider the following scenario in SQL Server:
> 
> ```
> 1> SET QUOTED_IDENTIFIER ON
> 2> GO
> 1> CREATE TABLE t1(a int, b as APP_NAME() PERSISTED)
> 2> GO
> Msg 4936, Level 16, State 1, Server EC2AMAZ-R3QO7RM, Line 1
> Computed column 'b' in table 't1' cannot be persisted because the column is non-deterministic.
> ```
> 
> It doesn't allow persisted column with APP_NAME() because the value will change across different connections. So, it will not be immutable or deterministic and the output differs based on the state of the database. Hence, in this case the function should be declared as STABLE.

That makes a lot of sense. Thanks for the clarification!

test/JDBC/upgrade/latest/schedule
```diff
@@ -339,3 +339,4 @@ BABEL-CHECK-CONSTRAINT
 BABEL-3640
 sys_server_principals_dep
 sys_database_principals_dep
+app_name
```
surendravishnoi, line 358: Add test to minor version update as well, 14.6 to latest

2jungkook: In 14.6, the preparation script cannot work because `app_name()` is added in 14.7. mVU (Minor Version Upgrade) will be tested by https://github.com/babelfish-for-postgresql/babelfish_extensions/blob/BABEL_2_X_DEV/.github/workflows/minor-version-upgrade.yml

2jungkook (dismissed):

2jungkook (approved):

PR #910, opened 2022-12-06
Convert babelfishpg_tsql to not use direct links to babelfishpg_common

rishabhtanwar29:
### Description
Historically, we have referenced functions from `babelfishpg_common` directly
into `babelfishpg_tsql` extension using `extern`keyword. Due to this `babelfishpg_tsql`
library couldn't be loaded independently (without explicitly loading `babelfishpg_common`
library first) due to unresolved/undefined symbols. To overcome this commit 06fdc75bfbb9ef8be16041ea052526ff761bea87
linked `babelfishpg_common` to `babelfishpg_tsql` with a direct dependency.

The direct linking fix mentioned above has it's own drawbacks and causes MVU failures
due to incorrect library linking between source and target data directories during MVU.
Instead, let's remove these `extern` references (unresolved symbols) and use function
pointers that are manually filled in by the module's _PG_init function and all are clubbed
together into one `rendezvous` variable called `common_utility_plugin`.
Additionally, change all occurrences of PG directory to `psql` so that directory names
do not contain `postgres` or `pgsql` keywords.

**Note**: I have temporarily disabled all upgrade workflows (except latest->latest) since
MVU to PG 15 is blocked on BABEL-3801 (which will be fixed soon). These workflows
will be enabled back with the fix for BABEL-3801 but I have manually tested these changes
and verified that MVU works fine.

Task: BABEL-3758

* **Use case based -**

* **Boundary conditions -**

* **Arbitrary inputs -**

* **Negative test cases -**

* **Performance tests -**

* **Tooling impact -**

* **Client tests -**

### Check List


.github/composite-actions/build-extensions/action.yml
```diff
@@ -3,7 +3,7 @@ inputs:
   install_dir:
     description: 'Engine install directory'
     required: no
-    default: postgres
+    default: psql
```
surendravishnoi, line 6: why is this change required? if keywords are issue, it would be better to use something other than psql

.github/workflows/major-version-upgrade.yml
```diff
@@ -1,13 +1,13 @@
...
 jobs:
   run-babelfish-mvu-tests:
     env:
-      OLD_INSTALL_DIR: postgres13
-      NEW_INSTALL_DIR: postgres14
-      ENGINE_BRANCH_FROM: BABEL_1_X_DEV__PG_13_X
-      EXTENSION_BRANCH_FROM: BABEL_1_X_DEV
+      OLD_INSTALL_DIR: psql_source
+      NEW_INSTALL_DIR: psql_target
+      ENGINE_BRANCH_FROM: BABEL_2_3_STABLE__PG_14_6
```
surendravishnoi, line 9: why is this 14_6? is it from 14_6 to 15 latest for empty database?

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -155,7 +154,7 @@ version(PG_FUNCTION_ARGS)
 	 * TODO: Return Build number with version string as well.
 	 */
 
-	info = tsql_varchar_input(temp.data, temp.len, -1);
+	info = (*common_utility_plugin_ptr->tsql_varchar_input)(temp.data, temp.len, -1);
```
Deepesh125, line 157: how are we ensuring that common_utility_plugin_ptr would always have initialised at this point?

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -236,7 +235,7 @@ InstallExtendedHooks(void)
 	logicalrep_modify_slot_hook = logicalrep_modify_slot;
 
 	prev_is_tsql_rowversion_or_timestamp_datatype_hook = is_tsql_rowversion_or_timestamp_datatype_hook;
-	is_tsql_rowversion_or_timestamp_datatype_hook = is_tsql_rowversion_or_timestamp_datatype;
+	is_tsql_rowversion_or_timestamp_datatype_hook = common_utility_plugin_ptr->is_tsql_rowversion_or_timestamp_datatype;
```
Deepesh125, line 240: we should have init'd this hook in common extension side.

contrib/babelfishpg_common/src/datetime.h
```diff
@@ -8,6 +8,10 @@
 #ifndef PLTSQL_DATETIME_H
 #define PLTSQL_DATETIME_H
 
+#include "postgres.h"
+
+#include "datatype/timestamp.h"
```
surendravishnoi, line 11: avoid including headers in headers, rather include in C files wherever needed (OK for cases where header file itself needs the def)

contrib/babelfishpg_common/src/sqlvariant.h
```diff
@@ -1,3 +1,10 @@
+#ifndef SQLVARIANT_H
+#define SQLVARIANT_H
+
+#include "postgres.h"
```
surendravishnoi: avoid headers inclusion in header, include in C files wherever needed

contrib/babelfishpg_tsql/runtime/functions.c
rishabhtanwar29: It is getting inited in `_pg_init` so it will always be initialized as we have an explicit load_library('babelfishpg_common') in `_pg_init`. For reference see [here](https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/910/files#diff-ce1101bab57d0e7f8b36f0df103ed3ae49f4b488e473ac87bf2320c32f2697e1R3348).

contrib/babelfishpg_tsql/src/pltsql_utils.c
```diff
@@ -889,3 +890,19 @@ get_pltsql_function_signature(PG_FUNCTION_ARGS)
...
+	{
+		common_utility_plugin **utility_plugin;
+		utility_plugin = (common_utility_plugin **) find_rendezvous_variable("common_utility_plugin"); 
+		common_utility_plugin_ptr = *utility_plugin;
+
+		/* common_utility_plugin_ptr is still not initialised */
+		if (!common_utility_plugin_ptr)
+			ereport(ERROR,
+				(errcode(ERRCODE_INTERNAL_ERROR),
+				 errmsg("common utility plugin pointer is not initialised properly.")));
```
surendravishnoi: Give a better error like "Failed to find common utility plugin"

contrib/babelfishpg_tsql/src/hooks.c
rishabhtanwar29: We will not get anything extra even if we do that since the specific plugin (`is_tsql_rowversion_or_timestamp_datatype`) is anyway needed in tsql extensions in other places as well so it's not like we are reducing one function from the structure. But I will address this separately.

contrib/babelfishpg_tsql/src/pltsql.h
```diff
@@ -34,6 +34,7 @@
 
 #include "dynavec.h"
 #include "dynastack.h"
+#include "../../contrib/babelfishpg_common/src/babelfishpg_common.h"
```
surendravishnoi, line 37: why is this required? We already have callbacks from TSQL to common (for collation etc.), how do they work?

.github/composite-actions/build-extensions/action.yml
rishabhtanwar29: Problematic keywords are `postgres` and `pgsql` only so `psql` will work fine but let me know if you have any preference.

.github/workflows/major-version-upgrade.yml
rishabhtanwar29: As we are not supporting MMVU so I am changing it as 14_6 to 15 latest MVU and yes it is for empty databases.

contrib/babelfishpg_common/src/datetime.h
rishabhtanwar29: It is needed in the header file itself (there is a reference for `Timestamp` in this file).

Deepesh125 (dismissed):

contrib/babelfishpg_tsql/src/pltsql.h
rishabhtanwar29: This is needed so that we don't have to declare the rendezvous variable structure at two places. Other examples [here](https://github.com/babelfish-for-postgresql/babelfish_extensions/blob/BABEL_3_X_DEV/contrib/babelfishpg_tds/src/include/tds_int.h#L37). For collation callbacks, we have declared the structure both in tsql and common extensions which is just duplicate code so I think it is better to keep the structure at one place and use this header include method to reduce duplication of the code.

contrib/babelfishpg_tsql/src/pltsql_utils.c
rishabhtanwar29: Updated!

contrib/babelfishpg_tsql/src/pltsql.h
surendravishnoi: fine for now but we need to think more. We have taken care of link time dependency but added a compile time dependency from TSQL extension to common extension. I guess such exchange should pass thorough engine as that is common to all extensions.

surendravishnoi (dismissed):

Deepesh125 (dismissed):

Deepesh125 (dismissed):

Deepesh125 (approved):

PR #952, opened 2022-12-20
Re-implement object_id() sql function to C type function to improve performance

sumitj824:
### Description
To improve the performance, this commit contains the re-implementation of OBJECT_ID sql function to C type function. This commit also add checks on user's permission of object, support cross database lookup and also fixed the expected behaviour in some scenarios.

### Issues Resolved

[BABEL-3798]

BABEL_3_X_DEV PR : https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/1052

* **Use case based -** Yes

* **Arbitrary inputs -** Yes

* **Performance tests -** N/A

* **Tooling impact -** N/A

* **Client tests -** N/A

### Check List


contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -928,6 +929,173 @@ checksum(PG_FUNCTION_ARGS)
...
+		PG_RETURN_NULL();
+
+	if (!strcmp(db_name, ""))
+		db_name = get_cur_db_name();
+
+	/* check if looking in current database, if not then return null value */
+	if (strcmp(db_name, get_cur_db_name()) && strcmp(db_name, "tempdb"))
+		ereport(ERROR,
+				(errcode(ERRCODE_UNDEFINED_DATABASE),
+				 errmsg("Can only do lookup in current database.")));
```
HarshLunagariya: - Is this according to TSQL behaviour? 
- Can you please check the cross-db behaviour of object_id()?
- Also check if output depends on the user's permission on the objects or not.

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -928,6 +929,173 @@ checksum(PG_FUNCTION_ARGS)
...
+		/* find the default schema for current user */
+		const char *user = get_user_for_database(db_name);
+		schema_name = get_authid_user_ext_schema_name((const char *) db_name, user);
+	}
+
+	/* get physical schema name from logical schema name */
+	physical_schema_name = get_physical_schema_name(db_name, schema_name);
+
+	/* get schema oid from physical schema name */
+	schema_oid = LookupExplicitNamespace(physical_schema_name, true);
```
HarshLunagariya: Any specific reason to use LookupExplicitNamespace API here? Apart from lookup it is also checks USAGE permission, other throws permission check error. Please add testcase for the same.

contrib/babelfishpg_tsql/src/pltsql_utils.c
```diff
@@ -889,3 +895,230 @@ get_pltsql_function_signature(PG_FUNCTION_ARGS)
...
+		strncpy(res, str, len);
+		res[len] =  '\0';
+		return res;
+	}
+}
+
+/*
+ * Split multiple-part object-name into list, it also extract the delimiter pairs. 
+ */
+List*
```
KushaalShroff: Why do we want a list here, seems unnecessary since you are creating it temporarily just for returning it. We can use an array of size 3 instead.

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -928,6 +929,177 @@ checksum(PG_FUNCTION_ARGS)
...
+			break;
+		case 2:
+			schema_name = (char *) linitial(splited_object_name);
+			object_name = (char *) lsecond(splited_object_name);
+			break;
+		case 3:
+			db_name = (char *) linitial(splited_object_name);
+			schema_name = (char *) lsecond(splited_object_name);
+			object_name = (char *) lthird(splited_object_name);
+			break;
```
KushaalShroff: If split_object_name returned an array of size 3 instead of List of dynamic length the we can skip this switch case

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -928,6 +929,177 @@ checksum(PG_FUNCTION_ARGS)
...
+{	
+	char 			*db_name = "";
+	char 			*schema_name = "";
+	char 			*physical_schema_name;
+	char 			*object_name;
+	bool 			is_temp_object;
+	Oid 			schema_oid;
+	Oid 			result = InvalidOid;
+	int 			i;
+	char 			*input = lowerstr(text_to_cstring(PG_GETARG_TEXT_P(0)));
```
KushaalShroff: Why are we using lowerstr?
String comparison is based on Database Collation.
I think pg_catalog will save the names in lowercase always, can you confirm any case where it can be otherwise?

contrib/babelfishpg_tsql/src/pltsql_utils.c
```diff
@@ -889,3 +895,230 @@ get_pltsql_function_signature(PG_FUNCTION_ARGS)
...
+static char*
+extract_and_remove_delimiter_pair(char *str, int len)
+{	
+	/* extract string from str of given length and remove delimiter pair */
+	if (len >= 2 && ((str[0] == '[' && str[len - 1] == ']') || (str[0] == '"' && str[len - 1] == '"')))
+	{	
+		if(len > 2)
+		{	
+			char *res = (char *) palloc((len - 1) * sizeof(char));
+			strncpy(res, &str[1], len - 2);
```
KushaalShroff: This could lead to security threat since User could send an infinitely long input string.
Even if we have a limitation in the parser(Please confirm if so), we should add a check at the start of object_id C function.
For all 3 parts I believe we can check for appropriate lengths allowed.

contrib/babelfishpg_tsql/src/pltsql_utils.c
```diff
@@ -889,3 +895,230 @@ get_pltsql_function_signature(PG_FUNCTION_ARGS)
...
+			return -1;
+		else
+			return (int)(ptr - str);
+	}
+	
+	return -1;
+}
+
+static char*
+extract_and_remove_delimiter_pair(char *str, int len)
```
KushaalShroff: use delimited_identifiers in the name instead since that makes more sense

contrib/babelfishpg_tsql/src/pltsql_utils.c
```diff
@@ -889,3 +895,230 @@ get_pltsql_function_signature(PG_FUNCTION_ARGS)
...
+		else
+			return "";
+			
+	}
+	else
+	{
+		char *res = (char *) palloc((len + 1) * sizeof(char));
+		strncpy(res, str, len);
+		res[len] =  '\0';
+		return res;
```
KushaalShroff: Why is this required?
Also please use strdup in such cases

contrib/babelfishpg_tsql/src/pltsql_utils.c
```diff
@@ -889,3 +895,230 @@ get_pltsql_function_signature(PG_FUNCTION_ARGS)
...
+ * Returns InvalidOid if there is no such proc.
+ */
+Oid
+tsql_get_temp_object_oid(char *relname)
+{
+	ListCell 		*lc;
+	char			*name;
+	int 			reloid;
+	Oid				result = InvalidOid;
+	List 			*enr_list = get_namedRelList();
```
KushaalShroff: I wonder if we should be checking for enr->md.enrtype ==  ENR_TSQL_TEMP, since we should allow only TSQL temp tables.
But in case of interoperability we will want PG side behaviour as well, please confirm with DBEs

contrib/babelfishpg_tsql/src/pltsql_utils.c
```diff
@@ -889,3 +895,230 @@ get_pltsql_function_signature(PG_FUNCTION_ARGS)
...
+
+	foreach (lc, enr_list)
+	{	
+		reloid = ((EphemeralNamedRelationMetadata)lfirst(lc))->reliddesc;
+		name = ((EphemeralNamedRelationMetadata)lfirst(lc))->name;
+		if (!strcmp(name, relname))
+		{
+			result = reloid;
```
KushaalShroff: There seems to be a function which already does what we are trying do here: get_ENR

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.3.0--2.4.0.sql
```diff
@@ -136,6 +136,11 @@ CREATE OR REPLACE VIEW information_schema_tsql.SEQUENCES AS
 
 GRANT SELECT ON information_schema_tsql.SEQUENCES TO PUBLIC;
 
+CREATE OR REPLACE FUNCTION sys.object_id(IN object_name TEXT, IN object_type char(2) DEFAULT '')
```
KushaalShroff: FYI:
```
' object_name '
Is the object to be used. object_name is either varchar or nvarchar. If object_name is varchar, it is implicitly converted to nvarchar. 

' object_type '
Is the schema-scoped object type. object_type is either varchar or nvarchar. If object_type is varchar, it is implicitly converted to nvarchar. 
```
I think based on this, its hight time we changed the argument types for object_id as well

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -928,6 +929,177 @@ checksum(PG_FUNCTION_ARGS)
...
+			break;
+		default:
+			list_free(splited_object_name);
+			PG_RETURN_NULL();
+	}
+	pfree(input);
+	list_free(splited_object_name);
+
+	/* truncate identifiers if needed */
+	truncate_tsql_identifier(db_name);
```
KushaalShroff, line 1008: Although we are doing this here, there should be a length check before split_object_name for safety

contrib/babelfishpg_tsql/runtime/functions.c
KushaalShroff: Output should depend on user's permission on object. We should check for that

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -928,6 +929,177 @@ checksum(PG_FUNCTION_ARGS)
...
+	truncate_tsql_identifier(schema_name);
+	truncate_tsql_identifier(object_name);
+
+	if (!strcmp(db_name, ""))
+		db_name = get_cur_db_name();
+
+	/* check if looking in current database, if not then return null value */
+	if (strcmp(db_name, get_cur_db_name()) && strcmp(db_name, "tempdb"))
```
KushaalShroff: Shouldnt this check happen before:
```
if (!strcmp(db_name, ""))
		db_name = get_cur_db_name();
```

test/JDBC/input/BABEL_OBJECT_ID-vu-prepare.sql
```diff
@@ -0,0 +1,36 @@
+
+-- To test table, procedure, function
+CREATE TABLE babel_object_id_t1 (a int);
```
KushaalShroff: Where are dependent objects on object_id?
We should add them as well for upgrade tests

test/JDBC/expected/BABEL_OBJECT_ID-vu-verify.out
```diff
@@ -0,0 +1,218 @@
+-- test table, procedure, function, views
+SELECT OBJECT_NAME(OBJECT_ID('babel_object_id_t1'))
```
KushaalShroff: We should add few negative test cases, for NULL input and also for excessively long name such that it gets truncated.

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -928,6 +929,177 @@ checksum(PG_FUNCTION_ARGS)
         PG_RETURN_INT32(result);
 }
 
+Datum
+object_id(PG_FUNCTION_ARGS)
```
KushaalShroff, line 941: Please add a Short description on top for inputs and also that it returns NULL for error scenarios

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -928,6 +929,177 @@ checksum(PG_FUNCTION_ARGS)
...
+				/* search in list of ENRs registered in the current query environment by name */
+				result = tsql_get_temp_object_oid(object_name);
+			}
+			else
+			{
+				/* search in pg_class by name and schema oid */
+				result = get_relname_relid((const char *) object_name, schema_oid); 
+			}
+		}
+		else if (!strcmp(object_type, "C") || !strcmp(object_type, "D") || !strcmp(object_type, "F") ||
```
KushaalShroff: Shouldnt we return NULL for other object types if their names start with '#'?
is_temp_object would be true, but we are ignoring it in further scenarios

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -928,6 +929,177 @@ checksum(PG_FUNCTION_ARGS)
...
+		else if (!strcmp(object_type, "TR") || !strcmp(object_type, "TA"))
+		{
+			/* search in pg_trigger by name */
+			result = tsql_get_trigger_oid(object_name);
+		}
+		else if (!strcmp(object_type, "R") || !strcmp(object_type, "EC") || !strcmp(object_type, "PG") ||
+				!strcmp(object_type, "SN") || !strcmp(object_type, "SQ") || !strcmp(object_type, "TT"))
+		{
+			ereport(ERROR,
+					(errcode(ERRCODE_UNDEFINED_OBJECT),
```
KushaalShroff: Use ERRCODE_FEATURE_NOT_SUPPORTED with "Object type currently unsupported in Babelfish"

contrib/babelfishpg_tsql/src/pltsql_utils.c
```diff
@@ -889,3 +895,230 @@ get_pltsql_function_signature(PG_FUNCTION_ARGS)
...
+}
+
+/*
+ * tsql_get_trigger_oid
+ *		Given name of a trigger, look up the OID.
+ *
+ * Returns InvalidOid if there is no such trigger.
+ */
+Oid
+tsql_get_trigger_oid(char *tgname)
```
KushaalShroff: Why dont we need to check for schema/namespace for triggers?

contrib/babelfishpg_tsql/runtime/functions.c
sumitj824: `split_object_name`  can be used be 4-part name and 2-part name also, I don't think we want to make this function specific for our usecases

contrib/babelfishpg_tsql/runtime/functions.c
sumitj824: If we have not specified db_name in input, we should load the db_name to current db. We can add this check in `else-if ` block such that if we have not specified db_name we will directly load and move.

contrib/babelfishpg_tsql/src/pltsql_utils.c
sumitj824: `split_object_name` can be used be 4-part name and 2-part name also, I don't think we want to make this function specific for our usecases

contrib/babelfishpg_tsql/src/pltsql_utils.c
sumitj824: Function `Datum get_enr_list()` unnecessarily creating result tuple for all rows, It will not be a good idea to use that when we are talking in terms of performance.

contrib/babelfishpg_tsql/src/pltsql_utils.c
```diff
@@ -906,3 +913,232 @@ void init_and_check_common_utility(void)
...
+ * Returns InvalidOid if there is no such proc.
+ */
+Oid
+tsql_get_temp_object_oid(char *relname)
+{
+	ListCell 		*lc;
+	char			*name;
+	int 			reloid;
+	Oid				result = InvalidOid;
+	List 			*enr_list = get_namedRelList();
```
surendravishnoi: do we have any locking to take care of? we are accessing list directly

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -927,6 +928,177 @@ checksum(PG_FUNCTION_ARGS)
...
+	char 			*object_type = text_to_cstring(PG_GETARG_TEXT_P(1));
+	List 			*splited_object_name;
+
+	/* strip trailing whitespace */
+	i = strlen(input);
+	while (i > 0 && isspace((unsigned char) input[i - 1]))
+		input[--i] = '\0';
+	
+	/* resolve the three part name */
+	splited_object_name = split_object_name(input);
```
surendravishnoi, line 987: this looks like a functionality needed else where as well. Could you please search the code if we are splitting three part name at other places as well and reuse that code

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -927,6 +928,177 @@ checksum(PG_FUNCTION_ARGS)
...
+
+	/* truncate identifiers if needed */
+	truncate_tsql_identifier(db_name);
+	truncate_tsql_identifier(schema_name);
+	truncate_tsql_identifier(object_name);
+
+	if (!strcmp(db_name, ""))
+		db_name = get_cur_db_name();
+
+	/* check if looking in current database, if not then return null value */
```
surendravishnoi: comment is not clear, what is the role of tmpdb? also we are throwing error rather than returning NULL

contrib/babelfishpg_tsql/src/pltsql_utils.c
KushaalShroff: Why cant we have NULL valued indexes? Can you explain why will it be specific to usecases?

contrib/babelfishpg_tsql/runtime/functions.c
KushaalShroff: Same as above, use NULL valued indexes for arrays and directly assign the variables here

contrib/babelfishpg_tsql/src/pltsql_utils.c
KushaalShroff: Again, I am specifically talking about get_ENR() function. We cant use that?

contrib/babelfishpg_tsql/runtime/functions.c
sumitj824: Earlier there was function written in sql which was doing the same, I have not found any similar function in our code which can be used.

contrib/babelfishpg_tsql/runtime/functions.c
sumitj824: addressed in the next commit

contrib/babelfishpg_tsql/runtime/functions.c
sumitj824: added check for input in next commit

contrib/babelfishpg_tsql/runtime/functions.c
sumitj824: addressed in the next commit

contrib/babelfishpg_tsql/runtime/functions.c
sumitj824: addressed in the next commit

contrib/babelfishpg_tsql/runtime/functions.c
sumitj824: Modified the logic accordingly, it will improve performance also.

contrib/babelfishpg_tsql/runtime/functions.c
sumitj824: addressed in the next commit

contrib/babelfishpg_tsql/src/pltsql_utils.c
sumitj824: added check in next commit

contrib/babelfishpg_tsql/src/pltsql_utils.c
sumitj824: addressed in next commit

contrib/babelfishpg_tsql/src/pltsql_utils.c
sumitj824: addressed

contrib/babelfishpg_tsql/src/pltsql_utils.c
sumitj824: addressed

contrib/babelfishpg_tsql/src/pltsql_utils.c
sumitj824: addressed

contrib/babelfishpg_tsql/src/pltsql_utils.c
sumitj824: addressed

contrib/babelfishpg_tsql/src/pltsql_utils.c
sumitj824: We don't need any locking while accessing the list, as these list will be created for every session. And in each session we are running queries sequentially so there will not be any cases where it can be modified while reading the list

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -927,6 +928,208 @@ checksum(PG_FUNCTION_ARGS)
...
+		const char *login;
+		const char *user = NULL;
+		int db_id = get_db_id(db_name);
+
+		if (!DbidIsValid(db_id))
+			PG_RETURN_NULL();
+		login = GetUserNameFromId(GetSessionUserId(), false);
+		user = get_authid_user_ext_physical_name(db_name, login);
+		if(!user)
+			PG_RETURN_NULL();
```
HarshLunagariya: Why do we need this logic?
Because it will return NULL(because won't have any entry in user_ext catalog) for 2 valid cases:
1. When login is member of sysadmin, in this case user will be dbo.
2. When that database has guest user enabled, in that case it will by default use guest.

For more info. please refer `Datum has_dbaccess(PG_FUNCTION_ARGS)` which contains logic around above 2 cases.

We can just use loginid in case of cross-db, it will handle all the cases.

HarshLunagariya (changes requested):

test/JDBC/expected/BABEL_OBJECT_ID-vu-verify.out
```diff
@@ -0,0 +1,449 @@
...
+GO
+~~START~~
+varchar
+babel_object_id_master_user1
+~~END~~
+
+
+-- following should return NULL, as user in db corresponding to current login still don't have permission
+SELECT OBJECT_ID('babel_object_id_db..babel_object_id_db_t1')
+GO
```
HarshLunagariya: Please add atleast 1 cross-db test for each object.

test/JDBC/expected/BABEL_OBJECT_ID-vu-verify.out
```diff
@@ -0,0 +1,449 @@
...
+~~START~~
+void
+
+~~END~~
+
+
+-- tsql
+DROP LOGIN babel_object_id_login1;
+GO
```
HarshLunagariya: Please move it to cleanup file. If you move it there then whole psql code block about terminating session is no longer needed.

contrib/babelfishpg_tsql/runtime/functions.c
sumitj824: addressed in next commit

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -927,6 +928,248 @@ checksum(PG_FUNCTION_ARGS)
         PG_RETURN_INT32(result);
 }
 
+/*
+ * object_id
+ * 		Returns the object ID with object name and object type as input where object type is option
```
sumitj824: will change option to optional in next commit

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -927,6 +928,248 @@ checksum(PG_FUNCTION_ARGS)
...
+	if(!PG_ARGISNULL(1))
+		object_type = text_to_cstring(PG_GETARG_TEXT_P(1));
+	
+	/* strip trailing whitespace */
+	i = strlen(input);
+	while (i > 0 && isspace((unsigned char) input[i - 1]))
+		input[--i] = '\0';
+	
+	/* length should be restricted to 4000 */
+	if(strlen(input) > 4000)
```
surendravishnoi: no need to call strlen twice for input

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -927,6 +928,248 @@ checksum(PG_FUNCTION_ARGS)
...
+	if (!strcmp(db_name, ""))
+		db_name = get_cur_db_name();
+	else if (strcmp(db_name, get_cur_db_name()) && strcmp(db_name, "tempdb"))
+	{
+		/* cross database lookup */
+		int db_id = get_db_id(db_name);
+		if (!DbidIsValid(db_id))
+		{
+			pfree(db_name);
+			pfree(object_name);
```
surendravishnoi, line 1022: what about scheme_name?

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -927,6 +928,248 @@ checksum(PG_FUNCTION_ARGS)
...
+	/* get physical schema name from logical schema name */
+	if (!strcmp(schema_name, ""))
+	{	
+		/* find the default schema for current user and get physical schema name */
+		const char *user = get_user_for_database(db_name);
+		const char		*guest_role_name = get_guest_role_name(db_name);
+		if(!user)
+		{	
+			pfree(db_name);
+			pfree(object_name);
```
surendravishnoi, line 1040: schema_name?

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.3.0--2.4.0.sql
```diff
@@ -817,6 +817,16 @@ and
   );
 GRANT SELECT ON sys.types TO PUBLIC;
 
+-- function sys.object_id(object_name, object_type) needs to change input type to VARCHAR
+ALTER FUNCTION sys.object_id RENAME TO object_id_deprecated_in_2_4_0;
```
surendravishnoi, line 821: what is out plan for 3.0 to 3.1? we can do the same thing there but that will result in unnecessary rename and user has to again recreate dependent views. May be we detect that existence of deprecated object and skip recreate in 3.0 to 3.1

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -927,6 +928,248 @@ checksum(PG_FUNCTION_ARGS)
...
+	/* resolve the three part name */
+	splited_object_name = split_object_name(input);
+	db_name = splited_object_name[1];
+	schema_name = splited_object_name[2];
+	object_name = splited_object_name[3];
+
+	/* downcase identifier if needed */
+	if (pltsql_case_insensitive_identifiers)
+	{	
+		db_name = downcase_identifier(db_name, strlen(db_name), false, false);
```
surendravishnoi, line 995: are we making a copy or using the same memory?

sumitj824: we are making copy here, in next commit I have freed the previous memory after use.

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.3.0--2.4.0.sql
sumitj824: Rishabh suggested that `We can check if column datatype of sys.object_id is already varchar , if yes then we skip drop+recreate step`.

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -927,6 +928,263 @@ checksum(PG_FUNCTION_ARGS)
...
+		schema_name = downcase_identifier(schema_name, strlen(schema_name), false, false);
+		object_name = downcase_identifier(object_name, strlen(object_name), false, false);
+	}
+	
+	for (int i = 0; i < 4; i++)
+		pfree(splited_object_name[i]);
```
surendravishnoi: this is incorrect if pltsql_case_insensitive_identifiers is false, we are freeing memory still referred by *_name objects

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -927,6 +928,263 @@ checksum(PG_FUNCTION_ARGS)
...
+			object_type[1] = '\0';
+		}
+	}
+	/* strip trailing whitespace from input */
+	i = strlen(input);
+	while (i > 0 && isspace((unsigned char) input[i - 1]))
+		input[--i] = '\0';
+	
+	/* length should be restricted to 4000 */
+	if (i > 4000)
```
surendravishnoi, line 981: so 4000 after truncating spaces?

sumitj824: yes

contrib/babelfishpg_tsql/runtime/functions.c
sumitj824: Thanks for pointing it.

surendravishnoi (dismissed):

HarshLunagariya (dismissed):

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -927,6 +928,265 @@ checksum(PG_FUNCTION_ARGS)
...
+	char			**splited_object_name;
+	Oid			schema_oid;
+	Oid			user_id = GetUserId();
+	Oid			result = InvalidOid;
+	bool			is_temp_object;
+	int			i;
+
+	if(PG_ARGISNULL(0))
+		PG_RETURN_NULL();
+	else
```
KushaalShroff: I think we do not require the "else", the else keyword can be removed

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -927,6 +928,265 @@ checksum(PG_FUNCTION_ARGS)
...
+	}
+	/* strip trailing whitespace from input */
+	i = strlen(input);
+	while (i > 0 && isspace((unsigned char) input[i - 1]))
+		input[--i] = '\0';
+	
+	/* length should be restricted to 4000 */
+	if (i > 4000)
+		ereport(ERROR,
+			(errcode(ERRCODE_STRING_DATA_RIGHT_TRUNCATION),
```
KushaalShroff: The error seems misleading, is it documented this way?
We should have a message like "value too long for parameter"
Also I think ERRCODE_STRING_DATA_LENGTH_MISMATCH should be used rather than ERRCODE_STRING_DATA_RIGHT_TRUNCATION

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -927,6 +928,265 @@ checksum(PG_FUNCTION_ARGS)
...
+				pfree(object_type);
+			PG_RETURN_NULL();
+		}
+		else if ((guest_role_name && strcmp(user, guest_role_name) == 0))
+		{
+			physical_schema_name = pstrdup(get_dbo_schema_name(db_name));
+		}
+		else
+		{
+			schema_name  = get_authid_user_ext_schema_name((const char *) db_name, user);
```
KushaalShroff, line 1052: We need to free schema_name before new assignment

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -927,6 +928,265 @@ checksum(PG_FUNCTION_ARGS)
...
+	if(!OidIsValid(schema_oid) || pg_namespace_aclcheck(schema_oid, user_id, ACL_USAGE) != ACLCHECK_OK)
+	{
+		pfree(object_name);
+		if (object_type)
+			pfree(object_type);
+		PG_RETURN_NULL();
+	}
+	
+	/* check if looking for temp object */
+	is_temp_object = (object_name[0] == '#'? true: false);
```
KushaalShroff, line 1078: Can we have a test for object_id('')

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -927,6 +928,265 @@ checksum(PG_FUNCTION_ARGS)
...
+
+	if (object_type) /* "object_type" is specified in-argument */
+	{	
+		if (is_temp_object)
+		{
+			if (!strcmp(object_type, "S") || !strcmp(object_type, "U") || !strcmp(object_type, "V") ||
```
KushaalShroff: Shouldnt these be case insensitive comparison?
A better solution would be to downcase-truncate object_type as well

KushaalShroff (changes requested):

KushaalShroff (approved):

PR #965, opened 2022-12-26
Add NULL checks for GUCs in tdsstat_bestart()

thephantomthief:
### Description

In this commit we add NULL checks for various GUCs that are relevant for the TDS statistics collector. Before comparing the GUC values, we first check whether they are NULL or not. If GUC value is NULL, we set its respective entry equal to the boot value explicitly.

Task: BABEL-3357

Currently the only way this crash can be reproduced is if the babelfishpg_tsql extension compilation is successful but the extension creation fails. This fix has been manually tested since such a case cannot be automated in our actions.

* **Use case based -** N/A

* **Arbitrary inputs -** N/A

* **Performance tests -** N/A

* **Tooling impact -** N/A

* **Client tests -** N/A

### Check List

### DIFF
--- a/contrib/babelfishpg_tds/src/backend/tds/tds.c
+++ b/contrib/babelfishpg_tds/src/backend/tds/tds.c
@@ -423,7 +423,6 @@ tdsstat_bestart(void)
 	char *library_name = NULL;
 	char *host_name = NULL;
 	const char *language = NULL;
-	const char *guc_val = NULL;
 
 	/*
 	 * To minimize the time spent modifying the TdsStatus entry, and
@@ -440,7 +439,6 @@ tdsstat_bestart(void)
 		   unvolatize(TdsStatus *, vtdsentry),
 		   sizeof(TdsStatus));
 
-
 	ltdsentry.st_procpid = MyProcPid;
 	ltdsentry.client_version = MyTdsClientVersion;
 	ltdsentry.client_pid = MyTdsClientPid;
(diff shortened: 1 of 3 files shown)


Deepesh125 (dismissed):

contrib/babelfishpg_tds/src/backend/tds/tds.c
```diff
@@ -447,17 +448,39 @@ tdsstat_bestart(void)
...
+
+	guc_val = GetConfigOption("babelfishpg_tsql.textsize", true, true);
+	ltdsentry.textsize = guc_val != NULL ? atoi(guc_val) : 0;
+
+	guc_val = GetConfigOption("babelfishpg_tsql.datefirst", true, true);
+	ltdsentry.datefirst = guc_val != NULL ? atoi(guc_val) : 7;
+
+	guc_val = GetConfigOption("babelfishpg_tsql.lock_timeout", true, true);
+	ltdsentry.lock_timeout = guc_val != NULL ? atoi(guc_val) : -1;
+
```
kuntalghosh: Since we're revisiting this, let's optimize the code path. All these variables are session level variables. So, you'll be able to get the value by accessing the variable directly.
Declare the GUC in the rendezvous variable shared between TSQL and TDS and access it directly.

kuntalghosh (changes requested):

kuntalghosh (approved):

PR #973, opened 2022-12-28
Correct the 'definition' column of sys.check_constraints

shalinilohia50:
### Description
Correct the 'definition' column of sys.check_constraints

### Issues Resolved

SSMS uses the view sys.check_constraints to get the definition of check constraints. 
_Issues to be resolved:_
1. The definition column of sys.check_constraints view uses the PG  function `pg_get_constraintdef` which gives the PG definition not the TSQL definition which is  incorrect. For example: PG definition replaces the datatype datetime2 with "without time zone". 

2. There is an explicit `CAST( %s AS %s)` for all implicit and explicit casts.
 This causes a difference when we recreate the object from the generated DDL scripts. 

Resolution:
1. Changed the `definition` column to use `tsql_get_constraintdef` function which complies with TSQL Syntax.
2. Removed the implicit casts and also some explicit casts (which are fine) which is not affecting the recreation of objects from the generated DDL scripts.

Task: BABEL-3812

* **Use case based - Added the tests, updated the expected output of the existing tests.**

* **Boundary conditions - N/A**

* **Arbitrary inputs - N/A**

* **Negative test cases - N/A**

* **Minor version upgrade tests - Updated the expected output**

* **Major version upgrade tests - Updated the expected output**

* **Performance tests - N/A**

* **Tooling impact - N/A**

* **Client tests - N/A**


test/JDBC/expected/latest__verification_cleanup__14_3__sys-check_constraints-vu-verify.out
```diff
@@ -4,7 +4,7 @@ GO
 nvarchar
 ((sck_date_col IS NOT NULL))
 ((column_a = 0))
-((column_b <> 'wrong'))
+((column_b <> 'wrong' COLLATE "default"))
```
2jungkook, line 7: Can this constraint be re-created from SSMS?

test/JDBC/input/views/sys-check_constraints.sql
```diff
@@ -2,11 +2,12 @@ DROP TABLE IF EXISTS sys_check_constraints
 GO
 
 CREATE TABLE sys_check_constraints (
-	sck_date_col DATETIME CHECK (sck_date_col IS NOT NULL)
+	sck_date_col DATETIME CHECK (sck_date_col IS NOT NULL),
+	sck_datetime_2 datetime2 check(sck_datetime_2 < cast('2020-10-20 09:00:00' as datetime2))
```
2jungkook: Could you move this test to `sys-check_constraints-vu` so that it can be tested by both normal mode and `latest->latest MVU`?

test/JDBC/expected/latest__verification_cleanup__14_3__sys-check_constraints-vu-verify.out
shalinilohia50: Yes, the DDL can be re-created with the same constraint.

test/JDBC/input/views/sys-check_constraints.sql
shalinilohia50: Updated the tests.

xhfanhe (approved):

2jungkook (approved):

jsudrik:
Correct the 'definition' column of sys.check_constraints

SSMS uses the view sys.check_constraints to get the definition of check constraints.
Currently, the definition column of sys.check_constraints view uses the PG function pg_get_constraintdef which gives the PG definition not the TSQL definition which is incorrect.

PR #985, opened 2023-01-03
Handle collation name in exported DDL's CHECK() clause

shalinilohia50:
### Description
When a DDL is exported where an object's CHECK() clause has an explicit COLLATE clause with SQL Server collation, the collation name is converted to its equivalent BBF name while recreating the definition which is valid for BBF but not for SQL Server.

Task: BABEL-3838

### Issues Resolved

1. When a SQL Server collation is specified with an object in the CHECK() clause of a table, the definition will show the SQL Server collation name (there might be a possibility it shows it's closest match) but not it's equivalent Babelfish Collation name which would work on both SQL Server and BBF.
2. When a BBF collation name is provided, it shows it's closed SQL Server collation name if any. Otherwise it shows the same BBF name.

* **Use case based - Updated the existing test cases.**

1. *  When a "default" collation is used.
2. * When a SQL Server collation is used.
3. * When a BBF collation is used. 

* **Boundary conditions -**

* **Arbitrary inputs -**

* **Negative test cases -**

* **Performance tests -**

* **Tooling impact -**

* **Client tests -**


contrib/babelfishpg_common/src/collation.c
```diff
@@ -119,6 +119,34 @@ coll_translate_t coll_translations[] =
 };
 #define TOTAL_COLL_TRANSLATION_COUNT (sizeof(coll_translations)/sizeof(coll_translations[0]))
 
+coll_translate_t coll_translations_tsql[] =
```
Deepesh125: let's give a meaningful name for this reverse translation, reverse_coll_translations. Also, add a comment about purpose of this C table. 

Plus, ensure that this info is sorted based on first column...

contrib/babelfishpg_common/src/collation.c
```diff
@@ -601,6 +629,33 @@ translate_collation(const char *collname, bool check_for_server_collation_name_g
...
+	int compare;
+
+	while (first <= last)
+	{
+		compare = pg_strcasecmp(coll_translations_tsql[middle].from_collname, collname);
+		if (compare < 0)
+			first = middle + 1;
+		else if (compare == 0)
+		{
+			return (char*)(coll_translations_tsql[middle].to_collname);
```
Deepesh125: Let's return const pointer instead.

contrib/babelfishpg_common/src/collation.c
```diff
@@ -601,6 +629,33 @@ translate_collation(const char *collname, bool check_for_server_collation_name_g
...
+		}
+		else
+			last = middle - 1;
+
+		middle = (first + last) / 2;
+	}
+	return res;
```
Deepesh125: we don't need res anywhere except this place, let's return NULL directly from here.

contrib/babelfishpg_common/src/collation.c
```diff
@@ -601,6 +629,33 @@ translate_collation(const char *collname, bool check_for_server_collation_name_g
 	return idx;
 }
 
+/* Translate BBF collation to it's closest SQL Server Collation. */
+char *
+translate_collation_bbf(const char *collname)
```
Deepesh125: translate_bbf_collation_to_tsql_collation makes more sense, right?

contrib/babelfishpg_common/src/collation.c
```diff
@@ -1400,6 +1455,7 @@ get_collation_callbacks(void)
 		collation_callbacks_var.find_cs_as_collation_internal = &find_cs_as_collation;
 		collation_callbacks_var.find_collation_internal = &find_collation;
 		collation_callbacks_var.has_ilike_node = &has_ilike_node;
+		collation_callbacks_var.translate_collation_bbf = &translate_collation_bbf;
```
Deepesh125: We have to initialise it at https://github.com/amazon-aurora/babelfish_extensions/blob/jira-BABEL-3838/contrib/babelfishpg_common/src/collation.c#L27

contrib/babelfishpg_tsql/src/pltsql_ruleutils.c
```diff
@@ -1359,7 +1360,7 @@ get_rule_expr(Node *node, deparse_context *context,
 					appendStringInfoChar(buf, '(');
 				get_rule_expr_paren(arg, context, showimplicit, node);
 				appendStringInfo(buf, " COLLATE %s",
-								 generate_collation_name(collate->collOid));
+								 print_collation_name(collate->collOid));
```
Deepesh125: User better naming convention for this new function.

contrib/babelfishpg_tsql/src/pltsql_ruleutils.c
```diff
@@ -2836,3 +2837,16 @@ tsql_printTypmod(const char *typname, int32 typmod, Oid typmodout)
 	return res;
 }
 
+char*
+print_collation_name(Oid collOid)
+{
+	char* res = NULL;
+	char* translated_res = NULL;
+	res = generate_collation_name(collOid);
+	translated_res = tsql_translate_collation_bbf(res);
```
Deepesh125: if res is null then only try to reverse translate it.

test/JDBC/input/ISC-Check-Constraints-vu-prepare.sql
```diff
@@ -29,8 +29,8 @@ create table test_tsql_cast(
 go
 
 create table test_tsql_collate(
-	c_varchar varchar check(c_varchar <> cast('sflkjasdlkfjf' as varchar(12)) COLLATE bbf_unicode_cp1_ci_as),
-	c_char char check(c_char <> cast('sflkjasdlkfjf' as char(7)) COLLATE bbf_unicode_cp1_ci_as),
+	c_varchar varchar check(c_varchar <> cast('sflkjasdlkfjf' as varchar(12)) COLLATE latin1_general_ci_as),
```
Deepesh125, line 32: Existing test cases are okay but let's add test cases for collation outside of this translation table, for example, japanese_ci_as

Deepesh125 (changes requested):

contrib/babelfishpg_common/src/collation.c
```diff
@@ -119,6 +119,34 @@ coll_translate_t coll_translations[] =
 };
 #define TOTAL_COLL_TRANSLATION_COUNT (sizeof(coll_translations)/sizeof(coll_translations[0]))
 
+coll_translate_t coll_translations_tsql[] =
+{
+	{ "bbf_unicode_bin2", "latin1_general_bin2", 1252 },
```
KushaalShroff, line 133: It makes sense to have this just to make it working with the other servers, but this is against our tenet/policy. The expectation from customer POV is that we get the exact DDL which user used for creation. We our not being transparent here, this should be approved by Senior DBEs once

contrib/babelfishpg_common/src/collation.c
```diff
@@ -601,6 +629,33 @@ translate_collation(const char *collname, bool check_for_server_collation_name_g
 	return idx;
 }
 
+/* Translate BBF collation to it's closest SQL Server Collation. */
```
KushaalShroff: Please replace SQL Server keyword appropriately

contrib/babelfishpg_common/src/collation.c
```diff
@@ -601,6 +629,33 @@ translate_collation(const char *collname, bool check_for_server_collation_name_g
...
+char *
+translate_collation_bbf(const char *collname)
+{
+	char *res = NULL;
+	int first = 0;
+	int last = TOTAL_COLL_TRANSLATION_COUNT_TSQL - 1;
+	int middle = 16; /* optimization: usually it's the default collation. */
```
KushaalShroff: Use TOTAL_COLL_TRANSLATION_COUNT_TSQL/2 instead of hardcoded value here

contrib/babelfishpg_tsql/src/pltsql_ruleutils.c
```diff
@@ -2836,3 +2837,16 @@ tsql_printTypmod(const char *typname, int32 typmod, Oid typmodout)
...
+
+	if (translated_res)
+		return translated_res;
```
KushaalShroff, line 2858: pfree res in this case

contrib/babelfishpg_common/src/collation.c
KushaalShroff: We have seen cases in the past that maintaining self-sorted ordered arrays could lead to issues. If we extend this array to have new translations and the developer doesnt add it in the right order, how can we ensure that the test fails in all cases? Can we add negative test case?

contrib/babelfishpg_common/src/collation.c
Deepesh125: We are choosing fast path here because we are expecting that default TSQL collation would mostly be sql_latin1_general_cp1_ci_as ~ bbf_unicode_cp1_ci_as.

KushaalShroff: Oh I see that now, we should add proper comment in that case. Use macro in this case then

contrib/babelfishpg_tsql/src/pltsql_ruleutils.c
shalinilohia50: When res is not NULL, only then we reverse translate it.

contrib/babelfishpg_common/src/collation.c
shalinilohia50: Talked to Rob about it, we can go with this approach. We need to document this properly.

test/JDBC/input/ISC-Check-Constraints-vu-prepare.sql
shalinilohia50: Done.

contrib/babelfishpg_common/src/collation.c
shalinilohia50: The test would fail in a case where the developer adds a new translation for TSQL collation T which is not in the right order and the original DDL has that particular TSQL collation T, then the exported DDL will not find the translation and give the BBF collation name instead of the TSQL name which will result in test failure.

contrib/babelfishpg_common/src/collation.c
```diff
@@ -119,6 +119,39 @@ coll_translate_t coll_translations[] =
 };
 #define TOTAL_COLL_TRANSLATION_COUNT (sizeof(coll_translations)/sizeof(coll_translations[0]))
 
+/* Collation translation table.  This must be sorted by the first column (BBF collation),
+ * which contains the name of the collation to be translated from.  The second
+ * column (TSQL Collation) is the name of the collation to translate to.
+ */
```
Deepesh125: Follow multiline style of comment,
```
/*
 * Comment 
 */
```

contrib/babelfishpg_common/src/collation.c
```diff
@@ -119,6 +119,39 @@ coll_translate_t coll_translations[] =
...
+	{ "bbf_unicode_cp1_ci_ai", "latin1_general_ci_ai", 1252 },
+	{ "bbf_unicode_cp1_ci_as", "latin1_general_ci_as", 1252 }, /* default */
+	{ "bbf_unicode_cp1_cs_ai", "latin1_general_cs_ai", 1252 },
+	{ "bbf_unicode_cp1_cs_as", "latin1_general_cs_as", 1252 },
+	{ "bbf_unicode_cp874_ci_as", "sql_latin1_general_cp874_ci_as", 874 },
+	{ "bbf_unicode_cp874_cs_as", "sql_latin1_general_cp874_cs_as", 874 },
+	{ "bbf_unicode_pref_cp1_cs_as", "sql_latin1_general_pref_cp1_cs_as", 1252 }
+};
+
+#define TOTAL_COLL_TRANSLATION_COUNT_TSQL (sizeof(reverse_coll_translations)/sizeof(reverse_coll_translations[0]))
```
Deepesh125: use TOTAL_REVERSE_COLL_TRANSLATION_COUNT for readability.

contrib/babelfishpg_tsql/src/pltsql_ruleutils.c
```diff
@@ -2836,3 +2837,20 @@ tsql_printTypmod(const char *typname, int32 typmod, Oid typmodout)
 	return res;
 }
 
+char*
+return_collation_name(Oid collOid)
```
Deepesh125: generate_tsql_collation_name would be better? and also, add a comment to explain the function.

Deepesh125 (changes requested):

contrib/babelfishpg_common/src/collation.c
shalinilohia50: Updated in the next commit

contrib/babelfishpg_tsql/src/pltsql_ruleutils.c
shalinilohia50: Done

contrib/babelfishpg_common/src/collation.c
```diff
@@ -119,6 +119,40 @@ coll_translate_t coll_translations[] =
...
 #define TOTAL_COLL_TRANSLATION_COUNT (sizeof(coll_translations)/sizeof(coll_translations[0]))
 
+/*
+ * Collation translation table.  This must be sorted by the first column (BBF collation),
+ * which contains the name of the collation to be translated from.  The second
+ * column (TSQL Collation) is the name of the collation to translate to.
+ */
+coll_translate_t reverse_coll_translations[] =
+{
+	{ "bbf_unicode_bin2", "latin1_general_bin2", 1252 },
```
surendravishnoi, line 133: keeping a sorted table in code is not a good idea. If someone adds a wrong entry lookup can fail. Given that there are 20 odd entries only, better to do a scan rather than binary search. Add frequently used entries up front in the table as optimization

contrib/babelfishpg_common/src/collation.c
```diff
@@ -119,6 +119,40 @@ coll_translate_t coll_translations[] =
 };
 #define TOTAL_COLL_TRANSLATION_COUNT (sizeof(coll_translations)/sizeof(coll_translations[0]))
 
+/*
+ * Collation translation table.  This must be sorted by the first column (BBF collation),
```
surendravishnoi: have we explored the solution to keep collation name somewhere in catalog and later use during SQL generation

shalinilohia50: Yes, we explored _pg_attribute.attoptions_ to store the column level collation name. But, it is complicated. 
There are few examples below which make it even more complex to handle:
1. When there are collate clause for column and check clause.
```
1> create table t1(A varchar(10) collate japanese_ci_as, check(A <> 'sflkjasdlkfjf' COLLATE sql_latin1_general_cp1_ci_as))
2> go
1> 
```
2. When there are multiple collate clause in the check constraint clause.
```
1> create table t100(a varchar, check(a <> 'sflkjasdlkfjf' COLLATE sql_latin1_general_cp1_ci_as AND a <> 'abc' COLLATE japanese_ci_as))
2> go
1> 
```

Should we go with the current approach of reverse translation for now until we find a good solution to this problem?

contrib/babelfishpg_common/src/collation.c
shalinilohia50: Kept the reverse collation table unsorted and moved the frequently used entries up in the front for optimization.

PR #1015, opened 2023-01-05
Enable TAP tests for Babelfish

kuntalghosh:
In PG, TAP test framework is used to drive tests for backup and restore, replication, etc - anything that can't really be expressed using pg_regress or the isolation test framework.

Basically, it can be used if we want something more than just comparing the expected outputs. It may include:

- Some engine/os configuration changes
- Start or stop the server
- grep log messages from log file and match with expected log file output
- system calls

In Babelfish, we can utilize the same test framework for the following scenarios:

- pg_hba.conf changes with password authentication
- Kerberos authentication by setting up MIT kerberos 
- SSL setup and test encrypted connection
- Different Provisioning scenarios

Task: BABEL-3852
Signed-off-by: Kuntal Ghosh <kuntalgh@amazon.com>

### DIFF
--- a/contrib/babelfishpg_tds/test/README.test
+++ b/contrib/babelfishpg_tds/test/README.test
@@ -1,10 +1,33 @@
-This folder contains TAP tests written for TDS endpoint. To enable the TAP tests,
-the server needs to be configured with --enable-tap-tests -flag.  Additionally,
-to run the Kerberos test, the server needs to be configured with --with-gssapi
-flag.
+This folder contains TAP tests written for TDS endpoint. 
+
+In PG, TAP test framework is used to drive tests for backup and restore,
+replication, etc - anything that can't really be expressed using pg_regress
+or the isolation test framework.
+https://github.com/postgres/postgres/blob/master/src/test/perl/README
+
+Basically, it can be used if we want something more than just comparing the
+expected outputs. It may include:
+
+- Some engine/os configuration changes
+- Start or stop the server
+- grep log messages from log file and match with expected log file output
+- system calls
+
+In Babelfish, we can utilize the same test framework for the following
+scenarios:
+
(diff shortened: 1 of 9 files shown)


contrib/babelfishpg_tds/test/README.test
```diff
@@ -0,0 +1,108 @@
...
+To enable the TAP tests, the server needs to be configured with --enable-tap-tests 
+flag.  Additionally, to run the Kerberos test, the server needs to be configured
+with --with-gssapi flag.
+
+1. Configure engine with --enable-tap-tests and --with-gssapi flag
+2. Build engine and extension
+3. Execute the following command:
+export PG_CONFIG=<engine installation directory>/pg_config
+cd babelfish_extensions/contrib/babelfishpg_tds/
+make installcheck
```
Deepesh125, line 57: Is there way to run particular set of tests?

contrib/babelfishpg_tds/test/README.test
```diff
@@ -0,0 +1,108 @@
...
+order.
+
+Each test script should begin with:
+
+    use strict;
+    use warnings;
+    use PostgreSQL::Test::Cluster;
+    use PostgreSQL::Test::Utils;
+    use Test::More;
+	use TDSNode;
```
Deepesh125: Nitpick: please fix this indentation here and in below section for better readability.

kuntalghosh: Fixed

contrib/babelfishpg_tds/test/README.test
kuntalghosh: Yes. Added a line for the same.

Deepesh125 (approved):

ahmed-shameem (approved):

PR #1018, opened 2023-01-06
Avoid password check if sspi data is present in login request

Deepesh125:
### Description

In T-SQL world, Authentication mechanism would be chosen by the client as opposed to Postgres where authentication method selection would be controlled through pg_hba.conf file on server side.

For example, TDS client such as sqlcmd would not specify any username in login packet when it wants to use Kerberos/NTLM authentication.

So in order to support Kerberos authentication, This commit avoids consulting pg_hba.conf file if sspi data is present in login request during authentication. 

### Issues Resolved
Task: BABEL-3847
Sub-task: BABEL-3849

* **Use case based -**

- Tested these changes manually. We will add more test case in future PRs for AD authentication support. 

* **Boundary conditions -**

* **Arbitrary inputs -**

* **Negative test cases -**

* **Performance tests -**

* **Tooling impact -**

* **Client tests -**

### Check List

### DIFF
--- a/contrib/babelfishpg_tds/src/backend/tds/tdslogin.c
+++ b/contrib/babelfishpg_tds/src/backend/tds/tdslogin.c
@@ -1767,10 +1767,10 @@ TdsClientAuthentication(Port *port)
 			}
 
 			/*
-			* If pg_hba.conf specifies that the entry should be authenticated using
-			* password and the request doesn't contain a password, we should
-			* throw an error.
-			*/
+			 * If pg_hba.conf specifies that the entry should be authenticated using
+			 * password and the request doesn't contain a password, we should
+			 * throw an error.
+			 */
 			if (!loginInfo->password)
 			{
 				char		hostinfo[NI_MAXHOST];
--- a/contrib/babelfishpg_tds/src/backend/tds/tdslogin.c
+++ b/contrib/babelfishpg_tds/src/backend/tds/tdslogin.c
@@ -1423,10 +1423,6 @@ CheckGSSAuth(Port *port)
 										  NULL,
 										  NULL);
 
-		/* gbuf no longer used */
-		pfree(request->sspi);
-		request->sspiLen = 0;
-
 		elog(DEBUG4, "gss_accept_sec_context major: %d, "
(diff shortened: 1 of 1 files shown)


contrib/babelfishpg_tds/src/backend/tds/tdslogin.c
```diff
@@ -1759,29 +1755,39 @@ TdsClientAuthentication(Port *port)
 			break;
 		case uaMD5:
 		case uaPassword:
-			/*
-			 * If pg_hba.conf specifies that the entry should be authenticated using
-			 * password and the request doesn't contain a password, we should
-			 * throw an error.
-			 */
-			if (!loginInfo->password)
+			/* if sspiLen > 0 then GSS auth is already done at this point */
```
kuntalghosh, line 1758: How about the following change only?
```
if (loginInfo->sspiLen > 0)
{
	Assert(loginInfo->sspi);

	/* Cleanup sspi data. */
	pfree(loginInfo->sspi);
	loginInfo->sspiLen = 0;
	break;
}

/*
 * If pg_hba.conf specifies that the entry should be authenticated using
 * password and the request doesn't contain a password, we should
 * throw an error.
 */
if (!loginInfo->password)
...

```
The assert is valid as loginInfo->sspi is not null if sspiLen>0. (Check FetchLoginRequest)

contrib/babelfishpg_tds/src/backend/tds/tdslogin.c
```diff
@@ -1769,15 +1776,15 @@ TdsClientAuthentication(Port *port)
 				char		hostinfo[NI_MAXHOST];
 
 				pg_getnameinfo_all(&port->raddr.addr, port->raddr.salen,
-								   hostinfo, sizeof(hostinfo),
```
kuntalghosh: Unnecessary changes. Please discard the same, otherwise the git blame becomes difficult.

Deepesh125: Done

kuntalghosh (approved):

PR #1027, opened 2023-01-08

shalinilohia50:
Currently, Babelfish stores the modified DDL for "CREATE VIEW" statement not the original DDL used by the customer to create that view.

Task: BABEL-3835

### Description

### Issues Resolved

* **Use case based -**

* **Boundary conditions -**

* **Arbitrary inputs -**

* **Negative test cases -**

* **Performance tests -**

* **Tooling impact -**

* **Client tests -**


contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.3.0--2.4.0.sql
```diff
@@ -726,6 +726,9 @@ WHERE has_schema_privilege(s.schema_id, 'USAGE')
 AND c.contype = 'c' and c.conrelid != 0;
 GRANT SELECT ON sys.check_constraints TO PUBLIC;
 
+-- For all the views created on previous versions, the definition in the catalog should be NULL.
+UPDATE sys.babelfish_view_def SET definition = NULL;
```
2jungkook, line 729: As 3_X_DEV is the default branch now, could you raise the PR for 3_X_DEV first?

Also please be careful that there are two different upgrade paths:
1. 2.3 --> 2.4 --> 3.0 --> 3.1
2. 2.3 --> 3.0 --> 3.1

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -1922,7 +1923,14 @@ pltsql_store_view_definition(const char *queryString, ObjectAddress address)
...
 	new_record[2] = CStringGetTextDatum(NameStr(form_reltup->relname));
-	new_record[3] = CStringGetTextDatum(queryString);
+	if (original_query)
+		new_record[3] = CStringGetTextDatum(original_query);
+	/*
+	 * sp_describe_first_result_set internally translates to a "CREATE VIEW" statement,
+	 * we don't want to save the view definition for that case.
+	 */
+	else
+		new_record[3] = CStringGetTextDatum("NULL");
```
2jungkook: 1. This will store a "NULL" string, not a NULL value. Is this intended?
2. Do we have a test case for this?

contrib/babelfishpg_tsql/src/tsqlIface.cpp
```diff
@@ -2905,6 +2905,7 @@ handleBatchLevelStatement(TSqlParser::Batch_level_statementContext *ctx, tsqlSel
 	result->body = list_make1(init);
 	// create PLtsql_stmt_execsql to wrap all query string
 	PLtsql_stmt_execsql *execsql = (PLtsql_stmt_execsql *) makeSQL(ctx);
+	execsql->original_query = (makeTsqlExpr(ctx, false))->query;
```
2jungkook: It seems that we are passing `char *` from here to `pltsql_store_view_definition` hook. Could you double check the life cycle of this pointer? In other words, `(makeTsqlExpr(ctx, false))->query` is in which memory context?

contrib/babelfishpg_tsql/src/pl_exec.c
```diff
@@ -10207,3 +10211,21 @@ bool reset_search_path(PLtsql_stmt_execsql *stmt, char *old_search_path, bool* r
 	}
 	return false;
 }
+
+/*
+ * Set the global variable original_query_string with the original query.
+ */
+static void
+set_original_query_string(char *queryString)
```
2jungkook: Is there a specific reason to have this function?

contrib/babelfishpg_tsql/src/pl_exec.c
```diff
@@ -142,6 +142,7 @@ static SimpleEcontextStackEntry *simple_econtext_stack = NULL;
 	MemoryContextAlloc(get_eval_mcontext(estate), sz)
 #define eval_mcontext_alloc0(estate, sz) \
 	MemoryContextAllocZero(get_eval_mcontext(estate), sz)
+char *original_query_string;
```
2jungkook: Let's initialize this to NULL. Otherwise, it could have a garbage value and `if (original_query)` may succeed.

test/JDBC/upgrade/14_6/schedule
```diff
@@ -252,7 +252,6 @@ format
 msdb-dbo-syspolicy_system_health_state
 dateadd_internal_df
 sys-all_columns
-sys-all_sql_modules
```
2jungkook: Why did you remove this?

2jungkook:
1. Please fill in PR description properly, e.g., https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/911)
2. Instead of using the preparation folder, please follow this trick (https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/270/files), i.e.,
```
- bbf_view_definition-before-14-7-vu-prepare/verify/cleanup
- bbf_view_definition-vu-prepare/verify/cleanup
- add those to proper schedule files including latest
```

test/JDBC/upgrade/14_6/schedule
shalinilohia50: This view used the catalog table babelfish_view_def for the view definition which has been modified, so we have to write separate test files in **sys-all_sql_modules-before-14_7** format. Until then, removing this test.

contrib/babelfishpg_tsql/src/pl_exec.c
shalinilohia50: Initialized "original_query" to NULL.

contrib/babelfishpg_tsql/src/pl_exec.c
shalinilohia50: It's not needed anymore. Removed it.

contrib/babelfishpg_tsql/src/hooks.c
shalinilohia50: In case the original_query is NULL, we have inserted "queryString" in the catalog.
This executes for a scenario where we internally translate "sp_describe_first_result_set" to a "CREATE VIEW" statement, though we don't save anything in the catalog since it's not an actual view. 
IMO we can't have a test case for this.

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.3.0--2.4.0.sql
shalinilohia50: #1046 https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/1046

shalinilohia50:
> 1. Please fill in PR description properly, e.g., [Remove GetConfigOption call from TDSSendDone function #911](https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/911))
> 2. Instead of using the preparation folder, please follow this trick (https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/270/files), i.e.,
> 
> ```
> - bbf_view_definition-before-14-7-vu-prepare/verify/cleanup
> - bbf_view_definition-vu-prepare/verify/cleanup
> - add those to proper schedule files including latest
> ```

Updated the PR description.
Updated the tests for bbf_view_def, there were redundant test files.

contrib/babelfishpg_tsql/src/tsqlIface.cpp
shalinilohia50: Sorry I did not understand what's your question here. 
At this point, _ctx_ is the batch_level_statement context and it contains the original statement throughout. I am making use of that _ctx_ to get the query and save it in _execsql->original_query_

PR #1033, opened 2023-01-10
Trigger function should be created in schema the of trigger

HarshLunagariya:
### Description

Previously, Trigger function was always created in the dbo schema, no matter whether we specify schema while creating a trigger or not.

This commit fixes above issue by providing parent trigger's schema while creating implicit trigger function.

Signed-off-by: Harsh Lunagariya <lunharsh@amazon.com>

### Issues Resolved

BABEL-3629

* **Use case based -**
Added test for verifying the schemas

* **Boundary conditions -**
NA

* **Arbitrary inputs -**
NA

* **Negative test cases -**
NA

NA

Added

* **Performance tests -**
NA

* **Tooling impact -**
NA

* **Client tests -**
NA

### Check List

### DIFF
--- a/contrib/babelfishpg_tsql/src/backend_parser/gram-tsql-rule.y
+++ b/contrib/babelfishpg_tsql/src/backend_parser/gram-tsql-rule.y
@@ -2747,7 +2747,7 @@ tsql_CreateTrigStmt:
 					 * trigger will be created as part of
 					 * this create trigger command.
 					 */
-					n1->funcname = list_make1(makeString(n1->trigname));
+					n1->funcname = $3;
  					if (list_length($3) > 1){
 	 					n1->trigname = ((String *)list_nth($3,1))->sval;
 						/*
@@ -2788,7 +2788,7 @@ tsql_CreateTrigStmt:
 
 					n2->is_procedure = false;
 					n2->replace = true;
-					n2->funcname = list_make1(makeString(n1->trigname));;
+					n2->funcname = $3;
 					n2->parameters = NIL;
 					n2->returnType = makeTypeName("trigger");
 					n2->options = list_make3(lang, body, trigStmt);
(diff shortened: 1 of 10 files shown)


contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -442,6 +443,16 @@ pltsql_pre_parse_analyze(ParseState *pstate, RawStmt *parseTree)
 					}
 					trigStmt->args = NIL;
 				}
+				else
+				{
+					/*
+					 * Add schemaname to trigger's function name.
+					 */
+					if (list_length(funcStmt->funcname) == 1 && trigStmt->relation->schemaname != NULL)
```
rishabhtanwar29: I think `list_length(funcStmt->funcname) == 1` should be an assert here, since grammar logic makes sure that when `trigStmt->args == NIL` then `list_length` is one.

contrib/babelfishpg_tsql/src/backend_parser/gram-tsql-rule.y
```diff
@@ -2747,7 +2747,7 @@ tsql_CreateTrigStmt:
 					 * trigger will be created as part of
 					 * this create trigger command.
 					 */
-					n1->funcname = list_make1(makeString(n1->trigname));
```
rishabhtanwar29, line 2750: The overall logic looks fine but I am wondering why we are treating `funcname` and `trigname` differently? Due to this only we introduced this issue in the first place. IMO let's directly assign (or duplicate the list) `n1->funcname = n2->trigname` somewhere at the end since both of them should be same at the end.

forestkeeper (changes requested):
Let's add a MVU test like this: 
1. create a trigger with schema ( but func still create in dbo_master ) , and verify it's working
after MVU
2. run some test to verify it works as 1

contrib/babelfishpg_tsql/src/backend_parser/gram-tsql-rule.y
HarshLunagariya: We can't because trigname is char* only, but funcname takes list of char*.

forestkeeper (approved):
LGTM

rishabhtanwar29 (approved):

shardgupta (approved):

PR #1036, opened 2023-01-10
Fix SELECT FOR XML/JSON behavior

Sairakan:
SELECT FOR XML/JSON were both originally implemented by converting the select query into a function call that then called SPI_execute() against the original SELECT statement. While that implementation works for simple SELECT statements, it fails to execute properly against queries where the SELECT statement is a subquery which references CTEs or correlated aliases outside of the scope of the sub-SELECT. This is due to the fact that the SELECT statement is passed as a string argument to the internal function call for SPI. Thus, any external references are lost when SPI_execute() is called.

We resolve this issue by changing the implementation of SELECT FOR XML/JSON to use aggregate functions. This way, there is no need to store any query strings, and the executor will fully execute the sub-SELECT prior to passing the result rows to the aggregate function, allowing us to retain the context of any outer CTEs or correlations.

Matching Engine PR: https://github.com/babelfish-for-postgresql/postgresql_modified_for_babelfish/pull/81

### Issues Resolved

BABEL-1876

### Testing

- [x] Regression testing against older tests
- [x] Test cases with correlated CTEs/subqueries

Performance testing results:
|# of rows | SQL Server | Old Implementation | New Implemenation
--- | --- | --- | ---
|10,000 | 10 ms | 180 ms | 176 ms|
|100,000 | 435 ms | 1816 ms | 1774 ms|
|1,000,000 | 8047 ms | 18568 ms | 18066 ms|

### Check List


contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--3.0.0--3.1.0.sql
```diff
@@ -1100,6 +1100,109 @@ AND sys.babelfish_get_pltsql_function_signature(ao.object_id) = f.funcsignature
 WHERE ao.type in ('P', 'RF', 'V', 'TR', 'FN', 'IF', 'TF', 'R');
 GRANT SELECT ON sys.all_sql_modules_internal TO PUBLIC;
 
+-- deprecate old FOR XML/JSON functions if they exist - if this is a fresh install of 3.1 they won't exist, so check for that
```
2jungkook: The comment needs be updated, e.g., old functions only exist when upgrade paths do not contain v2.4 or higher.

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--3.0.0--3.1.0.sql
```diff
@@ -1100,6 +1100,109 @@ AND sys.babelfish_get_pltsql_function_signature(ao.object_id) = f.funcsignature
...
+DO $$
+BEGIN IF (SELECT count(*) FROM pg_proc as p where p.proname = 'tsql_query_to_xml') = 1 THEN
```
2jungkook: This condition could be false in the following cases:
- when a customer creates a function with the same name in non-sys schema
- when customer created a function with the same name and a different signature in sys schema

If deprecating the old function is not mandatory, could you just use TRY/CATCH?

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--3.0.0--3.1.0.sql
```diff
@@ -1100,6 +1100,109 @@ AND sys.babelfish_get_pltsql_function_signature(ao.object_id) = f.funcsignature
...
+    ALTER FUNCTION sys.tsql_query_to_xml(text, int, text, boolean, text) RENAME TO tsql_query_to_xml_deprecated_in_2_4_0;
+    CALL sys.babelfish_drop_deprecated_object('function', 'sys', 'tsql_query_to_xml_deprecated_in_2_4_0');
+END IF;
+END $$;
+
+DO $$
+BEGIN IF (SELECT count(*) FROM pg_proc as p where p.proname = 'tsql_query_to_xml_text') = 1 THEN
+    ALTER FUNCTION sys.tsql_query_to_xml_text(text, int, text, boolean, text) RENAME TO tsql_query_to_xml_text_deprecated_in_2_4_0;
```
2jungkook: proper name would be `tsql_query_to_xml_text_deprecated_in_3_1_0`

2jungkook (commented):
Please add the engine PR link to the PR description

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--3.0.0--3.1.0.sql
Sairakan: renamed to 3_1_0

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--3.0.0--3.1.0.sql
Sairakan: good point, I've changed these blocks to use BEGIN...EXCEPTION

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--3.0.0--3.1.0.sql
Sairakan: updated

2jungkook (dismissed):

2jungkook (approved):

PR #1045, opened 2023-01-12
Change AVG aggregates for integer datatypes to return correct data types

shah-nirmit:
### Description
Problem:-
The AVG aggregates were having wrong return types for Integer datatypes

Fix Approach:-
New final functions were introduced to return correct datatype and  sys.avg(bigint),sys.avg(int),sys.avg(smallint),sys.avg(int) aggregates were created in babelfishpg_common extension  so that these aggregates are given more precedance than pg_catalog.avg when the dialect is `tsql` / called from tds endpoint

References 
[current-behaviour](https://www.postgresql.org/docs/current/functions-aggregate.html#:~:text=avg%20(%20smallint,bigint%20)%20%E2%86%92%20numeric)
[expected-behaviour](https://github.com/MicrosoftDocs/sql-docs/blob/live/docs/t-sql/functions/avg-transact-sql.md#return-types)

#### Engine PR
https://github.com/babelfish-for-postgresql/postgresql_modified_for_babelfish/pull/82

Signed-off-by: Nirmit Shah <nirmisha@amazon.com>

### Issues Resolved
BABEL-3507,BABEL-3672

### Tests scenarios covered
- Use case based
  - Tested the aggregates within functions procedures and views ,also with PARTITION and ORDER BY Clauses (Partial Aggregation)
- Boundary conditions
  - Tested for datatype overflow/underflow for int/bigint return types, errors match expected outputs
- Arbitrary inputs
  - Tested with arbitrary inputs , with PARTITION and OVER Clauses
- Minor version upgrade tests
  - Added dependency tests to see upgrade impact 
- Major version upgrade tests
  - Added dependency tests to all upgrade paths to check upgrade impact 

### Check List


contrib/babelfishpg_common/sql/numerics.sql
```diff
@@ -313,11 +313,23 @@ RETURNS BIGINT
...
+RETURNS BIGINT
+AS 'babelfishpg_common', 'bigint_avg'
+LANGUAGE C IMMUTABLE PARALLEL SAFE;
+
 CREATE OR REPLACE FUNCTION sys.int4int2_sum(BIGINT)
 RETURNS INT
 AS 'babelfishpg_common' , 'int4int2_sum'
 LANGUAGE C IMMUTABLE PARALLEL SAFE;
 
+CREATE OR REPLACE FUNCTION sys.int4int2_avg(pg_catalog._int8)
```
surendravishnoi, line 326: why pg_catalog._int8 and not BIGINT?

contrib/babelfishpg_common/src/numeric.c
```diff
@@ -1037,7 +1039,7 @@ tsql_numeric_get_typmod(Numeric num)
 Datum
 bigint_sum(PG_FUNCTION_ARGS)
 {
-	return bigint_poly_sum(fcinfo);
+	return bigint_utility(fcinfo, true);
```
surendravishnoi: I think it would be best to add an enum on engine side and use that. Boolean looks error prone.

contrib/babelfishpg_common/src/numeric.c
```diff
@@ -1065,3 +1067,38 @@ int4int2_sum(PG_FUNCTION_ARGS)
...
+} Int8TransTypeData;
+
+Datum
+int4int2_avg(PG_FUNCTION_ARGS)
+{
+	ArrayType *transarray = PG_GETARG_ARRAYTYPE_P(0);
+	Int8TransTypeData *transdata;
+
+	if (ARR_HASNULL(transarray) ||
+			ARR_SIZE(transarray) != ARR_OVERHEAD_NONULLS(1) + sizeof(Int8TransTypeData))
```
surendravishnoi, line 1086: how is this check validating 2 element array?

contrib/babelfishpg_common/src/numeric.c
```diff
@@ -1065,3 +1067,38 @@ int4int2_sum(PG_FUNCTION_ARGS)
...
+			elog(ERROR, "expected 2-element int8 array");
+	transdata = (Int8TransTypeData *) ARR_DATA_PTR(transarray);
+	
+	/* SQL defines AVG of no values to be NULL */
+	if (transdata->count == 0)
+		PG_RETURN_NULL();
+	
+	if (unlikely(transdata->sum < PG_INT32_MIN) || unlikely(transdata->sum > PG_INT32_MAX))
```
surendravishnoi: put in a single unlikely

test/python/expected/upgrade_validation/expected_dependency.out
```diff
@@ -223,6 +223,7 @@ Function sys.bbf_varbinary_cmp(sys.bbf_varbinary,sys.bbf_varbinary)
 Function sys.bbfbinary_sqlvariant(sys.bbf_binary)
 Function sys.bbfvarbinary(sys.bbf_varbinary,integer,boolean)
 Function sys.bbfvarbinary_sqlvariant(sys.bbf_varbinary)
+Function sys.bigint_avg(internal)
```
surendravishnoi, line 226: we are missing upgrade dependency checks for these

test/JDBC/input/BABEL-3507-vu-prepare.sql
```diff
@@ -0,0 +1,112 @@
+-- max values causing overflow in avg (BIGINT,INT) 
+CREATE TABLE babel_3507_vu_prepare_t1(
+avgbigint BIGINT, avgint INT , avgsmallint SMALLINT , avgtinyint TINYINT )
+GO
+
```
surendravishnoi: we should also add tests for negative/positive values mix, NULL values, divide by zero, result being zero

contrib/babelfishpg_common/sql/upgrades/babelfishpg_common--2.3.0--2.4.0.sql
```diff
@@ -0,0 +1,54 @@
+-- complain if script is sourced in psql, rather than via ALTER EXTENSION
```
Deepesh125: why do we need to update this upgrade script if this change is only meant for 3_X_DEV branch?

contrib/babelfishpg_common/src/numeric.c
```diff
@@ -1065,3 +1067,38 @@ int4int2_sum(PG_FUNCTION_ARGS)
...
+
+Datum
+int4int2_avg(PG_FUNCTION_ARGS)
+{
+	ArrayType *transarray = PG_GETARG_ARRAYTYPE_P(0);
+	Int8TransTypeData *transdata;
+
+	if (ARR_HASNULL(transarray) ||
+			ARR_SIZE(transarray) != ARR_OVERHEAD_NONULLS(1) + sizeof(Int8TransTypeData))
+			elog(ERROR, "expected 2-element int8 array");
```
Deepesh125: use ereport instead of elog.

test/JDBC/expected/BABEL-3507-vu-prepare.out
```diff
@@ -0,0 +1,148 @@
+-- max values causing overflow in avg (BIGINT,INT)
```
Deepesh125: what would happen if one of the value is NULL?

Deepesh125 (changes requested):

contrib/babelfishpg_common/sql/numerics.sql
shah-nirmit: The transition functions used for the aggregates (`int2_avg_accum`/`int4_avg_accum`) use `_int8` type as the state ,so this state is passed to the final functions; so the input to the final function is given as `pg_catalog._int8`

test/python/expected/upgrade_validation/expected_dependency.out
shah-nirmit: There are C type function (final functions) that are being used when the aggregate are called . There are dependency objects(functions,procedures,views) in the test files. do we need to have dependency tests for them explicitly?

contrib/babelfishpg_common/sql/numerics.sql
surendravishnoi: please add Jungkook for review, we need to make sure this is not a problem

test/python/expected/upgrade_validation/expected_dependency.out
surendravishnoi: Talk to Rishabh, see if we are taking in account indirect dependencies or not?

contrib/babelfishpg_common/src/numeric.c
shah-nirmit: Resolved in the latest commit to engine PR

contrib/babelfishpg_common/src/numeric.c
shah-nirmit: Resolved in the latest commit

test/JDBC/input/BABEL-3507-vu-prepare.sql
shah-nirmit: Added extra tests in the latest commit.
`divide-by-zero` situation would not arise since if all values are NULL we don't increment the `state->count` . and return NULL if `state->count` is 0; [Reference](https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/1045/files#diff-ff0c89ab284f5bb75c83374ee361d8b192bca9db405599bb4bbcd316eddea1afR1090-R1091)

contrib/babelfishpg_common/sql/upgrades/babelfishpg_common--2.3.0--2.4.0.sql
shah-nirmit: Fixed in latest commit, we need as to have all the 2_X sql upgrade script in 3_X dev as well

contrib/babelfishpg_common/src/numeric.c
shah-nirmit: Fixed in latest commit

test/JDBC/expected/BABEL-3507-vu-prepare.out
shah-nirmit: If one of the values are NULL , they are not used in avg calculation thus not causing overflow

contrib/babelfishpg_common/src/numeric.c
shah-nirmit: This check is also being used in engine code for function `int8_avg` which is the final function that is used by `pg_catalog.avg` , so I have used the same validation logic for the state received .
Reference [link](https://github.com/babelfish-for-postgresql/postgresql_modified_for_babelfish/blob/BABEL_3_X_DEV__PG_15_X/src/backend/utils/adt/numeric.c#L6672-L6693)

test/python/expected/upgrade_validation/expected_dependency.out
rishabhtanwar29: Currently we are not tracking indirect dependencies as even pg_depend does not track such dependencies. This looks like a function to function dependency where aggregate function depends upon other functions. But IMO, adding test for aggregate function should be enough since it's helper functions will anyway get executed/tested as part of it.

contrib/babelfishpg_common/sql/numerics.sql
2jungkook: It looks okay.

contrib/babelfishpg_common/sql/upgrades/babelfishpg_common--3.0.0--3.1.0.sql
```diff
@@ -0,0 +1,54 @@
+-- complain if script is sourced in psql, rather than via ALTER EXTENSION
+\echo Use "ALTER EXTENSION ""babelfishpg_common"" UPDATE TO '3.1.0'" to load this file. \quit
```
2jungkook: Please add `2.3.0--2.4.0` and `2.4.0--3.0.0` upgrade scripts.

test/JDBC/upgrade/13_6/schedule
```diff
@@ -294,3 +294,4 @@ ISC-Table_Constraints
 sys_server_principals_dep_for_13_x
 sys_database_principals_dep_for_13_x
 case_insensitive_collation
+BABEL-3507-before-15-1
```
2jungkook: 1. Please use a more descriptive test name other than BABEL-3507.
2. It should have `before-15-2-or-14-7` suffix because our current dev cycle is targeting PG15.2 and PG14.7 (cc: @rishabhtanwar29)
3. `BABEL-3507-before-15-1` and `BABEL-3507` has only one line difference. In that case, please consider creating three test files to avoid duplicates, e.g., `<test name>`, `<test name>-type-before-15-2-or-14-7` and `<test name>-type`.

test/JDBC/upgrade/14_7/schedule
```diff
@@ -365,5 +365,6 @@ sys_database_principals_dep
 atn2
 app_name
 case_insensitive_collation
+BABEL-3507-before-15-1
```
2jungkook: This should be `BABEL-3507`

contrib/babelfishpg_common/sql/upgrades/babelfishpg_common--3.0.0--3.1.0.sql
shah-nirmit: Fixed in the latest commit

test/JDBC/upgrade/13_6/schedule
shah-nirmit: Fixed in the latest commit,
Divide existing test into common and Dependency tests

test/JDBC/upgrade/14_7/schedule
shah-nirmit: Fixed in the latest commit

contrib/babelfishpg_common/sql/upgrades/babelfishpg_common--2.4.0--3.0.0.sql
```diff
@@ -0,0 +1,29 @@
+-- complain if script is sourced in psql, rather than via ALTER EXTENSION
```
surendravishnoi, line 1: why have we added collation related changes here?

contrib/babelfishpg_common/src/numeric.c
surendravishnoi: it is still the same

test/JDBC/input/AVG-Aggregate-Dep-before-15-2-or-14-7-vu-prepare.sql
```diff
@@ -0,0 +1,58 @@
+CREATE TABLE avg_dep_vu_prepare_t1(
```
surendravishnoi, line 1: why is this a separate file?

contrib/babelfishpg_common/sql/upgrades/babelfishpg_common--2.4.0--3.0.0.sql
shah-nirmit: With accordance to https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/1045#discussion_r1072608467
In case there is scenario where one upgrades first from 14.6->14.7 and then 14.7->15.2. 
We would need the collation changes that were also done for 2.3->3.0 . Both of the upgrade script should be identical

test/JDBC/input/AVG-Aggregate-Dep-before-15-2-or-14-7-vu-prepare.sql
shah-nirmit: We need the seperate file as the dependant objects (only dependant view) would see a change in return type returning (numeric for all avg aggregate before the fix). Since function and procedure are compiled on runtime they would have correct (new ) return types

rishabhtanwar29 (dismissed):

rishabhtanwar29:
Upgrade changes look good.

Deepesh125 (dismissed):

shah-nirmit (commented):
Note:- the 14_7->latest upgrade path will fail for AVG-Aggregate-Dep ,2X PR needs to be merged First

contrib/babelfishpg_common/sql/upgrades/babelfishpg_common--2.4.0--3.0.0.sql
Deepesh125: Yes, this changes are required to make the MVU work.

Deepesh125 (approved):

surendravishnoi (approved):

PR #1046, opened 2023-01-12

shalinilohia50:
### Description

Currently, Babelfish stores the modified DDL for "CREATE VIEW" statement not the original DDL used by the customer to create that view. With this change, we store the original DDL in the catalog babelfish_view_def

Since the DDL is modified by the Antlr, few of the DDLs might not be valid for SQL Server. To make sure all the exported DDLs are compatible with SQL Server, we have stored the original DDL in the catalog.

We have introduced a global variable to store the original query string before the Antlr modifies it and pass it on to the hook which stores the query string in the catalog.

### Issues Resolved
Task: BABEL-3835

* **Use case based - Tests to show that the original DDL is stored in the catalog**
* **Boundary conditions -**
* **Arbitrary inputs -**
* **Negative test cases -**
* **Minor version upgrade tests - In case of upgrades, when the views are created on previous versions the catalog will contain NULL. If the customer wants to restore the view definition, they would need to recreate the views.**
* **Performance tests - N/A**
* **Tooling impact - N/A**
* **Client tests - JDBC tests**


contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--3.0.0--3.1.0.sql
```diff
@@ -935,6 +935,9 @@ LEFT JOIN pg_foreign_data_wrapper AS w ON f.srvfdw = w.oid
 WHERE w.fdwname = 'tds_fdw';
 GRANT SELECT ON sys.linked_logins TO PUBLIC;
 
+-- For all the views created on previous versions, the definition in the catalog should be NULL.
+UPDATE sys.babelfish_view_def SET definition = NULL;
```
2jungkook: We should be careful about this because there are two different upgrade paths:
1. 2.3 --> 3.0 --> 3.1
2. 2.4 --> 3.0 --> 3.1

In the first case, `set definition = NULL` makes sense. However, in the second case, this will remove the correct view definitions created from v2.4.

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -1930,7 +1931,14 @@ pltsql_store_view_definition(const char *queryString, ObjectAddress address)
 	new_record[0] = Int16GetDatum(dbid);
 	new_record[1] = CStringGetTextDatum(logical_schemaname);
 	new_record[2] = CStringGetTextDatum(NameStr(form_reltup->relname));
-	new_record[3] = CStringGetTextDatum(queryString);
+	if (original_query)
+		new_record[3] = CStringGetTextDatum(original_query);
+	/*
+	 * sp_describe_first_result_set internally translates to a "CREATE VIEW" statement,
```
2jungkook: Please add a test for this, i.e., use `sp_describe_first_result_set` and check `babelfish_view_def`.

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -1930,7 +1931,14 @@ pltsql_store_view_definition(const char *queryString, ObjectAddress address)
...
+	 * we won't have the original view definition for that case.
+	 */
+	else
+		new_record[3] = CStringGetTextDatum(queryString);
```
2jungkook: If we store `queryString` here, what would be impact of DDL Export from SSMS? Will a T-SQL user get a non-executable SQL?

contrib/babelfishpg_tsql/src/tsqlIface.cpp
```diff
@@ -2905,6 +2905,7 @@ handleBatchLevelStatement(TSqlParser::Batch_level_statementContext *ctx, tsqlSel
 	result->body = list_make1(init);
 	// create PLtsql_stmt_execsql to wrap all query string
 	PLtsql_stmt_execsql *execsql = (PLtsql_stmt_execsql *) makeSQL(ctx);
+	execsql->original_query = (makeTsqlExpr(ctx, false))->query;
```
2jungkook: Inside `makeTsqlExpr()`, we use `palloc()` to create the query string. When we use `palloc()`, the memory is allocated in the `CurrentMemoryContext`. In this PR, we don't use any deep copy and just passing `char *` from ANTLR to `pltsql_store_view_definition` hook. If the life cycle of `CurrentMemoryContext` is different between here and `pltsql_store_view_definition`, it could be problematic.

For more details, please refer https://github.com/babelfish-for-postgresql/postgresql_modified_for_babelfish/blob/BABEL_3_X_DEV__PG_15_X/src/backend/utils/mmgr/README

test/JDBC/upgrade/14_6/schedule
```diff
@@ -259,7 +259,6 @@ format
 msdb-dbo-syspolicy_system_health_state
 dateadd_internal_df
 sys-all_columns
-sys-all_sql_modules
```
2jungkook: Please do not remove existing test files. If some modifications are needed due to this PR, then let's include that as well.

test/JDBC/upgrade/14_7/schedule
```diff
@@ -367,3 +366,4 @@ app_name
 case_insensitive_collation
 sys-has_perms_by_name
 sys-has_perms_by_name-dep
+bbf_view_def_upgrade-before-15_2
```
2jungkook: This should be `bbf_view_def` (let's also remove "upgrade" from the file name).

test/JDBC/jdbc_schedule
```diff
@@ -77,6 +77,12 @@ ignore#!#BABEL-CHECK-CONSTRAINT-before-14_6-vu-verify
 ignore#!#BABEL-CHECK-CONSTRAINT-before-14_6-vu-cleanup
 ignore#!#BABEL-3646-before-14_6-vu-prepare
 ignore#!#BABEL-3646-before-14_6-vu-verify
+ignore#!#bbf_view_def_upgrade-before-14_5-vu-prepare
+ignore#!#bbf_view_def_upgrade-before-14_5-vu-verify
+ignore#!#bbf_view_def_upgrade-before-14_7-vu-prepare
+ignore#!#bbf_view_def_upgrade-before-14_7-vu-verify
+ignore#!#bbf_view_def_upgrade-before-15_2-vu-prepare
+ignore#!#bbf_view_def_upgrade-before-15_2-vu-verify
```
2jungkook: 1. vu stands for "version upgrade", so you don't need "upgrade" in the file name.
2. It seems that `before-14_7` and `before_15_2` files are identical. Then let's use `before-14_7-or-15_2`.

test/JDBC/input/ISC-Views.sql
```diff
@@ -77,6 +77,12 @@ go
 create view vcnums as select * from nums with check option;
 go
 
+create view iscv as select * from information_schema.views;
+go
+
+create view sbv with schemabinding as select 1;
```
2jungkook: You don't need this test because `vu` tests will be executed in both normal and upgrade scenario.

test/JDBC/jdbc_schedule
shalinilohia50: Updated in the next commit.

test/JDBC/upgrade/14_7/schedule
shalinilohia50: Sure

test/JDBC/upgrade/14_6/schedule
shalinilohia50: Agreed

contrib/babelfishpg_tsql/src/tsqlIface.cpp
shalinilohia50: Used a duplicate string to avoid any problematic situation in future.

contrib/babelfishpg_tsql/src/hooks.c
shalinilohia50: Removed this condition. The value should be NULL in `else` case.

contrib/babelfishpg_tsql/src/hooks.c
shalinilohia50: Removed this condition, it should be NULL when the original_query is NULL.

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--3.0.0--3.1.0.sql
shalinilohia50: Yes, the upgrade script from 3.0 to 3.1 needs to be updated. 
The PR for 2_X_DEV needs to be merged before.

surendravishnoi: what it has to do with 2_x_dev PR? We need to distinguish cases where definition in table is original vs babelfish one

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -1866,6 +1866,7 @@ pltsql_store_view_definition(const char *queryString, ObjectAddress address)
 	uint64		flag_values = 0, flag_validity = 0;
 	char		*physical_schemaname;
 	const char  *logical_schemaname;
+	char *original_query = get_original_query_string();
```
surendravishnoi, line 1869: We should think if this can be done in better way. Having a global variable to track original query string is not most desirable.

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--3.0.0--3.1.0.sql
shalinilohia50: Yes, we are adding one more bit to an existing column "flag_values" which is set to 1 in the 2_X dev if view is created 2.4 and onwards.
In 3.0 to 3.1 upgrade script, we need to check if that bit is set to 1 ( which indicates that the view was created in or after 2.4) and we don't reset those definition to NULL in that cases.

contrib/babelfishpg_tsql/src/hooks.c
shalinilohia50: We are getting the original query from the Antlr and storing it in the `PLtsql_stmt_execsql` and need to pass it on to the hook `pltsql_store_view_definition` in hooks.c. 
Could you suggest something if not a global variable?

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -1930,7 +1931,10 @@ pltsql_store_view_definition(const char *queryString, ObjectAddress address)
 	new_record[0] = Int16GetDatum(dbid);
 	new_record[1] = CStringGetTextDatum(logical_schemaname);
 	new_record[2] = CStringGetTextDatum(NameStr(form_reltup->relname));
-	new_record[3] = CStringGetTextDatum(queryString);
+	if (original_query)
+		new_record[3] = CStringGetTextDatum(original_query);
+	else
+		new_record_nulls[3] = true;
```
2jungkook, line 1945: As discussed, please set the third bit to 0 explicitly to prevent confusions when we add another bit in the future. Also, please add detailed comments on the meaning of the 3rd bit.

test/JDBC/upgrade/14_7/schedule
```diff
@@ -364,6 +364,7 @@ app_name
 case_insensitive_collation
 sys-has_perms_by_name
 sys-has_perms_by_name-dep
+bbf_view_def-before-14_7-or-15_2
```
2jungkook: This should be changed after merging the 2_X_DEV PR.

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -1926,11 +1927,22 @@ pltsql_store_view_definition(const char *queryString, ObjectAddress address)
 	flag_validity |= BBF_VIEW_DEF_FLAG_USES_QUOTED_IDENTIFIER;
 	if (pltsql_quoted_identifier)
 		flag_values |= BBF_VIEW_DEF_FLAG_USES_QUOTED_IDENTIFIER;
+	/*
+	 * Setting this flag big to 0 to distinguish between the objects
```
2jungkook: flag bit

contrib/babelfishpg_tsql/src/hooks.c
shalinilohia50: done

test/JDBC/upgrade/14_7/schedule
shalinilohia50: Done

contrib/babelfishpg_tsql/src/catalog.h
```diff
@@ -176,6 +176,7 @@ typedef FormData_authid_user_ext *Form_authid_user_ext;
 #define BBF_VIEW_DEF_NUM_COLS 8
 #define BBF_VIEW_DEF_FLAG_IS_ANSI_NULLS_ON (1<<0)
 #define BBF_VIEW_DEF_FLAG_USES_QUOTED_IDENTIFIER (1<<1)
+#define BBF_VIEW_DEF_FLAG_CREATED_IN_OR_AFTER_2_4 (0<<2)
```
surendravishnoi, line 183: we should not use version number in macro, a comment can mention version number. Macro could be something like _CREATED_WITH_ORIGINAL_SQL_DEF

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -1926,11 +1927,22 @@ pltsql_store_view_definition(const char *queryString, ObjectAddress address)
 	flag_validity |= BBF_VIEW_DEF_FLAG_USES_QUOTED_IDENTIFIER;
 	if (pltsql_quoted_identifier)
 		flag_values |= BBF_VIEW_DEF_FLAG_USES_QUOTED_IDENTIFIER;
+	/*
+	 * Setting this flag bit to 0 to distinguish between the objects
+	 * created in 2.x or 3.x for future references. Let's not use
+	 * this bit in 3.x, as we are setting this to 1 in 2.x and will
+	 * be reserved for MVU.
+	 */
+	flag_validity |= BBF_VIEW_DEF_FLAG_CREATED_IN_OR_AFTER_2_4;
```
surendravishnoi, line 1936: why setting it to 0? why not set it to 1

shalinilohia50: This is to distinguish between 2.4 and 3.1 versions in 3.2, if we want to introduce a new bit in future.
We are keeping 3rd bit from the right to track the change from 2.x, explicitly keeping it 0 in 3.x.
When we upgrade from 2.4 -> 3.1 -> 3.2, the 3rd bit will be 1 which means the object was created in 2.x.

contrib/babelfishpg_tsql/src/catalog.h
shalinilohia50: The MACRO with the same name is merged in 2_X_DEV. Let's keep it same in 3_X

shalinilohia50:
After the PR for BABEL-3802 is merged to 3_X_DEV, the tests should rerun and all the tests should pass.

contrib/babelfishpg_tsql/src/hooks.c
surendravishnoi: 1 does not mean it is from 2.x, 1 means it is from 2.4 while 0 means it is from 2.3. Now with this change, 1 means it is from 2.4 and 0 means it is from 2.3/3.x which is confusing. Otherwise this could be used by customer to find all the views which needs recreation to get correct definition in the catalog (views with bit set to 0)

shalinilohia50: I agree. But this bit is not added for the customer to find out the views which needs to be recreated, it is solely for the MVU purpose. This is an internal table.

We can convey to customer "If the view definition comes out be NULL, then it needs to be recreated".

NOTE: If the customer is on 3.x,  the view definition would not be NULL even though the 3rd bit will be 0 except 3.0.
1. 2.3 -> 3.0 (create V1) -> 3.1 ( the view definition V1 would be NULL) which means that the view V1 needs to be recreated.

2jungkook (approved):

jsudrik:
Currently, Babelfish stores the modified DDL for "CREATE VIEW" statement not the original DDL used by the customer to create that view. With this change, we store the original DDL in the catalog babelfish_view_def

PR #1060, opened 2023-01-14
Handle collation name in exported DDL's CHECK() clause

shalinilohia50:
### Description

When a DDL is exported where an object's CHECK() clause has an explicit COLLATE clause with SQL Server collation, the collation name is converted to its equivalent BBF name while recreating the definition which is valid for BBF but not for SQL Server.

When a SQL Server collation is specified with an object in the CHECK() clause of a table, the definition will show the SQL Server collation name (there might be a possibility it shows it's closest match) but not it's equivalent Babelfish Collation name which would work on both SQL Server and BBF.
When a BBF collation name is provided, it shows it's closed SQL Server collation name if any. Otherwise it shows the same BBF name.

### Issues Resolved
Handle collation name in exported DDL's CHECK() clause
Task: BABEL-3838

* **Use case based - Updated the existing test cases.**
* When a "default" collation is used.
* When a SQL Server collation is used.
* When a BBF collation is used.
* **Boundary conditions -**
* **Arbitrary inputs -**
* **Negative test cases -**
* **Performance tests -**
* **Tooling impact -**
* **Client tests -**


contrib/babelfishpg_common/src/collation.c
```diff
@@ -601,6 +634,18 @@ translate_collation(const char *collname, bool check_for_server_collation_name_g
 	return idx;
 }
 
+/* Translate BBF collation to it's closest TSQL Collation. */
+const char *
+translate_bbf_collation_to_tsql_collation(const char *collname)
+{
+	int last = TOTAL_REVERSE_COLL_TRANSLATION_COUNT - 1;
```
surendravishnoi: aren't we missing the last collation due to count - 1?

test/JDBC/expected/ISC-Check-Constraints-vu-prepare.out
```diff
@@ -29,8 +29,8 @@ create table test_tsql_cast(
 go
 
 create table test_tsql_collate(
-	c_varchar varchar check(c_varchar <> cast('sflkjasdlkfjf' as varchar(12)) COLLATE bbf_unicode_cp1_ci_as),
```
surendravishnoi, line 32: we should have test case for each collation name in reverse transaction table, also we should have test case to make sure that we validate fallback case where transaction table does not find matching TSQL collation

contrib/babelfishpg_common/src/collation.c
shalinilohia50: Oh yes! Fixed it.

test/JDBC/expected/ISC-Check-Constraints-vu-prepare.out
shalinilohia50: We already have a test case where the collation name "japanese_ci_as" doesn't find a match in reverse translation table. 
Should we add test for all the ~20 collation names? We have added tests for first, last value in the table and a negative case where the collation name doesn't exist in the table.

surendravishnoi: we should if we do not have already, specially to guard against table getting messed up in future

shalinilohia50: Added in the next commit.

contrib/babelfishpg_common/src/collation.c
```diff
@@ -601,6 +634,17 @@ translate_collation(const char *collname, bool check_for_server_collation_name_g
 	return idx;
 }
 
+/* Translate BBF collation to it's closest TSQL Collation. */
+const char *
+translate_bbf_collation_to_tsql_collation(const char *collname)
+{
+	for (int i = 0; i < TOTAL_REVERSE_COLL_TRANSLATION_COUNT; i++)
+		if (pg_strcasecmp(reverse_coll_translations[i].from_collname, collname) == 0)
```
surendravishnoi, line 642: can we provide collation names in mixed/upper case? what is the expected behaviour in that case?

shalinilohia50: Yes, we can give the collation name in mixed/upper case but we always print the collation name in lower case as obtained by the reverse translation table.
The"collname" we pass to this function is always in the lower case as generated by the "generate_collation_name".

surendravishnoi: and we have tests cases to cover lowercase, upper case and mixed case?

shalinilohia50: Yes, added such tests.

surendravishnoi (approved):

Deepesh125 (approved):

PR #1061, opened 2023-01-14
Store original DDL for procedure and functions in the catalog

shalinilohia50:
### Description

* Instead of recreating the routine definition from the information stored in the **pg_proc** tuple, store the original DDL in the **sys.babelfish_function_ext** catalog including Table Valued Functions.
* Update the view ISC routines and sys.all_sql_modules_internal to use the original function/procedure definition stored in the catalog. 
* SSMS uses the view sys.all_sql_modules to show the function/procedure definition needed for DDL Export, it was failing previously due to the invalid SQL Server syntax in few cases. Since we were recreating the DDL from the data stored in pg_proc tuple and the sql body which was stored in the tuple is modiflied by Antlr, some of the syntax could be invalid for SQL Server. So we want to store the original DDL for DDL Export.

### Issues Resolved
Store original DDL for procedure and functions in the catalog 
Task: BABEL-3839, BABEL-3355

* **Use case based -**
* Added tests to show how the exact routine definition is being stored and retrieved from the catalog. 
*  Added tests to show the definition for Table Valued functions (both Inline and Multi statement)
* **Boundary conditions -**
* **Arbitrary inputs -**
* **Negative test cases -**
* **Performance tests -**
* **Tooling impact -**
* **Client tests -**
JDBC tests: The test structure has been changed a bit. Instead of creating redundant prepare, verify and cleanup scripts, we have created the tests in _before-14_5_  format to run the tests in previous versions and commented those tests in test/JDBC/jdbc_schedule file.


shalinilohia50:
NOTE: These tests are currently added in the upgrade path 14_7 -> latest. 

> routines_definition_before-14_7-or-15_2
> sys-all_sql_modules_before-14_7-or-15_2
> sys-sql_modules_before-14_7-or-15_2

Currently the respective changes are not in 2_X_DEV, so the tests are still picking the older definition of the views which is expected.
After the changes are merged in 2_X_Dev, we need to modify these tests to the following:
> routines_definition
> sys-all_sql_modules
> sys-sql_modules

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.3.0--2.4.0.sql
```diff
@@ -614,6 +614,163 @@ GRANT EXECUTE ON FUNCTION sys.has_perms_by_name(
     sub_securable sys.SYSNAME,
     sub_securable_class sys.nvarchar(60)) TO PUBLIC;
 
+-- Add one column to store definition of the function in the table.
+SET allow_system_table_mods = on;
+ALTER TABLE sys.babelfish_function_ext add COLUMN IF NOT EXISTS definition sys.NTEXT DEFAULT NULL;
+SET allow_system_table_mods = off;
```
rishabhtanwar29: Let's use `reset` statement here.

contrib/babelfishpg_tsql/sql/information_schema_tsql.sql
```diff
@@ -687,7 +687,9 @@ CREATE OR REPLACE VIEW information_schema_tsql.routines AS
 
        FROM sys.pg_namespace_ext nc LEFT JOIN sys.babelfish_namespace_ext ext ON nc.nspname = ext.nspname,
             pg_proc p inner join sys.schemas sch on sch.schema_id = p.pronamespace
-	    inner join sys.all_objects ao on ao.object_id = CAST(p.oid AS INT),
+	    inner join sys.all_objects ao on ao.object_id = CAST(p.oid AS INT)
+		left join sys.babelfish_function_ext f on p.proname = f.funcname and sch.schema_id::regnamespace::name = f.nspname
+			and sys.babelfish_get_pltsql_function_signature(p.oid) = f.funcsignature collate sys.database_default,
```
rishabhtanwar29: Can we just use `regnamespace` and function signature for the join and remove name since schema + signature is sufficient and make primary.

contrib/babelfishpg_tsql/sql/sys_views.sql
```diff
@@ -1760,6 +1760,8 @@ LEFT OUTER JOIN sys.babelfish_view_def bvd
       ao.name = bvd.object_name 
    )
 LEFT JOIN pg_proc p ON ao.object_id = CAST(p.oid AS INT)
+LEFT JOIN sys.babelfish_function_ext f on ao.name = f.funcname collate "C" and ao.schema_id::regnamespace::name = f.nspname
+and sys.babelfish_get_pltsql_function_signature(ao.object_id) = f.funcsignature collate "C"
```
rishabhtanwar29: Same as above.

rishabhtanwar29: * Why are we using different collate clauses? Here it is `collate "C"` but above it was `collate sys.database_default`. Let's use `collate "C"` throughout to be consistent.
* Nit: SQL tokens (`and`, `on` etc) should be capital.

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -2235,6 +2236,8 @@ pltsql_store_func_default_positions(ObjectAddress address, List *parameters, con
 	new_record[Anum_bbf_function_ext_flag_values - 1] = UInt64GetDatum(flag_values);
 	new_record[Anum_bbf_function_ext_create_date - 1] = TimestampGetDatum(GetSQLLocalTimestamp(3));
 	new_record[Anum_bbf_function_ext_modify_date - 1] = TimestampGetDatum(GetSQLLocalTimestamp(3));
+	if (original_query)
+		new_record[Anum_bbf_function_ext_definition - 1] = CStringGetTextDatum(original_query);
```
rishabhtanwar29, line 2243: Missing `else` condition, need to set new_record_nulls[Anum_bbf_function_ext_definition - 1] to true when `original_query` is NULL.

contrib/babelfishpg_tsql/src/pltsql.h
```diff
@@ -1874,7 +1875,7 @@ extern void pltsql_exec_get_datum_type_info(PLtsql_execstate *estate,
 
 extern int get_insert_bulk_rows_per_batch(void);
 extern int get_insert_bulk_kilobytes_per_batch(void);
-
+extern char* get_original_query_string(void);
```
rishabhtanwar29: Nit: `*` should be attached to function name (not datatype).

contrib/babelfishpg_tsql/src/pl_exec.c
```diff
@@ -142,6 +142,7 @@ static SimpleEcontextStackEntry *simple_econtext_stack = NULL;
 	MemoryContextAlloc(get_eval_mcontext(estate), sz)
 #define eval_mcontext_alloc0(estate, sz) \
 	MemoryContextAllocZero(get_eval_mcontext(estate), sz)
+char *original_query_string = NULL;
```
rishabhtanwar29: * We should put this statement with all the other variable declaration above.
* Let's make this `static` if it is used only in this file.

contrib/babelfishpg_tsql/src/pl_exec.c
```diff
@@ -10207,3 +10210,12 @@ bool reset_search_path(PLtsql_stmt_execsql *stmt, char *old_search_path, bool* r
 	}
 	return false;
 }
+
+/*
+ * Get the original_query_string which stores the original query.
+ */
+char*
```
rishabhtanwar29: Missing space b/w `char` and `*`.

rishabhtanwar29 (dismissed):

contrib/babelfishpg_tsql/sql/information_schema_tsql.sql
shalinilohia50: As discussed offline, I will file a new JIRA to track this separately.

rishabhtanwar29 (approved):

test/JDBC/upgrade/14_7/schedule
```diff
@@ -366,4 +364,6 @@ app_name
 case_insensitive_collation
 sys-has_perms_by_name
 sys-has_perms_by_name-dep
+sys-all_sql_modules
+sys-sql_modules
```
rishabhtanwar29, line 368: These changes look unnecessary to me, we are adding and removing the same tests.

shardgupta (approved):

PR #1067, opened 2023-01-17
Store original DDL for procedure and functions in the catalog

shalinilohia50:
Store original DDL for procedure and functions in the catalog

### Description

* Instead of recreating the routine definition from the information stored in the **pg_proc** tuple, store the original DDL in the **sys.babelfish_function_ext** catalog.
* Update the view ISC routines and sys.all_sql_modules_internal to use the original function/procedure definition stored in the catalog. 
* Previously, BBF did not show the routine definition for Table Valued Functions. Now we store the original definition for TVFs also.
* SSMS uses the view sys.all_sql_modules to show the function/procedure definition needed for DDL Export, it was failing previously due to the invalid SQL Server syntax in few cases.

### Issues Resolved

Task: BABEL-3839, BABEL-3355

* **Use case based -**
* Added tests to show how the exact routine definition is being stored and retrieved from the catalog. 
*  Added tests to show the definition for Table Valued functions (both Inline and Multi statement)

* **Boundary conditions -**

* **Arbitrary inputs -**

* **Negative test cases -**

* **Performance tests -**

* **Tooling impact -**

* **Client tests -**
JDBC tests: The test structure has been changed a bit. Instead of creating redundant prepare, verify and cleanup scripts, we have created the tests in _before-14_5_  format to run the tests in previous versions and commented those tests in test/JDBC/jdbc_schedule file.


contrib/babelfishpg_tsql/sql/ownership.sql
```diff
@@ -25,6 +25,7 @@ CREATE TABLE sys.babelfish_function_ext (
 	flag_values BIGINT,
 	create_date SYS.DATETIME NOT NULL,
 	modify_date SYS.DATETIME NOT NULL,
+  	definition sys.NTEXT,
```
rishabhtanwar29: Oh I missed to point this before, we need to specify `DEFAULT NULL` here otherwise it will be different from upgraded server.

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -2099,6 +2099,7 @@ pltsql_store_func_default_positions(ObjectAddress address, List *parameters, con
 	ListCell	*x;
 	int			idx;
 	uint64		flag_values = 0, flag_validity = 0;
+	const char *original_query = get_original_query_string();
```
rishabhtanwar29: Why `const`?

shalinilohia50: It's not needed, removed it

rishabhtanwar29 (approved):
LGTM

shardgupta (approved):

PR #1074, opened 2023-01-17
Support storing and retrieving TSQL trigger definition in Babelfish

shalinilohia50:
### Description

SSMS uses the views sys.sql_modules and sys.all_sql_modules to show the object definitions. This change is needed for the SSMS scripting so that users have a way to script the trigger definition in TSQL syntax which helps in DDL Export.

### Issues Resolved

Task: BABEL-3681

* **Use case based -**
The views sys.sql_modules and sys.all_sql_modules should show the trigger definition instead of NULL.

* **Boundary conditions -**
* **Arbitrary inputs -**
* **Negative test cases -**
* **Performance tests -**
* **Tooling impact -**
* **Client tests -**
JDBC Tests

### DIFF
--- a/contrib/babelfishpg_tsql/sql/sys_views.sql
+++ b/contrib/babelfishpg_tsql/sql/sys_views.sql
@@ -1739,7 +1739,7 @@ SELECT
   , CAST(0 as sys.bit)  AS uses_database_collation
   , CAST(0 as sys.bit)  AS is_recompiled
   , CAST(
-      CASE WHEN ao.type IN ('P', 'FN', 'IN', 'TF', 'RF', 'IF', 'TR') THEN
+      CASE WHEN ao.type IN ('P', 'FN', 'IN', 'TF', 'RF', 'IF') THEN
         CASE WHEN p.proisstrict THEN 1
         ELSE 0 
         END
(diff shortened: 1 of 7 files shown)


contrib/babelfishpg_tsql/sql/sys_views.sql
```diff
@@ -1740,7 +1739,7 @@ SELECT
   , CAST(0 as sys.bit)  AS uses_database_collation
   , CAST(0 as sys.bit)  AS is_recompiled
   , CAST(
-      CASE WHEN ao.type IN ('P', 'FN', 'IN', 'TF', 'RF', 'IF') THEN
```
HarshLunagariya: Is this changes is acc. to TSQL behaviour? If yes then can we please add a positive testcase as well.

rishabhtanwar29 (dismissed):

contrib/babelfishpg_tsql/sql/sys_views.sql
shalinilohia50: This was meant only for functions. Removed it in the next commit.

HarshLunagariya (approved):

shardgupta (approved):

PR #1082, opened 2023-01-18

shalinilohia50:
### Description

Currently, Babelfish stores the modified DDL(rewritten by Antlr) for "CREATE VIEW" statement and not the original DDL used by the customer to create that view. 
With this change, we store the original DDL in the catalog babelfish_view_def instead of the modified DDL.

Since the DDL gets modified by the Antlr and we were storing the modified DDL in the catalog earlier, few of the DDLs would not be valid for SQL Server. SSMS used the definition column of the view "babelfish_view_def" to export the DDL.
To make sure that all the exported DDLs are compatible with SQL Server, we have stored the original DDL in the catalog.

### Issues Resolved
Task: BABEL-3835

* **Use case based - Tests to show that the original DDL is stored in the catalog**
* **Boundary conditions -**
* **Arbitrary inputs -**
* **Negative test cases -**
* **Minor version upgrade tests - In case of upgrades, when the views are created on previous versions the catalog will contain NULL. If the customer wants to restore the view definition, they would need to recreate the views.**
* **Major version upgrade tests - **
* **Performance tests - N/A**
* **Tooling impact - N/A**
* **Client tests - JDBC tests**


test/JDBC/jdbc_schedule
```diff
@@ -71,6 +71,12 @@ ignore#!#sys_database_principals_dep_for_13_x-vu-verify
 ignore#!#sys_server_principals_dep_for_13_x-vu-cleanup
 ignore#!#sys_server_principals_dep_for_13_x-vu-prepare
 ignore#!#sys_server_principals_dep_for_13_x-vu-verify
+ignore#!#bbf_view_def-vu-prepare
+ignore#!#bbf_view_def-vu-verify
+ignore#!#bbf_view_def-before-14_5-vu-prepare
+ignore#!#bbf_view_def-before-14_5-vu-verify
+ignore#!#bbf_view_def-before-14_7-vu-prepare
```
2jungkook: To minimize the difference between 2_X_DEV and 3_X_DEV, let's use the same name, i.e., bbf_view_def-before-14_7-or-15_2

2jungkook (dismissed):

test/JDBC/jdbc_schedule
shalinilohia50: Done.

2jungkook (approved):

PR #1090, opened 2023-01-19
Correct text of TIMESTAMP/ROWVERSION/@@DBTS messages

shalinilohia50:
### Description

The TIMESTAMP/ROWVERSION feature is off by default, and the corresponding escape hatch must be set to 'ignore' in order to switch on the feature. However, the error messages that are raised need to be updated to be more correct and clear. Especially the mention of 'currently' is raising incorrect impressions.

Changed the error messages for TIMESTAMP/ROWVERSION/@@DBTS

### Issues Resolved
Task: BABEL-3896

* **Use case based -**
Updated the expected error message
* **Boundary conditions -**
* **Arbitrary inputs -**
* **Negative test cases -**
* **Performance tests -**
* **Tooling impact -**
* **Client tests - **
JDBC tests

### DIFF
--- a/contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.3.0--2.4.0.sql
+++ b/contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.3.0--2.4.0.sql
@@ -982,6 +982,23 @@ RETURNS INTEGER AS
 'babelfishpg_tsql', 'object_id'
 LANGUAGE C STABLE;
 
+CREATE OR REPLACE FUNCTION sys.DBTS()
+RETURNS sys.ROWVERSION AS
+$$
+DECLARE
+    eh_setting text;
+BEGIN
+    eh_setting = (select s.setting FROM pg_catalog.pg_settings s where name = 'babelfishpg_tsql.escape_hatch_rowversion');
+    IF eh_setting = 'strict' THEN
+        RAISE EXCEPTION 'To use @@DBTS, set ''babelfishpg_tsql.escape_hatch_rowversion'' to ''ignore''';
+    ELSE
+        RETURN sys.get_current_full_xact_id()::sys.ROWVERSION;
+    END IF;
+END;
+$$
+STRICT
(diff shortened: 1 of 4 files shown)


rishabhtanwar29 (approved):

xhfanhe (approved):

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.3.0--2.4.0.sql
```diff
@@ -982,6 +982,23 @@ RETURNS INTEGER AS
...
+    eh_setting = (select s.setting FROM pg_catalog.pg_settings s where name = 'babelfishpg_tsql.escape_hatch_rowversion');
+    IF eh_setting = 'strict' THEN
+        RAISE EXCEPTION 'To use @@DBTS, set ''babelfishpg_tsql.escape_hatch_rowversion'' to ''ignore''';
+    ELSE
+        RETURN sys.get_current_full_xact_id()::sys.ROWVERSION;
+    END IF;
+END;
+$$
+STRICT
+LANGUAGE plpgsql;
```
Deepesh125, line 1000: I am seeing original definition of sys.DBTS where we have not granted select permission. Is it expected?

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.3.0--2.4.0.sql
```diff
@@ -982,6 +982,23 @@ RETURNS INTEGER AS
 'babelfishpg_tsql', 'object_id'
```
surendravishnoi, line 982: do we already have upgrade test for this? Otherwise, we should add as it is minor upgrade use case

shalinilohia50: https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/1097

PR #1097, opened 2023-01-20
Correct error message for @@DBTS

shalinilohia50:
### Description

Changed the error messages for @@DBTS. 
Added upgrade tests for the change.

### Issues Resolved
Task: BABEL-3896

* **Use case based -**
Added use case based tests
* **Boundary conditions -**
* **Arbitrary inputs -**
* **Negative test cases -**
Added tests
* **Performance tests -**
* **Tooling impact -**
* **Client tests -**


test/JDBC/upgrade/14_6/preparation/TestRowVersion-vu-prepare.sql
```diff
@@ -0,0 +1,108 @@
+-- Throws error
```
surendravishnoi, line 1: why we need a separate file? can we use the common file and add it to schedule. We do not need to call @@DBTS in prepare, verify should be sufficient

shalinilohia50: I called @@DBTS in prepare and verify to show the difference between the errors before and after the upgrade.

PR #1098, opened 2023-01-20
Fix query rewriting unproperly with multibyte characters

ghost:
### Description

Currently, Babelfish doesn't rewrite query with multibyte characters properly.

#### Analysis
Babelfish preprocess the query string and remove unsupported syntax before sending the query to PG backend. The implementation didn’t consider multibyte unicode characters, so when unicode characters are used ahead of the unsupported syntax, Babelfish will emit a broken query. In specific, character offset is used instead of byte offset during the character replacement. For example:
Input T-SQL : select "你好世界" from tbl with(nolock);
Executed SQL: select "你好世界" f            (nolock);

#### Solution
Consolidate all rewriting behaviors to PLtsql_expr_query_mutator. It’s more maintainable because there will only be one interface for query rewriting. We support Chinese unicode charset as identifier in this patch. We also fix test case of pg_hint_plan.

* **Use case based -**
babel_unicode_charset

* **Boundary conditions -**

* **Arbitrary inputs -**

* **Negative test cases -**

* **Performance tests -**
I have done performance test with long sql, which is about 4KB:
```
select t0.col from unicode_test t0 with(nolock)
  join unicode_test t1 with(nolock) on t0.col=t1.col
  join unicode_test t2 with(nolock) on t1.col=t2.col
  ... 
  join unicode_test t99 with(nolock) on t98.col=t99.col
```
The SQL contains 100 tables' joins. Antlr parser will rewrite this SQL for 100 times. 

We only count the time cost in antlr parser. 

Execut 100 times, remove the first uncached cost. 

Before:    23879.653 ms 

After:       23606.084 ms

* **Tooling impact -**

* **Client tests -**

### Check List

### DIFF
--- a/contrib/babelfishpg_tsql/antlr/TSqlLexer.g4
+++ b/contrib/babelfishpg_tsql/antlr/TSqlLexer.g4
@@ -1238,94 +1238,95 @@ fragment LETTER
     | '\u03a3'..'\u03ce'
     | '\u03d0'..'\u03d7'
     | '\u03da'..'\u03f3'
-//    | '\u0400'..'\u0481'  // Cyrillic
-//    | '\u048c'..'\u04c4'
-//    | '\u04c7'..'\u04c8'
-//    | '\u04cb'..'\u04cc'
-//    | '\u04d0'..'\u04f5'
-//    | '\u04f8'..'\u04f9'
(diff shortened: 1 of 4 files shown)


markrui3:
@macwoody

macwoody (approved):

PR #1120, opened 2023-01-24
BABEL-3802 Bigint/Int/Smallint/Tinyint Power Function

pratikzode:
This Jira is a part of BABEL-736.
At first in Babel, the Transact-SQL functions power always return float instead of returning the same type in SQL server. In this Jira we have added Bigint, Int, Smallint and Tinyint to power function. Have added the test case to show Bigint, Int, Smallint and Tinyint to power. Now this 3 is giving same output as SQL server.

Task: BABEL-3802
Signed-off-by: pratikzode <pratik.zode@gmail.com>


contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -29,6 +29,7 @@
 #include "utils/float.h"
 #include "utils/xid8.h"
 #include <math.h>
+#include "utils/numeric.h"
```
macwoody, line 32: header files suppose to be ordered alphabetically.

PR #1135, opened 2023-01-25
Support and re-design UPDATE/DELETE

xhfanhe:
This commit re-designs UPDATE/DELETE DML statement support, including but not restricted to
1. Plain UPDATE/DELETE
2. UPDATE/DELETE with JOIN (ANSI or non-ANSI, all types of JOIN)
3. UPDATE/DELETE with correlation name

Corresponding test cases are also added in this commit.

OUTPUT clause and TOP clause in UPDATE/DELETE will be re-designed using the new approach
in future commits.

Task: BABEL-1330/1875/2675/3091/3684/3685/3775
Signed-off-by: Xiaohui Fanhe <hexiaohu@amazon.com>

### Check List


contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -1215,6 +1236,189 @@ pre_transform_target_entry(ResTarget *res, ParseState *pstate,
 			res->name = alias;
 		}
 	}
+	/* Update table set qualified column name, resolve qualifiers here */
```
xhfanhe, line 1237: This is column name handling in APG, controlled by `ansi_qualified_update_set_target`. Adding this in open source because it's required by TSQL UPDATE/DELETE.

test/JDBC/expected/babel_delete.out
```diff
@@ -207,7 +205,9 @@ go
 
 ~~ROW COUNT: 3~~
 
-~~ROW COUNT: 1~~
+~~ERROR (Code: 33557097)~~
+
+~~ERROR (Message: ctid is NULL)~~
```
xhfanhe: This is a regression that will happen in certain outer join cases, haven't got a full picture of the reason or pattern yet. Will file a Jira for this as bug fix.

.github/scripts/clone_engine_repo.conf
```diff
@@ -1 +1 @@
-DEFAULT_BRANCH=BABEL_2_X_DEV__PG_14_X
+DEFAULT_BRANCH=jira-babel-3682
```
xhfanhe: Temp change to test with engine PR change. Will remove this after it's ok to merge in whole.

contrib/babelfishpg_tsql/src/pltsql.h
```diff
@@ -1995,6 +1995,10 @@ extern Oid tsql_get_constraint_oid(char *conname, Oid connamespace, Oid user_id)
 extern Oid tsql_get_proc_oid(char *proname, Oid pronamespace, Oid user_id);
 extern char** split_object_name(char *name);
 extern bool is_schema_from_db(Oid schema_oid, Oid db_id);
+extern void pltsql_update_target_table_alias(RangeVar **target_table, List *fromClause);
+extern void pltsql_cte_update_target_table_alias(WithClause *withClause);
```
macwoody: I wonder if this is the right place? 

pltsql.h is already very messy.

contrib/babelfishpg_tsql/src/pltsql_utils.c
```diff
@@ -34,6 +35,10 @@ bool pltsql_suppress_string_truncation_error(void);
 bool is_tsql_any_char_datatype(Oid oid); /* sys.char / sys.nchar / sys.varchar / sys.nvarchar */
 bool is_tsql_text_ntext_or_image_datatype(Oid oid);
 
+static bool find_target_alias_in_tableref(RangeVar **target, Node *tblref);
+static bool find_relation_in_tableref(RangeVar *target, Node *tblref);
```
macwoody: again, pltsql_utils.c seems to be messy place as well.

contrib/babelfishpg_tsql/src/pltsql_utils.c
```diff
@@ -1153,4 +1156,226 @@ bool is_schema_from_db(Oid schema_oid, Oid db_id)
...
+			return true;
+		}
+		else if (rv->alias && pg_strcasecmp((*target)->relname, rv->alias->aliasname) == 0)
+		{
+			if ((*target)->schemaname)
+				ereport(ERROR,
+						(errcode(ERRCODE_SYNTAX_ERROR),
+						 errmsg("The correlation name \'%s\' has the same exposed name as table \'%s.%s\'.", 
+								rv->alias->aliasname, (*target)->schemaname, (*target)->relname)));
+			*target = rv;
```
macwoody: Just shallow copy?

contrib/babelfishpg_tsql/src/pltsql_utils.c
```diff
@@ -1153,4 +1156,226 @@ bool is_schema_from_db(Oid schema_oid, Oid db_id)
...
+	/* 
+ 	 * Currently we only consider RangeVar and JoinExpr cases. In the
+ 	 * future, if there are concrete use cases, we'll add support for 
+ 	 * more table reference types
+ 	 */
+	return false;
+}
+
+void
+pltsql_update_table_reference(RangeVar *target_table, List *fromClause, ParseState *pstate)
```
macwoody: This behavior is to revert part of addNSItemToQuery 
I wonder if it is possible that we could call 
setTargetTable (alsosource = false).

xhfanhe: Then instead of pre_parse_analyze and post_set_target_table, we would need to add a hook around setTargetTable. 
```
if (hook)
   /* 
    * if duplication exists, call setTargetTable(alsosouorce=false) 
    * and reset r_table
    */
   qry->resultRelation = (*hook) () 
else
   qry->resultRelation = setTargetTable()
```
Yes this might be a better way, and this also avoid dealing with CTEs

contrib/babelfishpg_tsql/src/pltsql_utils.c
xhfanhe: Should be fine because target relation (stmt->relation) and the source of RangeVar (stmt->fromClause) shares same parent structure. Will let PG do the management.

contrib/babelfishpg_tsql/src/pltsql.h
xhfanhe: Maybe put them in hooks.c and hooks.h instead?

macwoody (dismissed):

macwoody (approved):

PR #1160, opened 2023-01-31
Disable Savepoints for statements inside a function

KushaalShroff:
Currently we have savepoints for every statement inside a function in order to rollback safely in case of any errors. This was done to not destroy portals amidst function execution. These Savepoints added around 40% perfomance overhead to function execution.
We have verified that any errors that occur inside a function shall rollback the entire function execution, so we can safely disable Savepoints for statements inside a function and let the function call(exec stmt) handle the error.

Signed-off-by: Kushaal Shroff <kushaal@amazon.com>

### Issues Resolved
BABEL-3925

**Performance Testing**
For the following query:
```
create table numbers (a int primary key)
go
insert into numbers select * from generate_series(1,100000)
go
drop function if exists f_per_row
go
create function f_per_row (@num int)
returns int
AS
begin
    declare @i int
    select @i = 1
    select @i = 2
    select @i = 3
    select @i = 4
    select @i = 5
    select @i = 6
    select @i = 7
    select @i = 8
    select @i = 9
    select @i = 10
    return @i
end
go
select count(*) from (
    select dbo.f_per_row(a) as c
    from numbers
) as t
go
```

Before the Changes: Flamegraph perf-tsql-sp-orig-func
```
select count(*) from (    select dbo.f_per_row(a) as c    from numbers ) as t
go
count
-----
100000



(1 rows affected)
4096:1: 21504:21504.0:0.1
```
After the changes: Flamegraph perf-tsql-sp-func1

```
select count(*) from (    select dbo.f_per_row(a) as c    from numbers ) as t
go
count
-----
100000

(1 rows affected)
4096:1:12471:12471.0:0.1
```
Thus we see a speedup of 40%

* **Use case based -**

* **Boundary conditions -**

* **Arbitrary inputs -**

* **Negative test cases -**

* **Performance tests -**

* **Tooling impact -**

* **Client tests -**

### Check List

### DIFF
--- a/contrib/babelfishpg_tsql/src/iterative_exec.c
+++ b/contrib/babelfishpg_tsql/src/iterative_exec.c
@@ -1133,11 +1133,9 @@ int dispatch_stmt_handle_error(PLtsql_execstate *estate,
 		 * is active to handle undo of failed command
 		 * We do not start savepoint for batch commands as
 		 * error handling must be taken care of at statement
-		 * level
-		 * For RO functions start savepoint even when transaction
-		 * is not active to retain top level portals. A transaction
-		 * rollback will cleanup portal data which can lead to
-		 * problems when control returns back to portal level
+		 * level.
+		 * For statements inside an RO functions we do not start
+		 * savepoints and let the caller be responsible for handling the error.
 		 */
 		if (!ro_func && !pltsql_disable_internal_savepoint && !is_batch_command(stmt) && IsTransactionBlockActive())
 		{
@@ -1184,13 +1182,6 @@ int dispatch_stmt_handle_error(PLtsql_execstate *estate,
 		ErrorData *edata;
 		int last_error;
 		bool error_mapped;
-
-		if (ro_func)
-		{
-			MemoryContextSwitchTo(cur_ctxt);
-			PG_RE_THROW();
-		}
-
 		support_tsql_trans = pltsql_support_tsql_transactions();
 
 		/* Close trigger nesting in engine */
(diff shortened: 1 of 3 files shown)


contrib/babelfishpg_tsql/src/iterative_exec.c
```diff
@@ -1230,6 +1228,14 @@ int dispatch_stmt_handle_error(PLtsql_execstate *estate,
 
 		MemoryContextSwitchTo(cur_ctxt);
 
+		/* In case of RO functions we want the caller to handle errors. */
+		if (ro_func)
```
surendravishnoi, line 1232: we want to validate that error handing inside functions is same as that of SQL server. Given that no test has failed, we are missing that coverage. We should cover cases where a statement inside function throws various error codes like
- stmt terminating error
- batch terminating error
- transaction rollback error

We should also cover xact_abort ON/OFF, try/catch etc. The behaviour should match with SQL server.

In addition, there should be some generic testing where
func1 -> func2 -> func3 -> error
proc1 -> proc2 -> func1 -> error
trigger -> func1 -> error
tsqlfunc -> pgfunc -> error
pgfunc -> tsqlfunc -> error

Some of these might already be there but we should validate the same.

KushaalShroff: xact_abort ignore test cases are in mostly DDL phase so we cant have repro for function execution (function cant have dml either)
BABEL-3101.sql and TestErrorFunctions.sql have function/procedure nesting test cases so I am skipping those.

surendravishnoi (approved):

PR #1173, opened 2023-02-01
Support the function OBJECT_DEFINITION()

shalinilohia50:
### Description
Implement a function OBJECT_DEFINITION(object_id) which returns the Transact-SQL source text of the definition of a specified object.

List of objects supported:

Check constraints
Default constraints
Procedure
Scalar function
DML trigger - TODO: After BABEL-3927 is fixed, we get the correct trigger definition.
Views
Inline table valued functions
Table valued functions

Task: BABEL-3919

### Issues Resolved
BABEL-3919

* **Use case based -**
Added use case based tests in JDBC framework

* **Boundary conditions -**
Added edge cases where the object_id is either incorrect or NULL

* **Arbitrary inputs -**
* **Negative test cases -**
* **Performance tests -**
* **Tooling impact -**
* **Client tests -**


test/JDBC/upgrade/latest/schedule
```diff
@@ -391,4 +391,5 @@ BABEL-3828
 BABEL-3844
 test_windows_alter_login
 sys-systypes
+BABEL_OBJECT_DEFINITION
```
rishabhtanwar29, line 398: Are we going to support it in PG14 as well? If yes then we should also add this to 14_7/schedule file.

test/python/expected/upgrade_validation/expected_dependency.out
```diff
@@ -497,6 +497,7 @@ Function sys.nvarchar(sys.nvarchar,integer,boolean)
 Function sys.nvarchar_larger(sys.nvarchar,sys.nvarchar)
 Function sys.nvarchar_smaller(sys.nvarchar,sys.nvarchar)
 Function sys.nvarchar_sqlvariant(sys.nvarchar)
+Function sys.object_definition(integer)
```
rishabhtanwar29: Let's use this function in a view to get rid of this.

test/JDBC/expected/BABEL_OBJECT_DEFINITION-vu-verify.out
```diff
@@ -0,0 +1,111 @@
+-- Throws an error
+SELECT OBJECT_DEFINITION();
```
rishabhtanwar29, line 3: Please verify if it is the correct T-SQL behavior if you haven't already.

contrib/babelfishpg_tsql/sql/sys_functions.sql
```diff
@@ -3322,6 +3322,38 @@ $$
...
 LANGUAGE sql PARALLEL SAFE STABLE;
 
+CREATE OR REPLACE FUNCTION OBJECT_DEFINITION(IN object_id INT)
+RETURNS sys.NVARCHAR(4000)
+AS $$
+DECLARE
+    type char(2);
+BEGIN
+
+    IF NOT EXISTS(SELECT ao.object_id FROM sys.all_objects ao WHERE ao.object_id = $1)
```
rishabhtanwar29: Can we save this `all_objects` view lookup? I am thinking of directly searching the OID as follows:

1. Search in `sys.check_constraints` return if we get non-null result otherwise,
2. Search in `sys.default_constraints` return if we get non-null result otherwise,
3. Search in `sys.all_sql_modules` return if we get non-null result otherwise,
4. return null
I am asking this since `sys.all_objects` lookup can be quite expensive.

test/JDBC/expected/BABEL_OBJECT_DEFINITION-vu-prepare.out
```diff
@@ -0,0 +1,57 @@
+CREATE DATABASE object_definition_db;
```
rishabhtanwar29, line 2: I remember we have a test file for object definitions of various kind of objects, should we move these tests to that file instead if possible? We will not have to create new objects in that case.

shalinilohia50: Those files are named specifically on the view names and there is no single file where all of these objects supported by object_definition() are created. IMO, it's better to have a separate test file.

test/JDBC/expected/BABEL_OBJECT_DEFINITION-vu-verify.out
shalinilohia50: Yes, this is the expected behaviour. T-SQL throws an error if the OBJECT_DEFINITION() has no argument.

test/python/expected/upgrade_validation/expected_dependency.out
shalinilohia50: Since we don't support object_definition in previous versions, we need to create a new test file `object_def_dep-vu-*` which will run only in 14_7 and latest. As discussed offline, Let's not create unnecessary files just to add a simple dependency test.

rishabhtanwar29 (dismissed):

HarshLunagariya (dismissed):

shardgupta (approved):

PR #1175, opened 2023-02-01
Support the function OBJECT_DEFINITION()

shalinilohia50:
### Description
Implement a function OBJECT_DEFINITION(object_id) which returns the Transact-SQL source text of the definition of a specified object.

List of objects supported:

1. Check constraints
2. Default constraints
3. Procedure
4. Scalar function
5. DML trigger - TODO: After BABEL-3927 is fixed, we get the correct trigger definition.
6. Views
7. Inline table valued functions
8. Table valued functions

Task: BABEL-3919

### Issues Resolved
BABEL-3919

* **Use case based -**
Added use case based tests in JDBC framework

* **Boundary conditions -**
Added edge cases where the object_id is either incorrect or NULL

* **Arbitrary inputs -**
* **Negative test cases -**
* **Performance tests -**
* **Tooling impact -**
* **Client tests -**


rishabhtanwar29 (dismissed):
We should update the description with the object types which are covered with the current implementation.

contrib/babelfishpg_tsql/sql/sys_functions.sql
```diff
@@ -3377,3 +3377,30 @@ $$
     SELECT current_setting('application_name');
 $$
 LANGUAGE sql PARALLEL SAFE STABLE;
+
+CREATE OR REPLACE FUNCTION OBJECT_DEFINITION(IN object_id INT)
```
HarshLunagariya, line 3381: Is it supposed to work for cross-db calls also? If yes then using sys views won't be sufficient.

rishabhtanwar29 (dismissed):

contrib/babelfishpg_tsql/sql/sys_functions.sql
```diff
@@ -3377,3 +3377,30 @@ $$
...
+            IF (definition IS NULL)
+            THEN
+                RETURN NULL;
+            END IF;
+        END IF;
+    END IF;
+
+    RETURN definition;
+END;
+$$
```
HarshLunagariya, line 3404: It should throw error when caller doesn't have permission on object acc. documentation. Please add test cases for the same.

HarshLunagariya (changes requested):

contrib/babelfishpg_tsql/sql/sys_functions.sql
shalinilohia50: No, it should show NULL for cross-db call. Also, added test showing cross-db call.

contrib/babelfishpg_tsql/sql/sys_functions.sql
shalinilohia50: The OBJECT_DEFINITION() function doesn't throw error in that case. I have added tests for that case, it returns NULL.
From T-SQL doc:
```
Returns NULL on error or if a caller does not have permission to view the object.

HarshLunagariya (approved):

rishabhtanwar29 (approved):

shardgupta (approved):

PR #1179, opened 2023-02-01
Allow FOR XML/JSON to be able to return zero rows

Sairakan:

### Description

If FOR XML/JSON are used on a query that returns 0 rows, the result is supposed to be an empty result with 0 rows.
In order to mimic this behavior, we need to use an additional set-returning function and pass the result of the aggregate
function call to that SRF so that returning 0 rows is possible.

### Issues Resolved

BABEL-3569, BABEL-3690

regression tests + new tests for empty rowsets

### Check List

### DIFF
--- a/contrib/babelfishpg_tsql/sql/sys_functions.sql
+++ b/contrib/babelfishpg_tsql/sql/sys_functions.sql
@@ -59,7 +59,7 @@ ELSE
 END IF;
 END;
 $$
-LANGUAGE plpgsql;
+LANGUAGE plpgsql IMMUTABLE;
 
 CREATE OR REPLACE FUNCTION sys.tsql_select_for_xml_text_result(res NTEXT)
 RETURNS setof NTEXT AS
@@ -72,7 +72,7 @@ ELSE
 END IF;
 END;
 $$
-LANGUAGE plpgsql;
+LANGUAGE plpgsql IMMUTABLE;
 
 -- SELECT FOR JSON
 CREATE OR REPLACE FUNCTION sys.tsql_query_to_json_sfunc(
@@ -116,7 +116,7 @@ ELSE
 END IF;
 END;
 $$
-LANGUAGE plpgsql;
+LANGUAGE plpgsql IMMUTABLE;
 
 -- User and Login Functions
 CREATE OR REPLACE FUNCTION sys.user_name(IN id OID DEFAULT NULL)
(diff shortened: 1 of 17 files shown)


contrib/babelfishpg_tsql/sql/sys_functions.sql
```diff
@@ -48,6 +48,32 @@ CREATE OR REPLACE AGGREGATE sys.tsql_select_for_xml_text_agg(
...
+$$
+BEGIN
+IF res IS NOT NULL THEN
+    return next res;
+ELSE
+    return;
+END IF;
+END;
+$$
+LANGUAGE plpgsql;
```
2jungkook: Can it be IMMUTABLE or STABLE?

contrib/babelfishpg_tsql/src/backend_parser/gram-tsql-epilogue.y.c
```diff
@@ -1276,6 +1276,17 @@ TsqlForXMLMakeFuncCall(TSQL_ForClause* forclause)
...
 	fc = makeFuncCall(func_name, func_args, COERCE_EXPLICIT_CALL, -1);
 
+	/* In SQL Server if the result is empty then 0 rows are returned. Unfortunately it is not
+	 * possible to mimic this behavior solely using an aggregate, so we use an additional SRF
+	 * and pass the result to that function so that returning 0 rows is possible.
+	 */
+	func_name= list_make2(makeString("sys"), 
+							makeString(return_xml_type ? 
+								"tsql_select_for_xml_result" : 
+								"tsql_select_for_xml_text_result"));
```
2jungkook, line 1286: Can we have test cases for both `tsql_select_for_xml_result` and `tsql_select_for_xml_text_result` cases with 0 row output?

contrib/babelfishpg_tsql/sql/sys_functions.sql
Sairakan: I wasn't sure originally, but I think it should be okay to set it to IMMUTABLE

contrib/babelfishpg_tsql/src/backend_parser/gram-tsql-epilogue.y.c
Sairakan: sure, the current test is for text, let me add a test for xml as well

macwoody (approved):

PR #1180, opened 2023-02-01
Add tests for scope_identity on index

wboettge:
### Description

Currently, a problem arises when using scope_identity to lookup on an index of identities. Babelfish implemented identity columns using INT sequences, but scope_identity returns a numeric type. Numeric has higher precedence than int, so the index column is implicitly cast to a numeric. However, this means queries that should use an index lookup instead use a sequential scan, significantly degrading performance.

This change adds tests for an engine change that fixes this issue be rewriting specific queries to use babelfish_get_last_identity instead of scope_identity.

### Issues Resolved

Task: BABEL-3384

### Check List

### DIFF
--- a/test/JDBC/sql_expected/BABEL-IDENTITY.out
+++ b/test/JDBC/sql_expected/BABEL-IDENTITY.out
@@ -734,6 +734,14 @@ numeric
 ~~END~~
 
 
+select set_config('babelfishpg_tsql.explain_costs', 'off', false)
+go
+~~START~~
+text
+off
+~~END~~
+
+
 SET babelfish_showplan_all ON;
 go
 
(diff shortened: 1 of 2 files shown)


test/JDBC/sql_expected/BABEL-IDENTITY.out
```diff
@@ -709,6 +709,142 @@ int#!#int
...
+  Filter: ((id <= babelfish_get_last_identity()) OR (mycol = 11))
+~~END~~
+
+
+SELECT num_index, mycol FROM dbo.test_numeric_index WHERE num_index = scope_identity();
+go
+~~START~~
+text
+Query Text: SELECT num_index, mycol FROM dbo.test_numeric_index WHERE num_index = scope_identity()
+Index Scan using test_numeric_index_pkey on test_numeric_index  (cost=0.16..8.18 rows=1 width=24)
```
macwoody: We usually don't include cost in the expected result because it is subject to change in future. 
plz check with jungkook whether there is a way to hide cost in BBF explain plan .

macwoody (approved):

PR #1210, opened 2023-02-03
Fix outer join ctid is null error in update/delete

wboettge:
### Description
This fixes an issue with update and delete statements using outer joins in their from clauses. Previously, if the target table was on the 'other side' of an outer join (i.e. right side of a left outer join) an Babelfish would give a 'ctid is null' error while the statement would work in T-SQL.

To fix this issue, the queries is checked for update/delete targets present in from clauses on these particular sides of join statements. If so, the query is re-written by adding a WHERE clause that filters rows where the target's ctid is null. The offending queries now match the T-SQL behavior.

### Issues Resolved

Task: BABEL-3910

* **Use case based -**
Checks the scenario in the jira, as well as with other types of joins.

* **Boundary conditions -**
Checks with nested joins. Checks with aliasing tables

* **Arbitrary inputs -**
N/A

* **Negative test cases -**
Tests confirm other cases of update / delete with joins aren't affected.

N/A

N/A

* **Performance tests -**
N/A

* **Tooling impact -**
N/A

* **Client tests -**
N/A

### Check List

### DIFF
--- a/test/JDBC/input/dml/babel_delete.sql
+++ b/test/JDBC/input/dml/babel_delete.sql
@@ -94,9 +94,113 @@ delete babel_2020_delete_t1 from babel_2020_delete_t1 x left outer join babel_20
 go
 
 -- will be tracked in BABEL-3910
---exec babel_2020_delete_ct;
---delete babel_2020_delete_t1 from babel_2020_delete_t2 left outer join babel_2020_delete_t1 x on babel_2020_delete_t2.a = x.a;
---go
+drop procedure if exists babel_3910_init_tables
+go
+
+create procedure babel_3910_init_tables as
+begin
+    drop table if exists t1_3910
+    create table t1_3910 (a int)
+    insert into t1_3910 values (1), (2), (3), (4), (NULL)
+    drop table if exists t2_3910
+    create table t2_3910 (a int)
+    insert into t2_3910 values (2), (3), (4), (5), (NULL)
+    drop table if exists t3_3910
+    create table t3_3910 (a int)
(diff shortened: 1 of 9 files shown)


test/JDBC/input/dml/babel_delete.sql
```diff
@@ -94,9 +94,113 @@ delete babel_2020_delete_t1 from babel_2020_delete_t1 x left outer join babel_20
...
+    drop table if exists t3_3910
+    create table t3_3910 (a int)
+    insert into t3_3910 values (3), (4), (5), (6), (NULL)
+    drop table if exists t4_3910
+    create table t4_3910 (a int)
+    insert into t4_3910 values (4), (5), (6), (7), (NULL)
+end
+go
+
+exec babel_3910_init_tables;
```
macwoody, line 117: could you please also add a few 3-table join scenarios ?

test/JDBC/input/dml/babel_delete.sql
```diff
@@ -94,9 +94,113 @@ delete babel_2020_delete_t1 from babel_2020_delete_t1 x left outer join babel_20
...
+delete t1_3910 from t2_3910 left outer join t1_3910 on t2_3910.a = t1_3910.a;
+go
+
+select a from t1_3910;
```
macwoody, line 121: there is also a wired way of using join like this  (comma separated list that indicating inner join) 

update t3 set ... from T1, T2 left outer join t3 on ... where t1.a = t2.a

macwoody (approved):

PR #1233, opened 2023-02-07
Create table or proc should throw error for declaring non-permissible…

shalinilohia50:
### Description
Formerly, when specifying NUMERIC(precision, scale), the scale had to be in the range [0, precision], which was per SQL spec. This commit extends the range of allowed scales to [-1000, 1000], independent of the precision (whose valid range remains [1, 1000]). A negative scale implies rounding before the decimal point. For example, a column might be declared with a scale of -3 to round values to the nearest thousand. Note that the display scale remains non-negative, so in this case the display scale will be zero, and all digits before the decimal point will be displayed. A scale greater than the precision supports fractional values with zeros immediately after the decimal point.

From SQL spec aspect which means checking explicitly whether 0 <= scale <= precision and 1 <= precision <= 38, which was earlier done by underlying PG code for any numeric/decimal input data.

Along with the above change, it is found that all variable length datatypes like numeric, decimal, time, datetime2, datetimeoffset need explicit checking on their permissible datalength at the time of table or stored procedure creation and generate error if required. Before fix, Babelfish was not complaining for the above behaviour.

Task:  BABEL-3793
Co-authored-by:  Satarupa Biswas (satarupb@amazon.com)

### Issues Resolved
BABEL-3793

* **Use case based -**

* **Boundary conditions -**

* **Arbitrary inputs -**

* **Negative test cases -**

* **Performance tests -**

* **Tooling impact -**

* **Client tests -**


jsudrik (approved):
Identical changes to 3X.

PR #1252, opened 2023-02-08
Add error handler for T-SQL Linked Servers

thephantomthief:
### Description

The message handler for linked servers was not handling errors encountered in the TDS client library itself (eg. cannot connect to remote server etc.). In this commit, we introduce an error handler that will relay such errors back to client instead of fatally exiting.

Task: BABEL-3974

* **Use case based -** Yes

* **Arbitrary inputs -** Yes

* **Performance tests -** N/A

* **Tooling impact -** N/A

* **Client tests -** N/A

### Check List


contrib/babelfishpg_tsql/src/linked_servers.c
```diff
@@ -88,6 +88,33 @@ linked_server_msg_handler(LinkedServerProcess lsproc, int error_code, int state,
...
+
+	appendStringInfo(
+		&buf,
+		"TDS client library error: DB #: %i, DB Msg: %s, OS #: %i, OS Msg: %s, Level: %i",
+		db_error,
+		db_err_str ? db_err_str : "",
+		os_error,
+		os_err_str ? os_err_str : "",
+		severity
+	);
```
Deepesh125: we should construct error message in a way that includes key: value thing only if value is valid. Otherwise there is not point of adding empty string.

contrib/babelfishpg_tsql/src/linked_servers.c
```diff
@@ -88,6 +88,33 @@ linked_server_msg_handler(LinkedServerProcess lsproc, int error_code, int state,
...
+
+	ereport(ERROR,
+			(errcode(ERRCODE_FDW_UNABLE_TO_CREATE_EXECUTION),
+			errmsg("%s", buf.data)));
+
+	return LS_INT_CANCEL;
```
Deepesh125, line 123: what is meaning of this?

contrib/babelfishpg_tsql/src/linked_servers.c
thephantomthief: I kept it deliberately like this because this is how it gets handled in tds_fdw as well.

thephantomthief: https://github.com/tds-fdw/tds_fdw/blob/c24be29f5f4a9c5ba87b8e025fc898637e589583/src/tds_fdw.c#L3937)

contrib/babelfishpg_tsql/src/linked_servers.c
thephantomthief: https://www.freetds.org/reference/a00554.html#ga302a4ce6559b81808e3be51615cd1b67. See INT_CANCEL

Deepesh125: But we have used ereport here so control will never reach back to tds_fdw, right?

thephantomthief: Agreed but we should return based on the behaviour intended. Theoretically, the execution should be suspended (which is what INT_CANCEL indicates). Returning 0 isn't correct because it maps to INT_EXIT which is Sybase semantics.

Deepesh125 (approved):

PR #1271, opened 2023-02-13
Fix table variable usage with sp_executesql

Sairakan:

### Description

In Babelfish, table variables are always passed by value. Postgres, however, will default to passing table variables by reference, leading to server crashes due to invalid memory accesses. This was previously worked around by setting the `isnull` flag to true when setting up a table variable. However, this leads to an error if a table variable is used with `sp_executesql`, as the PG executor will then use the `isnull` flag to fill in null values for the parameter. 

This commit resolves the issue in 2 steps: 
1. We properly set the `isnull` flag to false (since the datum is not empty) 
2. We add additional code in `bbf_ProcessUtility()` to detect when a TSQL table type is created and modify the pg_type entry to set the column `typbyval` to true. With this, future accesses to the table variable will properly see that the type is pass-by-value.

### Issues Resolved

BABEL-3967

* **Use case based -**
test TVF, table type, table variable

* **Boundary conditions -**
N/A

* **Arbitrary inputs -**
tested

* **Negative test cases -**
N/A

tested

tested

* **Performance tests -**
N/A

* **Tooling impact -**
none

* **Client tests -**
repro from JIRA

### Check List


contrib/babelfishpg_tsql/src/pl_exec.c
```diff
@@ -6859,6 +6859,7 @@ exec_eval_datum(PLtsql_execstate *estate,
 				*typeid = tbl->tbltypeid;
 				*typetypmod = -1;
 				*value = CStringGetDatum(tbl->tblname);
+				*isnull = false;
```
2jungkook, line 6862: Please revive the previous comment as well (https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/1267/files) if it is still valid.

Sairakan: That comment isn't valid - it references setting `isnull` to true to force pass-by-val during the planner/executor, which is not applicable here since we're actually setting it to false (which it should have been in the first place since the value is not null)

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -3293,11 +3293,43 @@ static void bbf_ProcessUtility(PlannedStmt *pstmt,
...
+					memset(nulls, 0, sizeof(nulls));
+					memset(replaces, 0, sizeof(replaces));
+					replaces[Anum_pg_type_typbyval - 1] = true;
+					values[Anum_pg_type_typbyval - 1] = BoolGetDatum(true);
+
+					newtup = heap_modify_tuple(tup, RelationGetDescr(catalog), values, nulls, replaces);
+					CatalogTupleUpdate(catalog, &newtup->t_self, newtup);
+
+					/* Clean up */
+					ReleaseSysCache(tup);
```
macwoody: Unless above codes are exception free, need an exception handling to avoid tup/opened table leak

Sairakan: I've checked in a few other places in the code, and none of them have any exception handling around their catalog table accesses. I assume that the engine has logic in place to properly free locks/table fds when exceptions are raised.

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--3.0.0--3.1.0.sql
```diff
@@ -56,6 +56,14 @@ LANGUAGE plpgsql;
  * final behaviour.
  */
 
+-- 3.1 introduces a change to how table types are stored in pg_type. While MVU forces the catalog
```
2jungkook: Please add the same in 2.3.0--2.4.0 upgrade script because we'd like to see all upgrade scripts in the latest branch.

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--3.0.0--3.1.0.sql
```diff
@@ -56,6 +56,14 @@ LANGUAGE plpgsql;
...
+-- to be regenerated, mVU (minor version upgrade) does not, so we need to manually fix it here.
+UPDATE pg_type SET typbyval = 't' 
+FROM sys.babelfish_namespace_ext AS ns 
+WHERE typtype = 'c' AND
+    typacl IS NOT NULL AND
+    typacl::TEXT LIKE '%' || ns.nspname || '%';
```
2jungkook: Using `like '%nspname%'` is not 100% safe because non-babelfish namespace could contain some babelfish namespace strings. Let's use an exact match using `babelfish_namespace_ext.nspname`, `pg_namespace.nspname`, `pg_namespace.oid` and `pg_type.typnamespace`.

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -3329,6 +3335,45 @@ static void bbf_ProcessUtility(PlannedStmt *pstmt,
...
+	HeapTuple	tup;
+	
+	Datum		values[Natts_pg_type];
+	bool		nulls[Natts_pg_type];
+	bool		replaces[Natts_pg_type];
+	HeapTuple	newtup;
+
+	/*
+		* Table types need to set the typbyval column in pg_type to 't'
+		*/
```
2jungkook: indentation

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -3329,6 +3335,45 @@ static void bbf_ProcessUtility(PlannedStmt *pstmt,
...
+	memset(values, 0, sizeof(values));
+	memset(nulls, 0, sizeof(nulls));
+	memset(replaces, 0, sizeof(replaces));
+	replaces[Anum_pg_type_typbyval - 1] = true;
+	values[Anum_pg_type_typbyval - 1] = BoolGetDatum(byval);
+
+	newtup = heap_modify_tuple(tup, RelationGetDescr(catalog), values, nulls, replaces);
+	CatalogTupleUpdate(catalog, &newtup->t_self, newtup);
+
+	/* Clean up */
```
2jungkook, line 3370: what about `newtup`?

test/JDBC/expected/table-variable-vu-verify.out
```diff
@@ -204,3 +204,12 @@ int
 1
 ~~END~~
 
+
+-- double-check that the underlying type for table_variable_vu_type is pass-by-val
+select typbyval from pg_type where typname = 'table_variable_vu_type';
```
2jungkook, line 209: For pg_type field checks, please also add the following test cases:
1. `create type custom_schema.table_type as table ...`
2. `declare @tab_var as table ...`
3. table types created by TVF

contrib/babelfishpg_tsql/src/pl_handler.c
Sairakan: the other code examples I borrowed this implementation from do not clean newtup, so I am not sure if it is necessary/safe to manually do so in here.

test/JDBC/expected/table-variable-vu-verify.out
Sairakan: Table types from TVF are covered in TestTableType MVU tests:
```
create function TestTableType_vu_prepare_func1 (@number int, @tv TestTableType_vu_prepare_t4 READONLY) returns table as return (
select *, @number from @tv 
);
go
```
I will add tests for the other two cases

2jungkook: TestTableType test does not check typbyval field. Please check typbyval field directly, e.g.,
```
select typbyval from pg_type where typname like '@testtabletype_vu_prepare_func1%';
go

declare @tableVar as table (x int)
insert into @tableVar values (1),(2),(3)
select * from @tableVar;
select typbyval from pg_catalog.pg_type where typname like '@tablevar%';
go
```

Sairakan: Added a check for @tableVar. However, it appears that TVFs do not generate a pg_type entry, as their return type is set to `record` type, which is a built-in PG type.

Sairakan: Added example with TVF

2jungkook (approved):

macwoody (approved):

macwoody (approved):

PR #1285, opened 2023-02-16
BABEL-3478 Support the ROWCOUNT_BIG() function. 

pratikzode:
We have implemented function rowcount_big()
The ROWCOUNT_BIG() function is used to return the number of rows affected by the last statement that modified data, such as an INSERT, UPDATE, or DELETE statement. The expected behavior of ROWCOUNT_BIG() is to return a big integer value representing the number of rows affected by the previous data modification statement.

BBF
1> SELECT ROWCOUNT_BIG();
2> GO
rowcount_big

3

Task: BABEL-3802


test/JDBC/expected/BABEL-3478-vu-verify.out
```diff
@@ -0,0 +1,99 @@
+SELECT * FROM BABEL_3478_t1 ORDER BY LastName ASC;
```
macwoody: you have two "Doe"s, sorting order of these two rows are not guaranteed. 
Rule of thumb: we are adding "order by " to create consistent result order. 
For your case, consistent order can be order by "first and last name"

test/JDBC/upgrade/latest/schedule
```diff
@@ -407,5 +407,6 @@ BABEL-3657
 test_windows_alter_user
 BABEL-733
 BABEL-3938
+BABEL-3478
```
surendravishnoi, line 410: this only for Major version upgrade without any minor (3.1 to 3.1). If there is no schedule for 3 to 3.1, add it one of the MVU schedule like 14.7 to 3.1.

PR #1288, opened 2023-02-19
Implement CONTEXT_INFO() and SET CONTEXT_INFO

ghost:
Implement CONTEXT_INFO() and SET CONTEXT_INFO

Description
===========
CONTEXT_INFO is a session local parameter of type varbinary(128). It can be set
by 'SET CONTEXT_INFO 0x123456' and can be get by 'SELECT CONTEXT_INFO()'. It can
also be get with sys.sysprocesses, sys.dm_exec_requests, sys.dm_exec_sessions
views.

Solution
========
CONTEXT_INFO is similar to HOST_NAME/LANGUAGE, all of them are session local.
The difference is that CONTEXT_INFO supports int/numeric/float parameter and
dose implicit type casts.
We add a procedure bbf_set_context_info(IN context_info sys.VARBINARY(128)) and
rewrite 'SET CONTEXT_INFO 0x123456' to
'CALL bbf_set_context_info(convert(varbinary(128), 0x123456))'.

Limitation
==========
Some type casts are not supported in Babelfish, such as type cast from numeric
to varbinary. Some type casts have their own rules, e.g. integer constant
greater than 2,147,483,647 is converted to decimal instead of bigint.
They are not the issue of CONTEXT_INFO, so are not solved in this patch.


robverschoor (changes requested):
Can you please add test cases:

- SET CONTEXT_INFO NULL -- not allowed in SQL Server
- SET CONTEXT_INFO <128-byte hex constant>
- SET CONTEXT_INFO @v -- where @v is [var]binary(128) as well as longer than 128 and containing a longer value than 128 bytes

test/JDBC/expected/babel_context_info-vu-prepare.out
```diff
@@ -0,0 +1,14 @@
+CREATE PROCEDURE SET_CONTEXT_INFO
+    @context_info varbinary(128)
+AS
+BEGIN
+    EXEC sys.bbf_set_context_info @context_info
+END;
+GO
+
+CREATE PROCEDURE GET_CONTEXT_INFO
```
robverschoor: why do we need this procedure?

test/JDBC/expected/babel_context_info-vu-verify.out
```diff
@@ -0,0 +1,21 @@
...
+
+
+EXEC SET_CONTEXT_INFO 0x12345678
+GO
+
+SELECT CONTEXT_INFO()
+GO
+~~START~~
+varbinary
+1234567800000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000
```
robverschoor: The returned value should start with '0x': any [var]binary value should be formatted like that.
E.g. (in current Babelfish):
```
1> select cast(10 as varbinary)
2> go
varbinary
-------------------------------
0x0000000A
```

test/JDBC/expected/babel_context_info-vu-prepare.out
```diff
@@ -0,0 +1,14 @@
+CREATE PROCEDURE SET_CONTEXT_INFO
```
robverschoor: why do we need this procedure?

robverschoor (changes requested):
I added some inline comments.
Also, please add test cases for the following:
- SET CONTEXT_INFO 0
- SET CONTEXT_INFO @var, where @var is a variable of datatype [VAR]BINARY, INT  or any numeric datatype  such as NUMERIC(x,y) or FLOAT -- all of these are allowed by SQL Server; other datatypes like DATETIME or VARCHAR are not allowed by SQL Server and we should have negative test cases for those.
- when @v is a [VAR]BINARY longer than 128 there should be an error raised

test/JDBC/input/babel_context_info-vu-prepare.sql
```diff
@@ -0,0 +1,14 @@
...
+    EXEC sys.bbf_set_context_info @context_info
+END;
+GO
+
+CREATE PROCEDURE GET_CONTEXT_INFO
+AS
+BEGIN
+    SELECT CONTEXT_INFO()
+END;
+GO
```
macwoody, line 14: Please also help add a SQL view depending on them

test/JDBC/input/babel_context_info.sql
```diff
@@ -0,0 +1,75 @@
+-- 1. select without set
+SELECT CONTEXT_INFO()
```
macwoody: for simplicity, you could merge everything in this file into verify.sql

contrib/babelfishpg_tsql/sql/sys_views.sql
```diff
@@ -1029,7 +1029,7 @@ select
   , null::varchar(12) as net_address
   , null::varchar(12) as net_library
   , a.usename as loginname
-  , null::bytea as context_info
+  , t.context_info::bytea as context_info
```
macwoody, line 1032: upgrade script is missing for this object change. 
Without it, new definition will not take effect after minor version upgrade 
Please add new definition in 3.1.0 -> 3.2.0 upgrade script

test/JDBC/expected/babel_context_info-vu-verify.out
markrui3: Yeah, 0x0000000A is the format in sqlcmd, 0000000A is the format in JDBC or freetds.

test/JDBC/expected/babel_context_info-vu-prepare.out
markrui3: GET_CONTEXT_INFO and SET_CONTEXT_INFO is for upgrade test.

test/JDBC/upgrade/latest/schedule
```diff
@@ -410,4 +410,5 @@ BABEL-733
 BABEL-3938
 BABEL-NEXT-VALUE-FOR
 binary-index
+babel_context_info
```
surendravishnoi, line 415: latest schedule file is for MVU from self to self, no minor or major version change. To exercise 3.1 to 3.2 script, add to schedule file which exercises 3.1 to 3.2

robverschoor (dismissed):

macwoody (approved):

PR #1304, opened 2023-03-01
Support database-level dump/restore

rishabhtanwar29:
#### Engine
Added support for dumping a single Babelfish logical database. Changes are
summarized in following bullet points:
1. Added a new option `--bbf-database-name` in both pg_dumpall and pg_dump to
let user provide the name of logical database to be dumped.
2. Fetch all the physical schema names corresponding to the given logical database
and add all of them into `schema_include_patterns` list. With this, existing dump
logic will make sure to only dump all these schemas as well as all the objects
contained by these schemas.
3. Function `bbf_selectDumpableTableData` takes care of explicitly marking all
Babelfish catalog table data to be dumped.
4. We need to selectively dump the Babelfish catalog table data corresponding to
given logical database which is being handled by function `getCursorForBbfCatalogTableData`
for `COLUMN-INSERT` and `fixCopyCommand` for `COPY`. These function prepares
query using joins and conditions to filter the data.
5. We skip dumping the `dbid` column as it might conflict with the existing database's `dbid`
so we generate the new `dbid` during restore. It has been handled for both `COLUMN-INSERT`
and `COPY` options:
 a. `COPY`: Implemented a hook `fill_missing_values_in_copyfrom_hook` which will use the
 sequence to generate the new `dbid`.  
 b. `COLUMN-INSERT`: Used existing `pre_parse_analyze_hook` hook to handle this case.
5. Added a new file `dumpall_babel_utils.c` to implement Babelfish specific logic to
selectively dump users. New file is needed since existing `dump_babel_utils.c` can't
be used with pg_dumpall utility.
7. Logins do not belong to a single database so we will not be dumping the logins so it and
we will only dump default users of the database (dbo, db_owner and db_guest).
Due to this `CREATE` and `INSERT` might fail for certain logins but it should be fine.
8. Current implementation does not dump foreign servers and foreign data wrappers as
these are global objects and does not belong to a single logical database.
9. **Note**: A logical database with certain migration mode (single-db/multi-db) can only be 
restored on the Babelfish server with the same migration mode. Which means a single-db
database can be restored only on single-db Babelfish server, same thing applies to multi-db.
Although the builtin databases like `master` can be restored on any kind of server.

#### Extension
1. Modified existing instance dump-restore github action to work with database-level
dump-restore as well. We can use configuration file to specify the logical database name.
Currently we are testing with master database for which a directory has been in upgrade
folder.
2. Implemented `fill_missing_values_in_copyfrom_hook` to handle the case of generating
new dbid for `COPY`. Handled the same for `COLUMN-INSERT` in `pre_parse_analyze_hook`.

Task: BABEL-3870

* **Use case based -**
New github workflow has been added with existing upgrade test coverage.

* **Boundary conditions -**

* **Arbitrary inputs -**

* **Negative test cases -**

* **Performance tests -**

* **Tooling impact -**

* **Client tests -**

### Check List


contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -3403,3 +3412,64 @@ pltsql_set_target_table_alternative(ParseState *pstate, Node *stmt, CmdType comm
...
+							errmsg("cannot find an available ID for new database.")));
+		}
+		/*
+		 * For all the other catalog tables which contain dbid column, get dbid using current value of the
+		 * babelfish_db_seq sequence. It is ok to fetch current value of the sequence here since we already
+		 * have generated new dbid while inserting into sysdatabases catalog.
+		 */
+		else
+		{
+			RangeVar	*sequence = makeRangeVarFromNameList(stringToQualifiedNameList("sys.babelfish_db_seq"));
```
surendravishnoi: is this safe? there cannot be any race condition where sequence changed due to multiple get DB id calls? can there be multiple restore db calls in paraellel?

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -433,6 +434,98 @@ pltsql_pre_parse_analyze(ParseState *pstate, RawStmt *parseTree)
...
+				 * have generated new dbid while inserting into sysdatabases catalog.
+				 */
+				else
+				{
+					RangeVar	*sequence = makeRangeVarFromNameList(stringToQualifiedNameList("sys.babelfish_db_seq"));
+					Oid     	 seqid = RangeVarGetRelid(sequence, NoLock, false);
+
+					dbid = DirectFunctionCall1(currval_oid, seqid);
+				}
+				bbf_set_current_user(prev_current_user);
```
2jungkook: Looks like it is duplicate from here and `fill_missing_values_in_copyfrom()`. Can we make this as a separate function?

.github/configuration/dump-restore-test-configuration.yml
```diff
@@ -1,8 +1,20 @@
...
+  }
+],
+[
+  {
+    version: source.latest,
+    logical-database: master
+  },
+  {
+    version: target.latest,
+    logical-database: master
```
2jungkook, line 18: For dump-restore, do we have different github actions for single-db and multi-db?

contrib/babelfishpg_tsql/src/pl_handler.c
rishabhtanwar29: Done!

.github/configuration/dump-restore-test-configuration.yml
rishabhtanwar29: Currently, we are only running it for multi-db mode. Do you think we should be running tests for single-db as well? Asking as there is no specific change has been done dependent upon the migration_mode in the code.

contrib/babelfishpg_tsql/src/hooks.c
rishabhtanwar29: Yeah it is safe since the restore happens in a single session for which the `currval` of sequence will not change irrespective of other sessions.

test/JDBC/input/BABEL_OBJECT_ID-vu-cleanup.mix
```diff
@@ -33,12 +33,6 @@ GO
 DROP DATABASE babel_object_id_db;
 GO
 
-DROP TABLE babel_object_id_schema_longer_than_63_0abcdefgij1abcdefgij2abcdefgij3abcdefgij4abcdefgij5abcdefgij6abcdefgij7abcdefgij8abcdefghij9abcdefghij.babel_object_id_table_longer_than_63_0abcdefgij1abcdefgij2abcdefgij3abcdefgij4abcdefgij5abcdefgij6abcdefgij7abcdefgij8abcdefghij9abcdefghij;
```
rishabhtanwar29, line 36: This is not a valid test case even in sql-server so removing it for now.

test/JDBC/expected/BABEL_OBJECT_ID-vu-cleanup.out
```diff
@@ -33,12 +33,6 @@ GO
 DROP DATABASE babel_object_id_db;
 GO
 
-DROP TABLE babel_object_id_schema_longer_than_63_0abcdefgij1abcdefgij2abcdefgij3abcdefgij4abcdefgij5abcdefgij6abcdefgij7abcdefgij8abcdefghij9abcdefghij.babel_object_id_table_longer_than_63_0abcdefgij1abcdefgij2abcdefgij3abcdefgij4abcdefgij5abcdefgij6abcdefgij7abcdefgij8abcdefghij9abcdefghij;
```
rishabhtanwar29, line 36: This is not a valid test case even in sql-server so removing it for now.

.github/configuration/dump-restore-test-configuration.yml
2jungkook: It's okay for now, but when something is changed in the future, it might be hard to think about dump/restore cases.

Btw, I'm just wondering the following cases:
1. single-db --> multi-db
2. multi-db --> single-db

Do we allow those cases in a database-level dump/restore?

2jungkook (dismissed):

.github/configuration/dump-restore-test-configuration.yml
rishabhtanwar29: Nope, as discussed earlier we only support the following:
1. single-db --> single-db
2. multi-db --> multi-db  

Added this in the description as well.

surendravishnoi (approved):

2jungkook (approved):

PR #1333, opened 2023-03-14
Create proc/func/trigger not visible in the pg_stat_statements view

kritika-0601:
### Description

Currently Babelfish does not show `create proc/func/trigger` query inside the pg_stat_statements table. After my change now it will be visible in the view.
pg_stat_statements displays planning and execution statistics for all SQL statements executed by a server. Some statements like create proc/func was not being tracked. Mainly because subsequent processutility hooks after bbf_processutilty were not being called. 
Created a hook for that and called it inside relevant ProcessUtility function.

PR for engine code: https://github.com/babelfish-for-postgresql/postgresql_modified_for_babelfish/pull/113

### Issues Resolved

* **Use case based -**

* **Boundary conditions -**

* **Arbitrary inputs -**

* **Negative test cases -**

* **Performance tests -**

* **Tooling impact -**

* **Client tests -**

### Check List


contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -65,12 +65,18 @@
 #include "multidb.h"
 #include "tsql_analyze.h"
 
+#include "tcop/cmdtag.h"
+#include "tcop/tcopprot.h"
```
rishabhtanwar29: Let's move these above with the other includes in alphabetical order. Note, first comes all the engine headers and then the extensions' headers.

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -293,10 +310,18 @@ InstallExtendedHooks(void)
...
 
+
+	prev_CreateFunctionStmt_hook = CreateFunctionStmt_hook;
+	CreateFunctionStmt_hook = pltsql_CreateFunctionStmt;
+
 	prev_fill_missing_values_in_copyfrom_hook = fill_missing_values_in_copyfrom_hook;
 	fill_missing_values_in_copyfrom_hook = fill_missing_values_in_copyfrom;
+
+
+
```
rishabhtanwar29: Nit: unnecessary line breaks.

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -293,10 +310,18 @@ InstallExtendedHooks(void)
...
 	prev_check_rowcount_hook = check_rowcount_hook;
 	check_rowcount_hook = bbf_check_rowcount_hook;
+
```
rishabhtanwar29, line 329: Nit: not needed.

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -347,6 +373,169 @@ UninstallExtendedHooks(void)
  * 			Hook Functions
  *****************************************/
 
+void pltsql_CreateFunctionStmt(ParseState *pstate,
```
rishabhtanwar29: Nit: missing required line break after `void`. Also, the function needs to be `static`.

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -347,6 +373,169 @@ UninstallExtendedHooks(void)
...
+							   PlannedStmt *pstmt,
+							   const char *queryString,
+							   bool readOnlyTree,
+							   ProcessUtilityContext context,
+								ParamListInfo params,
```
rishabhtanwar29: Nit: Indentation look a bit off.

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -347,6 +373,169 @@ UninstallExtendedHooks(void)
  * 			Hook Functions
  *****************************************/
 
+void pltsql_CreateFunctionStmt(ParseState *pstate,
+							   PlannedStmt *pstmt,
+							   const char *queryString,
+							   bool readOnlyTree,
```
rishabhtanwar29: Looks redundant as we are always passing `readOnlyTree = false`.

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -347,6 +373,169 @@ UninstallExtendedHooks(void)
...
+							   ProcessUtilityContext context,
+								ParamListInfo params,
+							   QueryEnvironment *queryEnv,
+							   DestReceiver *dest,
+							   QueryCompletion *qc)
```
rishabhtanwar29: All three look redundant as we are not using them anywhere.

rishabhtanwar29:
One other we could do this without introducing a new hook is that tweak the logic in bbf_ProcessUtility itself:
```
// create table type if needed
if (tbltypStmt)
   *create table type*

// call standard utility to create the function.
if (prev_ProcessUtility)
    prev_ProcessUtility(pstmt, queryString, readOnlyTree, context,
						params, queryEnv, dest,
						qc);
else
    standard_ProcessUtility(pstmt, queryString, readOnlyTree, context,
						params, queryEnv, dest,
						qc);

//get address of the function since it has already been created.

// create trigger if needed
if (trigstmt)
   *create the trigger*
```
@2jungkook do you also agree with this?

HarshLunagariya:
> One other we could do this without introducing a new hook is that tweak the logic in bbf_ProcessUtility itself:
> 
> ```
> // create table type if needed
> if (tbltypStmt)
>    *create table type*
> 
> // call standard utility to create the function.
> if (prev_ProcessUtility)
>     prev_ProcessUtility(pstmt, queryString, readOnlyTree, context,
> 						params, queryEnv, dest,
> 						qc);
> else
>     standard_ProcessUtility(pstmt, queryString, readOnlyTree, context,
> 						params, queryEnv, dest,
> 						qc);
> 
> //get address of the function since it has already been created.
> 
> // create trigger if needed
> if (trigstmt)
>    *create the trigger*
> ```
> 
> @2jungkook do you also agree with this?

There are two problems with above solution:
1. If we do it by calling standard_ProcessUtility() for creating the function only as suggested above then statistics will be only collected for actual create physical function logic. If won't include the statistics for `create table type` or `create trigger` statements. This the main pain point why we can't avoid the hooks here.

2. Even if we go ahead with above approach, then getting the address of function(because we using it in the subsequent logic), we need to rebuild the function signature from the CreateFunctionStmt and search in the catalog.

IMO, To address 1st point for correctness of statistics we should go ahead with hooks approach only.

PR #1355, opened 2023-03-21
Fix the issue that context_info() cannot display complete varbinary in sqlcmd.

ghost:
Fix the issue that context_info() cannot display complete varbinary in sqlcmd.

Description
===========
context_info() cannot return complete varbinary in sqlcmd.

Analysis
========
Based on the TDS document, it is expected that sqlcmd would have the capability
to support both variable-length and partially length-prefixed data types.
However, in this particular scenario, sqlcmd does not support the partially
length-prefixed data type for varbinary.
The main distinguishing factor between these two data types is that the
variable-length data type has an associated length value, while the
partially length-prefixed data type does not require the full data length to be
specified before the actual data is streamed out.

Note: Neither freetds nor JDBC encounter any issues with regard to this matter.

Fix
===
Make context_info() immutable, which can get typmod and use variable-length
data type.


contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--3.1.0--3.2.0.sql
```diff
@@ -421,8 +421,18 @@ CALL sys.babelfish_drop_deprecated_object('view', 'sys', 'sysprocesses_deprecate
 CALL sys.babelfish_drop_deprecated_object('view', 'sys', 'dm_exec_sessions_deprecated_in_3_2_0');
 CALL sys.babelfish_drop_deprecated_object('function', 'sys', 'tsql_stat_get_activity_deprecated_in_3_2_0');
 
+CREATE OR REPLACE FUNCTION sys.bbf_get_context_info()
+RETURNS sys.VARBINARY(128) AS 'babelfishpg_tsql', 'bbf_get_context_info' LANGUAGE C STABLE;
+
 CREATE OR REPLACE FUNCTION sys.context_info()
-RETURNS sys.VARBINARY(128) AS 'babelfishpg_tsql' LANGUAGE C STABLE;
+RETURNS sys.VARBINARY(128)
```
macwoody, line 428: I wonder do we need this 2level structure?

kuntalghosh (approved):

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--3.1.0--3.2.0.sql
kuntalghosh: I don't think you can specify both AS 'definition' and AS 'obj_file', 'link_symbol' in the same function definition.

macwoody (approved):

PR #1340, opened 2023-03-16
DATEADD() gives wrong output with DATETIMEOFFSET

bsatarupa:
### Problem description

DATEADD() gives wrong output with DATETIMEOFFSET

Postgres outputs datetimeoffset data by removing/subtracting input timezone value from it.
Fix includes, working on the engine output to include timezone and thereby retain the input timestamp as is.

1> declare @dt datetimeoffset(6)
2> set @dt = '1912-10-25 12:24:32 +10:0'
3> select dateadd(month,1,@dt)
4> go
dateadd

 BABEL output: 1912-11-25 02:24:32.0000000 +10:00

   SQL output: 1912-11-25 12:24:32.0000000 +10:00

### Upgrade

This commit includes adding changes to dateadd_internal_df() to reflect sql behavior.

### Test scenarios covered

o Use case based: Original issue
o Null testcases: N/A
o Negative test scenarios: added negative timezone (Permissable range +14:00 to -14:00)
o Boundary conditions, Arbitrary inputs: Testcases including leap year dates, dates across years, dates having positive and negative timezones.
o Testcases calling dateadd function for datetimeoffset from procedures, views and functions
o Expected result correction on existing testcases
o Minor version upgrade tests, Major version upgrade tests : Done
o Performance tests, Tooling impact, Client tests: N/A

Task: BABEL-1625
Signed-off-by: Satarupa Biswas [satarupb@amazon.com](mailto:satarupb@amazon.com)

PR #1364, opened 2023-03-24
Fixing errors in existing expected file created due to changes added as part of BABEL-1625

bsatarupa:
Postgres outputs datetimeoffset data by removing/subtracting input timezone value from it.
Datetimeoffset change in PR https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/1340/ works on the engine output to include timezone and thereby retain the input timestamp as is.

In this PR, we have fixed errors in existing expected file generated due to changes added to BABEL-1625.

Task: BABEL-1625
Signed-off-by: Satarupa Biswas satarupb@amazon.com


surendravishnoi (commented):
Please do not put links to internal docs here. You can describe what fix is doing. Changing expected files to account for datetimeoffset behavior change.

surendravishnoi (approved):

rishabhtanwar29 (approved):

PR #1382, opened 2023-03-31
In CTE Null first order behavior is not correct

forestkeeper:
Previously we add the Null first check after analyze, and in that case it'll miss to add the Null first tag to the sort clause for a CTE query

change the place adding Null first tag to the place sortClause is initialized

Task: BABEL-3991

### Description

the following sql not produce the right order for NULL : 
```
drop table if exists t1;
create table t1(a int);
insert t1 values (1);
insert t1 values (3);
insert t1 values (null);
with t1cte AS (
select top(3) a from t1 order by 1
)
select * from t1cte; -- Returns NULL at bottom not expected
select top(3) a from t1 order by 1; -- Returns NULL at the TOP as expected
```

the root cause is we deal with sort clause inproperly

### Issues Resolved

Will return the right order for null include 
- asc order by return NULL first
- no order by , natural asc, return NULL last
- desc order by return NULL last

checked this behavior with 
- simple select
- select order by
- subquery order by 
- cte with inner query order by
- babel_functions
- etc other remainning explain query

### Check List
- [ x ] Commits are signed per the DCO using --signoff 


test/JDBC/expected/babel_function.out
```diff
@@ -2426,26 +2426,26 @@ SELECT a, b, COUNT(*) OVER (ORDER BY a) from t3;
 GO
```
forestkeeper, line 2426: order by a , nullfirst +  asc , null comes first

test/JDBC/expected/babel_function.out
```diff
@@ -2426,26 +2426,26 @@ SELECT a, b, COUNT(*) OVER (ORDER BY a) from t3;
...
+efg#!#b#!#5
+xyz#!#a#!#7
+xyz#!#b#!#7
 ~~END~~
 
 SELECT a, b, COUNT(*) OVER (ORDER BY a DESC) from t3;
 GO
 ~~START~~
 varchar#!#varchar#!#int
-<NULL>#!#<NULL>#!#1
```
forestkeeper, line 2442: order by desc , null comes last

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -3630,3 +3636,13 @@ bbf_check_rowcount_hook(int es_processed)
 	else
 		return false;
 }
+
+static void 
+sort_nulls_first(SortGroupClause * sortcl, bool reverse)
+{
+	if (sql_dialect == SQL_DIALECT_TSQL)
```
2jungkook, line 3643: To test this dialect check, could you also add a test case in PG connection?

forestkeeper: OK

2jungkook (approved):

PR #1383, opened 2023-03-31
BABEL-1199 Support DATABASE_PRINCIPAL_ID() T-SQL function 

pratikzode:
SELECT DATABASE_PRINCIPAL_ID();
Retrieve the ID of the current user.

SELECT DATABASE_PRINCIPAL_ID(‘principal_name’);
Retrieve the ID of the specific database principal.

Modified the file so that when called with NULL as argument it will return NULL and also modified for trailing spaces that will give proper output instead of NULL.

Task: BABEL-1199
Signed-off-by: Pratik Zode [pzode@amazon.com](mailto:pzode@amazon.com)


forestkeeper (approved):
LGTM

xhfanhe (approved):

PR #1412, opened 2023-04-10
Do not use session time zone as a default in datetimeoffset

### LINKED ISSUE #1411: [Bug]: Invalid default time zone in datetimeoffset
### What happened?

If I run the following on a machine running in `UTC+01:00` time zone:

```
create table test1(a datetimeoffset);
insert into test1(a) values('2000-06-01 00:00:00');
select cast(a as varchar(max)) from test1;
```

On MSSQL it gives the following result:

```
----------------------------------
2000-06-01 00:00:00.0000000 +00:00
```

On Babelfish it gives the following:

```
--------------------------
2000-06-01 00:00:00 +01:00
```

According to MSFT, in `datetimeoffset` [TIMEZONE defaults to +00:00](https://learn.microsoft.com/en-us/sql/t-sql/data-types/datetimeoffset-transact-sql?view=sql-server-ver16#converting-string-literals-to-datetimeoffset), but in Babelfish it currently defaults to session timezone (`+01:00` in my case).

It appeared to be easy to fix this, will create a PR shortly.

PS: stumbled upon this while investigating test failures that started to happen after the recent daylight saving switch. There are seems to be more timezone-related problems around `datetimeoffset`, investigation is in progress, will file more issues/PRs if I find something.

### Version

BABEL_3_X_DEV (Default)

### Extension

babelfishpg_common

Fedora

### Relevant log output

_No response_

### Code of Conduct

staticlibs:
### Description

This change checks whether time zone part was present in the input string by examining the output of a [ParseDateTime call](https://github.com/babelfish-for-postgresql/babelfish_extensions/blob/9a6b2cde9ede7ee47c6abdb02499db74d24e1d09/contrib/babelfishpg_common/src/datetimeoffset.c#L110). If time zone was not found, then it set to zero instead of the one from [DecodeDateTime call](https://github.com/babelfish-for-postgresql/babelfish_extensions/blob/9a6b2cde9ede7ee47c6abdb02499db74d24e1d09/contrib/babelfishpg_common/src/datetimeoffset.c#L114) (session time zone in this case).

### Issues Resolved

#1411

There are a number of existing test cases for `datetimeoffset` data type (`TestDatetimeoffset`, `babel_datetimeoffset` tests among others), but this specific problem is not easy to test because it depends on a time zone set on a server where DB runs.

### Check List


2jungkook:
Could you test this by changing a session time zone, e.g., `set timezone TO 'GMT';`? If this is not possible, we can consider to add a new Github action using a different server time zone.

staticlibs:
Something like this seems to allow to set the time zone for the specific test:

```
-- psql
ALTER DATABASE jdbc_testdb SET timezone TO 'Australia/Sydney'
GO

SELECT pg_reload_conf()
GO

-- tsql
SELECT CAST(CAST('2000-06-01 00:00:00' AS DATETIMEOFFSET) AS VARCHAR(MAX))
GO

-- psql
ALTER DATABASE jdbc_testdb RESET timezone
GO

SELECT pg_reload_conf()
GO
```

Without the patch TSQL output uses `Australia/Sydney` timezone:

```
2000-06-01 00:00:00 +10:00
```

This approach should also work for #1418 .

I intend to add such test to this PR, just would appreciate some guidance here on tests organization - should it be a new test (then a single `.mix` or `prepare/verify/cleanup`) or is it better to update existing `datetimeoffset` test, should all `datetimeoffset` tests run in a non-UTC timezone or only one check etc.


Note, this approach won't work for #1420, where JVM timezone needs to be specified to [Maven invocation](https://github.com/babelfish-for-postgresql/babelfish_extensions/blob/2f24a3765f65aa0683540a345cf753898e5c3119/.github/workflows/jdbc-tests.yml#L47) with something like this:

```
mvn test -DargLine="-Duser.timezone=..."
``` 

Would it be appropriate to set it to some custom timezone for the whole run? Or instead set it to UTC in a main run and add an some kind of an additional run with a custom timezone?

KushaalShroff:
> I intend to add such test to this PR, just would appreciate some guidance here on tests organization - should it be a new test (then a single .mix or prepare/verify/cleanup) or is it better to update existing datetimeoffset test, should all datetimeoffset tests run in a non-UTC timezone or only one check etc.

We can update the existing tests and add such tests there.

I also want to discuss the upgrade impact this change can have, what if an existing database(in Australia/Sydney timezone) has a table with a datetimeoffset column with default value `'2000-06-01 00:00:00'`, it would already be seeing `2000-06-01 00:00:00 +10:00` when NULL is inserted for this column. Now with your change, the database will be upgraded and for the new Values(NULL) inserted will be, it will see `2000-06-01 00:00:00 +00:00`. We shall add some upgrade tests like this as well.

@2jungkook This change in behaviour needs to be documented properly

staticlibs:
Added a test case:

 - new `.mix` test because existing `datetimeoffset` tests are not cross dialect
 - T-SQL connection is reopened after the DB-wide timezone change to refresh its session time zone
 - only printing time zone name in the custom case to not depend on a default time zone name in `.out` file
 - did not include the header comment because it is getting re-ordered with `-- tsql` comments in `.out` file
 - it is intended to extend the same new `babel_datetimeoffset_timezone.mix` file with the additional test case in #1418

I did not add any upgrade tests, would appreciate pointers on them if such test is necessary as part of this PR.

2jungkook:
@staticlibs here is a guide to write upgrade tests for this PR (once those upgrade tests are added, you can remove the current test file):

```
datetimeoffset-timezone-vu-prepare
-- create a table with a datetimeoffset column with default value '2000-06-01 00:00:00'
-- insert some default values
-- create a view from SELECT CAST('2000-06-01 00:00:00' AS DATETIMEOFFSET)
```

```
datetimeoffset-timezone-vu-verify
-- insert some default values to the table
-- check the table values
-- check the view result
```

```
datetimeoffset-timezone-vu-cleanup
-- remove the table and the view
```

Add `datetimeoffset-timezone` to `latest/schedule` file.

```
datetimeoffset-timezone-before-15_3-vu-prepare
-- same as in datetimeoffset-timezone-vu-prepare
```

```
datetimeoffset-timezone-before-15_3-vu-verify
-- same as in datetimeoffset-timezone-vu-verify, but the expected output could be different
```

```
datetimeoffset-timezone-before-15_3-vu-cleanup
-- same as in datetimeoffset-timezone-vu-cleanup
```

Add `datetimeoffset-timezone-before-15_3` to schedule files under `14_7`, `14_8` and `15_2`.
(For example, if it is added in `14_7` schedule file, the prepare script is executed in 14_7, and then upgrade the test instance to the latest and execute verify and cleanup scripts)

Lastly, you need to ignore `datetimeoffset-timezone-before-15_3-vu-prepare/verify/cleanup` in the `jdbc_schedule` file.

Reference: https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/1398/files

staticlibs:
@2jungkook thanks for detailed instructions! I've followed them closely, just the problem was not reproducible with a `DEFAULT` column, so I've used ordinary `datetimeoffset` column. Also have rebased the PR to the latest `BABEL_3_X_DEV` because of the merge problems in `schedule` files.

I can see upgrade failures in GitHub actions run, will look into details when all workflows are completed.

staticlibs:
Added a minor fix to `vu-verify` tests.

2jungkook:
Two more comments:
1. Let's add DEFAULT column to the test table so that others can understand and see this behavior.
2. In both verify scripts, please add INSERT statements so that we can ensure the correct value is added (Note that verify scripts are always executed in the latest version)

staticlibs:
Thanks for the comments!

>     1. Let's add DEFAULT column to the test table so that others can understand and see this behavior.

Added `DEFAULT` column `dto_default ` in addition to the existing ordinary `dto` column. The problem without the fix is only reproducible with inserts into ordinary non-default column. The problem does not occur when with the values inserted with `DEFAULT` specifier in `values` tuple.
 
>     2. In both verify scripts, please add INSERT statements so that we can ensure the correct value is added (Note that verify scripts are always executed in the latest version)

Added `INSERT` statements to both `vu-verify` scripts for default and custom time zones.

2jungkook (approved):

2jungkook:
> > I intend to add such test to this PR, just would appreciate some guidance here on tests organization - should it be a new test (then a single .mix or prepare/verify/cleanup) or is it better to update existing datetimeoffset test, should all datetimeoffset tests run in a non-UTC timezone or only one check etc.
> 
> We can update the existing tests and add such tests there.
> 
> I also want to discuss the upgrade impact this change can have, what if an existing database(in Australia/Sydney timezone) has a table with a datetimeoffset column with default value `'2000-06-01 00:00:00'`, it would already be seeing `2000-06-01 00:00:00 +10:00` when NULL is inserted for this column. Now with your change, the database will be upgraded and for the new Values(NULL) inserted will be, it will see `2000-06-01 00:00:00 +00:00`. We shall add some upgrade tests like this as well.
> 
> @2jungkook This change in behaviour needs to be documented properly

Seems like default values are not affected.

2jungkook:
Thanks @staticlibs! The change looks good to me.

staticlibs:
>> what if an existing database(in Australia/Sydney timezone) has a table with a datetimeoffset column with default value '2000-06-01 00:00:00', it would already be seeing 2000-06-01 00:00:00 +10:00 when NULL is inserted for this column

> Seems like default values are not affected.

Another clarification on `DEFAULT` values, apologies for confusion here, after re-reading the request about the `DEFAULT` values I can now understand the logic better:

The problem happens when the input string is being decoded into the `datetimeoffset` value. With the `DEFAULT` column this decoding happens when the table is created and then the saved `datetimeoffset` value is re-used when `values()` tuple with a `default` specifier is inserted. It was not reproducible in `datetimeoffset-timezone` test for me because the table was created in `UTC` (system default) time zone and the database-wide time zone was changed to `Australia/Sydney` only after that. 

If I switch the DB time zone to `Australia/Sydney` first and then create a table with a `datetimeoffset default '2000-06-01 00:00:00'` column, then this column will continue producing `2000-06-01 00:00:00 +10:00` for every subsequent default insert after the upgrade. I guess this way the test will match closer to the possible real-life deployment with "existing database(in Australia/Sydney timezone)".

I was intended to update the `datetimeoffset-timezone` test anyway for #1418 , so, if it makes sense, I can change the table creation to `Australia/Sydney` to cover `DEFAULT` columns better.

2jungkook:
> > > what if an existing database(in Australia/Sydney timezone) has a table with a datetimeoffset column with default value '2000-06-01 00:00:00', it would already be seeing 2000-06-01 00:00:00 +10:00 when NULL is inserted for this column
> 
> > Seems like default values are not affected.
> 
> Another clarification on `DEFAULT` values, apologies for confusion here, after re-reading the request about the `DEFAULT` values I can now understand the logic better:
> 
> The problem happens when the input string is being decoded into the `datetimeoffset` value. With the `DEFAULT` column this decoding happens when the table is created and then the saved `datetimeoffset` value is re-used when `values()` tuple with a `default` specifier is inserted. It was not reproducible in `datetimeoffset-timezone` test for me because the table was created in `UTC` (system default) time zone and the database-wide time zone was changed to `Australia/Sydney` only after that.
> 
> If I switch the DB time zone to `Australia/Sydney` first and then create a table with a `datetimeoffset default '2000-06-01 00:00:00'` column, then this column will continue producing `2000-06-01 00:00:00 +10:00` for every subsequent default insert after the upgrade. I guess this way the test will match closer to the possible real-life deployment with "existing database(in Australia/Sydney timezone)".
> 
> I was intended to update the `datetimeoffset-timezone` test anyway for #1418 , so, if it makes sense, I can change the table creation to `Australia/Sydney` to cover `DEFAULT` columns better.

Thanks for the detailed explanation! Yes, please update the test in #1418.

PR #1416, opened 2023-04-11
Add COLUMN for sp_rename

raydhkim:
Previously, sp_rename did not support renaming column in BBF. Now, renaming of column though sp_rename is also being supported.

Task: BABEL-3889

### Check List


lejaokri (commented):
A couple of nitpick comments. 

You do not need to make these changes but consider them in the future.

(1) babelfishpg_tsql.sql: I think we can replace some of the IF to IF-ELSE-IF

(2) babelfishpg_tsql.sql: Quite a few duplicated code here and that always suggests something can be improved a bit. eg: a helper function. Also, I notice that subname is always the last and relname is second to the last. You can modify the ORDER BY such that first row is subname and second row is relname, that way you can reuse the same query regardless of name_count.

(3) Adding comments help future readers as well.

(4) gen_sp_rename_subcmds(): In general, if we have a bunch of if/else if cases consider using a switch statement instead whenever possible. The compiler could optimize switch statements so it's more efficient plus switch throws a warning/error if not all cases are covered so it's less error prone.

(5) In sp_rename_internal, we can also introduce a objtype -> struct (objtype_code, errmsg) map instead of a if/else-if

contrib/babelfishpg_tsql/sql/babelfishpg_tsql.sql
```diff
@@ -2973,71 +2970,149 @@ BEGIN
...
+			IF @name_count = 2
+				BEGIN
+					WITH parseTable AS (
+						SELECT (ROW_NUMBER() OVER (ORDER BY NULL)) as row,*
+						FROM STRING_SPLIT(@objname, '.'))
+					SELECT @schemaname = value FROM parseTable WHERE row = 1;
+					WITH parseTable AS (
+						SELECT (ROW_NUMBER() OVER (ORDER BY NULL)) as row,*
+						FROM STRING_SPLIT(@objname, '.'))
+					SELECT @subname = value FROM parseTable WHERE row = 2;
```
xhfanhe: Let's create a helper proc or func to do the string split recursively. This sp_rename procedure block is too long and has too many duplications.

test/JDBC/input/storedProcedures/Test-sp_rename-vu-verify.sql
```diff
@@ -209,7 +209,37 @@ GO
...
+GO
+
+EXEC sp_rename 'sp_rename_vu_schema1.sp_rename_vu_table2_new.sp_rename_vu_s1_t2_col1', 'sp_rename_vu_s1_t2_col1_new', 'COLUMN';
+GO
+
+SELECT COLUMN_NAME, TABLE_SCHEMA, TABLE_NAME 
+FROM INFORMATION_SCHEMA.COLUMNS WHERE TABLE_NAME = 'sp_rename_vu_table2_new'
+ORDER BY COLUMN_NAME, TABLE_SCHEMA, TABLE_NAME;
+GO
+
```
xhfanhe, line 246: Please try some basic operations after renaming, like insert and select from the column

xhfanhe (approved):

lejaokri (approved):

PR #1421, opened 2023-04-11
Alter table crash in antlr parser unsupported feature

Jakeowen1:
### Description

1. Currently Babelfish crashes when alter table alter column is called with drop default where the column has no definition
2. This change was made to prevent customers instances from crashing when calling this query

### Issues Resolved

[Babel-3818]

* **Use case based -**

```
1> alter table cities alter column region drop default defRegion
2> go
'DROP DEFAULT' is not currently supported in Babelfish
```

```
1> alter table cities alter column region varchar(50) drop default defRegion
2> go
'DROP DEFAULT' is not currently supported in Babelfish
```

```
1> alter table cities alter region varchar(50) drop default defRegion
2> go
syntax error near 'region' at line 1 and character position 25
```

```
1> alter table cities alter region drop default defRegion
2> go
syntax error near 'region' at line 1 and character position 25
```

```
1> alter table cities alter region varchar(50) drop default
2> go
syntax error near 'region' at line 1 and character position 25
```

* **Arbitrary inputs -**

- Column not defined in table
```
1> alter table cities alter column fake_column drop default defRegion
2> go
'DROP DEFAULT' is not currently supported in Babelfish
```

Added to upgrade scripts

Added to upgrade scripts

* **Performance tests -**
NA

* **Tooling impact -**
NA

* **Client tests -**
Example tests used are reflective of client calls that caused error

### Check List


contrib/babelfishpg_tsql/src/tsqlUnsupportedFeatureHandler.cpp
```diff
@@ -737,16 +737,18 @@ antlrcpp::Any TsqlUnsupportedFeatureHandlerImpl::visitAlter_table(TSqlParser::Al
 
 	if (ctx->COLUMN()) // ALTER TABLE ... ALTER COLUMN
 	{
-		Assert(ctx->column_definition());
```
forestkeeper, line 740: Can we still keep the assert in here ? I think this change is not related to null column_def

Jakeowen1: The assert is what was causing the crash. When we call `alter table cities alter column region drop default defRegion` the antlr parser is expecting a column definition after "region" which is why there was no crash originally when running `alter table cities alter column region varchar(50) drop default defRegion`. Perhaps there is another way to handle the assert that could be better?

forestkeeper: So for alter table cities alter column region drop default defRegion , the ctx->column_definition() = null, why I didn't find the if cdctx == null clause in if conditon

forestkeeper (approved):
LGTM

PR #1427, opened 2023-04-13
Numeric receiver function TdsRecvTypeNumeric won't work for system that does not support 128bit INT

bsatarupa:
### Description

Numeric receiver function TdsRecvTypeNumeric won't work for system that does not support 128bit INT
[The problem statement is incorrect though in my opinion. GCC4.1 and above(today we mostly use GCC > 7) have defined type uint128_t to handle 128-bit integer data; even if we have an underlying 64-bit or 32-bit processor, uint128_t uses 2 64-bit (or, 4 32-bit) registers to store data of length 128 bit.
This can be a good read: https://cs.stackexchange.com/questions/140881/how-do-computers-perform-operations-on-numbers-that-are-larger-than-64-bits

Postgres code has "if defined(HAVE_INT128)"; since Numeric PG datatype can have up to 131072 digits before the decimal point; up to 16383 digits after the decimal point. For, SQL sever Numeric datatype can have upto 38 decimal digits that can always be accommodated using uint128_t datatype without any further alignment of data like PG does. PG uses a different approach for storing Numeric data than SQL server does.]

Nonetheless, code change includes, checking length of Numeric data from precision and using 128-bit integer(uint128) only if Numeric data can't be accommodated in a 64-bit integer.

Task: BABEL-3577

* **Use case based -**

* **Boundary conditions -**

* **Arbitrary inputs -**

* **Negative test cases -**

* **Performance tests -**

* **Tooling impact -**

* **Client tests -**

### Check List

### DIFF
--- a/contrib/babelfishpg_tds/src/backend/tds/tdstypeio.c
+++ b/contrib/babelfishpg_tds/src/backend/tds/tdstypeio.c
@@ -1047,7 +1047,24 @@ ReverseString(char *res)
 }
 
 static inline void
-Integer2String(uint128 num, char *str)
+Integer128_2_String(uint128 num, char *str)
+{
+	int			i = 0,
+				rem = 0;
+
+	while (num)
+	{
+		rem = num % 10;
+		str[i++] = rem + '0';
+		num = num / 10;
+	}
+	str[i++] = '-';
+	ReverseString(str);
+}
+
+
+static inline void
+Integer2String(uint64 num, char *str)
 {
 	int			i = 0,
 				rem = 0;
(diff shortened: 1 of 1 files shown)


bsatarupa:
Do we really need this change?

GCC4.1 and above(today we mostly use GCC > 7) have defined type uint128_t to handle 128-bit integer data; even if we have an underlying 64-bit or 32-bit processor, uint128_t uses 2 64-bit (or, 4 32-bit) registers to store data of length 128 bit.
This can be a good read: https://cs.stackexchange.com/questions/140881/how-do-computers-perform-operations-on-numbers-that-are-larger-than-64-bits

Postgres code has "if defined(HAVE_INT128)"; since Numeric PG datatype can have up to 131072 digits before the decimal point; up to 16383 digits after the decimal point. For, SQL sever Numeric datatype can have upto 38 decimal digits that can always be accommodated using uint128_t datatype without any further alignment of data like PG does. PG uses a different approach for storing Numeric data than SQL server does.

PR #1434, opened 2023-04-14
Fix Memory Corruption in update_DropRoleStmt

HarshLunagariya:
### Description
Previously, sys.drop_all_logins() function call was throwing error and failing to drop it when it encounters login or user with considerably long name due to memory corruption in update_DropRoleStmt(). It caused by unwanted pfree() on the first element of stmt->roles which was still being used in subsequent logic.

To fix above issue, We removed explicit memory deallocation of first element of stmt->roles list. It is not needed because list_delete_cell already handles the memory deallocation.

Task: BABEL-4041

### Issues Resolved
BABEL-4041

### Testing
Before
```
--tsql
1> create login lllllllllllllllllllllll with password='12345678';
2> go
1> exit

--psql
jdbc_testdb=# CALL sys.babel_drop_all_logins();
ERROR:  Cannot drop the login 'llllllllЯk', because it does not exist or you do not have permission.
```
After
```
--tsql
1> create login lllllllllllllllllllllll with password='12345678';
2> go
1> exit

--psql
jdbc_testdb=# CALL sys.babel_drop_all_logins();
CALL
```

* **Use case based -**
Included in Description

* **Boundary conditions -**
NA

* **Arbitrary inputs -**
NA

* **Negative test cases -**
NA

NA

NA

* **Performance tests -**
NA

* **Tooling impact -**
NA

* **Client tests -**
NA

### Check List


KushaalShroff (approved):

KushaalShroff:
Can we add some other test case which triggers and tests this change?

HarshLunagariya:
> 

Only sys.remove_babelfish() call sys.drop_all_logins() which internally uses update_DropRoleStmt function. So it is not possible to add any other scenario.

contrib/babelfishpg_tsql/src/pltsql_utils.c
```diff
@@ -693,8 +693,6 @@ update_DropRoleStmt(Node *n, const char *role)
 		if (strcmp(tmp->rolename, "is_role") == 0)
 			stmt->roles = list_delete_cell(stmt->roles, list_head(stmt->roles));
 
-		pfree(tmp);
```
surendravishnoi, line 696: we should add test case for this. We should have a way to test functions like sys.remove_babelfish() with different input parameters. May be to have such tests run strictly in the end after regular tests are done?

HarshLunagariya: Tracking effort of adding automated test as separate task.

KushaalShroff, line 696: we should think of adding this pfree back at the end of the function

PR #1440, opened 2023-04-17
Table Variables inside Transactions (extension code changes)

lejaokri:
This commit refactors how Table Variables work inside a Transaction especially when there is a rollback. The Table Variable metadata, on-disk files, and tuples should be able to survive rollback because Table Variables are non-transactional.

Also, Table Variables primary key, constraint and toast tables should be non-transactional as well.

On high-level there are three main components in this commit:

  1. Metadata: Table Variable metadata are added into ENR in order to bypass catalog MVCC.

  2. On-disk files: Table Variable relfilenode are not added to pending_deletes during table creation so it would not be deleted in case of rollback.

  3. Tuples: Introduced a new set of MVCC, Vacuum, Toast visibility functions.

(Remember to revert .github/scripts/clone_engine_repo.conf)

Task: [BABEL-579]

### Check List


.github/scripts/clone_engine_repo.conf
```diff
@@ -1 +1 @@
-DEFAULT_BRANCH=BABEL_3_X_DEV__PG_15_X
```
lejaokri, line 1: temporary only for testing purposes

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -1958,13 +1959,15 @@ PLTsqlProcessTransaction(Node *parsetree,
 				 * haven't implemented this yet so we throw an error if
 				 * ROLLBACK is used with table variables.
 				 */
+/*
```
surendravishnoi: cleanup

contrib/babelfishpg_tsql/src/table_variable_mvcc.c
```diff
@@ -0,0 +1,1094 @@
...
+
+                /* not LOCKED_ONLY, so it has to have an xmax */
+                Assert(TransactionIdIsValid(xmax));
+
+                /*
+                 * The updating subtransaction must have aborted.
+                 * The xmax does not appear as it still belong to the current transaction
+                 * because it has been removed from the list of subtransactions.
+                 */
+                if (!TransactionIdIsCurrentTransactionId(xmax) && find_failed_transaction(xmax))
```
surendravishnoi: this seems confusing. Why are we assuming that current transaction might not have failed? If xmax is current transaction, it could have still failed

contrib/babelfishpg_tsql/src/table_variable_mvcc.c
```diff
@@ -0,0 +1,1094 @@
...
+            return (HeapTupleHeaderGetCmax(tuple) >= snapshot->curcid);     /* Case 1.2.4 and 2.2.2 */
+        }
+        else if (find_failed_transaction(HeapTupleHeaderGetRawXmin(tuple))) /* This is case 2.2.3 in comment section */
+        {
+            /* xmin is aborted unexpectedly */
+            HeapTupleSetHintBits(tuple, buffer, HEAP_XMIN_INVALID, InvalidTransactionId);
+            return false;
+        }
+
+        HeapTupleSetHintBits(tuple, buffer, HEAP_XMIN_COMMITTED, HeapTupleHeaderGetRawXmin(tuple));
```
surendravishnoi: could this create problem for APG side consistency checks? We are marking row as committed even though xmin transaction might not have been committed. Wouldn't consistency check will detect it and report as problematic?

contrib/babelfishpg_tsql/src/table_variable_mvcc.c
```diff
@@ -0,0 +1,1094 @@
...
+    }
+
+    /* by here, the inserting transaction is "considered" committed */
+
+    if ((tuple->t_infomask & HEAP_XMAX_INVALID) || HEAP_XMAX_IS_LOCKED_ONLY(tuple->t_infomask)) /* Cases 1.1.1 and 1.1.3 in comment section */
+        return true;
+
+    if (tuple->t_infomask & HEAP_XMAX_IS_MULTI)
+    {
+        TransactionId xmax = HeapTupleGetUpdateXid(tuple);
```
surendravishnoi: can we write small utility functions for same functionality, like MULTI case is same for both xmin committed and not. Same should be done wherever possible

contrib/babelfishpg_tsql/src/table_variable_mvcc.c
```diff
@@ -0,0 +1,1094 @@
...
+        /* already checked above */
+        Assert(!HEAP_XMAX_IS_LOCKED_ONLY(tuple->t_infomask));
+
+        /* not LOCKED_ONLY, so it has to have an xmax */
+        Assert(TransactionIdIsValid(xmax));
+
+        if (TransactionIdIsCurrentTransactionId(xmax))
+            return (HeapTupleHeaderGetCmax(tuple) >= snapshot->curcid); /* Case 1.1.4 and 2.4.1 */
+
+        return (find_failed_transaction(xmax));  /* Case 1.1.2. It must have aborted unexpectedly */
```
surendravishnoi: why failed transaction check is after current txn xmax check? What if there was failure after deleting in the same transaction?

contrib/babelfishpg_tsql/src/table_variable_mvcc.c
```diff
@@ -0,0 +1,1094 @@
...
+}
+
+
+/*
+ * TVHeapTupleSatisfiesDirty
+ *		True iff heap tuple is valid including effects of open transactions.
+ *
+ * See SNAPSHOT_DIRTY's definition for the intended behaviour.
+ *
+ * This is essentially like HeapTupleSatisfiesSelf as far as effects of
```
surendravishnoi: is this even a valid use case for table variables? table variables are only visible to a single connection/transaction

contrib/babelfishpg_tsql/src/table_variable_mvcc.c
```diff
@@ -0,0 +1,1094 @@
...
+	Assert(ItemPointerIsValid(&htup->t_self));
+	Assert(htup->t_tableOid != InvalidOid);
+
+	if (!HeapTupleHeaderXminCommitted(tuple))
+	{
+		if (HeapTupleHeaderXminInvalid(tuple))
+			return false;
+
+		/*
+		 * An invalid Xmin can be left behind by a speculative insertion that
```
surendravishnoi, line 307: does speculative insert applies to TV?

contrib/babelfishpg_tsql/src/table_variable_mvcc.c
```diff
@@ -0,0 +1,1094 @@
...
+            return TVHeapTupleSatisfiesDirty(tup, snapshot, buffer);
+            break;
+        case SNAPSHOT_TOAST:
+             return TVHeapTupleSatisfiesToast(tup, snapshot, buffer);
+             break;
+        case SNAPSHOT_NON_VACUUMABLE:
+             return TVHeapTupleSatisfiesNonVacuumable(tup, snapshot, buffer);
+             break;
+        case SNAPSHOT_HISTORIC_MVCC:
+        //     return HeapTupleSatisfiesHistoricMVCC(tup, snapshot, buffer);
```
surendravishnoi: cleanup comments

contrib/babelfishpg_tsql/src/table_variable_mvcc.c
```diff
@@ -0,0 +1,1094 @@
...
+            if (tuple->t_infomask & HEAP_XMAX_INVALID)  /* xid invalid */
+                return TM_Ok;
+
+            if (HEAP_XMAX_IS_LOCKED_ONLY(tuple->t_infomask))
+            {
+                TransactionId xmax;
+
+                xmax = HeapTupleHeaderGetRawXmax(tuple);
+
+                /*
```
surendravishnoi: Does this apply to TVs? can there be any parallel update/locking?

contrib/babelfishpg_tsql/src/table_variable_mvcc.c
```diff
@@ -0,0 +1,1094 @@
...
+ * In contrast to HeapTupleSatisfiesVacuum this routine, when encountering a
+ * tuple that could still be visible to some backend, stores the xid that
+ * needs to be compared with the horizon in *dead_after, and returns
+ * HEAPTUPLE_RECENTLY_DEAD. The caller then can perform the comparison with
+ * the horizon.  This is e.g. useful when comparing with different horizons.
+ *
+ * Note: HEAPTUPLE_DEAD can still be returned here, e.g. if the inserting
+ * transaction aborted.
+ */
+HTSV_Result
```
surendravishnoi, line 534: can we make it simpler by just detecting not alive tuples and return live for the rest. Does it matter to return various insert/delete states?
xmin is invalid
xmin with an error
xmax is valid (without an error)

contrib/babelfishpg_tsql/src/table_variable_mvcc.c
lejaokri: `TransactionIdIsCurrentTransactionId(xid)` is true if xid is the current top transaction or xid is one the subxids in the current transaction. However, an aborted subxid is removed from the subxids so it won't appear as it belong to the current transaction anymore.  if xmax is current transaction then it shouldve not failed yet, it's either still in progress or one that already committed.

contrib/babelfishpg_tsql/src/table_variable_mvcc.c
lejaokri: good point. Let me surround this with is committed check same as PG.

contrib/babelfishpg_tsql/src/table_variable_mvcc.c
lejaokri: sure will do

contrib/babelfishpg_tsql/src/table_variable_mvcc.c
lejaokri: Yes this is used during index insert.

```
552                 /*
 553                  * Check if there's any table tuples for this index entry
 554                  * satisfying SnapshotDirty. This is necessary because for AMs
 555                  * with optimizations like heap's HOT, we have just a single
 556                  * index entry for the entire chain.
 557                  */
 558                 else if (table_index_fetch_tuple_check(heapRel, &htid,
 559                                                        &SnapshotDirty,
 560                                                        &all_dead))
```

contrib/babelfishpg_tsql/src/pl_handler.c
lejaokri: done

contrib/babelfishpg_tsql/src/table_variable_mvcc.c
lejaokri: xmax is either the top xid or a subxid and it should still be active or have committed otherwise `TransactionIdIsCurrentTransactionId(xmax)` wont return true

contrib/babelfishpg_tsql/src/table_variable_mvcc.c
lejaokri: Probably not but the check is cheap only checks the hint bit so keeping this one

contrib/babelfishpg_tsql/src/table_variable_mvcc.c
lejaokri: no parallel locking. But it could be locked by an aborted subtransaction at that point the transaction wont appear as it belongs to the current transaction and it's considered "another" transaction

contrib/babelfishpg_tsql/src/table_variable_mvcc.c
lejaokri: looking into this

lejaokri: done

contrib/babelfishpg_tsql/src/table_variable_mvcc.c
```diff
@@ -0,0 +1,587 @@
...
+            {
+                HeapTupleSetHintBits(tuple, buffer, HEAP_XMAX_INVALID, InvalidTransactionId);
+                return true;                                    /* Case 1.1.2 */
+            }
+
+            return false;   /* Case 2.3 in comment section */
+        }
+
+        /* Table Variables are per-session so if xmin is active it would be true above */
+        Assert(!TransactionIdIsInProgress(HeapTupleHeaderGetRawXmin(tuple)));
```
surendravishnoi, line 253: this could be problematic as it assumes that table variables are accesses only by self. It is true for user access but what about vacuum or certain utilities which look at the heap as a whole? They can still come and see these transactions running?

contrib/babelfishpg_tsql/src/table_variable_mvcc.c
```diff
@@ -0,0 +1,587 @@
...
+	{
+		if (HeapTupleHeaderXminInvalid(tuple))
+			return false;
+
+		/*
+		 * An invalid Xmin can be left behind by a speculative insertion that
+		 * is canceled by super-deleting the tuple.  This also applies to
+		 * TOAST tuples created during speculative insertion.
+		 */
+		else if (!TransactionIdIsValid(HeapTupleHeaderGetXmin(tuple)))
```
surendravishnoi, line 311: how is this condition diff from one using HeapTupleHeaderXminInvalid(tuple)?

contrib/babelfishpg_tsql/src/table_variable_mvcc.c
```diff
@@ -0,0 +1,587 @@
...
+            HeapTupleSetHintBits(tuple, buffer, HEAP_XMAX_COMMITTED, HeapTupleHeaderGetRawXmax(tuple));
+        else if (find_failed_transaction(HeapTupleHeaderGetRawXmax(tuple)))
+        {
+            HeapTupleSetHintBits(tuple, buffer, HEAP_XMIN_INVALID, InvalidTransactionId);
+            return HEAPTUPLE_LIVE;
+        }
+    }
+
+    /* Deleter is considered committed */
+    return HEAPTUPLE_DEAD;
```
surendravishnoi: why are we considering a deleted tuple as dead? The decision should be based on oldest xmin, not based on whether tuple is visible or not

test/JDBC/expected/BABEL-3865.out
```diff
@@ -0,0 +1,140 @@
...
+GO
+
+/* Test a variety of babelfish-specific datatypes */
+CREATE TABLE #temp_2(
+	a sysname,
+	b datetime,
+	c datetime2,
+	d nchar,
+	e nvarchar,
+	f varbinary
```
surendravishnoi: let us add numeric, money as well

test/JDBC/expected/BABEL-3865.out
```diff
@@ -0,0 +1,140 @@
...
+GO
+
+DROP TYPE typc
+GO
+
+/* Test Ownership + Roles, for pg_shdepend */
+CREATE ROLE role_a
+GO
+
+CREATE TABLE #temp_5 (a int)
```
surendravishnoi, line 126: does this table go into ENR or not? if yes, is it OK to have a dependency on role which is part of catalog?

test/JDBC/expected/table_variable_xact_basic.out
```diff
@@ -0,0 +1,756 @@
...
+    DECLARE @tv1 tv_table_primary_key;
+    INSERT INTO @tv1 VALUES(1, 1, 'First')
+    INSERT INTO @tv1 VALUES(1, 1, 'Second')     -- Heap insert succeeds, index insert fails.
+ROLLBACK
+INSERT INTO @tv1 VALUES(1, 1, 'Second')         -- index still valid. should fail with duplicate key
+SELECT * FROM @tv1                              -- Should only show one record
+GO
+~~ROW COUNT: 1~~
+
+~~ERROR (Code: 2627)~~
```
surendravishnoi, line 469: why there is only one error message instead of 2, one for each insert?

test/JDBC/expected/table_variable_xact_basic.out
```diff
@@ -0,0 +1,756 @@
...
+
+
+
+
+
+
+-------------------------------------------------------------------------------
+-- Test 12: Table Variable used with CURSOR
+-------------------------------------------------------------------------------
+DECLARE @source TABLE(c1 INT, c2 INT)
```
surendravishnoi, line 714: can we have a simple cursor test where cursor cannot see new updates/insert?

test/JDBC/expected/table_variable_xact_basic_isolation_snapshot.out
```diff
@@ -0,0 +1,758 @@
+SET TRANSACTION ISOLATION LEVEL SNAPSHOT
```
surendravishnoi, line 1: snapshot behaviour really comes into picture when there are parallel commits to this transaction. I am not sure if we can use multi connections in single JDBC test but if possible we should add those tests as well

test/JDBC/expected/table_variable_xact_errors.out
```diff
@@ -0,0 +1,170 @@
+
```
surendravishnoi, line 1: we should also add few delete and truncate test cases as well

contrib/babelfishpg_tsql/src/table_variable_mvcc.c
```diff
@@ -0,0 +1,587 @@
...
+ *    1.1.2. XMAX failed unexpectedly
+ *    1.1.3. XMAX is in-progress
+ *
+ * 2. A row is INVISIBLE when:
+ *  2.0. XMIN INVALID hint bit is set by previous scan operations
+ *  2.1. XMIN Failed unexpectedly
+ *  2.3. XMAX committed or clean Rollback
+ */
+static bool
+TVHeapTupleSatisfiesDirtyOrSelf(HeapTuple htup, Snapshot snapshot, Buffer buffer)
```
surendravishnoi, line 217: do we have test cases which lead to calling this in the context of both dirty and self?

contrib/babelfishpg_tsql/src/table_variable_mvcc.c
```diff
@@ -0,0 +1,587 @@
...
+}
+
+
+/*
+ * TVHeapTupleSatisfiesUpdate
+ *
+ * Counterpart of HeapTupleSatisfiesUpdate.
+ * The only difference is this function is not rollback sensitive.
+ */
+TM_Result TVHeapTupleSatisfiesUpdate(HeapTuple htup, CommandId curcid, Buffer buffer)
```
surendravishnoi, line 392: do we have sufficient test case for this to cover most of the if/else parts?

contrib/babelfishpg_tsql/src/table_variable_mvcc.c
```diff
@@ -0,0 +1,587 @@
...
+ *  running transaction.  If so, it can't be removed yet by VACUUM.
+ *
+ * OldestXmin is a cutoff XID (obtained from
+ * GetOldestNonRemovableTransactionId()).  Tuples deleted by XIDs >=
+ * OldestXmin are deemed "recently dead"; they might still be visible to some
+ * open transaction, so we can't remove them, even if we see that the deleting
+ * transaction has committed.
+ */
+HTSV_Result
+TVHeapTupleSatisfiesVacuum(HeapTuple htup, TransactionId OldestXmin,
```
surendravishnoi, line 500: this is not tested at all right now. We should look into the possibility of how to force vacuum in a test using a parallel connection. Also, use utilities to dump heap pages to make sure we have correct state (pg inspect or similar)

contrib/babelfishpg_tsql/src/table_variable_mvcc.c
lejaokri: the table is only accessible by the session that created it. other backends should not be able to access temp tables owned by other backends, several checks in the code to prevent this. For instance, autovacuum skips temp tables because it cannot access the table.

The only time we call vacuum/prune is during page scan but even at that point, it's the same session calling prune on its own temp tables

contrib/babelfishpg_tsql/src/table_variable_mvcc.c
lejaokri: `HeapTupleHeaderXminInvalid` checks for a hint bit. If the hint bit is not set, we go here.

lejaokri:
```
329 #define HeapTupleHeaderXminInvalid(tup) \
330 ( \
331     ((tup)->t_infomask & (HEAP_XMIN_COMMITTED|HEAP_XMIN_INVALID)) == \
332         HEAP_XMIN_INVALID \
333 )
```

contrib/babelfishpg_tsql/src/table_variable_mvcc.c
lejaokri: The table is session local. If it's not visible to the session then it is prunable.

test/JDBC/expected/BABEL-3865.out
lejaokri: @timchang514

test/JDBC/expected/BABEL-3865.out
lejaokri: @timchang514

test/JDBC/expected/table_variable_xact_basic_isolation_snapshot.out
lejaokri: We are dealing with temp tables. It's per session so I didnt need to test concurrency. It's under this assumption that we're able to simplify visibility. 

This test is still important for table variables because table variables are non-transactional and I wanted to test with reused snapshots with older xmin/xmax

contrib/babelfishpg_tsql/src/table_variable_mvcc.c
lejaokri: Yes. I have testcases with index which test both code path.

contrib/babelfishpg_tsql/src/table_variable_mvcc.c
lejaokri: update and delete testcases are included in the tests. I also have a case to test halloween problem is not hit which test this code path as well. Also the cursor test goes to the if/else cases when I ran through gdb

contrib/babelfishpg_tsql/src/table_variable_mvcc.c
lejaokri: It is tested. page scan calls `heap_page_prune_opt()` which calls this code path. I also have a testcase that forces sequence scan and makes sure the ordering of the data on the page is correct after updates. This is to verify which ctid is visible and which arent.

test/JDBC/expected/table_variable_xact_errors.out
lejaokri: I do have delete cases in basic.sql and nested.sql. Also truncate is not supported for table variables and query will report syntax error and wont even run.

test/JDBC/expected/table_variable_xact_basic.out
lejaokri: Good eyes. I think this is JDBC issue. RUnning this through sqlcmd prints two errors.

```
-------------------------------------------------------------------------------
-- Test 7: Primary Key metadata should be accessible after rollback
-------------------------------------------------------------------------------
BEGIN TRANSACTION
    DECLARE @tv1 tv_table_primary_key;
    INSERT INTO @tv1 VALUES(1, 1, 'First')
    INSERT INTO @tv1 VALUES(1, 1, 'Second')     -- Heap insert succeeds, index insert fails.
ROLLBACK
INSERT INTO @tv1 VALUES(1, 1, 'Second')         -- index still valid. should fail with duplicate key
SELECT * FROM @tv1                              -- Should only show one record


(1 rows affected)
Msg 2627, Level 14, State 1, Server BABELFISH, Line 8
duplicate key value violates unique constraint "@tv1_0_pkey"
Msg 2627, Level 14, State 1, Server BABELFISH, Line 10
duplicate key value violates unique constraint "@tv1_0_pkey"
c1          b           c
----------- ----------- ---------------
          1           1 First
                                                                                                                                                                      (1 rows affected)
                                                                                                                                                                      DROP TYPE tv_table_primary_key
```

test/JDBC/expected/table_variable_xact_basic.out
lejaokri: This is what the testcase is already doing.
It updates the records but cursor does not see the newly updated rows.
I also unit tested this on gdb. 

```
DECLARE @source TABLE(c1 INT, c2 INT)
INSERT INTO @source VALUES(1, 2), (2,3), (4,5), (6,7), (8,9)

DECLARE @target INT

DECLARE cur CURSOR
FOR SELECT c1 FROM @source
FOR UPDATE OF c1

OPEN cur
FETCH NEXT FROM cur INTO @target
WHILE @@FETCH_STATUS = 0 BEGIN

    UPDATE @source set c1 = c1 * 10 WHERE c1 = @target
    FETCH NEXT FROM cur INTO @target -- next item
End

-- housekeeping
CLOSE cur
DEALLOCATE cur

SELECT * FROM @source

GO
```

contrib/babelfishpg_tsql/src/table_variable_mvcc.c
surendravishnoi: still seems risky to me. There could be multiple visibility points in the same session (cursor) and not being visible to latest does not mean it is not visible anywhere else.

insert
@cursor1
delete
visibility check -> cursor1 can still see it.
You can argue that pruning will not happen in such a use case where cursor is involved but I think it is always better to be on safe side when vacuuming.

contrib/babelfishpg_tsql/src/table_variable_mvcc.c
surendravishnoi: that is good but still we can add couple of tests with explicit vacuum call.

lejaokri: I dont think we can run vacuum in sqlcmd

```
[clouddesk] % sqlcmd -S localhost -U jdbc_user -P 1
1> declare @tv table(c1 int, c2 int);
2> vacuum @tv
3> go
Msg 33557097, Level 16, State 1, Server BABELFISH, Line 2
syntax error near '@tv' at line 2 and character position 7
```

I also could not create table variables on psql port

```
jdbc_testdb=# set babelfishpg_tsql.sql_dialect='tsql';
SET
jdbc_testdb=# declare @tv table(c1 int);
ERROR:  syntax error at or near "table"
LINE 1: declare @tv table(c1 int);
```

lejaokri: I dont think we can run vacuum in sqlcmd

```
[clouddesk] % sqlcmd -S localhost -U jdbc_user -P 1
1> declare @tv table(c1 int, c2 int);
2> vacuum @tv
3> go
Msg 33557097, Level 16, State 1, Server BABELFISH, Line 2
syntax error near '@tv' at line 2 and character position 7
```

I also could not create table variables on psql port

```
jdbc_testdb=# set babelfishpg_tsql.sql_dialect='tsql';
SET
jdbc_testdb=# declare @tv table(c1 int);
ERROR:  syntax error at or near "table"
LINE 1: declare @tv table(c1 int);
```

test/JDBC/expected/BABEL-3865.out
timchang514: yes, it goes into ENR. so there is one peculiar case where a role could be dropped from a different session but not from the local session with temp tables. I think this is ok to have, since functionality isn't lost if the role is dropped from a separate session. I also don't see a simple way to resolve this

test/JDBC/expected/BABEL-3865.out
timchang514: added

surendravishnoi (approved):

macwoody (approved):

PR #1453, opened 2023-04-19
Rework logic of TDS SSL Pre-login Handshake

thephantomthief:
This commit rewrites the logic of SSL prelogin handshake read. Previously, whatever segments we were reading from the SSL prelogin handshake packet, if the first byte started with TDS_PRELOGIN, then we assumed that the segment is actually the first segment of the SSL handshake packet from the client containing the TDS prelogin header.

This assumption is incorrect. It is possible that the segment being read by the SSL Handshake read method might not contain the TDS prelogin header but still has its first byte as TDS_PRELOGIN. There is no restriction on this for the encrypted handshake message.

We rewrite this logic to check for the TDS prelogin header only if we have fully read the entire previous SSL handshake packet from client or if this is actually the very first SSL packet from client (client hello).

Task: BABEL-4096

### Check List

### DIFF
--- a/contrib/babelfishpg_tds/src/backend/tds/tdssecure.c
+++ b/contrib/babelfishpg_tds/src/backend/tds/tdssecure.c
@@ -41,6 +41,26 @@
 int			tds_ssl_min_protocol_version;
 int			tds_ssl_max_protocol_version;
 #ifdef USE_SSL
+
+/*
+ * The SSL packets is packed in a TDS prelogin packet.  The following
+ * structure is used to store the TDS prelogin header.  It's not safe
+ * to access inidividual union members until we read the entire header
+ * in the buf.
+ */
+typedef union TDSPacketHeader {
+	struct header {
+		uint8_t		pkt_type;
+		uint8_t		status;
+		uint16_t	length;
+	} header;
+	char buf[TDS_PACKET_HEADER_SIZE];
+} TDSPacketHeader;
+
+/* tracks how much packet data we've read */
+static int	pkt_bytes_read = 0;
+static TDSPacketHeader pkt_data = {0};
+
 /*
  * SslRead - TDS secure read function, similar to my_sock_read
  */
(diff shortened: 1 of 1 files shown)


Deepesh125 (approved):

contrib/babelfishpg_tds/src/backend/tds/tdssecure.c
```diff
@@ -69,72 +89,77 @@ SslRead(BIO * h, char *buf, int size)
...
-			if ((res = SslRead(h, buf, size)) <= 0)
-				return res;
+			header_left -= res;
 		}
-		else
-		{
-			int			tmp_res = 0;
-			int			i = TDS_PACKET_HEADER_SIZE;
 
-			for (i = TDS_PACKET_HEADER_SIZE; i < res; i++)
```
surendravishnoi, line 120: Earlier we were ignoring header data, why is it OK now?

contrib/babelfishpg_tds/src/backend/tds/tdssecure.c
```diff
@@ -69,72 +89,77 @@ SslRead(BIO * h, char *buf, int size)
...
+	Assert(pkt_bytes_read >= TDS_PACKET_HEADER_SIZE);
+
+	/*
+	 * If SSL packet expands the current TDS prelogin packet, we need to
+	 * read some data in next iteration.  For now, just read till end of the
+	 * packet.  The caller will try to read the remaining data in the next
+	 * iteration.
+	 * XXX: We've not really seen this scenario.
+	 */
+	if (pkt_bytes_read + size > pkt_data.header.length)
```
surendravishnoi, line 148: does this mean that data is send as multiple pre-login packets and we are reading single packet at a time? How does this align with SSL handshake data?

contrib/babelfishpg_tds/src/backend/tds/tdssecure.c
thephantomthief: We are reading the header data only for fetching the packet length. Other than that we are still ignoring it as before and reading remaining data to account for the header bytes.

contrib/babelfishpg_tds/src/backend/tds/tdssecure.c
thephantomthief: As mentioned we haven't really seen this scenario. It is just an attempt to handle and log if we ever see such a scenario in SSL handshake. Previous logic was not addressing this at all.

surendravishnoi: we are handling it with assumption that additional data must belong to next packet. Have we made sure that it is always the case for test cases we have seen so far? SSL handshake will always be done using 1 or more pre-login packets where each packet has header in the start and followed by data tracked with length field.

kuntalghosh: As per the TDS protocol, the SSL packets will always be encapsulated in a TDS prelogin packet. It means two possible scenario:
1. One or more SSL packets are sent in a single TDS prelogin packet. We've seen this.
2. If the combined length SSL packets that should be sent together is more than TDS prelogin length, it'll be split in multiple TDS prelogin packets. We've not seen this scenario yet.

These possibilities are not specified in the protocol doc. But, we can assume it's a possibility if not explicitly mentioned in the doc. The above code tries to handle the second scenario. Another option is to throw an error as 'unhandled'. If we get an issue, we can fix the same.

surendravishnoi (approved):

PR #1470, opened 2023-04-27
Add additional tests for DML w/ TOP

wboettge:
### Description

Adds a tests that ensure that Postgres behavior remains unchanged despite earlier changes to PG parsenodes for Update, Delete, and Insert statements.

Insert should allow LIMIT clauses, but Update and Delete should not.


forestkeeper (approved):
LGTM

PR #1501, opened 2023-05-09
Fix UNION fixed-len type issues

wboettge:
Currently, Babelfish gives an error when UNION is used with fixed-length types of different lengths or null. This change fixes how types and typmods are handled within UNION and VALUES via two new hooks.

### Issues Resolved

BABEL-3348, BABEL-3392, BABEL-4157, BABEL-1874

### Check List


contrib/babelfishpg_tsql/src/backend_parser/gram-tsql-rule.y
```diff
@@ -1957,8 +1957,14 @@ AexprConst:
 				{
 					/* This is to support N'str' in various locations */
 					TypeName *t = makeTypeNameFromNameList(list_make2(makeString("sys"), makeString("nvarchar")));
+					/* Include a typmod based on the length of the literal */
+					int32 typmod = strlen($2);
+					if (typmod == 0)
+						typmod = 2; /* typmod can't be 0 */
+					else if (typmod > 4000)
+						typmod = TSQLMaxTypmod;
```
Deepesh125, line 1979: This is length of input, typmode could be length + VARHDRSZ. Please double check if this is what we wanted to do

contrib/babelfishpg_tsql/src/pltsql_coerce.c
```diff
@@ -1084,6 +1096,120 @@ tsql_coerce_string_literal_hook(ParseCallbackState *pcbstate, Oid targetTypeId,
...
+	{
+		Node	*expr = (Node *) lfirst(lc);
+		Oid		type = exprType(expr);
+
+		if (expr_is_null(expr))
+			continue;
+		else if (is_tsql_str_const(expr))
+			type = common_utility_plugin_ptr->lookup_tsql_datatype_oid("varchar");
+		else if (!is_tsql_char_type_with_len(type))
+			return InvalidOid;
```
Deepesh125, line 1178: Didnt get this part. Can you please explain?

contrib/babelfishpg_tsql/src/pltsql_coerce.c
```diff
@@ -1084,6 +1096,120 @@ tsql_coerce_string_literal_hook(ParseCallbackState *pcbstate, Oid targetTypeId,
...
+	Oid			result_type = InvalidOid;
+	ListCell	*lc;
+
+	if (sql_dialect != SQL_DIALECT_TSQL)
+		return InvalidOid;
+
+	/* Limit changes for Set Operations and Values only for now
+	 * We may want to expand to other callers in the future */
+	if (context && strcmp(context, "UNION") != 0 && strcmp(context, "INTERSECT") != 0 
+		&& strcmp(context, "INTERSECT") != 0 && strcmp(context, "VALUES") != 0)
```
Deepesh125: Comparing "INTERSECT" twice

contrib/babelfishpg_tsql/src/tsql_analyze.c
```diff
@@ -313,34 +317,186 @@ pre_transform_setop_tree(SelectStmt *stmt, SelectStmt *leftmostSelect)
...
+		if (IsA(setOp, SetOperationStmt))
+		{
+			SetOperationStmt *op = (SetOperationStmt *) setOp;
+			setOpNodes = lappend(setOpNodes, op);
+			setOpTreeStack = lappend(setOpTreeStack, op->rarg);
+			setOpTreeStack = lappend(setOpTreeStack, op->larg);
+		} else if (IsA(setOp, RangeTblRef))
+		{
+			RangeTblEntry *rte = rt_fetch(((RangeTblRef*)setOp)->rtindex, pstate->p_rtable);
+			List *targetList = rte->subquery->targetList;
```
Deepesh125: will there always be subtree? does it make sense to check it before accessing it?

contrib/babelfishpg_tsql/src/tsql_analyze.c
```diff
@@ -313,34 +317,186 @@ pre_transform_setop_tree(SelectStmt *stmt, SelectStmt *leftmostSelect)
...
+			ListCell *tlistl, *collistl;
+
+			if(collist_list == NIL)
+			{
+				foreach(tlistl, targetList)
+				{
+					TargetEntry *tle = (TargetEntry*) lfirst(tlistl);
+					collist_list = lappend(collist_list, list_make1(tle));
+				}
+			} else
```
Deepesh125: Nit: else should be on next line

contrib/babelfishpg_tsql/src/tsql_analyze.c
```diff
@@ -313,34 +317,186 @@ pre_transform_setop_tree(SelectStmt *stmt, SelectStmt *leftmostSelect)
...
+		Var *top_expr = (Var*) top_tle->expr;
+
+
+		foreach(lc, col_tles)
+		{
+			TargetEntry *tle = (TargetEntry*) lfirst(lc);
+			col_exprs = lappend(col_exprs, (Node*)tle->expr);
+		}
+
+		common_type = select_common_type(pstate, col_exprs, "UNION", NULL);
```
Deepesh125: why "UNION" everytime?

contrib/babelfishpg_tsql/src/backend_parser/gram-tsql-rule.y
wboettge: Good point, I initially thought it should be length + VARHDRSZ as well. However, when casting to NVARCHAR, anychar_typmodin as the input function for the typmod. This is useful for cases where we explicitly set the typmod for VARCHAR and NVARCHAR, i.e. NVARCHAR(15). The input typmod would be 15, and anychar_typmodin would handle ensuring that the given value is valid, and is adjusted with VARHDRSZ. As such, we shouldn't add VARHDRSZ here since it will be added later.

contrib/babelfishpg_tsql/src/pltsql_coerce.c
wboettge: This set of conditionals applies the following logic:
1. If we see a NULL-const, move on.
2. If we see a string literal, 'foo', make the current expr type VARCHAR.
3. If we see a type that's not UNKNOWN, CHAR, NCHAR, VARCHAR, or NVARCHAR, let the engine logic handle the case. This is to limit the amount of cases this change will affect.
4. Make the result type the highest precedence seen so far.

I also updated the comment to describe this as well. Does this help explain?

forestkeeper (approved):
LGTM

PR #1506, opened 2023-05-16
Support extended property

ghost:
Description
===========
The extended property includes the following features: sys.sp_addextendedproperty: Adds a new extended property to a database object. sys.sp_updateextendedproperty: Updates the value of an existing extended property. sys.sp_dropextendedproperty: Drops an existing extended property. sys.fn_listextendedproperty: Returns extended property values of database objects. sys.extended_properties: Returns a row for each extended property in the current database.

Solution
========
The implementation of extended property is carried out in the following manner: Firstly, a new catalog table is added, which records details such as dbid, schema, major_name, minor_name, and type. 'sys.extended_properties' is view based on this table.
Secondly, we implement 'sp_addextendedproperty', 'sp_updateextendedproperty', 'sp_dropextendedproperty', and 'fn_listextendedproperty' procedure or function.
Lastly, we modify major_name and minor_name cascade when calling 'sp_rename' and delete cascade when dropping an object.

Limitation
==========
At present, we only support extended properties for the most commonly used objects, which include database, schema, table, table column, view, sequence, procedure, function, and type.

### DIFF
--- a/contrib/babelfishpg_tsql/sql/ownership.sql
+++ b/contrib/babelfishpg_tsql/sql/ownership.sql
@@ -612,3 +612,64 @@ GRANT EXECUTE ON PROCEDURE sys.babelfish_remove_domain_mapping_entry TO PUBLIC;
 CREATE OR REPLACE PROCEDURE sys.babelfish_truncate_domain_mapping_table()
   AS 'babelfishpg_tsql', 'babelfish_truncate_domain_mapping_table_internal' LANGUAGE C;
 GRANT EXECUTE ON PROCEDURE sys.babelfish_truncate_domain_mapping_table TO PUBLIC;
+
+CREATE TABLE sys.babelfish_extended_properties (
+  dbid smallint NOT NULL,
+  schema_name name NOT NULL,
+  major_name name NOT NULL,
+  minor_name name NOT NULL,
+  type sys.varchar(50) NOT NULL,
+  name sys.sysname NOT NULL,
+  orig_name sys.sysname NOT NULL,
+  value sys.sql_variant,
+  PRIMARY KEY (dbid, type, schema_name, major_name, minor_name, name)
+);
+GRANT SELECT on sys.babelfish_extended_properties TO PUBLIC;
+
+CREATE OR REPLACE VIEW sys.extended_properties
+AS
+SELECT
+	CAST((CASE
+		WHEN ep.type = 'DATABASE' THEN 0
+		WHEN ep.type = 'SCHEMA' THEN 3
+		WHEN ep.type IN ('TABLE', 'TABLE COLUMN', 'VIEW', 'SEQUENCE', 'PROCEDURE', 'FUNCTION') THEN 1
+		WHEN ep.type = 'TYPE' THEN 6
+		END) AS sys.tinyint) AS class,
+	CAST((CASE
+		WHEN ep.type = 'DATABASE' THEN 'DATABASE'
+		WHEN ep.type = 'SCHEMA' THEN 'SCHEMA'
+		WHEN ep.type IN ('TABLE', 'TABLE COLUMN', 'VIEW', 'SEQUENCE', 'PROCEDURE', 'FUNCTION') THEN 'OBJECT_OR_COLUMN'
+		WHEN ep.type = 'TYPE' THEN 'TYPE'
+	END) AS sys.nvarchar(60)) AS class_desc,
+	CAST((CASE
+		WHEN ep.type = 'DATABASE' THEN 0
+		WHEN ep.type = 'SCHEMA' THEN n.oid
+		WHEN ep.type IN ('TABLE', 'TABLE COLUMN', 'VIEW', 'SEQUENCE') THEN c.oid
+		WHEN ep.type IN ('PROCEDURE', 'FUNCTION') THEN p.oid
+		WHEN ep.type = 'TYPE' THEN t.oid
+	END) AS int) AS major_id,
+	CAST((CASE
+		WHEN ep.type = 'DATABASE' THEN 0
+		WHEN ep.type = 'SCHEMA' THEN 0
+		WHEN ep.type IN ('TABLE', 'VIEW', 'SEQUENCE', 'PROCEDURE', 'FUNCTION', 'TYPE') THEN 0
+		WHEN ep.type = 'TABLE COLUMN' THEN a.attnum
+	END) AS int) AS minor_id,
+	ep.orig_name AS name, ep.value AS value
+	FROM sys.babelfish_extended_properties ep
+		LEFT JOIN sys.babelfish_namespace_ext ne ON ne.dbid = sys.db_id() AND ne.orig_name = ep.schema_name COLLATE "C"
+		LEFT JOIN pg_catalog.pg_namespace n ON n.nspname = ne.nspname
+		LEFT JOIN pg_catalog.pg_class c ON c.relname = ep.major_name AND c.relnamespace = n.oid
+		LEFT JOIN pg_catalog.pg_proc p ON p.proname = ep.major_name AND p.pronamespace = n.oid
+		LEFT JOIN pg_catalog.pg_type t ON t.typname = ep.major_name AND t.typnamespace = n.oid
+		LEFT JOIN pg_catalog.pg_attribute a ON a.attrelid = c.oid AND a.attname = ep.minor_name
+	WHERE ep.dbid = sys.db_id() AND
+	(CASE
+		WHEN ep.type = 'DATABASE' THEN true
+		WHEN ep.type = 'SCHEMA' THEN has_schema_privilege(n.oid, 'USAGE, CREATE')
+		WHEN ep.type IN ('TABLE', 'VIEW', 'SEQUENCE') THEN (has_table_privilege(c.oid, 'SELECT, INSERT, UPDATE, DELETE, TRUNCATE, REFERENCES, TRIGGER'))
+		WHEN ep.type IN ('TABLE COLUMN') THEN (has_table_privilege(c.oid, 'SELECT, INSERT, UPDATE, DELETE, TRUNCATE, REFERENCES, TRIGGER') OR has_column_privilege(a.attrelid, a.attname, 'SELECT, INSERT, UPDATE, REFERENCES'))
+		WHEN ep.type IN ('PROCEDURE', 'FUNCTION') THEN has_function_privilege(p.oid, 'EXECUTE')
+		WHEN ep.type = 'TYPE' THEN has_type_privilege(t.oid, 'USAGE')
+	END)
+	ORDER BY class, class_desc, major_id, minor_id, ep.orig_name;
+GRANT SELECT ON sys.extended_properties TO PUBLIC;
--- a/contrib/babelfishpg_tsql/sql/sys_functions.sql
+++ b/contrib/babelfishpg_tsql/sql/sys_functions.sql
@@ -2069,61 +2069,6 @@ CREATE OR REPLACE FUNCTION sys.db_name() RETURNS sys.nvarchar(128)
 AS 'babelfishpg_tsql', 'babelfish_db_name'
 LANGUAGE C PARALLEL SAFE IMMUTABLE;
 
--- BABEL-1783: (partial) support for sys.fn_listextendedproperty
-create table if not exists sys.extended_properties (
-class sys.tinyint,
-class_desc sys.nvarchar(60),
-major_id int,
-minor_id int,
-name sys.sysname,
-value sys.sql_variant
-);
-GRANT SELECT ON sys.extended_properties TO PUBLIC;
-
-CREATE OR REPLACE FUNCTION sys.fn_listextendedproperty (
-property_name varchar(128),
-level0_object_type varchar(128),
-level0_object_name varchar(128),
-level1_object_type varchar(128),
-level1_object_name varchar(128),
-level2_object_type varchar(128),
-level2_object_name varchar(128)
-)
-returns table (
-objtype	sys.sysname,
-objname	sys.sysname,
-name	sys.sysname,
-value	sys.sql_variant
-) 
-as $$
-begin
--- currently only support COLUMN property
-IF (((SELECT coalesce(property_name COLLATE sys.database_default, '')) = '') or
-    ((SELECT UPPER(coalesce(property_name COLLATE sys.database_default, ''))) = 'COLUMN')) THEN
-	IF (((SELECT LOWER(coalesce(level0_object_type COLLATE sys.database_default, ''))) = 'schema') and
-	    ((SELECT LOWER(coalesce(level1_object_type COLLATE sys.database_default, ''))) = 'table') and
-	    ((SELECT LOWER(coalesce(level2_object_type COLLATE sys.database_default, ''))) = 'column')) THEN
-		RETURN query 
-		select CAST('COLUMN' AS sys.sysname) as objtype,
-		       CAST(t3.column_name AS sys.sysname) as objname,
-		       t1.name as name,
-		       t1.value as value
-		from sys.extended_properties t1, pg_catalog.pg_class t2, information_schema.columns t3
-		where t1.major_id = t2.oid and 
-			  t2.relname = cast(t3.table_name as sys.sysname) COLLATE sys.database_default and 
-		      t2.relname = (SELECT coalesce(level1_object_name COLLATE sys.database_default, '')) COLLATE sys.database_default and 
-			  t3.column_name = (SELECT coalesce(level2_object_name COLLATE sys.database_default, '')) COLLATE sys.database_default;
-	END IF;
-END IF;
-RETURN;
-end;
-$$
-LANGUAGE plpgsql
-STABLE;
-GRANT EXECUTE ON FUNCTION sys.fn_listextendedproperty(
-	varchar(128), varchar(128), varchar(128), varchar(128), varchar(128), varchar(128), varchar(128)
-) TO PUBLIC;
-
 CREATE OR REPLACE FUNCTION sys.exp(IN arg DOUBLE PRECISION)
 RETURNS DOUBLE PRECISION
 AS 'babelfishpg_tsql', 'tsql_exp'
@@ -3614,3 +3559,21 @@ CREATE OR REPLACE FUNCTION sys.EOMONTH(date,int DEFAULT 0)
 RETURNS date
 AS 'babelfishpg_tsql', 'EOMONTH'
 LANGUAGE C STABLE PARALLEL SAFE;
+
+CREATE OR REPLACE FUNCTION sys.fn_listextendedproperty
+(
+    IN "@name" sys.sysname DEFAULT NULL,
+    IN "@level0type" VARCHAR(128) DEFAULT NULL,
+    IN "@level0name" sys.sysname DEFAULT NULL,
+    IN "@level1type" VARCHAR(128) DEFAULT NULL,
+    IN "@level1name" sys.sysname DEFAULT NULL,
+    IN "@level2type" VARCHAR(128) DEFAULT NULL,
+    IN "@level2name" sys.sysname DEFAULT NULL,
+    OUT objtype sys.sysname,
+    OUT objname sys.sysname,
+    OUT name sys.sysname,
+    OUT value sys.sql_variant
+)
+RETURNS SETOF RECORD
+AS 'babelfishpg_tsql' LANGUAGE C STABLE;
+GRANT EXECUTE ON FUNCTION sys.fn_listextendedproperty TO PUBLIC;
--- a/contrib/babelfishpg_tsql/sql/sys_procedures.sql
+++ b/contrib/babelfishpg_tsql/sql/sys_procedures.sql
@@ -271,3 +271,44 @@ GRANT EXECUTE on PROCEDURE sys.sp_babelfish_volatility(IN sys.varchar, IN sys.va
 
 CREATE OR REPLACE PROCEDURE sys.bbf_set_context_info(IN context_info sys.VARBINARY(128))
 AS 'babelfishpg_tsql' LANGUAGE C;
+
+CREATE OR REPLACE PROCEDURE sys.sp_addextendedproperty
+(
+  "@name" sys.sysname,
+  "@value" sys.sql_variant = NULL,
+  "@level0type" VARCHAR(128) = NULL,
+  "@level0name" sys.sysname = NULL,
+  "@level1type" VARCHAR(128) = NULL,
+  "@level1name" sys.sysname = NULL,
+  "@level2type" VARCHAR(128) = NULL,
+  "@level2name" sys.sysname = NULL
+)
+AS 'babelfishpg_tsql' LANGUAGE C;
+GRANT EXECUTE ON PROCEDURE sys.sp_addextendedproperty TO PUBLIC;
+
+CREATE OR REPLACE PROCEDURE sys.sp_updateextendedproperty
+(
+  "@name" sys.sysname,
+  "@value" sys.sql_variant = NULL,
+  "@level0type" VARCHAR(128) = NULL,
+  "@level0name" sys.sysname = NULL,
+  "@level1type" VARCHAR(128) = NULL,
+  "@level1name" sys.sysname = NULL,
+  "@level2type" VARCHAR(128) = NULL,
+  "@level2name" sys.sysname = NULL
+)
+AS 'babelfishpg_tsql' LANGUAGE C;
+GRANT EXECUTE ON PROCEDURE sys.sp_updateextendedproperty TO PUBLIC;
+
+CREATE OR REPLACE PROCEDURE sys.sp_dropextendedproperty
+(
+  "@name" sys.sysname,
+  "@level0type" VARCHAR(128) = NULL,
+  "@level0name" sys.sysname = NULL,
+  "@level1type" VARCHAR(128) = NULL,
+  "@level1name" sys.sysname = NULL,
+  "@level2type" VARCHAR(128) = NULL,
+  "@level2name" sys.sysname = NULL
+)
+AS 'babelfishpg_tsql' LANGUAGE C;
+GRANT EXECUTE ON PROCEDURE sys.sp_dropextendedproperty TO PUBLIC;
--- a/contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--3.2.0--3.3.0.sql
+++ b/contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--3.2.0--3.3.0.sql
@@ -45,6 +45,132 @@ RETURNS date
 AS 'babelfishpg_tsql', 'EOMONTH'
 LANGUAGE C STABLE PARALLEL SAFE;
 
+ALTER TABLE sys.extended_properties RENAME TO extended_properties_deprecated_in_3_3_0;
+CREATE TABLE sys.babelfish_extended_properties (
+  dbid smallint NOT NULL,
+  schema_name name NOT NULL,
+  major_name name NOT NULL,
+  minor_name name NOT NULL,
+  type sys.varchar(50) NOT NULL,
+  name sys.sysname NOT NULL,
+  orig_name sys.sysname NOT NULL,
+  value sys.sql_variant,
+  PRIMARY KEY (dbid, type, schema_name, major_name, minor_name, name)
+);
+GRANT SELECT on sys.babelfish_extended_properties TO PUBLIC;
+
+CREATE OR REPLACE VIEW sys.extended_properties
+AS
+SELECT
+	CAST((CASE
+		WHEN ep.type = 'DATABASE' THEN 0
+		WHEN ep.type = 'SCHEMA' THEN 3
+		WHEN ep.type IN ('TABLE', 'TABLE COLUMN', 'VIEW', 'SEQUENCE', 'PROCEDURE', 'FUNCTION') THEN 1
+		WHEN ep.type = 'TYPE' THEN 6
+		END) AS sys.tinyint) AS class,
+	CAST((CASE
+		WHEN ep.type = 'DATABASE' THEN 'DATABASE'
+		WHEN ep.type = 'SCHEMA' THEN 'SCHEMA'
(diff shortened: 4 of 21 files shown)


markrui3:
@rcv-aws @macwoody 
There is a problem with type sys.sysname and sys.sql_variant in sys.extended_properties.
Refer to [SQL Server sys.extended_properties](https://learn.microsoft.com/en-us/sql/relational-databases/system-catalog-views/extended-properties-catalog-views-sys-extended-properties?view=sql-server-ver16), type of name and value are sys.sysname and sys.sql_variant, which don't recognize Chinese unicode character in Babelfish.
```
1> create table test(name sys.sysname, value sys.sql_variant)
2> go
1> insert into test values('你好', '世界')
2> go
1> select * from test
2> go
name    value
??      ??
(1 row affected)
```

robverschoor:
I am not very good with chinese charsets, but could this just be a charset conversion issue for the client you're using? I am pretty sure nvarchar should be able to handle chinese chars.

robverschoor:
Actually, I tried running this in SSMS against SQL Server itself (not Babelfish), and I am getting the same result. Not sure what that means (did I mention I'm not good with such charsets?)

image

contrib/babelfishpg_tsql/sql/ownership.sql
```diff
@@ -611,4 +613,16 @@ GRANT EXECUTE ON PROCEDURE sys.babelfish_remove_domain_mapping_entry TO PUBLIC;
...
-GRANT EXECUTE ON PROCEDURE sys.babelfish_truncate_domain_mapping_table TO PUBLIC;
\ No newline at end of file
+GRANT EXECUTE ON PROCEDURE sys.babelfish_truncate_domain_mapping_table TO PUBLIC;
+
+CREATE TABLE sys.babelfish_extended_properties (
+  dbid smallint NOT NULL,
+  schema_name name NOT NULL,
+  major_name name NOT NULL,
+  minor_name name NOT NULL,
+  type text NOT NULL COLLATE "C",
```
macwoody: do we really need text type?

contrib/babelfishpg_tsql/sql/sys_functions.sql
```diff
@@ -3715,3 +3660,100 @@ IN query text)
...
+    var_type := '';
+
+    -- DATABASE
+    IF "@level0type" IS NULL THEN
+        var_type := 'DATABASE';
+    END IF;
+
+    -- SCHEMA or object in SCHEMA
+    IF "@level0type" IN ('SCHEMA') THEN
+        var_schema_name := LOWER("@level0name");
```
macwoody: I wonder if trailing spaces are also ignored in this function. 
We see many cases that trailing spaces are ignore when comparing object names.

contrib/babelfishpg_tsql/sql/sys_procedures.sql
```diff
@@ -271,3 +271,223 @@ GRANT EXECUTE on PROCEDURE sys.sp_babelfish_volatility(IN sys.varchar, IN sys.va
...
+        var_object_type := ARRAY['SO'];
+      ELSIF "@level1type" = 'PROCEDURE' THEN
+        var_object_type := ARRAY['P', 'PC', 'RF', 'X'];
+      ELSIF "@level1type" = 'FUNCTION' THEN
+        var_object_type := ARRAY['AF', 'FN', 'FS', 'FT', 'IF', 'TF'];
+      ELSIF "@level1type" = 'TYPE' THEN
+        var_object_type := ARRAY['TT'];
+      END IF;
+      var_object_id := sys.object_id(var_object_name);
+      IF var_object_id IS NULL OR NOT EXISTS (SELECT * FROM sys.objects WHERE object_id = var_object_id AND type = ANY(var_object_type)) THEN
```
macwoody: be aware that sys.objects is currently a slow view. 
It may cause your feature slow when there are large amount of objects. 
Given the amount of call we are seeing, I suspect, this approach would not work for big migration project. 
As an alternative, if you know the object type, you could search dedicated view

contrib/babelfishpg_tsql/sql/sys_views.sql
```diff
@@ -2981,3 +2981,56 @@ LEFT JOIN pg_foreign_server AS f ON u.srvid = f.oid
...
+)
+AS $$
+DECLARE
+  rec record;
+  class sys.tinyint;
+  class_desc sys.nvarchar(60);
+  major_id int;
+  minor_id int;
+BEGIN
+  FOR rec IN SELECT * FROM sys.babelfish_extended_properties WHERE dbid = sys.db_id()
```
macwoody: I suspect this implementation would be very slow when large amount of objects exists. 
Why not just use a subquery?

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -5325,3 +5336,141 @@ pltsql_remove_current_query_env(void)
...
+			}
+
+			if (schema_name)
+				pfree((char *) schema_name);
+			if (major_name)
+				pfree(major_name);
+		}
+	}
+	else if (stmt->removeType == OBJECT_PROCEDURE ||
+			 stmt->removeType == OBJECT_FUNCTION ||
```
macwoody, line 5173: how about user defined type?

macwoody:
I didn't go through the test part, I feel we may need more people to screen this PR from multiple angles.

contrib/babelfishpg_tsql/sql/ownership.sql
markrui3: what type should I use if not text?

contrib/babelfishpg_tsql/sql/sys_views.sql
markrui3: sorry, I don't really know how to use a subquery. Do you mean 'case when'?

markrui3: sorry, I don't really know how to use a subquery. Do you mean 'case when'?

contrib/babelfishpg_tsql/src/pl_handler.c
markrui3: user defined type will be captured by 'stmt->removeType == OBJECT_TYPE'.

contrib/babelfishpg_tsql/sql/sys_functions.sql
robverschoor: I can confirm that fn_listextendedproperty(), as is common in SQL Server, ignores trailing spaces for its arguments, but not leading spaces (i.e. when leading spaces are present, these are not stripped, and that usually means the object is not found)

robverschoor: In addition ,please note that the examples for fn_listextendedproperty() use **default** as argument a lot, which Babelfish currently does not support. That is not an issue for this PR, but a generic issue that would have to be solved in a different context. However, this may mean the usefulness of fn_listextendedproperty() may be limited as a result

contrib/babelfishpg_tsql/sql/sys_procedures.sql
markrui3: I use sys.objects view because sys.object_id() is not accurate.
Maybe I should make sys.object_id() accurate. How do you think this way?

For example in Babelfish:
```
1> select sys.object_id('dbo.testtable', 'U')
2> GO
object_id
23104

1> select sys.object_id('dbo.testtable', 'V')
2> GO
object_id
23104
```

But in SQL Server:
```
1> select sys.object_id('dbo.testtable', 'U')
2> GO
object_id
23104

1> select sys.object_id('dbo.testtable', 'V')
2> GO
object_id
NULL
```

contrib/babelfishpg_tsql/sql/ownership.sql
macwoody: I think we just need a shortest type that could fit the intended values

contrib/babelfishpg_tsql/sql/sys_procedures.sql
macwoody: I see. Let's not workaround this bug. 
Let me file a JIRA internally and you could implement your feature assuming everything is correct

contrib/babelfishpg_tsql/sql/sys_views.sql
macwoody: Yes, just define the view with one single select statement with "case when"

test/JDBC/upgrade/latest/schedule
```diff
@@ -434,3 +434,4 @@ triggers_with_transaction
 typeproperty
 typeproperty-dep
 datetimeoffset-timezone
+BABEL-EXTENDEDPROPERTY-v2
```
macwoody: this file is supposed to be sorted alphabetically

contrib/babelfishpg_tsql/sql/babelfishpg_tsql.sql
```diff
@@ -3222,3 +3222,35 @@ BEGIN
...
+AS
+SELECT class::sys.tinyint, class_desc::sys.nvarchar(60), major_id::int,
+  CAST((CASE
+    WHEN sub.type = 'DATABASE' THEN 0
+    WHEN sub.type = 'SCHEMA' THEN 0
+    WHEN sub.type = 'TYPE' THEN 0
+    WHEN sub.type IN ('TABLE', 'TABLE COLUMN', 'VIEW', 'SEQUENCE', 'PROCEDURE', 'FUNCTION') THEN (CASE WHEN sub.type = 'TABLE COLUMN' THEN (SELECT attnum FROM pg_attribute WHERE attrelid = sub.major_id AND attname = sub.minor_name COLLATE "C") ELSE 0 END)
+  END) AS int) AS minor_id, name::sys.sysname, value::sys.sql_variant
+  FROM
+  (SELECT
```
macwoody: I feel that we don't really need this subquery, just put everything in one layer would be enough. 
But it is not a functional requirement. You can decide based on your taste.

contrib/babelfishpg_tsql/sql/sys_procedures.sql
```diff
@@ -271,3 +271,227 @@ GRANT EXECUTE on PROCEDURE sys.sp_babelfish_volatility(IN sys.varchar, IN sys.va
...
+    UPDATE sys.babelfish_extended_properties SET value = "@value" WHERE dbid = var_dbid AND schema_name = var_schema_name COLLATE "C" AND major_name = var_major_name COLLATE "C" AND minor_name = var_minor_name COLLATE "C" AND type = var_type COLLATE "C" AND name = "@name";
+  ELSIF procedure_name = 'sp_dropextendedproperty' THEN
+    DELETE FROM sys.babelfish_extended_properties WHERE dbid = var_dbid AND schema_name = var_schema_name COLLATE "C" AND major_name = var_major_name COLLATE "C" AND minor_name = var_minor_name COLLATE "C" AND type = var_type COLLATE "C" AND name = "@name";
+  END IF;
+
+  PERFORM sys.babelfish_set_role(var_user);
+END;
+$$ LANGUAGE plpgsql;
+
+CREATE OR REPLACE PROCEDURE sys.sp_addextendedproperty
```
macwoody, line 275: I wonder if there is any permission check on who could insert / update / delete an extended property?

contrib/babelfishpg_tsql/sql/sys_procedures.sql
```diff
@@ -271,3 +271,227 @@ GRANT EXECUTE on PROCEDURE sys.sp_babelfish_volatility(IN sys.varchar, IN sys.va
...
+  var_user TEXT;
+  var_dbid SMALLINT;
+  var_schema_name NAME;
+  var_major_name NAME;
+  var_minor_name NAME;
+  var_type TEXT;
+  var_object_name TEXT;
+  var_object_type TEXT[];
+BEGIN
+  "@level0type" := UPPER(RTRIM("@level0type"));
```
macwoody: I have concern about the performance of this PLPGSQL implementation. 
Could you please try how much time dose it take this stored procedure to create 10K extended properties?

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--3.1.0--3.2.0.sql
```diff
@@ -2202,6 +2202,382 @@ GRANT SELECT ON sys.servers TO PUBLIC;
 CALL sys.babelfish_drop_deprecated_object('procedure', 'sys', 'babelfish_sp_rename_internal_deprecated_in_3_2_0');
 CALL sys.babelfish_drop_deprecated_object('procedure', 'sys', 'sp_rename_deprecated_in_3_2_0');
 
+ALTER TABLE sys.extended_properties RENAME TO extended_properties_deprecated_in_3_2_0;
```
macwoody: this would not be released in 3.2, 
I think you may create a new upgrade file from 3.2 -> 3.3

test/JDBC/expected/BABEL-EXTENDEDPROPERTY-v2-vu-prepare.out
```diff
@@ -0,0 +1,92 @@
...
+BEGIN
+    SELECT * FROM fn_listextendedproperty(@name, @level0type, @level0name, @level1type, @level1name, @level2type, @level2name)
+END
+GO
+
+CREATE VIEW babel_extended_properties_view AS
+SELECT class, class_desc, IIF(major_id > 0, 1, 0) AS major_id, minor_id, name, value FROM sys.extended_properties
+GO
+
+CREATE PROCEDURE babel_extended_properties_proc AS
```
macwoody, line 82: I think you may need to clean up these created procedures

macwoody (commented):
I haven't check your test case yet.

test/JDBC/expected/BABEL-EXTENDEDPROPERTY-v2-vu-prepare.out
markrui3: These procedures are created in database babel_extended_properties_db. And database babel_extended_properties_db will be dropped at the end.

contrib/babelfishpg_tsql/sql/sys_procedures.sql
markrui3: object_id() do the permission check for us. It can return result only when the user has permission on the object.

contrib/babelfishpg_tsql/sql/sys_procedures.sql
markrui3: I called sp_addextendedproperty xx, xx, 'schema', 'dbo', 'table', 'test', 'column', 'id' 10K times and get 122 seconds.

image

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--3.1.0--3.2.0.sql
markrui3: done

contrib/babelfishpg_tsql/sql/babelfishpg_tsql.sql
markrui3: done

contrib/babelfishpg_tsql/sql/sys_procedures.sql
macwoody: Thanks! 12ms per call would be enough for the moment

contrib/babelfishpg_tsql/sql/sys_procedures.sql
macwoody: I think you are assuming that if a user could use or execute the object, the he will be able to add a property. 
Because that's how obect_id is implemented. But that's incorrect. 

According to the T-SQL doc: 

Permissions
Members of the db_owner and db_ddladmin fixed database roles can add extended properties to any object with the following exception: db_ddladmin cannot add properties to the database itself, or to users or roles.

Users can add extended properties to objects they own or have ALTER or CONTROL permissions on.

Permissions
Permissions to list extended properties of objects vary by object type.

contrib/babelfishpg_tsql/sql/sys_functions.sql
```diff
@@ -3620,3 +3565,104 @@ IN query text)
...
+                    RETURN;
+                END IF;
+
+                var_type := "@level1type" || ' ' || "@level2type";
+                var_minor_name := "@level2name";
+            END IF;
+        END IF;
+    END IF;
+
+    FOR rec IN (SELECT * FROM sys.babelfish_extended_properties ep WHERE ep.dbid = var_dbid AND ep.schema_name = coalesce(var_schema_name, ep.schema_name) COLLATE "C" AND ep.major_name = coalesce(var_major_name, ep.major_name) COLLATE "C" AND ep.minor_name = coalesce(var_minor_name, ep.minor_name) COLLATE "C" AND ep.type = var_type COLLATE "C" AND ep.name = coalesce("@name", ep.name) ORDER BY dbid, schema_name, major_name, minor_name, type, name)
```
macwoody: what are the permission rules you implemented ?

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -5325,3 +5336,169 @@ pltsql_remove_current_query_env(void)
...
+	{
+		foreach(cell, stmt->objects)
+		{
+			/* Get schema_name */
+			logical_schema_name = get_logical_schema_name(strVal(lfirst(cell)),
+														  true);
+
+			if (logical_schema_name)
+			{
+				schema_name = downcase_identifier(logical_schema_name,
```
macwoody, line 5090: why do we do downcase for identifier? 
Shouldn't the original case be stored in the extended property catalog table?

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -5325,3 +5336,169 @@ pltsql_remove_current_query_env(void)
...
+										 lfirst(cell),
+										 &relation,
+										 AccessShareLock,
+										 true);
+
+			if (!relation)
+				return;
+
+			/* Get major_name */
+			major_name = pstrdup(RelationGetRelationName(relation));
```
macwoody, line 5120: here we may still have the problem of original name issue.  
And I feel that we don't really need to get schema and object name through Relation. 
They could be obtained just based on the object name and current schema name if needed.

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -5325,3 +5336,169 @@ pltsql_remove_current_query_env(void)
...
+				delete_extended_property(3, db_id, schema_name, major_name,
+										 NULL, NULL);
+			}
+
+			if (schema_name)
+				pfree(schema_name);
+			if (major_name)
+				pfree(major_name);
+		}
+	}
```
macwoody, line 5248: Do we have any other objects to handle ? If not, how about add an error handling case?

contrib/babelfishpg_tsql/src/extendedproperty.h
```diff
@@ -0,0 +1,21 @@
...
+                                     const char *type);
+extern void update_extended_property(int nkeys,
+                                     int16 db_id,
+                                     const char *schema_name,
+                                     const char *major_name,
+                                     const char *minor_name,
+                                     const char *type,
+                                     int attnum,
+                                     const char *new_value);
+extern void babelfish_exec_extendedproperty(PG_FUNCTION_ARGS,
```
macwoody: I am not familiar with this coding pattern. What are the considerations behind it? 
My personal preference would just be putting everything in its own function and extract common functionalities as util functions. Would like to hear your thoughts

contrib/babelfishpg_tsql/src/pl_handler.c
markrui3: In PG, all name is in downcase and we should store downcase in the extended property catalog table.

sp_xxxextendedproperty can pass different case. For example,
```
-- we have schema dbo
sp_addextendedproperty 'name', 'value', 'schema', 'Dbo'
```

contrib/babelfishpg_tsql/src/pl_handler.c
markrui3: I can pass wrong schema name and wrong object name, which will cause useless search on the extended property catalog table.
It doesn't cost much through relation, cause the relation cache will be reused in PG utility function.

contrib/babelfishpg_tsql/src/extendedproperty.h
markrui3: Yes, the common functionalities are so huge that most code are common.

contrib/babelfishpg_tsql/sql/ownership.sql
```diff
@@ -611,4 +611,65 @@ GRANT EXECUTE ON PROCEDURE sys.babelfish_remove_domain_mapping_entry TO PUBLIC;
...
+	END) AS int) AS major_id,
+	CAST((CASE
+		WHEN ep.type = 'DATABASE' THEN 0
+		WHEN ep.type = 'SCHEMA' THEN 0
+		WHEN ep.type IN ('TABLE', 'VIEW', 'SEQUENCE', 'PROCEDURE', 'FUNCTION', 'TYPE') THEN 0
+		WHEN ep.type = 'TABLE COLUMN' THEN a.attnum
+	END) AS int) AS minor_id,
+	ep.orig_name AS name, ep.value AS value
+	FROM sys.babelfish_extended_properties ep
+		LEFT JOIN sys.babelfish_namespace_ext ne ON ne.dbid = sys.db_id() AND ne.orig_name = ep.schema_name COLLATE "C"
```
macwoody, line 659: This series of left outer join is semantically correct. 
It's only usage is to provide filter OID to the filter below. 
I have concern on it's worst case performance. 
Ideally, every row in the extended properties only need one pg_catalog look up based on its type. 
This implementation will join all the catalog tables for every row. 

Could you please load a DB with certain amount of various objects and check if execution plan is expected?

contrib/babelfishpg_tsql/sql/ownership.sql
```diff
@@ -611,4 +611,65 @@ GRANT EXECUTE ON PROCEDURE sys.babelfish_remove_domain_mapping_entry TO PUBLIC;
...
+		LEFT JOIN pg_catalog.pg_namespace n ON n.nspname = ne.nspname
+		LEFT JOIN pg_catalog.pg_class c ON c.relname = ep.major_name AND c.relnamespace = n.oid
+		LEFT JOIN pg_catalog.pg_proc p ON p.proname = ep.major_name AND p.pronamespace = n.oid
+		LEFT JOIN pg_catalog.pg_type t ON t.typname = ep.major_name AND t.typnamespace = n.oid
+		LEFT JOIN pg_catalog.pg_attribute a ON a.attrelid = c.oid AND a.attname = ep.minor_name
+	WHERE ep.dbid = sys.db_id() AND
+	(CASE
+		WHEN ep.type = 'DATABASE' THEN true
```
macwoody, line 667: Just for clarification: Is it true that any user could see current database's extended property?

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--3.2.0--3.3.0.sql
```diff
@@ -45,6 +45,132 @@ RETURNS date
 AS 'babelfishpg_tsql', 'EOMONTH'
 LANGUAGE C STABLE PARALLEL SAFE;
 
+ALTER TABLE sys.extended_properties RENAME TO extended_properties_deprecated_in_3_3_0;
+CREATE TABLE sys.babelfish_extended_properties (
```
macwoody, line 49: This implementation means, anything stored in the old table would be thrown away. 
We've never supported add/drop extended property, so I feel this is OK. 
But it is worth pointing out in your design. And please help double check with Rob

contrib/babelfishpg_tsql/src/catalog.h
```diff
@@ -292,6 +292,42 @@ extern Oid	bbf_domain_mapping_idx_oid;
...
+#define Anum_bbf_extended_properties_value 8
+#define BBF_EXTENDED_PROPERTIES_NUM_COLS 8
+
+extern Oid	bbf_extended_properties_oid;
+extern Oid	bbf_extended_properties_idx_oid;
+
+extern Oid	get_bbf_extended_properties_oid(void);
+extern Oid	get_bbf_extended_properties_idx_oid(void);
+
+typedef struct FormData_bbf_extended_properties
```
macwoody, line 317: This type of structure is needed only when you need to access the tuple in-memory. 
Usually it is associated with catalog cache. Since you are not using cache, it is optional

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -5046,3 +5057,193 @@ pltsql_remove_current_query_env(void)
...
+
+	if (stmt->removeType == OBJECT_SCHEMA)
+	{
+		foreach(cell, stmt->objects)
+		{
+			/* Get schema_name */
+			logical_schema_name = get_logical_schema_name(strVal(lfirst(cell)),
+														  true);
+
+			if (logical_schema_name)
```
macwoody, line 5088: when can it be NULL?

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -5046,3 +5057,193 @@ pltsql_remove_current_query_env(void)
...
+	ListCell		*cell;
+
+	db_id = get_cur_db_id();
+
+	if (stmt->removeType == OBJECT_SCHEMA)
+	{
+		foreach(cell, stmt->objects)
+		{
+			/* Get schema_name */
+			logical_schema_name = get_logical_schema_name(strVal(lfirst(cell)),
```
macwoody, line 5085: I think there are two cases
1. drop schema from "DROP database" statement
2. drop schema from DROP SCHEMA statement

I am not sure if get_logical_schema_name is always correct here.

contrib/babelfishpg_tsql/src/extendedproperty.c
```diff
@@ -0,0 +1,1139 @@
...
+	systable_endscan(scan);
+	table_close(rel, RowExclusiveLock);
+
+	CommandCounterIncrement();
+}
+
+Datum
+sp_addextendedproperty(PG_FUNCTION_ARGS)
+{
+	sp_execextended_property(fcinfo, "sp_addextendedproperty");
```
macwoody: if you would like to code this way, let's use ENUM instead of string.

contrib/babelfishpg_tsql/src/extendedproperty.c
```diff
@@ -0,0 +1,1139 @@
...
+	{
+		CatalogTupleDelete(rel, &tuple->t_self);
+	}
+
+	systable_endscan(scan);
+	table_close(rel, RowExclusiveLock);
+
+	CommandCounterIncrement();
+
+	if (type)
```
macwoody, line 691: Are they copied within this function? 
Usually we rely on the memory context, what is the consideration here?

contrib/babelfishpg_tsql/src/extendedproperty.c
```diff
@@ -0,0 +1,1139 @@
...
+				BTEqualStrategyNumber, F_NAMEEQ,
+				CStringGetDatum(minor_name));
+	ScanKeyInit(&scanKey[5],
+				Anum_bbf_extended_properties_name,
+				BTEqualStrategyNumber, F_TEXTEQ,
+				CStringGetTextDatum(name));
+
+	scan = systable_beginscan(rel, get_bbf_extended_properties_idx_oid(), true,
+							  NULL, 6, scanKey);
+	if (HeapTupleIsValid(tuple = systable_getnext(scan)))
```
macwoody, line 616: I am not familiar with the coding pattern, do we need to check if it is returning only one row ?

macwoody (commented):
This is really a large chunk of changes, I could hardly consume all at once. 
I suggest that we utilize the weekly meeting to break it into small logical chunks and proceed one by one

contrib/babelfishpg_tsql/sql/ownership.sql
markrui3: It is any user who can connect to the database could see current database's extended property.

contrib/babelfishpg_tsql/sql/ownership.sql
markrui3: I know this view has a slow performance. I wonder if I can use sys.object_id(), which will save a lot of left joins.
I remember you mentioned not to rely on Babelfish functions.

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--3.2.0--3.3.0.sql
markrui3: I have concerned about this problem. Maybe you can tell me how to throw an error and fail the upgrade.

image
https://github.com/babelfish-for-postgresql/babelfish-for-postgresql-priv-collaboration/discussions/55

contrib/babelfishpg_tsql/src/pl_handler.c
markrui3: When the schema doesn't exist at all.

contrib/babelfishpg_tsql/src/pl_handler.c
markrui3: In fact, 1 and 2 have same code path to call bbf_ExecDropStmt, so schema names are both physical.

contrib/babelfishpg_tsql/src/extendedproperty.c
markrui3: Sure.

contrib/babelfishpg_tsql/src/extendedproperty.c
markrui3: Refer to commonly realization in procedure.c. I think the code block has been in a memory context called "SPI exec".
If you really need memory context, please tell me directly.

contrib/babelfishpg_tsql/src/extendedproperty.c
markrui3: I only check if the extended property with given parameters has existed.

macwoody:
CREATE SCHEMA 'buYErs';   --- original name 

EXEC sys.sp_addextendedproperty   
@name = N'MS_Description',   
@value = N'Application Role for the Purchasing Department.',  
@level0type = N'SCHEMA',  
@level0name = 'Buyers'; 

list 
buYErs <--- implemented behavior

extended property catalog 
buyers <--- physicalname, original name , logical name in pg format

PR #1540, opened 2023-06-01
Support PARSENAME() T-SQL function

pratikzode:
The PARSENAME() function in T-SQL is used to parse a string representing a four-part SQL Server object name, such as "database.schema.object.column".

SELECT PARSENAME('tempdb.dbo.Employee', 1)
2> go

Employee

1> SELECT PARSENAME('tempdb.dbo.Employee', 2)
2> go

dbo

1> SELECT PARSENAME('tempdb.dbo.Employee', 3)
2> go

tempdb

Signed-off-by: Pratik Zode [pzode@amazon.com](mailto:pzode@amazon.com)


contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -2088,6 +2089,292 @@ numeric_radians(PG_FUNCTION_ARGS)
...
+				 errmsg("invalid UTF8 code point 0x%x", code)));
+
+	if (consumed_p)
+		*consumed_p = consumed;
+
+	return code;
+}
+
+Datum
+parsename(PG_FUNCTION_ARGS)
```
lejaokri, line 2099: It would be easier for reviewers to add a comment describing what the function does and a few examples.

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -2088,6 +2089,292 @@ numeric_radians(PG_FUNCTION_ARGS)
...
+{
+    text *object_name = PG_GETARG_TEXT_PP(0);
+    int object_piece = PG_GETARG_INT32(1);
+    char *object_name_str = text_to_cstring(object_name);
+    int len = strlen(object_name_str);
+    int state = 0;
```
lejaokri: Please use `enum` for state.

test/JDBC/expected/sys-parsename-vu-verify.out
```diff
@@ -0,0 +1,473 @@
...
+
+SELECT PARSENAME('tempdb.dbo.[Empl]]oyee]',1)
+GO
+~~START~~
+varchar
+Empl]oyee
+~~END~~
+
+
+SELECT PARSENAME('tempdb.dbo.[Empl]]oy]]ee]',1)
```
lejaokri, line 345: nitpick: please add a one line comment what is going on here

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -2088,6 +2089,292 @@ numeric_radians(PG_FUNCTION_ARGS)
...
+                    initial_state[current_part] = 1;
+                }
+
+                start_positions[current_part] = &object_name_str[i + 1];
+                i += consumed;
+                continue;
+            }
+            else if(c == ']')
+            {
+                pfree(object_name_str);
```
lejaokri: another nitpick: can remove the pfree call here. You're not doing it everytime before returning null anyway. I also checked this runs under `ExprContext` which is pretty disposable.

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -2088,6 +2089,292 @@ numeric_radians(PG_FUNCTION_ARGS)
 	PG_RETURN_NUMERIC(result);
 }
 
+static inline int32_t
```
xhfanhe: Please add comments to describe the functionality and usage of this function

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -2088,6 +2089,292 @@ numeric_radians(PG_FUNCTION_ARGS)
...
+GetUTF8CodePoint(const unsigned char *in, int len, int *consumed_p)
```
macwoody: It is OK to prototype with it, but we should generally avoid duplicating functions within BBF for maintainablity .
We need to have discussion on this duplicate.

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -2088,6 +2089,292 @@ numeric_radians(PG_FUNCTION_ARGS)
...
+            }
+            else if(c == ']')
+            {
+                pfree(object_name_str);
+                PG_RETURN_NULL();
+            }
+            else if (c == '[')
+            {
+                state = 2;
+                if (initial_state[current_part] == 0)
```
macwoody: how your implementation handle abc[abc] case?

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -2088,6 +2089,292 @@ numeric_radians(PG_FUNCTION_ARGS)
...
+                        PG_RETURN_NULL();
+                    }
+                    i += consumed;
+                    continue;
+                }
+            }
+        }
+
+        // just add the total characters and total length
+        if (state > 0 || (state == 0 && c != '.'))
```
xhfanhe: Please elaborate a bit regarding the UTF handling here

contrib/babelfishpg_tsql/runtime/functions.c
wboettge: Wondering why we need to worry about encodings here. The design doc also mentioned needing to calculate length in bytes under UTF-16. Is this the first time we've needed a function to get a UTF-8 code point?

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -2088,6 +2089,292 @@ numeric_radians(PG_FUNCTION_ARGS)
...
+        }
+        else
+        {
+            PG_RETURN_NULL();
+        }
+    }
+    else
+    {
+        PG_RETURN_NULL();
+    }
```
xhfanhe: Can omit them and just add one line `PG_RETURN_NULL();` at the end of this function

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -2088,6 +2089,292 @@ numeric_radians(PG_FUNCTION_ARGS)
...
+                    end_positions[current_part] = &object_name_str[i - 1];
+                }
+
+                current_part++;
+                if (current_part > 3)
+                {
+                    PG_RETURN_NULL();
+                }
+
+                start_positions[current_part] = &object_name_str[i + 1];
```
macwoody, line 2196: how do we even know if "i+1" is valid?

contrib/babelfishpg_tsql/runtime/functions.c
pratikzode: Yes by using this function we are getting the code and bytes in UTF8 of the character and then calculate how much bytes it takes in UTF16.

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -2088,6 +2089,292 @@ numeric_radians(PG_FUNCTION_ARGS)
...
+                i += consumed;
+                continue;
+            }
+            else if (c == '.')
+            {
+                // do not update the value of end_positions[current_part] if there is already a value in end_postions[current_part] & previous character is " or ].
+                if ( !((end_positions[current_part] != NULL) && (object_name_str[i - 1] == '"')) && !((end_positions[current_part] != NULL) && (object_name_str[i - 1] == ']')) )
```
macwoody, line 2185: I feel this condition is very confusing. Can the program reach line 2248? Under which input?

macwoody (commented):
I haven't check the test one by one. 
Please make sure scenario used in your test matches sql server.

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -2088,6 +2089,292 @@ numeric_radians(PG_FUNCTION_ARGS)
...
+    }
+
+    // Reverse the object piece index
+    object_piece = current_part + 1 - object_piece;
+    if (object_piece < 0 || object_piece > current_part)
+    {
+        PG_RETURN_NULL();
+    }
+    
+    if (object_piece >= 0 && object_piece <= current_part)
```
xhfanhe, line 2281: Already did the check, can omit this

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -2088,6 +2089,292 @@ numeric_radians(PG_FUNCTION_ARGS)
...
+        }
+        i += consumed;
+    }
+
+    if (state != 0)
+    {
+        PG_RETURN_NULL();
+    }
+
+    if (total_chars > 128 || total_length > 256)
```
abhihjoshi, line 2263: How does the parsing handle cases where there is an invalid UTF8 character? i.e. a new line character

contrib/babelfishpg_tsql/runtime/functions.c
pratikzode: SELECT PARSENAME('tempdb.[dbo].Employee',2)
In the above input, if you see line 2296 which has code `end_positions[current_part] = &object_name_str[i - 1];`, if the character is ], we update the end position for that part. Now, there is a dot(.) after we encounter a ']' we need the end_position to be ']' otherwise it will take the end_position as (.) and it will print an output as  dbo]

contrib/babelfishpg_tsql/runtime/functions.c
pratikzode: yes will add that

contrib/babelfishpg_tsql/runtime/functions.c
deepakshi-mittal: Also Is contrib/babelfishpg_tsql/runtime/functions.c the correct file for such a function definition?

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -2088,6 +2089,292 @@ numeric_radians(PG_FUNCTION_ARGS)
...
+					(errcode(ERRCODE_DATA_EXCEPTION),
+					 errmsg("truncated UTF8 byte sequence starting with 0x%02x",
+							in[0])));
+		if ((in[1] & 0xC0) != 0x80 || (in[2] & 0xC0) != 0x80 ||
+			(in[3] & 0xC0) != 0x80)
+			ereport(ERROR,
+					(errcode(ERRCODE_DATA_EXCEPTION),
+					 errmsg("invalid UTF8 byte sequence starting with 0x%02x",
+							in[0])));
+		code = ((in[0] & 0x07) << 18) | ((in[1] & 0x3F) << 12) |
```
forestkeeper: for those 0x07, and 18, a lot of bit operations, I can't figure out the real meaning when reading code
Is it possible to give some meaningful define ?

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -2088,6 +2089,292 @@ numeric_radians(PG_FUNCTION_ARGS)
...
+	}
+	else
+	{
+		ereport(ERROR,
+				(errcode(ERRCODE_DATA_EXCEPTION),
+				 errmsg("invalid UTF8 byte sequence starting with 0x%02x",
+						in[0])));
+	}
+
+	if (code > 0x10FFFF || (code >= 0xD800 && code < 0xE000))
```
forestkeeper: same, need some meaningful define for bit or bytes here

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -2088,6 +2089,292 @@ numeric_radians(PG_FUNCTION_ARGS)
...
+                    if (i + 1 < len && object_name_str[i + 1] != '.')
+                    {
+                        PG_RETURN_NULL();
+                    }
+                    i += consumed;
+                    continue;
+                }
+            }
+        }
+        else if (state == 2)
```
forestkeeper: could you add a meaningful define for state = 1, = 2 ?

contrib/babelfishpg_tsql/runtime/functions.c
pratikzode: Is it necessary to use enum here? I'm using state to keep track of which special character(" [ .) I've encountered initially.

contrib/babelfishpg_tsql/runtime/functions.c
pratikzode: Yes this case was missed. To handle this case, I'm checking in state 0 if the character is (" or [) and if total characters until now is greater than 0, it will return null.

contrib/babelfishpg_tsql/runtime/functions.c
pratikzode: If we take an input as SELECT PARSENAME('[',1) only in this input its storing it in NULL and as there are no more characters in the string and state is 2. After that its returning to for loop when its false and it will go to State!=0 and then returning NULL.

contrib/babelfishpg_tsql/runtime/functions.c
pratikzode: Yes, this check is required here as we are not checking it before.

contrib/babelfishpg_tsql/runtime/functions.c
pratikzode: The invalid UTF8 character will get handled in `GetUTF8CodePoint` function.

contrib/babelfishpg_tsql/runtime/functions.c
pratikzode: This line is decoding a 4-byte UTF-8 sequence by extracting the bits that encode the Unicode code point and combining them into a single integer.

contrib/babelfishpg_tsql/runtime/functions.c
pratikzode: It is checking if the Unicode code point is within the valid range.

contrib/babelfishpg_tsql/runtime/functions.c
pratikzode: This line increments total_chars by 1 if the current character's Unicode code point is less than or equal to 0xFFFF (i.e., it can be represented in UTF-16), and by 2 otherwise.

contrib/babelfishpg_tsql/runtime/functions.c
pratikzode: Do you suggestion or reference that I can look into?

test/JDBC/expected/sys-parsename-vu-verify.out
pratikzode: Added the comment.

contrib/babelfishpg_tsql/runtime/functions.c
pratikzode: Yes makes sense and have removed it.

contrib/babelfishpg_tsql/runtime/functions.c
pratikzode: Added the above comment.

contrib/babelfishpg_tsql/runtime/functions.c
pratikzode: Yes added comment.

contrib/babelfishpg_tsql/runtime/functions.c
pratikzode: Yes, I have made changes in the next revision.

contrib/babelfishpg_tsql/runtime/functions.c
pratikzode: Added Enum in my next revision.

contrib/babelfishpg_tsql/runtime/functions.c
pratikzode: I thought more about this and I think enum make more sense. So thank you for suggesting this and have made the changes in the next revision.

contrib/babelfishpg_tsql/sql/sys_functions.sql
```diff
@@ -551,22 +551,11 @@ RETURNS INTEGER AS
...
-***************************************************************/
-SELECT CASE
-		WHEN char_length($1) < char_length(pg_catalog.replace($1, '.', '')) + 4
-			AND $2 BETWEEN 1
-				AND 4
-			THEN reverse(split_part(reverse($1), '.', $2))
-		ELSE NULL
-		END $$ immutable LANGUAGE 'sql';
-        
+CREATE OR REPLACE FUNCTION sys.parsename(object_name character varying, object_piece int)
```
macwoody: I feel that your input type should be sys.varchar, please double check agains the specification.

pratikzode: Yes, will make the changes.

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -2088,6 +2089,313 @@ numeric_radians(PG_FUNCTION_ARGS)
...
+    } State;
+    State state = STATE_DEFAULT;
+    int consumed;
+    int32_t code;
+    int total_chars = 0;
+    int total_length = 0;
+    char c;
+    char *start_positions[4] = {NULL};
+    char *end_positions[4] = {NULL};
+    int initial_state[4] = {0};
```
macwoody: It should be using State type as well.

contrib/babelfishpg_tsql/runtime/functions.c
wboettge: I agree with the original comment, due to the check on line 2364, is the following if statement not guaranteed to always be true? Since it can't be false, it the check should be removed.

test/JDBC/input/functions/sys-parsename-vu-prepare.sql
```diff
@@ -0,0 +1,51 @@
...
+AS
+SELECT PARSENAME('tempdb.dbo.Employee', 1) AS [Table Name]
+GO
+
+CREATE PROCEDURE parsename_GetEmployeeDatabaseName3
+AS
+BEGIN
+SELECT PARSENAME('tempdb.dbo.Employee_parsename', 1) AS [Table Name]
+END
+GO
```
RIC06X, line 51: Should we also test PARSENAME in functions?

test/JDBC/input/functions/sys-parsename-vu-verify.sql
```diff
@@ -0,0 +1,242 @@
...
+GO
+
+--Checking PARSENAME to the fourth part of a four-part name, 'tempdb'.
+SELECT PARSENAME('tempdb.dbo.Employee.mytable',4)
+GO
+
+--Checking PARSENAME to return the fifth part of a four-part name which does not exist. Will return NULL.
+SELECT PARSENAME('tempdb.dbo.Employee.mytable',5)
+GO
+
```
RIC06X, line 47: Should we also cover the 0-part of the name?

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -2088,6 +2089,313 @@ numeric_radians(PG_FUNCTION_ARGS)
...
+		*consumed_p = consumed;
+
+	return code;
+}
+
+/* The PARSENAME() function in T-SQL is used to parse a string representing a four-part SQL Server object name, such as "database.schema.object.column".
+SELECT PARSENAME('tempdb.dbo.Employee', 1)
+2> go
+Employee
+*/
```
wboettge, line 2097: Nit: This comment should get cleaned up. I'd suggest instead of an example, perhaps explaining what the input arguments represent. An explanation of why '[' and '"' are special could also be helpful here. Also see https://www.postgresql.org/docs/current/source-format.html for comment formatting guidelines

test/JDBC/input/functions/sys-parsename-vu-verify.sql
pratikzode: Yes Thank you for suggesting have added it in the test case.

test/JDBC/input/functions/sys-parsename-vu-prepare.sql
pratikzode: If we give the input as parsename it will give the output as parsename.

contrib/babelfishpg_tsql/runtime/functions.c
pratikzode: Yes make sense have add the comments

contrib/babelfishpg_common/src/varchar.c
```diff
@@ -92,7 +92,7 @@ is_basetype_nchar_nvarchar(Oid typid)
  * GetUTF8CodePoint - extract the next Unicode code point from 1..4
  *					  bytes at 'in' in UTF-8 encoding.
  */
-static inline int32_t
+inline int32_t
```
macwoody: don't need inline here .

contrib/babelfishpg_tsql/runtime/functions.c
pratikzode: Ack. made the changes in the next revision.

contrib/babelfishpg_common/src/varchar.c
pratikzode: Ack. made the changes in the next revision.

macwoody (dismissed):

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--3.1.0--3.2.0.sql
```diff
@@ -1150,6 +1150,15 @@ CREATE AGGREGATE sys.VARP(float8) (
 CREATE OR REPLACE FUNCTION sys.rowcount_big()
 RETURNS BIGINT AS 'babelfishpg_tsql' LANGUAGE C STABLE;
 
+ALTER FUNCTION sys.parsename(VARCHAR,INT) RENAME TO parsename_deprecated_in_3_2_0;
```
macwoody: According to current schedule, you may need to deliver it next cycle, which is 3.3. 
Please move this to next upgrade script

pratikzode: Yes made the changes in the next revision.

macwoody (approved):

macwoody (approved):

PR #1541, opened 2023-06-02
like with bracket work improper for some corner cases

forestkeeper:
previously, like expression for bracket produce wrong result for :
1. [%, when there's no right bracket, % should not work as a wild character
2. the order range of character should not follow the ascii order but should follow the database natural order, for example : '='(ascii:62) should not between 0(ascii 48) to a(ascii 97)

Task: BABEL-4193

### Check List

### DIFF
--- a/contrib/babelfishpg_common/src/collation.c
+++ b/contrib/babelfishpg_common/src/collation.c
@@ -423,7 +423,7 @@ locale_info locales[] =
 	{0x043e, 1252, PG_WIN1252, "ms-MY"}, //Malay:Malaysia
 	/* {0x044e, 	0, 	"mr-IN"}, // Marathi: India */
 	{0x0450, 1251, PG_WIN1251, "mn-MN"}, //Mongolian:Mongolia
-	{0x0414, 1252, PG_WIN1252, "nb-NO"}, //Norwegian:Norway(Bokm� � l)
+	{0x0414, 1252, PG_WIN1252, "nb-NO"}, //Norwegian:Norway(BokmÃ ¥ l)
 	{0x0814, 1252, PG_WIN1252, "nn-NO"}, //Norwegian:Norway(Nynorsk)
 	{0x0415, 1250, PG_WIN1250, "pl-PL"}, //Polish:Poland
 	{0x0416, 1252, PG_WIN1252, "pt-BR"}, //Portuguese:Brazil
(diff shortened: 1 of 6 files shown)


test/JDBC/upgrade/15_3/schedule
```diff
@@ -227,7 +227,6 @@ nested_trigger_inside_proc
 nested_trigger_with_dml
 objectpropertyex
 openjson
-openquery_upgrd
```
forestkeeper: this one failed with diff : 
```
-~~ERROR (Message: server "invalid_server" does not exist)~~
+~~ERROR (Message: The server 'invalid_server' does not exist. Use sp_linkedservers to show available servers.)~~

```
in vu prepare

it's not caused by this commit, uncomment it to pass the test cases

contrib/babelfishpg_common/src/collation.c
```diff
@@ -423,7 +423,7 @@ locale_info locales[] =
 	{0x043e, 1252, PG_WIN1252, "ms-MY"}, //Malay:Malaysia
 	/* {0x044e, 	0, 	"mr-IN"}, // Marathi: India */
 	{0x0450, 1251, PG_WIN1251, "mn-MN"}, //Mongolian:Mongolia
-	{0x0414, 1252, PG_WIN1252, "nb-NO"}, //Norwegian:Norway(Bokm� � l)
+	{0x0414, 1252, PG_WIN1252, "nb-NO"}, //Norwegian:Norway(bokmål)
```
2jungkook, line 426: Is this a necessary change?

contrib/babelfishpg_common/src/collation.c
```diff
@@ -1587,3 +1587,15 @@ babelfish_update_server_collation_name(PG_FUNCTION_ARGS)
 	MemoryContextSwitchTo(oldContext);
 	PG_RETURN_VOID();
 }
+
+Oid like_cid = InvalidOid;
```
2jungkook, line 1591: I'm fine with the current approach, but usually hooks should be order-independent and call prev_hooks inside. An alternative could be to define `tsql_like_cid` in `like_match.c` and use it there.

contrib/babelfishpg_common/src/collation.c
forestkeeper: yeah, previously the comment has some error code

2jungkook (approved):

PR #1545, opened 2023-06-05
sp_execute_postgresql procedure

kritika-0601:
Currently users of allowed extensions will have to exit T-SQL and connect on PG port for installing the extension, altering or dropping the extension. This procedure will enable the users to execute certain PG statements from T-SQL endpoint. 
For now we are supporting it in a restricted manner by adding support for executing create extension, alter extension version and drop extension. 

### Check List

Test Scenarios Covered
Use case based -

Boundary conditions -

Arbitrary inputs -

Negative test cases -

Minor version upgrade tests -

Major version upgrade tests -

Performance tests -

Tooling impact -

Client tests -


contrib/babelfishpg_tsql/sql/sys_procedures.sql
```diff
@@ -202,6 +202,10 @@ CREATE OR REPLACE PROCEDURE sys.sp_droprolemember(IN "@rolename" sys.SYSNAME, IN
 AS 'babelfishpg_tsql', 'sp_droprolemember' LANGUAGE C;
 GRANT EXECUTE on PROCEDURE sys.sp_droprolemember(IN sys.SYSNAME, IN sys.SYSNAME) TO PUBLIC;
 
+CREATE OR REPLACE PROCEDURE sys.sp_execute_postgresql(IN "@extensionStmt" sys.SYSNAME)
+AS 'babelfishpg_tsql', 'sp_execute_postgresql' LANGUAGE C;
+GRANT EXECUTE on PROCEDURE sys.sp_execute_postgresql(IN sys.SYSNAME) TO PUBLIC;
```
HarshLunagariya: Change the datatype to nvarchar.

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -1588,6 +1590,234 @@ create_xp_instance_regread_in_master_dbo_internal(PG_FUNCTION_ARGS)
...
+	
+	PG_TRY();
+	{
+		SetCurrentRoleId(GetSessionUserId(), false);
+		if (!(superuser() || role_is_sa(GetSessionUserId())))
+		{
+			ereport(ERROR,
+					(errcode(ERRCODE_INSUFFICIENT_PRIVILEGE),
+					 errmsg("permission denied to create extension")));
+		}
```
HarshLunagariya: Let's not first set the current role to sessionuser. First check whether current statement is extension statement or not. if yes then check whether current sessionuser is superuser and sysadmin.

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -1588,6 +1590,234 @@ create_xp_instance_regread_in_master_dbo_internal(PG_FUNCTION_ARGS)
...
+sp_execute_postgresql(PG_FUNCTION_ARGS)
+{
+	List	   *parsetree_list;
+	ListCell   *parsetree_item;
+	char	   *extensionStmt;
+	Node	   *stmt;
+	size_t		len;
+	const char *saved_dialect = GetConfigOption("babelfishpg_tsql.sql_dialect", true, true);
+	Oid			current_user_id = GetUserId();
+	const char *saved_buffer = "%s, \"$user\", sys, pg_catalog";
```
HarshLunagariya: Save the existing search_path using GetConfigOption.

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -1588,6 +1590,234 @@ create_xp_instance_regread_in_master_dbo_internal(PG_FUNCTION_ARGS)
...
+			char *allowed_extns[] = {"pg_stat_statements", "tds_fdw"};
+			int allowed_extns_size = sizeof(allowed_extns) / sizeof(allowed_extns[0]);
+
+			/* need to make a wrapper PlannedStmt */
+			wrapper = makeNode(PlannedStmt);
+			wrapper->commandType = CMD_UTILITY;
+			wrapper->canSetTag = false;
+			wrapper->utilityStmt = stmt;
+			wrapper->stmt_location = 0;
+			wrapper->stmt_len = 16;
```
HarshLunagariya: Let's specify the actual statement length.

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -1588,6 +1590,234 @@ create_xp_instance_regread_in_master_dbo_internal(PG_FUNCTION_ARGS)
...
+				case  T_AlterExtensionContentsStmt:
+				{
+					ereport(ERROR,
+					(errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
+					 errmsg("Alter extension to Add/Drop object in extension is not currently supported in Babelfish")));
+					break;
+				}
+				default:
+				elog(ERROR, "unrecognized node type: %d",
+					 (int) nodeTag(parsetree));
```
HarshLunagariya: What if user executes `CREATE TABLE ...` statement? It is a valid statement. So in that case it doesn't make sense to throw unrecognized node type error. We should change it 'not supported' error.

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -1588,6 +1590,234 @@ create_xp_instance_regread_in_master_dbo_internal(PG_FUNCTION_ARGS)
...
+				break;
+			}
+		}
+		SetCurrentRoleId(current_user_id, false);
```
HarshLunagariya: We should be including it in PG_CATCH() and after end of the try. Just how we are resetting the config.

contrib/babelfishpg_tsql/src/session.c
```diff
@@ -118,7 +118,7 @@ check_session_db_access(const char *db_name)
 void
 set_cur_user_db_and_path(const char *db_name)
 {
-	const char *user = get_user_for_database(db_name);
+	const char *user = get_user_for_database(db_name);
```
HarshLunagariya: Unwanted change.

test/JDBC/expected/Test-sp_execute_postgresql.out
```diff
@@ -0,0 +1,173 @@
+-- tsql
+exec sp_execute_postgresql 'create extension if not exists pg_stat_statements';
+go
```
HarshLunagariya: Add negatives tests - no args supplied, arg supplied as NULL, string with 0 length is supplied, some other statements supplied.

Verify search_path, suser_name, dialect is not changes after executing the procedure (with error and w/o error).

HarshLunagariya (changes requested):

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -1588,6 +1590,245 @@ create_xp_instance_regread_in_master_dbo_internal(PG_FUNCTION_ARGS)
...
+sp_execute_postgresql(PG_FUNCTION_ARGS)
+{
+	List	   *parsetree_list;
+	ListCell   *parsetree_item;
+	char	   *extensionStmt;
+	Node	   *stmt;
+	size_t		len;
+	const char *saved_dialect = GetConfigOption("babelfishpg_tsql.sql_dialect", true, true);
+	Oid			current_user_id = GetUserId();
+	const char *saved_buffer = pstrdup(GetConfigOption("search_path", true, true));
```
surendravishnoi: why name it buffer?

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -1588,6 +1590,245 @@ create_xp_instance_regread_in_master_dbo_internal(PG_FUNCTION_ARGS)
...
+	PG_TRY();
+	{	
+		set_config_option("babelfishpg_tsql.sql_dialect", "postgres",
+						GUC_CONTEXT_CONFIG,
+						PGC_S_SESSION, GUC_ACTION_SAVE, true, 0, false);
+		
+		extensionStmt = PG_ARGISNULL(0) ? NULL : TextDatumGetCString(PG_GETARG_TEXT_PP(0));
+
+		if (extensionStmt == NULL)
+				ereport(ERROR, (errcode(ERRCODE_NULL_VALUE_NOT_ALLOWED),
```
surendravishnoi, line 1618: Error message should not have "." at the end

.github/composite-actions/install-extensions/action.yml
```diff
@@ -21,7 +21,7 @@ runs:
         ~/${{inputs.install_dir}}/bin/pg_ctl -D ~/${{inputs.install_dir}}/data/ -l logfile start
         cd ${{inputs.install_dir}}/data
         sudo sed -i "s/#listen_addresses = 'localhost'/listen_addresses = '*'/g" postgresql.conf
-        sudo sed -i "s/#shared_preload_libraries = ''/shared_preload_libraries = 'babelfishpg_tds'/g" postgresql.conf
+        sudo sed -i "s/#shared_preload_libraries = ''/shared_preload_libraries = 'babelfishpg_tds,  pg_stat_statements'/g" postgresql.conf
```
2jungkook, line 24: Please add the corresponding change in `dev-tools.sh` file as well.

contrib/babelfishpg_tsql/sql/sys_procedures.sql
```diff
@@ -202,6 +202,10 @@ CREATE OR REPLACE PROCEDURE sys.sp_droprolemember(IN "@rolename" sys.SYSNAME, IN
 AS 'babelfishpg_tsql', 'sp_droprolemember' LANGUAGE C;
 GRANT EXECUTE on PROCEDURE sys.sp_droprolemember(IN sys.SYSNAME, IN sys.SYSNAME) TO PUBLIC;
 
+CREATE OR REPLACE PROCEDURE sys.sp_execute_postgresql(IN "@extensionStmt" sys.nvarchar)
```
2jungkook: Even though we only support extension stmts at the moment, let's try to be more generic to avoid future changes, e.g., `@postgresStmt` instead of `@extensionStmt`.

contrib/babelfishpg_tsql/src/procedures.c
2jungkook, line 1618: Please use consistent indentations.

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -1588,6 +1590,245 @@ create_xp_instance_regread_in_master_dbo_internal(PG_FUNCTION_ARGS)
...
+								errmsg("Statement cannot be NULL.")));
+
+		/* Remove trailing whitespaces */
+			len = strlen(extensionStmt);
```
2jungkook: indentation

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -1588,6 +1590,245 @@ create_xp_instance_regread_in_master_dbo_internal(PG_FUNCTION_ARGS)
...
+				extensionStmt[--len] = 0;
+
+			len = strlen(extensionStmt);
+
+		/* check if input statement is empty after removing trailing spaces */
+			if (len == 0)
+				ereport(ERROR, (errcode(ERRCODE_NULL_VALUE_NOT_ALLOWED),
+											errmsg("Statement cannot be NULL.")));
+
+		parsetree_list = raw_parser(extensionStmt, RAW_PARSE_DEFAULT);
```
2jungkook: Instead of calling `raw_parser` and `ProcessUtility`, did we consider using `SPI`, e.g., https://www.postgresql.org/docs/current/spi-spi-execute.html?

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -1588,6 +1590,245 @@ create_xp_instance_regread_in_master_dbo_internal(PG_FUNCTION_ARGS)
...
+		/* Remove trailing whitespaces */
+			len = strlen(extensionStmt);
+			while (isspace(extensionStmt[len - 1]))
+				extensionStmt[--len] = 0;
+
+			len = strlen(extensionStmt);
+
+		/* check if input statement is empty after removing trailing spaces */
+			if (len == 0)
+				ereport(ERROR, (errcode(ERRCODE_NULL_VALUE_NOT_ALLOWED),
```
2jungkook: What happens if we don't error out a space-only string here?

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -1588,6 +1590,245 @@ create_xp_instance_regread_in_master_dbo_internal(PG_FUNCTION_ARGS)
...
+		stmt = parsetree_nth_stmt(parsetree_list, 0);
+		rewrite_object_refs(stmt);
+
+		if (list_length(parsetree_list) != 1)
+			ereport(ERROR,
+					(errcode(ERRCODE_SYNTAX_ERROR),
+					errmsg("Expected 1 statement but got %d statements after parsing.", list_length(parsetree_list))));
+
+
+		foreach(parsetree_item, parsetree_list)
```
2jungkook: Let's remove the unnecessary for-loop.

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -1588,6 +1590,245 @@ create_xp_instance_regread_in_master_dbo_internal(PG_FUNCTION_ARGS)
...
+{
+	List	   *parsetree_list;
+	ListCell   *parsetree_item;
+	char	   *extensionStmt;
+	Node	   *stmt;
+	size_t		len;
+	const char *saved_dialect = GetConfigOption("babelfishpg_tsql.sql_dialect", true, true);
+	Oid			current_user_id = GetUserId();
+	const char *saved_buffer = pstrdup(GetConfigOption("search_path", true, true));
+	const char *new_buffer = "public, %s, \"$user\", sys, pg_catalog";
```
2jungkook: Where is the place for binding `%s`?

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -1588,6 +1590,245 @@ create_xp_instance_regread_in_master_dbo_internal(PG_FUNCTION_ARGS)
...
+				default:
+				ereport(ERROR,
+					(errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
+					 errmsg("Only Create/alter/drop extension statements are currently supported in Babelfish.")));
+				break;
+			}
+		}
+		
+	}
+	PG_CATCH();
```
2jungkook: Please consider using `PG_FINALLY`, e.g.,
```
 * For the common case that the error recovery code and the cleanup in the
 * normal code path are identical, the following can be used instead:
 *
 *		PG_TRY();
 *		{
 *			... code that might throw ereport(ERROR) ...
 *		}
 *		PG_FINALLY();
 *		{
 *			... cleanup code ...
 *		}
 *              PG_END_TRY();
 *
 * The cleanup code will be run in either case, and any error will be rethrown
 * afterwards.
```

test/JDBC/expected/Test-sp_execute_postgresql.out
```diff
@@ -0,0 +1,231 @@
...
+-- throw an error if more than one statement is passed
+exec sp_execute_postgresql 'create extension pg_stat_statements with schema sys; create extension pg_stat_statements';
+go
+~~ERROR (Code: 33557097)~~
+
+~~ERROR (Message: Expected 1 statement but got 2 statements after parsing.)~~
+
+
+--alter extension version
+exec sp_execute_postgresql 'alter extension pg_stat_statements update to "1.10"';
```
2jungkook, line 316: To verify this, we need to install a version lower than 1.10 and update it to 1.10.

contrib/babelfishpg_tsql/src/procedures.c
kritika-0601: If we don't error it out it will try to create the parsetree for a space " " and it will cause segmentation fault.

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -1588,6 +1590,229 @@ create_xp_instance_regread_in_master_dbo_internal(PG_FUNCTION_ARGS)
 	PG_RETURN_INT32(0);
 }
 
+Datum
+sp_execute_postgresql(PG_FUNCTION_ARGS)
+{
+	List	   *parsetree_list;
+	char	   *postgresStmt;
+	Node	   *stmt;
+    Node	   *parsetree;
```
surendravishnoi: Indentation is incorrect

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -1588,6 +1590,229 @@ create_xp_instance_regread_in_master_dbo_internal(PG_FUNCTION_ARGS)
...
+	size_t		len;
+    PlannedStmt *wrapper;
+    char *allowed_extns[] = {"pg_stat_statements", "tds_fdw"};
```
surendravishnoi: should be const char *

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -1588,6 +1590,229 @@ create_xp_instance_regread_in_master_dbo_internal(PG_FUNCTION_ARGS)
...
+	const char *saved_buffer = pstrdup(GetConfigOption("search_path", true, true));
+	const char *new_buffer = "public, %s, \"$user\", sys, pg_catalog";
+	
+	PG_TRY();
+	{
+		set_config_option("babelfishpg_tsql.sql_dialect", "postgres",
+						GUC_CONTEXT_CONFIG,
+						PGC_S_SESSION, GUC_ACTION_SAVE, true, 0, false);
+		
+		postgresStmt = PG_ARGISNULL(0) ? NULL : TextDatumGetCString(PG_GETARG_TEXT_PP(0));q
```
surendravishnoi: ;q ? at the end

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -1588,6 +1590,229 @@ create_xp_instance_regread_in_master_dbo_internal(PG_FUNCTION_ARGS)
...
+
+		/* Remove trailing whitespaces */
+		len = strlen(postgresStmt);
+		while (isspace(postgresStmt[len - 1]))
+			postgresStmt[--len] = 0;
+
+		len = strlen(postgresStmt);
```
surendravishnoi: this is unnecessary

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -45,6 +45,7 @@
 #include "tcop/tcopprot.h"
 #include "tcop/utility.h"
 #include "tsearch/ts_locale.h"
+#include "../src/rolecmds.h"
```
surendravishnoi: why this syntax?

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -1588,6 +1590,229 @@ create_xp_instance_regread_in_master_dbo_internal(PG_FUNCTION_ARGS)
...
+		len = strlen(postgresStmt);
+		while (isspace(postgresStmt[len - 1]))
+			postgresStmt[--len] = 0;
+
+		len = strlen(postgresStmt);
+
+		/* check if input statement is empty after removing trailing spaces */
+		if (len == 0)
+			ereport(ERROR, (errcode(ERRCODE_NULL_VALUE_NOT_ALLOWED),
+							errmsg("Statement cannot be NULL.")));
```
surendravishnoi: Please follow https://www.postgresql.org/docs/current/error-style-guide.html for all error messages

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -1588,6 +1590,229 @@ create_xp_instance_regread_in_master_dbo_internal(PG_FUNCTION_ARGS)
...
+						schemaName = defGetString(d_schema);
+					}
+					if (strcmp(defel->defname, "cascade") == 0)
+					{
+						ereport(ERROR, (errcode(ERRCODE_INVALID_PARAMETER_VALUE),
+							errmsg("'cascade' is not yet supported in Babelfish.")));
+					}
+				}
+
+				if(schemaName != NULL && (strcmp(schemaName, "sys") && strcmp(schemaName, "public")))
```
surendravishnoi: I think decision was to allow user to create extension in any schema of their choice. We will just document the recommendation. If there is confusion, please verify with Rob.

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -1588,6 +1590,229 @@ create_xp_instance_regread_in_master_dbo_internal(PG_FUNCTION_ARGS)
...
+				}
+
+				for(int i = 0; i < allowed_extns_size; i++)
+				{
+					if((strcmp(crstmt->extname, allowed_extns[i])))
+					{
+						if(i == allowed_extns_size - 1)
+						{
+							ereport(ERROR, (errcode(ERRCODE_INVALID_PARAMETER_VALUE),
+								errmsg("'%s' is not a valid name.", crstmt->extname)));
```
surendravishnoi: %s extension creation is not supported

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -1588,6 +1590,229 @@ create_xp_instance_regread_in_master_dbo_internal(PG_FUNCTION_ARGS)
...
+					}
+				}
+
+				if(schemaName != NULL && (strcmp(schemaName, "sys") && strcmp(schemaName, "public")))
+				{
+					ereport(ERROR, (errcode(ERRCODE_INVALID_PARAMETER_VALUE),
+							errmsg("'%s' is not a valid schema, only sys and public schema is allowed in Babelfish.", schemaName)));
+				}
+
+				for(int i = 0; i < allowed_extns_size; i++)
```
surendravishnoi: It would be simpler to use a boolean to indicate if there was match or not. After loop, if no match is found, raise error

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -1588,6 +1590,229 @@ create_xp_instance_regread_in_master_dbo_internal(PG_FUNCTION_ARGS)
...
+                            PROCESS_UTILITY_SUBCOMMAND,
+                            NULL,
+                            NULL,
+                            None_Receiver,
+                            NULL);
+
+                /* make sure later steps can see the object created here */
+                    CommandCounterIncrement();
+                    SetCurrentRoleId(current_user_id, false);
+                    SetConfigOption("search_path",
```
surendravishnoi: these are unnecessary as we are taking care of it in finally

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -1588,6 +1590,229 @@ create_xp_instance_regread_in_master_dbo_internal(PG_FUNCTION_ARGS)
...
+                                PROCESS_UTILITY_SUBCOMMAND,
+                                NULL,
+                                NULL,
+                                None_Receiver,
+                                NULL);
+
+                    /* make sure later steps can see the object created here */
+                    CommandCounterIncrement();
+                }
+                SetCurrentRoleId(current_user_id, false);
```
surendravishnoi: not required, finally will take care of it

contrib/babelfishpg_tsql/src/procedures.c
kritika-0601: It is for the function we are using role_is_sa().

contrib/babelfishpg_tsql/src/procedures.c
kritika-0601: Addressed

contrib/babelfishpg_tsql/src/procedures.c
kritika-0601: Yes, we did consider this as one of the approaches but since we are only supporting utility statements for now, the raw_parser approach seems to suit more. https://quip-amazon.com/2t9mAd8vGqTE#temp:C:OWb9a4ca38c96b84c898ac8ff4af

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -1588,6 +1590,220 @@ create_xp_instance_regread_in_master_dbo_internal(PG_FUNCTION_ARGS)
...
+		postgresStmt = PG_ARGISNULL(0) ? NULL : TextDatumGetCString(PG_GETARG_TEXT_PP(0));
+
+		if (postgresStmt == NULL)
+				ereport(ERROR, (errcode(ERRCODE_NULL_VALUE_NOT_ALLOWED),
+								errmsg("statement cannot be NULL")));
+
+		/* Remove trailing whitespaces */
+		len = strlen(postgresStmt);
+		while (isspace(postgresStmt[len - 1]))
+			postgresStmt[--len] = 0;
```
HarshLunagariya, line 1624: What if len is 0? Shouldn't it be crashing?

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -1588,6 +1590,220 @@ create_xp_instance_regread_in_master_dbo_internal(PG_FUNCTION_ARGS)
...
+
+		/* check if input statement is empty after removing trailing spaces */
+		if (len == 0)
+			ereport(ERROR, (errcode(ERRCODE_NULL_VALUE_NOT_ALLOWED),
+							errmsg("statement cannot be NULL")));
+
+		parsetree_list = raw_parser(postgresStmt, RAW_PARSE_DEFAULT);
+		stmt = parsetree_nth_stmt(parsetree_list, 0);
+		rewrite_object_refs(stmt);
```
HarshLunagariya: Please remove this function call. It TSQL specific rewrite. We don't need it here.

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -1588,6 +1590,220 @@ create_xp_instance_regread_in_master_dbo_internal(PG_FUNCTION_ARGS)
...
+							postgresStmt,
+							false,
+							PROCESS_UTILITY_SUBCOMMAND,
+							NULL,
+							NULL,
+							None_Receiver,
+							NULL);
+
+				/* make sure later steps can see the object created here */
+					CommandCounterIncrement();
```
HarshLunagariya: Fix indent.

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -1588,6 +1590,220 @@ create_xp_instance_regread_in_master_dbo_internal(PG_FUNCTION_ARGS)
...
+					{
+						ereport(ERROR, (errcode(ERRCODE_INVALID_PARAMETER_VALUE),
+							errmsg("'cascade' is not yet supported in Babelfish")));
+					}
+				}
+
+				if(schemaName != NULL && !(is_shared_schema(schemaName)))
+				{
+					ereport(ERROR, (errcode(ERRCODE_INVALID_PARAMETER_VALUE),
+							errmsg("'%s' is not a valid schema", schemaName)));
```
HarshLunagariya: Please verify this error message with Rob.

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -1588,6 +1590,220 @@ create_xp_instance_regread_in_master_dbo_internal(PG_FUNCTION_ARGS)
...
+			{
+				DropStmt *drstmt = (DropStmt *) parsetree;
+
+				SetCurrentRoleId(GetSessionUserId(), false);
+
+				if(drstmt->behavior == DROP_CASCADE)
+				{
+					ereport(ERROR, (errcode(ERRCODE_INVALID_PARAMETER_VALUE),
+								errmsg("'cascade' is not yet supported in Babelfish")));
+				}
```
HarshLunagariya: We should be checking that the object is extension or not. There no point setting sessionuser and checking the cascade if it's not drop extension statement. We need to throw unsupported error in case of other drop statements.
Can you please add a testcase where we do drop table ? In that our current implementation won't throw any error.

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -1588,6 +1590,220 @@ create_xp_instance_regread_in_master_dbo_internal(PG_FUNCTION_ARGS)
...
+
+				/* make sure later steps can see the object created here */
+				CommandCounterIncrement();
+				break;
+			} 
+			case T_AlterObjectSchemaStmt:
+			{
+				ereport(ERROR,
+				(errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
+					errmsg("Alter extension schema is not currently supported in Babelfish")));
```
HarshLunagariya: We need to check the object type here. There can be any object possible. Please add a test case where we execute ALTER TABLE , in this case current implementation won't throw any error.

HarshLunagariya (changes requested):

contrib/babelfishpg_tsql/src/procedures.c
kritika-0601: No, before executing this, I've added the relevant check and it will error out before only if length will be 0.

contrib/babelfishpg_tsql/src/procedures.c
surendravishnoi: I mean just use #include "rolecmds.h"

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -1588,6 +1590,229 @@ create_xp_instance_regread_in_master_dbo_internal(PG_FUNCTION_ARGS)
...
+		len = strlen(postgresStmt);
+		while (isspace(postgresStmt[len - 1]))
+			postgresStmt[--len] = 0;
+
+		/* check if input statement is empty after removing trailing spaces */
+		if (len == 0)
+			ereport(ERROR, (errcode(ERRCODE_NULL_VALUE_NOT_ALLOWED),
+							errmsg("statement cannot be NULL")));
+
+		parsetree_list = raw_parser(postgresStmt, RAW_PARSE_DEFAULT);
```
surendravishnoi, line 1631: is it possible to get a NULL list? we should check it anyway to be safe. Also add a test case where create extension has syntax error

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -1588,6 +1590,229 @@ create_xp_instance_regread_in_master_dbo_internal(PG_FUNCTION_ARGS)
...
+		stmt = parsetree_nth_stmt(parsetree_list, 0);
```
surendravishnoi: what is the purpose of this statement?

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -1588,6 +1590,229 @@ create_xp_instance_regread_in_master_dbo_internal(PG_FUNCTION_ARGS)
...
+				CreateExtensionStmt *crstmt = (CreateExtensionStmt *) parsetree;
+				DefElem    *d_schema = NULL;
+				ListCell   *lc;
+				char	   *schemaName = NULL;
+
+				for(int i = 0; i < allowed_extns_size; i++)
+				{
+					if(!(strcmp(crstmt->extname, allowed_extns[i])))
+					{
+						flag = true;
```
surendravishnoi: use a local variable in current scope and name it ext_found or similar, also break after found

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -1588,6 +1590,229 @@ create_xp_instance_regread_in_master_dbo_internal(PG_FUNCTION_ARGS)
...
+	Node	   *stmt;
+	Node	   *parsetree;
+	size_t		len;
+	PlannedStmt *wrapper;
+	bool flag = false;
+	const char *allowed_extns[] = {"pg_stat_statements", "tds_fdw", "fuzzystrmatch"};
+	int allowed_extns_size = sizeof(allowed_extns) / sizeof(allowed_extns[0]);
+	const char *saved_dialect = GetConfigOption("babelfishpg_tsql.sql_dialect", true, true);
+	Oid			current_user_id = GetUserId();
+	const char *saved_path = pstrdup(GetConfigOption("search_path", true, true));
```
surendravishnoi, line 1606: why make a copy for search path?

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -1588,6 +1590,229 @@ create_xp_instance_regread_in_master_dbo_internal(PG_FUNCTION_ARGS)
...
+				{
+					ereport(ERROR, (errcode(ERRCODE_INVALID_PARAMETER_VALUE),
+					 		errmsg("'%s' extension creation is not supported", crstmt->extname)));
+				}
+
+				if (!superuser_arg(GetSessionUserId()) || !role_is_sa(GetSessionUserId()))
+				{
+					ereport(ERROR,
+						(errcode(ERRCODE_INSUFFICIENT_PRIVILEGE),
+	 					errmsg("permission denied to create extension")));
```
surendravishnoi, line 1679: Please add a test case for this, one with super user, one with sa and one user which result in error

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -1588,6 +1590,229 @@ create_xp_instance_regread_in_master_dbo_internal(PG_FUNCTION_ARGS)
...
+				break;
+			} 
+			case T_AlterObjectSchemaStmt:
+			{
+				AlterObjectSchemaStmt *alstmt = (AlterObjectSchemaStmt *) parsetree;
+				if (alstmt->objectType != OBJECT_EXTENSION)
+				{
+					ereport(ERROR,
+					(errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
+						errmsg("only create/alter/drop extension statements are currently supported in Babelfish")));
```
surendravishnoi: do we have test case for this?

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -1588,6 +1590,229 @@ create_xp_instance_regread_in_master_dbo_internal(PG_FUNCTION_ARGS)
...
+			{
+				ereport(ERROR,
+				(errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
+					errmsg("alter extension to Add/Drop object in extension is not currently supported in Babelfish")));
+				break;
+			}
+			default:
+				ereport(ERROR,
+					(errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
+						errmsg("only create/alter/drop extension statements are currently supported in Babelfish")));
```
surendravishnoi, line 1788: do we have test case for this error due to some other command?

contrib/babelfishpg_tsql/src/procedures.c
kritika-0601: ALTER SCHEMA is currently not supported in Babelfish, so I was not able to add a test case for that.

contrib/babelfishpg_tsql/src/procedures.c
kritika-0601: Yes, I have added a check after this for the list_length to be 1.

surendravishnoi: I think error is still confusing, when we have invalid stmt vs more or less than 1 stmt. So, if user gets error that expected 1 statement but got 0, it is confusing when they have given 1 stmt only (with invalid syntax)

contrib/babelfishpg_tsql/src/procedures.c
surendravishnoi: if you mean that code can never come here, we should remove this let it go to default error

kritika-0601: I added it thinking, if we support ALTER SCHEMA in future. Yeah ok I will remove it.

contrib/babelfishpg_tsql/src/procedures.c
kritika-0601: If there will be an invalid syntax say `'create extens pg_stat_statements'`, it will throw syntax error.

surendravishnoi (dismissed):

HarshLunagariya (approved):

surendravishnoi (approved):

HarshLunagariya (approved):

PR #1556, opened 2023-06-14
Support EOMONTH Transact-SQL function

pratikzode:
The EOMONTH function is a Transact-SQL function in SQL Server that returns the last day of the month of a specified date, with an optional offset.

1> SELECT EOMONTH ('2023-01-20')
2> go

     2023-01-31


contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -2423,3 +2426,59 @@ pg_extension_config_remove(PG_FUNCTION_ARGS)
...
+    int year, month, day;
+    DateADT date = PG_GETARG_DATEADT(0);
+    int offset = PG_GETARG_INT32(1);
+
+    /* Convert the date to year, month, day */
+    j2date(date + POSTGRES_EPOCH_JDATE, &year, &month, &day);
+
+    /* Adjust the month based on the offset */
+    month += offset;
+    // Check if the new month is greater than 0.
```
macwoody: please don't mix different types of comments

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -2423,3 +2426,59 @@ pg_extension_config_remove(PG_FUNCTION_ARGS)
...
+        month = month % 12 + 12;
+    }
+    /* Now move to the first day of the next month */
+    month++;
+    /* If the new year is less than 1 or greater than 9999, report an error that an overflow occurred. */
+    if (year < 1 || year > 9999) 
+    {
+        ereport(ERROR,
+                (errcode(ERRCODE_DATETIME_FIELD_OVERFLOW),
+                 errmsg("Adding a value to a 'date' column caused an overflow.")));
```
xhfanhe, line 2517: Is this the same error msg as SQL server would throw?

contrib/babelfishpg_tsql/runtime/functions.c
xhfanhe: Please modify the comment style here

test/JDBC/expected/sys-eomonth-vu-verify.out
```diff
@@ -0,0 +1,322 @@
...
+EXEC GetEndOfNextMonthDate_EOMONTH 1;
+GO
+~~START~~
+date
+1996-02-29
+~~END~~
+
+
+--Edge case testing:
+--when offset is positive edge cases where yyyy can go upto 9999 after that it will throw an error.
```
xhfanhe, line 38: Is it supposed to throw an error?

test/JDBC/expected/sys-eomonth-vu-verify.out
```diff
@@ -0,0 +1,322 @@
...
+
+--when offset is positive edge cases where yyyy can go upto 9999 after that it will throw an error.
+SELECT EOMONTH ('2023-05-10',95720)
+GO
+~~ERROR (Code: 33557097)~~
+
+~~ERROR (Message: Adding a value to a 'date' column caused an overflow.)~~
+
+
+--when offset is negative edge cases where yyyy can go upto 0000 after that it will throw an error.
```
xhfanhe, line 55: Same as above, it's not throwing error here

test/JDBC/expected/sys-eomonth-vu-verify.out
```diff
@@ -0,0 +1,322 @@
...
+~~START~~
+date
+1996-01-31
+~~END~~
+
+
+
+--Checking when separator is ‘/’ 
+--Checking valid date format for ‘YYYY/MM/DD’.
+SELECT EOMONTH ('1996/01/20',30)
```
macwoody: All of these tests that tries different formats , doesn't add value to your implementation. 
As explained, they belongs to Date data type. 
Please redesign your tests to cover your own code.

contrib/babelfishpg_tsql/runtime/functions.c
pratikzode: Yes

contrib/babelfishpg_tsql/sql/sys_functions.sql
```diff
@@ -3609,3 +3609,8 @@ IN query text)
 RETURNS SETOF RECORD
 AS 'babelfishpg_tsql', 'openquery_internal'
 LANGUAGE C VOLATILE;
+
+CREATE OR REPLACE FUNCTION sys.EOMONTH(date,int DEFAULT 0)
+RETURNS date
+AS 'babelfishpg_tsql', 'EOMONTH'
+LANGUAGE C IMMUTABLE STRICT;
```
macwoody: I think this function is also parallel safe and stable .

contrib/babelfishpg_tsql/runtime/functions.c
pratikzode: Yes, thanks for pointing that. Will do that in the next revision.

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -2423,3 +2426,59 @@ pg_extension_config_remove(PG_FUNCTION_ARGS)
...
+    /* Adjust the month based on the offset */
+    month += offset;
+    // Check if the new month is greater than 0.
+    if(month > 0)
+    {
+        /* 
+         * If it is, calculate the new year by adding the number of full years in the offset to the current year.
+         * The number of full years in the offset is (month - 1) / 12. 
+         */
+        year += (month - 1) / 12;
```
macwoody, line 2482: Could you please help explain these for calculations? 
if (month > 0 )
{
year += (month - 1) / 12;
 month = (month - 1) % 12 + 1; 
 }
 else 
{
 year += month / 12 - 1;
month = month % 12 + 12;
}

test/JDBC/expected/sys-eomonth-vu-verify.out
```diff
@@ -0,0 +1,322 @@
...
+SELECT EOMONTH ('960120')
+GO
+~~START~~
+date
+1996-01-31
+~~END~~
+
+
+--Checking for NULL it should return NULL
+SELECT EOMONTH (NULL)
```
macwoody: You may also explicitly test all combinations of NULL for your 2 arguments

test/JDBC/expected/sys-eomonth-vu-verify.out
```diff
@@ -0,0 +1,322 @@
...
+2#!#2000-06-15#!#2000-07-31
+3#!#2022-04-30#!#2022-05-31
+~~END~~
+
+
+EXEC GetEndOfNextMonthDate_EOMONTH 1;
+GO
+~~START~~
+date
+1996-02-29
```
wboettge, line 33: I recommend adding a test for February in a leap year if that's not already included.

contrib/babelfishpg_tsql/runtime/functions.c
pratikzode: yes, will make the change in next revision

contrib/babelfishpg_tsql/runtime/functions.c
wboettge: I feel several of the comments here are just restating the code. They could potentially be more helpful if they instead explained why we have the if/else here (i.e. different case with positive / negative offset). And why we need to adjust the month by one only in the positive case.

test/JDBC/expected/sys-eomonth-vu-verify.out
pratikzode: yes, thanks will add that case in next revision

contrib/babelfishpg_tsql/sql/sys_functions.sql
pratikzode: Yes, will make the change in the next revision.

contrib/babelfishpg_tsql/runtime/functions.c
pratikzode: In the expression (month - 1) / 12, the month - 1 operation is performed to account for the fact that in calendar, months are 1-indexed, meaning they start at 1 (January) and end at 12 (December).

When we add an offset to the current month, and that value is greater than 12, it means we crossed a new year. The (month - 1) / 12 operation is basically asking, "how many full cycles of 12 months are there in our offset?". So we have to think in terms of zero-based indices, hence the month - 1.

Example: If the current month is January (1) and we add an offset of 11, the result is 12 (December). In terms of full years passed, the answer is zero, even though the number is 12, because we’re still within the first year. By subtracting 1 before the division, we correctly calculate that no full year has passed.

The same logic applies to (month - 1) % 12 + 1. We subtract 1 before taking the modulus so that we're working with a zero (where January is 0 and December is 11), perform the modulus operation to find the remainder months that didn't form a full year, and then add 1 back to convert back to a 1.

test/JDBC/expected/sys-eomonth-vu-verify.out
pratikzode: Yes, will add them in next revision.

contrib/babelfishpg_tsql/sql/sys_functions.sql
pratikzode: Will make the change

contrib/babelfishpg_tsql/runtime/functions.c
pratikzode: In the expression (month - 1) / 12, the month - 1 operation is performed to account for the fact that in calendar, months are 1-indexed, meaning they start at 1 (January) and end at 12 (December).

When we add an offset to the current month, and that value is greater than 12, it means we crossed a new year. The (month - 1) / 12 operation is basically asking, "how many full cycles of 12 months are there in our offset?". So we have to think in terms of zero-based indices, hence the month - 1.

Example: If the current month is January (1) and we add an offset of 11, the result is 12 (December). In terms of full years passed, the answer is zero, even though the number is 12, because we’re still within the first year. By subtracting 1 before the division, we correctly calculate that no full year has passed.

The same logic applies to (month - 1) % 12 + 1. We subtract 1 before taking the modulus so that we're working with a zero (where January is 0 and December is 11), perform the modulus operation to find the remainder months that didn't form a full year, and then add 1 back to convert back to a 1.

test/JDBC/expected/sys-eomonth-vu-verify.out
pratikzode: Yes, will all the test for different month which gives different last date.

contrib/babelfishpg_tsql/runtime/functions.c
pratikzode: Yes,I have added comments in this new revision and have explain in that.

test/JDBC/expected/sys-eomonth-vu-verify.out
pratikzode: Yes I have added the 2 tests for the month of feb 
--Check if its leap year. If it is a leap year and the month is feb it will return 29 otherwise 28.

test/JDBC/expected/sys-eomonth-vu-verify.out
pratikzode: I have added the tests for all the combinations of NULL for your 2 arguments.

test/JDBC/expected/sys-eomonth-vu-verify.out
pratikzode: Have changed it to `LANGUAGE C IMMUTABLE STRICT PARALLEL SAFE;
`

contrib/babelfishpg_tsql/sql/sys_functions.sql
pratikzode: Have changed it to `LANGUAGE C IMMUTABLE STRICT PARALLEL SAFE;
`

test/JDBC/expected/sys-eomonth-vu-verify.out
pratikzode: I have added more test and removed most of the test that belongs to Date data type.
The test which I added in prepare were different month returning different last date.

test/JDBC/expected/sys-eomonth-vu-verify.out
pratikzode: I have added all possible combinations for NULL except 1 that is mentioned below. Because in SQL server It returns me the value for that months last date. In BBF the NULL is not getting casted to integer and so it's returning me NULL.
```
1> SELECT EOMONTH('1990-01-01',NULL)
2> go
eomonth         
------------
         NULL
```

contrib/babelfishpg_tsql/sql/sys_functions.sql
pratikzode: Have Changed from `IMMUTABLE` to `STABLE`.

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -2423,3 +2426,98 @@ pg_extension_config_remove(PG_FUNCTION_ARGS)
...
+        month = month % 12 + 12;
+    }
+
+    /* Now move to the first day of the next month */
+    month++;
+
+    /* If the new year is less than 1 or greater than 9999, report an error that an overflow occurred. */
+    if (year < 1 || year > 9999) 
+    {
+        if (year < -4173 || year > 5874897)
```
macwoody: what does this condition mean? Please double check

pratikzode: yes, have made the changes in the next revision

macwoody (approved):

PR #1562, opened 2023-06-16
Removing duplicate entry of timestamp and correcting is_nullable column entry

hash-16:
### Description

Currently there are duplicate entry for timestamp which while querying for typeproperty function throw issues  and also sql server has a single entry of timestamp.
Ans also is_nullable columns for timestamp ans sysname should be zero in accordance with that of sql server (required for typeproperty function).

1. Added a filter to sys.babelfish_typecode_list() to get single entry of timestamp
2. Throw 0 in case of  timestamp and sysname datatypes.

### Performance Testing

With Changes

`select  tbl.name AS [Name] , clmns.name , usrt.name as tname from sys.tables AS tbl inner join sys.all_columns as clmns on clmns.object_id = tbl.object_id inner join sys.types as usrt on usrt.user_type_id = clmns.user_type_id`
(24492 rows affected)
8637 ms
 

Without changes

`select  tbl.name AS [Name] , clmns.name , usrt.name as tname from sys.tables AS tbl inner join sys.all_columns as clmns on clmns.object_id = tbl.object_id inner join sys.types as usrt on usrt.user_type_id = clmns.user_type_id`
(24492 rows affected)
8589 ms

### Issues Resolved

* **Use case based -**

* **Boundary conditions -**

* **Arbitrary inputs -**

* **Negative test cases -**

* **Performance tests -**

* **Tooling impact -**

* **Client tests -**

### Check List


test/JDBC/input/views/sys-types-before.sql
```diff
@@ -0,0 +1,88 @@
+select cast(name as varchar(20))
```
shalinilohia50: Why do we need this test file if we already have the tests in `sys_types_before_vu_prepare/verify/cleanup` format?

hash-16: I actually did not add any test files just renamed it . They were present there actually.

shalinilohia50: Why did you append `before` to its original name? 
Shouldn't it be `sys-types` only?

hash-16: Few test suite in github actions required original sys-types file and few required updated one , so I did this to ensure both.

shalinilohia50: 1. Why do we even need `sys-type-before` file if we have the same `sys_types_before_vu_prepare/verify/cleanup` format? These are just extra tests running. 
2. I don't see any jdbc schedule where you have added these tests `sys_types_before_vu_prepare/verify/cleanup`, for upgrade scenarios. Are these even running in any upgrade workflow?

contrib/babelfishpg_tsql/sql/sys_views.sql
```diff
@@ -1077,7 +1077,11 @@ select
     WHEN 'default' THEN default_collation_name
     ELSE  c.collname
     END as collation_name
-  , case when typnotnull then 0 else 1 end as is_nullable
+  , CASE ti.tsql_type_name
+    WHEN 'timestamp' THEN 0
```
shalinilohia50: Indentation

test/JDBC/input/views/sys-types-before.sql
hash-16: 1. Actually we were initially having sys-types and sys-types-vu-prepare both  so I just need renamed it .
2. Yeah we are running updated files now , but I thought to keep older files there , if needed further.
Should  I remove these files ?

contrib/babelfishpg_tsql/sql/sys_views.sql
```diff
@@ -1107,9 +1111,11 @@ select cast(t.typname as text) as name
     ELSE  c.collname 
     END as collation_name
   , case when tt.typrelid is not null then 0
-         else case when typnotnull then 0 else 1 end
-    end
-    as is_nullable
+        else CASE t.typbasetype
```
shalinilohia50: Why are we not doing the same thing as above here?

```
CASE ti.tsql_type_name
      WHEN 'timestamp' THEN 0
      WHEN 'sysname' THEN 0
    ELSE case when typnotnull then 0 else 1 end
    end as is_nullable

Comment (hash-16) [reply contrib/babelfishpg_tsql/sql/sys_views.sql]: We cannot use ti.tsql_type_name to compare here as this is taken from sys.babelfish_typecode_list() which contains only system data types names which will be  different from user defined data type names  , so I used the system_type_id to identify which system data type it is created from and if that system_type_id matches that of  timestamp and sysname system_type_id then it should return 0.

**Review by shalinilohia50 — APPROVED** [review]

**Comment by KushaalShroff** [comment]
Your description does not mention why you require this change and how its going to resolve it

**Comment by KushaalShroff** [comment]
> Changed is_nullable column value from 1 to 0 for timestamp and sysname datatypes.

There are is_nullable columns in other views as well, how are we making sure that we are fixing them as well?

File: contrib/babelfishpg_tsql/sql/sys_views.sql [hunk 4]
```diff
@@ -1057,7 +1057,7 @@ create or replace view sys.types As
 with RECURSIVE type_code_list as
 (
     select distinct  pg_typname as pg_type_name, tsql_typname as tsql_type_name
-    from sys.babelfish_typecode_list()
+    from sys.babelfish_typecode_list() where pg_typname <> 'rowversion'
```
KushaalShroff: Please explain why we have made this change here?
Seems like you are fixing duplicate entry for rowversion datatype since we already have timestamp in babelfish_typecode_list. We should make this change in babelfish_typecode_list itself then, right?

contrib/babelfishpg_tsql/sql/sys_views.sql
```diff
@@ -1077,7 +1077,11 @@ select
     WHEN 'default' THEN default_collation_name
     ELSE  c.collname
     END as collation_name
-  , case when typnotnull then 0 else 1 end as is_nullable
+  , CASE ti.tsql_type_name
+      WHEN 'timestamp' THEN 0
+      WHEN 'sysname' THEN 0
+    ELSE case when typnotnull then 0 else 1 end
```
KushaalShroff: To summarise So we were getting typnotnull as 0 for these types and now we want to only display these datatypes as not nullable in this view. I think this change doesnt make sense to me since we our displaying a behaviour of the datatype which is different to the created type.
We need to make change in the domain itself and not here.

KushaalShroff: We are falsely displaying information of a datatype for a behaviour which is not supported

KushaalShroff:
In all, if you take into consideration all of my comments - we wont require change in sys.types

contrib/babelfishpg_tsql/sql/sys_views.sql
hash-16: I made this change to get single timestamp (not that with rowversion pg_typname). I think this change here is ok as only here we don't require rowversion datatype , but it maybe required at other places.

contrib/babelfishpg_tsql/sql/sys_views.sql
hash-16: I tried making change in domain like:-

- CREATE DOMAIN sys.SYSNAME AS sys.VARCHAR(128) NOT NULL;
- CREATE DOMAIN sys.TIMESTAMP AS sys.ROWVERSION NOT NULL;

But this will cause error everywhere where we have assigned null to sysname/timestamp datatype . Eg:-
For sys.systypes view for this line :-
`, CAST(collation_name as sys.sysname) as collation`

we are assigning NULL here and it will throw error as :- `domain sysname does not allow null values`

contrib/babelfishpg_tsql/sql/sys_views.sql
KushaalShroff: I am not convinced at all by that reply, please elaborate on "other places"

contrib/babelfishpg_tsql/sql/sys_views.sql
KushaalShroff: I'm not convinced either, repeating this again to point out the gravity of your change - WE ARE DISPLAYING a feature of the datatype we do not support

hash-16: Ok, then we need to think on this.

contrib/babelfishpg_tsql/sql/sys_views.sql
hash-16: Ok , I got this point here that we have two entry like 

```
pg_typname.    tsql_typname
rowversion         timestamp
timestamp          timestamp
```

Need to remove the second entry.

contrib/babelfishpg_common/sql/rowversion.sql
```diff
@@ -172,7 +172,7 @@ LANGUAGE C IMMUTABLE STRICT PARALLEL SAFE;
 CREATE CAST (sys.ROWVERSION as INT8)
 WITH FUNCTION sys.rowversionint8 (sys.ROWVERSION) AS ASSIGNMENT;
 
-CREATE DOMAIN sys.TIMESTAMP AS sys.ROWVERSION;
```
KushaalShroff, line 175: Why havent we added NOT NULL to sys.ROWVERSION itself?

AFAIK even rowversion should be not null, how are we ensuring that is fixed?

contrib/babelfishpg_common/src/typecode.c
```diff
@@ -200,6 +200,9 @@ typecode_list(PG_FUNCTION_ARGS)
 		Datum		values[7];
 		bool		nulls[7];
 
+		if(strcmp(info->pg_typname,"rowversion") == 0)
+		continue;
```
KushaalShroff, line 204: Please explain why you chose to remove duplicate entry here rather than removing it from type_infos itself?

contrib/babelfishpg_common/sql/rowversion.sql
```diff
@@ -172,7 +172,7 @@ LANGUAGE C IMMUTABLE STRICT PARALLEL SAFE;
...
+CREATE DOMAIN sys.TIMESTAMP AS sys.ROWVERSION NOT NULL;
```
KushaalShroff, line 175: No upgrade test for testing this change?

contrib/babelfishpg_common/sql/rowversion.sql
hash-16: `A nonnullable rowversion column is semantically equivalent to a binary(8) column. A nullable rowversion column is semantically equivalent to a varbinary(8) column.` 
Got this from sql server documentation .
From this we can see that rowversion can be null.

contrib/babelfishpg_common/src/typecode.c
hash-16: I actually did that change ( removing rowversion datatype) in type_infos array , but that was causing failure in bcp query in dotnet test framework.

contrib/babelfishpg_common/sql/upgrades/babelfishpg_common--3.2.0--3.3.0.sql
```diff
@@ -0,0 +1,9 @@
+-- complain if script is sourced in psql, rather than via ALTER EXTENSION
+\echo Use "ALTER EXTENSION ""babelfishpg_common"" UPDATE TO '3.3.0'" to load this file. \quit
+
+SELECT set_config('search_path', 'sys, '||current_setting('search_path'), false);
+
+ALTER DOMAIN sys.TIMESTAMP SET NOT NULL;
```
surendravishnoi, line 6: Please make sure to test it with both minor and major upgrade. Specifically, if a view is using this domain, what happens as part of minor version upgrade.

PR #1565, opened 2023-06-16
Fix the bug: like expression will error out unexpectly

forestkeeper:
Previously, like expression unexpectly error out in certain cases:
1. Like expression inside Join on condition
2. Like expression inside one of the multiple case whens
3. Like expression as one of the sublink condition in sublink query

This fix has solved those three issues by adding corresponding Like collation change transform into Like expression node.

Task: BABEL-4046

### Check List

### DIFF
--- a/test/JDBC/expected/BABEL-4046.out
+++ b/test/JDBC/expected/BABEL-4046.out
@@ -1,0 +1,197 @@
+create table options(name varchar(30) collate bbf_unicode_cp1_ci_as)
+insert options values('abc')
+go
+~~ROW COUNT: 1~~
+
+
+create table svc_defs(svc_name varchar(30) collate bbf_unicode_cp1_ci_as)
+insert svc_defs values('def')
+go
+~~ROW COUNT: 1~~
+
+
+select options.name
+  from options
+ inner join svc_defs
+    on options.name like 'UM\_%' + svc_defs.svc_name escape '\'
+go
+~~START~~
+varchar
+~~END~~
+
+
+create table t (a varchar(30));
+go
+
+insert into t values ('aaa');
+go
+~~ROW COUNT: 1~~
+
+
+insert into t values ('AAA');
+go
+~~ROW COUNT: 1~~
+
+
+create table t1 (a varchar(30));
+go
+
+insert into t1 values ('a%');
+go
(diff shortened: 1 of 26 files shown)


test/JDBC/expected/BABEL-4046.out
```diff
@@ -0,0 +1,197 @@
+create table options(name varchar(30) collate bbf_unicode_cp1_ci_as)
```
2jungkook: Could you use T-SQL collation instead of BBF's? I'd like to make test files to be executed in SQL Servers as much as possible.

test/JDBC/expected/BABEL-4046.out
```diff
@@ -0,0 +1,197 @@
...
+go
+~~ROW COUNT: 1~~
+
+
+insert into t1 values ('A%');
+go
+~~ROW COUNT: 1~~
+
+
+select * from t join t1 on t.a like t1.a;
```
2jungkook: Could you make this as an upgrade test with views in it?

test/JDBC/expected/BABEL-4046.out
```diff
@@ -0,0 +1,197 @@
...
+int#!#nvarchar#!#text
+7369#!#SMITH#!#BOSTON
+7499#!#ALLEN#!#CHICAGO
+7521#!#WARD#!#CHICAGO
+7566#!#JONES#!#A
+7654#!#MARTIN#!#A
+7698#!#BLAKE#!#BOSTON
+7782#!#CLARK#!#NEW YORK
+7788#!#SCOTT#!#NEW YORK
+7839#!#KING#!#A
```
2jungkook: It could be better to have `order by` clause to prevent future flakeness.

test/JDBC/expected/BABEL-4046.out
forestkeeper: I can delete collate bbf_unicode_cp1_ci_as in this case, since it's the default collation

test/JDBC/expected/BABEL-4046.out
forestkeeper: OK

2jungkook (approved):

PR #1568, opened 2023-06-20
Strip parens from column name in pre_transform_target_entry hook

### LINKED ISSUE #1567: [Bug]: Incorrect column name when column is enclosed in parens
### What happened?

When column is enclosed in parens (round brackets) and no alias is specified, then the column name in result set is returned incorrectly:

```
create table tab1(val1 int);
insert into tab1 values(42);
select (val1) from tab1;
+------+
| (val |
+------+
| 42   |
+------+
```

In real code bases this can happen with this use of `distinct` (that [is misleading](https://stackoverflow.com/a/31679169), but still not infrequent):

```
select distinct(val1) from tab1;
+------+
| (val |
+------+
| 42   |
+------+
```

I have a fix that seems to work, will create a PR shortly.

### Version

BABEL_3_X_DEV (Default)

### Extension

babelfishpg_tsql (Default)

Fedora

### Relevant log output

_No response_

### Code of Conduct

staticlibs:
### Description

It appeared that [pre_transform_target_entry](https://github.com/babelfish-for-postgresql/babelfish_extensions/blob/1b59741c437af4c6a8afe4647d9a30a9df9659c2/contrib/babelfishpg_tsql/src/hooks.c#L1384) can handle double quotes and square bracket quotes, but cannot handle parens (round brackets) that also can be present there.

The patch attempts to strip leading parens, handling multiple levels of parens and spaces in-between them.

### Issues Resolved

#1567 

* **Use case based -**

The test is included with the patch, but it is ignored by default because it requires `outputColumnName` to be enabled in `config.txt`.

* **Boundary conditions -**

I am a bit concerned about the possible buffer overrun because of unbounded `while`, perhaps additional check should be added there.

* **Arbitrary inputs -**

* **Negative test cases -**

* **Performance tests -**

* **Tooling impact -**

* **Client tests -**

### Check List


staticlibs:
Added a check for possible latest `column_start` position inside `while`, added a cast in `isspace`, added initializers to bool vars. Rebased the branch due to accidentally pushing wrong commit.

test/JDBC/jdbc_schedule
```diff
@@ -16,6 +16,9 @@ all
 ignore#!#insertbulk
 ignore#!#BABEL-SQLvariant
 
+# These tests check output colunm names and require 'outputColumnName' config parameter to be set to 'true'.
+ignore#!#select-clause-parens
```
2jungkook: It means there will be no automated test for this change. Could you try another way of testing this fix without changing `outputColumnName`? What about using subqueries, e.g.,`select field1 from (select ((field1)) from select_clause_parens1) a;`?

test/JDBC/expected/select-clause-parens.out
```diff
@@ -0,0 +1,71 @@
+create table select_clause_parens1(field1 int);
```
2jungkook: Please write a version-upgrade test (Ref: https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/1412/files), i.e., create view/proc in `prepare` and execute those in `verify`.

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -1429,12 +1429,24 @@ pre_transform_target_entry(ResTarget *res, ParseState *pstate,
...
+
+			while (colname_start < colname_limit &&
+				(*colname_start == '(' || isspace((unsigned char) *colname_start)))
+			{
+				colname_start++;
+			}
+
+			dq = *colname_start == '"';
+			sqb = *colname_start == '[';
+			sq = *colname_start == '\'';
```
2jungkook: We need more tests combining with the existing logic, e.g., `select ( "field1   "   ) from select_clause_parens1;`

staticlibs:
Thanks for the comments, and apologies for the delay - I will be able to implement the required changes in about 2 weeks.

test/JDBC/jdbc_schedule
staticlibs: Rewrote tests using subqueries and removed the ignore record from jdbc_schedule.

test/JDBC/expected/select-clause-parens.out
staticlibs: Added version upgrade tests, `select-strip-parens-vu-*` for latest and `select-strip-parens-before-15_4-vu-*` for previous versions. The difference on upgrade is that queries with parentheses work correctly for tables created on earlier versions, but if a view was created with parentheses in field names - these parentheses will remain there after the upgrade.

contrib/babelfishpg_tsql/src/hooks.c
staticlibs: Added tests combining parentheses with brackets and double quotes with spaces. Please note, that the logic for spaces for parentheses and for brackets/double quotes is different. For parentheses all spaces are stripped. For brackets and double quotes spaces are preserved in the field name (this logic remains the same as before).

staticlibs:
Added upgrade tests and rebased the branch to the latest version.

2jungkook (dismissed):

2jungkook:
Thanks for revising the PR! I guess there are some trivial merge conflict happened. Other than that, it looks good to me.

staticlibs:
Thanks for the review! I've rebased the branch to the latest one fixing merge conflicts in schedule files.

staticlibs:
Rebased the branch to the latest one taking into account `14_10` and `15_4` upgrade versions.

staticlibs:
About the `JDBC Tests` action failure, that seems to be unrelated to this change (this test passes for me locally from this branch):
```
TOTAL TESTS:	2028
TESTS PASSED:	2027
TESTS FAILED:	1
```
Output diff:
```
--- /home/runner/work/babelfish_extensions/babelfish_extensions/test/JDBC/./expected/BABEL_4367.out	2023-09-06 20:00:19.008893623 +0000
+++ /home/runner/work/babelfish_extensions/babelfish_extensions/test/JDBC/./output/BABEL_4367.out	2023-09-06 20:14:23.549736980 +0000
@@ -236,6 +236,10 @@
 
 CREATE VIEW BABEL_4367_VIEW_2 AS SELECT * FROM BABEL_4367_1 WHERE id <3
 GO
+~~ERROR (Code: 2627)~~
+
+~~ERROR (Message: duplicate key value violates unique constraint "babelfish_view_def_pkey")~~
+
 
 DROP VIEW IF EXISTS BABEL_4367_VIEW_1, BABEL_4367_VIEW_2
 GO
```

staticlibs:
Rebased the branch merging `upgrade/latest/schedule` changes.

2jungkook:
Sorry for the late response and frequent merge conflicts. Since the change looks good to me, I will merge it once you resolve the conflict and all tests are green.
1. We fixed the ODBC test failure today, so it should be green once you rebase it.
2. You can consider adding a test in the middle of schedule files to avoid frequent merge conflicts by others.

staticlibs:
@2jungkook thanks for your comments!

Just the situation with this fix seems to become complicated: [c549c79](https://github.com/babelfish-for-postgresql/babelfish_extensions/commit/c549c794f174b3a95588b02bfd4efdfabda3d82a) was pushed recently, AFAIU it is supposed to fix (among others) the issue with parens. So the actual code change to `contrib/babelfishpg_tsql/src/hooks.c` from this PR is outdated now. Though the `select-strip-parens-*` tests from this PR are still valid and I think it makes sense to push them. But `c549c79` change breaks these tests in non obvious manner, I've added comments with details to #1567 and #1853.

To minimize the confusion I will close this PR for now. Assuming someone is going to follow-up on `c549c79` - I suggest to include `select-strip-parens-*` tests in such follow-up. I will use #1567 for additional comments on this.

staticlibs:
It appeared that using `isspace` in `pre_transform_target_entry` hook instead of direct comparison with a space allows `select-strip-parens` tests to run. Just tests before #1812 and after #1812 (and before `15_5`) must be separate because of different handling of spaces before and after #1812.

Not sure if [isspace](https://en.cppreference.com/w/c/string/byte/isspace) is appropriate here or a direct comparison with all possible (allowed by grammar) whitespace symbols needs to be done instead.

Added tests in the middle of upgrade schedule files and (for 13 and 14) only for versions that actually being run in upgrade tests.

2jungkook:
Thanks a lot for diving deep in this complex issue ;)

Your point makes sense and I'd suggest to use scanner_isspace(). I found this function from https://postgrespro.com/list/thread-id/2653515.

staticlibs:
Thanks for the pointer! I've changed `isspace` to `scanner_isspace`.

2jungkook (approved):

Deepesh125:
Thank you @staticlibs for catching this issue with https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/1812/. I believe your current changes are also required in BABEL_2_X_DEV. Are you planning the same? If not then please let me know. Thanks again.

Deepesh125:
We have cherry-picked these changes in BABEL_2_X_DEV branch, https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/1888

PR #1582, opened 2023-06-28
Do not access function fields after deleting its memory context

### LINKED ISSUE #1581: [Bug]: Intermittent crash in ResetTDSConnection
### What happened?

I am observing an intermittent backend crash that happens when [ResetTDSConnection](https://github.com/babelfish-for-postgresql/babelfish_extensions/blob/aaa7127d18f01188b8945029eb5a27dc2298f378/contrib/babelfishpg_tds/src/backend/tds/tdsprotocol.c#L113) is requested by client application.

`ResetTDSConnection` calls [reset_session_properties](https://github.com/babelfish-for-postgresql/babelfish_extensions/blob/aaa7127d18f01188b8945029eb5a27dc2298f378/contrib/babelfishpg_tds/src/backend/tds/tdsprotocol.c#L144) that eventually calls [reset_cached_batch](https://github.com/babelfish-for-postgresql/babelfish_extensions/blob/aaa7127d18f01188b8945029eb5a27dc2298f378/contrib/babelfishpg_tsql/src/pl_comp.c#L3223). And the crash happens in `pltsql_free_function_memory` on [this assignment](https://github.com/babelfish-for-postgresql/babelfish_extensions/blob/aaa7127d18f01188b8945029eb5a27dc2298f378/contrib/babelfishpg_tsql/src/pl_funcs.c#L868) when the `func` itself is already freed.

I am unable to get a stable reproduction or reproduce it without the client app, the crash happens for me only when the client app runs under load and ODBC connection pooling is enabled (so TDS reset is called on existing connection taken from the pool).

It seems that there can be a simple workaround for the problematic assignment, will submit a PR with it shortly.

### Version

BABEL_3_X_DEV (Default)

### Extension

babelfishpg_tsql (Default)

Other

### Relevant log output

_No response_

### Code of Conduct

staticlibs:
### Description

When an anonymous code block is being executed, the compiled function for it [is allocated](https://github.com/babelfish-for-postgresql/babelfish_extensions/blob/aaa7127d18f01188b8945029eb5a27dc2298f378/contrib/babelfishpg_tsql/src/pl_comp.c#L1058) in the same memory context that is used for the contents of the function (`fn_cxt` field).

When `pltsql_free_function_memory` is called on such function, the memory context is being deleted first and [fn_cxt field is assigned after that](https://github.com/babelfish-for-postgresql/babelfish_extensions/blob/aaa7127d18f01188b8945029eb5a27dc2298f378/contrib/babelfishpg_tsql/src/pl_funcs.c#L868). That is effectively a "use after free" situation, but due to the memory contexts (and I assume libc allocator too) caching, this write almost never leads to a crash.

Proposed patch moves the assignment to `fn_cxt` to happen before the `MemoryContextDelete` call, otherwise the logic is kept the same.

I believe this fixes the root cause for #1581 (at least I cannot reproduce the problem with patch applied).

### Issues Resolved

#1581

* **Use case based -**

Unable to create a simple reproducer.

* **Boundary conditions -**

* **Arbitrary inputs -**

* **Negative test cases -**

* **Performance tests -**

* **Tooling impact -**

* **Client tests -**

### Check List

### DIFF
--- a/contrib/babelfishpg_tsql/src/pl_funcs.c
+++ b/contrib/babelfishpg_tsql/src/pl_funcs.c
@@ -863,6 +863,9 @@ pltsql_free_function_memory(PLtsql_function *func)
 	 * And finally, release all memory except the PLtsql_function struct
 	 * itself (which has to be kept around because there may be multiple
 	 * fn_extra pointers to it).
+	 * It is also possible that PLtsql_function struct was allocated in
+	 * the same func->fn_cxt memory context, so we avoid accessing struct
+	 * fields after memory context is released.
 	 */
 	if (func->fn_cxt)
 	{
--- a/contrib/babelfishpg_tsql/src/pl_funcs.c
+++ b/contrib/babelfishpg_tsql/src/pl_funcs.c
@@ -798,6 +798,7 @@ void
 pltsql_free_function_memory(PLtsql_function *func)
 {
 	int			i;
+	MemoryContext func_cxt = NULL;
 
 	/* Better not call this on an in-use function */
 	Assert(func->use_count == 0);
@@ -864,8 +865,11 @@ pltsql_free_function_memory(PLtsql_function *func)
 	 * fn_extra pointers to it).
 	 */
 	if (func->fn_cxt)
-		MemoryContextDelete(func->fn_cxt);
-	func->fn_cxt = NULL;
+	{
+		func_cxt = func->fn_cxt;
+		func->fn_cxt = NULL;
+		MemoryContextDelete(func_cxt);
+	}
 }
 
 
--- a/contrib/babelfishpg_tsql/src/pl_comp.c
+++ b/contrib/babelfishpg_tsql/src/pl_comp.c
@@ -166,6 +166,7 @@ pltsql_compile(FunctionCallInfo fcinfo, bool forValidator)
 	PLtsql_func_hashkey hashkey;
 	bool		function_valid = false;
 	bool		hashkey_valid = false;
+	bool		function_in_use = false;
 
 	/*
 	 * Lookup the pg_proc tuple by Oid; we'll need it in any case
@@ -204,7 +205,12 @@ recheck:
 			/*
 			 * Nope, so remove it from hashtable and try to drop associated
 			 * storage (if not done already).
+			 * Check whether function is in use or not before deleting it, if
+			 * function is not in use - its memory context will be freed.
+			 * PLtsql_function struct itself could have been allocated in the same
+			 * context, so we avoid accessing struct fields after the delete call.
 			 */
+			function_in_use = function->use_count != 0;
 			delete_function(function);
 
 			/*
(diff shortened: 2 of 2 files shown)


contrib/babelfishpg_tsql/src/pl_funcs.c
```diff
@@ -864,8 +865,11 @@ pltsql_free_function_memory(PLtsql_function *func)
 	 * fn_extra pointers to it).
 	 */
 	if (func->fn_cxt)
-		MemoryContextDelete(func->fn_cxt);
```
kuntalghosh, line 867: The variable func->fn_cxt is a pointer. MemoryContextDelete frees up the memory pointed by func->fn_cxt and it doesn't free func->fn_cxt itself. You can't access func->fn_cxt if you do a pfree(func). But, that's not happening in this case.
```
func->fn_cxt = NULL
```
The above code just changes the pointer value to NULL. So, I don't see why it'll cause an usage of memory after free issue.

staticlibs: I well could have misread the existing logic in `pl_comp.c`, my reasoning was based on the following understanding:
 1. [func_cxt is set as a current memory context](https://github.com/babelfish-for-postgresql/babelfish_extensions/blob/aaa7127d18f01188b8945029eb5a27dc2298f378/contrib/babelfishpg_tsql/src/pl_comp.c#L1055)
 2. [function structure itself is allocated in current memory contex](https://github.com/babelfish-for-postgresql/babelfish_extensions/blob/aaa7127d18f01188b8945029eb5a27dc2298f378/contrib/babelfishpg_tsql/src/pl_comp.c#L1058)
 3. [the same memory context is assigned to fn_cxt field of the function](https://github.com/babelfish-for-postgresql/babelfish_extensions/blob/aaa7127d18f01188b8945029eb5a27dc2298f378/contrib/babelfishpg_tsql/src/pl_comp.c#L1066)

I assumed that, after "MemoryContextDelete will free up the memory pointed by func->fn_cxt" and when the function struct itself was allocated in this memory (using the same context) - then writes to the fields of this function structure will be "use after free" situation.

kuntalghosh: Actually, you are right. The function memory itself is being allocated in func->fn_cxt. And, we're accessing it from more than one place after it's freed. Apart from the one you've fixed, we also need to fix [this](https://github.com/babelfish-for-postgresql/babelfish_extensions/blob/aaa7127d18f01188b8945029eb5a27dc2298f378/contrib/babelfishpg_tsql/src/pl_comp.c#L224). Here, delete_function is freeing up the memory if usage count is 0, then again accessing the variable. Could you please include this in your fix as well?

staticlibs: Sorry for delay, will be able to update the PR in a few days.

staticlibs: Moved `use_count` check to happen before the `delete_function` call, the logic is kept the same.

staticlibs:
Besides the `use_count` check, added some comments and rebased the branch to the latest.

kuntalghosh (approved):

2jungkook (approved):

PR #1590, opened 2023-07-04
Support extended property

ghost:
Description
===========
The extended property includes the following features: sys.sp_addextendedproperty: Adds a new extended property to a database object. sys.sp_updateextendedproperty: Updates the value of an existing extended property. sys.sp_dropextendedproperty: Drops an existing extended property. sys.fn_listextendedproperty: Returns extended property values of database objects. sys.extended_properties: Returns a row for each extended property in the current database.

Solution
========
The implementation of extended property is carried out in the following manner: Firstly, a new catalog table is added, which records details such as dbid, schema, major_name, minor_name, and type. 'sys.extended_properties' is view based on this table.
Secondly, we implement 'sp_addextendedproperty', 'sp_updateextendedproperty', 'sp_dropextendedproperty', and 'fn_listextendedproperty' procedure or function. Lastly, we modify major_name and minor_name cascade when calling 'sp_rename' and delete cascade when dropping an object.

Limitation
==========
At present, we only support extended properties for the most commonly used objects, which include database, schema, table, table column, view, sequence, procedure, function, and type.

### Description

### Issues Resolved

* **Use case based -**

* **Boundary conditions -**

* **Arbitrary inputs -**

* **Negative test cases -**

* **Performance tests -**

* **Tooling impact -**

* **Client tests -**

### Check List

### DIFF
--- a/contrib/babelfishpg_tsql/sql/ownership.sql
+++ b/contrib/babelfishpg_tsql/sql/ownership.sql
@@ -618,3 +618,66 @@ GRANT EXECUTE ON PROCEDURE sys.babelfish_remove_domain_mapping_entry TO PUBLIC;
 CREATE OR REPLACE PROCEDURE sys.babelfish_truncate_domain_mapping_table()
   AS 'babelfishpg_tsql', 'babelfish_truncate_domain_mapping_table_internal' LANGUAGE C;
 GRANT EXECUTE ON PROCEDURE sys.babelfish_truncate_domain_mapping_table TO PUBLIC;
+
+CREATE TABLE sys.babelfish_extended_properties (
+  dbid smallint NOT NULL,
+  schema_name name NOT NULL,
+  major_name name NOT NULL,
+  minor_name name NOT NULL,
+  type sys.varchar(50) NOT NULL,
+  name sys.sysname NOT NULL,
+  orig_name sys.sysname NOT NULL,
+  value sys.sql_variant,
+  PRIMARY KEY (dbid, type, schema_name, major_name, minor_name, name)
+);
+GRANT SELECT on sys.babelfish_extended_properties TO PUBLIC;
+SELECT pg_catalog.pg_extension_config_dump('sys.babelfish_extended_properties', '');
+
+-- This view contains many outer joins that could potentially impair performance.
+-- To optimize this, it may be beneficial to begin by acquiring the object identifier (OID) and verifying permissions.
+CREATE OR REPLACE VIEW sys.extended_properties
+AS
+SELECT
+	CAST((CASE
+		WHEN ep.type = 'DATABASE' THEN 0
+		WHEN ep.type = 'SCHEMA' THEN 3
+		WHEN ep.type IN ('TABLE', 'TABLE COLUMN', 'VIEW', 'SEQUENCE', 'PROCEDURE', 'FUNCTION') THEN 1
+		WHEN ep.type = 'TYPE' THEN 6
+		END) AS sys.tinyint) AS class,
+	CAST((CASE
+		WHEN ep.type = 'DATABASE' THEN 'DATABASE'
+		WHEN ep.type = 'SCHEMA' THEN 'SCHEMA'
+		WHEN ep.type IN ('TABLE', 'TABLE COLUMN', 'VIEW', 'SEQUENCE', 'PROCEDURE', 'FUNCTION') THEN 'OBJECT_OR_COLUMN'
+		WHEN ep.type = 'TYPE' THEN 'TYPE'
+	END) AS sys.nvarchar(60)) AS class_desc,
+	CAST((CASE
+		WHEN ep.type = 'DATABASE' THEN 0
+		WHEN ep.type = 'SCHEMA' THEN n.oid
+		WHEN ep.type IN ('TABLE', 'TABLE COLUMN', 'VIEW', 'SEQUENCE') THEN c.oid
+		WHEN ep.type IN ('PROCEDURE', 'FUNCTION') THEN p.oid
+		WHEN ep.type = 'TYPE' THEN t.oid
+	END) AS int) AS major_id,
+	CAST((CASE
+		WHEN ep.type = 'DATABASE' THEN 0
+		WHEN ep.type = 'SCHEMA' THEN 0
+		WHEN ep.type IN ('TABLE', 'VIEW', 'SEQUENCE', 'PROCEDURE', 'FUNCTION', 'TYPE') THEN 0
+		WHEN ep.type = 'TABLE COLUMN' THEN a.attnum
+	END) AS int) AS minor_id,
+	ep.orig_name AS name, ep.value AS value
+	FROM sys.babelfish_extended_properties ep
+		LEFT JOIN pg_catalog.pg_namespace n ON n.nspname = ep.schema_name
+		LEFT JOIN pg_catalog.pg_class c ON c.relname = ep.major_name AND c.relnamespace = n.oid
+		LEFT JOIN pg_catalog.pg_proc p ON p.proname = ep.major_name AND p.pronamespace = n.oid
+		LEFT JOIN pg_catalog.pg_type t ON t.typname = ep.major_name AND t.typnamespace = n.oid
+		LEFT JOIN pg_catalog.pg_attribute a ON a.attrelid = c.oid AND a.attname = ep.minor_name
+	WHERE ep.dbid = sys.db_id() AND
+	(CASE
+		WHEN ep.type = 'DATABASE' THEN true
+		WHEN ep.type = 'SCHEMA' THEN has_schema_privilege(n.oid, 'USAGE, CREATE')
+		WHEN ep.type IN ('TABLE', 'VIEW', 'SEQUENCE') THEN (has_table_privilege(c.oid, 'SELECT, INSERT, UPDATE, DELETE, TRUNCATE, REFERENCES, TRIGGER'))
+		WHEN ep.type IN ('TABLE COLUMN') THEN (has_table_privilege(c.oid, 'SELECT, INSERT, UPDATE, DELETE, TRUNCATE, REFERENCES, TRIGGER') OR has_column_privilege(a.attrelid, a.attname, 'SELECT, INSERT, UPDATE, REFERENCES'))
+		WHEN ep.type IN ('PROCEDURE', 'FUNCTION') THEN has_function_privilege(p.oid, 'EXECUTE')
+		WHEN ep.type = 'TYPE' THEN has_type_privilege(t.oid, 'USAGE')
+	END)
+	ORDER BY class, class_desc, major_id, minor_id, ep.orig_name;
+GRANT SELECT ON sys.extended_properties TO PUBLIC;
--- a/contrib/babelfishpg_tsql/src/extendedproperty.c
+++ b/contrib/babelfishpg_tsql/src/extendedproperty.c
@@ -1,0 +1,1167 @@
+/*-------------------------------------------------------------------------
+ *
+ * extendedproperty.c
+ *	  support extended property
+ *
+ *-------------------------------------------------------------------------
+ */
+
+#include "postgres.h"
+
+#include "access/genam.h"
+#include "access/skey.h"
+#include "access/table.h"
+#include "access/xact.h"
+#include "catalog/indexing.h"
+#include "catalog/pg_namespace.h"
+#include "catalog/pg_proc.h"
+#include "catalog/pg_type.h"
+#include "funcapi.h"
+#include "miscadmin.h"
+#include "tsearch/ts_locale.h"
+#include "utils/builtins.h"
+#include "utils/catcache.h"
+#include "utils/datum.h"
+#include "utils/formatting.h"
+#include "utils/syscache.h"
+#include "utils/fmgroids.h"
+#include "utils/rel.h"
+
+#include "catalog.h"
+#include "extendedproperty.h"
+#include "multidb.h"
+#include "pltsql.h"
+#include "session.h"
+
+typedef enum ExtendedPropertyProc
+{
+	SP_ADDEXTENDEDPROPERTY = 0,
+	SP_UPDATEEXTENDEDPROPERTY,
+	SP_DROPEXTENDEDPROPERTY
+} ExtendedPropertyProc;
+
+const char *const ExtendedPropertyTypeNames[] = {
+    "DATABASE",
+    "SCHEMA",
+    "TABLE",
+    "VIEW",
+    "SEQUENCE",
+    "PROCEDURE",
+    "FUNCTION",
+    "TYPE",
+    "TABLE COLUMN"
+};
+
+StaticAssertDecl(lengthof(ExtendedPropertyTypeNames) == EXTENDED_PROPERTY_MAX,
+				 "array length mismatch");
+
+PG_FUNCTION_INFO_V1(sp_addextendedproperty);
+PG_FUNCTION_INFO_V1(sp_updateextendedproperty);
+PG_FUNCTION_INFO_V1(sp_dropextendedproperty);
+PG_FUNCTION_INFO_V1(fn_listextendedproperty);
+
+static void init_scan_key(ScanKeyData *scanKey,
+						  int *nkeys,
+						  int16 db_id,
+						  const char *type,
+						  const char *schema_name,
+						  const char *major_name,
+						  const char *minor_name,
+						  const char *name);
+static void sp_execextended_property(PG_FUNCTION_ARGS, ExtendedPropertyProc proc);
+static bool get_extended_property_from_tuple(Relation relation, HeapTuple tuple,
+											 Datum *values, bool *nulls, int len);
+static char* get_value_by_name_from_array(ArrayType *array, const char *name);
+
+static void
+init_scan_key(ScanKeyData *scanKey,
+			  int *nkeys,
+			  int16 db_id,
+			  const char *type,
+			  const char *schema_name,
+			  const char *major_name,
+			  const char *minor_name,
+			  const char *name)
+{
+	*nkeys = 0;
+	ScanKeyInit(&scanKey[*nkeys],
+				Anum_bbf_extended_properties_dbid,
+				BTEqualStrategyNumber, F_INT2EQ,
+				Int16GetDatum(db_id));
+	(*nkeys)++;
+
+	if (type)
+	{
+		ScanKeyInit(&scanKey[*nkeys],
+					Anum_bbf_extended_properties_type,
+					BTEqualStrategyNumber, F_TEXTEQ,
+					CStringGetTextDatum(type));
+		(*nkeys)++;
+	}
+
+	if (schema_name)
+	{
+		ScanKeyInit(&scanKey[*nkeys],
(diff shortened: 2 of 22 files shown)


contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -2989,7 +2993,13 @@ bbf_ProcessUtility(PlannedStmt *pstmt,
 				DropStmt   *drop_stmt = (DropStmt *) parsetree;
 
 				if (drop_stmt->removeType != OBJECT_SCHEMA)
+				{
+					if (sql_dialect == SQL_DIALECT_TSQL)
+						bbf_ExecDropStmt(drop_stmt);
```
HarshLunagariya, line 2998: Here we are cleaning up the extended property catalog without making sure that statement execution is successful or not. Please consider the scenario that what if drop statement fail due to some error like permission error or anything else - ideally catalog shouldn't be cleaned up in that case but according to current implementation catalog entry will be removed. Please double-check and add a test-case for the same.

HarshLunagariya (changes requested):

contrib/babelfishpg_tsql/sql/ownership.sql
```diff
@@ -617,4 +617,66 @@ GRANT EXECUTE ON PROCEDURE sys.babelfish_remove_domain_mapping_entry TO PUBLIC;
...
+  type sys.varchar(50) NOT NULL,
+  name sys.sysname NOT NULL,
+  orig_name sys.sysname NOT NULL,
+  value sys.sql_variant,
+  PRIMARY KEY (dbid, type, schema_name, major_name, minor_name, name)
+);
+GRANT SELECT on sys.babelfish_extended_properties TO PUBLIC;
+
+-- This view contains many outer joins that could potentially impair performance.
+-- To optimize this, it may be beneficial to begin by acquiring the object identifier (OID) and verifying permissions.
```
robverschoor, line 637: It is not a-typical to see customer applications with thousands of extended properties being defined (see one with 57k of these). We should know whether this view is still practically usable whith 100, 1000, 10000 attributes.

contrib/babelfishpg_tsql/sql/ownership.sql
```diff
@@ -617,4 +617,66 @@ GRANT EXECUTE ON PROCEDURE sys.babelfish_remove_domain_mapping_entry TO PUBLIC;
...
+  schema_name name NOT NULL,
+  major_name name NOT NULL,
+  minor_name name NOT NULL,
+  type sys.varchar(50) NOT NULL,
+  name sys.sysname NOT NULL,
+  orig_name sys.sysname NOT NULL,
+  value sys.sql_variant,
+  PRIMARY KEY (dbid, type, schema_name, major_name, minor_name, name)
+);
+GRANT SELECT on sys.babelfish_extended_properties TO PUBLIC;
```
robverschoor, line 633: By granting SELECT to PUBLIC on the base table, what happens to user X trying to see attributes on an object owned by a different user? Wouldn't that always be possible through this base table? SQL Server allows you to see only attributes on objects which you have permissions on.

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -3450,6 +3458,119 @@ sp_rename_internal(PG_FUNCTION_ARGS)
...
+			truncate_tsql_identifier(minor_name);
+			truncate_tsql_identifier(new_minor_name);
+
+			type = ExtendedPropertyTypeNames[EXTENDED_PROPERTY_TABLE_COLUMN];
+			update_extended_property(db_id, type, schema_name,
+									 major_name, minor_name,
+									 Anum_bbf_extended_properties_minor_name,
+									 new_minor_name);
+		}
+	}
```
robverschoor, line 3589: must catch the case that objtype is none of the object types tested for

test/JDBC/input/BABEL-EXTENDEDPROPERTY-v2-vu-cleanup.mix
```diff
@@ -0,0 +1,24 @@
+-- tsql
+USE babel_extended_properties_db
```
robverschoor, line 2: Comments on the combined test cases files:

- Please add test cases for object/column (etc) names that use bracketed identifiers and double-quoted identifiers, as well as such identifiers containing non-standard characters (like space, comma, etc)

- All objects in the test cases seem to be created with lowercase names, please soem test caes where the object is created in mixed-case and the sp_XXXextendedproperty procs are called with a different case.

- Please add test cases where identifiers and attribute descriptions contains non-western characters (Chinese characters would be a good test case)

- Are there any negative tests where an identifier does not exist or the executing user has no permissions to access the object?

robverschoor (commented):
Please see my comments on permissions, performance and additional test cases.

contrib/babelfishpg_tsql/src/pl_handler.c
markrui3: Firstly, dropping catalog and dropping statement are in the same transaction. If dropping statement fails, dropping catalog will be rolled back.
Secondly, if dropping statement fails due to permission error, it's very likely that dropping catalog will fail due to permission error, too.

contrib/babelfishpg_tsql/sql/ownership.sql
markrui3: It’s surely usable with 100, 1000, 10000 attributes. In my test, it costs about 6 seconds to return 100000 attributes and 1 second to return 10000 attributes.

The comments here only follow Di's advise, to mark potentially performance decreasing.

contrib/babelfishpg_tsql/sql/ownership.sql
markrui3: Yes, the base table is visible to everyone. However, most base tables in Babelfish are visible to everyone, too. I have the impression that it's possibly for major version upgrade.
Maybe it's not actually a problem?

contrib/babelfishpg_tsql/src/procedures.c
markrui3: If objtype is none of the object types tested for, ProcessUtility will fail and ereport error, rename_extended_property won't be executed.

contrib/babelfishpg_tsql/src/pl_handler.c
HarshLunagariya: Got it. Wanted to confirm 1st point.

markrui3: 1st point is based on the MVCC realization of PG.
Simple verification is to add elog(ERROR, ...) after bbf_ExecDropStmt, and you will see the same result as I said.

test/JDBC/input/BABEL-EXTENDEDPROPERTY-v2-vu-cleanup.mix
markrui3: 1. done.
2. What you want is the test case for table column. Since table column has the most detailed object types, I add test case of trailing space, uppercase/lowercase. And I add test case for long name(more than 64) of shema, table, table column.
3. I have tested Chinese characters before, and you give the below answer. 
image
I think it's not the problem of my implementation.
4. Yes, all of negative test are considered, including not exist objects and permission check.
I create two users to test permissions, one is sysadmin, the other one is normal user.

contrib/babelfishpg_tsql/src/extendedproperty.c
```diff
@@ -0,0 +1,1165 @@
...
+			if (strncmp(type, ExtendedPropertyTypeNames[EXTENDED_PROPERTY_TABLE], 5) == 0 ||
+				strncmp(type, ExtendedPropertyTypeNames[EXTENDED_PROPERTY_VIEW], 4) == 0 ||
+				strncmp(type, ExtendedPropertyTypeNames[EXTENDED_PROPERTY_SEQUENCE], 8) == 0)
+			{
+				Form_pg_class	classform;
+				ArrayType		*reloptions;
+
+				heaptuple = SearchSysCache2(RELNAMENSP,
+											CStringGetDatum(major_name),
+											Int16GetDatum(schema_id));
```
HarshLunagariya: Please use `ObjectIdGetDatum()`, Oid is 32 bit and unsigned, so even Int32GetDatum() can't be utilised.

markrui3: You are right. Even if the implementations of ObjectIdGetDatum and Int16GetDatum are same, ObjectIdGetDatum is better in this case.

Done.

contrib/babelfishpg_tsql/sql/ownership.sql
```diff
@@ -617,4 +617,66 @@ GRANT EXECUTE ON PROCEDURE sys.babelfish_remove_domain_mapping_entry TO PUBLIC;
 
 CREATE OR REPLACE PROCEDURE sys.babelfish_truncate_domain_mapping_table()
   AS 'babelfishpg_tsql', 'babelfish_truncate_domain_mapping_table_internal' LANGUAGE C;
-GRANT EXECUTE ON PROCEDURE sys.babelfish_truncate_domain_mapping_table TO PUBLIC;
\ No newline at end of file
+GRANT EXECUTE ON PROCEDURE sys.babelfish_truncate_domain_mapping_table TO PUBLIC;
+
+CREATE TABLE sys.babelfish_extended_properties (
```
HarshLunagariya, line 622: Please mark this relation as configuration relation using  `pg_extension_config_dump()` function, so that data/content is also dumped along with relation. Every babelfish catalog should be marked as configuration relation.

test/JDBC/upgrade/latest/schedule
```diff
@@ -133,6 +133,7 @@ babel_datetime
 Babel_domain_mapping_test
 BABEL-EXECUTE_AS_CALLER
 BABEL-EXTENDEDPROPERTY
+BABEL-EXTENDEDPROPERTY-v2
```
HarshLunagariya, line 136: We should add this to older version's schedule files as well to cover other m\Mvu workflows. @2jungkook @rishabhtanwar29 can comment better.

rishabhtanwar29: Yep, since `prepare` script is using newly added procedures inside test procedures, it could be executed on older versions without any issue. So we should add this test all previous upgrade scripts.

contrib/babelfishpg_tsql/sql/ownership.sql
markrui3: I think `pg_dump` has already dumped data/content. `pg_extension_config_dump()` is used to mark special babelfish catalog and dump part of data during dumping.
If we mark every newly babelfish catalog, we should also append code in PG. Should we always modify PG?

test/JDBC/upgrade/latest/schedule
markrui3: @HarshLunagariya @rishabhtanwar29 
As @rishabhtanwar29 said, prepare script is using newly added procedures inside test procedures, BABEL-EXTENDEDPROPERTY-v2-vu-prepare.out will fail during any of Version Upgrade Test. Error is below:

```
CREATE PROCEDURE babel_sp_addextendedproperty
    @name sys.sysname,
    @value sys.sql_variant = NULL,
    @level0type VARCHAR(128) = NULL,
    @level0name sys.sysname = NULL,
    @level1type VARCHAR(128) = NULL,
    @level1name sys.sysname = NULL,
    @level2type VARCHAR(128) = NULL,
    @level2name sys.sysname = NULL
AS
BEGIN
    EXEC sp_addextendedproperty @name, @value, @level0type, @level0name, @level1type, @level1name, @level2type, @level2name
END
GO
~~ERROR (Code: 33557097)~~

~~ERROR (Message: 'sp_addextendedproperty' is not currently supported in Babelfish)~~


CREATE PROCEDURE babel_sp_updateextendedproperty
    @name sys.sysname,
    @value sys.sql_variant = NULL,
    @level0type VARCHAR(128) = NULL,
    @level0name sys.sysname = NULL,
    @level1type VARCHAR(128) = NULL,
    @level1name sys.sysname = NULL,
    @level2type VARCHAR(128) = NULL,
    @level2name sys.sysname = NULL
AS
BEGIN
    EXEC sp_updateextendedproperty @name, @value, @level0type, @level0name, @level1type, @level1name, @level2type, @level2name
END
GO
~~ERROR (Code: 33557097)~~

~~ERROR (Message: 'sp_updateextendedproperty' is not currently supported in Babelfish)~~


CREATE PROCEDURE babel_sp_dropextendedproperty
    @name sys.sysname,
    @level0type VARCHAR(128) = NULL,
    @level0name sys.sysname = NULL,
    @level1type VARCHAR(128) = NULL,
    @level1name sys.sysname = NULL,
    @level2type VARCHAR(128) = NULL,
    @level2name sys.sysname = NULL
AS
BEGIN
    EXEC sp_dropextendedproperty @name, @level0type, @level0name, @level1type, @level1name, @level2type, @level2name
END
GO
~~ERROR (Code: 33557097)~~

~~ERROR (Message: 'sp_dropextendedproperty' is not currently supported in Babelfish)~~
```

HarshLunagariya: This should be happening for the versions for which sp_xxxxextendedproperty is not supported. You can update the version specific expected output.

rishabhtanwar29: Ahh, generally we doesn't check the objects inside a procedure's body at `CREATE` time but reason for the above errors is that we are detecting unsupported procedure's in ANTLR parser itself [here](https://github.com/babelfish-for-postgresql/babelfish_extensions/blob/a7e79421f66030d1d3b97cf09136da4a4fce8958/contrib/babelfishpg_tsql/src/tsqlUnsupportedFeatureHandler.cpp#L1490). I was not aware of this before so looks like we won't be able to add this test in previous versions since it looks like all these procedures are new.

contrib/babelfishpg_tsql/sql/ownership.sql
rishabhtanwar29: @markrui3 If we want the data of a particular catalog table to be dumped by `pg_dump` then we would need to mark that table as config table (using `pg_extension_config_dump()`). This is needed to support `pg_dump/pg_dumpall/pg_restore` utilities, `pg_upgrade` works with/without this change. Now, short answer of your question is "Yes, we need to modify PG as well". The reason for this that we have added logic in `pg_dump` such that dump fails if it detects a new unhandled catalog table ([reference1](https://github.com/babelfish-for-postgresql/postgresql_modified_for_babelfish/blob/BABEL_3_X_DEV__PG_15_X/src/bin/pg_dump/dump_babel_utils.c#L897) and [reference2](https://github.com/babelfish-for-postgresql/postgresql_modified_for_babelfish/blob/BABEL_3_X_DEV__PG_15_X/src/bin/pg_dump/dump_babel_utils.c#L844)), which is because we want to make sure that the developer adding a new catalog considers it's dump/restore impacts so that we don't blindly dump whole contents of the table.  
`Hint`: You can look at the above two references (functions `addFromClauseForLogicalDatabaseDump` and `addFromClauseForPhysicalDatabaseDump`) and add logic to dump the contents of this new catalog table. (Comments around these functions should tell you how to do that)
Finally, to test this, we can add our test file in `test/JDBC/upgrade/master/schedule` file which will be executed in pg_dump/restore github actions.

markrui3: Thanks. The modification for `pg_dump` is at: https://github.com/babelfish-for-postgresql/postgresql_modified_for_babelfish/pull/162

The pg_dump/resotre test in this patch will be passed after above pr is merged.

test/JDBC/upgrade/master/schedule
```diff
@@ -107,6 +107,7 @@ BABEL-CHECK-CONSTRAINT
 babel_datetime
 BABEL-EXECUTE_AS_CALLER
 BABEL-EXTENDEDPROPERTY
+BABEL-EXTENDEDPROPERTY-v2
```
rishabhtanwar29: This test is probably going to fail in database-level dump/restore github action since we only dump `master` database in that action and we don't dump logins as well so it's better to not add this test here directly. Although this test should already be getting executed in instance-level dump/restore (in this we dump everything).
Is it possible to write a minimal test in which we just create objects in master database so that we can add that test here?

markrui3: Now `BABEL-EXTENDEDPROPERTY-v2` is used to test permissions between sysadmin and normal user. sysadmin can add, update, drop, list extended property of six objects, while normal user can only list extended property if he has SELECT privilege.
So, is it necessary to add an individual minimal test only in `test/JDBC/upgrade/master/schedule`? If the answer is yes, what kind of test should I add in this minimal test?

rishabhtanwar29: Actually, my idea is to have an automated test to verify if the changes we have made in `pg_dump` function `addFromClauseForLogicalDatabaseDump` works correctly. Which is that, we only dump extended properties of a particular logical database we are dumping. So, a minimal test can add extended properties for multiple user databases in prepare script and then verify that only extended properties related to a particular database are dumped with logical database dump.
Having said that, your pg_dump changes look quite simple so no need to create a whole new test file but we can just add such test in existing BABEL-EXTENDEDPROPERTY file if it is not already there.

markrui3: We can't add such test in existing BABEL-EXTENDEDPROPERTY file because `sp_addextendedproperty` is the only way to insert into sys.babelfish_extended_properties catalog and `sp_addextendedproperty` is not supported in lower version like 13.x, 14.x. 
If I add such test, many major version upgrade tests will be failed.

timchang514:
If BABEL-4161 test is failing, you might be missing this commit locally: https://github.com/babelfish-for-postgresql/postgresql_modified_for_babelfish/commit/00d2d4e8f3fa5ba714b6ff765726422c07193b92

timchang514:
The failing 4161 test is probably because this commit https://github.com/babelfish-for-postgresql/postgresql_modified_for_babelfish/commit/00d2d4e8f3fa5ba714b6ff765726422c07193b92 is missing. It doesn't reproduce locally so perhaps github actions is not pulling the latest engine branch, for some reason. 

I wonder, could github actions be pulling the engine branch from the closed corresponding engine PR for this change? https://github.com/babelfish-for-postgresql/postgresql_modified_for_babelfish/commit/346a98bf2e3c257321b60a0c0f899aed6a32d2d8?

markrui3:
> The failing 4161 test is probably because this commit [babelfish-for-postgresql/postgresql_modified_for_babelfish@00d2d4e](https://github.com/babelfish-for-postgresql/postgresql_modified_for_babelfish/commit/00d2d4e8f3fa5ba714b6ff765726422c07193b92) is missing. It doesn't reproduce locally so perhaps github actions is not pulling the latest engine branch, for some reason.
> 
> I wonder, could github actions be pulling the engine branch from the closed corresponding engine PR for this change? [babelfish-for-postgresql/postgresql_modified_for_babelfish@346a98b](https://github.com/babelfish-for-postgresql/postgresql_modified_for_babelfish/commit/346a98bf2e3c257321b60a0c0f899aed6a32d2d8)?

You are right. GIthub actions is pulling commit from my private repository which is not synchronized with master.

macwoody (approved):

PR #1592, opened 2023-07-04
OPENQUERY for Linked Servers does not work when used from user created database

thephantomthief:
### Description

Previously the user mapping that got created as part of sp_addlinkedsrvlogin was only working in master database. In T-SQL the linked login is not specific to login unlike the user mappings in PG. To fix this we create the user mapping for the public role so that the linked login is not specific to any single role.

Task: BABEL-4229

* **Use case based -** Yes

* **Arbitrary inputs -** N/A

* **Performance tests -** N/A

* **Tooling impact -** N/A

* **Client tests -** N/A

### Check List


basasairohan (approved):

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -2826,17 +2827,34 @@ sp_droplinkedsrvlogin_internal(PG_FUNCTION_ARGS)
 				(errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
 				 errmsg("Only @locallogin = NULL is supported. Configuring remote server access specific to local login is not yet supported")));
 
+	remove_trailing_spaces(servername);
```
rishabhtanwar29, line 2830: Suggestion: can be moved at the top.

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -2826,17 +2827,34 @@ sp_droplinkedsrvlogin_internal(PG_FUNCTION_ARGS)
...
+
+	/* Check if servername is valid */
+	get_foreign_server_oid(servername, false);
```
rishabhtanwar29, line 2833: Question: This will throw error in case foreign server does not exist at CREATE time, have we verified if this is expected in T-SQL? If yes, then can we have a test case for it?

contrib/babelfishpg_tsql/src/procedures.c
thephantomthief: IMO it is better to check for errors first before we remove the trailing spaces. Thoughts?

rishabhtanwar29 (approved):

shardgupta (approved):

PR #1600, opened 2023-07-10
GETDATE() and other date functions incorrectly return different values in the same query

KushaalShroff:
### Description
Ideally GETDATE() and other date functions are executed before the rows are fetched. The result is cached and re-used in all places within the query, to align with customer expectation of such behaviour.
However, in Babelfish the functions return different results per-row and per instance of the function in the query. This causes customers to receive wrong results in some cases.
To fix this we use statement_timestamp internally which returns the start time of the current statement and its value does not change inside the current command being executed.

Signed-off-by: Kushaal Shroff <kushaal@amazon.com>

### Issues Resolved
BABEL-4188

* **Use case based -**
*Added JDBC tests and we do not require upgrade tests since there is no change in signature of the function. Moreover we cannot test the actual value being returned by the function since it is going to be dynamic, we can only test for the stability of the function in the manner in which tests are added in this PR*

* **Boundary conditions -**

* **Arbitrary inputs -**

* **Negative test cases -**

* **Performance tests -**

* **Tooling impact -**

* **Client tests -**

### Check List

### DIFF
--- a/contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--3.2.0--3.3.0.sql
+++ b/contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--3.2.0--3.3.0.sql
@@ -139,6 +139,34 @@ inner join sys.babelfish_namespace_ext b on t.schema_name = b.orig_name
 inner join pg_catalog.pg_namespace ns on b.nspname = ns.nspname;
 GRANT SELECT ON sys.shipped_objects_not_in_sys TO PUBLIC;
 
+CREATE OR REPLACE FUNCTION sys.sysdatetime() RETURNS datetime2
+    AS $$select statement_timestamp()::datetime2;$$
+    LANGUAGE SQL;
+GRANT EXECUTE ON FUNCTION sys.sysdatetime() TO PUBLIC;
+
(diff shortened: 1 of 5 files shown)


test/JDBC/expected/getdatetest.out
```diff
@@ -0,0 +1,88 @@
+
+-- NOTE: Each test is expected to tale a lot of time
```
rishabhtanwar29: Nit: Tale -> Take?

rishabhtanwar29 (approved):

shardgupta (approved):

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--3.2.0--3.3.0.sql
```diff
@@ -139,6 +139,34 @@ inner join sys.babelfish_namespace_ext b on t.schema_name = b.orig_name
 inner join pg_catalog.pg_namespace ns on b.nspname = ns.nspname;
 GRANT SELECT ON sys.shipped_objects_not_in_sys TO PUBLIC;
 
+CREATE OR REPLACE FUNCTION sys.sysdatetime() RETURNS datetime2
```
surendravishnoi, line 142: Do we already have upgrade tests for these functions? If not, we should add them.

PR #1604, opened 2023-07-11
Fix backward scan flag for parallel queries

rl626:
### Description

Currently there is a bug in Babelfish that allows the backward scan flag to be set for parallel queries. However, backward scans cannot be parallelized. This bug results in assertion failures in the Postgres engine and leads Babelfish queries to crash.

### Issues Resolved

In this commit, we fix the issue by disallowing backward scan flag to be set when a query is in parallel mode. This issue occurs during the implementation of full text search in Babelfish. Fixing the issue unblocks full text search implementation.

### Check List

### DIFF
--- a/test/JDBC/input/BABEL-4261.sql
+++ b/test/JDBC/input/BABEL-4261.sql
@@ -6,7 +6,6 @@ select set_config('parallel_tuple_cost', 0, false);
 select set_config('min_parallel_table_scan_size', 0, false);
 GO
 
--- execution
 select a, count(*) from t_babel4261 group by a order by 2; -- should not crash
 GO
 
--- a/test/JDBC/input/BABEL-4261.sql
+++ b/test/JDBC/input/BABEL-4261.sql
@@ -6,7 +6,26 @@ select set_config('parallel_tuple_cost', 0, false);
 select set_config('min_parallel_table_scan_size', 0, false);
 GO
 
+-- show explicitly this is a parallel query plan
+select set_config('babelfishpg_tsql.explain_timing', 'off', false);
+GO
+
+select set_config('babelfishpg_tsql.explain_summary', 'off', false);
+GO
+
+SET BABELFISH_STATISTICS PROFILE ON
+GO
+
 select a, count(*) from t_babel4261 group by a order by 2; -- should not crash
+GO
+
+SET BABELFISH_STATISTICS PROFILE OFF
+GO
+
+select set_config('babelfishpg_tsql.explain_timing', 'on', false);
+GO
+
+select set_config('babelfishpg_tsql.explain_summary', 'on', false);
 GO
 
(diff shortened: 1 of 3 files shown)


test/JDBC/input/BABEL-4261.sql
```diff
@@ -0,0 +1,14 @@
+CREATE TABLE t_babel4261 (a int, b int);
+GO
+
+select set_config('parallel_setup_cost', 0, false);
+select set_config('parallel_tuple_cost', 0, false);
+select set_config('min_parallel_table_scan_size', 0, false);
+GO
+
+select a, count(*) from t_babel4261 group by a order by 2; -- should not crash
```
forestkeeper, line 24: could you use 
```
SET BABELFISH_STATISTICS PROFILE ON
```
at this test place, to explicit show it's a parallel query plan at here

test/JDBC/input/BABEL-4261.sql
```diff
@@ -0,0 +1,14 @@
...
+GO
+
+DROP TABLE t_babel4261;
```
forestkeeper, line 43:
```
select set_config('parallel_setup_cost', 0, false);
select set_config('parallel_tuple_cost', 0, false);
select set_config('min_parallel_table_scan_size', 0, false);

```

could you set those three parameters back to default value in this place

test/JDBC/input/BABEL-4261.sql
rl626: Done

test/JDBC/input/BABEL-4261.sql
rl626: I'm not quite sure how to set them back since I don't know what the default value is. Also, it seems that in this test https://github.com/babelfish-for-postgresql/babelfish_extensions/blob/48d416c52d445080022b738bc9f5d5fd7c0b777c/test/JDBC/input/pg_hint_plan/BABEL-3294.sql#L14 they are not set back. So I guess not setting them back is okay?

forestkeeper: No it's not, 3294 didn't do the right thing in here

```
EXEC sp_babelfish_configure 'babelfishpg_tsql.xxxx', 'default';
```

Would you try this to reset.

rl626: I tried doing that, but it doesn't seem to work:
```
1> EXEC sp_babelfish_configure 'babelfishpg_tsql.parallel_setup_cost', 'default';
2> go
unknown configuration: babelfishpg_tsql.parallel_setup_cost
```

rl626: I have done this by putting set_config in a transaction and make the third parameter true, so that the configurations will be set back to default when the transaction is committed.

forestkeeper (approved):

PR #1620, opened 2023-07-17
Support HOST_ID() T-SQL function

robverschoor:
## Description

The **HOST_ID()** function returns the PID of the client app PID on the client machine. This PID can be useful when dealing with situations where a client session needs to be terminated but this doesn't work from the server side. 
This client PID is transmitted to the server in the TDS login packet. Currently in babelfish, this value is already exposed through the catalog column **sys.dm_exec_sessions.host_process_id**. This also needs to be exposed through catalog column **sys.sysprocesses.hostprocess** (which is currently NULL) as well as through the function **HOST_ID()**. For a particular connection, these should all return the same value.
Note that for JDBC connections, the client PID is not currently available (this is a separate issue). In that case, **HOST_ID()** and **sys.sysprocesses.hostprocess** currently return 0 and **sys.dm_exec_sessions.host_process_id** currently returns NULL.

## Issues Resolved

BABEL-4272 Support HOST_ID() T-SQL function

## Test Scenarios Covered

- Since the actual value of the client PID cannot be predicted, unit tests are focused on testing for a numeric value and that values in HOST_ID(), **sys.dm_exec_sessions.host_process_id** and **sys.sysprocesses.hostprocess** are identical.

- Use case based - Yes

- Boundary conditions - N/A

- Arbitrary inputs - N/A

- Negative test cases - N/A

- Performance tests - N/A

- Tooling impact - N/A

- Client tests - N/A

### Check List


contrib/babelfishpg_tsql/sql/sys_functions.sql
```diff
@@ -3442,6 +3442,10 @@ RETURNS sys.NVARCHAR(128)  AS 'babelfishpg_tsql' LANGUAGE C STABLE;
 CREATE OR REPLACE FUNCTION sys.host_name()
 RETURNS sys.NVARCHAR(128)  AS 'babelfishpg_tsql' LANGUAGE C IMMUTABLE PARALLEL SAFE;
 
+CREATE OR REPLACE FUNCTION sys.host_id()
+RETURNS sys.VARCHAR(10)  AS 'babelfishpg_tsql' LANGUAGE C IMMUTABLE PARALLEL SAFE;
```
2jungkook, line 3446: https://github.com/MicrosoftDocs/sql-docs/blob/live/docs/t-sql/functions/host-id-transact-sql.md says its return type is `char(10)`. Is there a reason to use `varchar(10)`?

2jungkook:
For the PR description, we are using a different format. You can refer https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/1624.

contrib/babelfishpg_tsql/sql/sys_functions.sql
robverschoor: When using sys.CHAR(10), the following error is raised when creating the extension:

```
jdbc_testdb=# CREATE EXTENSION IF NOT EXISTS "babelfishpg_tds" CASCADE;
NOTICE:  installing required extension "babelfishpg_tsql"
NOTICE:  installing required extension "uuid-ossp"
NOTICE:  installing required extension "babelfishpg_common"
ERROR:  type modifier cannot be specified for shell type "sys.char"
```

So, that's why sys.VARCHAR(10) is used.

2jungkook (approved):

PR #1630, opened 2023-07-19
Support TSQL PIVOT operator 

RIC06X:
### Description

This PR brings PIVOT operator to Babelfish. 

Currently PIVOT support 
- 2 column
- 3 column
- 3+ column
- top pivot
- group by
- having
- where
- insert into
- select into
- union
- Pivot on table variable 
- Pivot on temp table
- Subquery 

Existing bugs:
* value column with table name will not work 
    * solution available
* execute function, procedure will fail in second run (CachedPlan bug)
    * need to investigate CachedPlan
* View cannot be created from pivot (grammar rule bug)
    * need investigation
* Join cannot produce correct result (grammar rule bug)
    * need investigation

### Issues Resolved
Task: BABEL-284

Column Based test:

* 2 column
* 3 column
* 3+ column

Select support

* top pivot
* group by
* having
* where

Output related test

* insert into
* select into
* union
* join usage

Temp table use cases

* Include PIVOT on a #temp table 

Table variable use cases

* @tablevariable
* Include PIVOT on a @tablevariable

Column test

* column with table name schema name

Pivot inside of other stmt:

* subquery contaning PIVOT
* table valued function
* procedure

### Check List


contrib/babelfishpg_tsql/src/hooks.h
```diff
@@ -10,6 +10,10 @@ extern IsExtendedCatalogHookType PrevIsExtendedCatalogHook;
 extern IsToastRelationHookType PrevIsToastRelationHook;
 extern IsToastClassHookType PrevIsToastClassHook;
 
+/* TODO: (pivot) Remove these later*/
+extern RawStmt	   *bbf_pivot_sql1;
+extern RawStmt	   *bbf_pivot_sql2;
```
RIC06X: will remove in upcoming commit

contrib/babelfishpg_tsql/src/pltsql.h
```diff
@@ -1894,6 +1897,7 @@ extern int	insert_bulk_rows_per_batch;
 extern int	insert_bulk_kilobytes_per_batch;
 extern bool insert_bulk_keep_nulls;
 
+
```
RIC06X: will remove in next commit

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -3435,3 +3510,415 @@ objectproperty_internal(PG_FUNCTION_ARGS)
...
+
+	plancache_list = lappend(plancache_list, plansource);
+	plan.plancache_list = plancache_list;
+	plan.oneshot = true;
+
+	ret = SPI_execute_plan_with_paramlist(&plan, NULL, read_only, tcount);
+
+	/* reset sql_dialect */
+	sql_dialect = prev_sql_dialect;
+
```
lejaokri: if `SPI_execute_plan_with_paramlist` hits an ERROR then you wouldnt be able to reset sql_dialect to the original value, unless you surround this with `PG_TRY` and `CATCH`. Also, do we ever get here when sql_dialect != SQL_DIALECT_TSQL?

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -3435,3 +3510,415 @@ objectproperty_internal(PG_FUNCTION_ARGS)
...
+	if (!tsql_outmost_context)
+		ereport(ERROR,
+			(errcode(ERRCODE_SYNTAX_ERROR),
+				errmsg("pivot outer context not found")));
+
+	oldcontext = MemoryContextSwitchTo(tsql_outmost_context);
+	per_pivot_list = list_nth_node(List, tsql_outmost_estat->pivot_parsetree_list, tsql_outmost_estat->pivot_number - 1);
+	bbf_pivot_src_sql = list_nth_node(RawStmt, per_pivot_list, 0);
+	bbf_pivot_cat_sql = list_nth_node(RawStmt, per_pivot_list, 1);
+	MemoryContextSwitchTo(oldcontext);
```
lejaokri: Add a comment what's happening here. 
Also, add Assert check on the length of the lists you're trying to access.

For instance:
```
Assert(list_length(per_pivot_list) >= 2); 
bbf_pivot_src_sql = list_nth_node(RawStmt, per_pivot_list, 0);
bbf_pivot_cat_sql = list_nth_node(RawStmt, per_pivot_list, 1);
```

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -3435,3 +3510,415 @@ objectproperty_internal(PG_FUNCTION_ARGS)
...
+
+	if (!per_pivot_list)
+		ereport(ERROR,
+			(errcode(ERRCODE_SYNTAX_ERROR),
+				errmsg("pivot parsetree list not found")));
```
lejaokri: shouldn't you do this check before accessing `per_pivot_list`?

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -3435,3 +3510,415 @@ objectproperty_internal(PG_FUNCTION_ARGS)
...
+	 * tuples are in our tuplestore and passed back through rsinfo->setResult.
+	 * rsinfo->setDesc is set to the tuple description that we actually used
+	 * to build our tuples with, so the caller can verify we did what it was
+	 * expecting.
+	 */
+	rsinfo->setDesc = tupdesc;
+	MemoryContextSwitchTo(oldcontext);
+
+	oldcontext = MemoryContextSwitchTo(tsql_outmost_context);
+	/* TODO: do we need to free the list */
```
lejaokri: Have you completed this TODO?

Also, add more comments what you're doing here.

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -3435,3 +3510,415 @@ objectproperty_internal(PG_FUNCTION_ARGS)
...
+							"return 1 column of at least one row")));
+
+		for (i = 0; i < proc; i++)
+		{
+			bbf_pivot_cat_desc *catdesc;
+			char	   *catname;
+			HeapTuple	spi_tuple;
+
+			/* get the next sql result tuple */
+			spi_tuple = spi_tuptable->vals[i];
```
lejaokri, line 3735: You are looping from `i=0` to `i<proc` and accessing `spi_tuptable->vals[i]`.
Make sure you dont go out of bounds. I suggest add Assertion here.

```
Assert(spi_tuptable->numvals >= proc);
```

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -3435,3 +3510,415 @@ objectproperty_internal(PG_FUNCTION_ARGS)
...
+			catname = SPI_getvalue(spi_tuple, spi_tupdesc, 1);
+			if (catname == NULL)
+				ereport(ERROR,
+						(errcode(ERRCODE_SYNTAX_ERROR),
+						 errmsg("provided \"categories\" SQL must " \
+								"not return NULL values")));
+			
+			SPIcontext = MemoryContextSwitchTo(per_query_ctx);
+			catdesc = (bbf_pivot_cat_desc *) palloc(sizeof(bbf_pivot_cat_desc));
+			catdesc->catname = catname;
```
lejaokri, line 3747: You need to do a deep copy of `catname` because you are doing a `SPI_finish()` before returning `bbf_pivot_hash`. You can try to confirm this by running one of your tests that go to this codepath with valgrind enabled.

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -3435,3 +3510,415 @@ objectproperty_internal(PG_FUNCTION_ARGS)
...
+					 errdetail("Query-specified return " \
+							   "tuple has %d columns but bbf_pivot " \
+							   "returns %d.", tupdesc->natts, result_ncols)));
+
+		/* allocate space and make sure it's clear */
+		values = (char **) palloc0(result_ncols * sizeof(char *));
+		columngroup = (char **) palloc0((ncols - 2) * sizeof(char *));
+		lastcolumngroup = (char **) palloc0((ncols - 2) * sizeof(char *));
+
+		for (i = 0; i < proc; i++)
```
lejaokri, line 3846: add assert before the for loop

```
Assert(proc >= spi_tuptable->numvals); 
```

Also, let me know once you have testcases that exercise this code path and I can help you run valgrind.
The memory management in this codepath is a bit complicated so let's see if valgrind can help us detect some issues that can be missed easily.

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -3435,3 +3510,415 @@ objectproperty_internal(PG_FUNCTION_ARGS)
...
+			ereport(ERROR,
+					(errcode(ERRCODE_SYNTAX_ERROR),
+					 errmsg("invalid return type"),
+					 errdetail("Query-specified return " \
+							   "tuple has %d columns but bbf_pivot " \
+							   "returns %d.", tupdesc->natts, result_ncols)));
+
+		/* allocate space and make sure it's clear */
+		values = (char **) palloc0(result_ncols * sizeof(char *));
+		columngroup = (char **) palloc0((ncols - 2) * sizeof(char *));
```
lejaokri: I see a bunch of `ncols-2`. Put this in a variable and use that variable instead and add a comment why the -2.

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -3435,3 +3510,415 @@ objectproperty_internal(PG_FUNCTION_ARGS)
...
+					*/
+					if (!firstpass)
+					{
+						for (j = 0; j < result_ncols; j++)
+						{
+							if (values[j] == NULL)
+								values[j] = pstrdup("0");
+						}
+						/* rowid changed, flush the previous output row */
+						tuple = BuildTupleFromCStrings(attinmeta, values);
```
lejaokri, line 3895: It looks like `BuildTupleFromCStrings` is not doing a memcpy of `values` and you are doing a `xpfree` at ;om 3876 this means that the tuple would be pointing to a freed memory. 

Let's run valgrind here. :)

lejaokri (dismissed):

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -3435,3 +3510,415 @@ objectproperty_internal(PG_FUNCTION_ARGS)
 
 	PG_RETURN_NULL();
 }
+
+int 
+SPI_execute_raw_parsetree(RawStmt *parsetree, bool read_only, long tcount)
```
forestkeeper, line 3521: Pls add comment for this function illustrate why we have to add this function ( Since we have a discussion about why other SPI_execute not working)

forestkeeper (approved):

contrib/babelfishpg_tsql/runtime/functions.c
RIC06X: SPI_tuptable->values,  SPI_tuptable->numvals, and SPI_processed (proc) are managed by the same SPI_Current context, and SPI_processed is always equals to SPI_tuptable->numvals. So there is no need to do extra assertion.

contrib/babelfishpg_tsql/runtime/functions.c
RIC06X: SPI_tuptable->values,  SPI_tuptable->numvals, and SPI_processed (proc) are managed by the same SPI_Current context, and SPI_processed is always equals to SPI_tuptable->numvals. So there is no need to do extra assertion.

contrib/babelfishpg_tsql/runtime/functions.c
RIC06X: tuple = BuildTupleFromCStrings(attinmeta, values) use values(c string) and attinmeta to build a brand new datum not even in string type, it is more like atoi() (convert string to int) so there is no need to worry about memcpy, and free that value list is necessary later.

contrib/babelfishpg_tsql/runtime/functions.c
RIC06X: This part was originally from tablefunc.c’s load_categories_hash, and it is a good finding. The reason why it does not causing any problem is we only use catname as a argument of snprintf to generate a key for the hash table. snprintf(key, MAX_CATNAME_LEN - 1, "%s", CATNAME); \And later on the retrieval from bbf_pivot_HashTableLookup, we got catdesc, and we only used the catdesc->attidx but not the catdesc->catname. catname at this point is pointing to an random address but we are not using it, so we are not breaking anything, but it is still a very good finding.

PR #1642, opened 2023-07-23
Implement sp_who

robverschoor:
## Description

Implement T-SQL procedure **sp_who**. This procedure is frequently used by SQL Server DBAs to display the currently connected client sessions.
As the output of this procedure would be displayed in a user-unfriendly format wrapping around the screen in sqlcmd, a helper procedure **sp_babelfish_auutoformat** is used  which formats the contents of a table (or view) as narrowly as possible given its actual column contents. This way, the output of **sp_who** is nicely formatted and easily readable. 
In addition, 'postgres' can be specified as argument or as an optional second argument: in this case, active PG connections will also be reported; by default only TDS connections are reported. This was added to support practical Babelfsh situations in which it may be useful to see which PG sessions are connected as well. 

## Issues Resolved

BABEL-4131  Support for sp_who T-SQL stored procedure

## Test Scenarios Covered

Since the 'spid' and 'hostname' columns in the output of sp_who have unpredictable values, the tests should avoid showing those.
Normally, a workaround for this issue is to call the procedure as INSERT..EXECUTE <proc> and insert the result set into a table
and show only the columns that are predictable.
However, this does not work for procedures where the final SELECT is inside EXECUTE() -- and that is the case for sp_who.
All of this means we currently cannot actually run sp_who tests for the normal use cases, and we can only do negative tests and test where no output is produced.

- Use case based - Yes

- Boundary conditions - N/A

- Arbitrary inputs - N/A

- Negative test cases - Yes

- Performance tests - N/A

- Tooling impact - N/A

- Client tests - N/A

### Check List

For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/babelfish-for-p

### DIFF
--- a/contrib/babelfishpg_tsql/sql/babelfishpg_tsql.sql
+++ b/contrib/babelfishpg_tsql/sql/babelfishpg_tsql.sql
@@ -3230,6 +3230,9 @@ LANGUAGE pltsql STABLE;
 -- catalogs, so we have to use some trickery to make things work.
 -- Not all datatypes are handled as well as might be possible, but it is sufficient for 
 -- the current purposes.
+-- Note that this proc may increase the response time for the first execution of sp_who, but 
+-- we're looking at prioritizing user-friendliness (easy-to-read output) here. Also, sp_who 
+-- is very unlikely to be part of performance-critical workload.
 CREATE OR REPLACE PROCEDURE sys.sp_babelfish_autoformat(
 	IN "@tab"        VARCHAR(257) DEFAULT NULL,
 	IN "@orderby"    VARCHAR(1000) DEFAULT '',
--- a/contrib/babelfishpg_tsql/sql/babelfishpg_tsql.sql
+++ b/contrib/babelfishpg_tsql/sql/babelfishpg_tsql.sql
@@ -3230,6 +3230,9 @@ LANGUAGE pltsql STABLE;
 -- catalogs, so we have to use some trickery to make things work.
 -- Not all datatypes are handled as well as might be possible, but it is sufficient for 
 -- the current purposes.
+-- Note that this proc may increase the response time for the first execution of sp_who, but 
+-- we're looking at prioritizing user-friendliness (easy-to-read output) here. Also, sp_who 
+-- is very unlikley to be part of performance-critical workload.
 CREATE OR REPLACE PROCEDURE sys.sp_babelfish_autoformat(
 	IN "@tab"        VARCHAR(257) DEFAULT NULL,
 	IN "@orderby"    VARCHAR(1000) DEFAULT '',
--- a/contrib/babelfishpg_tsql/sql/babelfishpg_tsql.sql
+++ b/contrib/babelfishpg_tsql/sql/babelfishpg_tsql.sql
@@ -3222,3 +3222,369 @@ BEGIN
 END;
 $$
 LANGUAGE pltsql STABLE;
+
+-- sp_babelfish_autoformat is a helper procedure which formats the contents of a table (or view)
+-- as narrowly as possible given its actual column contents.
+-- This proc is currently only used by sp_who but could be applied more generically.
+-- A complication is that the metadata from #tmp tables cannot be found in the babelfish
+-- catalogs, so we have to use some trickery to make things work.
+-- Not all datatypes are handled as well as might be possible, but it is sufficient for 
+-- the current purposes.
+CREATE OR REPLACE PROCEDURE sys.sp_babelfish_autoformat(
+	IN "@tab"        VARCHAR(257) DEFAULT NULL,
+	IN "@orderby"    VARCHAR(1000) DEFAULT '',
+	IN "@printrc"    sys.bit DEFAULT 1,
+	IN "@hiddencols" VARCHAR(1000) DEFAULT NULL)
+LANGUAGE 'pltsql'
+AS $$
+BEGIN
+	SET NOCOUNT ON
+	DECLARE @rc INT
+	DECLARE @id INT
+	DECLARE @objtype VARCHAR(2)	
+	DECLARE @msg VARCHAR(200)	
+	
+	IF @tab IS NULL
+	BEGIN
+		RAISERROR('Must specify table name', 16, 1)
+		RETURN		
+	END
+	
+	IF TRIM(@tab) = ''
+	BEGIN
+		RAISERROR('Must specify table name', 16, 1)
+		RETURN		
+	END	
+	
+	-- Since we cannot find #tmp tables in the Babelfish catalogs, we cannot check 
+	-- their existence other than by trying to select from them
+	-- NB: not handling uncommon but valid syntax '<schemaname>.#tmp' for #tmp tables
+	IF SUBSTRING(@tab,1,1) <> '#'
+	BEGIN
+		SET @id = OBJECT_ID(@tab)
+		IF @id IS NULL
+		BEGIN
+			IF SUBSTRING(UPPER(@tab),1,4) = 'DBO.'
+			BEGIN
+				SET @id = OBJECT_ID('SYS.' + SUBSTRING(@tab,5))
+			END
+			IF @id IS NULL
+			BEGIN		
+				SET @msg = 'Table or view '''+@tab+''' not found'
+				RAISERROR(@msg, 16, 1)
+				RETURN		
+			END
+		END
+	END
+	
+	SELECT @objtype = type FROM sys.sysobjects WHERE id = @id
+	IF @objtype NOT IN  ('U', 'S', 'V')
+	BEGIN
+		SET @msg = ''''+@tab+''' is not a table or view'
+		RAISERROR(@msg, 16, 1)
+		RETURN		
+	END
+	
+	-- check for 'ORDER BY', if specified
+	SET @orderby = TRIM(@orderby)
+	IF @orderby <> ''
+	BEGIN
+		IF UPPER(@orderby) NOT LIKE 'ORDER BY%'
+		BEGIN
+			RAISERROR('@orderby parameter must start with ''ORDER BY''', 16, 1)
+			RETURN
+		END
+	END
+	
+	-- columns to hide in final client output
+	-- assuming delimited column names do not contain spaces or commas inside the name
+	WHILE (CHARINDEX(' ,', @hiddencols) > 0) or (CHARINDEX(', ', @hiddencols) > 0)
+	BEGIN
+		SET @hiddencols = UPPER(REPLACE(@hiddencols, ' ,', ','))
+		SET @hiddencols = UPPER(REPLACE(@hiddencols, ', ', ','))
+	END
+	IF LEN(@hiddencols) IS NOT NULL SET @hiddencols = ',' + @hiddencols + ','
+
+	-- Need to use a guaranteed-uniquely named table as intermediate step since we cannot 
+	-- access the metadata in case a #tmp table is passed as argument
+	-- But when we copy the #tmp table into another table, we do get all the attributes and metadata
+	DECLARE @tmptab VARCHAR(63) = 'sp_babelfish_autoformat' + REPLACE(NEWID(), '-', '')
+	DECLARE @tmptab2 VARCHAR(63) = 'sp_babelfish_autoformat' + REPLACE(NEWID(), '-', '')
+	DECLARE @cmd VARCHAR(1000) = 'SELECT * INTO ' + @tmptab + ' FROM ' + @tab
+	EXECUTE(@cmd)
+
+	-- Get the columns
+	SELECT 
+	   c.name AS colname, c.colid AS colid, t.name AS basetype, 0 AS maxlen
+	INTO #sp_bbf_autoformat
+	FROM sys.syscolumns c left join sys.systypes t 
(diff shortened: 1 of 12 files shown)


contrib/babelfishpg_tsql/sql/babelfishpg_tsql.sql
```diff
@@ -3222,3 +3222,372 @@ BEGIN
...
+	IF TRIM(@tab) = ''
+	BEGIN
+		RAISERROR('Must specify table name', 16, 1)
+		RETURN		
+	END	
+	
+	-- Since we cannot find #tmp tables in the Babelfish catalogs, we cannot check 
+	-- their existence other than by trying to select from them
+	-- NB: not handling uncommon but valid syntax '<schemaname>.#tmp' for #tmp tables
+	IF SUBSTRING(@tab,1,1) <> '#'
```
surendravishnoi, line 3268: sys.babelfish_get_enr_list() this view lists oid and names for objects in ENR (table variables and temp tables), may be it can help

robverschoor: Function sys.babelfish_get_enr_list() could be used to determine if a #tmp table exists but the columns and datatypes can still not be retrieved, so it would be of little use here. I added a comment to the code to mention this.
In case a #tmp table is specified which does not exist, the current code raises the same error message as we would be able to raise when using  sys.babelfish_get_enr_list() .

contrib/babelfishpg_tsql/sql/babelfishpg_tsql.sql
```diff
@@ -3222,3 +3222,375 @@ BEGIN
...
+		PRINT @cmd
+	END
+	
+	-- Cleanup
+	EXECUTE('DROP TABLE ' + @tmptab)
+	EXECUTE('DROP TABLE ' + @tmptab2)
+	RETURN
+END
+$$;
+GRANT EXECUTE ON PROCEDURE sys.sp_babelfish_autoformat(IN VARCHAR(257), IN VARCHAR(1000), sys.bit, VARCHAR(1000)) TO PUBLIC;
```
surendravishnoi, line 3401: is this intended? sp_babelfish_autoformat being accessible to everyone?

surendravishnoi (approved):

shalinilohia50 (approved):

shalinilohia50:
Please fix the JDBC tests. They are failing.

contrib/babelfishpg_tsql/sql/babelfishpg_tsql.sql
robverschoor: Yes, this is intentional. It's a convenient way to display the contents of any table or view in sqlcmd (try 'sysobjects'). The added value of this proc is just the formatting, it does not provide access to data that would otherwise not be accessible, so any user can benefit from this.

contrib/babelfishpg_tsql/sql/babelfishpg_tsql.sql
```diff
@@ -3222,3 +3222,375 @@ BEGIN
...
+-- sp_babelfish_autoformat is a helper procedure which formats the contents of a table (or view)
+-- as narrowly as possible given its actual column contents.
+-- This proc is currently only used by sp_who but could be applied more generically.
+-- A complication is that the metadata from #tmp tables cannot be found in the babelfish
+-- catalogs, so we have to use some trickery to make things work.
+-- Not all datatypes are handled as well as might be possible, but it is sufficient for 
+-- the current purposes.
+-- Note that this proc may increase the response time for the first execution of sp_who, but 
+-- we're looking at prioritizing user-friendliness (easy-to-read output) here. Also, sp_who 
+-- is very unlikely to be part of performance-critical workload.
```
KushaalShroff, line 3235: This has scope to be rewritten in C because the implementation seems lengthy for SQL

contrib/babelfishpg_tsql/sql/babelfishpg_tsql.sql
```diff
@@ -3222,3 +3222,375 @@ BEGIN
...
+CREATE OR REPLACE PROCEDURE sys.sp_babelfish_autoformat(
+	IN "@tab"        VARCHAR(257) DEFAULT NULL,
+	IN "@orderby"    VARCHAR(1000) DEFAULT '',
+	IN "@printrc"    sys.bit DEFAULT 1,
+	IN "@hiddencols" VARCHAR(1000) DEFAULT NULL)
```
KushaalShroff, line 3240: Since this is customer visible/usable shouldnt the argument type be sys.VARCHAR?

contrib/babelfishpg_tsql/sql/babelfishpg_tsql.sql
```diff
@@ -3222,3 +3222,375 @@ BEGIN
...
+	
+	-- Since we cannot find #tmp tables in the Babelfish catalogs, we cannot check 
+	-- their existence other than by trying to select from them
+	-- Function sys.babelfish_get_enr_list() could be used to determine if a #tmp table
+	-- exists but the columns and datatypes can still not be retrieved, so it would be of 
+	-- little use here. 
+	-- NB: not handling uncommon but valid syntax '<schemaname>.#tmp' for #tmp tables
+	IF SUBSTRING(@tab,1,1) <> '#'
+	BEGIN
+		SET @id = OBJECT_ID(@tab)
```
KushaalShroff, line 3270: We should add proper TSQL schema qualifiers to objects (in this case 'sys') since if a user wants to execute this procedure from PSQL they might get an error that the functions does not exist if 'sys' is not in the search_path

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--3.2.0--3.3.0.sql
```diff
@@ -424,6 +424,375 @@ END
...
+	END
+	
+	-- columns to hide in final client output
+	-- assuming delimited column names do not contain spaces or commas inside the name
+	WHILE (CHARINDEX(' ,', @hiddencols) > 0) or (CHARINDEX(', ', @hiddencols) > 0)
+	BEGIN
+		SET @hiddencols = UPPER(REPLACE(@hiddencols, ' ,', ','))
+		SET @hiddencols = UPPER(REPLACE(@hiddencols, ', ', ','))
+	END
+	IF LEN(@hiddencols) IS NOT NULL SET @hiddencols = ',' + @hiddencols + ','
```
KushaalShroff, line 507: Can we do an upper and trim at the very beginning of the functions and for other variables we can do it once at time of initialisation to avoid doing it repeatedly.

Also, can you explain why we need to use UPPER here? Is it just for consistency in string comparison functions?

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--3.2.0--3.3.0.sql
```diff
@@ -424,6 +424,375 @@ END
...
+		spid, 
+		MAX(blocked) AS blocked, 
+		0 AS ecid, 
+		CAST('' AS VARCHAR(100)) AS status,
+		CAST('' AS VARCHAR(100)) AS loginname,
+		CAST('' AS VARCHAR(100)) AS hostname,
+		0 AS dbid,
+		CAST('' AS VARCHAR(100)) AS cmd,
+		0 AS request_id,
+		CAST('TDS' AS VARCHAR(20)) AS connection
```
KushaalShroff, line 672: We need to use sys.varchar

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--3.2.0--3.3.0.sql
```diff
@@ -424,6 +424,375 @@ END
...
+	IF @printrc = 1
+	BEGIN
+		PRINT '   '
+		SET @cmd = '(' + CAST(@rc AS VARCHAR) + ' rows affected)'
+		PRINT @cmd
+	END
+	
+	-- Cleanup
+	EXECUTE('DROP TABLE ' + @tmptab)
+	EXECUTE('DROP TABLE ' + @tmptab2)
```
KushaalShroff, line 592: Should we be executing the entire logic inside a try/catch? Incase of unwanted/unknown errors before dropping, the tables could be left stale

KushaalShroff (changes requested):

contrib/babelfishpg_tsql/sql/babelfishpg_tsql.sql
robverschoor: Agree. But that is not urgent since this is intended as a formatting feature for ease of use.

contrib/babelfishpg_tsql/sql/babelfishpg_tsql.sql
robverschoor: VARCHAR and sys.VARCHAR are not used consistently in this source file. For example, sp_helpdb and and sp_getapplock (and others) also uses VARCHAR (without sys.) and I took these as guidance since these procedures have existed since v.1.0. However sys.VARCHAR is also used in other objects so doing some code cleanup would make sense but that is an action beyond the scope of this PR.

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--3.2.0--3.3.0.sql
robverschoor: see earlier comment

contrib/babelfishpg_tsql/sql/babelfishpg_tsql.sql
robverschoor: Since the dialect is T-SQL, wouldn't that imply that sys is in the search path?

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--3.2.0--3.3.0.sql
robverschoor: 1: TRIM() won't do the job: we're removing an arbitrary number of spaces inside a string here, either before or after a comma. This has to use a loop.
2: UPPER() is used just for defensive coding. It could indeed be moved outside the loop.

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--3.2.0--3.3.0.sql
robverschoor: I guess we could put SET XACT_ABORT OFF at the top of the proc, so that execution still reaches the end of the procedure (where the DROP stmts are) in case of errors. Only an error that would abort the proc execution would miss those DROPs, but I think that is also true for TRY-CATCH? 
If the column metadata for #tmp tables would be accessible, we could (and should) use #tmp tables instead of these uniquely-named permanent ones. 
Note that the tables are named such that their purpose is clear so their accidental existence should not be too confusing. One thing we could do is add a timestamp to the table name and drop any such tables older than, say, 24 hours. But I wonder if all that is worth it.

contrib/babelfishpg_tsql/sql/babelfishpg_tsql.sql
KushaalShroff: Earlier we did not follow proper guidelines in creating and declaring objects, now with new objects or modifications to the current objects, We are making sure to use sys.VARCHAR instead of VARCHAR

contrib/babelfishpg_tsql/sql/babelfishpg_tsql.sql
KushaalShroff: From pg endoint we do not add sys to search path even on changing the dialect:
```
demo=# set babelfishpg_tsql.sql_dialect = 'TSQL';
SET
demo=# show babelfishpg_tsql.sql_dialect;
 babelfishpg_tsql.sql_dialect 
------------------------------
 tsql
(1 row)

demo=# show search_path;
   search_path   
-----------------
 "$user", public
(1 row)
```
We will need explicit sys qualifiers for every TSQL object referenced

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--3.2.0--3.3.0.sql
KushaalShroff: Defensive coding might lead to performance impact. Taking the UPPER out of the loop might be a better choice.

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--3.2.0--3.3.0.sql
KushaalShroff: `Only an error that would abort the proc execution would miss those DROPs, but I think that is also true for TRY-CATCH`: Any error aborting in the body of the try will take the control flow to the catch block.

`One thing we could do is add a timestamp to the table name and drop any such tables older than, say, 24 hours. But I wonder if all that is worth it.` : 
Drawbacks: 1. Might be a potential to Bloating the storage. 2. We would need to run some cleanup

PR #1646, opened 2023-07-25
Support for DBCC CHECKIDENT

aadityavermaa:
### Description

This changes adds support for DBCC CHECKIDENT command in Babelfish that allows user to reset the current value of the identity column in a table. Earlier user had to do the same with Setval() command from pg end-point, with this change they can do it directly from T-SQL end-point. As this is the first time a DBCC command is implemented in Babelfish, this change also provides support to syntax for DBCC commands, and the design is kept scalable so that new DBCC commands can be better introduced in future. 

### Issues Resolved
BABEL-3201

* **Use case based -**
Yes

* **Boundary conditions -**
Yes

* **Arbitrary inputs -**
Yes

* **Negative test cases -**
Yes

N/A

N/A

* **Performance tests -**
N/A

* **Tooling impact -**
N/A

* **Client tests -**
N/A

### Check List


contrib/babelfishpg_tsql/src/iterative_exec.c
```diff
@@ -828,6 +828,15 @@ dispatch_stmt(PLtsql_execstate *estate, PLtsql_stmt *stmt)
 			}
 			exec_stmt_restore_ctx_partial(estate, (PLtsql_stmt_restore_ctx_partial *) stmt);
 			break;
+		case PLTSQL_STMT_DBCC:
+			if (pltsql_explain_only)
+				{
+				ereport(ERROR,
+						(errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
+						 errmsg("Showing Estimated Execution Plan for DBCC statment is not yet supported")));
+				}
```
kuntalghosh: DBCC is utility command. It doesn't make sense to show an execution plan for the same. What's the SQL Server behavior?

contrib/babelfishpg_tsql/src/pl_funcs-2.c
```diff
@@ -717,6 +723,12 @@ dump_stmt_insert_bulk(PLtsql_stmt_insert_bulk *stmt_insert_bulk)
 	printf("INSERT BULK %s\n", stmt_insert_bulk->table_name);
 }
 
+void
+dump_stmt_dbcc(PLtsql_stmt_dbcc *stmt_dbcc)
```
kuntalghosh, line 689: We should also mention the type of the DBCC statement being dumped.

contrib/babelfishpg_tsql/src/tsqlUnsupportedFeatureHandler.cpp
```diff
@@ -1239,6 +1239,29 @@ antlrcpp::Any TsqlUnsupportedFeatureHandlerImpl::visitTransaction_statement(TSql
 	return visitChildren(ctx);
 }
 
+antlrcpp::Any TsqlUnsupportedFeatureHandlerImpl::visitDbcc_statement(TSqlParser::Dbcc_statementContext *ctx)
+{
+
+	if (ctx->dbcc_command())
```
kuntalghosh, line 1261: We should add some telemetry for DBCC statements. Can be addressed separately.

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -2955,6 +2959,225 @@ exec_stmt_insert_bulk(PLtsql_execstate *estate, PLtsql_stmt_insert_bulk *stmt)
 	return PLTSQL_RC_OK;
 }
 
+int exec_stmt_dbcc(PLtsql_execstate *estate, PLtsql_stmt_dbcc *stmt)
+{
+	switch (stmt->dbcc_stmt_type)
```
kuntalghosh, line 3124: If we don't expect the function to be called for other dbcc statement type, we should add an assert(0) for the default case.
```
case default:
		Assert(0);
  
```

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -2955,6 +2959,225 @@ exec_stmt_insert_bulk(PLtsql_execstate *estate, PLtsql_stmt_insert_bulk *stmt)
...
+	table_oid = get_relname_relid(dbcc_stmt.table_name, nsp_oid);
+	if(!OidIsValid(table_oid))
+	{
+		ereport(ERROR,
+		(errcode(ERRCODE_UNDEFINED_TABLE),
+			errmsg("relation \"%s\" does not exist", dbcc_stmt.table_name)));
+	}
+
+	/* Permission check */
+	if (pg_class_aclcheck(table_oid, GetUserId(), ACL_SELECT | ACL_USAGE) != ACLCHECK_OK)
```
kuntalghosh: Does this check also cover the sysadmin and dbowner permission?

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -2955,6 +2959,225 @@ exec_stmt_insert_bulk(PLtsql_execstate *estate, PLtsql_stmt_insert_bulk *stmt)
...
+			errmsg("'%s.%s' does not contain an identity column.",
+				nsp_name, dbcc_stmt.table_name)));	
+
+	pfree(nsp_name);
+
+	/*
+	 * We need to find out the last value of the identity column in the table
+	 * and last value of the identity sequence to compare and decide on the
+	 * action that needs to be taken.
+	 */
```
kuntalghosh: It's not clear what CHECKIDENT behavior we're trying to implement here. Also, we shouldn't do the relation scan if customer has already provided the new seed value.

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -2955,6 +2959,225 @@ exec_stmt_insert_bulk(PLtsql_execstate *estate, PLtsql_stmt_insert_bulk *stmt)
...
+	scan = table_beginscan_catalog(rel, 0, NULL);
```
kuntalghosh: Why are we using catalog scan API to scan an user relation?

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -2955,6 +2959,225 @@ exec_stmt_insert_bulk(PLtsql_execstate *estate, PLtsql_stmt_insert_bulk *stmt)
...
+	tuple = heap_getnext(scan, BackwardScanDirection);
+
+	if (HeapTupleIsValid(tuple))
+	{
+		max_identity_value = DatumGetInt64(heap_getattr(tuple, attnum+1,
```
kuntalghosh: The last value need not be the max value. We should execute the select max(id) from table query and fetch the result.

test/JDBC/input/BABEL-3201-vu-verify.sql
```diff
@@ -0,0 +1,35 @@
+DBCC CHECKIDENT(babel_3201_t1);
```
kuntalghosh: Please try to add at least one test case for all the errors these changes are introducing.

kuntalghosh (changes requested):

test/JDBC/expected/BABEL-3201-vu-verify.out
```diff
@@ -0,0 +1,59 @@
+DBCC CHECKIDENT(babel_3201_t1);
```
surendravishnoi: Please also add some tests to validate transactional behaviour.
Running DBCC inside a transaction
DBCC running into error inside transaction
Calling DBCC from procedure/functions (with/without errors)

contrib/babelfishpg_tsql/src/iterative_exec.c
aadityavermaa: Executing DBCC CHECKIDENT after SET SHOWPLAN_ALL ON in SQL Server does not throw and error but most of the fields come out to be NULL, I have removed the condition for now.

contrib/babelfishpg_tsql/src/pl_exec-2.c
aadityavermaa: Made changes in the latest commits, Thanks.

contrib/babelfishpg_tsql/src/pl_exec-2.c
aadityavermaa: Updated the comments and finding the max value of the column where it's required ( not when new_reseed_value ) is already provided.

contrib/babelfishpg_tsql/src/pl_exec-2.c
aadityavermaa: Since we have to use MAX() to find out the maximum value, I have removed the catalog scan API.

contrib/babelfishpg_tsql/src/pl_exec-2.c
aadityavermaa: Made the change in the latest commit. Thanks.

test/JDBC/input/BABEL-3201-vu-verify.sql
aadityavermaa: I have added test cases for all the functional scenarios, corner cases ( no rows have been inserted in table, rows are removed using TRUNCATE TABLE, rows are deleted using DELETE TABLE), and negative tests for syntax and datatype bounds.

test/JDBC/expected/BABEL-3201-vu-verify.out
aadityavermaa: Added tests for transactions and procedures. Please take a look. Thanks.

contrib/babelfishpg_tsql/src/iterative_exec.c
kuntalghosh: Sorry, which condition exactly have you removed? What's the Babelfish behavior with your changes?

aadityavermaa: Apologies, I just saw that I hadn't included this file in the commit, I meant I have removed the pltsql_explain_only condition for now.

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -3112,6 +3119,297 @@ exec_stmt_insert_bulk(PLtsql_execstate *estate, PLtsql_stmt_insert_bulk *stmt)
...
+
+int exec_stmt_dbcc(PLtsql_execstate *estate, PLtsql_stmt_dbcc *stmt)
+{
+	switch (stmt->dbcc_stmt_type)
+	{
+		case PLTSQL_DBCC_CHECKIDENT:
+			exec_stmt_dbcc_checkident(stmt);
+			break;
+		default:
+			Assert(0);
```
kuntalghosh, line 3130: What is the BBF behavior for other DBCC statement types? If we're not throwing an error already, we can't put this assert here.

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -3112,6 +3119,297 @@ exec_stmt_insert_bulk(PLtsql_execstate *estate, PLtsql_stmt_insert_bulk *stmt)
...
+	volatile bool cur_value_is_null = true;
+	bool	login_is_db_owner;
+	char		message[200];
+
+
+    bool    is_cross_db = false;
+	if(dbcc_stmt.new_reseed_value)
+	{
+		/* If float value is passed as reseed_value, only decimal part is considered.*/
+		token = strtok(dbcc_stmt.new_reseed_value, ".");
```
kuntalghosh: Do we have a testcase for this?

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -3112,6 +3119,297 @@ exec_stmt_insert_bulk(PLtsql_execstate *estate, PLtsql_stmt_insert_bulk *stmt)
...
+			reseed_value = floor(reseed_value);
+		else
+			reseed_value = -floor(-reseed_value);
+		pfree(token);
+	}
+
+    db_name = get_cur_db_name();
+    if (dbcc_stmt.db_name)
+    {
+        if (pg_strcasecmp(db_name, dbcc_stmt.db_name) != 0)
```
kuntalghosh: Let's use pg_strncase and provide NAMEDATALEN as length

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -3112,6 +3119,297 @@ exec_stmt_insert_bulk(PLtsql_execstate *estate, PLtsql_stmt_insert_bulk *stmt)
...
+
+		if (attr->attidentity)
+		{
+			attname = NameStr(attr->attname);
+			seqid = getIdentitySequence(table_oid, attnum + 1, false);
+			break;
+		}
+	}
+
+	RelationClose(rel);
```
kuntalghosh, line 3304: We already have helper function for this. Let's use the same - get_table_identity().

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -3112,6 +3119,297 @@ exec_stmt_insert_bulk(PLtsql_execstate *estate, PLtsql_stmt_insert_bulk *stmt)
...
+	if (!OidIsValid(seqid))
+		ereport(ERROR,
+		(errcode(ERRCODE_UNDEFINED_COLUMN),
+			errmsg("'%s.%s' does not contain an identity column.",
+				nsp_name, table_name)));
+
+	PG_TRY();
+	{
+		// Lock table to ensure concurrency.
+		if(dbcc_stmt.is_reseed)
```
kuntalghosh: Please add a comment why this is needed

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -3112,6 +3119,297 @@ exec_stmt_insert_bulk(PLtsql_execstate *estate, PLtsql_stmt_insert_bulk *stmt)
...
+
+				DirectFunctionCall3(setval3_oid,
+					ObjectIdGetDatum(seqid),
+					Int64GetDatum(reseed_value),
+					BoolGetDatum(false));
+			}
+			else
+			{
+				snprintf(message, sizeof(message), "Checking identity information: current identity value 'NULL', current column value 'NULL'.\n");
+			}
```
kuntalghosh, line 3363: Do we have test cases for these two scenarios?

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -3112,6 +3119,297 @@ exec_stmt_insert_bulk(PLtsql_execstate *estate, PLtsql_stmt_insert_bulk *stmt)
...
+	PG_TRY();
+	{
+		// Lock table to ensure concurrency.
+		if(dbcc_stmt.is_reseed)
+		{
+			LockRelationOid( table_oid, AccessExclusiveLock);
+		}
+		else
+		{
+			LockRelationOid( table_oid, ShareLock);
```
kuntalghosh: Please add a concurrency test for this scenario where other backends are blocked from accessing this table while a DBCC CHECKIDENT has reseeded the identity column. You can use isolation test framework.

test/JDBC/input/BABEL-3201-vu-verify.mix
```diff
@@ -0,0 +1,247 @@
...
+-- current identity value should be NULL (TO-DO), identity value inserted in
+-- next INSERT operation should be new_reseed_value.
+DBCC CHECKIDENT(babel_3201_t1, RESEED, 5);
+GO
+
+INSERT INTO babel_3201_t_int VALUES (5);
+GO
+
+SELECT * FROM babel_3201_t_int;
+GO
```
kuntalghosh: The tests don't match the comment. The table is not empty

test/JDBC/input/BABEL-3201-vu-verify.mix
```diff
@@ -0,0 +1,247 @@
...
+
+-- Remove all rows in table using TRUNCATE TABLE, identity value inserted in
+-- next INSERT operation should be new_reseed_value.
+TRUNCATE TABLE babel_3201_t_int;
+GO
+
+DBCC CHECKIDENT(babel_3201_t_int, RESEED, 10);
```
kuntalghosh, line 122: Also, add a test with noreseed.

test/JDBC/input/BABEL-3201-vu-verify.mix
```diff
@@ -0,0 +1,247 @@
...
+SELECT * FROM babel_3201_t_decimal;
+GO
+
+BEGIN TRAN;
+    DBCC CHECKIDENT(babel_3201_t_decimal, RESEED, 20);
+    INSERT INTO babel_3201_t_decimal VALUES (11);
+ROLLBACK;
+GO
+
+SELECT * FROM babel_3201_t_decimal;
```
kuntalghosh, line 294: For ROLLBACK, we need to check what's the current identity value? In PG, it'll not be rolled-back - different from SQL Server.

kuntalghosh (dismissed):

contrib/babelfishpg_tsql/src/pl_exec-2.c
skumawat2025: We are throwing error already in the `tsqlUnsupportedFeatureHandler.cpp` for other dbcc statements. 
[Link to Code](https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/1646/files#diff-da105a520d3a543e68071851f0d373f13f9153b9d416627c00a950ed46e81151:~:text=throw%20PGErrorWrapperException(ERROR%2C%20ERRCODE_FEATURE_NOT_SUPPORTED,getLineAndPos(ctx%2D%3Edbcc_command()))%3B)

contrib/babelfishpg_tsql/src/pl_exec-2.c
skumawat2025: yes

contrib/babelfishpg_tsql/src/pl_exec-2.c
skumawat2025: Done.

contrib/babelfishpg_tsql/src/pl_exec-2.c
skumawat2025: Although we can use the `get_table_identity()` api to find `seqid` but we will also need the name of this sequence  col  that is `attname` in the code  to find maximum value. See line no. 3361.

contrib/babelfishpg_tsql/src/pl_exec-2.c
skumawat2025: Fixed.

contrib/babelfishpg_tsql/src/pl_exec-2.c
skumawat2025: Added.

contrib/babelfishpg_tsql/src/pl_exec-2.c
skumawat2025: added test for concurrency using isolation test framework.

test/JDBC/input/BABEL-3201-vu-verify.mix
skumawat2025: Fixed.

test/JDBC/input/BABEL-3201-vu-verify.mix
skumawat2025: added.

test/JDBC/input/BABEL-3201-vu-verify.mix
skumawat2025: Added a `checkident(<table>, noreseed)` after transection rollback to check the current identity values, which is not rolled back.

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -3112,6 +3119,309 @@ exec_stmt_insert_bulk(PLtsql_execstate *estate, PLtsql_stmt_insert_bulk *stmt)
...
+		/* If float value is passed as reseed_value, only decimal part is considered.*/
+		token = strtok(dbcc_stmt.new_reseed_value, ".");
+		reseed_value = pg_strtoint64(token);
+		pfree(token);
+	}
+
+    db_name = get_cur_db_name();
+    if (dbcc_stmt.db_name)
+    {
+        if (pg_strncasecmp(db_name, dbcc_stmt.db_name, NAMEDATALEN) != 0)
```
HarshLunagariya: Please add testcase when db_name is more than 64 chars and we are trying to do cross-db call.

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -3112,6 +3119,309 @@ exec_stmt_insert_bulk(PLtsql_execstate *estate, PLtsql_stmt_insert_bulk *stmt)
...
+		}
+		else if ((guest_role_name && strcmp(user, guest_role_name) == 0))
+		{
+			nsp_name = pstrdup(get_guest_schema_name(db_name));
+		}
+		else
+		{
+			schema_name = get_authid_user_ext_schema_name((const char *) db_name, user);
+			nsp_name = get_physical_schema_name(db_name, schema_name);
+		}
```
HarshLunagariya: Missing dbo user handling. Please refer set_search_path_for_user_schema() for more details.

HarshLunagariya (changes requested):

contrib/babelfishpg_tsql/src/tsqlIface.cpp
```diff
@@ -1889,6 +1890,22 @@ class tsqlBuilder : public tsqlCommonMutator
 		clear_rewritten_query_fragment();
 	}
 
+	void enterDbcc_statement(TSqlParser::Dbcc_statementContext *ctx) override
+	{
+		if (ctx->CHECKIDENT())
+        	graft(makeDbccCheckidentStatement(ctx), peekContainer());
```
surendravishnoi: indentation

contrib/babelfishpg_tsql/src/tsqlIface.cpp
```diff
@@ -1889,6 +1890,22 @@ class tsqlBuilder : public tsqlCommonMutator
...
+
+		clear_rewritten_query_fragment();
+		// PLtsql_stmt_execsql *stmt = (PLtsql_stmt_execsql *) getPLtsql_fragment(ctx);
```
surendravishnoi: cleanup if not required

contrib/babelfishpg_tsql/src/tsqlIface.cpp
```diff
@@ -5416,6 +5432,72 @@ makeExecBodyBatch(TSqlParser::Execute_body_batchContext *ctx)
...
+	stmt->cmd_type = PLTSQL_STMT_DBCC;
+	stmt->dbcc_stmt_type = PLTSQL_DBCC_CHECKIDENT;
+	if (ctx->table_name())
+	{
+		if (ctx->table_name()->database)
+			db_name = stripQuoteFromId(ctx->table_name()->database);
+		if (ctx->table_name()->schema)
+			schema_name = stripQuoteFromId(ctx->table_name()->schema);
+		if (ctx->table_name()->table)
+		table_name = stripQuoteFromId(ctx->table_name()->table);
```
surendravishnoi: indentaion

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -3112,6 +3119,309 @@ exec_stmt_insert_bulk(PLtsql_execstate *estate, PLtsql_stmt_insert_bulk *stmt)
...
+	char	*nsp_name = NULL;
+	char	*table_name;
+	char	*query;
+	char	*attname;
+	const char	*user;
+	const char	*guest_role_name;
+	int64	max_identity_value = 0;
+	int64	cur_identity_value = 0;
+	char	*max_identity_value_str = NULL;
+    char	*login = GetUserNameFromId(GetSessionUserId(), false);
```
surendravishnoi: indentation

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -3112,6 +3119,309 @@ exec_stmt_insert_bulk(PLtsql_execstate *estate, PLtsql_stmt_insert_bulk *stmt)
...
+	Oid		nsp_oid;
+	Oid 	table_oid;
+	Oid		seqid = InvalidOid;
+	Oid		current_user_id = GetUserId();
+	volatile bool cur_value_is_null = true;
+	bool	login_is_db_owner;
+	char		message[200];
+
+
+    bool    is_cross_db = false;
```
surendravishnoi: indentation

contrib/babelfishpg_tsql/src/pl_exec-2.c
surendravishnoi: can we handle cases like ".123", is it a valid use case?

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -3112,6 +3119,309 @@ exec_stmt_insert_bulk(PLtsql_execstate *estate, PLtsql_stmt_insert_bulk *stmt)
...
+	if(dbcc_stmt.new_reseed_value)
+	{
+		/* If float value is passed as reseed_value, only decimal part is considered.*/
```
surendravishnoi: /* comment */

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -3112,6 +3119,309 @@ exec_stmt_insert_bulk(PLtsql_execstate *estate, PLtsql_stmt_insert_bulk *stmt)
...
+		reseed_value = pg_strtoint64(token);
+		pfree(token);
+	}
+
+    db_name = get_cur_db_name();
+    if (dbcc_stmt.db_name)
+    {
+        if (pg_strncasecmp(db_name, dbcc_stmt.db_name, NAMEDATALEN) != 0)
+            is_cross_db = true;
+        db_name = pstrdup(dbcc_stmt.db_name);
```
surendravishnoi: do we need to free earlier db_name?

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -3112,6 +3119,309 @@ exec_stmt_insert_bulk(PLtsql_execstate *estate, PLtsql_stmt_insert_bulk *stmt)
...
+    }
+
+	if (!DbidIsValid(get_db_id(db_name)))
```
surendravishnoi: indentation is messed up at number of places

test/JDBC/expected/BABEL-3201-vu-prepare.out
```diff
@@ -0,0 +1,146 @@
...
+GO
+~~ROW COUNT: 1~~
+
+
+INSERT INTO babel_3201_t_decimal VALUES (7);
+GO
+~~ROW COUNT: 1~~
+
+
+CREATE PROCEDURE babel_3201_proc1
```
surendravishnoi, line 146: can we call DBCC from a function/view?

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -3112,6 +3119,309 @@ exec_stmt_insert_bulk(PLtsql_execstate *estate, PLtsql_stmt_insert_bulk *stmt)
...
+										get_owner_of_db(db_name), NAMEDATALEN);
+	user = get_user_for_database(db_name);
+	guest_role_name = get_guest_role_name(db_name);
+
+    if(is_cross_db)
+    {
+        if (user)
+            SetCurrentRoleId(GetSessionUserId(), false);
+        else
+            ereport(ERROR,
```
surendravishnoi: please add a test case for this error if not already

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -3112,6 +3119,309 @@ exec_stmt_insert_bulk(PLtsql_execstate *estate, PLtsql_stmt_insert_bulk *stmt)
...
+	}
+	else
+	{
+		/* 
+		 * If schema_name is not provided, find default schema for current user
+		 * and get physical schema name
+		 */
+		if (!user)
+		{
+			ereport(ERROR,
```
surendravishnoi: please add a test case

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -3112,6 +3119,309 @@ exec_stmt_insert_bulk(PLtsql_execstate *estate, PLtsql_stmt_insert_bulk *stmt)
...
+				{
+					DirectFunctionCall2(setval_oid,
+						ObjectIdGetDatum(seqid),
+						Int64GetDatum(max_identity_value));
+				}
+			}
+		}
+		
+		if (is_cross_db)
+            SetCurrentRoleId(current_user_id, false);
```
surendravishnoi: identation

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -3112,6 +3119,309 @@ exec_stmt_insert_bulk(PLtsql_execstate *estate, PLtsql_stmt_insert_bulk *stmt)
...
+	}
+	PG_CATCH();
+	{
+		PG_RE_THROW();
+	    if (is_cross_db)
+            SetCurrentRoleId(current_user_id, false);
```
surendravishnoi: control will never come here, we should first do set and then rethrow

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -3112,6 +3119,309 @@ exec_stmt_insert_bulk(PLtsql_execstate *estate, PLtsql_stmt_insert_bulk *stmt)
...
+				}
+			}
+		}
+		
+		if (is_cross_db)
+            SetCurrentRoleId(current_user_id, false);
+	}
+	PG_CATCH();
+	{
+		PG_RE_THROW();
```
surendravishnoi, line 3443: who will release lock in case of errors?

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -3112,6 +3119,309 @@ exec_stmt_insert_bulk(PLtsql_execstate *estate, PLtsql_stmt_insert_bulk *stmt)
...
+	
+	if (max_identity_value_str)
+		pfree(max_identity_value_str);
+	pfree(schema_name);
+	pfree(table_name);
+	pfree(nsp_name);
+
+	if(!dbcc_stmt.new_reseed_value)
+	{
+		UnlockRelationOid(table_oid, ShareLock);
```
surendravishnoi, line 3460: what about exclusive lock?

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -3112,6 +3119,309 @@ exec_stmt_insert_bulk(PLtsql_execstate *estate, PLtsql_stmt_insert_bulk *stmt)
...
+	}
+	PG_CATCH();
+	{
+		PG_RE_THROW();
+	    if (is_cross_db)
+            SetCurrentRoleId(current_user_id, false);
+	}
+	PG_END_TRY();
+	
+	if (max_identity_value_str)
```
surendravishnoi, line 3449: we want to free memory for error cases as well

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -3112,6 +3119,309 @@ exec_stmt_insert_bulk(PLtsql_execstate *estate, PLtsql_stmt_insert_bulk *stmt)
...
+						ObjectIdGetDatum(seqid),
+						Int64GetDatum(max_identity_value));
+				}
+			}
+		}
+		
+		if (is_cross_db)
+            SetCurrentRoleId(current_user_id, false);
+	}
+	PG_CATCH();
```
surendravishnoi, line 3422: do we have test cases with different error and lock types?

contrib/babelfishpg_tsql/src/pl_exec-2.c
skumawat2025: as per t-sql its is valid case. with value .123 and -.123 it will treat it as 0 value. Made changes in the latest commit.

contrib/babelfishpg_tsql/src/pl_exec-2.c
skumawat2025: So, we are trying to hold this shareLock on table only for the execution of this statement. We try to hold Exclusive Lock on a transaction level. We want to release this share lock as soon as we are done with the execution of DBCC CHECKIDENT() stmt. But Exclusive Lock will get release when transection gets commited or rollbacked.

contrib/babelfishpg_tsql/src/pl_exec-2.c
skumawat2025: done. made changes in the next commit. please take a look.

contrib/babelfishpg_tsql/src/pl_exec-2.c
skumawat2025: So there are test-cases in isolation test framwork, when one transection made update to table t1 (by reseeding identity value via dbcc checkident) and other transection is not able to use t1 inside a dbcc checkident stmt.

contrib/babelfishpg_tsql/src/pl_exec-2.c
skumawat2025: made changes to release locks in case of errors.

contrib/babelfishpg_tsql/src/pl_exec-2.c
skumawat2025: made changes. please have a look.

contrib/babelfishpg_tsql/src/pl_exec-2.c
skumawat2025: added.

contrib/babelfishpg_tsql/src/pl_exec-2.c
skumawat2025: made changes.

contrib/babelfishpg_tsql/src/pl_exec-2.c
skumawat2025: made changes. Please have a look.

contrib/babelfishpg_tsql/src/pl_exec-2.c
skumawat2025: added. please have a look.

test/JDBC/expected/BABEL-3201-vu-prepare.out
skumawat2025: We can call DBCC inside a transaction and proc but not inside a function. test-cases for proc/tran are there.

contrib/babelfishpg_tsql/src/pl_exec-2.c
skumawat2025: I think adding test-case for this error is not possible. `get_user_for_database(db_name)` will check if the login is the db owner, or login is member of sysadmin, or guest is enabled in login. Otherwise it will return NULL. This case we have covered under corss_db calls. And in cross-db calls when user is NULL we will throw error `The server principal \"%s\" is not able to access the database \"%s\" under the current security context`

contrib/babelfishpg_tsql/antlr/TSqlParser.g4
```diff
@@ -4271,15 +4305,21 @@ keyword
     | CHANGETABLE
     | CHANGE_RETENTION
     | CHANGE_TRACKING
+    | CHECKDB
+    | CHECKFILEGROUP
+    | CHECKIDENT
```
HarshLunagariya: Do we still need these changes ? since in the dbcc_command rule we have already explicitly specified them.

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -19,6 +20,8 @@
 #include "pl_explain.h"
 #include "session.h"
 
+#include <math.h>
```
HarshLunagariya: Is it needed?
I see we are using pg_strtoint64 which is there 'utils/builtins.h' which is indirectly included by some other header.

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -3113,6 +3120,347 @@ exec_stmt_insert_bulk(PLtsql_execstate *estate, PLtsql_stmt_insert_bulk *stmt)
...
+			nsp_name = get_dbo_schema_name(db_name);
+		}
+		else if ((guest_role_name && strcmp(user, guest_role_name) == 0))
+		{
+			nsp_name = get_guest_schema_name(db_name);
+		}
+		else
+		{
+			nsp_name = get_physical_schema_name(db_name, dbcc_stmt.schema_name);
+		}
```
HarshLunagariya: This is incorrect. Here dbcc_stmt will be NULL We are missing coverage for this case otherwise it would have failed.
It should be schema_name instead. Move schema_name evaluation inside this condition.

contrib/babelfishpg_tsql/antlr/TSqlParser.g4
skumawat2025: Correct. Removing.

contrib/babelfishpg_tsql/src/pl_exec-2.c
skumawat2025: Yes. Removing it .

contrib/babelfishpg_tsql/src/pl_exec-2.c
skumawat2025: Thanks for pointing this out. Fixing it and also adding a test-case for the coverage. However as we do need schema_ name later ahead in the code, keeping schema_name evaluation out of this else condition

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -3113,6 +3118,347 @@ exec_stmt_insert_bulk(PLtsql_execstate *estate, PLtsql_stmt_insert_bulk *stmt)
...
+	bool    is_cross_db = false;
+
+	if(dbcc_stmt.new_reseed_value)
+	{
+		/* If float value is passed as reseed_value, only part before decimal is considered */
+		is_float_value = strchr(dbcc_stmt.new_reseed_value, '.') != NULL;
+
+		if (is_float_value)
+		{
+			if (dbcc_stmt.new_reseed_value[0] == '.' ||
```
surendravishnoi, line 3175: have we tried input like ".", "-" etc for testing?

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -3113,6 +3118,347 @@ exec_stmt_insert_bulk(PLtsql_execstate *estate, PLtsql_stmt_insert_bulk *stmt)
...
+    	db_name = get_cur_db_name();
+    	if (dbcc_stmt.db_name)
+    	{
+		if (!DbidIsValid(get_db_id(dbcc_stmt.db_name)))
+		{
+			ereport(ERROR,
+			(errcode(ERRCODE_UNDEFINED_DATABASE),
+				errmsg("database \"%s\" does not exist", dbcc_stmt.db_name)));
+		}
+       		if (pg_strncasecmp(db_name, dbcc_stmt.db_name, NAMEDATALEN) != 0)
```
surendravishnoi: make sure we have test cases for mixed case db name and DB name more than 64 chars (if allowed)

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -3113,6 +3118,347 @@ exec_stmt_insert_bulk(PLtsql_execstate *estate, PLtsql_stmt_insert_bulk *stmt)
...
+        	{
+			is_cross_db = true;
+			pfree(db_name);
+        		db_name = pstrdup(dbcc_stmt.db_name);
+		}
+    	}
+
+	user = get_user_for_database(db_name);
+	login_is_db_owner = 0 == strncmp(GetUserNameFromId(GetSessionUserId(), false),
```
surendravishnoi, line 3207: make sure we have test cases for mixed case user name and user name more than 64 chars (if allowed)

contrib/babelfishpg_tsql/src/pl_exec-2.c
surendravishnoi: not clear. can you test without cross db? also, if we cannot write a test case means that code path is not possible. If that is the case, add an assert (at-least for debug)

skumawat2025: No, user can  not be null without cross-db calls.  
For example we can see that api  `get_user_for_database(db_name)` will return NULL when we are trying to do a cross db call and there is no user present for that db for the current login. Adding an assert for debugging.

contrib/babelfishpg_tsql/src/pl_exec-2.c
skumawat2025: tested and added cases.

contrib/babelfishpg_tsql/src/pl_exec-2.c
skumawat2025: There are test cases with db name longer then 64 char, mixed case db name.

contrib/babelfishpg_tsql/src/pl_exec-2.c
skumawat2025: added a test case where user name is longer then 64 chars.

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -3113,6 +3118,359 @@ exec_stmt_insert_bulk(PLtsql_execstate *estate, PLtsql_stmt_insert_bulk *stmt)
...
+		 * If schema_name is not provided, find default schema for current user
+		 * and get physical schema name
+		 */
+		guest_role_name = get_guest_role_name(db_name);
+		dbo_role_name = get_dbo_role_name(db_name);
+		
+		/* user will never be null here as cross-db calls are already handled */
+		Assert(user != NULL);
+
+		schema_name = get_authid_user_ext_schema_name((const char *) db_name, user);
```
HarshLunagariya, line 3245: Please move it inside the third condition, It is not needed for every condition. Otherwise permissions related changes looks good.

surendravishnoi (approved):

HarshLunagariya (commented):
Verified user and permissions related changes. Looks good!

PR #1676, opened 2023-08-02
Support prefix terms in full text search CONTAINS

rl626:
### Description
Currently full text search CONTAINS only supports basic simple terms. This commit implements prefix terms for full text search CONTAINS. Concretely, CONTAINS(col_name, '"word*"') is rewritten into to_tsvector('simple', col_name) @@ to_tsquery('simple', 'word:*'), and CONTAINS(col_name, '"word1 word2 word3*"') is rewritten into to_tsvector('simple', col_name) @@ to_tsquery('simple', 'word1<->word2<->word3:*'). We choose to use the 'simple' full text configuration in Postgres because it gives the closest matching to SQL Server.

### Issues Resolved
Task: BABEL-4245

We tested CONTAINS on a single column with the query keyword being prefix terms as either words or phrases. 

### Check List


contrib/babelfishpg_tsql/sql/fts_contains_pgconfig.sql
```diff
@@ -1,12 +1,20 @@
...
+-- They are the configurations that provide closest matching according to our experiments
 CREATE OR REPLACE FUNCTION sys.babelfish_fts_contains_pgconfig(IN phrase text)
   RETURNS regconfig AS
 $$
 DECLARE
   joined_text text;
   word text;
 BEGIN
+  -- Prefix term
+  IF (phrase COLLATE C) SIMILAR TO ('[ ]*"%\*"[ ]*' COLLATE C) THEN
```
forestkeeper: Could you explain where is this string ''[ ]*"%\*"[ ]*'' comes from ?

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--3.2.0--3.3.0.sql
```diff
@@ -484,17 +485,19 @@ BEGIN
...
     -- Append the word to the joined_text variable
     joined_text := joined_text || word || '<->';
   END LOOP;
 
   -- Remove the trailing "<->" from the joined_text
   joined_text := substring(joined_text COLLATE "C", 1, length(joined_text) - 3) COLLATE "C";
 
+  -- If the last word has the form 'lastword*', then this is a prefix term (the last word cannot be '*')
+  -- We need to rewrite the last word into 'lastword:*'
+  IF ((joined_text COLLATE "C") SIMILAR TO ('%\*' COLLATE C)) AND (NOT (joined_text COLLATE "C") SIMILAR TO ('%<->\*' COLLATE C)) AND (NOT (joined_text COLLATE "C") SIMILAR TO ('\*' COLLATE C)) THEN
```
forestkeeper: Did you aware similar to operator in PG is not the same with regular expression match ? refer : https://www.postgresql.org/docs/current/functions-matching.html

contrib/babelfishpg_tsql/sql/fts_contains_rewrite.sql
```diff
@@ -54,17 +55,19 @@ BEGIN
...
     -- Append the word to the joined_text variable
     joined_text := joined_text || word || '<->';
   END LOOP;
 
   -- Remove the trailing "<->" from the joined_text
   joined_text := substring(joined_text COLLATE "C", 1, length(joined_text) - 3) COLLATE "C";
 
+  -- If the last word has the form 'lastword*', then this is a prefix term (the last word cannot be '*')
+  -- We need to rewrite the last word into 'lastword:*'
+  IF ((joined_text COLLATE "C") SIMILAR TO ('%\*' COLLATE C)) AND (NOT (joined_text COLLATE "C") SIMILAR TO ('%<->\*' COLLATE C)) AND (NOT (joined_text COLLATE "C") SIMILAR TO ('\*' COLLATE C)) THEN
```
forestkeeper: Are we sure all the similar to pattern match has a test cases ?

contrib/babelfishpg_tsql/sql/fts_contains_pgconfig.sql
rl626: SELECT * FROM t WHERE CONTAINS(txt, '"blu*"')
is equivalent to
SELECT * FROM t WHERE CONTAINS(txt, '     "blu*"   ') in SQL Server, presumably because the string constant is parsed by a grammar internally. This means that we can allow spaces in the search term. 
'[ ]*"%\*"[ ]*' says that the search term can has any number of preceding or trailing white spaces. And it should have something surrounded by a pair of double quote. Inside the double quote, it should be any string followed by a * character.

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--3.2.0--3.3.0.sql
rl626: Yes. I'm following the syntax in the documentation

contrib/babelfishpg_tsql/sql/fts_contains_rewrite.sql
rl626: Not really. This is an expedient solution. The ideal way is to parse those search terms by writing parser rules in a parser generator. However, current ANTLR parser/Bison parser do not fit our needs to parse it, because the search term is lexed as a single string constant token by the lexer and there is no way for a parser to look into a single token. To parse the search term we will need to add another parser generator specifically for full text search which requires substantial change to the existing software architecture.

rl626: I just added test cases for the patter match

forestkeeper (approved):
LGTM

PR #1681, opened 2023-08-03
Handle exception gracefully inside JDBC test framework if remote server is crashed

Deepesh125:
Current behaviour of test framework in event of remote server crash is that framework will get stuck without any error or anything. Recently we understood that it may also got stuck in infinite loop of exception which would keep on getting buffered in corresponding test output file. This will keep on bloating output file if java process is not terminated which may in theory can cause out of disk space situation. This commit fixes that issue by identifying such exception and exiting the process.

Task: None

### Check List

### DIFF
--- a/test/JDBC/src/main/java/com/sqlsamples/HandleException.java
+++ b/test/JDBC/src/main/java/com/sqlsamples/HandleException.java
@@ -36,16 +36,23 @@ public class HandleException {
                         errorMsg += "\n  ";
                     }
                     bw.write("~~ERROR (Message: "+ errorMsg + "  Server SQLState: " + e.getSQLState() + ")~~");
-                } else if (e instanceof SQLServerException && "08S01".equals(e.getSQLState())){
-                    // Handling this particular exception where server might have crashed.
-                    System.out.println("Remote server might have crashed.");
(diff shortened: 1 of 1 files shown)


test/JDBC/src/main/java/com/sqlsamples/HandleException.java
```diff
@@ -34,7 +36,16 @@ static void handleSQLExceptionWithFile(SQLException e, BufferedWriter bw, Logger
                         errorMsg += "\n  ";
                     }
                     bw.write("~~ERROR (Message: "+ errorMsg + "  Server SQLState: " + e.getSQLState() + ")~~");
-                } else {
+                } else if (e instanceof SQLServerException && "08S01".equals(e.getSQLState())){
```
thephantomthief: We should also dump something in the output file because this will print only on the console. We can move this logic inside the original else.

Also, should we do something similar for PSQLException?

thephantomthief (approved):

test/JDBC/src/main/java/com/sqlsamples/HandleException.java
Deepesh125: Done

shardgupta (approved):

PR #1704, opened 2023-08-09
Fixed power function to handle smallint, tinyint and null values.

siddhijain2:
### Description

This pull request modifies power function :

**Issue 1 :** To handle the smallint, tinyint and null values correctly.

Currently, power function is throwing error out of range of smallint/tinyint for some values.

```
1> SELECT POWER( CAST( 100 AS SMALLINT ), 3) 
2> GO
Msg 220, Level 16, State 1, Server BABELFISH, Line 1
smallint out of range
```
It is an incorrect result, as the required return type of power function for input value smallint and tinyint is int. And 1000000 is under the range of int. In newer version, it will return the value if it is under the range of int. If the number or power is NULL, then it will return NULL.

**Issue 2 :** To resolve the round off error.

Currently, power function rounds off the value to the nearest integer. 

```
1> SELECT POWER( 2,-1)
2> GO

power      
-----------
          1
```
But the correct result of the above query is 0. This issue is occurring as conversion of numeric to int8/int4 is rounding off the results to nearest integer. In the newer version, this issue is resolve by truncating the numeric to have 0 digits after the decimal point before it's conversion to int8/int4.

### Issues Resolved

BABEL-4013

* **Use case based -** Added test case from original bug

* **Boundary conditions -** Have added testcases inside and outside of the range, and for null inputs.

* **Arbitrary inputs -** Tests already cover arbitrary input.

* **Negative test cases -** Added.

* **Performance tests -** N/A

* **Tooling impact -** N/A

* **Client tests -** N/A

Signed-off-by: Siddhi Jain <siddhij@amazon.com>

### Check List


shah-nirmit (approved):

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -2064,12 +2064,19 @@ smallint_radians(PG_FUNCTION_ARGS)
...
-	Numeric		arg2 = PG_GETARG_NUMERIC(1);
+	int64		arg1;
+	Numeric		arg2;
 	int64 result;
 	Numeric		arg1_numeric,
 				result_numeric;
+	
+	/* when num or power is NULL, return NULL */
+
+	if (PG_ARGISNULL(0) | PG_ARGISNULL(1))
```
sumitj824: Instead of checking NULL for each arguments in C function, we can direclty make sys.power function as `STRICT`.

siddhijain2: Addressed

contrib/babelfishpg_tsql/sql/sys_functions.sql
```diff
@@ -3477,19 +3477,19 @@ RETURNS int  AS 'babelfishpg_tsql','smallint_radians' LANGUAGE C STRICT IMMUTABL
 GRANT EXECUTE ON FUNCTION sys.radians(TINYINT) TO PUBLIC;
 
 CREATE OR REPLACE FUNCTION sys.power(IN arg1 BIGINT, IN arg2 NUMERIC)
-RETURNS bigint  AS 'babelfishpg_tsql','bigint_power' LANGUAGE C IMMUTABLE PARALLEL SAFE;
+RETURNS bigint  AS 'babelfishpg_tsql','bigint_power' LANGUAGE C STRICT IMMUTABLE PARALLEL SAFE;
```
sumitj824, line 3373: We need to add this in upgrade script also.

sumitj824 (changes requested):

sumitj824 (approved):

jsudrik (dismissed):
Based on other approvals.

Deepesh125 (changes requested):
Power operator also needs to be fixes a d need to add test cases

test/JDBC/expected/BABEL-3802-vu-prepare.out
```diff
@@ -63,7 +67,10 @@ CREATE PROCEDURE BABEL_3802_vu_prepare_t_p3 AS (
         power(CAST(NULL AS SMALLINT),1),
         power(CAST(8969.32 AS SMALLINT),1),
         power(CAST(8962 AS SMALLINT),1.1),
-        power(CAST(8962 AS SMALLINT),1.2)
+        power(CAST(8962 AS SMALLINT),1.2),
+        power(CAST(NULL AS SMALLINT),NULL),
```
hash-16, line 89: Can you pls check for this test case as well :-
`Select power(cast(2 as smallint),-1)`
And pls add some testcase with negative power. Thanks

siddhijain2: Addressed

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -2103,12 +2103,13 @@ bigint_power(PG_FUNCTION_ARGS)
...
 	int64 result;
 	Numeric		arg1_numeric,
+				result_floor,
 				result_numeric;
 
 	arg1_numeric = DatumGetNumeric(DirectFunctionCall1(int8_numeric, arg1));
 	result_numeric = DatumGetNumeric(DirectFunctionCall2(numeric_power, NumericGetDatum(arg1_numeric), NumericGetDatum(arg2)));
-
-	result = DatumGetInt64(DirectFunctionCall1(numeric_int8, NumericGetDatum(result_numeric)));
+	result_floor = DatumGetNumeric(DirectFunctionCall1(numeric_floor, NumericGetDatum(result_numeric)));
```
Deepesh125: why do we need to get the floor or result? wouldn't be converting to int8 (begin done on next line) be sufficient?

Deepesh125: Applicable for other functions also

test/JDBC/expected/BABEL-3802-vu-prepare.out
```diff
@@ -41,7 +41,11 @@ CREATE PROCEDURE BABEL_3802_vu_prepare_t_p1 AS (
         power(CAST(NULL AS BIGINT),1),
         power(CAST(8969.32 AS BIGINT),1),
         power(CAST(896932 AS BIGINT),1.1),
-        power(CAST(896932 AS BIGINT),1.2)
+        power(CAST(896932 AS BIGINT),1.2),
+        power(CAST(NULL AS BIGINT),NULL),
+        power(CAST(896932 AS BIGINT),NULL),
+        power(CAST(2 AS BIGINT), -1),
+        power(CAST(0 AS BIGINT),0)
```
Deepesh125: add the similar tests for view? also, add test cases like `power(2, 2.2)` and `power(2, 2.8)` . overflow, underflow tests are missing.

Deepesh125 (changes requested):

contrib/babelfishpg_tsql/runtime/functions.c
siddhijain2: Converting Numeric to int8 is rounding of the results to the nearest integer. Example - power(2,-1) result is 0.5000, if we directly convert it into int8 it will be rounded off to 1, but the correct answer is 0. That's why, used the numeric_floor function before the conversion.

Deepesh125 (dismissed):

hash-16 (approved):

test/JDBC/expected/BABEL-3802-vu-prepare.out
siddhijain2: Yes, Added the similar tests for view.

Deepesh125 (approved):

PR #1709, opened 2023-08-09
Fixed handling of input arguments for ISNULL()

Deepesh125:
### Description

Previously, we were handling ISNULL using COALESCE semantics which is not right thing to do because return datatype of T-SQL ISNULL function is not same as COALESCE in all the cases. So this commit fixes this issue by appropriately implementing hook to select the common type, select_common_type_hook_type. Logic to select common type or return type is very straight forward and described below:
1. If first argument is literal NULL then return the data type of second argument. 
2. If second argument is also literal NULL then return INT. 

There could be major version upgrade due to this change. Have opened BABEL-4263 internally to handle it when we support major version upgrade from version 15 to version 16.

Task: BABEL-917

### Check List


contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -214,6 +215,27 @@ static table_variable_satisfies_vacuum_hook_type prev_table_variable_satisfies_v
 static table_variable_satisfies_vacuum_horizon_hook_type prev_table_variable_satisfies_vacuum_horizon = NULL;
 static drop_relation_refcnt_hook_type prev_drop_relation_refcnt_hook = NULL;
 
+
+/* To cache oid of sys.varchar */
+static Oid sys_varcharoid = InvalidOid;
+
+static Oid get_sys_varcharoid()
```
KushaalShroff: This is good, Can we declare this as extern and make use in 3 other places in extension?

KushaalShroff:
Update PR title appropriately

contrib/babelfishpg_tsql/src/hooks.c
Deepesh125: Done

KushaalShroff (approved):

PR #1714, opened 2023-08-10
Change volatility of sys.replace function from STABLE to IMMUTABLE

sumitj824:
### Description

Earlier, VOLATILITY of sys.replace function was STABLE due which we were getting error while using this function as computed column in table. This commit changes the VOLATILITY of sys.replace function from STABLE to IMMUTABLE.

### Issues Resolved

BABEL-4330

* **Use case based -** yes

* **Arbitrary inputs -** yes

* **Performance tests -** NA

* **Tooling impact -** NA

* **Client tests -** NA

### Check List


test/JDBC/expected/BABEL_4330.out
```diff
@@ -0,0 +1,86 @@
+
+-- We should be able to use replace function in computed column.
```
KushaalShroff, line 2: I think we need to add tests for replace as well.
Then we need to add test case to check if it is immutable
Then we should also add upgrade tests

KushaalShroff (changes requested):

test/JDBC/expected/BABEL_4330.out
sumitj824: addressed

KushaalShroff (approved):

PR #1727, opened 2023-08-11
Handle different behaviour with like escape null and like escape ''

vikashprajapati-cell:
### Description

This PR adds the test cases for the following changes in `postgresql_modified_for_babelfish`:-

When query `select 1 where 'ABCD' LIKE 'AB[C]D' ESCAPE ''` executes on babelfish, it should return an error as expected behaviour. But it was not throwing error, this is actually there was no condition added to return error in case of empty ESCAPE char. So i added a condition to return error for this case.
Another query is `select 1 where 'ABCD' LIKE 'AB[C]D' ESCAPE null`. This should return 1 row as expected behaviour and before this fix, this query returned 0 row on execution on babelfish. So i added one condition to capture this case and revert the behaviour as expected. (Behaviour same as no ESCAPE char used - as per Rob's comment on Jira).

Task: BABEL-4271
Authored-by: Vikash Prajapati [vikasprj@amazon.com](mailto:vikasprj@amazon.com)
Signed-off-by: Vikash Prajapati [vikasprj@amazon.com](mailto:vikasprj@amazon.com)

### Issues Resolved

BABEL-4271

* **Use case based -**
Added test cases for both the queries mentioned in JIRA and covered different possible scenerios.

* **Boundary conditions -** **NA**

* **Arbitrary inputs -** **NA**

* **Negative test cases -** 
  Added 

* **Minor version upgrade tests -** **NA**

* **Major version upgrade tests -** **NA**

* **Performance tests -** **NA**

* **Tooling impact -** **NA**

* **Client tests -** **NA**

### Check List


test/JDBC/input/BABEL-4271.sql
```diff
@@ -0,0 +1,25 @@
+select 1 where 'ABCD' LIKE 'AB[C]D' ESCAPE ''
```
ahmed-shameem: Let's have some tests with table operations as well. For example,
create table t1(a varchar(30), b varchar(30));
insert into t1 values ('abc','abc');

select a,b from t1 where a like b escape '';

ahmed-shameem (changes requested):

test/JDBC/input/BABEL-4271-vu-prepare.sql
```diff
@@ -0,0 +1,8 @@
+-- Test to check like escape null and like escape ''
+create table testvikasprj(a varchar(30), b varchar(30));
```
ahmed-shameem: We follow the convention of giving table names like babel--vu-prepare-. For this case it can be something like babel-4271-vu-prepare-t1. So please update all the names.

test/JDBC/input/BABEL-4271-vu-prepare.sql
```diff
@@ -0,0 +1,8 @@
...
+go
+
+insert into testvikasprj values ('cbc','[c-a]bc');
```
ahmed-shameem: Let us have more edge case. One of them can be something like:
insert into t1 values('abc',' abc")
The difference between the inputs is, the second column has an extra space before abc. So, think of cases like that. And make sure we are getting expected results.

test/JDBC/input/BABEL-4271-vu-verify.sql
```diff
@@ -0,0 +1,30 @@
...
+go
+select 1 where 'abc' LIKE '[0-a]bc' ESCAPE null;
+go
+select 1 where 'abc' LIKE '[abc]bc' ESCAPE null;
+go
+select 1 where 'abc' LIKE '[a-c]bc' ESCAPE null;
+go
+select 1 where 'bbc' LIKE '[a-c]bc' ESCAPE null;
+go
+select a, b from testvikasprj where testvikasprj.a LIKE testvikasprj.b ESCAPE null;
```
ahmed-shameem: Let us have case like:

SELECT a, 'abc' from t1 where t1.a LIKE 'abc' ESCAPE null;
SELECT a, '' from t1 where t1.a LIKE '' ESCAPE null;
SELECT a, '' from t1 where t1.a LIKE '' ESCAPE '';
SELECT a, 'abc' from t1 where t1.a LIKE '' ESCAPE '';
So please try to think of such different scenarios to make sure that we are generating expected output.

ahmed-shameem (changes requested):

test/JDBC/input/BABEL-4271.sql
vikashprajapati-cell: Done

test/JDBC/input/BABEL-4271-vu-prepare.sql
vikashprajapati-cell: updated in the  next revision.

test/JDBC/input/BABEL-4271-vu-verify.sql
vikashprajapati-cell: Added more test cases !

forestkeeper (approved):

Deepesh125 (commented):
Upgrade tests are not required? If some views/functions/proc contains `like .. escape ''` in the definition then does it get dumped properly?

test/JDBC/input/BABEL-4271-vu-prepare.sql
vikashprajapati-cell: Added

PR #1763, opened 2023-08-22
Revert: Return syntax error for identity in select-into

deepakshi-mittal:
Need to do this temporarily because order by clause behavior is not as expected

Task: BABEL-539

### Check List

### DIFF
--- a/contrib/babelfishpg_tsql/src/backend_parser/gram-tsql-epilogue.y.c
+++ b/contrib/babelfishpg_tsql/src/backend_parser/gram-tsql-epilogue.y.c
@@ -233,39 +233,8 @@ TsqlFunctionConvert(TypeName *typename, Node *arg, Node *style, bool try, int lo
 Node *
 TsqlFunctionIdentityInto(TypeName *typename, Node *seed, Node *increment, int location)
 {
-	ereport(ERROR, (errcode(ERRCODE_SYNTAX_ERROR), errmsg("syntax error near 'identity'")));
-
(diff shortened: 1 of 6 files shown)


contrib/babelfishpg_tsql/src/backend_parser/gram-tsql-epilogue.y.c
```diff
@@ -233,35 +233,39 @@ TsqlFunctionConvert(TypeName *typename, Node *arg, Node *style, bool try, int lo
...
-			ereport(ERROR,
-					(errcode(ERRCODE_SYNTAX_ERROR),
-					errmsg("Parameter or variable '' has an invalid data type.")));
-			break;
-	}
-	return result;
+	ereport(ERROR, (errcode(ERRCODE_SYNTAX_ERROR), errmsg("syntax error near 'identity'")));
+
+	// Node *result;
+	// List *args;
```
wboettge: Two small nits: IMO an error message like "Insert Into with Identity is not yet supported" might make it more clear that this is a valid syntax, just one that's not yet supported in BBF. 
And, since we can still have this code saved in a separate branch, I'd suggest just deleting this instead of commenting it out.

deepakshi-mittal: Done both
DBE suggested message  "IDENTITY() function in SELECT INTO is not supported."

xhfanhe (approved):

PR #1764, opened 2023-08-23
Handle different behavior with escape char('\') for BBF

RIC06X:
### Description

Postgres uses escape clause with empty string to turn off default escape. Since TSQL does not allow empty string as a valid escape pattern. We use non-UTF8 character in escape clause as a flag and reuse the code logic of escape clause with empty string to disable default escape character ’\‘. 

### Issues Resolved
Solved behavior difference when using '\' in like clause in babelfish
Task: BABEL-4270

like expression without escape clause

1. regular like expression
2. like expression that pattern has default escape character ‘\’ 
3. like expression that pattern has default invalid UTF-8 character ‘\xFE’
4. like pattern from result of a select clause
    1. select result with ‘\’
    2. select result with ‘\xFE’
5. like pattern from a variable/function call
6. like expression with non-varchar pattern

like expression with escape clause

1. regular like expression with escape clause
    1. with any escape charcter 
2.  like expression with special escape clause
    1. ‘\’
    2. ‘\xFE’ 
3. expect escape all ‘\’ rather than ‘\xFE’
4. should we expect user to escape this character? *‘0xFE’ is an non-UTF-8 Character
5. like pattern from result of a select clause?
6. like pattern not from a variable and escape character from a variable
7. like pattern from a variable and escape character from a variable
8. like pattern from a variable and escape character from a function
9. like pattern from a variable and escape character from a non-varchar value 

like expression within a procedure

### Check List

### DIFF
--- a/test/JDBC/expected/BABEL-4046-vu-prepare.out
+++ b/test/JDBC/expected/BABEL-4046-vu-prepare.out
@@ -41,7 +41,7 @@ select * from t join t_babel4046 on a like b;
 GO
 
 CREATE VIEW babel4046_2 as
-select * from t join t_babel4046 on a like 'aa%';
+select * from t join t_babel4046 on a like 'aa%' order by 1 offset 0 ROWS;;
 GO
 
 CREATE TABLE [dbo].[t3](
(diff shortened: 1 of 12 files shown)


forestkeeper (approved):
lgtm

PR #1769, opened 2023-08-23
Babel should not allow VIEW in OUTPUT INTO clause

vikashprajapati-cell:
### Description

when query `UPDATE table1 SET c1 = 23 OUTPUT INSERTED.c2, 22 INTO view1` executes on babelfish, it executes without throwing any error. Ideally it should throw error. So, to fix this, the `relkind` is fetched using `relid` and then added condition to check whether this relation is view or not. And error is thrown in this if block.
This PR contains the changes for VIEW relation only, for common table expression babel does not support the syntax right now. 

### Issues Resolved

BABEL-2144

* **Use case based -**
Added

* **Boundary conditions -**

* **Arbitrary inputs -**

* **Negative test cases -**

* **Performance tests -**

* **Tooling impact -**

* **Client tests -**

### Check List

### DIFF
--- a/contrib/babelfishpg_tsql/src/hooks.c
+++ b/contrib/babelfishpg_tsql/src/hooks.c
@@ -2324,10 +2324,19 @@ modify_insert_stmt(InsertStmt *stmt, Oid relid)
 	HeapTuple	tuple;
 	List	   *insert_col_list = NIL,
 			   *temp_col_list;
+	char		relkind = get_rel_relkind(relid);
 
 	if (!output_into_insert_transformation)
+	{
+		if(relkind == RELKIND_VIEW || relkind == RELKIND_MATVIEW)
+		{
+			ereport(ERROR,
+					(errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
+					 errmsg("The target '%s' of the OUTPUT INTO clause cannot be a view or common table expression.", stmt->relation->relname)));
+		}
 		return;
-
+	}
+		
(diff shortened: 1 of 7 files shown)


test/JDBC/input/BABEL-2144-vu-verify.sql
```diff
@@ -0,0 +1,18 @@
+UPDATE babel_2144_vu_prepare_t1 SET c1 = 23 OUTPUT INSERTED.c2, 22 INTO babel_2144_vu_prepare_v1;
```
rishabhtanwar29, line 1: UPDATE is not the only query where OUTPUT INTO clause can be used, it can used in INSERT and DELETE statements as well. So we need to add test cases for these as well.

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -2324,10 +2324,19 @@ modify_insert_stmt(InsertStmt *stmt, Oid relid)
 	HeapTuple	tuple;
 	List	   *insert_col_list = NIL,
 			   *temp_col_list;
+	char		relkind = get_rel_relkind(relid);
 
 	if (!output_into_insert_transformation)
```
rishabhtanwar29, line 2329: Issue (blocking): Hmm, this is not going to work for OUTPUT INTO clause in a INSERT statement as `output_into_insert_transformation` is going to be true in that case. We need to have a proper conditional to verify if this the `insert statement` of a OUTPUT INTO clause we are dealing with here.

Issue #1791, opened 2023-08-28
[Bug]: view sys.index_columns reponds with a incorrect value in field column_id


tophfr:
### What happened?

Using DBeaver to browse the database connecting with and SQL Server driver to an AWS Aurora instance with Babelfish enabled, I noticed that columns in indexed are always shown as the first columns in the table.

I finaly found that this is due to the query of the view sys.index_columns which use the attnum field from the table pg_attribute. But this table joined like it is currently give no information of the column id (or position) inside the table.

I added a extra join to pg_attribute to fix this (pull request will follow...)

### Version

BABEL_3_X_DEV (Default)

### Extension

None

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct.

robverschoor:
Could you please include a short reproduction of the issue?

tophfr:
Create a table with indices in columns other than first ones:

```sql
create table t1 (
  c1 int
, c2 int not null
, c3 int
, c4 int
, constraint t1_pk primary key (c2)
);
create index t1_i1 on t1 (c3, c1);
create index t1_i2 on t1 (c4);
```

Then query the sys tables:

```sql
select i.name as index_name
     , ic.key_ordinal
     , ic.column_id
     , c.name as column_name
  from sys.indexes i
  join sys.index_columns ic on i.object_id = ic.object_id and i.index_id = ic.index_id
  join sys.columns c ON ic.object_id = c.object_id and ic.column_id = c.column_id
  join sys.tables t ON i.object_id = t.object_id 
 where t.name = 't1'
;
```

Expected result:
| index_name | key_ordinal | column_id | column_name |
|-------------|-------------|------------|---------------|
| t1_pk | 2 | 1 | c2 |
| t1_i1 | 3 | 1 | c3 |
| t1_i1 | 1 | 2 | c1 |
| t1_i2 | 4 | 1 | c4 |


Writing this made me realize that index_id in sys.index_columns is incorrect for columns with key_ordinal > 1.

```sql
select name
         , index_id 
  from sys.indexes
where name like 't1%';
```
| name | index_id |
|-------|---------|
| t1_pkt1ad527cedc91b740e0f35d5161fdc69e3 | 2 |
| t1_i1t1fc3ad6f1506b2a13371857879dfdbad8 | 3 |
| t1_i2t104a836e77ad57814e46a4ebb6606df0e | 4 |

=> ok

```sql
select ic.index_id
[...]

PR #401, opened 2022-07-18
Get T-SQL DATEPART working for string arguments

thephantomthief:
### Description

In this commit, we modified the SQL definition of sys.datepart() such
that if the arg is of type text, then we explicitly cast it to datetime
before we call datepart_internal.

Added test cases for the same.

Task: BABEL-2308
Signed-off-by: Sharu Goel <goelshar@amazon.com>

### Check List
- [x] Commits are signed per the DCO using --signoff 

By submitting this pull request, I confirm that my contribution is under the terms of the Apache 2.0 and PostgreSQL licenses, and grant any person obtaining a copy of the contribution permission to relicense all or a portion of my contribution to the PostgreSQL License solely to contribute all or a portion of my contribution to the PostgreSQL open source project.

For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/babelfish-for-postgresql/babelfish_extensions/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #363, opened 2022-07-06
Added entries to sys.configurations view

faviansamatha:
### Description
Previously, there were missing values in the sys.configurations which blocked the Table - Create To feature in SSMS. This commit fixes that issue by adding values to sys.babelfish_configurations, the underlying table for sys.configurations view.

Task: BABELFISH-296
Signed-off-by: Favian (Ian) Samatha <ians@bitquilltech.com>

### Check List
- [X] Commits are signed per the DCO using --signoff 

By submitting this pull request, I confirm that my contribution is under the terms of the Apache 2.0 and PostgreSQL licenses, and grant any person obtaining a copy of the contribution permission to relicense all or a portion of my contribution to the PostgreSQL License solely to contribute all or a portion of my contribution to the PostgreSQL open source project.

For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/babelfish-for-postgresql/babelfish_extensions/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #1792, opened 2023-08-28
fixing column_id in view sys.index_columns

tophfr:
### Description

Using DBeaver to browse the database connecting with and SQL Server driver to an AWS Aurora instance with Babelfish enabled, I noticed that columns in indexed are always shown as the first columns in the table.

I finaly found that this is due to the query of the view sys.index_columns which use the attnum field from the table pg_attribute. But this table joined like it is currently give no information of the column id (or position) inside the table.

I added a extra join to pg_attribute to fix this (pull request will follow...)

### Issues Resolved

Issue #1791 

* **Use case based -**

* **Boundary conditions -**

* **Arbitrary inputs -**

* **Negative test cases -**

* **Performance tests -**

* **Tooling impact -**

* **Client tests -**

### Check List

### DIFF
--- a/contrib/babelfishpg_tsql/sql/sys_views.sql
+++ b/contrib/babelfishpg_tsql/sql/sys_views.sql
@@ -1960,7 +1960,7 @@ GRANT SELECT ON sys.endpoints TO PUBLIC;
 create or replace view sys.index_columns
 as
 select i.indrelid::integer as object_id
-  , CAST(CASE WHEN i.indisclustered THEN 1 ELSE 1+row_number() OVER(PARTITION BY c.oid) END AS INTEGER) AS index_id
+  , imap.index_id
   , a.attrelid::integer as index_column_id
   , col.attnum::integer as column_id
   , a.attnum::sys.tinyint as key_ordinal
@@ -1972,6 +1972,7 @@ inner join pg_catalog.pg_attribute a on i.indexrelid = a.attrelid
 inner join pg_catalog.pg_attribute col on col.attname = a.attname and col.attrelid = i.indrelid
 inner join pg_class c on i.indrelid = c.oid
 inner join sys.schemas sch on sch.schema_id = c.relnamespace
+inner join (select indexrelid, CAST(CASE WHEN indisclustered THEN 1 ELSE 1+row_number() OVER(PARTITION BY indrelid) END AS INTEGER) AS index_id from pg_index) as imap on imap.indexrelid = i.indexrelid
 where has_schema_privilege(sch.schema_id, 'USAGE')
 and has_table_privilege(c.oid, 'SELECT,INSERT,UPDATE,DELETE,TRUNCATE,TRIGGER');
 GRANT SELECT ON sys.index_columns TO PUBLIC;
--- a/contrib/babelfishpg_tsql/sql/sys_views.sql
+++ b/contrib/babelfishpg_tsql/sql/sys_views.sql
@@ -1962,13 +1962,14 @@ as
 select i.indrelid::integer as object_id
   , CAST(CASE WHEN i.indisclustered THEN 1 ELSE 1+row_number() OVER(PARTITION BY c.oid) END AS INTEGER) AS index_id
   , a.attrelid::integer as index_column_id
-  , a.attnum::integer as column_id
+  , col.attnum::integer as column_id
   , a.attnum::sys.tinyint as key_ordinal
   , 0::sys.tinyint as partition_ordinal
   , 0::sys.bit as is_descending_key
   , 1::sys.bit as is_included_column
 from pg_index as i
 inner join pg_catalog.pg_attribute a on i.indexrelid = a.attrelid
+inner join pg_catalog.pg_attribute col on col.attname = a.attname and col.attrelid = i.indrelid
 inner join pg_class c on i.indrelid = c.oid
 inner join sys.schemas sch on sch.schema_id = c.relnamespace
 where has_schema_privilege(sch.schema_id, 'USAGE')


contrib/babelfishpg_tsql/sql/sys_views.sql
```diff
@@ -1960,17 +1960,19 @@ GRANT SELECT ON sys.endpoints TO PUBLIC;
...
   , a.attnum::sys.tinyint as key_ordinal
   , 0::sys.tinyint as partition_ordinal
   , 0::sys.bit as is_descending_key
   , 1::sys.bit as is_included_column
 from pg_index as i
 inner join pg_catalog.pg_attribute a on i.indexrelid = a.attrelid
+inner join pg_catalog.pg_attribute col on col.attname = a.attname and col.attrelid = i.indrelid
 inner join pg_class c on i.indrelid = c.oid
 inner join sys.schemas sch on sch.schema_id = c.relnamespace
+inner join (select indexrelid, CAST(CASE WHEN indisclustered THEN 1 ELSE 1+row_number() OVER(PARTITION BY indrelid) END AS INTEGER) AS index_id from pg_index) as imap on imap.indexrelid = i.indexrelid
```
KushaalShroff, line 1975: Can this be optimised by rather doing a join on pg_index and then leaving `CAST(CASE WHEN indisclustered THEN 1 ELSE 1+row_number() OVER(PARTITION BY indrelid) END AS INTEGER)`  as it was before, part of SELECT index_id.
The above mentioned would be a better optimisation _BUT_ if not possible we can think of adding this as a materialized CTE instead.
The performance of these catalog views gets exacerbated when we join multiple tsql views in a client facing query(with a lot of objects). These queries can also start taking as much time in the order of minutes and even hours. Hence, we should be very careful while making such changes.

jsudrik (commented):
The upgrade code and the tests will need to be added. 

MVU test framework is integrated to run with JDBC test framework. At present, it runs as part of Github push/pull actions but we cannot run it on our local machine. Framework works as follows.

* Read the config file for upgrade path to test
* For source version
    * Checkout and build the corresponding code
    * Run prepare scrips for the version to populate database
    * Run upgrade action
* For intermediate versions
    * Checkout and build the corresponding code
    * Run upgrade version
* For target version
    * Checkout and build the corresponding code
    * Run final upgrade action
    * Run verify/cleanup scripts for the source version to validate database

Upgrade test file format

Upgrade test will consist of three files.

* Prepare file to create desired objects which populate the database
* Verify file to validate database state
* Cleanup file to destroy desired objects

MVU test directory structure

* Config file : .github/configuration/upgrade-test-configuration.yml
    * This file has list of all the the upgrade paths to be tested. User can add more paths by specifying the release version and upgrade type (minor/major)
* Main test directory : babelfish_extensions/test/JDBC
    * input directory
        * Contains all the normal JDBC tests
        * Contains release independent upgrade tests in *-vu-prepare.sql, *-vu-verify.sql and *-vu-cleanup.sql format
        * An upgrade test files here will also run as part of normal JDBC run. If upgrade test provides required coverage, you should not add a separate JDBC test
    * upgrade directory
        * One sub directory per release
            * Contains release version specific prepare file *vu-prepare.sql where this release is source
            * Contains schedule file to specify the upgrade test to run from input directory
                * This is useful when same tests can work across releases and avoid duplicating files across different release directories
            * One sub directory per release
                * Contains verify/cleanup scripts for specific release source version when this release is target

MVU test writing guidelines

* If same tests will work same across different releases, add them to input directory and corresponding entry into schedule file for various releases
* If a particular test needs release specific changes or its output is release specific, add it to upgrade/release/ directory
* If a particular test needs release specific verification tests or its output is release specific, add it to upgrade/release/source_release/directory
* Use file as prefix for objects name created inside tests (should be applicable to non upgrade tests as well)
* If new system object definitions are created which are not available in previous releases, we should still create upgrade tests to run as part of latest to latest upgrade
* When updating/creating system object definitions, make sure to add tests which exercise object dependency like placing object inside views, functions, procedures, tables and so on.

Example PRs

1. [Sample PR1](https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/363/): Where adding test entry under upgrade/<version>/schedule was enough as expected output is identical across different versions
2. [Sample PR2](https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/347/files): Where we had to add test cases under upgrade/base_version/preparation and upgrade/target_version/verification_cleanup as output is varying between different versions
3. https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/401
4. https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/365
5. https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/270/files (Nice trick to reduce the number of version-specific test files)

contrib/babelfishpg_tsql/sql/sys_views.sql
tophfr: On my side, I can't think of a alternative to get the correct row_number() value...

contrib/babelfishpg_tsql/sql/sys_views.sql
```diff
@@ -1960,17 +1960,19 @@ GRANT SELECT ON sys.endpoints TO PUBLIC;
...
   , a.attrelid::integer as index_column_id
-  , a.attnum::integer as column_id
+  , col.attnum::integer as column_id
   , a.attnum::sys.tinyint as key_ordinal
   , 0::sys.tinyint as partition_ordinal
   , 0::sys.bit as is_descending_key
   , 1::sys.bit as is_included_column
 from pg_index as i
 inner join pg_catalog.pg_attribute a on i.indexrelid = a.attrelid
+inner join pg_catalog.pg_attribute col on col.attname = a.attname and col.attrelid = i.indrelid
```
Deepesh125, line 1972: This condition col.attname = a.attname and col.attrelid = i.indrelid will ensure that col.attnum will be same as a.attnum and col.relid will be same as a.relid so I am not convinced why do we need this change.

contrib/babelfishpg_tsql/sql/sys_views.sql
Deepesh125: I agree with Kushaal here. If we can use CTE then we should use it

Deepesh125 (changes requested):
Overall please add upgrade script for this change and add some test cases to prove the correctness. Thanks!

contrib/babelfishpg_tsql/sql/sys_views.sql
tanscorpio7: column_id will return the position of the column from the table definition,
key_ordinal will return the position of the column from the index definiton.

tanscorpio7:
```
create table t4 (c1 int, c2 int, c3 int, c4 int, c5 int, primary key clustered (c3, c5))
select 't4' as table_name, column_id, key_ordinal from sys.index_columns where object_id =  object_id('t4')
table_name column_id   key_ordinal
---------- ----------- -----------
t4                   3           1
t4                   5           2
```

contrib/babelfishpg_tsql/sql/sys_views.sql
tanscorpio7: In the original definition
```
CAST(CASE 
    WHEN i.indisclustered THEN 1 
    ELSE 1+dense_rank() OVER(PARTITION BY i.indrelid ORDER BY i.indexrelid)
END AS INT) AS index_id,
```
Could this work ?
Rank gives unique values to rows with same oder by column value, and dense rank makes sure there are no gaps
So in this case, we first partition based on table OID
Then for within each partition we start assigning ranks to each row. (rows with same index OID will get same rank)

tanscorpio7:
addressed in https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/2424

PR #1797, opened 2023-08-29
Refinement of FOR JSON PATH & Nested Functionality

northaxosky:
Currently FOR JSON PATH is implemented in Babelfish using string manipulation. This change will implement FOR JSON PATH using PostgreSQL's JsonbValue and HTAB. The reason for this change is because the current implementation of FOR JSON PATH does not support the full functionality of the feature. Nested outputs determined by aliases in the SELECT statement are not supported. By implementing FOR JSON PATH using JsonbValue and HTAB, we provide an efficient solution that also supports nested json objects, while also setting the foundation for FOR JSON AUTO to be implemented.

**BABELFISH_EXTENSIONS CHANGES:**
- Switched from String Manipulation implementation of FOR JSON PATH to an implementation using JsonbValue.
- Wrapper function to add a value to JsonbValue and handle datatype checks.
- Supporting functions to determine path to insert a JsonbValue object and key for HashTable.
- Create json and insert into existing json functions that are used when dealing with nested json objects.
- Implemented Nested Output tracking using PostgreSQL's HTAB library (HashTable)
- Corrected the tests to showcase the new functionality of nested json objects
- Added another set of tests to more thoroughly test the nested json object functionality.

**POSTGRESQL_MODIFIED_FOR_BABELFISH CHANGES:**
- Created a public wrapper function in jsonb.h  to call add_jsonb() from forjson.c 

**TEST CASES:**
- Added a test that checks the functionality and accuracy of the nested json feature. Includes cases with multiple layers, deeply nested layers, values in different stages of json layers, nested null outputs.
- Organized version upgrade tests so previous versions use previous implementation of FOR JSON PATH and updated the schedule for each version.
- Modified existing test cases to correctly showcase the nested functionality if the given test uses alias that cause nested json objects to be created.
- Changed the expected outputs to include spaces as JsonbValue does when converted to a string.


contrib/babelfishpg_tsql/src/tsql_for/forjson.c
```diff
@@ -1,10 +1,4 @@
-/*-------------------------------------------------------------------------
- *
- * forjson.c
- *   For JSON clause support for Babel
- *
- *-------------------------------------------------------------------------
- */
+// forjson.c
```
Jakeowen1: Please replace this comment with the original forjson.c header comment

contrib/babelfishpg_tsql/src/tsql_for/forjson.c
```diff
@@ -15,21 +9,55 @@
...
+// Function to append into existing JsonbValue
+static void insert_existing_json(JsonbValue *exists, JsonbValue* parent, JsonbValue *val, int idx, char *key);
 
 PG_FUNCTION_INFO_V1(tsql_query_to_json_sfunc);
 
 Datum
 tsql_query_to_json_sfunc(PG_FUNCTION_ARGS)
 {
-	StringInfo	state;
+    forjson_state *state;
```
Jakeowen1: nit: indentation

contrib/babelfishpg_tsql/src/tsql_for/forjson.c
```diff
@@ -265,12 +299,261 @@ tsql_row_to_json(StringInfo state, Datum record, bool include_null_values)
...
+				else	{
+					hashEntry->parent = jsonbRow;
+					hashEntry->idx = jsonbRow->val.object.nPairs;
+				}
+
+				pfree(hashKey);
+			}
+		}
+		// Already inserted into existing json object
+		if (found)
```
Jakeowen1: Should this line be inside of the for loop?

contrib/babelfishpg_tsql/src/tsql_for/forjson.c
```diff
@@ -265,12 +299,261 @@ tsql_row_to_json(StringInfo state, Datum record, bool include_null_values)
...
+
+				pfree(hashKey);
+			}
+		}
+		// Already inserted into existing json object
+		if (found)	
+			continue;
+		
+		// JsonbValue was created in loop and needs to be inserted to the row jsonbValue
+		if (num > 1) {
```
Jakeowen1: This check is already on line 320, can we move this logic after the loop?

contrib/babelfishpg_tsql/src/tsql_for/forjson.c
```diff
@@ -265,12 +299,261 @@ tsql_row_to_json(StringInfo state, Datum record, bool include_null_values)
...
+			resStr->data[resStr->len - 1] = '\0';
+			resStr->len--;
+		}
+	}
+	
+	PG_RETURN_TEXT_P(cstring_to_text_with_len(resStr->data, resStr->len));
+}
+
+static char**
+determine_parts(const char* str, int* num)
```
Jakeowen1, line 458: For this function, can we just run strtok on copy_str and return that? Is there a reason why we need a nested array?

contrib/babelfishpg_tsql/src/tsql_for/forjson.c
```diff
@@ -265,12 +299,261 @@ tsql_row_to_json(StringInfo state, Datum record, bool include_null_values)
...
+{
+	int			i;
+	char		**parts;
+	char		*copy_str;
+	char 		*token;
+
+	// Determine how many parts there are (words seperated by ".")
+	*num = 1;
+	for (i = 0; str[i]; i++)	{
```
Jakeowen1, line 467: We can get num by taking the length of token so we can delete this logic

Jakeowen1 (changes requested):

contrib/babelfishpg_tsql/src/tsql_for/forjson.c
```diff
@@ -15,21 +9,55 @@
...
+
+// Entry struct for use in HashTable
+typedef struct {
+	char path[NAMEDATALEN];
+	JsonbValue *value;
+	JsonbValue *parent;
+	int idx;
+} JsonbEntry;
+
+// Main row to JSON function
```
Sairakan: No need to add these comment stubs for the function declarations- just make sure to add a proper header comment for the actual function implementation

contrib/babelfishpg_tsql/src/tsql_for/forjson.c
```diff
@@ -15,21 +9,55 @@
...
+#include "utils/hsearch.h"
 #include "catalog/pg_type.h"
 #include "catalog/namespace.h"
 
 #include "tsql_for.h"
 
-static void tsql_row_to_json(StringInfo state, Datum record, bool include_null_values);
+#define TABLE_SIZE 100
+
+// For holding information regarding the state of the FOR JSON call
```
Sairakan, line 29: This comment is redundant with the name of the struct - you can either remove this or expand on why this struct is needed if it is not obvious

contrib/babelfishpg_tsql/src/tsql_for/forjson.c
```diff
@@ -15,21 +9,55 @@
...
+typedef struct {
+	bool without_array_wrapper;
+	char *root_name;
+	JsonbValue* jsonbArray;
+} forjson_state;
+
+// Entry struct for use in HashTable
```
Sairakan, line 38: This is a helpful comment

Sairakan (approved):

contrib/babelfishpg_tsql/src/tsql_for/forjson.c
```diff
@@ -103,45 +136,28 @@ tsql_query_to_json_sfunc(PG_FUNCTION_ARGS)
...
+// Main row to json function. 
+// Creates a Jsonb row object, processes the row, determines if it should be inserted as a nested json object
+// inserts json object to row and then into the main jsonbArray.
+static void
+tsql_row_to_json(JsonbValue* jsonbArray, Datum record, bool include_null_values)
 {
-	StringInfo	res = makeStringInfo();
-	char	   *state = ((StringInfo) PG_GETARG_POINTER(0))->data;
+	// HashTable
+	HTAB	   *jsonbHash;
```
Sairakan, line 149: Is it possible to have this hash table persist between invocations? Because I believe that the nested structure should be the same for each row. You could initialize it in the initfunc and have it stored in the state struct. This would eliminate unnecessarily recreating the hash table for each row.

Jakeowen1 (approved):

test/JDBC/upgrade/14_6/schedule
```diff
@@ -175,8 +175,8 @@ dateadd_internal_df
 datediff_internal_date-before-14_7-or-15_2
 datepart
 datetime2fromparts
-forjson
-forjson-datatypes
+forjson-before-14_9
```
2jungkook: Since we are targeting 14.10 and 15.5, we should use `before-14_10-or-15_5`. Also, please add tests in `15_x` (all applicable schedule files for PG15), `14_9` and `latest` schedule files.

contrib/babelfishpg_tsql/src/tsql_for/forjson.c
```diff
@@ -265,12 +305,270 @@ tsql_row_to_json(StringInfo state, Datum record, bool include_null_values)
...
+					// function call
+					current = hashEntry->value;
+					insert_existing_json(current, hashEntry->parent, nestedVal, hashEntry->idx, colname);
+					pfree(hashKey);
+					break;
+				}
+
+				// If it does not exist
+				hashEntry = (JsonbEntry *) hash_search(jsonbHash, (void *) hashKey, HASH_ENTER, NULL);
+				strlcpy(hashEntry->path, hashKey, NAMEDATALEN);
```
2jungkook, line 344: How can we guarantee the length of `hashKey` is smaller than `NAMEDATALEN`?

2jungkook (commented):
Since this is a major change on processing `FOR JSON` in Babelfish, I'd like to see a performance comparison with and without this change against very large json outputs.

contrib/babelfishpg_tsql/src/tsql_for/forjson.c
Jakeowen1: The antlr parser enforces a 63 character limit on IDENT strings that it reads. If the path is >= 64 characters it will hashed it to be 63 characters. This is a general problem that affects multiple JSON functions, and would require modifying the way we initially parse the input. This will prevent the path from being greater than NAMEDATALEN, and I will create a jira to track fixing the 64 character bug issue

test/JDBC/upgrade/14_6/schedule
Jakeowen1: I just ran a select for json query on a table with 80,000 rows: the old implementation took 1263.478 ms and the new implementation took 2909.017 ms. I will start looking into ways that we can optimize the new implementation to try and bridge the gap

```
-- WITH CHANGE

~START~
text
Query Text: SELECT S.Id AS "States.id",
       S.name AS "States.name",
       T.name AS "States.Towns.name"
  FROM States S
      JOIN Towns T
          ON (T.id = S.Id)
  FOR JSON PATH
ProjectSet  (cost=3452.81..3458.07 rows=1000 width=32) (actual time=2552.838..2555.300 rows=1 loops=1)
  ->  Aggregate  (cost=3452.81..3452.82 rows=1 width=32) (actual time=2539.417..2539.426 rows=1 loops=1)
        ->  Hash Join  (cost=1963.47..3291.20 rows=64643 width=68) (actual time=129.860..335.677 rows=80000 loops=1)
              Hash Cond: (s.id = t.id)
              ->  Seq Scan on states s  (cost=0.00..1157.70 rows=64770 width=36) (actual time=0.049..66.808 rows=80000 loops=1)
              ->  Hash  (cost=1155.43..1155.43 rows=64643 width=36) (actual time=129.549..129.552 rows=80000 loops=1)
                    Buckets: 131072 (originally 65536)  Batches: 1 (originally 1)  Memory Usage: 5083kB
                    ->  Seq Scan on towns t  (cost=0.00..1155.43 rows=64643 width=36) (actual time=0.049..68.113 rows=80000 loops=1)
Planning Time: 0.025 ms
Execution Time: 2909.017 ms
~END~

-- WITHOUT CHANGE

~START~
text
Query Text: SELECT S.Id AS "States.id",
       S.name AS "States.name",
       T.name AS "States.Towns.name"
  FROM States S
      JOIN Towns T
          ON (T.id = S.Id)
  FOR JSON PATH
ProjectSet  (cost=3452.81..3458.07 rows=1000 width=32) (actual time=903.355..905.714 rows=1 loops=1)
  ->  Aggregate  (cost=3452.81..3452.82 rows=1 width=32) (actual time=890.383..890.412 rows=1 loops=1)
        ->  Hash Join  (cost=1963.47..3291.20 rows=64643 width=68) (actual time=132.007..312.548 rows=80000 loops=1)
              Hash Cond: (s.id = t.id)
              ->  Seq Scan on states s  (cost=0.00..1157.70 rows=64770 width=36) (actual time=0.042..61.139 rows=80000 loops=1)
              ->  Hash  (cost=1155.43..1155.43 rows=64643 width=36) (actual time=131.692..131.714 rows=80000 loops=1)
                    Buckets: 131072 (originally 65536)  Batches: 1 (originally 1)  Memory Usage: 5083kB
                    ->  Seq Scan on towns t  (cost=0.00..1155.43 rows=64643 width=36) (actual time=0.044..67.042 rows=80000 loops=1)
Planning Time: 0.026 ms
Execution Time: 1263.478 ms
~END~
```

Jakeowen1: SQL Server execution times on the same query

```
 SQL Server Execution Times:
   CPU time = 1000 ms,  elapsed time = 1285 ms.
SQL Server parse and compile time: 
   CPU time = 0 ms, elapsed time = 0 ms.
```

Jakeowen1: Created this jira to track https://jira.rds.a2z.com/browse/BABEL-4490

2jungkook (approved):

PR #1798, opened 2023-08-29
Add TAP tests for Babelfish dump/restore

rishabhtanwar29:
### Description
Babelfish currently only supports dump and restore from databases with
same migration mode ( i.e. either singledb->singledb or multidb->multidb)
and same versions. This has already been blocked in pg_dump/pg_dumpall
utilities. This commit adds tests to verify the above in TAP tests framework.

TASK: BABEL-4055

Engine PR: https://github.com/babelfish-for-postgresql/postgresql_modified_for_babelfish/pull/204

* **Use case based -**

* **Boundary conditions -**

* **Arbitrary inputs -**

* **Negative test cases -**

* **Performance tests -**

* **Tooling impact -**

* **Client tests -**

### Check List


contrib/babelfishpg_tds/test/t/004_bbfdumprestore.pl
```diff
@@ -0,0 +1,175 @@
...
+# dump/restore is not yet supported.
+$newnode->command_fails_like(
+	[
+		'psql',
+		'-d',         'testdb',
+		'-U',         'test_master',
+		'-p',         $newnode->port,
+		'-f',         $dump1_file,
+	],
+	qr/Backup and restore across different Postgres versions is not yet supported./,
```
surendravishnoi: why backup and restore? rather than dump and restore

contrib/babelfishpg_tds/test/t/004_bbfdumprestore.pl
```diff
@@ -0,0 +1,175 @@
...
+	'postgresql.conf', qq{
+	log_connections = on
+	listen_addresses='127.0.0.1'
+	shared_preload_libraries = 'babelfishpg_tds'
+	lc_messages = 'C'
+});
+$newnode2->start;
+# Initialize Babelfish in new node
+my $tsql_newnode2 = new TDSNode($newnode2);
+$tsql_newnode2->init_tsql('test_master', 'testdb', 'multi-db');
```
surendravishnoi, line 130: where are we making modes diff?

.github/workflows/tap-tests.yml
```diff
@@ -33,37 +36,64 @@ jobs:
           sudo -E apt-get install krb5-admin-server krb5-kdc krb5-user libkrb5-dev -y -qq
         shell: bash
 
-      - name: Build Modified Postgres
-        id: build-modified-postgres
+      - name: Build Modified Postgres using ${{env.ENGINE_BRANCH_OLD}}
```
surendravishnoi, line 39: are we always creating both old and new installations?

.github/workflows/tap-tests.yml
```diff
@@ -4,7 +4,10 @@ on: [push, pull_request]
 jobs:
   run-babelfish-tap-tests:
     env:
-      INSTALL_DIR: psql
```
surendravishnoi, line 7: why do we use tap tests instead of JDBC test framework?

rishabhtanwar29: We can't test dump/restore binaries in JDBC framework. JDBC framework is there to add only SQL level tests which is not the target in this case.

.github/workflows/tap-tests.yml
rishabhtanwar29: Yep, need both installations to verify if we correctly blocking cross-version dump/restore.

contrib/babelfishpg_tds/test/t/004_bbfdumprestore.pl
rishabhtanwar29: So here we are passing `multi-db` for `newnode2` which we make it a multi-db server and `newnode1` (at line 53) is already started with `single-db` mode (default is `single-db` if we don't specify anything to `init_tsql`).

contrib/babelfishpg_tds/test/t/004_bbfdumprestore.pl
rishabhtanwar29: No particular reason as such, this error message was there in the previous PR as well we reviewed and nothing stroked my mind to change it as it still makes sense. Do you think we should change it to `dump`?

surendravishnoi: I think dump and restore would be better if we are running dump/restore utility.

surendravishnoi (dismissed):

surendravishnoi (approved):

PR #1802, opened 2023-08-30
Fix DB_NAME() returning downcased values

anju15bharti:
**Description**

When a database name has uppercase characters. the Babelfish's DB_NAME() was returning the value in lowercase, so implemented the fix that will now onwards return the original case value that the user gives and will return lowercase value for already existed database.

Task: BABEL-2564
Authored-by: Anju Bharti [abanju@amazon.com](mailto:abanju@amazon.com)
Signed-off-by: Anju Bharti [abanju@amazon.com](mailto:abanju@amazon.com)

**Issues Resolved**
JIRA BABEL-2564
The BIFs DB_NAME() returns the value in lowercase.

 **Check List**


contrib/babelfishpg_tsql/sql/information_schema_tsql.sql
```diff
@@ -756,17 +756,17 @@ GRANT SELECT ON information_schema_tsql.sequences TO PUBLIC;
  * SCHEMATA view
  */
 CREATE OR REPLACE VIEW information_schema_tsql.schemata AS
-	SELECT CAST(sys.db_name() AS sys.sysname) AS "CATALOG_NAME",
-	CAST(CASE WHEN np.nspname LIKE CONCAT(sys.db_name(),'%') THEN RIGHT(np.nspname, LENGTH(np.nspname) - LENGTH(sys.db_name()) - 1)
+	SELECT CAST(LOWER(sys.db_name()) AS sys.sysname) AS "CATALOG_NAME",
```
HarshLunagariya: Please verify whether this LOWER() for CATALOG_NAME column is needed or not?

contrib/babelfishpg_tsql/sql/ownership.sql
```diff
@@ -9,6 +9,7 @@ CREATE TABLE sys.babelfish_sysdatabases (
 	name TEXT NOT NULL COLLATE "C",
 	crdate timestamptz NOT NULL,
 	properties TEXT NOT NULL COLLATE "C",
+  orig_name TEXT NOT NULL COLLATE "C",
```
HarshLunagariya: Fix indent

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--3.2.0--3.3.0.sql
```diff
@@ -862,6 +862,5 @@ IMMUTABLE;
 
 ALTER FUNCTION sys.replace (in input_string text, in pattern text, in replacement text) IMMUTABLE;
 
-
 -- Reset search_path to not affect any subsequent scripts
-SELECT set_config('search_path', trim(leading 'sys, ' from current_setting('search_path')), false);
\ No newline at end of file
+SELECT set_config('search_path', trim(leading 'sys, ' from current_setting('search_path')), false);
```
HarshLunagariya: Unwanted changes

contrib/babelfishpg_tsql/sql/ownership.sql
```diff
@@ -64,7 +65,7 @@ GRANT SELECT ON sys.sysdatabases TO PUBLIC;
 
 -- PG_NAMESPACE_EXT
 CREATE VIEW sys.pg_namespace_ext AS
-SELECT BASE.* , DB.name as dbname FROM
+SELECT BASE.* , DB.orig_name as dbname FROM
```
HarshLunagariya, line 68: This internal view is being used by many other system objects. Are we sure that we need orig_name here?

contrib/babelfishpg_tsql/src/backend_parser/gram-tsql-rule.y
```diff
@@ -268,6 +268,26 @@ tsql_CreateRoleStmt:
 				}
 			;
 
+tsql_CreatedbStmt:
+			CREATE DATABASE name opt_with createdb_opt_list
+				{
+					CreatedbStmt  *n = makeNode(CreatedbStmt);
+					n->dbname = $3;
+
+					/* If there are specified options, this is PSQL syntax */
```
HarshLunagariya: In case of TSQL CREATE DB, We can get non-null options list as well. We should remove this comment.

contrib/babelfishpg_tsql/src/catalog.h
```diff
@@ -39,9 +39,10 @@ extern Oid	sysdatabaese_idx_oid_oid;
 extern Oid	sysdatabaese_idx_name_oid;
 
 /* MUST comply with babelfish_sysdatabases table */
-#define SYSDATABASES_NUM_COLS 8
+#define SYSDATABASES_NUM_COLS 9
 #define Anum_sysdatabaese_oid 1
 #define Anum_sysdatabaese_name 6
+#define Anum_sysdatabaese_name_original 9
```
HarshLunagariya: Can we keep it xxxx_original_name same as column name?

contrib/babelfishpg_tsql/src/dbcmds.c
```diff
@@ -368,7 +369,7 @@ do_create_bbf_db(const char *dbname, List *options, const char *owner)
 	int16		dbid;
```
HarshLunagariya, line 373: Let's get rid of multiple downcase_identifier() calls, since case sensitive db_name is only required for only one instance.

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -471,7 +471,39 @@ pltsql_bbfCustomProcessUtility(ParseState *pstate, PlannedStmt *pstmt, const cha
...
+						foreach(option, db_options)
+							{
+								stmt->options = list_delete_ptr(stmt->options,
+																lfirst(option));
+							}
+						
+						pfree(stmt->dbname);
+						stmt->dbname = convertToUPN(orig_dbname);
+		
+						}
```
HarshLunagariya: Fix indent

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -471,7 +471,39 @@ pltsql_bbfCustomProcessUtility(ParseState *pstate, PlannedStmt *pstmt, const cha
 		{
 			if (sql_dialect == SQL_DIALECT_TSQL)
 			{
-				create_bbf_db(pstate, (CreatedbStmt *) parsetree);
+					CreatedbStmt *stmt = (CreatedbStmt *) parsetree;
+					List	   *db_options = NIL;
+					ListCell   *option;
+
+					/* Check if creating login or role. Expect islogin first */
```
HarshLunagariya: Unrelated comment.

HarshLunagariya (changes requested):

contrib/babelfishpg_tsql/sql/ownership.sql
anju15bharti: done

contrib/babelfishpg_tsql/src/backend_parser/gram-tsql-rule.y
anju15bharti: done

contrib/babelfishpg_tsql/src/catalog.h
anju15bharti: done

contrib/babelfishpg_tsql/src/hooks.c
anju15bharti: done

contrib/babelfishpg_tsql/src/hooks.c
anju15bharti: removed

contrib/babelfishpg_tsql/sql/babelfishpg_tsql.sql
```diff
@@ -1965,8 +1965,8 @@ BEGIN
 		THROW 33557097, N'Primary or foreign key table name must be given.', 1;
 	END
 	
-	IF (@pktable_qualifier != '' AND (SELECT sys.db_name()) != @pktable_qualifier) OR 
-		(@fktable_qualifier != '' AND (SELECT sys.db_name()) != @fktable_qualifier) 
+	IF (@pktable_qualifier != '' AND (SELECT LOWER(sys.db_name())) != @pktable_qualifier) OR
+		(@fktable_qualifier != '' AND (SELECT LOWER(sys.db_name())) != @fktable_qualifier)
```
Deepesh125: should use COLLATE sys.database_default

contrib/babelfishpg_tsql/sql/babelfishpg_tsql.sql
```diff
@@ -2024,7 +2024,7 @@ WHERE has_schema_privilege(s1.schema_id, 'USAGE')
 
 UNION 
 
-SELECT CAST((SELECT sys.db_name()) AS sys.sysname) COLLATE sys.database_default AS PROCEDURE_QUALIFIER,
+SELECT CAST((SELECT LOWER(sys.db_name())) AS sys.sysname) COLLATE sys.database_default AS PROCEDURE_QUALIFIER,
```
Deepesh125: why to convert this to lower?

contrib/babelfishpg_tsql/sql/babelfishpg_tsql.sql
```diff
@@ -2281,9 +2281,9 @@ BEGIN
 		LEFT OUTER JOIN sys.babelfish_authid_user_ext AS Ext2 ON Base2.rolname = Ext2.rolname
 		LEFT OUTER JOIN sys.babelfish_authid_login_ext As LogExt ON LogExt.rolname = Ext1.login_name
 		LEFT OUTER JOIN pg_catalog.pg_roles AS Base3 ON Base3.rolname = LogExt.rolname
-		LEFT OUTER JOIN sys.babelfish_sysdatabases AS Bsdb ON Bsdb.name = DB_NAME()
+		LEFT OUTER JOIN sys.babelfish_sysdatabases AS Bsdb ON Bsdb.name = LOWER(DB_NAME())
```
Deepesh125: use `collate sys.database_default` everywhere instead of converting db_name to lowercase

contrib/babelfishpg_tsql/sql/babelfishpg_tsql.sql
```diff
@@ -2605,7 +2605,7 @@ GRANT SELECT ON sys.syslogins TO PUBLIC;
 CREATE OR REPLACE VIEW sys.sp_sproc_columns_view
 AS
 SELECT
-CAST(sys.db_name() AS sys.sysname) AS PROCEDURE_QUALIFIER -- This will always be objects in current database
+CAST(LOWER(sys.db_name()) AS sys.sysname) AS PROCEDURE_QUALIFIER -- This will always be objects in current database
```
Deepesh125: Is this really needed and expected? I don't think so

contrib/babelfishpg_tsql/sql/information_schema_tsql.sql
Deepesh125: +1

contrib/babelfishpg_tsql/sql/ownership.sql
```diff
@@ -9,6 +9,7 @@ CREATE TABLE sys.babelfish_sysdatabases (
 	name TEXT NOT NULL COLLATE "C",
 	crdate timestamptz NOT NULL,
 	properties TEXT NOT NULL COLLATE "C",
+	orig_name TEXT NOT NULL COLLATE "C",
```
Deepesh125: why COLLATE "C"?

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -471,7 +471,32 @@ pltsql_bbfCustomProcessUtility(ParseState *pstate, PlannedStmt *pstmt, const cha
...
+				if (stmt->options != NIL)
+				{
+					char	   *orig_dbname = NULL;
+					foreach(option, stmt->options)
+					{
+						DefElem    *defel = (DefElem *) lfirst(option);
+						if (strcmp(defel->defname, "name_location") == 0)
+						{
+							int			location = defel->location;
+							orig_dbname = extract_identifier(queryString + location);
```
Deepesh125, line 487: this logic may not work well with multibyte chars, have we tested it with Japanese, Chines, etc chars?

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -471,7 +471,32 @@ pltsql_bbfCustomProcessUtility(ParseState *pstate, PlannedStmt *pstmt, const cha
...
+						}
+					}
+					
+					foreach(option, db_options)
+					{
+						stmt->options = list_delete_ptr(stmt->options,lfirst(option));
+					}
+					
+					pfree(stmt->dbname);
+					stmt->dbname = convertToUPN(orig_dbname);
```
Deepesh125: why?

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -283,6 +283,46 @@ get_db_id(const char *dbname)
 	return db_id;
 }
 
+char *
+get_db_original_name(const char *dbname)
```
Deepesh125: Add one liner explanation for these two functions

contrib/babelfishpg_tsql/src/dbcmds.c
```diff
@@ -408,6 +409,7 @@ create_bbf_db_internal(const char *dbname, List *options, const char *owner, int
 	const char *guest;
 	const char *prev_current_user;
 	int			stmt_number = 0;
+	char	   *lower_dbname = downcase_identifier(dbname, strlen(dbname), false, false);
```
Deepesh125: what would happen when you supply dbname whose length is > 64 bytes?

Deepesh125 (changes requested):

contrib/babelfishpg_tsql/sql/babelfishpg_tsql.sql
rishabhtanwar29: +1, isn't this view needs to return the database name in upper case? Correct me if I am wrong but I think all the T-SQL procedures/views return original name (case sensitive). Let's verify this if not already.

contrib/babelfishpg_tsql/sql/babelfishpg_tsql.sql
```diff
@@ -3592,7 +3592,7 @@ BEGIN
 		CAST(LEFT(p.loginname,40) AS sys.VARCHAR(40)) AS loginame,
 		CAST(LEFT(p.hostname,60) AS sys.VARCHAR(60)) AS hostname,
 		p.blocked AS blk, 
-		CAST(LEFT(db_name(p.dbid),40) AS sys.VARCHAR(40)) AS dbname,
+		CAST(LEFT(LOWER(db_name(p.dbid)),40) AS sys.VARCHAR(40)) AS dbname,
```
rishabhtanwar29: Again, have we verified if sys.sp_who returns lower-cased dbname according to T-SQL documentation?

contrib/babelfishpg_tsql/sql/ownership.sql
rishabhtanwar29: I guess it's fine to use orig_name in the view so that it's dependencies can output the name in original case but we need to make sure that the dbname comparison case insensitive everywhere wherever we are using this view.

Kushaal-EDIT: @rishabhtanwar29 we should do comparison based on collation, check my comment

contrib/babelfishpg_tsql/src/catalog.c
rishabhtanwar29: +1

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -283,6 +283,46 @@ get_db_id(const char *dbname)
...
+
+	name_datum = SysCacheGetAttr(SYSDATABASEOID, tuple, Anum_sysdatabaese_orig_name, &isNull);
+	original_name = TextDatumGetCString(name_datum);
+	ReleaseSysCache(tuple);
+
+	return original_name;
+}
+
+char *
+get_original_db_name(int16 dbid)
```
rishabhtanwar29: This and the above function names sound quite similar, looks like we trying to confuse the reader? ;-). Anyway, let's do better at naming these two, something like: `dbname_get_original_name` and `dbid_get_original_name` so that name gives a hint about what it does.

test/JDBC/upgrade/15_4/schedule
```diff
@@ -60,7 +60,7 @@ BABEL-3000
 BABEL-3000-dep
 BABEL-3010
 BABEL-3116
-BABEL-3117
+BABEL-3117-before-latest
```
rishabhtanwar29: There is nothing like `before-latest`, latest is always varying. Should be `before-15_5`

contrib/babelfishpg_tsql/sql/babelfishpg_tsql.sql
KushaalShroff: @Deepesh125 Will sys.babelfish_sysdatabases AS Bsdb ON Bsdb.name be case sensitive?
If it doesnt preserve casing then we need to lower for comparison

test/JDBC/upgrade/latest/schedule
```diff
@@ -187,7 +187,6 @@ babel_varbinary_int4_div
 bbf_view_def
 binary-index
 bitwise_not-operator
-case_insensitive_collation
```
rishabhtanwar29, line 190: Why have we removed it from latest schedule? issue is fixed now so the result should be consistent from current version and all future versions.

contrib/babelfishpg_tsql/sql/babelfishpg_tsql.sql
KushaalShroff: @anju15bharti rather we might want to use orginal name column for comparison with default collation

test/JDBC/upgrade/13_6/preparation/case_insensitive_collation-vu-prepare.sql
```diff
@@ -0,0 +1,32 @@
+-- BABEL-3094
```
rishabhtanwar29: I am not quite sure why do we need version specific test files here, shouldn't before-XX and after-XX upgrade file convention work here?

contrib/babelfishpg_tsql/sql/information_schema_tsql.sql
KushaalShroff: By this logic user queries like select * from information_schema.schemata where CATALOG_NAME = db_name() will stop working for case sensitive db names

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--3.3.0--3.4.0.sql
```diff
@@ -657,5 +657,1012 @@ $BODY$
 LANGUAGE plpgsql
 IMMUTABLE;
 
+SET allow_system_table_mods = on;
+
+ALTER TABLE sys.babelfish_sysdatabases ADD COLUMN IF NOT EXISTS orig_name TEXT COLLATE "C";
+
+UPDATE sys.babelfish_sysdatabases SET orig_name = name WHERE orig_name IS NULL;
+
+ALTER TABLE sys.babelfish_sysdatabases ALTER COLUMN orig_name SET NOT NULL;
```
rishabhtanwar29: Now this kind of changes make multi-version dump/restore complicated since we are adding a new column and expect it to be NOT NULL (unfortunately we have done this in past). Is it possible to skip this and not make this column NOT NULL. As a result we should be able to write the logic in system views so that if `orig_name` is NULL than take it from `dbname`?

contrib/babelfishpg_tsql/sql/ownership.sql
KushaalShroff: All system objects shall now return case sensitive catalog name, I think what Anju did is right.
Since user queries like select * from information_schema.tables where TABLE_CATALOG = db_name() will stop working for case sensitive db names (though this query does not makes sense but is still valid).
By the above change, all objects consuming ``sys.pg_namespace_ext`` to fetch orig_name will get the case sensitive names.

contrib/babelfishpg_tsql/sql/babelfishpg_tsql.sql
KushaalShroff: ++We should not lower

contrib/babelfishpg_tsql/src/catalog.c
KushaalShroff: The function names are confusing, why cant we name it appropriately as per input/arg type

KushaalShroff (changes requested):

contrib/babelfishpg_tsql/src/hooks.c
HarshLunagariya: +1, Please check convertToUPN(). it's not applicable here.

PR #1803, opened 2023-08-30
DROP TABLE & DROP TABLE IF EXISTS table1, table2 fix

tanscorpio7:
### Description

DROP TABLE IF EXISTS table1, table2 fails and crashes server if multiple table names are provided and a table which actually exist is first in the list followed by atleat one table name which does not actually exist.

1.   If --enable-cassert was set while postgres configure -----> a server crash will be seen on drop table command
2.   If --enable-cassert was not set ( that is assertions are off ) -----> error when dropping tables 'relcache reference testtable is not owned by resource owner TopTransaction'

bbf_ExecDropStmt -> function for dropping extended properties for objects ( e.x. called along with DROP TABLE )
A foreach loop is used for each table included in the drop table statement.  Every loop tries to fetch relation for given object.
Currently when object type is TABLE, the get_object_address is called which sets the relation for the current object, but relation is not set to NULL before the next call to get_object_address in next loop. As a result relation continues with the previous loop value if not set in current loop (ideally relation should be set to null for non existent objects). This later leads to assertion error when trying to close the relation.

FIX :
relation set to NULL at the beginning of each loop.
If relation is null for current object we skip the current loop. 

### Issues Resolved

[BABEL-4367]

1. DROP TABLE 
--> check with single existing table
--> check with multiple table ( existing and non existing ) --error expected 

2. DROP TABLE IF EXISTS
--> check with multiple tables existing and non existing table, existing first
--> non existing first

3 DROP PROCEDURE / DROP PROCEDURE IF EXISTS
4 DROP VIEW / DROP VIEW IF EXISTS
5 DROP TYPE / DROP TYPE IF EXISTS
5 DROP FUNCTION / DROP FUNCTION IF EXISTS

### Check List


contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -5447,14 +5447,15 @@ bbf_ExecDropStmt(DropStmt *stmt)
 	{
 		foreach(cell, stmt->objects)
 		{
+			relation = NULL;
```
surendravishnoi, line 5450: We are also returning at line 5438, please see if problem can happen there where we bail out early if an schema is non-existent

Also look from line 5500 as well, see if we have similar problem or not

test/JDBC/expected/BABEL_4367.out
```diff
@@ -0,0 +1,47 @@
+CREATE TABLE BABEL_4367_1 (
+    id INT
+)
+GO
+
+DROP TABLE BABEL_4367_1, BABEL_4367_2
```
surendravishnoi, line 13: let us also add test case where we have 3 tables and middle one does not exists, first one does not exists (both with and without if exists)

contrib/babelfishpg_tsql/src/pl_handler.c
tanscorpio7: When case is schema: We only have one object
Since DROP SCEHMA allows only one entry.

For procedure, functions, will need changes

tanscorpio7: made changes at two places
1st one covers case for table, view, sequence
2nd one covers procedure, function, type

for schema no change is needed since only one schema is allowed to be dropped at a time

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -5499,6 +5500,7 @@ bbf_ExecDropStmt(DropStmt *stmt)
 
 		foreach(cell, stmt->objects)
 		{
+			relation = NULL;
```
surendravishnoi, line 5503: let us add tests for TYPE as well. Also, add test for schemas (multiple schema as well even if it gives error)

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -5516,7 +5518,7 @@ bbf_ExecDropStmt(DropStmt *stmt)
 			if (!HeapTupleIsValid(tuple))
 			{
 				table_close(relation, AccessShareLock);
-				return;
+				continue;
```
surendravishnoi, line 5521: should we continue at 5511 as well?

contrib/babelfishpg_tsql/src/pl_handler.c
tanscorpio7: done

test/JDBC/expected/BABEL_4367.out
```diff
@@ -246,10 +246,99 @@ varchar
...
+CREATE TYPE BABEL_4367_TYPE_1 from datetime
+GO
+
+DROP TYPE BABEL_4367_TYPE_1, BABEL_4367_TYPE_2
+GO
+~~ERROR (Code: 33557097)~~
+
+~~ERROR (Message: syntax error near ',' at line 1 and character position 27)~~
+
+DROP TYPE IF EXISTS BABEL_4367_TYPE_1, BABEL_4367_TYPE_2
```
surendravishnoi, line 259: why is there syntax error, atleast PG allows it

test/JDBC/expected/BABEL_4367.out
```diff
@@ -246,10 +246,99 @@ varchar
...
+DROP TYPE BABEL_4367_TYPE_1
+GO
+
+CREATE TYPE BABEL_4367_TYPE_1 from datetime
+GO
+
+DROP TYPE IF EXISTS BABEL_4367_TYPE_1
+GO
+
+SELECT name FROM sys.views WHERE name in ('BABEL_4367_TYPE_1', 'BABEL_4367_TYPE_2') ORDER BY name ASC
```
surendravishnoi: why selecting from sys.views?

test/JDBC/expected/BABEL_4367.out
```diff
@@ -246,10 +246,99 @@ varchar
...
+
+
+DROP PROCEDURE BABEL_4367_PROCEDURE_3, BABEL_4367_PROCEDURE_1
+GO
+~~ERROR (Code: 3701)~~
+
+~~ERROR (Message: could not find a procedure named "babel_4367_procedure_3")~~
+
+
+DROP PROCEDURE IF EXISTS BABEL_4367_PROCEDURE_1, BABEL_4367_3
```
surendravishnoi: are we mixing table here BABEL_4367_3?

test/JDBC/expected/BABEL_4367.out
tanscorpio7: SQL server does not allow multiple TYPE drop

surendravishnoi (approved):

Issue #1571, opened 2023-06-21
[Bug]: Error with implicit cast for type BIT in select parameters


staticlibs:
### What happened?

Type [BIT](https://learn.microsoft.com/en-us/sql/t-sql/data-types/bit-transact-sql?view=sql-server-ver16) is expected to recognize "string values TRUE and FALSE".

Implicit cast to `bit` for `true` and `false` literals appeared to work in most scenarios, but it fails when such literal is passed as a parameter of type `varchar` in `select` query parameters.

With this table:

```
create table tab1(name1 varchar(42), flag1 bit)
```

The following works correctly:

```
insert into tab1 values('foo', 'true')

declare @out int
exec sp_prepexec @out output, N'@P1 varchar(43)', N'insert into tab1 values(''foo'', @P1)', @P1 = 'true'
exec sp_unprepare @out

select * from tab1 where flag1 = 'true'
```

But the following fails:

```
declare @out int
exec sp_prepexec @out output, N'@P1 varchar(43)', N'select * from tab1 where flag1 = @P1', @P1 = 'true'
exec sp_unprepare @out

ERROR: invalid input syntax for type integer: "true"
```

This breaks existing client code  that passes `true` string parameters like this:

```
conn = pyodbc.connect(...)
cur = conn.cursor()
cur.execute("select * from tab1 where flag1 = ?", ["true"])

pyodbc.ProgrammingError: ('42000', '[42000] [Microsoft][ODBC Driver 18 for SQL Server][SQL Server]invalid input syntax for type integer: "true" (33557097) (SQLExecDirectW)')
```

Note, the problem only happens when `true` literal is passed as a parameter of type `varchar`, it works correctly if the same literal is passed as a parameter of type `bit`:

```
declare @out int
exec sp_prepexec @out output, N'@P1 bit', N'select * from tab1 where flag1 = @P1', @P1 = 'true'
exec sp_unprepare @out
```

The problem happens because the following cast:

```
CREATE CAST (sys.VARCHAR AS INT4)
WITH FUNCTION sys.varchar2int4(sys.VARCHAR) AS IMPLICIT;
```

is incorrectly taken instead of the correct one:

```
CREATE CAST (sys.VARCHAR AS sys.BIT)
WITH FUNCTION sys.varchar2bit(sys.VARCHAR) AS IMPLICIT;
```

[...]

PR #1806, opened 2023-08-30
Fix wrong operator for bit=varchar

wboettge:
### Description

BBF has operators for BIT=INT and BIT=BIT. It must chose to cast the varchar to BIT or to INT. Currently it picks INT as it has a higher precedence, but the cast fails. We should prioritize same-typed operators when picking which to use.

* Checks issue where 'true' was cast to int instead of bit
* Checks the casting applied to cases with date and money types, which also use cross-type operators
* Checks a case where the operator selected for int to varbinary changed. Output type is now int, and out-of-range int consts must now be explicitly cast to bigint, matching T-SQL.

### Issues Resolved

Fixes Issue #1571 
BABEL-4264

### Check List


test/JDBC/input/BABEL-4264.sql
```diff
@@ -0,0 +1,20 @@
+create table babel4264(name1 varchar(42), flag1 bit)
+go
+
+select * from babel4264 where flag1 = CAST('true' as VARCHAR(20))
```
Deepesh125, line 11: Please insert some data

test/JDBC/expected/babel_operators.out
```diff
@@ -3,12 +3,12 @@ SELECT (123 + 0x42);
 SELECT (0x42 + 123);
 GO
 ~~START~~
-bigint
+int
```
Deepesh125, line 6: Make sure that these all changes are expected T-SQL behaviour

contrib/babelfishpg_tsql/src/pltsql_coerce.c
```diff
@@ -742,6 +744,19 @@ tsql_has_func_args_higher_precedence(int n, Oid *inputtypes, FuncCandidateList c
...
 		return true;
 
+	/* Prioritize candidates with same-typed arguments for operators only*/
+	if (candidates_are_opers)
+	{
+		for (i = 1; i < n; ++i)
+		{
+			can1_is_sametype &= argtypes1[i-1] == argtypes1[i];
+			can2_is_sametype &= argtypes2[i-1] == argtypes2[i];
+		}
```
Deepesh125, line 759: there could be unary/prefix operators possible, should we handle it?

Deepesh125 (dismissed):

contrib/babelfishpg_tsql/src/pltsql_coerce.c
wboettge: If we have a unary operator, then n should be 1, meaning we won't enter the for loop and can1/2_is_sametype are both true.

test/JDBC/expected/babel_operators.out
wboettge: T-SQL does choose int as the result type here. Note that as a result of using INT instead of BIGINT for the result type, we can see integer overflows that didn't previously occur in babelfish. But, these overflows match T-SQL behavior.

forestkeeper (approved):

PR #1810, opened 2023-09-01
Allow users to directly insert binary data in octal format from the PG endpoint

Sairakan:
### Description

When connecting through the PG endpoint and attempting to insert binary data in octal form into a table, the data gets incorrectly interpreted as an ASCII string and becomes jumbled garbage. This change allows users to directly insert binary data as octal strings from the PG endpoint.

### Issues Resolved

BABEL-4377

* **Use case based -**

* **Boundary conditions -**

* **Arbitrary inputs -**

* **Negative test cases -**

* **Performance tests -**

* **Tooling impact -**

* **Client tests -**

### Check List


contrib/babelfishpg_common/src/varbinary.c
```diff
@@ -34,6 +34,19 @@
...
+ * that we can reference pltsql_protocol_plugin_ptr to check whether
+ * we are coming from the TDS or PG client. This is required to support
+ * COPY inputs coming from DMS.
+ */
+typedef struct PLtsql_protocol_plugin
+{
+	/* True if Protocol being used by client is TDS. */
+	bool		is_tds_client;
+} PLtsql_protocol_plugin;
+PLtsql_protocol_plugin **pltsql_protocol_plugin_ptr;
```
KushaalShroff: Can this be done in a header file in common?
Also why is it a double pointer?

contrib/babelfishpg_common/src/varbinary.c
```diff
@@ -190,6 +203,12 @@ varbinaryin(PG_FUNCTION_ARGS)
 		PG_RETURN_BYTEA_P(result);
 	}
 
+	if ((!(*pltsql_protocol_plugin_ptr && (*pltsql_protocol_plugin_ptr)->is_tds_client)) &&
```
KushaalShroff: Your condition will be true if pltsql_protocol_plugin_ptr will be NULL as well which is wrong behaviour

contrib/babelfishpg_common/src/varbinary.c
```diff
@@ -190,6 +203,12 @@ varbinaryin(PG_FUNCTION_ARGS)
 		PG_RETURN_BYTEA_P(result);
 	}
 
+	if ((!(*pltsql_protocol_plugin_ptr && (*pltsql_protocol_plugin_ptr)->is_tds_client)) && 
+		(len >= 4 && inputText[0] == '\\')) /* COPY command from DMS, use base byteain() directly */
```
KushaalShroff: Can you explain why are you checking for an arbit len >=4?

contrib/babelfishpg_common/src/varbinary.c
```diff
@@ -34,6 +34,19 @@
 
 #include "instr.h"
 
+/* We need to create a minimal stub of PLtsql_protocol_plugin so
+ * that we can reference pltsql_protocol_plugin_ptr to check whether
+ * we are coming from the TDS or PG client. This is required to support
+ * COPY inputs coming from DMS.
```
KushaalShroff: Such a comment should not go in opensource! Please check the guidelines

contrib/babelfishpg_common/src/varbinary.c
```diff
@@ -190,6 +203,12 @@ varbinaryin(PG_FUNCTION_ARGS)
 		PG_RETURN_BYTEA_P(result);
 	}
 
+	if ((!(*pltsql_protocol_plugin_ptr && (*pltsql_protocol_plugin_ptr)->is_tds_client)) && 
+		(len >= 4 && inputText[0] == '\\')) /* COPY command from DMS, use base byteain() directly */
+	{
+		PG_RETURN_BYTEA_P(DirectFunctionCall1(byteain, (Datum) inputText));
```
KushaalShroff: After this change, is the data stored/retrieved during COPY from PSQL and insert from TDS have no difference? Did we validate this?

KushaalShroff (changes requested):
Even simple insert from PSQL should trigger this change, We should add tests for them.

KushaalShroff:
The title of your PR is also misleading, we were able to insert "binary data" before as well. BYTEA is a stream of bytes and not the actual binary value, so you are now supporting BYTEA values to be inserted in Binary columns (from pg endpoint).

contrib/babelfishpg_common/src/varbinary.c
```diff
@@ -34,6 +34,19 @@
...
+/* We need to create a minimal stub of PLtsql_protocol_plugin so
+ * that we can reference pltsql_protocol_plugin_ptr to check whether
+ * we are coming from the TDS or PG client. This is required to support
+ * COPY inputs coming from DMS.
+ */
+typedef struct PLtsql_protocol_plugin
+{
+	/* True if Protocol being used by client is TDS. */
+	bool		is_tds_client;
+} PLtsql_protocol_plugin;
```
KushaalShroff: I am still not sure how this is working, even in tds we are including the same plugins.
Check tds_int.h
```
#include "../../contrib/babelfishpg_tsql/src/pltsql.h"
#include "../../contrib/babelfishpg_tsql/src/pltsql-2.h"
```

You have not linked the plugins appropriately. Check my comment where pltsql_protocol_plugin_ptr being NULL will be the reason why its working now

Also we are adding dependency between TSQL and common, have we consulted any senior dev? Please include @surendravishnoi

contrib/babelfishpg_common/src/varbinary.c
Sairakan: I have manually validated that the data is identical between COPY and a normal TDS insert of the equivalent data, e.g.
TDS> INSERT INTO t1 VALUES (51)
PG> COPY dbo.t1 FROM STDIN CSV; -- insert \000\000\000\000\000\000\000\000\000\063
both end up with the same data in the table:
```
data                   
----------------------
0x00000000000000000033
```

contrib/babelfishpg_common/src/varbinary.c
Sairakan: This was a safeguard since any octal data must be at minimum `\xxx', so I wanted to a) avoid a segfault in case the string is null or invalid, and b) ensure that it is being executed against a somewhat sane string

contrib/babelfishpg_common/src/varbinary.c
Sairakan: As mentioned, this is a stub that is intended to recreate what is in `pltsql.h` from babelfishpg_tsql. It could be added in a header file, but I think putting it here minimizes the potential blast radius since this is technically a global variable, but I would not want it to be accessible outside of this file.

contrib/babelfishpg_common/src/varbinary.c
Sairakan: This is simply a copy of the `IS_TDS_CLIENT()` macro defined in `pltsql.h`:
```
#define IS_TDS_CLIENT() (*pltsql_protocol_plugin_ptr && \
						 (*pltsql_protocol_plugin_ptr)->is_tds_client)
```
I don't see why it is wrong to have the condition be true if the pointer is NULL - the point is that if we are not in a TDS connection, we want to use the default BYTEA conversion.

contrib/babelfishpg_common/src/varbinary.c
Sairakan: Just checked, and there is one behavior change with this - from the PG endpoint, inserting into a table with an octal string will now use the BYTEA conversion rather than the VARBINARY conversion - however, this can be worked around by directly inserting the VARBINARY value. See below:
```
-- old behavior
babelfish_db=> insert into dbo.t1 values ('\000\063');
INSERT 0 1
babelfish_db=> select * from dbo.t1;
        data
--------------------
 0x5c3030305c303633
(1 row)

-- new behavior
jdbc_testdb=# insert into dbo.t1 values ('\000\063');
INSERT 0 1
jdbc_testdb=# insert into dbo.t1 values ('\x5c3030305c303633');
INSERT 0 1
jdbc_testdb=# select * from dbo.t1;
        data        
--------------------
 0x0033
 0x5c3030305c303633
(2 rows)
```

Ultimately, I don't think this is too much of an issue since this only affects inserting data from the PG endpoint and has a workaround, but it may need to be documented.

contrib/babelfishpg_common/src/varbinary.c
Sairakan: I tried including the headers like you mentioned, but this results in a build failure while trying to load the extension:
```
NOTICE:  installing required extension "babelfishpg_tsql"
NOTICE:  installing required extension "uuid-ossp"
NOTICE:  installing required extension "babelfishpg_common"
ERROR:  could not load library "/local/jasonten/BabelOpenSource/src/postgres/lib/babelfishpg_common.so": /local/jasonten/BabelOpenSource/src/postgres/lib/babelfishpg_common.so: undefined symbol: pltsql_protocol_plugin_ptr
GRANT ALL ON SCHEMA sys to jdbc_user;
ERROR:  schema "sys" does not exist
```

I agree that this is a potentially risky method to getting this code to work, but so far it is the only solution I have found.

Sairakan: Updated to use a hook function. Note that this means that there is now an accompanying change in the engine code to define the hook.

contrib/babelfishpg_common/src/varbinary.c
KushaalShroff: The workaround you mentioned doesnt look feasible and not user friendly. How is the user gonna derive the binary value always?

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -91,6 +91,14 @@ static ObjectAddress get_trigger_object_address(List *object, Relation *relp, bo
 Oid			get_tsql_trigger_oid(List *object, const char *tsql_trigger_name, bool object_from_input);
 static Node *transform_like_in_add_constraint(Node *node);
 
+/*****************************************
+ * 			Protocol Hooks
+ *****************************************/
+
+IsTdsClientHookType PrevIsTdsClientHook = NULL;
```
KushaalShroff: We cannot have such a hook in opensource engine, please consult a senior dev like @kuntalghosh @surendravishnoi

contrib/babelfishpg_common/src/varbinary.c
```diff
@@ -190,6 +191,13 @@ varbinaryin(PG_FUNCTION_ARGS)
 		PG_RETURN_BYTEA_P(result);
 	}
 
+	if (!(IsTdsClientHook && (*IsTdsClientHook)()) &&
```
KushaalShroff: What you did before - using a rendevu variable - was better than adding a hook between 2 extensions.
And we are doing the same mistake here that we are gonna use byteain for IsTdsClientHook == NULL, why would we want to keep that behvaiour in varbinaryin? We should not be intrusive with our changes, you are technically making varbinary unusable for people wanting to use only common extension. 
With your change anyone having only common extension and using binary will now be using byteain which is absolutely not what it is intended for

KushaalShroff (changes requested):

contrib/babelfishpg_common/src/varbinary.c
KushaalShroff: Can you create a macro rather than hardcoded 4

contrib/babelfishpg_common/src/varbinary.c
KushaalShroff: We are defeating the purpose of varbinary datatype altogether, check my other comment where I have explained this in detail

contrib/babelfishpg_common/src/varbinary.c
surendravishnoi: we can have direct hook between common and TDS/TSQL using rendevouz variables. No need to go via engine. Also, ideal way to do it is

Find out what logic is specific to TDS vs PSQL
Add a function in TDS to do TDS specific work
Add a rendevouz variable between common and TDS to access the above function
in common
if (tds_hook && tds_hook())
else
   PSQL logic

However, if we have time constraints, just add a hook via rendevouz variable

contrib/babelfishpg_tsql/src/hooks.c
Sairakan: This hook was added at the suggestion of Surendra. I will contact Kuntal to get his opinion as well.

contrib/babelfishpg_common/src/varbinary.c
Sairakan: added a macro for this

contrib/babelfishpg_common/src/varbinary.c
Sairakan: updated to use rendezvous variable instead

contrib/babelfishpg_common/src/varbinary.c
Sairakan: Reworked to use a rendezvous variable again.

contrib/babelfishpg_tsql/src/hooks.c
Sairakan: Reworked to use a rendezvous variable, no longer need the hook

contrib/babelfishpg_common/src/varbinary.c
Sairakan: Oh cool, I didn't know about the existence of rendezvous variables. I'll update the implementation to use them.

contrib/babelfishpg_common/src/varbinary.c
Sairakan: The binary value can be derived by doing an ASCII to hex conversion - for example, they could use https://www.rapidtables.com/convert/number/ascii-to-hex.html to get the hex value to insert.

contrib/babelfishpg_common/src/varbinary.c
Sairakan: The entire point is to use byteain when connecting through the PG endpoint - I just tested this, and even with all extensions installed, when connecting through the PG endpoint `PLtsql_protocol_plugin **protocol_ptr` is NULL. So there is no way to determine whether someone is using only the common extension or not. As mentioned before, there is a workaround if the user manually converts the input to hex first and inserts that as a BYTEA directly.

contrib/babelfishpg_common/src/varbinary.c
Sairakan: Updated to use the existing GUC `psql_logical_babelfish_db_name` - this is a DMS-specific GUC that is not publicly documented anywhere, so it is extremely unlikely that anyone would accidentally set this GUC (unless they were inspecting the open source code). Chatted with Surendra and Bill about this, and they both agree that this is a sufficient restriction on the behavior override.

contrib/babelfishpg_common/src/varbinary.c
Sairakan: Updated to check `psql_logical_babelfish_db_name` GUC instead

contrib/babelfishpg_common/src/varbinary.c
Sairakan: Removed this code, as a hook/rendezvous function is not needed anymore since we now just check a GUC.

PR #1812, opened 2023-09-04
Implemented fix for when column is enclosed in parens (round brackets) and single byte as well as multibyte table and column aliases. 

riyajain39:
### Description

This pull request fix the issue when column name is enclosed in parens (round brackets) and no alias is specified. It also fix the implementation for column and table aliases contain both single byte and multibyte characters whose length less than or more than 64 bytes.

**Issue 1:** Handling of column name when enclosed by any number of parens.
When column is enclosed in parens (round brackets) and no alias is specified, then the column name in result set is returned incorrectly. In such cases the actual column name should come, instead from first characters of query till length of column name. To resolve this issue I added a condition in [pre_transform_target_entry](https://github.com/babelfish-for-postgresql/babelfish_extensions/blob/BABEL_3_X_DEV/contrib/babelfishpg_tsql/src/hooks.c#L1421) that will handle multiple levels of parens and spaces in-between parens.

**Issue 2 :** Handle column and table aliases. 
Column and table alias does not work when alias length is more than or equals to 64 bytes in case of multibyte aliases. When query select 1 as '您对“数据一览“中的车型，颜色，内饰，选装, '; executes through T-SQL endpoint, it causes client to hang and throws protocol error in TDS stream. The logic of pre_transform_target_entry was not handling the aliases contain multibyte characters whose length is more than 64 bytes accurately and overwriting the alias name in case of truncation. To address this problem, I redefined the function and implemented the conditions which will extract identifier name using extract_identifier() and handle both the case when aliases are not delimited as well as delimited with square bracket, single quotes and double quotes.

**Issue 3 :** Aliases with multibyte chars cause MVU failure.
When establishing a view of alias with multibyte chars through T-SQL endpoint and retrieving it's definition from PG end point, extra unreadable chars are appearing at the end of the alias and even we can not use this view through tsql endpoint as it throws protocol error in TDS stream. But while we use similar alias on PG end point, we are able to retrieve it's correct definition. This discrepancy in behaviour is causing MVU failure. After implementing the fix, the correct alias definition is obtained from the PG endpoint, which rectifies the issue.

 
### Issues Resolved

BABEL-4384
BABEL-4231

Signed-off-by: Riya Jain [riyaajn@amazon.com](mailto:riyaajn@amazon.com)

* **Use case based -** Added test case from original bug.

* **Boundary conditions -**

* **Arbitrary inputs -**

* **Negative test cases -**

* **Major version upgrade tests -** Added testcases for 14.latest to 15.latest. 

* **Performance tests -**

* **Tooling impact -**

* **Client tests -**

### Check List


contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -1447,70 +1447,64 @@ pre_transform_target_entry(ResTarget *res, ParseState *pstate,
...
+                memcpy(alias, identifier_name, alias_len);
+            }
+            else
+            {
+                if(sq)
+                {
+                    colname_start++;
+                }
+                /*
+                * After truncation, maximum truncated length of alias can be 60
```
Deepesh125: add details bit more here, it could be 61 to 63 in case of multibyte chars and we need to handle those cases only.

.github/configuration/upgrade-test-configuration.yml
```diff
@@ -76,7 +76,7 @@ upgrade-version: [[
 ],
 [
   {
-    version: 14.9,
+    version: '14.10',
```
Deepesh125, line 79: Make sure we get approval for this change from @rohit01010 or @basasairohan

test/JDBC/expected/14_10__preparation__BABEL-2877-vu-prepare.out
```diff
@@ -0,0 +1,100 @@
+-- Function
+-- Defaults at different positions
```
Deepesh125, line 2: Make sure we get approval from original writer of BABEL-2877 tests, same for other test files

test/JDBC/upgrade/14_10/schedule
```diff
@@ -409,3 +407,4 @@ babel_varbinary_int4_div
 sys-sql_expression_dependencies
 smalldatetimefromparts-dep
 BABEL_4330
+BABEL-4231
```
Deepesh125: what about upgrade tests for BABEL-4384

Deepesh125 (changes requested):

.github/configuration/upgrade-test-configuration.yml
riyajain39: Yes, I have confirmed from them.

test/JDBC/expected/14_10__preparation__BABEL-2877-vu-prepare.out
riyajain39: Yes, I discussed with Rishabh about it. He suggested me to do this.

test/JDBC/upgrade/14_10/schedule
riyajain39: Resolved.

Deepesh125 (approved):

PR #1813, opened 2023-09-04
Added the fix for column and table aliases contain single and multibyte characters.

riyajain39:
### Description

This pull request fixed the implementation for **column** and **table aliases** contain both single byte and multibyte characters whose length is more than 64 bytes for 14.latest version.

**Issue 1 :** Handle column and table aliases.
Column alias does not work when alias length is more than or equals to 64 bytes in case of multibyte aliases. When query select 1  as '您对“数据一览“中的车型，颜色，内饰，选装, '; executes through T-SQL endpoint, it causes client to hang and throws protocol error in TDS stream. The logic of [pre_transform_target_entry](https://github.com/babelfish-for-postgresql/babelfish_extensions/blob/BABEL_3_X_DEV/contrib/babelfishpg_tsql/src/hooks.c#L1501C1-L1501C1) was not handling the aliases contain multibyte characters whose length is more than 64 bytes accurately and overwriting the alias name in case of truncation. To address this problem, I redefined the function and implemented the conditions which will handle aliases delimited with square bracket and double quotes separately with other aliases. In other aliases, I added the conditions which will handle both truncated and untruncated aliases separately.

**Issue 2 :** Aliases with multibyte chars cause MVU failure.
When establishing a view of alias with multibyte chars through T-SQL endpoint and retrieving it's definition from PG end point, extra unreadable chars are appearing at the end of the alias and even we can not use this view through tsql endpoint as it throws protocol error in TDS stream. But while we use similar alias on PG end point, we are able to retrieve it's correct definition. This discrepancy in behaviour is causing MVU failure. After implementing a fix, the correct alias definition is obtained from the PG endpoint, which rectifies the issue.

### Issues Resolved

BABEL-4231

Signed-off-by: Riya Jain <riyaajn@amazon.com>

### Check List

### DIFF
--- a/contrib/babelfishpg_tsql/src/hooks.c
+++ b/contrib/babelfishpg_tsql/src/hooks.c
@@ -1275,7 +1275,7 @@ find_attr_by_name_from_relation(Relation rd, const char *attname, bool sysColOK)
 
 static void
 pre_transform_target_entry(ResTarget *res, ParseState *pstate,
-						ParseExprKind exprKind)
+							ParseExprKind exprKind)
 {
 	if (prev_pre_transform_target_entry_hook)
 		(*prev_pre_transform_target_entry_hook) (res, pstate, exprKind);
@@ -1321,7 +1321,7 @@ pre_transform_target_entry(ResTarget *res, ParseState *pstate,
 		/*
 		 * Case 1 : Handle both singlebyte and multibyte aliases when delimited by 
 		 * square bracket(sqb) and double quoutes(dq).
-		 * For instance, queries like : SELECT 1 AS '您对“数据一览“中的车型，颜色，内饰，选装, ';
+		 * For instance, queries like : SELECT 1 AS "您对“数据一览“中的车型，颜色，内饰，选选装";
 		 * Case 2 : Preserve the case of aliases with ascii characters when there is no sqb and dq.
 		 * For instance, queries like: SELECT 1 AS ABCD;
 		 * Case 3 : Handle both singlebyte and multibyte aliases whose length is
--- a/contrib/babelfishpg_tsql/src/hooks.c
+++ b/contrib/babelfishpg_tsql/src/hooks.c
@@ -1328,7 +1328,7 @@ pre_transform_target_entry(ResTarget *res, ParseState *pstate,
 
             if (dq || sqb)
             {
-                memcpy(alias, identifier_name, alias_len);
+                return;
             }
             else
             {
@@ -1337,21 +1337,29 @@ pre_transform_target_entry(ResTarget *res, ParseState *pstate,
                     colname_start++;
                 }
                 /*
-                * After truncation, maximum truncated length of alias can be 60
-                * If length is less than 60, it means alias is not truncated
-                */
+                 * After truncation, maximum truncated length of alias can be 60
+                 * If length is less than 60, it means alias is not truncated
+                 */
                 if(alias_len < 60){
                     memcpy(alias, colname_start, alias_len);
                 }
                 else
                 {
                     /*
-                    * To check whether last 32 bytes are equal to identifier_name or not.
-                    * If they are not equal, this means identifier_name is truncated
-                    */
+                     * To handle identifier when length is between 60 to 63, 
+                     * it is needed to check whether last 32 bytes are equal to identifier_name or not.
+                     * If they are not equal, this means identifier_name is truncated.
+                     */
 
                     for(int x = alias_len - 32; x < alias_len; x++){
                         colname_end = colname_start + x;
+
+						/*
+                         * Check if colname_end is in upper case then does uppercase of identifier_name 
+                         * matches to colname_end or not in case of ascii values. 
+                         * If colname_end is in lowercase, then simply check colname_end is equals to 
+                         * identifier_name or not.
+                         */ 
                         if((*colname_end == identifier_name[x] || *colname_end == toupper(identifier_name[x])))
                         {
                             identifier_truncated = false;
@@ -1359,9 +1367,11 @@ pre_transform_target_entry(ResTarget *res, ParseState *pstate,
                         else
                         {
                             identifier_truncated = true;
-                            memcpy(alias, colname_start, (alias_len - 32) );
-                            memcpy(alias + (alias_len + 1 - 1) - 32,
-                            identifier_name + (alias_len + 1 - 1) - 32, 
+							/* First 32 characters of colname_start are assigned to alias */
+                            memcpy(alias, colname_start, (alias_len - 32));
+							/* Last 32 characters of identifier_name are assigned to alias, as actual alias is truncated */
+                            memcpy(alias + (alias_len) - 32,
+                            identifier_name + (alias_len) - 32, 
                             32);
                             alias[alias_len+1] = '\0';
                             break;
--- a/contrib/babelfishpg_tsql/src/hooks.c
+++ b/contrib/babelfishpg_tsql/src/hooks.c
@@ -1275,7 +1275,7 @@ find_attr_by_name_from_relation(Relation rd, const char *attname, bool sysColOK)
 
 static void
 pre_transform_target_entry(ResTarget *res, ParseState *pstate,
-						   ParseExprKind exprKind)
+						ParseExprKind exprKind)
 {
 	if (prev_pre_transform_target_entry_hook)
 		(*prev_pre_transform_target_entry_hook) (res, pstate, exprKind);
@@ -1301,7 +1301,7 @@ pre_transform_target_entry(ResTarget *res, ParseState *pstate,
 			/*
 			 * If no alias is specified on a ColumnRef, then get the length of
 			 * the name from the ColumnRef and copy the column name from the
-			 * sourcetext
+			 * sourcetext.
 			 */
 			if (list_length(cref->fields) == 1 &&
 				IsA(linitial(cref->fields), String))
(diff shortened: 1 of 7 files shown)


contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -1315,72 +1315,64 @@ pre_transform_target_entry(ResTarget *res, ParseState *pstate,
...
-			{
-				memcpy(alias + (NAMEDATALEN - 1) - 32,
-					   identifier_name + (NAMEDATALEN - 1) - 32,
-					   32);
-				alias[NAMEDATALEN] = '\0';
-			}
-
-			res->name = alias;
-		}
+        {
```
Deepesh125: Indentation seems odd

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -1315,72 +1315,64 @@ pre_transform_target_entry(ResTarget *res, ParseState *pstate,
...
+            bool        dq = *colname_start == '"';
+            bool        sqb = *colname_start == '[';
+            bool        sq = *colname_start == '\'';
+            bool        identifier_truncated;
+            const char *colname_end;
+
+            if (dq || sqb)
+            {
+                memcpy(alias, identifier_name, alias_len);
+            }
```
Deepesh125: There's scope of optimisation here. We first read identifier_name from res->name and then copy content in alias and re-assign res->name with alias. In summation, this is kind of no-op. So just return from here in this case.

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -1315,72 +1315,64 @@ pre_transform_target_entry(ResTarget *res, ParseState *pstate,
...
+            else
+            {
+                if(sq)
+                {
+                    colname_start++;
+                }
+                /*
+                * After truncation, maximum truncated length of alias can be 60
+                * If length is less than 60, it means alias is not truncated
+                */
```
Deepesh125: Fix indentation. Plus, we could elaborate more on this comment. We need to say that we need to handle identifier in special way only when truncated identifier length is b/w 61 to 63 to handle multibytes chars correctly.

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -1315,72 +1315,64 @@ pre_transform_target_entry(ResTarget *res, ParseState *pstate,
...
+                else
+                {
+                    /*
+                    * To check whether last 32 bytes are equal to identifier_name or not.
+                    * If they are not equal, this means identifier_name is truncated
+                    */
+
+                    for(int x = alias_len - 32; x < alias_len; x++){
+                        colname_end = colname_start + x;
+                        if((*colname_end == identifier_name[x] || *colname_end == toupper(identifier_name[x])))
```
Deepesh125: Check if colname_end (also original char) is upper case then does uppercase of identifier_name matches? otherwise just compare colname_end with identifier_name.

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -1315,72 +1315,64 @@ pre_transform_target_entry(ResTarget *res, ParseState *pstate,
...
+                        }
+                        else
+                        {
+                            identifier_truncated = true;
+                            memcpy(alias, colname_start, (alias_len - 32) );
+                            memcpy(alias + (alias_len + 1 - 1) - 32,
+                            identifier_name + (alias_len + 1 - 1) - 32, 
+                            32);
+                            alias[alias_len+1] = '\0';
+                            break;
```
Deepesh125: Add comment to explain this. Also, `+ 1 - 1` is not needed just explain it in comments.

Deepesh125 (changes requested):

contrib/babelfishpg_tsql/src/hooks.c
riyajain39: Addressed.

contrib/babelfishpg_tsql/src/hooks.c
riyajain39: Addressed.

contrib/babelfishpg_tsql/src/hooks.c
riyajain39: Addressed.

contrib/babelfishpg_tsql/src/hooks.c
riyajain39: Addressed.

contrib/babelfishpg_tsql/src/hooks.c
riyajain39: Addressed.

contrib/babelfishpg_tsql/src/hooks.c
Deepesh125: This is not resolved.

contrib/babelfishpg_tsql/src/hooks.c
Deepesh125: Yet to be resolved

contrib/babelfishpg_tsql/src/hooks.c
riyajain39: Resolved now.

contrib/babelfishpg_tsql/src/hooks.c
riyajain39: Now resolved.

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -1316,71 +1316,113 @@ pre_transform_target_entry(ResTarget *res, ParseState *pstate,
 			colname_start = pstate->p_sourcetext + res->name_location;
 		}
 
+		/*
+         * Case 1 : Handle both singlebyte and multibyte aliases when delimited by
```
Deepesh125: Indentation:

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -1316,71 +1316,113 @@ pre_transform_target_entry(ResTarget *res, ParseState *pstate,
...
-			bool		sq = *colname_start == '\'';
-			int			a = 0;
-			const char *colname_end;
-			bool		closing_quote_reached = false;
-
-			if (dq || sqb || sq)
-			{
-				colname_start++;
-			}
+        {
```

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -1316,71 +1316,113 @@ pre_transform_target_entry(ResTarget *res, ParseState *pstate,
...
-					{
-						if ((*(++colname_end) != '"'))
-						{
-							closing_quote_reached = true;
-							break;	/* end of dbl-quoted identifier */
-						}
-					}
-					else if (sq && *colname_end == '\'')
+						/*
+                         * Check if colname_end is in upper case then does uppercase of identifier_name
```

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -1316,71 +1316,113 @@ pre_transform_target_entry(ResTarget *res, ParseState *pstate,
...
+            {
+                if(sq)
+                {
+                    colname_start++;
+                }
+                /*
+                 * After truncation, maximum truncated length of alias can be 60
+                 * If length is less than 60, it means alias is not truncated
+                 */
+                if(alias_len < 60){
```
Deepesh125: Open bracket should be on next line

riyajain39: Resolved

contrib/babelfishpg_tsql/src/hooks.c
riyajain39: Resolved.

contrib/babelfishpg_tsql/src/hooks.c
riyajain39: Resolved.

contrib/babelfishpg_tsql/src/hooks.c
riyajain39: Resolved.

Deepesh125 (dismissed):

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -1273,7 +1273,7 @@ find_attr_by_name_from_relation(Relation rd, const char *attname, bool sysColOK)
 
 static void
 pre_transform_target_entry(ResTarget *res, ParseState *pstate,
-						   ParseExprKind exprKind)
+						ParseExprKind exprKind)
```
kuntalghosh: Unnecessary change

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -1316,67 +1316,114 @@ pre_transform_target_entry(ResTarget *res, ParseState *pstate,
...
+		 * more than or equals to 63 when not delimited by sqb and dq.
+		 * For instance, queries like : SELECT 1 AS 您对您对您对您对您对您对您对您对您对您对您对您对您对;
+		 */
 		if (alias_len > 0)
 		{
-			char	   *alias = palloc0(alias_len + 1);
-			bool		dq = *colname_start == '"';
-			bool		sqb = *colname_start == '[';
-			bool		sq = *colname_start == '\'';
-			int			a = 0;
```
kuntalghosh: Unnecessary change

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -1316,67 +1316,114 @@ pre_transform_target_entry(ResTarget *res, ParseState *pstate,
 			colname_start = pstate->p_sourcetext + res->name_location;
 		}
 
+		/*
+		 * Case 1 : Handle both singlebyte and multibyte aliases when delimited by 
+		 * square bracket(sqb) and double quoutes(dq).
+		 * For instance, queries like : SELECT 1 AS '您对“数据一览“中的车型，颜色，内饰，选装, ';
```
kuntalghosh: The example is neither using square bracket nor using double quotes

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -1316,67 +1316,114 @@ pre_transform_target_entry(ResTarget *res, ParseState *pstate,
...
+		 * Case 2 : Preserve the case of aliases with ascii characters when there is no sqb and dq.
+		 * For instance, queries like: SELECT 1 AS ABCD;
+		 * Case 3 : Handle both singlebyte and multibyte aliases whose length is
+		 * more than or equals to 63 when not delimited by sqb and dq.
```
kuntalghosh, line 1328: Why don't we have to worry about singlebyte and multibyte aliases whose length is more than or equals to 63 when not delimited by sqb and dq?

contrib/babelfishpg_tsql/src/hooks.c
riyajain39: Resolved.

contrib/babelfishpg_tsql/src/hooks.c
riyajain39: Resolved.

contrib/babelfishpg_tsql/src/hooks.c
riyajain39: Aliases when delimited with sqb and dq are already handled in grammer rule. These type of aliases are truncated properly and also the case is preserved. In such case, we are just returning. So now we need to handle aliases which are not delimited with sqb and dq whose length is less than 63 and more than 63. In case of single byte aliases, we need to preserve the case while in the case of multibyte aliases if actual alias length is more than 63, we should retrieve the alias name after truncation.

contrib/babelfishpg_tsql/src/hooks.c
riyajain39: Did these changes to fix the indentation.

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -1318,67 +1318,114 @@ pre_transform_target_entry(ResTarget *res, ParseState *pstate,
...
+		 * For instance, queries like : SELECT 1 AS 您对您对您对您对您对您对您对您对您对您对您对您对您对;
+		 */
 		if (alias_len > 0)
 		{
 			char	   *alias = palloc0(alias_len + 1);
 			bool		dq = *colname_start == '"';
 			bool		sqb = *colname_start == '[';
 			bool		sq = *colname_start == '\'';
-			int			a = 0;
+			bool		identifier_truncated;
```
Deepesh125: Dont leave this uninitialised. Initialise it with false

riyajain39: Resolved.

Deepesh125 (approved):

kuntalghosh (approved):

PR #1832, opened 2023-09-11
Add collation when initialising scan keys for sys.varchar

tanscorpio7:
### Description
Orphan entries in tables uisng atleast one sys.varchar as a column.
Errors like : "duplcate key error" in view definition catalog table (sys.bablefish_view_def) and 
domain mapping entry corresponding to supplied argument could not be found" for domain 
mapping catalog (sys.babelfish_domain_mapping).

When we init scan keys without a collation, default collation "C" is used and hence comparison 
take place as per "C" collation whereas comparison as per server collation is expected. And 
hence index scan does not return the correct row.

We will now fetch the collation value from using tsql_get_server_collation_oid_internal(false) 
and forward it to the scan key init function (ScanKeyEntryInitialize()).

tables fixed - sys.bablefish_view_def, sys.babelfish_domain_mapping

Note - We will fix similar issue for other system catalog in subsequent changes.

Test Scenarios Covered -
We use cases which fail currently, that is using view/domain names with underscore character to force a situation which would fail if wrong collation is picked up when searching the index.

Task: BABEL-4389

### Check List

### DIFF
--- a/contrib/babelfishpg_tsql/src/catalog.c
+++ b/contrib/babelfishpg_tsql/src/catalog.c
@@ -1101,7 +1101,6 @@ HeapTuple
 search_bbf_view_def(Relation bbf_view_def_rel, int16 dbid, const char *logical_schema_name, const char *view_name)
 {
 
-	Relation 	bbf_view_def_idx_rel;
 	ScanKeyData scanKey[3];
 	SysScanDesc scan;
 	HeapTuple	scantup,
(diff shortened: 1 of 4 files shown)


contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -1109,30 +1110,34 @@ search_bbf_view_def(Relation bbf_view_def_rel, int16 dbid, const char *logical_s
...
 
+	bbf_view_def_idx_rel = index_open(get_bbf_view_def_idx_oid(), AccessShareLock);
 
 	/* Search and drop the definition */
-	ScanKeyInit(&scanKey[0],
-				Anum_bbf_view_def_dbid,
-				BTEqualStrategyNumber, F_INT2EQ,
+	ScanKeyEntryInitialize(&scanKey[0], 0, Anum_bbf_view_def_dbid,
+				BTEqualStrategyNumber, InvalidOid,
+				bbf_view_def_idx_rel->rd_indcollation[0], F_INT2EQ,
```
Deepesh125: datatype of this field is SMALLINT so we do nt need to specify the collation here.

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -1109,30 +1110,34 @@ search_bbf_view_def(Relation bbf_view_def_rel, int16 dbid, const char *logical_s
...
 				Int16GetDatum(dbid));
 
-	ScanKeyInit(&scanKey[1],
-				Anum_bbf_view_def_schema_name,
-				BTEqualStrategyNumber, F_TEXTEQ,
+	ScanKeyEntryInitialize(&scanKey[1], 0, Anum_bbf_view_def_schema_name,
+				BTEqualStrategyNumber, InvalidOid,
+				bbf_view_def_idx_rel->rd_indcollation[1], F_TEXTEQ,
```
Deepesh125: data type of this field is sys.sysname whose collation would always be same as `tsql_get_server_collation_oid_internal(false)` so we could avoid opening table index to gather the collation.

Deepesh125: Applicable to all other places

Deepesh125 (changes requested):

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -1116,14 +1116,14 @@ search_bbf_view_def(Relation bbf_view_def_rel, int16 dbid, const char *logical_s
 				BTEqualStrategyNumber, F_INT2EQ,
 				Int16GetDatum(dbid));
 
-	ScanKeyInit(&scanKey[1],
```
surendravishnoi, line 1119: have we validated that no other babel catalogs have this issue?

Deepesh125 (approved):

contrib/babelfishpg_tsql/src/catalog.c
Deepesh125: @tanscorpio7 Is preparing document where we need similar kind of fix.

PR #1841, opened 2023-09-13
Do not dump Babelfish initialize user

rishabhtanwar29:
### Description
When Babelfish initialize user is same between source and target servers
for dump/restore, the restore throws errors like "role ... already exists" and
"duplicate key value violates unique constraint..." etc while inserting catalog
entry for initialize user in babelfish_authid_login_ext catalog. These errors are
actually expected and makes sense as we know the user do already exists on
the target server but the whole restore gets rolled back in case the restore
process has been made transactional (`--single-transaction` option is used)
and there is no way for a user to fix this manually.

To overcome the above issue, the engine commit implements the logic to skip
dumping Babelfish initialize user as well as the "owner" column of
`sys.babelfish_sysdatabases` catalog table.

This commit implements logic to re-populate "owner" column of
`sys.babelfish_sysdatabases` catalog table during restore with the owner of
current database (which is going to be initialize user of target Babelfish database).

#### Test changes:
* Updated github dump/restore action to take two more options as input:
  1. dump-data-as: Dump table data using COPY or INSERT
  2. dump-format: Dump format (plain/custom/tar/directory)
* Updated dump-restore configurations to test all four dump formats and
  both data formats.
* Enabled `--single-transaction` option for restore so that any error during restore
  would result in restore getting rolledback and failed.
* Added TAP test coverage for pg_restore.
* Fixed BABEL-1206-vu-prepare test to insert valid data into varchar column. 

Task: BABEL-4404 

Engine PR: https://github.com/babelfish-for-postgresql/postgresql_modified_for_babelfish/pull/216

* **Use case based -**

* **Boundary conditions -**

* **Arbitrary inputs -**

* **Negative test cases -**

* **Performance tests -**

* **Tooling impact -**

* **Client tests -**

### Check List


.github/composite-actions/dump-restore-util/action.yml
```diff
@@ -63,35 +78,16 @@ runs:
...
-    - name: Temporarily disable failing tests
-      if: always() && steps.run-pg_dump-restore.outcome == 'success' && inputs.is_final_ver == 'true'
-      env:
-        migr_mode: ${{ inputs.migration_mode }}
-      run: |
-        if [[ "$migr_mode" == "multi-db" ]];then
-          base_dir=${{ matrix.upgrade-path.path[0] }}
-          if [[ "$base_dir" == *"latest"* ]]; then
-            base_dir="latest"
-          fi
```
HarshLunagariya, line 84: We don't need this because we modified testcase to use master, right?

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -3827,18 +3827,37 @@ fill_missing_values_in_copyfrom(Relation rel, Datum *values, bool *nulls)
...
+
+		attnum = (AttrNumber) attnameAttNum(rel, "owner", false);
+		Assert(attnum != InvalidAttrNumber);
 
-		dbid = getDbidForLogicalDbRestore(relid);
-		values[attnum - 1] = Int16GetDatum(dbid);
-		nulls[attnum - 1] = false;
+		if (nulls[attnum - 1])
+		{
+			const char *owner = GetUserNameFromId(GetSessionUserId(), false);
```
HarshLunagariya: Will it be correct always? Or should we fetch current physical database's owner?

test/JDBC/input/pg_hint_plan/BABEL-3513-vu-verify.mix
```diff
@@ -6,6 +6,9 @@ analyze master_dbo.babel_3513_t2;
 go
 
 -- tsql
+exec sp_babelfish_configure 'enable_pg_hint', 'on', 'server'
+go
```
HarshLunagariya, line 10: Why this change?

.github/composite-actions/dump-restore-util/action.yml
rishabhtanwar29: Yeah so, this step was initially added to temporarily skip the tests which were failing as part of dump/restore as we were still working on fixing them, but now this step is unnecessary since there are no test failures remaining and we do want to fail the workflow if there are any test failures.

test/JDBC/input/pg_hint_plan/BABEL-3513-vu-verify.mix
rishabhtanwar29: Oh this is needed because we don't preserve the database settings across dump/restore, so we basically need to reset the value of `enable_pg_hint`.

contrib/babelfishpg_tsql/src/hooks.c
rishabhtanwar29: Although both ways should be fine as restore must always be performed with babelfish initialize user but since we skipped dumping physical database's owner then it makes more sense to populate it here with physical database's owner here again.

test/JDBC/input/pg_hint_plan/BABEL-3513-vu-verify.mix
HarshLunagariya: Got it. But then won't we be missing coverage where it should be preserved  in case of M/mVU.

HarshLunagariya (approved):

shardgupta (approved):

PR #1843, opened 2023-09-15
Modify babel_cursor test to match SQL Server usage

xhfanhe:
### Description

Babelfish cursor has the following problems:
1. In SQL Server, cursor can only be a procedure argument of type VARYING OUTPUT, and cannot be an INPUT. However both are allowed in Babelfish.
2. In current Babelfish, when cursor is used as a VARYING OUTPUT argument, it is not behaving the same as SQL Server. This issue is tracked in BABEL-3350.

Due to the above reasons, rewrote/removed some test cases in babel_cursor.sql test to match SQL Server usage. Output has been verified against SQL Server test result. For the same reason, `sp_cursor_list` and `sp_describe_cursor` usage does not match SQL Server behavior either, removing related tests till BABEL-3350 is fixed.

Task: BABEL-4391
Signed-off-by: Xiaohui Fanhe <hexiaohu@amazon.com>

### Check List

### DIFF
--- a/test/JDBC/expected/babel_cursor.out
+++ b/test/JDBC/expected/babel_cursor.out
@@ -17,14 +17,16 @@ GO
 
 
 
-CREATE PROCEDURE babel_fetch_cursor_helper_int_proc(@cur CURSOR, @num_fetch int)
+CREATE PROCEDURE babel_cursor_proc(@num_fetch int)
 AS
 BEGIN
 	DECLARE @var_i int;
 	DECLARE @cnt int = 0;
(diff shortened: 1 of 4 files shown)


2jungkook (approved):

forestkeeper (approved):

robverschoor (commented):
I think it should be possible to create a test case where a cursor variable is used as a parameter for a procedure call.

xhfanhe:
> I think it should be possible to create a test case where a cursor variable is used as a parameter for a procedure call.

(Somehow I cannot directly reply to the comment) Yes and it's been tested in some other test cases like BABEL-2262.sql and routines_definition.sql.

PR #1845, opened 2023-09-15
Adding Unit Tests For SSL Operations

vikashprajapati-cell:
### Description
SSL (Secure Sockets Layer) is a cryptographic protocol used to establish secure connections between clients and servers. Testing SSL functionality ensures the secure transmission of data between clients and servers. Adding tests for SSL involves evaluating how the Babelfish codebase handles SSL connections.
We are keen on conducting thorough unit testing of the SslHandshake process that occurs during the prelogin stage. Our unit testing approach involves meticulously examining each step of the handshake, simulating various scenarios, and verifying the correct implementation of the SSL handshake. Through this comprehensive testing, we aim to identify any potential vulnerabilities that may exist.

* **Note-**
To execute SSL-related tests, it is essential to enable the SSL flag during the build process using the --with-openssl command. This ensures that the necessary components for SSL testing are included. In the event that the SSL flag is not enabled, only non-SSL tests will be executed.

Authored-by: Vikash Prajapati [vikasprj@amazon.com](mailto:vikasprj@amazon.com)
Signed-off-by: Vikash Prajapati [vikasprj@amazon.com](mailto:vikasprj@amazon.com)

```
jdbc_testdb=# CREATE EXTENSION IF NOT EXISTS babelfishpg_unit;
CREATE EXTENSION
jdbc_testdb=# select * from babelfishpg_unit.babelfishpg_unit_run_tests();
                  test_name                  | status |                   message                   | runtime | enabled 
---------------------------------------------+--------+---------------------------------------------+---------+---------
 GreaterThanOrEqualToCheck_INT4_FIXEDDECIMAL | pass   |                                             |       6 | enabled
 LesserThanOrEqualToCheck_INT4_FIXEDDECIMAL  | pass   |                                             |       1 | enabled
 NotEqualToCheck_INT4_FIXEDDECIMAL           | pass   |                                             |       1 | enabled
 EqualToCheck_INT4_FIXEDDECIMAL              | pass   |                                             |       2 | enabled
 LesserThanCheck_INT4_FIXEDDECIMAL           | pass   |                                             |       1 | enabled
 Comparison_INT4_FIXEDDECIMAL                | pass   |                                             |       3 | enabled
 FIXEDDECIMALUM                              | pass   | fixeddecimal out of range                   |      11 | enabled
 GreaterThanOrEqualToCheck_FIXEDDECIMAL_INT2 | pass   |                                             |       1 | enabled
 LesserThanOrEqualToCheck_FIXEDDECIMAL_INT2  | pass   |                                             |       1 | enabled
 GreaterThanCheck_FIXEDDECIMAL_INT2          | pass   |                                             |       1 | enabled
 NotEqualToCheck_FIXEDDECIMAL_INT2           | pass   |                                             |       1 | enabled
 Comparison_FIXEDDECIMAL_INT2                | pass   |                                             |       2 | enabled
 Testing_SSL_HANDSHAKE_READ                  | pass   |                                             |      21 | enabled
 TestingOverSize_SSL_HANDSHAKE_READ          | pass   |  SSL packet expand more than one TDS packet |      52 | enabled
 TestingPacketType_SSL_HANDSHAKE_READ        | pass   |                                             |       6 | enabled
 Testing_SSL_HANDSHAKE_WRITE                 | pass   |                                             |       4 | enabled
 TestingSizeCheck_SSL_HANDSHAKE_WRITE        | pass   |  There is nothing to write                  |       1 | enabled
(17 rows)

```

### Check List


contrib/babelfishpg_unit/test_ssl_read.c
```diff
@@ -0,0 +1,179 @@
...
+    PG_TRY();
+    {
+        test_ssl_handshake_read(h, buf, 0, mock_socket_read, ReadPointer);
+    }
+    PG_CATCH();
+    {
+        MemoryContextSwitchTo(oldcontext);
+        errorData = CopyErrorData();
+        FlushErrorState();
+        snprintf(testResult->message, MAX_TEST_MESSAGE_LENGTH, "%s, %s", testResult->message, errorData->message);
```
ahmed-shameem: I observed in description that in the message column there is an extra ',' at the beginning of the message. This is because %s is used 2 times. Please fix this.

contrib/babelfishpg_tds/src/backend/tds/tdssecure.c
```diff
@@ -40,6 +40,10 @@
 
 int			tds_ssl_min_protocol_version;
 int			tds_ssl_max_protocol_version;
+
+static TdsSecureSocketApi tds_secure_raw_read = secure_raw_read;
+static TdsSecureSocketApiConst tds_secure_raw_write = secure_raw_write;
```
ahmed-shameem, line 45: Is there any specific reason regarding why are we using TdsSecureSocketApi for read and TdsSecureSocketApiConst for write? Can't we use same APIs for both? If there is some specific reason using different APIs , please add proper comment.

contrib/babelfishpg_tds/src/backend/tds/tdssecure.c
```diff
@@ -163,6 +169,29 @@ SslHandShakeRead(BIO * h, char *buf, int size)
 	return res;
 }
 
+
+ssize_t
+test_ssl_handshake_read(BIO * h, char *buf, int size, TdsSecureSocketApi mock_socket_read, int ReadPointer)
```
ahmed-shameem, line 174: I see we have introduced 2 different functions to call static functions in this files. If we can use same APIs for read and write i.e., TdsSecureSocketApi or TdsSecureSocketApiConst, then let us narrow down to using one function only.

contrib/babelfishpg_unit/test_ssl_write.c
```diff
@@ -0,0 +1,108 @@
...
+}
+
+
+TestResult*
+test_ssl_handshakeWrite(void)
+{
+
+    /*
+     *  We will generate a message and pass it to the handshake write function
+     *  We will be comparing the number of bytes written against the expected value
```
ahmed-shameem: Please update the comment saying we are also checking the actual bytes which are written

contrib/babelfishpg_unit/test_ssl_write.c
```diff
@@ -0,0 +1,108 @@
...
+    strncpy(expected_str, buf, expected);
+    expected_str[expected] = '\0';
+
+    obtained = test_ssl_handshake_write(h, buf, expected, mock_socket_write);
+
+    obtained_str = (char *)malloc(obtained + 1);
+    strncpy(obtained_str, buf, obtained);
+    obtained_str[obtained] = '\0';
+
+    TEST_ASSERT_TESTCASE(expected == expected, testResult);
```
ahmed-shameem: nit: please add comment saying which ASSERT is checking what? I mean add comment saying this ASSERT checks the number of bytes, this ASSERT checks the actual byte message. Also, as we are checking the number of bytes, I think we should check the actual message only iff the number of bytes matches or else there is no need to write. So, better to put and if condition maybe.

contrib/babelfishpg_unit/test_ssl_read.c
```diff
@@ -0,0 +1,179 @@
...
+}
+
+
+TestResult*
+test_ssl_handshakeRead(void)
+{
+
+    /*
+     *  We will generate a prelogin request message and pass it to the handshake read function for processing
+     *  By comparing the number of bytes read against the expected value, we can verify the accuracy of the read operation
```
ahmed-shameem: Please update the comment saying we are also checking the actual bytes which are read

contrib/babelfishpg_unit/test_ssl_read.c
```diff
@@ -0,0 +1,179 @@
...
+    
+    obtained = test_ssl_handshake_read(h, buf, expected, mock_socket_read, ReadPointer);
+
+    size_buf = strlen(buf);
+    size_binaryData = strlen(expected_str);
+    buf[size_buf+1] = '\0';
+    expected_str[size_binaryData+1] = '\0';
+    
+    TEST_ASSERT_TESTCASE(obtained == expected, testResult);
+    TEST_ASSERT_TESTCASE(*((unsigned char*)expected_str) == *((unsigned char*)buf), testResult);
```
ahmed-shameem: nit: please add comment saying which ASSERT is checking what? I mean add comment saying this ASSERT checks the number of bytes, this ASSERT checks the actual byte message. Also, as we are checking the number of bytes, I think we should check the actual message only iff the number of bytes matches or else there is no need to read. So, better to put and if condition maybe.

ahmed-shameem (changes requested):

contrib/babelfishpg_tds/src/backend/tds/tdssecure.c
vikashprajapati-cell: Yes, the second argument of `secure_raw_write` is a const pointer and hence we are using `TdsSecureSocketApiConst`. While this is not the case for `secure_raw_read`.

vikashprajapati-cell: Yes, the second argument in the function `secure_raw_write` is a constant pointer. and hence the API `TdsSecureSocketApiConst` is used. While this is not the case for `secure_raw_read` function.

contrib/babelfishpg_tds/src/backend/tds/tdssecure.c
vikashprajapati-cell: The second argument in the function secure_raw_write is a constant pointer. and hence the API `TdsSecureSocketApiConst` is used. While this is not the case for `secure_raw_read` function.
So we cant do this.

contrib/babelfishpg_unit/test_ssl_write.c
vikashprajapati-cell: Updated in the next revision.

contrib/babelfishpg_unit/test_ssl_write.c
vikashprajapati-cell: done in the next revision.

contrib/babelfishpg_unit/test_ssl_read.c
vikashprajapati-cell: Updated in the next revision.

contrib/babelfishpg_unit/test_ssl_read.c
vikashprajapati-cell: Done in the next revision.

contrib/babelfishpg_unit/test_ssl_read.c
vikashprajapati-cell: Done.

ahmed-shameem (approved):

thephantomthief (commented):
We should be running these new tests as part of github actions. Currently I think that is not happening

contrib/babelfishpg_unit/test_ssl_read.c
```diff
@@ -0,0 +1,184 @@
...
+    buf = malloc(strlen(prelogin_request)/2);
+    expected = strlen(prelogin_request)/2 - 8;
+
+    expected_str = (char *)malloc(strlen(prelogin_request)/2);
+    for (int i = 0; i < expected; i++) 
+        sscanf(&prelogin_request [i * 2], "%2hhx", (char *)expected_str + i);
+    
+    obtained = test_ssl_handshake_read(h, buf, expected, mock_socket_read, ReadPointer);
+
+    size_buf = strlen(buf);
```
surendravishnoi: strlen only works in string is null terminated

contrib/babelfishpg_unit/test_ssl_read.c
```diff
@@ -0,0 +1,184 @@
...
+    size_binaryData = strlen(expected_str);
+    buf[size_buf+1] = '\0';
+    expected_str[size_binaryData+1] = '\0';
+
+    /*
+     * if number of bytes are same then we will compare actual data
+     */
+    if(expected == obtained)
+    {
+        TEST_ASSERT_TESTCASE(*((unsigned char*)expected_str) == *((unsigned char*)buf), testResult);
```
surendravishnoi: why are we only comparing first character?

contrib/babelfishpg_unit/test_ssl_read.c
```diff
@@ -0,0 +1,184 @@
...
+    h = BIO_new(BIO_s_mem());
+    ReadPointer = 0;
+
+    expected = strlen(prelogin_request)/2 - 8;
+    obtained = test_ssl_handshake_read(h, buf, expected, mock_socket_read, ReadPointer);
+
+    snprintf(expected_str, MAX_TEST_MESSAGE_LENGTH, "%d", expected);
+    snprintf(obtained_str, MAX_TEST_MESSAGE_LENGTH, "%d", obtained);
+    
+    TEST_ASSERT_TESTCASE(expected != obtained, testResult);
```
surendravishnoi, line 102: this is not clear, what do we expect here that multi packet message should be handled properly, if yes why results should not match

contrib/babelfishpg_tds/src/backend/tds/tdssecure.c
```diff
@@ -245,6 +274,28 @@ SslHandShakeWrite(BIO * h, const char *buf, int size)
...
+test_ssl_handshake_write(BIO * h, char *buf, int size, TdsSecureSocketApiConst mock_socket_write)
+{
+
+    /*
+     *  Accessing SslHandShakeWrite() function directly from our testing functions created in the test_ssl_write.c file is not possible due to its static nature
+     *  This intermediary function serves as a gateway, allowing us to invoke the SslHandShakeWrite() function from our testing function
+     */
+
+    int res;
+    tds_secure_raw_write = mock_socket_write;
```
surendravishnoi: should we reset it after the unit testing?

contrib/babelfishpg_tds/src/backend/tds/tdssecure.c
```diff
@@ -245,6 +274,28 @@ SslHandShakeWrite(BIO * h, const char *buf, int size)
...
+    unit_testing = true;
```
surendravishnoi: what if there is an error during testing and unit_testing remains true? I think not an issue in current set of tests but could be a problem later

surendravishnoi (dismissed):

contrib/babelfishpg_tds/src/backend/tds/tdssecure.c
vikashprajapati-cell: Yes, have updated in the next revision.

contrib/babelfishpg_tds/src/backend/tds/tdssecure.c
vikashprajapati-cell: Handled this using PG_TRY() and PG_FINALLY() block in the latest commit.

contrib/babelfishpg_unit/test_ssl_read.c
vikashprajapati-cell: Updated in the latest commit.

contrib/babelfishpg_unit/test_ssl_read.c
vikashprajapati-cell: Updated in the latest commit with complete string comparision.

contrib/babelfishpg_unit/test_ssl_read.c
vikashprajapati-cell: Updated the logic in the latest commit.

contrib/babelfishpg_tds/src/backend/tds/tdssecure.c
```diff
@@ -126,10 +132,10 @@ SslHandShakeRead(BIO * h, char *buf, int size)
 		}
 
 		if (unlikely(pkt_data.header.pkt_type != TDS_PRELOGIN))
-			ereport(FATAL,
+			ereport(unit_testing ? ERROR : FATAL,
 					(errcode(ERRCODE_ADMIN_SHUTDOWN),
 					 errmsg("terminating connection due to unexpected ssl packet header")));
-
+
```
surendravishnoi: unnecessary white spaces

contrib/babelfishpg_tds/src/backend/tds/tdssecure.c
```diff
@@ -163,6 +169,35 @@ SslHandShakeRead(BIO * h, char *buf, int size)
...
+    /*
+     *  Accessing SslHandShakeRead() function directly from our testing functions created in the test_ssl_read.c file is not possible due to its static nature
+     *  This intermediary function serves as a gateway, allowing us to invoke the SslHandShakeRead() function from our testing function
+     */
+
+    int res;
+    tds_secure_raw_read = mock_socket_read;
+    unit_testing = true;
+    pkt_bytes_read = ReadPointer;
+	PG_TRY();
```
surendravishnoi, line 186: indentation

contrib/babelfishpg_tds/src/backend/tds/tdssecure.c
```diff
@@ -163,6 +169,35 @@ SslHandShakeRead(BIO * h, char *buf, int size)
...
+    {
+        res = SslHandShakeRead(h, buf, size);
+    }
+    PG_FINALLY();
+    {
+        unit_testing = false;
+    }
+    PG_END_TRY();
+
+	tds_secure_raw_read = secure_raw_read;
```
surendravishnoi: put this also in finally block

contrib/babelfishpg_tds/src/backend/tds/tdssecure.c
```diff
@@ -245,6 +280,35 @@ SslHandShakeWrite(BIO * h, const char *buf, int size)
...
+    /*
+     *  Accessing SslHandShakeWrite() function directly from our testing functions created in the test_ssl_write.c file is not possible due to its static nature
+     *  This intermediary function serves as a gateway, allowing us to invoke the SslHandShakeWrite() function from our testing function
+     */
+
+    int res;
+    tds_secure_raw_write = mock_socket_write;
+    unit_testing = true;
+
+	PG_TRY();
```
surendravishnoi, line 296: indentation, also put write API reset in finally block

contrib/babelfishpg_unit/test_ssl_read.c
```diff
@@ -0,0 +1,198 @@
...
+    buf = malloc(strlen(prelogin_request)/2);
+    expected = strlen(prelogin_request)/2 - 8;
+
+    expected_str = (char *)malloc(strlen(prelogin_request)/2);
+    for (int i = 0; i < expected; i++) 
+        sscanf(&prelogin_request [i * 2], "%2hhx", (char *)expected_str + i);
+    
+    obtained = test_ssl_handshake_read(h, buf, expected, mock_socket_read, ReadPointer);
+
+    null_terminated_expected = strdup(expected_str);
```
surendravishnoi: why these copies? we can compare expected_str and buf directly. However, is buf guaranteed to be null terminated?

contrib/babelfishpg_unit/test_ssl_read.c
```diff
@@ -0,0 +1,198 @@
...
+
+    expected_str = (char *)malloc(strlen(prelogin_request)/2);
+    for (int i = 0; i < expected; i++) 
+        sscanf(&prelogin_request [i * 2], "%2hhx", (char *)expected_str + i);
+
+    obtained_str = (char *)malloc(expected+1);
+    memcpy(obtained_str, buf, obtained);
+    obtained_next = test_ssl_handshake_read(h, buf, expected - obtained, mock_socket_read, ReadPointer);
+    memcpy(obtained_str + obtained, buf, obtained_next);
+    null_terminated_expected = strdup(expected_str);
```
surendravishnoi: again why this copy? also if obtained_str is not null terminated how will strdup will work. You just need to put at '\0' at the end

contrib/babelfishpg_unit/test_ssl_read.c
```diff
@@ -0,0 +1,198 @@
...
+    int obtained_next;
+    char *expected_str;
+    char *obtained_str;
+    char *null_terminated_expected;
+    char *null_terminated_obtained;
+
+    TestResult* testResult = palloc0(sizeof(TestResult));
+    testResult->result = true;
+
+    prelogin_request = strdup("1201000F0000010011A25E4571");
```
surendravishnoi: can you explain the message format? how does it force two packet reads?

contrib/babelfishpg_unit/test_ssl_write.c
```diff
@@ -0,0 +1,116 @@
...
+
+    int expected;
+    int obtained;
+
+    TestResult* testResult = palloc0(sizeof(TestResult));
+    testResult->result = true;
+
+    h = BIO_new(BIO_s_mem());
+
+    prelogin = malloc(strlen(buf)/2 + 8);
```
surendravishnoi: why are we writing buf/2 rather than buf?

contrib/babelfishpg_unit/test_ssl_write.c
```diff
@@ -0,0 +1,116 @@
...
+    h = BIO_new(BIO_s_mem());
+
+    expected = -1;
+    obtained = test_ssl_handshake_write(h, buf, expected, mock_socket_write);
+
+    snprintf(expected_str, MAX_TEST_MESSAGE_LENGTH, "%d", expected);
+    snprintf(obtained_str, MAX_TEST_MESSAGE_LENGTH, "%d", obtained);
+
+    TEST_ASSERT_TESTCASE(expected == obtained, testResult);
+    if(testResult->result == true)
```
surendravishnoi, line 98: why both if/else cases?

contrib/babelfishpg_unit/test_ssl_write.c
```diff
@@ -0,0 +1,116 @@
...
+
+    TestResult* testResult = palloc0(sizeof(TestResult));
+    testResult->result = true;
+
+    h = BIO_new(BIO_s_mem());
+
+    expected = -1;
+    obtained = test_ssl_handshake_write(h, buf, expected, mock_socket_write);
+
+    snprintf(expected_str, MAX_TEST_MESSAGE_LENGTH, "%d", expected);
```
surendravishnoi: these are not used anywhere?

contrib/babelfishpg_tds/src/backend/tds/tdssecure.c
```diff
@@ -163,6 +169,35 @@ SslHandShakeRead(BIO * h, char *buf, int size)
...
+		res = SslHandShakeRead(h, buf, size);
+	}
+	PG_FINALLY();
+	{
+		unit_testing = false;
+		tds_secure_raw_read = secure_raw_read;
+	}
+	PG_END_TRY();
+
+		return res;
```
surendravishnoi, line 197: indentation

contrib/babelfishpg_unit/test_ssl_read.c
```diff
@@ -0,0 +1,173 @@
...
+    BIO *h = NULL;
+    char *buf = NULL;
+
+    int expected;
+    int obtained;
+
+    TestResult* testResult = palloc0(sizeof(TestResult));
+    testResult->result = true;
+
+    prelogin_request = strdup("1201000B00000100115461A23E");
```
surendravishnoi, line 93: again, why are we manipulating length here?

contrib/babelfishpg_unit/test_ssl_write.c
```diff
@@ -0,0 +1,106 @@
...
+    expected = strlen(buf);
+    expected_str = (char *)malloc(expected + 1);
+    strncpy(expected_str, buf, expected);
+    expected_str[expected] = '\0';
+
+    obtained = test_ssl_handshake_write(h, buf, expected, mock_socket_write);
+
+    obtained_str = (char *)malloc(obtained + 1);
+    strncpy(obtained_str, buf, obtained);
+    obtained_str[obtained] = '\0';
```
surendravishnoi, line 56: what are we achieving with this comparison? both expected and obtained strings are copy of buf only. We should compared expected against prelogin  string where mock socket write happens (after skipping header part)

PR #1847, opened 2023-09-18
BABEL-4411 set transaction isolation cmd change

tanscorpio7:
### Description
Currently babelfish transactions start with session default isolation level. Users can not change this isolation level after transaction has started. On running the set transaction isolation inside a transaction, current transaction isolation does not change but all subsequent transaction will pick up this new isolation level.

Changes in this PR:

Allow changing the current transaction isolation level if no DML query has been run inside transaction.
Abort the transaction if set transaction isolation level is run after a DML query in a transaction block unless the new isolation level is same as current isolation level.
Introduce an escape hatch to ignore any invalid set transaction isolation level stmt so that transaction does not abort.

#### Engine Change:
Add check hook to guc default_transaction_isolation (all errors should be taken care here, that is assign hook should never fail)
If the conditions for setting isolation level are true returns true
If conditions are false and escape hatch is not set (strict - default value) - abort transaction with error
but if escape hatch is set to ignore for same situation - we simply ignore the set tran isolation command by setting the new isolation level to current isolation level. Which is apparently the same as ignoring the cmd.

#### Extension Change:
Introduce escape hatch for set transaction isolation 
New attribute in PLtsql_stmt_execsql structure. This attribute is needed to identify if the stmt is set transaction isolation during initialisation of internal savepoint and if set, we skip the internal save point for the current stmt.
This needs to be done since postgres does not allow setting isolation level inside a save point.

### Issues Resolved

BABEL-4411

### Check List

### DIFF
--- a/contrib/babelfishpg_tsql/src/iterative_exec.c
+++ b/contrib/babelfishpg_tsql/src/iterative_exec.c
@@ -980,6 +980,19 @@ is_batch_command(PLtsql_stmt *stmt)
 }
 
 static
+bool
+is_set_tran_isolation(PLtsql_stmt *stmt)
+{
+	if( stmt->cmd_type == PLTSQL_STMT_EXECSQL)
+	{
+		PLtsql_stmt_execsql	*execsql = (PLtsql_stmt_execsql *) stmt;
+		return execsql->is_set_tran_isolation;
+	}
+	else
+		return true;
+}
+
+static
 void
 record_error_state(PLtsql_execstate *estate)
 {
@@ -1192,7 +1205,7 @@ dispatch_stmt_handle_error(PLtsql_execstate *estate,
 		 * savepoints and let the caller be responsible for handling the
 		 * error.
 		 */
-		if (!ro_func && !pltsql_disable_internal_savepoint && !is_batch_command(stmt) && IsTransactionBlockActive())
+		if (!ro_func && !pltsql_disable_internal_savepoint && !is_batch_command(stmt) && IsTransactionBlockActive() && !is_set_tran_isolation(stmt) )
 		{
 			elog(DEBUG5, "TSQL TXN Start internal savepoint");
 			BeginInternalSubTransaction(NULL);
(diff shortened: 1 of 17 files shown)


contrib/babelfishpg_tsql/src/iterative_exec.c
```diff
@@ -979,6 +979,19 @@ is_batch_command(PLtsql_stmt *stmt)
...
 }
 
+static
+bool
+is_set_tran_isolation(PLtsql_stmt *stmt)
+{
+	if( stmt->cmd_type == PLTSQL_STMT_EXECSQL)
+	{
+		PLtsql_stmt_execsql	*execsql = (PLtsql_stmt_execsql *) stmt;
+		if(execsql->is_set_tran_isolation)
```
surendravishnoi, line 989: just do return execsql->is_set_tran_isolation

contrib/babelfishpg_tsql/src/pltsql.h
```diff
@@ -1048,6 +1048,7 @@ typedef struct PLtsql_stmt_execsql
 	char	   *db_name;		/* db_name: only for cross db query */
 	bool		is_schema_specified;	/* is schema name specified? */
 	bool		is_create_view; /* CREATE VIEW? */
+	bool		is_set_tran_isolation; /* SET TRANSACTION ISOLATION*/
```
surendravishnoi, line 1051: /* comment */

contrib/babelfishpg_tsql/src/tsqlIface.cpp
```diff
@@ -4561,6 +4561,13 @@ makeSetStatement(TSqlParser::Set_statementContext *ctx, tsqlBuilder &builder)
 		}
 		else if (set_special_ctx->BABELFISH_STATISTICS() && set_special_ctx->PROFILE())
 			return makeSetExplainModeStatement(ctx, false);
+		else if(set_special_ctx->ISOLATION())
+		{
+			PLtsql_stmt_execsql *stmt = (PLtsql_stmt_execsql *) makeSQL(ctx);
+			stmt->is_set_tran_isolation = true;
+			attachPLtsql_fragment(ctx, (PLtsql_stmt *) stmt);
```
surendravishnoi, line 4568: why this attach is required, in default case we only do makeSQL?

test/JDBC/expected/BABEL_4411.out
```diff
@@ -0,0 +1,569 @@
+
```
surendravishnoi, line 1: please add test cases to test isolation levels inside proc/func/views

PR #1848, opened 2023-09-18
Support GRANT .. ON SCHEMA .. in Babelfish

shalinilohia50:
### Description
Support GRANT/REVOKE .. ON SCHEMA .. in Babelfish

1. Supported syntax `GRANT <permission> ON SCHEMA::<schema_name> TO <user_name>`
2. Supported syntax `REVOKE <permission> ON SCHEMA::<schema_name> FROM <user_name>`
3. Added one SQL statement `PLTSQL_STMT_GRANTSCHEMA` to store relevant information to execute GRANT/REVOKE .. ON SCHEMA .. statements.
4. Created one catalog table `sys.babelfish_schema_permissions` to hold the details about schema name, database name, object name, permission name and user name for each GRANT/REVOKE statements.
5. GRANT on schema/objects adds a row in the catalog table if not exists already.
6. REVOKE on schema/objects removes the corresponding row in the catalog table if it exists already.
7. REVOKE on schema internally grants permission to all the objects if there are explicit permissions granted to the objects belonging to the same schema.
8. GRANT ALL on objects work as it is and add rows in the catalog for each relevant permission depending on the object type.
9. REVOKE ALL on object should do nothing, if the relevant schema permission exists in the catalog. But, it should remove the rows from the catalog if the object level permission is granted.
10. Drop statement for OBJECT/SCHEMA removes all the relevant object entries from the catalog. 

Design Doc: https://quip-amazon.com/cWZYAprElLog/GRANT-ON-SCHEMA-for-Babelfish 
Test plan: https://quip-amazon.com/pYXHA4sy2Og1/Test-Plan-for-GRANTREVOKE-on-SCHEMA

### Issues Resolved

Task: [BABEL-4344](https://jira.rds.a2z.com/browse/BABEL-4344)

* **Use case based -** Yes

* **Arbitrary inputs -** N/A

* **Performance tests -** N/A

* **Tooling impact -** N/A

* **Client tests -** N/A


contrib/babelfishpg_tsql/sql/ownership.sql
```diff
@@ -14,6 +14,18 @@ CREATE TABLE sys.babelfish_sysdatabases (
 
 GRANT SELECT on sys.babelfish_sysdatabases TO PUBLIC;
 
+-- BABELFISH_SCHEMA_PERMISSIONS
+CREATE TABLE sys.babelfish_schema_permissions (
+  db_name NAME NOT NULL,
```
HarshLunagariya: In other catalogs, dbname is stored using text datatype. Can we use dbid instead to avoid all this datatype complexity?

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -83,6 +83,12 @@ Oid			bbf_servers_def_idx_oid;
 Oid			bbf_function_ext_oid;
 Oid			bbf_function_ext_idx_oid;
 
+/*****************************************
+ *			SCHEMA
+ *****************************************/
+Oid			bbf_schema_oid;
+Oid			bbf_schema_idx_oid;
```
HarshLunagariya: Please rename it to some meaningful name.

contrib/babelfishpg_tsql/sql/ownership.sql
```diff
@@ -14,6 +14,18 @@ CREATE TABLE sys.babelfish_sysdatabases (
...
+  schema_name NAME NOT NULL,
+  object_name NAME NOT NULL,
+  permission NAME NOT NULL,
+  grantee NAME NOT NULL,
+  object_type NAME,
+  PRIMARY KEY(db_name, schema_name, object_name, permission, grantee)
```
HarshLunagariya: We should include object_type in primary key as well.

contrib/babelfishpg_tsql/sql/ownership.sql
```diff
@@ -14,6 +14,18 @@ CREATE TABLE sys.babelfish_sysdatabases (
...
+);
+GRANT SELECT ON sys.babelfish_schema_permissions TO PUBLIC;
```
HarshLunagariya: Please avoid GRANTing SELECT on catalog tables to PUBLIC.

contrib/babelfishpg_tsql/src/pltsql_utils.c
```diff
@@ -1683,4 +1683,57 @@ Oid get_sys_varcharoid(void)
...
+			appendStringInfo(&query, "REVOKE \"%s\" ON ALL PROCEDURES IN SCHEMA \"%s\" FROM \"%s\"; ", privilege, schema, rolname);
+		}
+		else
+		{
+			appendStringInfo(&query, "REVOKE \"%s\" ON ALL TABLES IN SCHEMA \"%s\" FROM \"%s\"; ", privilege, schema, rolname);
+			appendStringInfo(&query, "ALTER DEFAULT PRIVILEGES IN SCHEMA \"%s\" REVOKE \"%s\" ON TABLES FROM \"%s\"; ", schema, privilege, rolname);
+		}
+	}
+	stmt_list = raw_parser(query.data, RAW_PARSE_DEFAULT);
+	if (list_length(stmt_list) != expected_stmts)
```
HarshLunagariya: There can be possibility of SQL injection. Ideal way is to replace these variable in parse tree.
Will the length check for expected_stmt be enough to make sure that there is no SQL injection possiblity?

contrib/babelfishpg_tsql/src/pltsql.h
```diff
@@ -185,8 +185,9 @@ typedef enum PLtsql_stmt_type
 	PLTSQL_STMT_RESTORE_CTX_FULL,
 	PLTSQL_STMT_RESTORE_CTX_PARTIAL,
 	PLTSQL_STMT_INSERT_BULK,
-	PLTSQL_STMT_GRANTDB
-} PLtsql_stmt_type;
+	PLTSQL_STMT_GRANTDB,
+	PLTSQL_STMT_GRANTSCHEMA
+	} PLtsql_stmt_type;
```
HarshLunagariya: Fix indent

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -3287,3 +3288,92 @@ get_insert_bulk_kilobytes_per_batch()
...
+	 * sysadmin or login is not the schema owner, then it doesn't have the permission to GRANT/REVOKE.
+	 */
+	login_is_db_owner = 0 == strncmp(login, get_owner_of_db(dbname), NAMEDATALEN);
+	datdba = get_role_oid("sysadmin", false);
+	schema_name = get_physical_schema_name(dbname, stmt->schema_name);
+	schemaOid = LookupExplicitNamespace(schema_name, true);
+	
+	if (!is_member_of_role(GetSessionUserId(), datdba) && !login_is_db_owner && !pg_namespace_ownercheck(schemaOid, GetUserId()))
+		ereport(ERROR,
+				(errcode(ERRCODE_INTERNAL_ERROR),
```
HarshLunagariya: Use correct errorcode.

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -3287,3 +3288,92 @@ get_insert_bulk_kilobytes_per_batch()
...
+	ListCell	*lc1;
+	Oid 		schemaOid;
+
+	/*
+	 * If the login is not the db owner or the login is not the member of
+	 * sysadmin or login is not the schema owner, then it doesn't have the permission to GRANT/REVOKE.
+	 */
+	login_is_db_owner = 0 == strncmp(login, get_owner_of_db(dbname), NAMEDATALEN);
+	datdba = get_role_oid("sysadmin", false);
+	schema_name = get_physical_schema_name(dbname, stmt->schema_name);
```
HarshLunagariya, line 3313: Will it have Cross-db scenarios?

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -3287,3 +3288,92 @@ get_insert_bulk_kilobytes_per_batch()
...
+	schemaOid = LookupExplicitNamespace(schema_name, true);
```
HarshLunagariya, line 3314: What if Oid is not valid?

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -3432,6 +3473,241 @@ bbf_ProcessUtility(PlannedStmt *pstmt,
 				}
 				break;
 			}
+		case T_GrantStmt:
+			{
+				GrantStmt *grant = (GrantStmt *) parsetree;
+				char	   *dbname = get_cur_db_name();
+				const char *current_user = GetUserNameFromId(GetUserId(), false);
+				/* If object is table. */
```
HarshLunagariya: ACL_TARGET_OBJECT is for named objects specified in GRANT statement. Please update the comment to reflect the same.

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -3432,6 +3473,241 @@ bbf_ProcessUtility(PlannedStmt *pstmt,
...
+						if (prev_ProcessUtility)
+							prev_ProcessUtility(pstmt, queryString, readOnlyTree, context, params,
+								queryEnv, dest, qc);
+						else
+							standard_ProcessUtility(pstmt, queryString, readOnlyTree, context, params,
+								queryEnv, dest, qc);
+						return;
+					}
+					/* Ignore CREATE database subcommands */
+					if (!(pstmt->stmt_len == 18 || pstmt->stmt_len == 19))
```
HarshLunagariya: We should compare the querystring. It can be broken if some new subcommand is added with same length.

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -3432,6 +3473,241 @@ bbf_ProcessUtility(PlannedStmt *pstmt,
...
+				/* If object is table. */
+				if (sql_dialect != SQL_DIALECT_TSQL || grant->targtype != ACL_TARGET_OBJECT)
+					break;
+				Assert(list_length(grant->objects) == 1);
+				if (grant->objtype == OBJECT_SCHEMA)
+						break;
+				else if (grant->objtype == OBJECT_TABLE)
+				{
+					/* Don't add any entry for the GRANT statements executed while REVOKE on Schema. */
+					if (pstmt->stmt_len == 1 && grant->is_grant)
```
HarshLunagariya: Didn't get the context. Can you please explain this scenarios?

HarshLunagariya (changes requested):

contrib/babelfishpg_tsql/src/pltsql_utils.c
shalinilohia50: IMO this would be enough. 
We have been updating the query string directly for GRANT/REVOKE without actually replacing the variables in parse tree.

contrib/babelfishpg_tsql/src/pl_exec-2.c
shalinilohia50: No, it's not allowed. We can only grant or revoke permissions on objects in the current database.

contrib/babelfishpg_tsql/sql/ownership.sql
shalinilohia50: We have only stored object_type in case of function since that in needed at the time of `REVOKE .. ON SCHEMA`. It will be NULL in other cases now.

contrib/babelfishpg_tsql/src/pl_exec-2.c
shalinilohia50: Addressed it.

contrib/babelfishpg_tsql/src/pl_handler.c
shalinilohia50: It was needed. Removed this part.

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -2799,6 +2828,346 @@ rename_procfunc_update_bbf_catalog(RenameStmt *stmt)
...
+add_entry_to_bbf_schema(const char *schema_name,
+				const char *object_name,
+				const char *permission,
+				const char *grantee,
+				const char *object_type)
+{
+	Relation	bbf_schema_rel;
+	TupleDesc	bbf_schema_dsc;
+	HeapTuple	tuple_bbf_schema;
+	Datum		new_record_bbf_schema[6];
```
xhfanhe: Better to use a macro instead of "6", in case we want to extend this catalog in the future. Same applies for the attribute numbers.

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -2799,6 +2828,346 @@ rename_procfunc_update_bbf_catalog(RenameStmt *stmt)
...
+{
+	Relation	bbf_schema_rel;
+	HeapTuple	tuple_bbf_schema;
+	ScanKeyData	key[5];
+	TableScanDesc	scan;
+	bool	catalog_entry_exists = false;
+	int16	dbid = get_cur_db_id();
+
+	bbf_schema_rel = table_open(get_bbf_schema_perms_oid(),
+									RowExclusiveLock);
```
xhfanhe: RowExclusiveLock is only needed in UPDATE, DELETE, INSERT and LOCK TABLE scenarios. Here we only need AccessShareLock.

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -2799,6 +2828,346 @@ rename_procfunc_update_bbf_catalog(RenameStmt *stmt)
...
+{
+	Relation	bbf_schema_rel;
+	HeapTuple	tuple_bbf_schema;
+	ScanKeyData key[4];
+	TableScanDesc scan;
+	bool		catalog_entry_exists = false;
+	int16	dbid = get_cur_db_id();
+
+	bbf_schema_rel = table_open(get_bbf_schema_perms_oid(),
+									RowExclusiveLock);
```
xhfanhe: Same as above

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -2799,6 +2828,346 @@ rename_procfunc_update_bbf_catalog(RenameStmt *stmt)
...
+	HeapTuple	tuple_bbf_schema;
+	const char	*object_name;
+	const char	*object_type;
+	ScanKeyData scanKey[4];
+	int16		dbid = get_cur_db_id();
+	const char *db_name = get_cur_db_name();
+
+	/* Fetch the relation */
+	bbf_schema_rel = table_open(get_bbf_schema_perms_oid(),
+									RowExclusiveLock);
```
xhfanhe: Same as above

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -3239,7 +3240,45 @@ bbf_ProcessUtility(PlannedStmt *pstmt,
...
+								Node	   *stmt = ((RawStmt *) lfirst(parsetree_item))->stmt;
+								PlannedStmt *wrapper;
+
+								/* need to make a wrapper PlannedStmt */
+								wrapper = makeNode(PlannedStmt);
+								wrapper->commandType = CMD_UTILITY;
+								wrapper->canSetTag = false;
+								wrapper->utilityStmt = stmt;
+								wrapper->stmt_location = 0;
+								wrapper->stmt_len = 16;
```
xhfanhe: Where does this number come from?

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -3518,6 +3559,233 @@ bbf_ProcessUtility(PlannedStmt *pstmt,
...
+					if (strcmp("(CREATE LOGICAL DATABASE )", queryString) != 0)
+					{
+						RangeVar   *rv = (RangeVar *) linitial(grant->objects);
+						const char *logical_schema = NULL;
+						char	   *obj = rv->relname;
+						ListCell   *lc;
+						ListCell	*lc1;
+						const char *obj_type = "r";
+						if (rv->schemaname != NULL)
+							logical_schema = get_logical_schema_name(rv->schemaname, true);	/* this is physical name */
```
xhfanhe: What is physical name?

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -5634,6 +5904,8 @@ bbf_ExecDropStmt(DropStmt *stmt)
 											 major_name, NULL);
 				}
 			}
+			if (logicalschema != NULL)
```
xhfanhe, line 5929: When will logicalschema be NULL? Should we allow that?

contrib/babelfishpg_tsql/src/pltsql_utils.c
xhfanhe: We should definitely avoid this style... `expected_stmt` is not enough, we tried that before but raised third-party concern and we ended up fixing them all ;)

test/python/expected/sql_validation_framework/expected_create.out
```diff
@@ -72,6 +72,7 @@ Could not find tests for procedure sys.printarg
 Could not find tests for procedure sys.sp_cursor_list
 Could not find tests for procedure sys.sp_describe_cursor
 Could not find tests for table sys.babelfish_helpcollation
+Could not find tests for table sys.babelfish_schema_permissions
```
xhfanhe, line 75: Can we add tests for this?

test/JDBC/jdbc_schedule
```diff
@@ -33,6 +33,9 @@ ignore#!#BABEL-3117-vu-prepare
 ignore#!#BABEL-3117-vu-verify
 ignore#!#BABEL-3655-vu-prepare
 ignore#!#BABEL-3655-vu-verify
+ignore#!#GRANT_SCHEMA-vu-prepare
```
xhfanhe, line 36: What would cause failure in normal JDBC run?

test/python/expected/sql_validation_framework/expected_create.out
shalinilohia50: No, we have not granted PUBLIC access to this catalog table since it is only meant for internal use.

test/JDBC/jdbc_schedule
shalinilohia50: `GRANT_SCHEMA-vu-verify` would fail due to difference in few cases while handling upgrade since. Those scenarios are mentioned in the test plan added in the description.

contrib/babelfishpg_tsql/src/pl_handler.c
shalinilohia50: Oh! Removed this misleading comment.

contrib/babelfishpg_tsql/src/pl_handler.c
shalinilohia50: Made it 0.

contrib/babelfishpg_tsql/src/pl_handler.c
shalinilohia50: While dropping table valued function, when we drop the internal table then the logical schema becomes NULL.
We have to avoid that case.
Example from JDBC test framework:
```
CREATE FUNCTION my_splitstring_3101 ( @stringToSplit VARCHAR(MAX) )
RETURNS
@returnList TABLE ([Value] [nvarchar] (50))
AS
BEGIN
	DECLARE @name NVARCHAR(255)
	DECLARE @pos INT
	WHILE CHARINDEX(',', @stringToSplit) > 0
		BEGIN
			SELECT @pos  = CHARINDEX(',', @stringToSplit)
			SELECT @name = SUBSTRING(@stringToSplit, 1, @pos-1)
			INSERT INTO @returnList SELECT @name
			SELECT @stringToSplit = SUBSTRING(@stringToSplit, @pos+1, LEN(@stringToSplit)-@pos)
		END
		INSERT INTO @returnList SELECT @stringToSplit
		RETURN
END
GO
```
It executes `DROP TABLE @returnlist_1` internally where makes logicalschema = NULL.

xhfanhe (dismissed):

xhfanhe (approved):

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -3287,3 +3288,98 @@ get_insert_bulk_kilobytes_per_batch()
...
+							NULL,
+							None_Receiver,
+							NULL);
+
+				/* make sure later steps can see the object created here */
+				CommandCounterIncrement();
+			}
+			/* Add entry for each grant statement. */
+			if (stmt->is_grant && !check_bbf_schema_for_entry(stmt->schema_name, "ALL", priv_name, rolname))
+				add_entry_to_bbf_schema(stmt->schema_name, "ALL", priv_name, rolname, NULL);
```
KushaalShroff, line 3374: Why is permission and grantee column of table not an array?  This is going to bloat our memory drastically.
In one query it is possible for the customer to assign m permissions to n grantees. In that case you are adding m*n rows.
In PG CATALOG we have these as an array with a bit of performance overhead

contrib/babelfishpg_tsql/src/tsqlIface.cpp
```diff
@@ -5357,6 +5403,108 @@ makeGrantdbStatement(TSqlParser::Security_statementContext *ctx)
...
+					char *grantee_name = pstrdup(downcase_truncate_identifier(id_str.c_str(), id_str.length(), true));
+					grantee_list = lappend(grantee_list, grantee_name);
+				}
+			}
+			List *privilege_list = NIL;
+			for (auto perm: ctx->grant_statement()->permissions()->permission())
+			{
+				auto single_perm = perm->single_permission();
+				if (single_perm->EXECUTE())
+					privilege_list = lappend(privilege_list, (void *)"execute");
```
KushaalShroff, line 5436: Anju -- You can directly store the bitmap masks here

KushaalShroff: You can also do the operations here itself

PR #1860, opened 2023-09-22
support type_id and type_name t-sql functions

vaibhav412:
### Description

This commit intends to support TYPE_ID() & TYPE_NAME() T-SQL Functions.

(a) TYPE_ID()
Input: [schema_name].[type_name] OR [type_name]
Output: user_type_id of the datatype
Info: Datatype can be user defined or system default datatype.
-> User Defined datatype: 
If new type is created in default schema & same typename is created in another schema, then both will have different type_id.
Example: TYPE_NAME('NewSchema.NewType'), TYPE_ID('NewType') - Both will have different type_id

-> System datatype:
Example: TYPE_ID('int'), TYPE_ID('sys.int'), TYPE_ID('pg_catalog.int') - All will give same type_id.
Reference: [Doc](https://github.com/MicrosoftDocs/sql-docs/blob/live/docs/t-sql/functions/type-id-transact-sql.md)

(b) TYPE_NAME()
Input: [type_id]
Output: name of the datatype
Info: It will give the name of the datatype as mapped according to the input type_id.
Example: TYPE_NAME(36)
Reference: [Doc](https://github.com/MicrosoftDocs/sql-docs/blob/live/docs/t-sql/functions/type-name-transact-sql.md)

Authored-by: Vaibhav Agarwal [vaibhagr@amazon.com](mailto:vaibhagr@amazon.com)
Signed-off-by: Vaibhav Agarwal [vaibhagr@amazon.com](mailto:vaibhagr@amazon.com)

### Issues Resolved

[BABEL-803]

* **Use case based -**
Added test cases and covered different possible scenarios.

* **Arbitrary inputs -** N/A

* **Performance tests -** N/A

* **Tooling impact -** N/A

* **Client tests -** N/A

### Check List


contrib/babelfishpg_tsql/sql/sys_functions.sql
```diff
@@ -4418,3 +4418,52 @@ BEGIN
 END;
 $body$
 LANGUAGE plpgsql IMMUTABLE;
+
+CREATE OR REPLACE FUNCTION TYPE_NAME(IN type_id INT)
+RETURNS VARCHAR AS $$
```
rishabhtanwar29: Issue: T-SQL documentation states that the return type should be `sysname`.

contrib/babelfishpg_tsql/sql/sys_functions.sql
```diff
@@ -4418,3 +4418,52 @@ BEGIN
...
+    BEGIN
+        RETURN (SELECT ty.name FROM sys.types ty WHERE ty.user_type_id = type_id);
+    END;
+$$
+LANGUAGE plpgsql IMMUTABLE
+STRICT;
+
+CREATE OR REPLACE FUNCTION TYPE_ID(IN typename VARCHAR)
```
rishabhtanwar29: 1. Please check the T-SQL documentation. argument type should be `sys.nvarchar`.
2. Let's have the schema qualified names here, ie: `SYS.TYPE_ID`.

contrib/babelfishpg_tsql/sql/sys_functions.sql
```diff
@@ -4418,3 +4418,52 @@ BEGIN
 END;
 $body$
 LANGUAGE plpgsql IMMUTABLE;
+
+CREATE OR REPLACE FUNCTION TYPE_NAME(IN type_id INT)
```
rishabhtanwar29: Let's have the schema qualified names here, ie: `SYS.TYPE_NAME`

contrib/babelfishpg_tsql/sql/sys_functions.sql
```diff
@@ -4418,3 +4418,52 @@ BEGIN
...
+
+CREATE OR REPLACE FUNCTION TYPE_ID(IN typename VARCHAR)
+RETURNS INT AS $$
+    DECLARE
+        schemaid int;
+        schema_name VARCHAR;
+        type_name VARCHAR;
+    BEGIN
+        IF typename LIKE '%.%' THEN
+            schema_name := TRIM(LOWER(split_part(typename COLLATE "C", '.', 1)));
```
rishabhtanwar29: Question: Why collate "C" here and collate sys.database_default below?

contrib/babelfishpg_tsql/sql/sys_functions.sql
```diff
@@ -4418,3 +4418,52 @@ BEGIN
...
+        IF NOT EXISTS (SELECT ty.name FROM sys.types ty WHERE ty.name = type_name COLLATE sys.database_default)
+        THEN
+            RETURN NULL;
+        END IF;
+        
+        -- Check if schema_name exists
+        IF NOT EXISTS (SELECT sc.name FROM sys.schemas sc WHERE sc.name = schema_name COLLATE sys.database_default OR schema_name = 'sys' OR schema_name = 'pg_catalog')
+        THEN
+            RETURN NULL;
+        END IF;
```
rishabhtanwar29: Can we get rid of these two steps and return NULL or result directly below so as to reduce unnecessary catalog lookups?

contrib/babelfishpg_tsql/sql/sys_functions.sql
```diff
@@ -4418,3 +4418,52 @@ BEGIN
...
+
+        IF NOT EXISTS (SELECT ty.name FROM sys.types ty WHERE ty.name = type_name COLLATE sys.database_default AND ty.is_user_defined = 0) 
+        THEN
+            schemaid := (SELECT sc.schema_id FROM sys.schemas sc WHERE sc.name = schema_name COLLATE sys.database_default);
+            RETURN (SELECT ty.user_type_id FROM sys.types ty WHERE ty.schema_id = schemaid AND ty.name = type_name COLLATE sys.database_default);
```
rishabhtanwar29: Suggestion: Can be done with only one query using joins.

contrib/babelfishpg_tsql/sql/sys_functions.sql
vaibhav412: Updated!

contrib/babelfishpg_tsql/sql/sys_functions.sql
vaibhav412: Updated!

contrib/babelfishpg_tsql/sql/sys_functions.sql
vaibhav412: Done

contrib/babelfishpg_tsql/sql/sys_functions.sql
vaibhav412: Used Collate "C" everywhere in the new commit

contrib/babelfishpg_tsql/sql/sys_functions.sql
vaibhav412: Done

contrib/babelfishpg_tsql/sql/sys_functions.sql
vaibhav412: We can remove type_name check but I think schema_name check will be required, as for system datatypes we are directly returning type_name from sys.types, so to handle the case of entering random schema name with system datatype, we need the check. eg: schema_id('random.int')

rishabhtanwar29 (approved):

test/JDBC/input/functions/typeid-typename-dep-vu-prepare.sql
```diff
@@ -0,0 +1,25 @@
...
+GO
+
+CREATE VIEW typeid_typename_dep_vu_prepare_view2 AS
+SELECT TYPE_NAME(TYPE_ID('datetime')) AS [TYPE_NAME];
+GO
+
+CREATE PROC typeid_typename_dep_vu_prepare_proc2 AS
+SELECT TYPE_NAME(TYPE_ID('datetime')) AS [TYPE_NAME];
+GO
+
```
ahmed-shameem, line 25: Let us have one more login:
```CREATE LOGIN log1 WITH PASSWORD = '12345678'```
Login using the newly created login, and then call type_id, type_name functions for the valid types which we have created. It should return null.

File: contrib/babelfishpg_tsql/sql/sys_functions.sql [hunk 8]
```diff
@@ -4418,3 +4418,45 @@ BEGIN
...
+$$
+LANGUAGE plpgsql IMMUTABLE
+STRICT;
+
+CREATE OR REPLACE FUNCTION SYS.TYPE_ID(IN typename SYS.NVARCHAR)
+RETURNS INT AS $$
+    DECLARE
+        schemaid int;
+        schema_name VARCHAR;
+        type_name VARCHAR;
```
ahmed-shameem: Rather than using ```VARCHAR```, we should use ```NVARCHAR(128)``` as there is a possibility that user may have a schema or type name with a different collation.

contrib/babelfishpg_tsql/sql/sys_functions.sql
```diff
@@ -4418,3 +4418,45 @@ BEGIN
...
+    BEGIN
+        IF typename LIKE '%.%' THEN
+            schema_name := TRIM(LOWER(split_part(typename COLLATE "C", '.', 1)));
+            type_name :=  TRIM(LOWER((split_part(typename COLLATE "C",'.', 2))));
```
ahmed-shameem: Why collate 'C'? Can't it be any other collation?

ahmed-shameem (changes requested):

contrib/babelfishpg_tsql/sql/sys_functions.sql
```diff
@@ -4418,3 +4418,45 @@ BEGIN
...
 LANGUAGE plpgsql IMMUTABLE;
+
+CREATE OR REPLACE FUNCTION SYS.TYPE_NAME(IN type_id INT)
+RETURNS SYSNAME AS $$
+    BEGIN
+        RETURN (SELECT ty.name FROM sys.types ty WHERE ty.user_type_id = type_id);
+    END;
+$$
+LANGUAGE plpgsql IMMUTABLE
+STRICT;
```
Deepesh125: why can't we implement it in pltsql?

contrib/babelfishpg_tsql/sql/sys_functions.sql
```diff
@@ -4418,3 +4418,45 @@ BEGIN
...
+
+CREATE OR REPLACE FUNCTION SYS.TYPE_ID(IN typename SYS.NVARCHAR)
+RETURNS INT AS $$
+    DECLARE
+        schemaid int;
+        schema_name VARCHAR;
+        type_name VARCHAR;
+    BEGIN
+        IF typename LIKE '%.%' THEN
```
Deepesh125: what if typename is passed as lets say, "ab.c"."my.type"?

contrib/babelfishpg_tsql/sql/sys_functions.sql
```diff
@@ -4418,3 +4418,45 @@ BEGIN
...
+            schema_name := TRIM(LOWER(split_part(typename COLLATE "C", '.', 1)));
+            type_name :=  TRIM(LOWER((split_part(typename COLLATE "C",'.', 2))));
+        ELSE
+            schema_name := 'dbo';
+            type_name := TRIM(LOWER(typename COLLATE "C"));
+        END IF;
+
+        -- Check if schema_name exists
+        IF NOT EXISTS (SELECT sc.name FROM sys.schemas sc WHERE sc.name = schema_name OR schema_name = 'sys' OR schema_name = 'pg_catalog')
```
Deepesh125: can we use our sys extended catalog like babelfish_namespace_ext? it would faster than this reference to view

contrib/babelfishpg_tsql/sql/sys_functions.sql
```diff
@@ -4418,3 +4418,45 @@ BEGIN
...
+        THEN
+            RETURN NULL;
+        END IF;
+
+        IF NOT EXISTS (SELECT ty.name FROM sys.types ty WHERE ty.name = type_name COLLATE "C" AND ty.is_user_defined = 0)
```
Deepesh125: why COLLATE "C"? should have used COLLATE sys.database_default

contrib/babelfishpg_tsql/sql/sys_functions.sql
```diff
@@ -4418,3 +4418,45 @@ BEGIN
...
+        END IF;
+
+        IF NOT EXISTS (SELECT ty.name FROM sys.types ty WHERE ty.name = type_name COLLATE "C" AND ty.is_user_defined = 0) 
+        THEN
+            RETURN (SELECT ty.user_type_id FROM sys.schemas sc JOIN sys.types ty ON sc.schema_id = ty.schema_id WHERE sc.name = schema_name AND ty.name = type_name COLLATE "C");
+        ELSE
+            RETURN (SELECT ty.user_type_id FROM sys.types ty WHERE ty.name = type_name COLLATE "C");
+        END IF;
+    END;
+$$
```
Deepesh125: as per T-SQL documentation, TYPE_ID returns NULL if the type name is not valid, or if the caller does not have sufficient permission to reference the type. Where are we checking it?

contrib/babelfishpg_tsql/sql/sys_functions.sql
```diff
@@ -4418,3 +4418,45 @@ BEGIN
...
+LANGUAGE plpgsql IMMUTABLE
```
Deepesh125: should try to implement in TSQL language

test/JDBC/input/functions/typeid-typename-dep-vu-prepare.sql
```diff
@@ -0,0 +1,25 @@
+CREATE TYPE typeid_typename_dep_vu_prepare_t1 FROM INT;
+GO
+
+CREATE SCHEMA typeid_typename_dep_vu_prepare_s1;
+GO
+
+CREATE TYPE typeid_typename_dep_vu_prepare_s1.typeid_typename_dep_vu_prepare_t1 FROM INT;
```
Deepesh125, line 7: overall test cases should involve, 
1/ delimited identifiers along with multi byte and single byte chars with all kind of variations
2/ permission checks on type usage

Deepesh125 (changes requested):

test/JDBC/input/functions/typeid-typename-dep-vu-prepare.sql
vaibhav412: Added

contrib/babelfishpg_common/src/babelfishpg_common.h
```diff
@@ -55,6 +55,7 @@ typedef struct common_utility_plugin
 	Datum		(*datetime2sqlvariant) (PG_FUNCTION_ARGS);
 	Datum		(*tinyint2sqlvariant) (PG_FUNCTION_ARGS);
 	Datum		(*translate_pg_type_to_tsql) (PG_FUNCTION_ARGS);
+	Oid         (*get_tsql_datatype_oid) (char *type_name);
```
ahmed-shameem: nit: indentation

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -1673,6 +1675,200 @@ object_name(PG_FUNCTION_ARGS)
...
+ */
+Datum
+type_id(PG_FUNCTION_ARGS)
+{
+    char       *db_name,
+               *schema_name,
+               *object_name;
+    char       *physical_schema_name;
+    char       *input;
+    char      **splited_object_name;
```
ahmed-shameem: nit: indentation

ahmed-shameem: *splitted_object_name;
Update the same in other places as well.

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -1673,6 +1675,200 @@ object_name(PG_FUNCTION_ARGS)
...
+    Oid         schema_oid;
+    Oid         user_id = GetUserId();
+    Oid result = InvalidOid;
```
ahmed-shameem: nit: indentation

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -1673,6 +1675,200 @@ object_name(PG_FUNCTION_ARGS)
...
+    int         i;
+
+    if (PG_ARGISNULL(0))
+        PG_RETURN_NULL();
+    input = text_to_cstring(PG_GETARG_TEXT_P(0));
```
ahmed-shameem: Better to use PG_GETARG_TEXT_PP. Ref: https://postgrespro.com/list/thread-id/2604215

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -1673,6 +1675,200 @@ object_name(PG_FUNCTION_ARGS)
...
+
+    /* strip trailing whitespace from input */
+    i = strlen(input);
+    while (i > 0 && isspace((unsigned char) input[i - 1]))
+        input[--i] = '\0';
+
+    /* length should be restricted to 4000 */
+    if (i > 4000)
```
ahmed-shameem: Try to find whether there is any existing macro that we can use which specifies the object name length or define a macro and set the value to 4000

test/JDBC/upgrade/master/schedule
```diff
@@ -333,6 +333,7 @@ TestVariableDataLength
 TestXML
 timefromparts
 triggers_with_transaction
+typeid-typename
```
ahmed-shameem, line 336: Why are we not adding typeid-typename-dep here?

ahmed-shameem (changes requested):

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -1673,6 +1675,200 @@ object_name(PG_FUNCTION_ARGS)
...
+        PG_RETURN_NULL();
+    
+    InitFunctionCallInfoData(*fcinfo1, NULL, 0, InvalidOid, NULL, NULL);
+    fcinfo1->args[0].value = type_id;
+    fcinfo1->args[0].isnull = false;
+    // Search in typecode list, it will return type name if system datatype, else will return null.
+    tsql_typename = (*common_utility_plugin_ptr->translate_pg_type_to_tsql) (fcinfo1);
+    if (tsql_typename)
+    {
+        PG_RETURN_VARCHAR_P(tsql_typename);
```
rishabhtanwar29: Question: Don't we need aclcheck in this case?

vaibhav412: System Datatypes should be visible to all users, so I think it is not required, right?

test/JDBC/upgrade/master/schedule
vaibhav412: Dump restore check was failing due to logging in from different user, so I confirmed it with rishabh to remove it.

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -1673,6 +1675,200 @@ object_name(PG_FUNCTION_ARGS)
...
+    int         i;
+
+    if (PG_ARGISNULL(0))
+        PG_RETURN_NULL();
+    input = text_to_cstring(PG_GETARG_TEXT_P(0));
+
+    /* strip trailing whitespace from input */
+    i = strlen(input);
+    while (i > 0 && isspace((unsigned char) input[i - 1]))
+        input[--i] = '\0';
```
Deepesh125: instead find the end of string and set input[i] = '\0' only once

contrib/babelfishpg_tsql/runtime/functions.c
Deepesh125: Do we have test cases to show this? Input is sys.nvarchar, what would be default length/typmod?

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -1673,6 +1675,200 @@ object_name(PG_FUNCTION_ARGS)
...
+    /* length should be restricted to 4000 */
+    if (i > 4000)
+        ereport(ERROR,
+                (errcode(ERRCODE_STRING_DATA_LENGTH_MISMATCH),
+                 errmsg("input value is too long for object name")));
+
+    /* resolve the two part name */
+    splited_object_name = split_object_name(input);
+    /* If three part name(db_name also included in input) then return null */
+    if(strcmp(splited_object_name[1], ""))
```
Deepesh125: check if splited_object_name[1] is not null before attempting to do strcmp

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -1673,6 +1675,200 @@ object_name(PG_FUNCTION_ARGS)
...
+    {
+        pfree(input);
+        pfree(splited_object_name);
```
Deepesh125: splited_object_name is double pointer, clear out inner pointers first and then outer one

contrib/babelfishpg_tsql/runtime/functions.c
Deepesh125: And better to use strlen here

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -1673,6 +1675,200 @@ object_name(PG_FUNCTION_ARGS)
...
+
+    if (!strcmp(schema_name, ""))
+    {
+        // To check if it is a system datatype, search in typecode list and it will give result oid, else if not it will return null.
+        result = (*common_utility_plugin_ptr->get_tsql_datatype_oid) (object_name);
+
+        // If null result, then it is not system datatype and now search with default schema in pg_type
+        if (!OidIsValid(result))
+        {
+            /* find the default schema for current user and get physical schema name */
```
Deepesh125, line 1893: Have we added test cases to show this behaviour? For example, 

```
create type dbo.int from int
```
trying to find int for above should return oid corresponding to sys.int

```
create type dbo.mytype frm int
create type  my_sch.mytype from int
```
typeid should return corresponding to dbo one

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -1673,6 +1675,200 @@ object_name(PG_FUNCTION_ARGS)
...
+    if (!OidIsValid(schema_oid) || pg_namespace_aclcheck(schema_oid, user_id, ACL_USAGE) != ACLCHECK_OK)
+    {
+        pfree(schema_name);
+        pfree(object_name);
+        PG_RETURN_NULL();
+    }
+    // If schema is 'sys' or 'pg_catalog' then search in typecode list, else in pg_type.
+    if(!strcmp(schema_name, "sys") || !strcmp(schema_name, "pg_catalog"))
+    {
+        result = (*common_utility_plugin_ptr->get_tsql_datatype_oid) (object_name);
```
Deepesh125: does table type_infos contains all the system typeS?

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -1673,6 +1675,200 @@ object_name(PG_FUNCTION_ARGS)
...
+    HeapTuple   tuple;
+    Oid         user_id = GetUserId();
+    char       *result = NULL;
+
+    LOCAL_FCINFO(fcinfo1, 1);
+
+    if (type_id < 0)
+        PG_RETURN_NULL();
+    
+    InitFunctionCallInfoData(*fcinfo1, NULL, 0, InvalidOid, NULL, NULL);
```
Deepesh125: nargs is zero?

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -1673,6 +1675,200 @@ object_name(PG_FUNCTION_ARGS)
...
+    }
+    // If schema is 'sys' or 'pg_catalog' then search in typecode list, else in pg_type.
+    if(!strcmp(schema_name, "sys") || !strcmp(schema_name, "pg_catalog"))
+    {
+        result = (*common_utility_plugin_ptr->get_tsql_datatype_oid) (object_name);
+    }
+    else
+    {
+        result = GetSysCacheOid2(TYPENAMENSP, Anum_pg_type_oid,
+                                    CStringGetDatum(object_name), ObjectIdGetDatum(schema_oid));
```
Deepesh125: object_name would be compared case sensitive manner. Have we tested if this is exhibiting the expected/case-insensitive behaviour? If so then add test cases

Deepesh125 (changes requested):

contrib/babelfishpg_tsql/runtime/functions.c
vaibhav412: Yes

contrib/babelfishpg_tsql/runtime/functions.c
vaibhav412: Will update

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -56,6 +56,7 @@
 
 #define TSQL_STAT_GET_ACTIVITY_COLS 26
 #define SP_DATATYPE_INFO_HELPER_COLS 23
+#define TYPE_NAME_MAX_LENGTH 4000
```
Deepesh125: SYSVARCHAR_MAX_LENGTH

contrib/babelfishpg_common/src/typecode.c
```diff
@@ -241,6 +241,18 @@ translate_pg_type_to_tsql(PG_FUNCTION_ARGS)
 	PG_RETURN_NULL();
 }
 
+Oid
+get_tsql_datatype_oid(char *type_name)
+{    
+    for (int i = 0; i < TOTAL_TYPECODE_COUNT; i++)
```
Deepesh125: Check if type_name is NULL before going into loop

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -1727,6 +1730,204 @@ object_name(PG_FUNCTION_ARGS)
...
+
+    if (PG_ARGISNULL(0))
+        PG_RETURN_NULL();
+    input = text_to_cstring(PG_GETARG_TEXT_PP(0));
+
+    /* strip trailing whitespace from input */
+    i = pg_mbstrlen(input);
+    while (i > 0 && isspace((unsigned char) input[i - 1]))
+        i--;
+    if(i < pg_mbstrlen(input))
```
Deepesh125: store len on line 1761 and reuse it here

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -1727,6 +1730,204 @@ object_name(PG_FUNCTION_ARGS)
...
+            PG_RETURN_INT32(result);
+        }
+    }
+    else
+    {
+        physical_schema_name = get_physical_schema_name(db_name, schema_name);
+    }
+
+    /* get schema oid from physical schema name, it will return InvalidOid if user don't have lookup access */
+    schema_oid = get_namespace_oid(physical_schema_name, true);
```
Deepesh125: just make sure before this line that physical_schema_name is non-null and non-empty

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -1727,6 +1730,204 @@ object_name(PG_FUNCTION_ARGS)
...
+    pfree(db_name);
+    pfree(physical_schema_name);
+
+    if (!OidIsValid(schema_oid) || pg_namespace_aclcheck(schema_oid, user_id, ACL_USAGE) != ACLCHECK_OK)
```
Deepesh125: at this point, we must find schema_oid, if we do not then there is some consistency issue and better to error out. Even Assert would make sense here.

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -1727,6 +1730,204 @@ object_name(PG_FUNCTION_ARGS)
...
+    {
+        pfree(schema_name);
+        pfree(object_name);
+        PG_RETURN_NULL();
+    }
+    // If schema is 'sys' or 'pg_catalog' then search in typecode list, else in pg_type.
+    if(!strcmp(schema_name, "sys") || !strcmp(schema_name, "pg_catalog"))
```
Deepesh125: Can we move this condition to before line 1850, we should save some execution in case of sys and pg_catalog

Deepesh125 (dismissed):

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--3.3.0--3.4.0.sql
```diff
@@ -516,6 +516,16 @@ $$
 LANGUAGE plpgsql IMMUTABLE
 STRICT;
 
+CREATE OR REPLACE FUNCTION SYS.TYPE_NAME(IN type_id INT)
```
ahmed-shameem, line 534: Let us update the input datatype to OID instead of INT. Also, add test for this check, i.e., pass negative value to this function.

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -1727,6 +1730,204 @@ object_name(PG_FUNCTION_ARGS)
...
+{
+    Datum       type_id = PG_GETARG_DATUM(0);
+    Datum       tsql_typename;
+    HeapTuple   tuple;
+    Oid         user_id = GetUserId();
+    char        *result = NULL;
+
+    LOCAL_FCINFO(fcinfo1, 1);
+
+    if (type_id < 0)
```
ahmed-shameem, line 1984: Let us take oid as input datatype instead of int. I think then this check would be redundant. After this change, what is the behaviour for negative inputs?

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -1727,6 +1730,204 @@ object_name(PG_FUNCTION_ARGS)
...
+        PG_RETURN_NULL();
+    
+    InitFunctionCallInfoData(*fcinfo1, NULL, 1, InvalidOid, NULL, NULL);
+    fcinfo1->args[0].value = type_id;
+    fcinfo1->args[0].isnull = false;
+    // Search in typecode list, it will return type name if system datatype, else will return null.
+    tsql_typename = (*common_utility_plugin_ptr->translate_pg_type_to_tsql) (fcinfo1);
+    if (tsql_typename)
+    {
+        PG_RETURN_VARCHAR_P(tsql_typename);
```
ahmed-shameem: Let us return PG_RETURN_TEXT_P instead of PG_RETURN_VARCHAR_P as we have marked the return type of the corresponding sql function as NVARCHAR. Similar signature can be found in the implementation of user_name, suser_name etc. Similarly update the return type to PG_RETURN_TEXT_P for all the c functions whose corresponding sql function's return type is NVARCHAR

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -1727,6 +1730,204 @@ object_name(PG_FUNCTION_ARGS)
...
+    Oid        result = InvalidOid;
+    int        i;
+
+    if (PG_ARGISNULL(0))
+        PG_RETURN_NULL();
+    input = text_to_cstring(PG_GETARG_TEXT_PP(0));
+
+    /* strip trailing whitespace from input */
+    i = pg_mbstrlen(input);
+    while (i > 0 && isspace((unsigned char) input[i - 1]))
```
ahmed-shameem: Can't there be spaces before the starting of the input? I mean can't we call this function like ```SELECT TYPE_ID('   type_a  ');```
If so, we need update this condition as well.

ahmed-shameem (changes requested):

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--3.3.0--3.4.0.sql
vaibhav412: Used input datatype as mentioned in [TSQL documentation](https://github.com/MicrosoftDocs/sql-docs/blob/live/docs/t-sql/functions/type-name-transact-sql.md).

contrib/babelfishpg_tsql/runtime/functions.c
vaibhav412: Spaces in the starting is not supported as per tsql behaviour.

ahmed-shameem (approved):

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -1802,6 +1805,216 @@ object_name(PG_FUNCTION_ARGS)
...
+	int        len;
+
+    if (PG_ARGISNULL(0))
+        PG_RETURN_NULL();
+    input = text_to_cstring(PG_GETARG_TEXT_PP(0));
+
+    /* strip trailing whitespace from input */
+    len = pg_mbstrlen(input);
+    i = len;
+	while (i > 0 && isspace((unsigned char) input[i - 1]))
```
Deepesh125: Should we use `scanner_isspace` in place of `isspace`?

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -1802,6 +1805,216 @@ object_name(PG_FUNCTION_ARGS)
...
+		{
+			physical_schema_name = get_physical_schema_name(db_name, schema_name);
+		}
+    }
+
+    /* get schema oid from physical schema name, it will return InvalidOid if user don't have lookup access */
+    if (physical_schema_name != NULL && pg_mbstrlen(physical_schema_name) != 0)
+		schema_oid = get_namespace_oid(physical_schema_name, true);
+	
+	pfree(schema_name);
```
Deepesh125: indentation

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -1802,6 +1805,216 @@ object_name(PG_FUNCTION_ARGS)
...
+		result = GetSysCacheOid2(TYPENAMENSP, Anum_pg_type_oid,
+										CStringGetDatum(object_name), ObjectIdGetDatum(schema_oid));
+		if (OidIsValid(result) && pg_type_aclcheck(result, user_id, ACL_USAGE) == ACLCHECK_OK)
+		{
+			pfree(object_name);
+			PG_RETURN_INT32(result);
+		}
+	}
+	pfree(object_name);
+	PG_RETURN_NULL();
```
Deepesh125: fix indentation

Deepesh125 (changes requested):

Deepesh125 (approved):

PR #1821, opened 2023-09-05
Support unquoted string parameters

robverschoor:
### Description

T-SQL allows string arguments to stored procedure calls to be specified without quotes if the string conforms to the rules for names (identifiers). This is also allowed for parameters defaults in CREATE PROCEDURE and CREATE FUNCTION.
In T-SQL, it is possible to do **EXECUTE myproc mytable**, without quotes around **mytable**. In Babelfish, **mytable** is interpreted as an identifier, so this becomes a columnRef node in the tree. 
This fix implements this functionality by intercepting parse tree nodes of type 'column' in the backend, and building a node of type 'string' instead. This uses a temporary nodetag **T_TSQL_UnquotedString** which is set/reset in **transformCallStmt()** and **interpret_function_parameter_list()**, and processed in **transformExprRecurse()**. 

Engine PR: https://github.com/babelfish-for-postgresql/postgresql_modified_for_babelfish/pull/206 

Signed-off-by: Rob Verschoor [rcv@amazon.com](mailto:rcv@amazon.com)

### Issues Resolved

BABEL-334  Support unquoted string parameter values in procedure calls/declarations
BABEL-2883 Error raised when invoking sproc w string parameter using argument in square brackets

## Test Scenarios Covered

- Note that test case babel_cursor had to be removed as it implemented incorrect functionality in the first place

- Use case based - Yes

- Boundary conditions - N/A

- Arbitrary inputs - N/A

- Negative test cases - Yes

- Minor version upgrade tests - N/A

- Major version upgrade tests - N/A

- Performance tests - N/A

- Tooling impact - N/A

- Client tests - N/A

### Check List

- [x] Commits are signed per the DCO using --signoff 

By submitting this pull request, I confirm that my contribution is under the terms of the PostgreSQL license, and grant any person obtaining a copy of the contribution permission to relicense all or a portion of my contribution to the PostgreSQL License solely to contribute all or a portion of my contribution to the PostgreSQL open source project.

For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/babelfish-for-postgresql/postgresql_modified_for_babelfish/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #1874, opened 2023-09-30
Rewriting of T-SQL double-quoted strings, and expresssions in DECLARE, SET

robverschoor:
### Description

Double-quoted strings with embedded quotes were not handled correctly. For PG, double quotes as string delimiters have to be replaced by single quotes. In the process, embedded double quotes in the original string must be un-escaped (by removing one double quote), and embedded single quotes must be escaped (by doubling each single quote).
Until now, embedded double quotes were not un-escaped (thus leaving one double quote too many) and embedded single quotes couldn't be processed at all and an error was raised. With the fix for [unquoted string arguments](https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/1821), the handling of double-quoted strings, which originally was implemented only for `sp_tables`, can now be simplified and double-quoted strings are now processed correctly by rewriting them in the listener methods of the ANTLR parser.
In the process, it was discovered that expressions being assigned to variables in a `DECLARE `statement, and assigned expressions in `SET @var =`, were not properly processed by the ANTLR query rewriting logic (and there were no test cases for these).
For `DECLARE `statements, another issue is that there can be multiple variable declarations in one statement, and consequently, multiple assigned expressions. The ANTLR query rewriting code could however handle only one expression, and if there were more, all but the last one would not be processed. This is fixed by using the context pointer for the expression, since this will be unique inside the statement (as opposed using the context pointer for the `Another_statement` rule for each assignment expression as previously).
In short, the query fragments created for these expressions (`'SELECT <expression>'`) result in different offsets than the original query at the moment when rewriting takes place, causing the rewriting to fail (in most cases, this means that the rewrite just does not happen). This PR captures the position of the expressions to be rewritten and applies these when the expression is rewritten. Additional offsets must be maintained for `SET @var =` assignments with a compound operator (e,g. `SET @v +=` ).
While making rewriting work for double-quoted strings, it became apparent that this could not be decoupled from rewriting of other aspects, like built-in functions such as `SESSION_USER` or `SYSTEM_USER`. As a result, this fix also addresses a broader category of incorrect results/errors due to expressions not being rewritten in various scenarios.

### Issues Resolved

BABEL-2442: Handle embedded double quotes in a double-quoted string
BABEL-4387: Support double-quoted strings containing single-quote
BABEL-4431: SESSION_USER/SYSTEM_USER in DECLARE variable assignment return wrong result/error
BABEL-4432: SESSION_USER/SYSTEM_USER in SET variable assignment return wrong result/error
BABEL-4435: SESSION_USER/SYSTEM_USER in PRINT return wrong result/error

* **Use case based -** yes

* **Arbitrary inputs -** N/A

* **Performance tests -** N/A

* **Tooling impact -** N/A

* **Client tests -** N/A

### Check List


test/JDBC/input/doublequoted_string-vu-prepare.sql
forestkeeper, line 1: If we don't need any sql in prepare test file, then we should put all tests into single sql file and out file.

contrib/babelfishpg_tsql/src/tsqlIface.cpp
```diff
@@ -398,6 +399,25 @@ std::pair<int,int> getLineAndPos(TerminalNode *node)
...
 static ParseTreeProperty<PLtsql_stmt *> fragments;
 
+// Keeps track of location of expressions being rewritten into a fragment 'SELECT <expr>'
+static std::map<ParseTree *,  std::pair<int, std::pair<int, int>>> selectFragmentOffsets;
+	
+// Record the offsets for a 'SELECT <expr>' fragment
+void
+recordSelectFragmentOffsetsIx(ParseTree *ctx, int ixStart, int ixEnd, int ixShift)
+{
+	assert(ctx);
```
timchang514: is `assert()` equivalent to `Assert()`? The latter is using the PG macro - not sure if this is also pointing to that or if it's using a Cpp library fn

timchang514: what is Ix - internal expression?

contrib/babelfishpg_tsql/src/tsqlIface.cpp
```diff
@@ -1814,65 +1885,125 @@ class tsqlBuilder : public tsqlCommonMutator
...
+				rewritten_query_fragment.emplace(std::make_pair(ctx->start->getStartIndex(), std::make_pair(getFullText(ctx->id()), str)));
+			}
+			else if (str.front() == '\'')
+			{
+				// This is a single-quoted string, no further action needed	
+				Assert(str.back() == '\'');		    						
+			}
+			else 
+			{
+				// This is an unquoted string parameter which has been parsed as a column name.
```
forestkeeper, line 2011: No need to keep an empty 
{}
If nothing need to be done here

robverschoor: I'd like to keep it as it documents the logic in the code.

contrib/babelfishpg_tsql/src/tsqlIface.cpp
```diff
@@ -1814,65 +1885,125 @@ class tsqlBuilder : public tsqlCommonMutator
 		}
 
 		// Declare table type statement might need rewriting in column definition list.
-		if (ctx->declare_statement() && ctx->declare_statement()->table_type_definition())
+		else if (ctx->declare_statement() && ctx->declare_statement()->table_type_definition())
```
forestkeeper, line 1892: I have a bit concern about the else in here
Since I'm not 100% sure there'll be one statement match two if condition in below.

contrib/babelfishpg_tsql/src/tsqlIface.cpp
robverschoor: - changed to Assert()
- changed recordSelectFragmentOffsetsIx() to recordSelectFragmentOffsets()  (identifier recordSelectFragmentOffsetsIx was a leftover)

contrib/babelfishpg_tsql/src/tsqlIface.cpp
robverschoor: There are many other examples of such constructs in the code already, and these work fine. For example:
- https://github.com/babelfish-for-postgresql/babelfish_extensions/blob/BABEL_3_X_DEV/contrib/babelfishpg_tsql/src/tsqlIface.cpp#L1648
- https://github.com/babelfish-for-postgresql/babelfish_extensions/blob/BABEL_3_X_DEV/contrib/babelfishpg_tsql/src/tsqlIface.cpp#L1766
- https://github.com/babelfish-for-postgresql/babelfish_extensions/blob/BABEL_3_X_DEV/contrib/babelfishpg_tsql/src/tsqlIface.cpp#L2550
- https://github.com/babelfish-for-postgresql/babelfish_extensions/blob/BABEL_3_X_DEV/contrib/babelfishpg_tsql/src/tsqlIface.cpp#L2995
- (and more)

test/JDBC/input/doublequoted_string-vu-prepare.sql
robverschoor: done

timchang514 (approved):

forestkeeper (approved):

PR #1885, opened 2023-10-05
Schema Resolution for multiple functions in single stmt

tanscorpio7:
### Description

In SQL Server the default schema inside  functions is same as functions schema, followed by the actual query default schema. We previously handled this behaviour on a statement level which worked correctly when only one schema & one function was used in the statement. When multiple functions and schemas are specified or the schema is specified with some other object along with a non schema specified function, this approach becomes ineffective, since we would resolve everything to the one schema.

To solve this we must resolve the schema (pg search path) at function execution level. We should only do this for user defined functions. All procedures & triggers will be excluded from the change since their schema resolution follows different strategy and is already implemented else where in the code. Cross DB functions call are not allowed in babelfish, so need not to look at those cases. System defined functions (sys, pg_catalog, ...) should be excluded from this change since they are expected to run in postgres environment.

The implementation uses the SET option available for PG function definitions. This SET cmd is picked up from PG catalog and locally set at run time(only for function execution). We do not store this information in the catalog itself. Instead we search the function owner schema and call the SET command during function initialisation.
But instead of adding the new search path to proconfig attribute of functions, we set it directly using the search_path global variable, to avoid performance drop.

We also store the new search path in fcache->prosearchpath to avoid multiple calls.
And we skip setting search path if language is not pltsql 

ex: Select s2.f();

s2.f() {
   -- first we search objects in s2
   -- then we search in database default schema s2.dbo
}

### Issues Resolved

[BABEL-4442]

### Check List


shalinilohia50:
Please update the PR title from "BABEL 4442" to something meaningful like "Fix incorrect schema resolution in case of multiple function calls within a single query"

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -224,6 +225,7 @@ static table_variable_satisfies_update_hook_type prev_table_variable_satisfies_u
 static table_variable_satisfies_vacuum_hook_type prev_table_variable_satisfies_vacuum = NULL;
 static table_variable_satisfies_vacuum_horizon_hook_type prev_table_variable_satisfies_vacuum_horizon = NULL;
 static drop_relation_refcnt_hook_type prev_drop_relation_refcnt_hook = NULL;
+static set_local_schema_for_func_hookType prev_set_local_schema_for_func_hook = NULL;
```
surendravishnoi: hookType -> hook_type

contrib/babelfishpg_tsql/src/pl_exec.c
```diff
@@ -10327,8 +10327,8 @@ reset_search_path(PLtsql_stmt_execsql *stmt, char **old_search_path, bool *reset
 	 * object. If not found, then search the dbo schema. Don't update the path
 	 * for "sys" schema.
 	 */
-	if ((stmt->func_call || stmt->is_create_view) && stmt->schema_name != NULL &&
-		(strcmp(stmt->schema_name, "sys") != 0 && strcmp(stmt->schema_name, "pg_catalog") != 0))
+	if (stmt->is_create_view && stmt->schema_name != NULL && (strcmp(stmt->schema_name, "sys") != 0 
+												&& strcmp(stmt->schema_name, "pg_catalog") != 0))
```
shalinilohia50: Nit: Indentation

test/JDBC/expected/BABEL_4442.out
```diff
@@ -0,0 +1,195 @@
+CREATE DATABASE db1
```
shalinilohia50: Use names like `BABEL_4442_db1`, `BABEL_4442_s1` etc. like the pattern we follow in other tests.

surendravishnoi (approved):

PR #1894, opened 2023-10-06
OID Wraparound

timchang514:
### Description

This change adds a separate OID buffer for temp objects. Since temp objects are already session-local, we can allocate a separate buffer of OIDs set aside only for temp objects. This same buffer range can be used across all sessions.

Unit tests in `babelfishpg_unit` for testing of individual functions. Standard JDBC tests for temp tables. In addition, a new JDBC module written that helps us test specific temp-table cases and inspecting OIDs of generated tables, among other things.

### Check List


contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -793,7 +793,7 @@ exec_stmt_exec(PLtsql_execstate *estate, PLtsql_stmt_exec *stmt)
 	volatile LocalTransactionId before_lxid;
 	LocalTransactionId after_lxid;
 	volatile int rc;
-	SimpleEcontextStackEntry *topEntry;
+	SimpleEcontextStackEntry *volatile topEntry;
```
surendravishnoi: why are we making all entries as volatile? are they all read/write across TRY/CATCH block

contrib/babelfishpg_tsql/src/guc.c
```diff
@@ -601,6 +601,16 @@ assign_datefirst(int newval, void *extra)
 		(*pltsql_protocol_plugin_ptr)->set_guc_stat_var("babelfishpg_tsql.datefirst", false, NULL, newval);
 }
 
+static void
+assign_temp_oid_buffer_size(int newval, void *extra)
+{
+	if (pltsql_protocol_plugin_ptr && *pltsql_protocol_plugin_ptr && (*pltsql_protocol_plugin_ptr)->set_guc_stat_var)
+	{
+		(*pltsql_protocol_plugin_ptr)->set_guc_stat_var("babelfishpg_tsql.temp_oid_buffer_size", false, NULL, newval);
```
surendravishnoi: why are we doing this?

contrib/babelfishpg_tsql/src/guc.c
```diff
@@ -601,6 +601,16 @@ assign_datefirst(int newval, void *extra)
 		(*pltsql_protocol_plugin_ptr)->set_guc_stat_var("babelfishpg_tsql.datefirst", false, NULL, newval);
```
surendravishnoi: we should add test cases, specifically around to cover corner cases

robverschoor: What is the relation of DATEFIRST with OID assignment?

contrib/babelfishpg_tsql/src/pl_exec-2.c
timchang514: They're all referenced in TRY/CATCH blocks below; I saw a similar situation as in BABEL-1292 with corrupt topEntry pointers.

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -520,7 +520,7 @@ get_logical_schema_name(const char *physical_schema_name, bool missingOk)
 	TupleDesc	dsc;
 	bool		isnull;
 
-	if (get_namespace_oid(physical_schema_name, missingOk) == InvalidOid)
+	if (!physical_schema_name || get_namespace_oid(physical_schema_name, missingOk) == InvalidOid)
```
lejaokri, line 517: Why is this change needed for the OID rework?

lejaokri: I get it now. There was a recently added call to get_logical_schema_name  by e302bcac97 that uncovered this issue. Good catch on this :)

contrib/babelfishpg_tsql/src/guc.c
```diff
@@ -1119,6 +1119,24 @@ define_custom_variables(void)
...
+							NULL,
+							&temp_oid_buffer_start,
+							InvalidOid, INT_MIN, INT_MAX, /* Add INT_MIN to size, since Oid is uint32. */
+							PGC_INTERNAL,
+							GUC_NOT_IN_SAMPLE,
+							NULL, NULL, NULL);
+
+	DefineCustomIntVariable("babelfishpg_tsql.temp_oid_buffer_size",
+							gettext_noop("Temp oid buffer size"),
+							NULL,
```
surendravishnoi, line 1166: this can be a unit32?

contrib/babelfishpg_tsql/src/guc.c
```diff
@@ -1119,6 +1119,24 @@ define_custom_variables(void)
 							   GUC_NOT_IN_SAMPLE | GUC_DISALLOW_IN_FILE | GUC_DISALLOW_IN_AUTO_FILE,
 							   check_babelfish_dump_restore_min_oid, NULL, NULL);
 
+	DefineCustomIntVariable("babelfishpg_tsql.temp_oid_buffer_start",
+							gettext_noop("Internal. Specifies the start range of the buffer for temp oids"),
+							NULL,
+							&temp_oid_buffer_start,
```
surendravishnoi, line 1155: can we have int to uint32 conversion one time as part of GUC assignment itself. Convert incoming value into equivalent unit32 value. This will simply next oid logic.

contrib/babelfishpg_tsql/src/guc.c
timchang514: is there any reason to make it uint32 as well? I can't imagine a user would ever want to set the temp buffer size to greater than INT_MAX - and currently the limit is much lower.

contrib/babelfishpg_tsql/src/guc.c
timchang514: I think it is no different from having a #define, if I'm understanding you correctly? We can also save a little bit on function overhead

.github/workflows/major-version-upgrade.yml
```diff
@@ -6,8 +6,8 @@ jobs:
     env:
       OLD_INSTALL_DIR: psql_source
       NEW_INSTALL_DIR: psql_target
-      ENGINE_BRANCH_FROM: BABEL_2_X_DEV__PG_14_X
-      EXTENSION_BRANCH_FROM: BABEL_2_X_DEV
+      ENGINE_BRANCH_FROM: oid-rework-14-x-mvu
```
surendravishnoi: this is for empty database only. Main MVU tests are driven form config while which specified source and target versions.

contrib/babelfishpg_tsql/src/guc.c
surendravishnoi: I still think it will simply the logic on engine side. Right now it is more difficult to reason (and also error prone) with INT_MIN added at different places.

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -556,11 +580,226 @@ pltsql_GetNewObjectId(VariableCache variableCache)
...
+{
+	Oid			result;
+	Oid			tempOidStart;
+	static Oid 	nextTempOid = InvalidOid;
+
+	/* safety check, we should never get this far in a HS standby */
+	if (RecoveryInProgress())
+		elog(ERROR, "cannot assign OIDs during recovery");
+
+	if (!OidIsValid(BUFFER_START_TO_OID)) /* InvalidOid means it needs assigning */
```
lejaokri, line 609: Since `temp_oid_buffer_start`  is not persisted, this means we're initializing this for every instance start (Eg: after crash, etc). So technically we could be skipping a bunch of normal OIDs during every restart, correct?

test/JDBC/src/main/java/com/sqlsamples/JDBCTempTable.java
```diff
@@ -0,0 +1,278 @@
+package com.sqlsamples;
```
lejaokri, line 1: thanks for adding this

contrib/babelfishpg_tsql/src/guc.c
```diff
@@ -1139,6 +1139,31 @@ define_custom_variables(void)
 							   GUC_NOT_IN_SAMPLE | GUC_DISALLOW_IN_FILE | GUC_DISALLOW_IN_AUTO_FILE,
 							   check_babelfish_dump_restore_min_oid, NULL, NULL);
 
+	/* 
+	 * This is an int here but actually represents an Oid (uint32). It's accessed only in varsup.c using the
```
surendravishnoi: comments are stale

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -556,11 +580,226 @@ pltsql_GetNewObjectId(VariableCache variableCache)
...
+			if (tempOidStart < FirstNormalObjectId)
+				tempOidStart = FirstNormalObjectId;
+
+			/* If the OID range would wraparound, start from beginning instead. */
+			if (tempOidStart + temp_oid_buffer_size < tempOidStart) 
+				tempOidStart = FirstNormalObjectId;
+
+			temp_oid_buffer_start = OID_TO_BUFFER_START(tempOidStart);
+			ShmemVariableCache->tempOidStart = tempOidStart;
+			ShmemVariableCache->tempOidBufferSize = temp_oid_buffer_size;
```
surendravishnoi: what is the use of this variable?

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -556,11 +580,226 @@ pltsql_GetNewObjectId(VariableCache variableCache)
...
+		}
+		else
+		{
+			/* We need to pick a new start for the buffer range. */
+			tempOidStart = ShmemVariableCache->nextOid;
+
+			/*
+			 * Decrement ShmemVariableCache->oidCount to take into account the new buffer we're allocating
+			 */
+			if (ShmemVariableCache->oidCount < temp_oid_buffer_size)
```
surendravishnoi, line 636: let us consider two cases:
1) oidCount is 0. It means that we do not have any remaining oid for current chunk. We come here, assign oids equal to buffer size and keep oid count as 0. However, we are not updating control file. If there is restart, both temp oid start and next oid will intersect. 
2) Similarly if oidCount is < buffer size, same problem could happen as we can go beyond the next oid in control file and a restart will lead to oid intersection among temp and next oid range

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -556,11 +580,226 @@ pltsql_GetNewObjectId(VariableCache variableCache)
...
+			/* Skip nextOid ahead to end of range here as well.  */
+			ShmemVariableCache->nextOid = (Oid) (tempOidStart + temp_oid_buffer_size);
+		}
+		LWLockRelease(OidGenLock);
+	}
+
+	/*
+	 * Check for wraparound of the temp OID buffer.
+	 */
+	if (nextTempOid >= (Oid) (BUFFER_START_TO_OID + temp_oid_buffer_size)
```
surendravishnoi, line 695: we should add an assert/check that temp oid range does not intersect with main oid range

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -556,11 +580,226 @@ pltsql_GetNewObjectId(VariableCache variableCache)
...
+			if (ShmemVariableCache->oidCount < temp_oid_buffer_size)
+				ShmemVariableCache->oidCount = 0;
+			else
+				ShmemVariableCache->oidCount -= temp_oid_buffer_size;
+
+			/*
+			 * If ShmemVariableCache->nextOid is below FirstNormalObjectId then we can start at FirstNormalObjectId here and
+			 * GetNewObjectId will return the right value on the next call.  
+			 */
+			if (tempOidStart < FirstNormalObjectId)
```
surendravishnoi, line 645: for regular get oid, we set oid count to 0 in case of such wraparound. However, we are not doing the same here

contrib/babelfishpg_tsql/src/hooks.c
timchang514: It will be needed when user-adjustable buffer size is added, but not necessary yet

contrib/babelfishpg_tsql/src/hooks.c
timchang514: Why would we do that here? We have no guarantee that the oidCount should be 0 at this point. That is handled in the above if statement. For instance, if we are at OID wraparound but in fact most of the OID buffer has already been freed, then oidCount will be very large.

contrib/babelfishpg_tsql/src/hooks.c
timchang514: What do you mean by control file? In the case you're giving here: 

- If oidCount is 0, then we are out of oids for the entire database. 

So first, let's say it's wraparound, ie ShmemVariableCache->nextOid = OID_MIN, since we will wrap around. If we create a temp table next, we will go through this code to pick a new tempOidStart. But tempOidStart will be pushed forward to FirstNormalObjectId by the code below. So the buffer will span from FirstNormalObjectId to FirstNormalObjectId + 65535. 

Then on line 625, we see that ShmemVariableCache->nextOid will be set to the end of the range. There will not be OID intersection. 

The other case to consider then that would be extraordinary is if ShmemVariableCache->nextOid = OID_MIN - 65535, so that we will now pick OID_MIN - 65535 as the beginning and OID_MIN as the end. That will bump ShmemVariableCache->nextOid to OID_MIN. The normal OID range will get bumped to FirstNormalObjectId by the code in GetNewObjectId, so there is no issue there as well.

timchang514: Discussed on call, we will add a write to control file for when oidCount = 0

contrib/babelfishpg_tsql/src/hooks.c
timchang514: Resolved, we now persist properly after crashes

lejaokri (dismissed):

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -75,6 +82,11 @@ extern char *babelfish_dump_restore_min_oid;
 extern bool pltsql_quoted_identifier;
 extern bool pltsql_ansi_nulls;
 
+#define OID_TO_BUFFER_START(oid) 		((oid) + INT_MIN)
+#define BUFFER_START_TO_OID 			((Oid) (temp_oid_buffer_start) - INT_MIN)
+
+/* For unit testing, to avoid concurrent heap update issues. */
+bool persist_temp_oid_buffer_start_disable_catalog_update = false;
```
surendravishnoi: please add a test prefix if this is for test only

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -563,6 +588,188 @@ pltsql_GetNewObjectId(VariableCache variableCache)
...
+
+	/* safety check, we should never get this far in a HS standby */
+	if (RecoveryInProgress())
+		elog(ERROR, "cannot assign OIDs during recovery");
+
+	/*
+	 * temp_oid_buffer_size = 0 would indicate that the feature is 
+	 * disabled, so we shouldn't even reach this code.
+	 */
+	Assert(temp_oid_buffer_size > 0);
```
surendravishnoi: we can add a fatal error instead of asserts

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -4157,6 +4364,81 @@ pltsql_validate_var_datatype_scale(const TypeName *typeName, Type typ)
...
+	char 	   *new_oid_str = NULL;
+	Oid			babelfish_db_id = InvalidOid;
+	int 		translated_oid = OID_TO_BUFFER_START(new_oid);
+	Datum		repl_val[Natts_pg_db_role_setting];
+	bool		repl_null[Natts_pg_db_role_setting];
+	bool		repl_repl[Natts_pg_db_role_setting];
+	Datum		datum;
+	ArrayType  *a;
+
+	babelfish_db_name = GetConfigOption("babelfishpg_tsql.database_name", true, false);
```
surendravishnoi, line 4392: see if there is better way to get babelfish db name rather than using GUC (cached), otherwise this is fine

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -4157,6 +4364,81 @@ pltsql_validate_var_datatype_scale(const TypeName *typeName, Type typ)
...
+	if (!babelfish_db_name)
+		return false;
+
+	babelfish_db_id = get_database_oid(babelfish_db_name, true);
```
surendravishnoi, line 4396: Please check that db id is valid

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -563,6 +588,188 @@ pltsql_GetNewObjectId(VariableCache variableCache)
...
+	{
+		/* First check to see if another connection has already picked a start, then update. */
+		LWLockAcquire(OidGenLock, LW_EXCLUSIVE);
+		if (OidIsValid(ShmemVariableCache->tempOidStart))
+		{
+			/*
+			 * Persist the newfound value of temp_oid_buffer_start to disk (via pg_settings).
+			 *
+			 * If for whatever reason this fails, we will fallback to manually updating the GUC,
+			 * which won't be crash-resilient, but won't cause loss of functionality.
```
surendravishnoi, line 619: Given that temp tables can function without persisted start oid, do we really need to persist it? Only benefit is that in case of restart we avoid one time lock contention on OID lock. It will also significantly reduces chances of temp and actual table oid range overlap.

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -563,6 +588,188 @@ pltsql_GetNewObjectId(VariableCache variableCache)
...
+			ShmemVariableCache->nextOid = (Oid) (tempOidStart + temp_oid_buffer_size);
+		}
+
+		/*
+		 * Invariant checks:
+		 * 	- We should not be in the restricted OID range
+		 *	- We should be continuous (IE the end of the buffer shouldn't wrap around to restricted OID range)
+		 * 	- We should be separate from the normal OID range
+		 */
+		Assert(BUFFER_START_TO_OID >= FirstNormalObjectId);
```
surendravishnoi: can just add errors instead of crashing the server? as long as we do not use oids, it should be OK

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -563,6 +588,188 @@ pltsql_GetNewObjectId(VariableCache variableCache)
...
+		Assert(((Oid) (BUFFER_START_TO_OID + temp_oid_buffer_size)) > BUFFER_START_TO_OID);
+		Assert(BUFFER_START_TO_OID != ShmemVariableCache->nextOid);
+
+		/* If we run out of logged for use oids then we must log more */
+		if (ShmemVariableCache->oidCount == 0)
```
surendravishnoi, line 683: this might not be required if we do not persist oid start.
also, why are we updating xlog instead of control file

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -563,6 +588,188 @@ pltsql_GetNewObjectId(VariableCache variableCache)
...
+
+		/*
+		 * Invariant checks:
+		 * 	- We should not be in the restricted OID range
+		 *	- We should be continuous (IE the end of the buffer shouldn't wrap around to restricted OID range)
+		 * 	- We should be separate from the normal OID range
+		 */
+		Assert(BUFFER_START_TO_OID >= FirstNormalObjectId);
+		Assert(((Oid) (BUFFER_START_TO_OID + temp_oid_buffer_size)) > BUFFER_START_TO_OID);
+		Assert(BUFFER_START_TO_OID != ShmemVariableCache->nextOid);
```
surendravishnoi: this is not sufficient to check if there is intersection between temp and regular table oid range. 
temp range is <start oid, temp buffer size>
regular table range is <next oid, oid count>

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -4157,6 +4364,81 @@ pltsql_validate_var_datatype_scale(const TypeName *typeName, Type typ)
...
+	SysScanDesc scan;
+	const char *babelfish_db_name = NULL;
+	char 	   *new_oid_str = NULL;
+	Oid			babelfish_db_id = InvalidOid;
+	int 		translated_oid = OID_TO_BUFFER_START(new_oid);
+	Datum		repl_val[Natts_pg_db_role_setting];
+	bool		repl_null[Natts_pg_db_role_setting];
+	bool		repl_repl[Natts_pg_db_role_setting];
+	Datum		datum;
+	ArrayType  *a;
```
surendravishnoi, line 4390: we should use arr_type or similar

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -4157,6 +4364,81 @@ pltsql_validate_var_datatype_scale(const TypeName *typeName, Type typ)
...
+	repl_repl[Anum_pg_db_role_setting_setconfig - 1] = true;
+	repl_null[Anum_pg_db_role_setting_setconfig - 1] = false;
+
+	Assert(strlen(new_oid_str) > 0);
+	datum = CStringGetTextDatum(psprintf("%s=%s", "babelfishpg_tsql.temp_oid_buffer_start", new_oid_str));
+	a = construct_array(&datum, 1,
+							TEXTOID,
+							-1, false, TYPALIGN_INT);
+
+	repl_val[Anum_pg_db_role_setting_setconfig - 1] =
```
surendravishnoi, line 4430: this is not clear. What if there are other database level settings? wouldn't it update all those setting with this single setting?

contrib/babelfishpg_unit/test_temp_tables.c
```diff
@@ -0,0 +1,308 @@
...
+        persist_temp_oid_buffer_start_disable_catalog_update = false;
+    }
+    PG_END_TRY();
+
+    TEST_ASSERT_TESTCASE(result == FirstNormalObjectId, testResult);
+    TEST_ASSERT(result == FirstNormalObjectId, testResult);
+
+    /* Check oidCount as well. */
+    LWLockAcquire(OidGenLock, LW_EXCLUSIVE);
+    TEST_ASSERT_TESTCASE(ShmemVariableCache->oidCount == 900, testResult);
```
surendravishnoi, line 47: we should also validate nextOid which should be FirstNormalObjectId + 100

contrib/babelfishpg_unit/test_temp_tables.c
```diff
@@ -0,0 +1,308 @@
...
+        persist_temp_oid_buffer_start_disable_catalog_update = true;
+        result = pltsql_GetNewTempObjectId();
+    }
+    PG_FINALLY();
+    {
+        persist_temp_oid_buffer_start_disable_catalog_update = false;
+    }
+    PG_END_TRY();
+
+    TEST_ASSERT_TESTCASE(result == 222222, testResult);
```
surendravishnoi, line 86: we should also validate that temp_oid_buffer_start is set to correct value

contrib/babelfishpg_unit/test_temp_tables.c
```diff
@@ -0,0 +1,308 @@
...
+        pltsql_GetNewTempObjectId();
+    }
+    PG_FINALLY();
+    {
+        persist_temp_oid_buffer_start_disable_catalog_update = false;
+    }
+    PG_END_TRY();
+
+    LWLockAcquire(OidGenLock, LW_EXCLUSIVE);
+    TEST_ASSERT_TESTCASE(ShmemVariableCache->oidCount == VAR_OID_PREFETCH, testResult);
```
surendravishnoi: validate nextOid as well

contrib/babelfishpg_unit/test_temp_tables.c
```diff
@@ -0,0 +1,308 @@
...
+        persist_temp_oid_buffer_start_disable_catalog_update = true;
+        result = pltsql_GetNewTempObjectId();
+    }
+    PG_FINALLY();
+    {
+        persist_temp_oid_buffer_start_disable_catalog_update = false;
+    }
+    PG_END_TRY();
+
+    TEST_ASSERT_TESTCASE(result == FirstNormalObjectId, testResult);
```
surendravishnoi, line 160: also validate oidCount, nextOid and temp_oid_buffer_start

contrib/babelfishpg_unit/test_temp_tables.c
```diff
@@ -0,0 +1,308 @@
...
+            result = pltsql_GetNewTempObjectId();
+        }
+        PG_FINALLY();
+        {
+            persist_temp_oid_buffer_start_disable_catalog_update = false;
+        }
+        PG_END_TRY();
+    }
+
+    TEST_ASSERT_TESTCASE(result == (OID_MAX - 10), testResult);
```
surendravishnoi, line 225: why is result OID_MAX - 10? if we expect errors, there is way to validate errors as well

contrib/babelfishpg_unit/test_temp_tables.c
```diff
@@ -0,0 +1,308 @@
...
+
+    LWLockAcquire(OidGenLock, LW_EXCLUSIVE);
+    TEST_ASSERT_TESTCASE(ShmemVariableCache->oidCount == VAR_OID_PREFETCH, testResult);
+    TEST_ASSERT(ShmemVariableCache->oidCount == VAR_OID_PREFETCH, testResult);
+    LWLockRelease(OidGenLock);
+
+    return testResult;
+}
+
+TestResult *test_pltsql_GetNewTempObjectId_endWraparound(void)
```
surendravishnoi, line 132: we should also add cases for
wraparound in temp range only (even though actual oid is not wrap around)
Errors due to no available temp oid
we should also call regular get oid methods validate the returned oid against nextOid value in all functions

test/JDBC/input/temp_tables/temp_tables_concurrent-vu-verify.mix
```diff
@@ -0,0 +1,35 @@
+--tsql user=u1 password=123
```
surendravishnoi, line 1: we need to add PG test cases which use table names starting with # and @ for verification

test/JDBC/src/main/java/com/sqlsamples/JDBCTempTable.java
```diff
@@ -0,0 +1,339 @@
...
+            }
+        } catch (Exception e) {
+            if (!e.getMessage().startsWith("Unable to allocate oid for temp table.")) {
+                bw.write(e.getMessage());
+                bw.newLine();
+            }
+        }
+
+        /* If the table was created, throw an error. */
+        ResultSet rs = s.executeQuery("SELECT * FROM babelfish_get_enr_list() WHERE relname = \'#table_cant_be_created\'");
```
surendravishnoi, line 195: not clear, why do expect to find this name for failed case?

test/JDBC/src/main/java/com/sqlsamples/JDBCTempTable.java
```diff
@@ -0,0 +1,339 @@
...
+        /* We need a new connection here to pick up the updated guc. */
+        Connection c2 = DriverManager.getConnection(connectionString);
+        Statement s = c2.createStatement();
+
+        try {
+            for (int i = 0; i < 11; i++) {
+                String queryString = "CREATE TABLE #tab" + i + " (a int)";
+                s.execute(queryString);
+            }
+        } catch (Exception e) {
```
surendravishnoi, line 187: test is not full proof, what if we create all 11 tables?

test/JDBC/src/main/java/com/sqlsamples/JDBCTempTable.java
```diff
@@ -0,0 +1,339 @@
...
+
+        /* Create two new connections to check the GUC value from postgres. */
+        Connection c = DriverManager.getConnection(connectionString);
+        JDBCCrossDialect cx = new JDBCCrossDialect(c);
+        Connection psql = cx.getPsqlConnection("-- psql", bw, logger);
+
+        Statement check_guc = psql.createStatement();
+        ResultSet rs = check_guc.executeQuery("SHOW babelfishpg_tsql.temp_oid_buffer_start");
+        if (!rs.next()) {
+            bw.write("Table is missing.");
```
surendravishnoi, line 251: why table is missing message?

test/JDBC/src/main/java/com/sqlsamples/JDBCTempTable.java
```diff
@@ -0,0 +1,339 @@
...
+        }
+
+        if (buffer_start_2 != buffer_start) {
+            bw.write("Oid buffer guc was different between two connections!");
+            bw.newLine();
+        }
+
+        /* Ensure that psql still can't create tables with '#'. */
+        try {
+            check_guc.execute("CREATE TABLE #psql_table_with_hash(a int)");
```
surendravishnoi, line 287: we want to create PG tables with # and @ using "" for testing

contrib/babelfishpg_tsql/src/hooks.c
timchang514: This is how it's done in GetNewObjectId.

contrib/babelfishpg_tsql/src/hooks.c
timchang514: FATAL might be too aggressive. I will leave it at ERROR, enough to stop the current transaction. I don't see a reason to kill the entire server because of this.

contrib/babelfishpg_tsql/src/hooks.c
timchang514: As mentioned before, if we don't persist then we could potentially be incrementing ShmemVariableCache->nextOid by TEMP_OID_BUFFER_SIZE each time it crashes. It would not actually consume OIDs of course, but it might induce wraparound earlier if crashes are happening consistently. Either way, making that change would be a distinct departure from the design document. I think if we want to consider a change now it's too late.

contrib/babelfishpg_tsql/src/hooks.c
timchang514: This is the correct way to do it, see AlterSetting(). We're providing the key above, so it should only alter the one setting.

contrib/babelfishpg_unit/test_temp_tables.c
timchang514: >Errors due to no available temp oid
> wraparound in temp range only

This is taken care of in JDBC tests below, we actually need to consume the OID to get this error ather than just calling the function here.

contrib/babelfishpg_unit/test_temp_tables.c
timchang514: Wraparound inside the OID buffer.

test/JDBC/input/temp_tables/temp_tables_concurrent-vu-verify.mix
timchang514: This is added below in JDBC.

test/JDBC/src/main/java/com/sqlsamples/JDBCTempTable.java
timchang514: In psql, '#' is not an allowed character in table names.

test/JDBC/src/main/java/com/sqlsamples/JDBCTempTable.java
timchang514: I guess it's a slight misnomer, but we shouldn't expect this to ever fail, since it's just displaying a GUC value for us.

lejaokri (approved):

timchang514:
Closing, this is merged in 16.x only for now.

PR #1899, opened 2023-10-09
Create DB Performance improvements

tanscorpio7:
### Description

In Multi DB mode, as the number of databases increases, so does the time to create the next new DB.
This is because we create three internal roles for each new DB and internally when run the DB subcommands, multiple calls to roles_is_member_of("sysadmin") is made. Now that output of this list contains all the three roles of every db created. This is the major reason for the perfomance degradation of CREATE DB command.

We fix this in three different places.

1. getAvailDbid - this functions makes a call to nextval function, which by default checks for current user's permission and makes a call to roles_is_member_of. Instead we could call the **nextval_internal** which is the same function but with the additional option of check permissions flag which we will set to false. To double check we can just ensure that the current user is "sysadmin" when getAvailDbid is called. (Currently we only call this when user is sysadmin)

2. Set temporary user when creating schema - when we create the dbo and guest schema for the new database, the create schema function fetches all the roles that current role is member of (recursively) to check if if current role can actually become the target schema owner role. To bypass this we can **assume the newdb_dbo role** when creating these schemas. In this case all the roles that newdb_dbo is member of will be fetched, but this list is much smaller than sysadmin.

3. Select best grantor - Select best grantor first fetches the roles_list that sysadmin is member of and then start checking for permissions. But sysadmin is always the first to be checked. That is sysadmin is always top of the roles_list. 
We can add a quick check to this. That is, first check if current role is sysadmin and can it give us all the permission needed. If yes, simply return. Note** This does not change any behaviour since this will anyway be done in the first loop after fetching roles_list. We are instead running the first loop before fetching the whole list.

4. Set newdb_dbo user when creating sysdatabases view in new db

| Create DB              | After | Before |
| DBs10        |   2ms  | 5ms |
| DBs500     |   3ms   | 163ms |
| DBs1000    |  4ms   | 740ms |
| DBs3000   |  9ms   | NA(~18secs) |

### Issues Resolved

[BABEL-3869](https://jira.rds.a2z.com/browse/BABEL-3869)

### Check List

### DIFF
--- a/contrib/babelfishpg_tsql/src/dbcmds.c
+++ b/contrib/babelfishpg_tsql/src/dbcmds.c
@@ -42,6 +42,9 @@
 #include "pltsql.h"
 #include "extendedproperty.h"
 
+Oid sys_babelfish_db_seq_oid = InvalidOid;
+
+static Oid get_sys_babelfish_db_seq_oid(void);
 static bool have_createdb_privilege(void);
 static List *gen_createdb_subcmds(const char *schema,
 								  const char *dbo,
@@ -56,6 +59,20 @@ static List *gen_dropdb_subcmds(const char *schema,
 static Oid	do_create_bbf_db(const char *dbname, List *options, const char *owner);
 static void create_bbf_db_internal(const char *dbname, List *options, const char *owner, int16 dbid);
 static void drop_related_bbf_namespace_entries(int16 dbid);
+
+static Oid
+get_sys_babelfish_db_seq_oid()
+{
+	if(!OidIsValid(sys_babelfish_db_seq_oid))
+	{
+		RangeVar	*sequence = makeRangeVarFromNameList(stringToQualifiedNameList("sys.babelfish_db_seq"));
+		Oid			seqid = RangeVarGetRelid(sequence, NoLock, false);
+		
+		Assert(OidIsValid(seqid));
+		sys_babelfish_db_seq_oid = seqid;
+	}
+	return sys_babelfish_db_seq_oid;
+}
 
 static bool
 have_createdb_privilege(void)
@@ -108,11 +125,6 @@ gen_createdb_subcmds(const char *schema, const char *dbo, const char *db_owner, 
 
 	appendStringInfo(&query, "CREATE SCHEMA dummy AUTHORIZATION dummy; ");
 
-	/* create sysdatabases under current DB's DBO schema */
-	appendStringInfo(&query, "CREATE VIEW dummy.sysdatabases AS SELECT * FROM sys.sysdatabases; ");
-	appendStringInfo(&query, "ALTER VIEW dummy.sysdatabases OWNER TO dummy; ");
-	appendStringInfo(&query, "GRANT SELECT ON dummy.sysdatabases TO dummy; ");
-
 	/* create guest schema in the database. This has to be the last statement */
 	if (guest)
 		appendStringInfo(&query, "CREATE SCHEMA dummy AUTHORIZATION dummy; ");
@@ -120,9 +132,9 @@ gen_createdb_subcmds(const char *schema, const char *dbo, const char *db_owner, 
 	res = raw_parser(query.data, RAW_PARSE_DEFAULT);
 
 	if (guest)
-		expected_stmt_num = list_length(logins) > 0 ? 10 : 9;
+		expected_stmt_num = list_length(logins) > 0 ? 7 : 6;
 	else
-		expected_stmt_num = 7;
+		expected_stmt_num = 6;
 
 	if (list_length(res) != expected_stmt_num)
 		ereport(ERROR,
@@ -159,15 +171,6 @@ gen_createdb_subcmds(const char *schema, const char *dbo, const char *db_owner, 
 
 	stmt = parsetree_nth_stmt(res, i++);
 	update_CreateSchemaStmt(stmt, schema, db_owner);
-
-	stmt = parsetree_nth_stmt(res, i++);
-	update_ViewStmt(stmt, schema);
-
-	stmt = parsetree_nth_stmt(res, i++);
-	update_AlterTableStmt(stmt, schema, db_owner);
-
-	stmt = parsetree_nth_stmt(res, i++);
-	update_GrantStmt(stmt, NULL, schema, db_owner);
 
 	if (guest)
 	{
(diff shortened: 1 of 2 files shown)


shalinilohia50:
Could you please mention how much improvement we have observed after these changes?

contrib/babelfishpg_tsql/src/dbcmds.c
```diff
@@ -301,9 +301,15 @@ getAvailDbid(void)
...
+	if(GetUserId() != get_role_oid("sysadmin", true))
+		return InvalidDbid;
+
 	do
 	{
-		dbid = DirectFunctionCall1(nextval, CStringGetTextDatum("sys.babelfish_db_seq"));
+		RangeVar	*sequence = makeRangeVarFromNameList(stringToQualifiedNameList("sys.babelfish_db_seq"));
+		Oid			seqid = RangeVarGetRelid(sequence, NoLock, false);
+
+		dbid = nextval_internal(seqid, false);
```
Deepesh125: just to be safe, add assert before calling nextval_internal - `Assert(OidIsValid(seqid))`

contrib/babelfishpg_tsql/src/dbcmds.c
```diff
@@ -108,21 +125,16 @@ gen_createdb_subcmds(const char *schema, const char *dbo, const char *db_owner,
...
 		appendStringInfo(&query, "CREATE SCHEMA dummy AUTHORIZATION dummy; ");
 
 	res = raw_parser(query.data, RAW_PARSE_DEFAULT);
 
 	if (guest)
-		expected_stmt_num = list_length(logins) > 0 ? 10 : 9;
+		expected_stmt_num = list_length(logins) > 0 ? 7 : 6;
 	else
-		expected_stmt_num = 7;
+		expected_stmt_num = 6;
```
HarshLunagariya, line 141: Shouldn't it be 4?

contrib/babelfishpg_tsql/src/dbcmds.c
```diff
@@ -545,7 +557,9 @@ create_bbf_db_internal(const char *dbname, List *options, const char *owner, int
 						   None_Receiver,
 						   NULL);
 
-			/* make sure later steps can see the object created here */
```
HarshLunagariya, line 548: Unwanted change

HarshLunagariya (changes requested):

contrib/babelfishpg_tsql/src/dbcmds.c
shalinilohia50: Yes, it should be 4.

contrib/babelfishpg_tsql/src/dbcmds.c
```diff
@@ -108,21 +125,16 @@ gen_createdb_subcmds(const char *schema, const char *dbo, const char *db_owner,
 
 	appendStringInfo(&query, "CREATE SCHEMA dummy AUTHORIZATION dummy; ");
 
-	/* create sysdatabases under current DB's DBO schema */
-	appendStringInfo(&query, "CREATE VIEW dummy.sysdatabases AS SELECT * FROM sys.sysdatabases; ");
-	appendStringInfo(&query, "ALTER VIEW dummy.sysdatabases OWNER TO dummy; ");
-	appendStringInfo(&query, "GRANT SELECT ON dummy.sysdatabases TO dummy; ");
-
```
HarshLunagariya: We shouldn't do it. It may break existing applications and tooling support.
Refer - https://github.com/MicrosoftDocs/sql-docs/blob/live/docs/relational-databases/system-compatibility-views/sys-sysdatabases-transact-sql.md

contrib/babelfishpg_tsql/src/dbcmds.c
```diff
@@ -108,21 +125,16 @@ gen_createdb_subcmds(const char *schema, const char *dbo, const char *db_owner,
 
 	appendStringInfo(&query, "CREATE SCHEMA dummy AUTHORIZATION dummy; ");
 
-	/* create sysdatabases under current DB's DBO schema */
-	appendStringInfo(&query, "CREATE VIEW dummy.sysdatabases AS SELECT * FROM sys.sysdatabases; ");
```
shalinilohia50, line 112: Since we are removing this create view statement from here, we should create this view `sysdatabases` in `ownership.sql`.

shalinilohia50 (changes requested):

contrib/babelfishpg_tsql/src/dbcmds.c
tanscorpio7: Select * from sys.sysdatabases
Select * from sysdatabases
stiil work, post this change

contrib/babelfishpg_tsql/src/dbcmds.c
tanscorpio7: seems we already create it in ownership

shalinilohia50: Right.
We need to resolve `dbo.sysdatabases` to `sys.sysdatabases` as discussed.

contrib/babelfishpg_tsql/src/dbcmds.c
tanscorpio7: Not removing the create view stmts for dbo.sysdatabases
Instead just setting the current_user to dbo for create and alter view stmts, similar to create schema stmts
cc: @shalinilohia50

contrib/babelfishpg_tsql/src/dbcmds.c
tanscorpio7: This approach requires more planning, dropping it for now

shalinilohia50 (approved):

contrib/babelfishpg_tsql/src/dbcmds.c
shalinilohia50: I see the performance numbers in the description:
```
This removes almost all the increase seen in create db execution time.
DBs10 created in 2 ms
DBs500 created in 3 ms
DBs1000 created in 4 ms
DBs3000 created in 9 ms
(local stats)
```
Are these the latest numbers after all 4 fixes done in this PR(including the engine change)?

tanscorpio7: These numbers can ONLY be seen with all 4 changes, I missed fix number 4 in earlier revisions.
Also numbers are for local dev dsk

forestkeeper (approved):

PR #1900, opened 2023-10-09
Redoing unquoted string handling

robverschoor:
### Description

Re-implements unquoted string handling, as the previous fix (https://github.com/babelfish-for-postgresql/postgresql_modified_for_babelfish/pull/206 , https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/1821) did not address all use cases, notably strings longer than 63 characters; also all unquoted strings were downcased. 
This fix rolls back the previous code changes in the backend and implements the the unquoted string handling fully in the Babelfish extension (`tsqlIface.cpp`), restructuring various parts of ANTLR tree rewriting in the process, and adding many test cases.

Engine PR: https://github.com/babelfish-for-postgresql/postgresql_modified_for_babelfish/pull/235

### Issues Resolved

BABEL-334 Support unquoted string parameter values in procedure calls/declarations
BABEL-2883 Error raised when invoking sproc w string parameter using argument in square brackets
BABEL-4452 Incorrect results with long unquoted string parameter values in procedure calls/declarations
 
## Test Scenarios Covered

- Use case based - Yes

- Boundary conditions - Yes

- Arbitrary inputs - N/A

- Negative test cases - Yes

- Performance tests - N/A

- Tooling impact - N/A

- Client tests - N/A

### Check List


contrib/babelfishpg_tsql/src/tsqlIface.cpp
```diff
@@ -996,16 +1004,31 @@ class tsqlCommonMutator : public TSqlParserBaseListener
...
 	void exitFunc_proc_name_server_database_schema(TSqlParser::Func_proc_name_server_database_schemaContext *ctx) override
 	{
 		GetCtxFunc<TSqlParser::Func_proc_name_server_database_schemaContext *> getDatabase = [](TSqlParser::Func_proc_name_server_database_schemaContext *o) { return o->database; };
 		GetCtxFunc<TSqlParser::Func_proc_name_server_database_schemaContext *> getSchema = [](TSqlParser::Func_proc_name_server_database_schemaContext *o) { return o->schema; };
 		std::string rewritten_name = rewrite_object_name_with_omitted_db_and_schema_name(ctx, getDatabase, getSchema);
 		std::string rewritten_schema_name = rewrite_information_schema_to_information_schema_tsql(ctx, getSchema);
 		if (!rewritten_name.empty())
-			rewritten_query_fragment.emplace(std::make_pair(ctx->start->getStartIndex(), std::make_pair(::getFullText(ctx), rewritten_name)));
+		{
+			int EXEC_prepend_length = (in_execute_body_batch) ? fragment_EXEC_prefix.length() : 0;  // add length of prefix prepended internally for execute_body_batch
```
forestkeeper, line 1024: What is a exec_preend_length ? Would you mind give some explain?

contrib/babelfishpg_tsql/src/tsqlIface.cpp
```diff
@@ -1506,7 +1529,7 @@ class tsqlBuilder : public tsqlCommonMutator
 	void exitCreate_or_alter_procedure(TSqlParser::Create_or_alter_procedureContext *ctx) override
 	{
 		// just throw away all PLtsql_stmt in the container.
-		// procedure body is a sequenece of sql_clauses. In create-procedure, we don't need to genereate PLtsql_stmt.
```
forestkeeper, line 1509: Thanks for fixing those typos

contrib/babelfishpg_tsql/src/tsqlIface.cpp
```diff
@@ -1954,40 +1977,43 @@ class tsqlBuilder : public tsqlCommonMutator
 
 	void exitChar_string(TSqlParser::Char_stringContext *ctx) override
 	{
-		if (ctx->STRING())
-		{
-			std::string str = ctx->STRING()->getSymbol()->getText();
+		std::string str = getFullText(ctx);
```
forestkeeper, line 2025: Would you point out in which condition it'll be something different other than ctx->STRING() ?

contrib/babelfishpg_tsql/src/tsqlIface.cpp
```diff
@@ -1954,40 +1977,43 @@ class tsqlBuilder : public tsqlCommonMutator
...
 
-			if (str.front() == 'N')
-			{
-				// This is only to make the assertion on str.front() easy (below)
-				str.erase(0, 1);	
-			}
+		if (str.front() == 'N')
+		{
+			// This is only to make the assertion on str.front() easy (below)
+			str.erase(0, 1);
```
forestkeeper, line 2030: I just remembered for pure char part erase or replace, if the length of replaced char didn't exceed the original char, then we can do that in enterXXX implements, not in exitXXX implements. Like enterConstant function.

contrib/babelfishpg_tsql/src/tsqlIface.cpp
```diff
@@ -1954,40 +1977,43 @@ class tsqlBuilder : public tsqlCommonMutator
...
+			str = rewriteDoubleQuotedString(str);	
+			size_t startPosition = ctx->start->getStartIndex();
+			if (in_execute_body_batch_parameter) startPosition += fragment_EXEC_prefix.length(); // add length of prefix prepended internally for execute_body_batch  
+			rewritten_query_fragment.emplace(std::make_pair(startPosition, std::make_pair(::getFullText(ctx), str)));
+		}
+		else
+		{
+			// This is a single-quoted string, no further action needed
+			Assert(str.front() == '\'');
+			Assert(str.back() == '\'');
```
forestkeeper, line 2047: We should move this part to enterChar_string() too.

contrib/babelfishpg_tsql/src/tsqlIface.cpp
robverschoor: EXEC_prepend_length is fragment_EXEC_prefix.length() (as assigned above), which is the length of the string "EXEC ". This is prepended in case of Execute_body_batch, i.e. execution of a stored proc as the first thing in a batch, so without an EXEC[UTE] keyword. We need to prepend EXEC but need to compensate for the additional length thus inserted or the rewriting will not work.  The comment at the end of the line basically says this.

contrib/babelfishpg_tsql/src/tsqlIface.cpp
robverschoor: The test for STRING was redundant given the grammar rule. It cannot be anything other than STRING:
```
char_string
    : STRING // single-quoted string; may also be a double-quoted string in case of SET QUOTED_IDENTIFIER OFF
    ;
```

contrib/babelfishpg_tsql/src/tsqlIface.cpp
robverschoor: This is more of a sanity check but actually it does not add any value given the grammar rules.

contrib/babelfishpg_tsql/src/tsqlIface.cpp
robverschoor: We are not actually erasing anything here. This only removes the leading 'N' locally so that the subsequent test is easier -- as the comment on line 1984 indicates (will update the comment to clarify more).

test/JDBC/expected/unquoted_string-vu-verify.out
```diff
@@ -244,29 +392,263 @@ go
...
 go
 ~~START~~
 varchar
-abc
+aBc
+~~END~~
+
+
+-- strings longer than 63 should not be truncated, hashed or downcased
+create procedure unqStr_proc_2a(
```
2jungkook: Usually, we put `CREATE` statements in `prepare` file instead of `verify` file. Is there any reason for not putting these statements in `prepare` file?

test/JDBC/expected/BABEL-2844.out
```diff
@@ -235,7 +235,7 @@ execute babel_2844_proc 3;
 go
 ~~START~~
 text
-Query Text: EXEC babel_2844_proc 3
+Query Text: EXEC    babel_2844_proc 3
```
2jungkook, line 238: Seems like this is a side effect. Do you have an explanation for this?

contrib/babelfishpg_tsql/src/tsqlIface.cpp
```diff
@@ -7043,3 +7256,27 @@ std::string rewriteDoubleQuotedString(const std::string strDoubleQuoted)
...
 }
+
+// Escape double quotes by doubling them
+std::string 
+escapeDoubleQuotes(const std::string strWithDoubleQuote)
+{
+	std::string quote = "\"";
+	std::string str = strWithDoubleQuote;
+
+	if (str.find(quote) == std::string::npos)
```
2jungkook: `find()` is called twice:
- `str.find(quote)`
- `str.find(quote, 1)`

Can we just start from the below for-loop?

contrib/babelfishpg_tsql/src/tsqlIface.cpp
```diff
@@ -1954,64 +1977,85 @@ class tsqlBuilder : public tsqlCommonMutator
...
+			{
+				// When it's a bracketed identifier, remove the delimiters as T-SQL treats it as a string								
+				Assert(str.back() == ']');
+
+				// Temporarily turn this into a double-quoted string so that we can handle embedded quotes.
+				// Since embedded double quotes inside a bracketed identifier are not escaped (as they would be in a 
+				// double-quoted string), escape them first
+				str = escapeDoubleQuotes(str);
+				str = '"' + str.substr(1,str.length()-2) + '"';
+				str = rewriteDoubleQuotedString(str);
```
2jungkook: It looks like `escapeDoubleQuotes()` adds `\"` and `rewriteDoubleQuotedString()` removes `\"`. Could you explain why do we need to add and then remove it? A concrete example would be helpful as well.

contrib/babelfishpg_tsql/src/tsqlIface.cpp
```diff
@@ -772,6 +779,7 @@ class tsqlCommonMutator : public TSqlParserBaseListener
 public:
 	explicit tsqlCommonMutator() = default;
 	bool in_create_or_alter_function = false;
+	//bool in_execute_body_batch = false;
```
2jungkook: Let's remove it instead of commenting out.

test/JDBC/expected/unquoted_string-vu-verify.out
robverschoor: There is conflicting guidance on this topic. Anyway, the reason for putting the CREATE statements in the VERIFY file is that it is very difficult to validate whether the results are correct when splitting it into two files. This is because the many testcases have subtle differences only. Whenever a test case might fail in the future, it will be much more difficult to determine exactly what is going on.

test/JDBC/expected/BABEL-2844.out
robverschoor: This is because of https://github.com/babelfish-for-postgresql/babelfish_extensions/blob/c9c7a4b2012afe7492267faf0b0c1e405e80569f/contrib/babelfishpg_tsql/src/tsqlIface.cpp#L5531 : Because the original keyword was EXECUTE, the EXEC keyword is padded with three space to keep the length , and therewith, the positions of subsequent tokens, the same. This is required for the subsequent rewriting steps.
Apparently the code in this test case prints out the intermediate string passed into raw_parser()).

contrib/babelfishpg_tsql/src/tsqlIface.cpp
robverschoor: Please see the comment preceding these lines. An example is `EXECUTE myproc [a"b]`

contrib/babelfishpg_tsql/src/tsqlIface.cpp
robverschoor: done

contrib/babelfishpg_tsql/src/tsqlIface.cpp
robverschoor: done

test/JDBC/expected/unquoted_string-vu-verify.out
2jungkook: The reason for putting `CREATE` in `prepare` is to test MVU cases. When we test MVU, we execute `prepare` file in an old version, and then execute `verify` after MVU. In this way, we can guarantee view/proc/func created in an older version works fine in a newer version.

I understand this makes developers' life a little bit more uncomfortable, but it has a clear benefit of having wider test coverage.

contrib/babelfishpg_tsql/src/tsqlIface.cpp
2jungkook: What's going wrong if we do only
```
str = '"' + str.substr(1,str.length()-2) + '"';
str = rewriteDoubleQuotedString(str);
```
for `EXECUTE myproc [a"b]`?

robverschoor: When we have two adjacent embedded double quotes, your proposed change would remove one of them. 
I.e.   [a""b]   would then become [a"b] so we'd lose one quote.

test/JDBC/expected/unquoted_string-vu-verify.out
robverschoor: OK, will split up

forestkeeper (approved):

2jungkook (approved):

Issue #1317, opened 2023-03-04
[Bug]: sp_describe_undeclared_parameters is slow


zekiguven:
### What happened?

sp_describe_undeclared_parameters is very slow, 50sec waiting. 
Tested with 2.3.0. 
Works good with 2.1.2

sample code:
```
SET ANSI_NULLS ON
GO

SET QUOTED_IDENTIFIER ON
GO

CREATE TABLE [dbo].[TESTTBL](
	[PKREF] [int] IDENTITY(1,1) NOT NULL,
	[F1] [smallint] NULL,
	[F2] [smallint] NULL,
	[F3] [varchar](25) NULL,
	[F4] [varchar](51) NULL,
	[F5] [varchar](11) NULL,
	[F6] [varchar](11) NULL,
	[F7] [float] NULL,
	[F8] [int] NULL,
	[F9] [int] NULL,
	[F10] [int] NULL,
	[F11] [smallint] NULL,
	[F12] [datetime] NULL,
	[F13] [smallint] NULL,
	[F14] [smallint] NULL,
	[F15] [smallint] NULL,
	[F16] [smallint] NULL,
	[F17] [datetime] NULL,
	[F18] [smallint] NULL,
	[F19] [smallint] NULL,
	[F20] [smallint] NULL,
	[F21] [smallint] NULL,
	[F22] [smallint] NULL,
	[F23] [int] NULL,
	[F24] [int] NULL,
	[F25] [float] NULL,
	[F26] [smallint] NULL,
	[F27] [smallint] NULL,
	[F28] [int] NULL,
	[F29] [smallint] NULL,
	[F30] [int] NULL,
	[F31] [smallint] NULL,
	[F32] [smallint] NULL,
	[F33] [smallint] NULL,
	[F34] [smallint] NULL,
	[F35] [smallint] NULL,
	[F36] [smallint] NULL,
	[F37] [varchar](11) NULL,
	[F38] [int] NULL,
	[F39] [varchar](201) NULL,
	[F40] [int] NULL,
	[F41] [int] NULL,
	[F42] [int] NULL,
	[F43] [int] NULL,
	[F44] [int] NULL,
	[F45] [int] NULL,
	[F46] [int] NULL,
	[F47] [int] NULL,
	[F48] [int] NULL,
	[F49] [int] NULL,
	[F50] [int] NULL,
	[F51] [varchar](11) NULL,
	[F52] [varchar](11) NULL,
[...]

PR #1905, opened 2023-10-10
Improve performance of sp_describe_undeclared_parameters

ahmed-shameem:
### Description
sp_describe_undeclared_parameters is slow. The issue is reported here: https://github.com/babelfish-for-postgresql/babelfish_extensions/issues/1317. 

Initially, we used to create a query  in C for every parameter passed as an argument to sp_describe_undeclared_parameters and then execute it. This was taking huge amount of time as it used JOINS among multiple views like sys.objects, sys.columns, sys.types T. In this commit, we have replaced the use of such views as much as possible with pg catalogs. This aids to the improvement in performance.

The changes are tested locally.

Performance improvements (All the times are in ms):
-------------- INSERT --------------
| No of Params | Before | After |
|--------------|-------|--------|
| 1            | 1109  | 298    |
| 2            | 1619  | 39     |
| 8            | 6061  | 562    |
| 10           | 7600  | 573    |
| 20           | 15498 | 967    |
| 30           | 23112 | 1170   |
| 40           | 31076 | 1563   |
| 50           | 39429 | 1745   |
| 60           | 49043 | 2628   |
| 70           | 55999 | 3318   |
| 80           | 65040 | 4381   |
| 90           | 73428 | 4978   |
| 100          | 83244 | 6072   |

```Comparison on INSERT```
Comparison on INSERT

-------------- UPDATE --------------
| No of Params | Before | After |
|--------------|-------|--------|
| 1            | 1681  | 177    |
| 2            | 2323  | 55     |
| 8            | 6919  | 556    |
| 10           | 8406  | 594    |
| 20           | 16280 | 976    |
| 30           | 23911 | 1195   |
| 40           | 31936 | 1581   |
| 50           | 40249 | 1765   |
| 60           | 49052 | 2606   |
| 70           | 57128 | 3318   |
| 80           | 65178 | 4365   |
| 90           | 73632 | 5028   |
| 100          | 83183 | 6102   |

```Comparison on UPDATE```
Comparison on UPDATE

-------------- DELETE --------------
| No of Params | Before | After |
|--------------|-------|--------|
| 1            | 803   | 71     |
| 2            | 1507  | 37     |
| 8            | 6135  | 538    |
| 10           | 7710  | 575    |
| 20           | 15561 | 975    |
| 30           | 23361 | 1173   |
| 40           | 31125 | 1570   |
| 50           | 39824 | 1772   |
| 60           | 48933 | 2620   |
| 70           | 57075 | 3362   |
| 80           | 65444 | 4481   |
| 90           | 73762 | 5127   |
| 100          | 83991 | 6248   |

```Comparison on DELETE```
Comparison on DELETE

```Approx Linear increment in execution time based on parameter size```
Linear increment in execution time based on parameter size

QUERY PLAN AFTER MODIFICATION WITH EXPLICIT JOINS AND REMOVING UNNECESSARY CHECK ON t2.NAME:
1. INSERT: https://explain.dalibo.com/plan/dh97223b15e92h0d
2. UPDATE: https://explain.dalibo.com/plan/d28ec392d560b25c
3. DELETE: https://explain.dalibo.com/plan/348269ffgb6ae9g9

Task: BABEL-3705
Signed-off-by: Shameem Ahmed <shmeeh@amazon.com>

* **Use case based -**

* **Boundary conditions -**

* **Arbitrary inputs -**

* **Negative test cases -**

* **Performance tests -**

* **Tooling impact -**

* **Client tests -**

### Check List

### CODE REVIEW

File: contrib/babelfishpg_tsql/src/procedures.c [hunk 1]
```diff
@@ -1283,146 +1287,299 @@ sp_describe_undeclared_parameters_internal(PG_FUNCTION_ARGS)
...
+		", CASE T2.name COLLATE sys.database_default "
+		"WHEN \'image\' THEN -1 "
+		"WHEN \'ntext\' THEN -1 "
+		"WHEN \'text\' THEN -1 "
+		"ELSE CASE "
+				"WHEN a.atttypmod != -1 THEN "
+					"sys.tsql_type_max_length_helper(coalesce(tsql_type_name, tsql_base_type_name), a.attlen, a.atttypmod) "
+				"ELSE "
+					"sys.tsql_type_max_length_helper(coalesce(tsql_type_name, tsql_base_type_name), a.attlen, t.typtypmod) "
+			"END  "
```
sumitj824, line 1448: sys.types view already contains max_length, precision and scale columns. Can we check if can reuse those instead of calculating it again and again.

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -1283,146 +1287,299 @@ sp_describe_undeclared_parameters_internal(PG_FUNCTION_ARGS)
...
-		"AND O.schema_id = %d " /* -- INPUT schema Oid */
-		"AND O.type = \'U\'";	/* -- User tables only for the time being */
-
-		char	   *query = psprintf(tempq,
-									 undeclaredparams->targetcolnames[undeclaredparams->paramindexes[call_cntr]],
-									 undeclaredparams->tablename,
-									 undeclaredparams->schemaoid);
-
+		"FROM sys.types T2, pg_attribute a, pg_type t "
+		", sys.translate_pg_type_to_tsql(a.atttypid) AS tsql_type_name "
```
sumitj824: `name` column in sys.types already gives you `tsql_type_name`, right?

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -1283,146 +1287,299 @@ sp_describe_undeclared_parameters_internal(PG_FUNCTION_ARGS)
...
+		", sys.translate_pg_type_to_tsql(t.typbasetype) AS tsql_base_type_name "
+        "WHERE a.attrelid = %d "
+        "AND a.atttypid = T2.user_type_id " 
+        "AND T2.system_type_id = T2.user_type_id "
```
sumitj824, line 1581: By this, Are we restricting the output to system types only?

sumitj824 (changes requested):

contrib/babelfishpg_tsql/src/procedures.c
ahmed-shameem: No we can not because initially we were taking max_lenght, precision and scale from sys.columns. Now, the definition of these parameters are different in sys.columns and sys.types, which is why we have modified the query this way. Initially, I used these parameters from sys.types only, but got test failrues. Then I notice the difference in definition.
We have 2 options now:
1. Use these parameters from sys.types (and make sure the results are as expected or not, if yes we have to modify the expected output file). But I suspect the values won't be correct.
2. Update the definition of sys.types for these parameters if that is acceptable.

contrib/babelfishpg_tsql/src/procedures.c
ahmed-shameem: Yes, currently our implementation of this sp is limited to system types only as this sp is not fully implemented yet. User defined types produces unexpected output. Which is not addressed here as this task only focuses on improving the perf. For the correctness of output, we have other tasks.

contrib/babelfishpg_tsql/src/procedures.c
ahmed-shameem: Updated in latest revision.

Deepesh125:
Change the title of PR to indicate that the performance improvement being done

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -1283,146 +1287,298 @@ sp_describe_undeclared_parameters_internal(PG_FUNCTION_ARGS)
...
-		"WHEN T2.name COLLATE sys.database_default = \'datetime2\' THEN \'datetime2(\' + CAST( C.scale AS sys.VARCHAR(10) ) + \')\' "
-		"WHEN T2.name COLLATE sys.database_default = \'datetimeoffset\' THEN \'datetimeoffset(\' + CAST( C.scale AS sys.VARCHAR(10) ) + \')\' "
-		"WHEN T2.name COLLATE sys.database_default = \'time\' THEN \'time(\' + CAST( C.scale AS sys.VARCHAR(10) ) + \')\' "
-		"WHEN T2.name COLLATE sys.database_default = \'varchar\' THEN "
-		"CASE WHEN C.max_length = -1 THEN \'varchar(max)\' "
-		"ELSE \'varchar(\' + CAST( C.max_length AS sys.VARCHAR(10) ) + \')\' "
+		", CASE T2.name COLLATE sys.database_default "
+		"WHEN \'decimal\' THEN \'decimal(\' + CAST( CASE "
+				"WHEN a.atttypmod != -1 THEN "
+					"sys.tsql_type_precision_helper(coalesce(T2.name, tsql_base_type_name), a.atttypmod) "
```
KushaalShroff: can we rather call coalesce(T2.name, tsql_base_type_name) in the join statement and then use the alias for that here?

contrib/babelfishpg_tsql/src/procedures.c
KushaalShroff: Are we tracking this limitation?

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -1283,146 +1287,298 @@ sp_describe_undeclared_parameters_internal(PG_FUNCTION_ARGS)
...
-		"AND O.name = \'%s\' COLLATE sys.database_default " /* -- INPUT table name */
-		"AND O.schema_id = %d " /* -- INPUT schema Oid */
-		"AND O.type = \'U\'";	/* -- User tables only for the time being */
-
-		char	   *query = psprintf(tempq,
-									 undeclaredparams->targetcolnames[undeclaredparams->paramindexes[call_cntr]],
-									 undeclaredparams->tablename,
-									 undeclaredparams->schemaoid);
-
+		"FROM sys.types T2, pg_attribute a, pg_type t "
```
KushaalShroff: We need to join system_type_id of T2 with pg_type's oid otherwise rows will be jumbled and tsql_base_type_name would not always be the same corresponding type

KushaalShroff (changes requested):

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -1283,146 +1287,298 @@ sp_describe_undeclared_parameters_internal(PG_FUNCTION_ARGS)
...
+		"WHEN \'smallmoney\' THEN 122 "
+		"WHEN \'text\' THEN 35 "
+		"WHEN \'time\' THEN 41 "
+		"WHEN \'tinyint\' THEN 48 "
+		"WHEN \'uniqueidentifier\' THEN 36 "
+		"WHEN \'varbinary\' THEN 165 "
+		"WHEN \'varchar\' THEN 167 "
+		"WHEN \'xml\' THEN 241 "
+		"ELSE CASE "
+				"WHEN T2.name IS NOT NULL OR t.typbasetype = 0 THEN "
```
Deepesh125: T2.name (sys.types.name) will never be NULL, why this check then?

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -1283,146 +1287,298 @@ sp_describe_undeclared_parameters_internal(PG_FUNCTION_ARGS)
...
-		"ELSE C.xml_collection_id "
-		"END "					/* AS "suggested_xml_collection_id" */
+		", CAST( NULL AS INT )  "				/* AS "suggested_xml_collection_id" */
 		", CAST( NULL AS sysname ) "	/* AS
 										 * "suggested_xml_collection_database" */
 		", CAST( NULL AS sysname ) "	/* AS
 										 * "suggested_xml_collection_schema" */
 		", CAST( NULL AS sysname ) "	/* AS "suggested_xml_collection_name" */
-		", C.is_xml_document "	/* AS "suggested_is_xml_document" */
+		", CAST(0 AS sys.bit) "	/* AS "suggested_is_xml_document" */
```
Deepesh125, line 1478: why?

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -1283,146 +1287,298 @@ sp_describe_undeclared_parameters_internal(PG_FUNCTION_ARGS)
...
-		", CASE WHEN T.user_type_id = T.system_type_id THEN CAST( NULL AS sysname) ELSE T.name END "	/* AS
+		", CASE WHEN T2.user_type_id = T2.system_type_id THEN CAST( NULL AS sysname) ELSE T2.name END "	/* AS
 																										 * "suggested_user_type_name" */
 		", CAST( NULL AS NVARCHAR(4000) ) " /* AS
 											 * "suggested_assembly_qualified_type_name" */
-		", CASE "
-		"WHEN C.xml_collection_id = 0 THEN CAST( NULL AS INT ) "
-		"ELSE C.xml_collection_id "
-		"END "					/* AS "suggested_xml_collection_id" */
+		", CAST( NULL AS INT )  "				/* AS "suggested_xml_collection_id" */
```
Deepesh125, line 1472: why?

Deepesh125 (changes requested):

contrib/babelfishpg_tsql/src/procedures.c
ahmed-shameem: sys.types.name is obtained from translate_pg_type_to_tsql function. If invalid oid is provided, then the function will return NULL. To handle that case, this check is kept. Also, previously we had this check as well.

contrib/babelfishpg_tsql/src/procedures.c
ahmed-shameem: is_xml_document column is hardcoded as 0 in sys.columns. If we have any way to obtain this value in future, then we have to update this definition as same as sys.columns_internal

contrib/babelfishpg_tsql/src/procedures.c
ahmed-shameem: xml_collection_id column is hardcoded as 0 in sys.columns. If we have any way to obtain this value in future, then we have to update this definition as same as sys.columns_internal. Previously, if this value was 0 then we return NULL cast into INT. This is why this change is made.

contrib/babelfishpg_tsql/src/procedures.c
ahmed-shameem: Added a JOIN condition between T2.user_type_id = t.oid as system_type_id of sys.types is union of t.oid and t.typbasetype. Now, currently we are only returning rows for system types only so we can certainly do a join between system_type_id of T2 with pg_type's oid. However, in future when we extend the output rows for user defined types as well, then we will have update it to this anyway.

contrib/babelfishpg_tsql/src/procedures.c
ahmed-shameem: Updated in latest revision.

contrib/babelfishpg_tsql/src/procedures.c
Deepesh125: This is still not a right way. We should group the join appropriately with appropriate join condition. What you have done is called qual condition on join which will be applied later as a filter on the result of join but adding it as a join condition (`on` clause) will cause executor to fetch lesser no of rows. Hence some gain in the performance.

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -1283,145 +1287,299 @@ sp_describe_undeclared_parameters_internal(PG_FUNCTION_ARGS)
...
-		"AND O.name = \'%s\' COLLATE sys.database_default " /* -- INPUT table name */
-		"AND O.schema_id = %d " /* -- INPUT schema Oid */
-		"AND O.type = \'U\'";	/* -- User tables only for the time being */
-
-		char	   *query = psprintf(tempq,
-									 undeclaredparams->targetcolnames[undeclaredparams->paramindexes[call_cntr]],
-									 undeclaredparams->tablename,
-									 undeclaredparams->schemaoid);
+		"FROM sys.types AS T2 "
+		"JOIN pg_attribute AS a ON a.atttypid = T2.user_type_id "
```
Deepesh125: pg_attribute join sys.type would make more sense

ahmed-shameem: Updated in latest rev.

KushaalShroff:
6 seconds is still a lot for a response time for 100 params

KushaalShroff (approved):

contrib/babelfishpg_tsql/src/procedures.c
ahmed-shameem: Updated in latest rev

contrib/babelfishpg_tsql/src/procedures.c
ahmed-shameem: Updated in latest rev

Deepesh125 (approved):

Deepesh125 (approved):

PR #1919, opened 2023-10-13
DROP DB Performance improvements

tanscorpio7:
### Description

Similar fixes to [Create DB PR](https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/1899)
Relies on this as well [PR LINK](https://github.com/babelfish-for-postgresql/postgresql_modified_for_babelfish/pull/234)

For DROP DB we currently,
-- assume role sysadmin
-- drop all related objects (roles, schema, etc)
-- reset role to prev user

1. DROP SCHEMA -  the owner can drop the schema (and thereby all contained objects) even if they do not own some of the objects within the schema. [Reference Link](https://www.postgresql.org/docs/current/sql-dropschema.html)
2. DROP OWNED BY role1, role2 - drops all the objects within the current database that are owned by one of the specified roles. Any privileges granted to the given roles on objects in the current database or on shared objects (databases, tablespaces, configuration parameters) will also be revoked. [Reference Link](https://www.postgresql.org/docs/current/sql-drop-owned.html)

Based on the above definitions we can set the current_user to dbo role and still be able to drop everything inside the babelfish database with the only exception being the PG Database privileges granted to any roles being dropped. Since only sysadmin has the grant rights to PG Database, we add an additional revoke statement for dropping these privileges under the assumed role sysadmin.

This improves the performance by reducing the time taken to run roles_is_member_of( roleId )
When no of DBs >> 1
  roles_is_member_of( 'any_db_name_dbo' ) << roles_is_member_of( 'sysadmin' )

So we set the current user to dbo wherever possible.

| DROP DB              | After | Before |
| DBs2990   |   15 ms  | NA |
| DBs1000   |    7 ms  | 329 ms |
| DBs500     |   6 ms  | 98 ms |
| DBs10        |   4 ms  | 6 ms |

### Issues Resolved

[BABEL-4438]

### Check List


shalinilohia50 (approved):

contrib/babelfishpg_tsql/src/dbcmds.c
```diff
@@ -225,6 +225,7 @@ gen_dropdb_subcmds(const char *schema,
 	}
 	/* Then drop db_owner and dbo in that order */
 	appendStringInfo(&query, "DROP OWNED BY dummy, dummy CASCADE; ");
```
forestkeeper, line 227: I'm confused why we use dummy in this case ?

forestkeeper:
DROP SCHEMA - the owner can drop the schema (and thereby all contained objects) even if they do not own some of the objects within the schema.
     -> Would add some reference where is this assume comes from.
DROP OWNED BY role1, role2 - drops all the objects within the current database that are owned by one of the specified roles. Any privileges granted to the given roles on objects in the current database or on shared objects (databases, tablespaces, configuration parameters) will also be revoked.
    -> Would add some reference where is this assume comes from.

contrib/babelfishpg_tsql/src/dbcmds.c
tanscorpio7: just a placeholder, we replace it with the correct value after the statement has been converted into a node

contrib/babelfishpg_tsql/src/dbcmds.c
```diff
@@ -716,7 +723,15 @@ drop_bbf_db(const char *dbname, bool missing_ok, bool force_drop)
 		{
 			Node	   *stmt = ((RawStmt *) lfirst(parsetree_item))->stmt;
 			PlannedStmt *wrapper;
+			is_set_userid = false;
 
+			if(stmt->type != T_DropRoleStmt && stmt->type != T_GrantStmt)
```
Deepesh125, line 728: why cant we simply check if stmt->type is dropdb cmd?

tanscorpio7: We are iterating through the sub-commands of bbf drop db cmd. None of which are actually T_Dropdb since bbf db is actually a pg namespace.

We skip setuser(dbo) for 
T_DropRoleStmt : Current user should not be same as the user being dropped
T_GrantStmt : dbo does not have enough permission to run this stmt (REVOKE stmt)

tanscorpio7: So the stmts for which we want to set current_user = dbo are
"DROP SCHEMA dummy CASCADE"
"DROP OWNED BY dummy, .. CASCADE"

forestkeeper (approved):

Deepesh125 (approved):

PR #1933, opened 2023-10-18
Support error code 11506 in sp_describe_undeclared_parameters

### LINKED ISSUE #1354: [Bug]: sp_describe_undeclared_parameters breaks transaction on failure
### What happened?

[PyODBC](https://github.com/mkleehammer/pyodbc) uses `SQLDescribeParam` [ODBC call](https://learn.microsoft.com/en-us/sql/odbc/reference/syntax/sqldescribeparam-function?view=sql-server-ver16) to get the type of the parameter when `None` is specified as a parameter value ([details](https://github.com/mkleehammer/pyodbc/blob/4829107c39fdd066b3d86ed6c226fb933b712eba/src/params.cpp#L2083)). `SQLDescribeParam` calls `sp_describe_undeclared_parameters` stored procedure. These calls are done on a "best effort" basis, it is not always possible for MSSQL/Babelfsh to derive the correct parameter type based on the query text. PyODBC expects possible failure of `SQLDescribeParam` and [handles it gracefully](https://github.com/mkleehammer/pyodbc/blob/4829107c39fdd066b3d86ed6c226fb933b712eba/src/params.cpp#L2124). The problem is that the transactional behaviour after `sp_describe_undeclared_parameters` is different with Babelfsh comparing to MSSQL.

Consider the following example:

```
import pyodbc
conn = pyodbc.connect("DRIVER={ODBC Driver 18 for SQL Server};...")
cursor = conn.cursor();
cursor.execute("select ?", [None]);
records = cursor.fetchall();
print(records[0][0])
# do some inserts/updates
conn.commit()
print("success")
```

In this case `sp_describe_undeclared_parameters` will fail for both MSSQL an Babelfish. Just with MSSQL the commit call on the same connection is handled correctly, MSSQL output:

```
$ python test.py 
None
success
```

But Babelfish, while it returns the query resulting data correctly, it does not allow to perform commit on the same connection, its output:

```
$ python test.py 
None
Traceback (most recent call last):
  File "/home/alex/[...]/test.py", line 11, in <module>
    conn.commit()
pyodbc.ProgrammingError: ('25000', '[25000] [Microsoft][ODBC Driver 18 for SQL Server][SQL Server]COMMIT can only be used in transaction blocks (3902) (SQLEndTran)')
```

Please note, that the problem is not that this is an `Unsupported use case in sp_describe_undeclared_parameters` (the example above is also unsupported in MSSQL), but in the transactional behaviour after the `sp_describe_undeclared_parameters` failure.

Besides that, PyODBC gives little flexibility to work with transactions ([docs](https://github.com/mkleehammer/pyodbc/wiki/Objects#commit)), excluding `autocommit` mode, the transaction is always started when the connection is opened. While `connection.commit()` is called explicitly from python code, after the select query is run (and its correct data is returned) there is no information on Python level whether `sp_describe_undeclared_parameters` failure happened or not (because the failure is handled inside PyODBC) and it is not known whether `commit()` can be called on this connection or not.

It breaks the Python application, where there are many queries like the following (with `None`s used for parameters not supplied by end users):

```
select * from tbl1
where
((? is null) or (id = ?))
and 
((? is null) or (name = ?))
...
```

As I understand (please correct me if I am wrong), it is not possible to change the `sp_describe_undeclared_parameters` behaviour by changing its `plpgsql` definition. And patching `procedures.c` is not an option for us because we'd like to use AWS Babelfish on Aurora.

As a workaround we are currently using a locally patched version of PyODBC (with `SQLDescribeParam` patched out), comments on any other possible workarounds are highly appreciated.

### Version

BABEL_3_X_DEV (Default)

### Extension

babelfishpg_tsql (Default)

Fedora

### Relevant log output

```shell
2023-03-21 13:38:21.341 GMT [28197] LOG:  Unmapped error found. Code: 1088, Message: Unsupported use case in sp_describe_undeclared_parameters, File: procedures.c, Line: 1031, Context: babelfishpg_tsql
2023-03-21 13:38:21.341 GMT [28197] CONTEXT:  PL/tsql function sp_describe_undeclared_parameters(nvarchar,nvarchar) line 3 at SQL statement
	SQL statement "EXEC sp_describe_undeclared_parameters @p1"
	PL/tsql function inline_code_block line 1 at EXEC
	TDS Protocol: Message Type: RPC, SP Type: SP_CUSTOMTYPE, Phase: TDS_REQUEST_PHASE_PROCESS. Preparing to Send Back the Tds response
2023-03-21 13:38:21.341 GMT [28197] LOG:  Unmapped error found. Code: 1088, Message: Unsupported use case in sp_describe_undeclared_parameters, File: procedures.c, Line: 1031, Context: babelfishpg_tsql
2023-03-21 13:38:21.341 GMT [28197] CONTEXT:  SQL statement "EXEC sp_describe_undeclared_parameters @p1"
	PL/tsql function inline_code_block line 1 at EXEC
	TDS Protocol: Message Type: RPC, SP Type: SP_CUSTOMTYPE, Phase: TDS_REQUEST_PHASE_PROCESS. Preparing to Send Back the Tds response
2023-03-21 13:38:21.341 GMT [28197] LOG:  Unmapped error found. Code: 1088, Message: Unsupported use case in sp_describe_undeclared_parameters, File: procedures.c, Line: 1031, Context: babelfishpg_tsql
2023-03-21 13:38:21.341 GMT [28197] CONTEXT:  PL/tsql function inline_code_block line 1 at EXEC
	TDS Protocol: Message Type: RPC, SP Type: SP_CUSTOMTYPE, Phase: TDS_REQUEST_PHASE_PROCESS. Preparing to Send Back the Tds response
2023-03-21 13:38:21.341 GMT [28197] LOG:  Unmapped error found. Code: 1088, Message: Unsupported use case in sp_describe_undeclared_parameters, File: procedures.c, Line: 1031, Context: babelfishpg_tsql
2023-03-21 13:38:21.341 GMT [28197] CONTEXT:  TDS Protocol: Message Type: RPC, SP Type: SP_CUSTOMTYPE, Phase: TDS_REQUEST_PHASE_PROCESS. Preparing to Send Back the Tds response
2023-03-21 13:38:21.341 GMT [28197] ERROR:  Unsupported use case in sp_describe_undeclared_parameters
2023-03-21 13:38:21.341 GMT [28197] CONTEXT:  PL/tsql function sp_describe_undeclared_parameters(nvarchar,nvarchar) line 3 at SQL statement
	SQL statement "EXEC sp_describe_undeclared_parameters @p1"
	PL/tsql function inline_code_block line 1 at EXEC
	TDS Protocol: Message Type: RPC, SP Type: SP_CUSTOMTYPE, Phase: TDS_REQUEST_PHASE_PROCESS. Preparing to Send Back the Tds response
2023-03-21 13:38:21.341 GMT [28197] LOG:  Unmapped error found. Code: 1088, Message: Unsupported use case in sp_describe_undeclared_parameters, File: procedures.c, Line: 1031, Context: TDS
2023-03-21 13:38:21.341 GMT [28197] CONTEXT:  TDS Protocol: Message Type: RPC, SP Type: SP_CUSTOMTYPE, Phase: TDS_REQUEST_PHASE_PROCESS. Preparing to Send Back the Tds response
2023-03-21 13:38:21.342 GMT [28197] LOG:  message_type: Remote Procedure Call (3) rpc_packet_type: sp_prepexec
2023-03-21 13:38:21.342 GMT [28197] CONTEXT:  TDS Protocol: Message Type: RPC, SP Type: SP_PREPEXEC, Phase: TDS_REQUEST_PHASE_FETCH. Fetching RPC Request
2023-03-21 13:38:21.342 GMT [28197] LOG:  sp_prepexec handle: 0, statement: select @P1
2023-03-21 13:38:21.342 GMT [28197] DETAIL:  Parameters: $1 = NULL
2023-03-21 13:38:21.343 GMT [28197] LOG:  message_type: Transaction Management Request (14) txn_request_type: TM_COMMIT_XACT
2023-03-21 13:38:21.343 GMT [28197] CONTEXT:  TDS Protocol: Message Type: Txn Manager, Txn Type: TM_COMMIT_XACT, Phase: TDS_REQUEST_PHASE_FETCH. Fetching Transaction Management Request
2023-03-21 13:38:21.376 GMT [28197] ERROR:  COMMIT can only be used in transaction blocks
2023-03-21 13:38:21.376 GMT [28197] CONTEXT:  SQL statement "COMMIT TRANSACTION ;"
	PL/tsql function inline_code_block line 1 at SQL statement
	TDS Protocol: Message Type: Txn Manager, Txn Type: TM_COMMIT_XACT, Phase: TDS_REQUEST_PHASE_PROCESS. Processing Transaction Management Request
2023-03-21 13:38:21.438 GMT [28197] LOG:  message_type: Remote Procedure Call (3) rpc_packet_type: sp_unprepare
2023-03-21 13:38:21.438 GMT [28197] CONTEXT:  TDS Protocol: Message Type: RPC, SP Type: SP_UNPREPARE, Phase: TDS_REQUEST_PHASE_FETCH. Fetching RPC Request
2023-03-21 13:38:21.438 GMT [28197] LOG:  could not receive data from client: Connection reset by peer
2023-03-21 13:38:21.438 GMT [28197] CONTEXT:  TDS Protocol: Message Type: RPC, SP Type: Unknown (Pre-Parsing Request), Phase: TDS_REQUEST_PHASE_FETCH. Reading data from socket
2023-03-21 13:38:21.438 GMT [28197] LOG:  EOF on TDS socket
2023-03-21 13:38:21.438 GMT [28197] CONTEXT:  TDS Protocol: EOF reached on TDS socket
```

### Code of Conduct

timchang514:
### Description

Support error code 11506 in sp_describe_undeclared_parameters. In cases where a transaction block is supplied (such as with pyodbc driver), Babelfish would throw an unsupported error message. We map the supported error message to an expected value. 

### Issues Resolved

#1354 

Additional manual testing with the script provided. The result is as expected (transaction is not aborted, an error message is thrown)

```
$ python test.py 
None
success
```

### Check List

### DIFF
--- a/contrib/babelfishpg_tds/error_mapping.txt
+++ b/contrib/babelfishpg_tds/error_mapping.txt
@@ -186,4 +186,4 @@ XX000 ERRCODE_INTERNAL_ERROR	"The table-valued parameter \"%s\" must be declared
 22008 ERRCODE_DATETIME_VALUE_OUT_OF_RANGE	"Adding a value to a \'%s\' column caused an overflow." SQL_ERROR_517 16
 42P01 ERRCODE_UNDEFINED_TABLE	"FOR JSON AUTO requires at least one table for generating JSON objects. Use FOR JSON PATH or add a FROM clause with a table name."		SQL_ERROR_13600	16
 42P01 ERRCODE_FEATURE_NOT_SUPPORTED	"Values for json auto is not currently supported."		SQL_ERROR_13600	16
-
+22023 ERRCODE_INVALID_PARAMETER_VALUE "The parameter type could not be uniquely deduced." SQL_ERROR_11506 16
--- a/contrib/babelfishpg_tsql/src/err_handler.c
+++ b/contrib/babelfishpg_tsql/src/err_handler.c
@@ -137,6 +137,7 @@ is_ignorable_error(int pg_error_code, uint8_t override_flag)
 		case SQL_ERROR_535:
 		case SQL_ERROR_13600:
 		case SQL_ERROR_15003:
+		case SQL_ERROR_11506:
 			{
 				elog(DEBUG1, "TSQL TXN is_ignorable_error %d", latest_error_code);
 				return true;
--- a/contrib/babelfishpg_tsql/src/err_handler.h
+++ b/contrib/babelfishpg_tsql/src/err_handler.h
@@ -142,6 +142,7 @@ uint8_t		override_txn_behaviour(PLtsql_stmt *stmt);
 #define SQL_ERROR_10727 10727
 #define SQL_ERROR_10733 10733
 #define SQL_ERROR_10793 10793
+#define SQL_ERROR_11506 11506
 #define SQL_ERROR_11555 11555
 #define SQL_ERROR_11700 11700
 #define SQL_ERROR_11701 11701
--- a/test/JDBC/input/BABEL-2408.sql
+++ b/test/JDBC/input/BABEL-2408.sql
@@ -1,0 +1,14 @@
+sp_describe_undeclared_parameters N'SELECT @p1', NULL
+GO
+
+sp_describe_undeclared_parameters N'SELECT @p1', 1
+GO
+
+sp_describe_undeclared_parameters N'SELECT @p1', N'text_test'
+GO
+
+sp_describe_undeclared_parameters N'SELECT * FROM t1'
+GO
+
+sp_describe_undeclared_parameters N'SELECT COUNT(*) FROM sys.assemblies'
+GO
--- a/contrib/babelfishpg_tsql/src/procedures.c
+++ b/contrib/babelfishpg_tsql/src/procedures.c
@@ -945,7 +945,6 @@ sp_describe_undeclared_parameters_internal(PG_FUNCTION_ARGS)
 		char	   *params;
 		int			sql_dialect_value_old;
 
-		SelectStmt *select_stmt;
 		List	   *values_list;
 		ListCell   *lc;
 		int			numresults = 0;
@@ -954,6 +953,7 @@ sp_describe_undeclared_parameters_internal(PG_FUNCTION_ARGS)
 		InsertStmt *insert_stmt = NULL;
 		UpdateStmt *update_stmt = NULL;
 		DeleteStmt *delete_stmt = NULL;
+		SelectStmt *select_stmt = NULL;
 		RangeVar   *relation;
 		Oid			relid;
 		Relation	r;
@@ -1104,6 +1104,23 @@ sp_describe_undeclared_parameters_internal(PG_FUNCTION_ARGS)
 				extra_restargets = handle_where_clause_restargets_left(pstate, delete_stmt->whereClause, extra_restargets);
 
 				cols = list_concat_copy(cols, extra_restargets);
+				break;
+			case T_SelectStmt:
+				/*
+				 * Select is not yet supported, but we can throw a mapped error message to avoid aborting the transaction in known cases.
+				 */
+				rewrite_object_refs(parsetree->stmt);
+				sql_dialect = sql_dialect_value_old;
+				select_stmt = (SelectStmt *)parsetree->stmt;
+				if (!select_stmt->fromClause)
+				{
+					ereport(ERROR,
+							(errcode(ERRCODE_INVALID_PARAMETER_VALUE),
+							errmsg("The parameter type could not be uniquely deduced.")));
+				}
+				ereport(ERROR,
+						(errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
+						 errmsg("Unsupported use case in sp_describe_undeclared_parameters")));
 				break;
 			default:
 				ereport(ERROR,
--- a/test/JDBC/expected/BABEL-2408.out
+++ b/test/JDBC/expected/BABEL-2408.out
@@ -1,0 +1,44 @@
+sp_describe_undeclared_parameters N'SELECT @p1', NULL
+GO
+~~START~~
+int#!#varchar#!#int#!#nvarchar#!#smallint#!#tinyint#!#tinyint#!#int#!#varchar#!#varchar#!#varchar#!#nvarchar#!#int#!#varchar#!#varchar#!#varchar#!#bit#!#bit#!#bit#!#bit#!#bit#!#varchar#!#int#!#int
+~~ERROR (Code: 11506)~~
+
+~~ERROR (Message: The parameter type could not be uniquely deduced.)~~
+
+
+sp_describe_undeclared_parameters N'SELECT @p1', 1
+GO
+~~START~~
+int#!#varchar#!#int#!#nvarchar#!#smallint#!#tinyint#!#tinyint#!#int#!#varchar#!#varchar#!#varchar#!#nvarchar#!#int#!#varchar#!#varchar#!#varchar#!#bit#!#bit#!#bit#!#bit#!#bit#!#varchar#!#int#!#int
+~~ERROR (Code: 33557097)~~
+
+~~ERROR (Message: syntax error at or near "PROC")~~
+
+
+sp_describe_undeclared_parameters N'SELECT @p1', N'text_test'
+GO
+~~START~~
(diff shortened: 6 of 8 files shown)


contrib/babelfishpg_tds/error_mapping.txt
```diff
@@ -179,4 +179,4 @@ XX000 ERRCODE_INTERNAL_ERROR	"The table-valued parameter \"%s\" must be declared
 0A000 ERRCODE_FEATURE_NOT_SUPPORTED	"Column name or number of supplied values does not match table definition."	SQL_ERROR_213	16
 42501 ERRCODE_INSUFFICIENT_PRIVILEGE	"Only members of the sysadmin role can execute this stored procedure."	SQL_ERROR_15003	16
 42809 ERRCODE_WRONG_OBJECT_TYPE	 "The target \"%s\" of the OUTPUT INTO clause cannot be a view or common table expression."	SQL_ERROR_330	16
-
+0A000 ERRCODE_FEATURE_NOT_SUPPORTED "Unsupported statement type for sp_describe_undeclared_parameters" SQL_ERROR_11506 16
```
KushaalShroff: We are mapping 11506 for a generic feature not supported PG error code, I dont think that's the right approach

test/JDBC/input/BABEL-2408.sql
```diff
@@ -0,0 +1,2 @@
+EXEC sp_describe_undeclared_parameters 'COMMIT TRANSACTION;'
```
KushaalShroff: The customer doesnt seem to have an issue with 'COMMIT TRANSACTION' stmt used with sp_describe_undeclared_parameters.
It seems that sp_describe_undeclared_parameters is being used for some datatype which is not valid for SQL Server too. 
So the python API fails and then the individual query for 'COMMIT TRANSACTION' is executed.
In babelfish we throw generic unsupported error message which is intentionally transaction aborting.
Either we introduce a better error catch scenario or change the PG error code from being unsupported and then map that error code appropriately

contrib/babelfishpg_tsql/src/err_handler.h
```diff
@@ -139,6 +139,7 @@ uint8_t		override_txn_behaviour(PLtsql_stmt *stmt);
 #define SQL_ERROR_10727 10727
 #define SQL_ERROR_10733 10733
 #define SQL_ERROR_10793 10793
+#define SQL_ERROR_11506 11506
```
KushaalShroff, line 145: Introducing a new error code also requires more dive deep to get what behaviour it aligns to.
On the top of my head we have:
1. Statement aborting
2. batch aborting
3. transaction aborting
Based on any of these we need to add them appropriately in the switch cases of err_handler.c

KushaalShroff (changes requested):
We should also try to add python tests since we already have a Py/pyodbc framework

contrib/babelfishpg_tds/error_mapping.txt
timchang514: So I think there's not really a good solution here. The specific error code `11506` that is being run into seems to be thrown in this case because the select statement is too generic and so can't be properly deduced. We don't yet support SELECT statements in `sp_dup`, and if we add proper support here, it's not getting to the core of the problem. We could try to support this one specific error message, but it's right to question how much value there is to fix just this one specific use case. 

The general issue mentioned in #1354 is that I guess most errors thrown in `sp_dup` are unmapped, so that when an error is encountered, the whole transaction is aborted. One middle-ground solution could be to create a new PG error code specific to `sp_dup` that we can then map it to existing TSQL error codes?

timchang514:
Have addressed other comments, but I don't think it makes a lot of sense to add a test in pyodbc for this. The original case that showed the bug is due to the specific SQL that pyodbc is sending behind the scenes. To simulate this, we would have to write pure python code rather than provide some input files in a `.txt` file. 

In place of that, I have added tests in JDBC that mimic the behavior that we are actually trying to support - the newly added error message, and not aborting a transaction if the error message is encountered.

contrib/babelfishpg_tsql/src/err_handler.c
```diff
@@ -137,6 +137,7 @@ is_ignorable_error(int pg_error_code, uint8_t override_flag)
 		case SQL_ERROR_535:
 		case SQL_ERROR_13600:
 		case SQL_ERROR_15003:
+		case SQL_ERROR_11506:
```
surendravishnoi, line 140: Please sync with Dipesh to find out how to deduce transactional/error behaviour for an error. We cannot map it without knowing full behaviour. It will also give us relevant test cases to validate the expected behaviour.

PR #1937, opened 2023-10-19
Resolve TDS protocol violation occurring due to parallel query enforced

Deepesh125:
### Description

This issue was mainly caused by incorrect  inference of typmod information for numeric expression where its being set to default when
Parallel worker is planned. Reason for the same is explained below: 

When parallel query is enforced, Postgres engine planes Gather node on the top of top plan. For example, consider following query - 
SELECT amount + 100 FROM overflow_test where id = 1   


```
-- If parallel query is not enforced- 
Query Text: SELECT amount + 100 FROM overflow_test where id = 1
Index Scan using overflow_test_pkey on master_dbo.overflow_test  (cost=0.15..8.17 rows=1 width=32)
  Output: ((amount)::numeric + '100'::numeric)
  Index Cond: (overflow_test.id = 1)

-- If parallel query is enforced then Postgres will wrap the above plan under Gather node - 
Query Text: SELECT amount + 100 FROM overflow_test where id = 1                                                                                                                                                                                 
Gather  (cost=0.15..8.17 rows=1 width=32)                                                                                                                                                                                                       
  Output: (((amount)::numeric + '100'::numeric))                                                                                                                                                                                                
  Workers Planned: 1                                                                                                                                                                                                                            
  Single Copy: true                                                                                                                                                                                                                             
  ->  Index Scan using overflow_test_pkey on master_dbo.overflow_test  (cost=0.15..8.17 rows=1 width=32)                                                                                                                                        
        Output: ((amount)::numeric + '100'::numeric)                                                                                                                                                                                            
        Index Cond: (overflow_test.id = 1)    
```                                                                                                

When Postgres does this, it also modifies targetlist of gather node to refer to the tuples returned by its lefttree subplan or outerplan (Indexscan in this case). Hence, targetlist of gather node will contain only one tle in above case which is Var with attno = OUTER_VAR (it indicates reference to its outer plan) and vartypmod being -1. 

Now, While sending metadata from TDS layer we use this  typmod to deduce max_scale and precision for numeric. This would be set to default since vartypmod = -1 whereas actual data being computed may not fit in it hence it will run into an error while sending numeric response. Hence, hang or crash of end client software.

This commit fixes this issue by taking special Var into account while We extract the information of outer plan from planned stmt and use the same to get the original/referenced tle from outer plan. 

This commit further improves the implementation of resolve_numeric_typmod_from_exp to handle reference to outer var correctly. It can now handle Append and MergeAppend node correctly which was required for following test case - 

```
-- setup
create table t1 (a numeric(6,4), b numeric(6,3));
insert into t1 values (4, 16);
insert into t1 values (10.1234, 10.123);
insert into t1 values (1.2, 6);
insert into t1 values (NULL, 101.123);

-- test
select a from t1 UNION All
select b from t1;
```

For this UNION ALL expression, typmod was always computed to be -1 which is not right and it should (7,4) ideally. This commit introduces helped function called resolve_numeric_typmod_from_append_or_mergeappend to compute typmod for expression correctly.

Task: BABEL-4424, BABEL-4359

### Check List


contrib/babelfishpg_tds/src/backend/tds/tdsprinttup.c
```diff
@@ -60,6 +60,13 @@ TdsPrinttupStartup(DestReceiver *self, int operation, TupleDesc typeinfo)
 {
 	DR_printtup *myState = (DR_printtup *) self;
 	Portal		portal = myState->portal;
+	PlannedStmt *plannedStmt = PortalGetPrimaryStmt(portal);
+	List 		*targetList = NIL;
+
+	if (portal->strategy != PORTAL_MULTI_QUERY)
```
surendravishnoi, line 66: can TdsSendRowDescription handle NULL targetList?

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
```diff
@@ -401,9 +404,86 @@ PrintTupPrepareInfo(DR_printtup *myState, TupleDesc typeinfo, int numAttrs)
...
+	{
+		planlist = ((Append *) plan)->appendplans;
+	}
+	else if(IsA(plan, MergeAppend))
+	{
+		planlist = ((MergeAppend *) plan)->mergeplans; 
+	}
+
+	Assert(planlist != NIL);
+	foreach(lc, ((Append *) plan)->appendplans)
```
surendravishnoi: can you explain what do multiple plans indicate here?

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
```diff
@@ -401,9 +404,86 @@ PrintTupPrepareInfo(DR_printtup *myState, TupleDesc typeinfo, int numAttrs)
...
+				typmod = resolve_numeric_typmod_outer_var(outerplan, var->varattno);
+			}
+		}
+		typmod = resolve_numeric_typmod_from_exp(outerplan, (Node *)tle->expr);
+		if (typmod == -1)
+			continue;
+		scale = (typmod - VARHDRSZ) & 0xffff;
+		precision = ((typmod - VARHDRSZ) >> 16) & 0xffff;
+		integralDigitCount = Max(precision - scale, max_precision - max_scale);
+		max_scale = Max(max_scale, scale);
```
surendravishnoi, line 460: can there be an overflow here?

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
```diff
@@ -401,9 +404,86 @@ PrintTupPrepareInfo(DR_printtup *myState, TupleDesc typeinfo, int numAttrs)
...
+		max_precision = integralDigitCount + max_scale;
```
surendravishnoi, line 461: what is the logic behind mixing precision and scale across the plans?

test/JDBC/input/babel_613.sql
```diff
@@ -60,8 +60,12 @@ select coalesce(a, b) from t1;
 go
 
 -- test Union All
-select a from t1 Union All
-select b from t1;
+select * from
```
surendravishnoi, line 63: are we planning to add more tests? from code changes it seem there should be good number of new tests

contrib/babelfishpg_tds/src/backend/tds/tdsprinttup.c
Deepesh125: This is taken from FetchPortalTargetList where it could return NIL.

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
Deepesh125: It indicates separate plan for all the subqueries that involves in UNION operation. For example,
```
select t1 from table1 
union
select t2 from table2
```
In above case, we would have two append plan one for each select .

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
Deepesh125: This logic will ensure that it can accommodate max scale and max precision across the plans. For example, numeric(6,4) and numeric(6,3), final result would be numeric(7,4)

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
Deepesh125: As per pg, It supports up to 131072 digits before the decimal point; up to 16383 digits after the decimal point. So I think we need to handle it. Will do in next revision.

test/JDBC/input/babel_613.sql
Deepesh125: Added some test cases

contrib/babelfishpg_tds/src/backend/tds/tdsprinttup.c
surendravishnoi: do we have a test case where we call TdsSendRowDescription with NULL trargetList? FetchPortalTargetList can return NULL but we should validate if TdsSendRowDescription can handle NULL (if it can and we have tests cases, we are good)

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
surendravishnoi: I was talking more from data type perspective. Can we allow a numeric result for which SQL server will throw an error due to overflow?

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
surendravishnoi: does SQL server have same behavior? What if result is beyond max supported by SQL server?

test/JDBC/input/BABEL-4424.sql
```diff
@@ -0,0 +1,104 @@
+create table babel_4359_t1 (a numeric(6,4), b numeric(6,3), c numeric);
+go
+
+insert into babel_4359_t1 values (4, 16, 1111);
```
surendravishnoi, line 4: let us add some test cases with boundary values for numeric, also have numeric type with min/max precision/scale (from SQL server perspective)

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
```diff
@@ -401,9 +404,105 @@ PrintTupPrepareInfo(DR_printtup *myState, TupleDesc typeinfo, int numAttrs)
...
+		tle = get_tle_by_resno(outerplan->targetlist, attno);
+		if (IsA(tle->expr, Var))
+		{
+			Var *var = (Var *)tle->expr;
+			if (var->varno == OUTER_VAR)
+			{
+				typmod = resolve_numeric_typmod_outer_var(outerplan, var->varattno);
+			}
+		}
+		typmod = resolve_numeric_typmod_from_exp(outerplan, (Node *)tle->expr);
```
surendravishnoi: we are overriding earlier typmod from outer_var function?

test/JDBC/parallel_query_jdbc_schedule
```diff
@@ -342,3 +331,6 @@ ignore#!#TestSimpleErrors
 ignore#!#ISC-Views
 ignore#!#TestSimpleErrorsWithXactAbort
 ignore#!#BABEL-2513
+ignore#!#Test-sp_rename-dep-vu-prepare
```
surendravishnoi: let us add Jira and reason to ignore

test/JDBC/expected/BABEL-3943.out
```diff
@@ -16,7 +16,6 @@ GO
 
 
 -- tsql
--- TODO: Fix BABEL-4359
```
surendravishnoi, line 19: why there is no output change after fix?

test/JDBC/expected/BABEL-4424.out
```diff
@@ -0,0 +1,445 @@
...
+10.1230
+10.1230
+10.1234
+16.0000
+101.1230
+1111.0000
+222222.0000
+33333333333333333.0000
+444444444444444444.0000
+444444444444444444.0000
```
surendravishnoi, line 75: where is scale part 00 instead of 44?

test/JDBC/expected/BABEL-4424.out
```diff
@@ -0,0 +1,445 @@
+create table babel_4359_t1 (a numeric(6,4), b numeric(6,3), c numeric);
```
surendravishnoi, line 1: have we validated it against SQL server?

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -2910,8 +2910,13 @@ pltsql_detect_numeric_overflow(int weight, int dscale, int first_block, int nume
 	if (sql_dialect != SQL_DIALECT_TSQL)
 		return false;
 
-	total_digit_count = (dscale == 0) ? (weight * numeric_base) :
-		((weight + 1) * numeric_base);
+	if (weight < 0)
```
surendravishnoi, line 2920: did any test fail due to this fix? also, we should add extensive tests for this around numeric overflow. Having both integer and scale being 1-4 5-8 9-12 and so on

test/JDBC/expected/BABEL-3943.out
Deepesh125: Check line no 43--17

test/JDBC/expected/BABEL-4424.out
Deepesh125: Because default scale is 0

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
Deepesh125: Resolved.

test/JDBC/parallel_query_jdbc_schedule
Deepesh125: Its under common jira open for time out issues

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
```diff
@@ -401,9 +404,112 @@ PrintTupPrepareInfo(DR_printtup *myState, TupleDesc typeinfo, int numAttrs)
...
+	{
+		planlist = ((Append *) plan)->appendplans;
+	}
+	else if(IsA(plan, MergeAppend))
+	{
+		planlist = ((MergeAppend *) plan)->mergeplans; 
+	}
+
+	Assert(planlist != NIL);
+	foreach(lc, ((Append *) plan)->appendplans)
```
surendravishnoi: shouldn't we use planlist here? If yes, why didn't any of tests fail due to this miss?

contrib/babelfishpg_tds/src/backend/tds/tdsprinttup.c
```diff
@@ -60,6 +60,13 @@ TdsPrinttupStartup(DestReceiver *self, int operation, TupleDesc typeinfo)
...
+	{
+		targetList = FetchStatementTargetList((Node *) plannedStmt);
```
surendravishnoi, line 68: let us double check that we can handle null plannedStmt here and null targetList in next call

contrib/babelfishpg_tds/src/backend/tds/tdsrpc.c
```diff
@@ -2334,10 +2334,15 @@ static void
 SendCursorResponse(TDSRequestSP req)
 {
 	int			cmd_type = TDS_CMD_UNKNOWN;
-	Portal		portal;
-
 	/* fetch the portal */
-	portal = GetPortalFromCursorHandle(req->cursorHandle, false);
+	Portal		portal = GetPortalFromCursorHandle(req->cursorHandle, false);
```
surendravishnoi, line 2338: have we added test case for this code path?

test/JDBC/expected/BABEL-4424.out
surendravishnoi: have we validated union query output against SQL server?

test/JDBC/expected/BABEL-4424.out
```diff
@@ -0,0 +1,725 @@
...
+~~END~~
+
+
+DROP TABLE babel_4424_t1;
+go
+DROP TABLE babel_4424_t2;
+go
+
+
+create table babel_4424_t1 (n3_0 numeric(3,0), n3_1 numeric(3,1), n6_0 numeric(6,0), n10_0 numeric(10,0), n10_9 numeric(10, 9),
```
surendravishnoi, line 487: 1) test case where weight < 0, fraction only 2) test case where we have overflow due to integer values only 3) NULL values in mix

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
Deepesh125: No, test was not failing because initial few field of both the data structures are same:
```
typedef struct Append
{
	Plan		plan;
	Bitmapset  *apprelids;		/* RTIs of appendrel(s) formed by this node */
	List	   *appendplans;
	int			nasyncplans;	/* # of asynchronous plans */

	/*
	 * All 'appendplans' preceding this index are non-partial plans. All
	 * 'appendplans' from this index onwards are partial plans.
	 */
	int			first_partial_plan;

	/* Info for run-time subplan pruning; NULL if we're not doing that */
	struct PartitionPruneInfo *part_prune_info;
} Append;
```

```

typedef struct MergeAppend
{
	Plan		plan;
	Bitmapset  *apprelids;		/* RTIs of appendrel(s) formed by this node */
	List	   *mergeplans;
	/* these fields are just like the sort-key info in struct Sort: */
	int			numCols;		/* number of sort-key columns */
	AttrNumber *sortColIdx;		/* their indexes in the target list */
	Oid		   *sortOperators;	/* OIDs of operators to sort them by */
	Oid		   *collations;		/* OIDs of collations */
	bool	   *nullsFirst;		/* NULLS FIRST/LAST directions */
	/* Info for run-time subplan pruning; NULL if we're not doing that */
	struct PartitionPruneInfo *part_prune_info;
} MergeAppend;
```

contrib/babelfishpg_tds/src/backend/tds/tdsrpc.c
Deepesh125: This is internal implementation of `FetchPortalTargetList` only. Since we require plannedstmt, we implemted similar logic so that we can get both plannedstmt and targetlist for our requirement. I dont know how to test this change though for targetList = NIL. Does it make sense to put `assert(targetList != NIL)` here

contrib/babelfishpg_tsql/src/hooks.c
Deepesh125: Nope.. We have added additional test cases. But you can check the table with `create table babel_4424_t6 (a numeric(10, 10));` where weight would be -1  and dscale = 10
Screenshot 2023-11-10 at 10 19 36 AM

test/JDBC/input/BABEL-4424.sql
Deepesh125: Added

contrib/babelfishpg_tds/src/backend/tds/tdsprinttup.c
Deepesh125: I dont know how to produce such test cases. Does it make sense to safe guard it using assert?

Deepesh125: We have coverage for this.

test/JDBC/expected/BABEL-4424.out
Deepesh125: Already added

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
```diff
@@ -1782,7 +1894,9 @@ PrepareRowDescription(TupleDesc typeinfo, List *targetlist, int16 *formats,
 					 * than -1.
 					 */
 					if (atttypmod == -1 && tle != NULL)
-						atttypmod = resolve_numeric_typmod_from_exp((Node *) tle->expr);
+						atttypmod = plannedstmt ? 
+									resolve_numeric_typmod_from_exp(plannedstmt->planTree, (Node *) tle->expr) :
+									resolve_numeric_typmod_from_exp(NULL, (Node *) tle->expr);
```
surendravishnoi: here we are passing plan as NULL but have assert for plan inside resolve_numeric_typmod_from_exp?

contrib/babelfishpg_tds/src/backend/tds/tdsrpc.c
surendravishnoi: Either we should have exiting test case for this or we should add new. Even though it is same logic, we need to make sure some tests are going through this code path. We can only put assert if we are 100% sure that targetList is not NULL. Otherwise, we are making change without fully understanding all the use cases.

contrib/babelfishpg_tsql/src/hooks.c
surendravishnoi: why total digit count is dscale + 1? it means we cannot store more than 37 fraction part?

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
Deepesh125: Yes, it is handling of sp_prepare where we do not need to send numeric in anyway. only integer handle of prepare stmt needs to be sent.

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
```diff
@@ -1782,7 +1894,15 @@ PrepareRowDescription(TupleDesc typeinfo, List *targetlist, int16 *formats,
 					 * than -1.
 					 */
 					if (atttypmod == -1 && tle != NULL)
-						atttypmod = resolve_numeric_typmod_from_exp((Node *) tle->expr);
+					{
+						if (!plannedstmt)
+						{
+							ereport(ERROR,
```
surendravishnoi, line 1900: should test for both, plannedstmt and plannedstmt->planTree both should be not NULL. Also, in error message add context that plan stmt is NULL

Deepesh125: At the moment, we are not able to create test case which would hit this error.

contrib/babelfishpg_tds/src/backend/tds/tdsrpc.c
Deepesh125: Existing test cases are sufficient to test this

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
surendravishnoi: I mean check both plannedstmt and plannedstmt->planTree for error condition

surendravishnoi (approved):

PR #1950, opened 2023-10-25
Mark is_rolemember functions PARALLEL RESTRICTED

ahmed-shameem:
### Description

JDBC tests which involve IS_ROLEMEMBER function fails with parallel query enabled. This is because internally this function calls function like get_physical_user_name which in turn checks the database name. As this involves database state, so we should not mark IS_ROLEMEMBER function parallel safe. As per [postgres documentation](https://www.postgresql.org/docs/current/parallel-safety.html#:~:text=miscellaneous%20backend%2Dlocal%20state%20that%20the%20system%20cannot%20synchronize%20across%20workers), a function must be marked as PARALLEL RESTRICTED if there is a possibility of miscellaneous backend-local state that the system cannot synchronize across workers. As get_physical_user_name function is accessing the database name, there is a possibility that the parallel workers may not be able to synchronize the database name among themselves.

### Issues Resolved

BABEL-4481, BABEL-4421

* **Use case based -**

* **Boundary conditions -**

* **Arbitrary inputs -**

* **Negative test cases -**

* **Performance tests -**

* **Tooling impact -**

* **Client tests -**

### Check List


contrib/babelfishpg_tsql/sql/sys_functions.sql
```diff
@@ -3713,7 +3713,7 @@ CREATE OR REPLACE FUNCTION sys.is_rolemember_internal(
 	IN database_principal sys.SYSNAME
 )
 RETURNS INT AS 'babelfishpg_tsql', 'is_rolemember'
-LANGUAGE C STABLE PARALLEL SAFE;
+LANGUAGE C STABLE PARALLEL UNSAFE;
```
KushaalShroff: We need to add these in upgrade files as well, right?

ahmed-shameem: Updated in latest rev.

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--3.3.0--3.4.0.sql
```diff
@@ -2994,6 +2994,31 @@ $BODY$
 LANGUAGE plpgsql IMMUTABLE PARALLEL UNSAFE;
 GRANT EXECUTE ON FUNCTION sys.FORMAT(IN anyelement, IN NVARCHAR, IN VARCHAR) TO PUBLIC;
 
+-- Role member functions
+CREATE OR REPLACE FUNCTION sys.is_rolemember_internal(
```
surendravishnoi, line 4092: how we reached to the conclusion that database name cannot be sync between workers?

kuntalghosh: Yeah, worker will use the same role as the leader. So, we should be able to pushdown the rolename to parallel worker and make it parallel safe.

PR #1957, opened 2023-10-26
fix conversion of varchar to binary varbinary and vice versa

tanscorpio7:
### Description

Binary data type simply stores hex codes but when transformed to and from varchar, use of correct encoding becomes important. For example the symbol '™' is stored as
0xE284A2 in UTF-8 enc. while
0x99 in Win encoding
From users perspective they must see the hex value which is congruent to their server encoding.

To fix this we do necessary encoding to source data in varchar <--> varbinary internal functions.
Length checks should be done when data (hex string) is in server encoding.

We must also handle the case for string literal to binary types. So we explicitly call the varcharvarbinary conversion when a string literal is being casted to binary data type.

Also added CAST functions for sys.BBF_VARBINARY to sys.BBF_BINARY and vice versa
these cast are being used in geography and geometry cast functions
made necessary changes to geography and geometry casts as well
i.e.  CAST (CAST ($1 AS sys.VARCHAR) AS sys.bbf_varbinary) --> CAST ($1 AS sys.bbf_varbinary)

### Issues Resolved
[BABEL-1940]

### Sign-off
Signed-off-by: Tanzeel Khan [tzlkhan@amazon.com](mailto:tzlkhan@amazon.com)
Co-authored-by: Rohit Bhagat [rohitbgt@amazon.com](mailto:rohitbgt@amazon.com)

### Check List


sumitj824:
Please add tests.

contrib/babelfishpg_common/src/varbinary.c
```diff
@@ -636,20 +641,26 @@ varcharvarbinary(PG_FUNCTION_ARGS)
 						"varbinary is not allowed. Use the CONVERT function "
 						"to run this query.")));
 
-	/* If typmod is -1 (or invalid), use the actual length */
+	collInfo = lookup_collation_table(get_server_collation_oid_internal(false));
```
Deepesh125: We can have different collation on this source varchar column

test/JDBC/expected/BABEL_1940.out
```diff
@@ -0,0 +1,217 @@
...
+~~START~~
+varbinary
+3F
+43
+3F3F
+3F
+63
+3F
+3F
+3F
```
Deepesh125: This is not a right output

contrib/babelfishpg_common/src/varbinary.c
tanscorpio7: Collation does not affect the encoding unless we add the UTF8 option in collation ex: Japanese_Bushu_Kakusu_100_CS_AS_KS_WS_UTF8
[Reference Link](https://github.com/MicrosoftDocs/sql-docs/blob/live/docs/relational-databases/collations/collation-and-unicode-support.md#examples-of-collation-options)

test/JDBC/expected/BABEL_1940.out
tanscorpio7: We lack the mapping for this UTF8 byte sequence to WIN1252 enc.
To support this we will have to add the desired char in the mapping (in this case mapping could be LATIN CAPITAL LETTER C 0x43)

postgres=# select convert('Ｃ','UTF8','WIN1252');  
(note** this char is 'FULLWIDTH LATIN CAPITAL LETTER C' (U+FF23)
ERROR:  character with byte sequence 0xef 0xbc 0xa3 in encoding "UTF8" has no equivalent in encoding "WIN1252"

This lack of mapping could also be the root cause for BABEL-1916

test/JDBC/expected/BABEL_1940.out
```diff
@@ -0,0 +1,217 @@
...
+41
+61
+416243
+614263
+~~END~~
+
+
+SELECT CONVERT(VARBINARY(10), CONVERT(VARCHAR(10), CONVERT(VARCHAR(10), a))) FROM babel_1940_t2
```
sumitj824: How this test different from above? We should this like `CONVERT(VARCHAR(10), CONVERT(VARBINARY(10), a)))` and it should output the same values inserted in table.

tanscorpio7: Output will not be same since we do not have the required mapping for some characters. UTF8 chars are replaced with 0x3F ('?') when mapping does not exists in radix_tree. (this is happens for cast directly to varchar as well currently)
The best option for now seems to be removing these not mapped characters and do a
SELECT CONVERT(VARCHAR(10), CONVERT(VARBINARY(10), a)))
cc: @Deepesh125

contrib/babelfishpg_common/src/encoding/mb/conv.c
```diff
@@ -401,10 +401,6 @@ TsqlLocalToUtf(const unsigned char *iso, int len,
 		unsigned char b3 = 0;
 		unsigned char b4 = 0;
 
-		/* "break" cases all represent errors */
-		if (*iso == '\0')
-			break;
-
```
Deepesh125: As discussed last time, we need to handle this differently.

contrib/babelfishpg_tsql/src/pltsql_coerce.c
```diff
@@ -1100,6 +1102,34 @@ tsql_coerce_string_literal_hook(ParseCallbackState *pcbstate, Oid targetTypeId,
...
 		}
+		else if ((*common_utility_plugin_ptr->is_tsql_binary_datatype) (baseTypeId) ||
+				 (*common_utility_plugin_ptr->is_tsql_varbinary_datatype) (baseTypeId))
+		{
+			/*
+			 * binary datatype should be passed in client encoding
+			 * when explicit cast is called
+			 */
+
+			TypeName 	*varcharTypeName = makeTypeName("varchar");
```
Deepesh125: better to use `makeTypeNameFromNameList` with `sys` as schema

contrib/babelfishpg_tsql/src/pltsql_coerce.c
```diff
@@ -1100,6 +1102,34 @@ tsql_coerce_string_literal_hook(ParseCallbackState *pcbstate, Oid targetTypeId,
...
+			Const 		*tempcon;
+
+			typenameTypeIdAndMod(NULL, (const TypeName *)varcharTypeName, &baseTypeId, &baseTypeMod);
+			baseType = typeidType(baseTypeId);
+			pfree(varcharTypeName);
+
+			tempcon = makeConst(TEXTOID, -1,
+								tsql_get_server_collation_oid_internal(false),
+								-1, PointerGetDatum(cstring_to_text(value)),
+								false, false);
```
Deepesh125: why not sys.varchar?

test/JDBC/input/babel_function_string-before-15-5-or-14-10-vu-verify.sql
```diff
@@ -0,0 +1,292 @@
+-- test REPLICATE function
+SELECT REPLICATE(' abc ', 3)
```
Deepesh125, line 2: why do we need to add this test cases?

Deepesh125 (changes requested):

test/JDBC/input/babel_function_string-before-15-5-or-14-10-vu-verify.sql
tanscorpio7: These are old test cases, moved this to a new file because of version upgrade failure, line 149 in verify.out file

contrib/babelfishpg_common/src/encoding/mb/conv.c
tanscorpio7: Something like this ?

test/JDBC/expected/BABEL_1940.out
```diff
@@ -0,0 +1,224 @@
...
+
+
+-- 0x81 does not exist is empty in some encodings
+SELECT CONVERT(VARCHAR(10), 0x81)
+GO
+~~START~~
+varchar
+~~ERROR (Code: 33557097)~~
+
+~~ERROR (Message: character with byte sequence 0x81 in encoding "WIN1252" has no equivalent in encoding "UTF8")~~
```
Deepesh125: This error is probably not right behaviour, we should either replace it to '?' or '0x81' depending on where this error occurring

tanscorpio7: adding a missing flag to conversion functions ?

rohit01010:
```
1> SELECT CONVERT(VARBINARY(10), CONVERT(VARCHAR(10), 0x330033))
2> go
invalid byte sequence for encoding "UTF8": 0x00
```
Getting this above error with the changes from this PR, Kindly resolve and add this query to test-cases

contrib/babelfishpg_common/src/varbinary.c
```diff
@@ -636,20 +641,36 @@ varcharvarbinary(PG_FUNCTION_ARGS)
...
+	PG_TRY();
+	{
+		collInfo = lookup_collation_table(get_server_collation_oid_internal(false));
+		encoded_data = encoding_conv_util(data, len, PG_UTF8, collInfo.enc, &encodedByteLen);
+	}
+	PG_CATCH();
+	{
+		ereport(ERROR,
+			   (errcode(ERRCODE_INTERNAL_ERROR),
+				errmsg("Failed to convert from data type varchar to varbinary")));
```
tanscorpio7: Could happens when NULL byte exists in input varchar. Cannot think of any test case for this right now.

contrib/babelfishpg_common/src/varbinary.c
```diff
@@ -697,21 +718,45 @@ Datum
...
 	int32		maxlen = typmod - VARHDRSZ;
-	VarChar    *result;
+	coll_info	collInfo;
+	int			encodedByteLen;
 
 	/*
-	 * Cast the entire input binary data if maxlen is invalid or supplied data
-	 * fits it
+	 * Allow trailing null bytes 
+	 * Its safe since multi byte UTF-8 does not contain 0x00
```
Deepesh125, line 743: make this more clearer that it is applicable for binary --> var/char cast. And add example also in comment

contrib/babelfishpg_common/src/varbinary.c
```diff
@@ -697,21 +718,45 @@ Datum
...
+	 * Cast the entire input binary data if maxlen is 
+	 * invalid or supplied data fits it
+	 * Else truncate it
+	 */
+	PG_TRY();
+	{
+		if (maxlen < 0 || len <= maxlen)
+			encoded_result = encoding_conv_util(data, len, collInfo.enc, PG_UTF8, &encodedByteLen);
+		else
+			encoded_result = encoding_conv_util(data, maxlen, collInfo.enc, PG_UTF8, &encodedByteLen);
```
Deepesh125, line 763: what would happen when `len = 0`? Can we add some test for the same?

contrib/babelfishpg_common/src/varbinary.c
```diff
@@ -697,21 +718,45 @@ Datum
...
+	}
+	PG_CATCH();
+	{
+		ereport(ERROR,
+			   (errcode(ERRCODE_INTERNAL_ERROR),
+				errmsg("Failed to convert from data type varbinary to varchar")));
```
Deepesh125: we somehow need to preserve the original error as log or error in the logfile

test/JDBC/expected/BABEL_1940.out
```diff
@@ -0,0 +1,242 @@
+
```
Deepesh125, line 1: Basic test cases mentioned in the Jira is missing.

Deepesh125 (changes requested):

contrib/babelfishpg_common/src/varbinary.c
tanscorpio7: empty varchar string returned ("")

contrib/babelfishpg_common/src/varbinary.c
rohit01010: can u add tests for this, where collation of source varchar column is different than server collation.

test/JDBC/expected/BABEL_1940.out
tanscorpio7: Added one, everything else is covered

contrib/babelfishpg_common/src/varbinary.c
tanscorpio7: throwing the original error alongside are our own.

contrib/babelfishpg_common/src/varbinary.c
tanscorpio7: added test case

contrib/babelfishpg_common/src/varbinary.c
```diff
@@ -636,20 +642,45 @@ varcharvarbinary(PG_FUNCTION_ARGS)
...
+		ereport(ERROR,
+			   (errcode(ERRCODE_INTERNAL_ERROR),
+				errmsg("Failed to convert from data type varchar to varbinary, %s",
+				errorData->message)));
+	}
+	PG_END_TRY();
+
+	/* 
+	 * If typmod is -1 (or invalid), use the actual length
+	 * Length should be checked after encoding into server enc
```
Deepesh125: server enc --> server encoding

contrib/babelfishpg_common/src/varbinary.c
```diff
@@ -697,21 +728,60 @@ Datum
...
 	 */
-	if (maxlen < 0 || len <= maxlen)
-		result = (VarChar *) cstring_to_text_with_len(data, len);
-	/* Else truncate it */
-	else
-		result = (VarChar *) cstring_to_text_with_len(data, maxlen);
+	while(len>0 && data[len-1] == '\0')
+		len -= 1;
+	
+	collInfo = lookup_collation_table(get_server_collation_oid_internal(false));
```
Deepesh125: include this inside pg_try

Deepesh125 (dismissed):

Deepesh125 (approved):

contrib/babelfishpg_common/sql/binary.sql
```diff
@@ -46,6 +46,12 @@ CREATE TYPE sys.BBF_BINARY (
     COLLATABLE     = false
 );
 
+CREATE CAST (sys.BBF_BINARY AS sys.BBF_VARBINARY)
+WITHOUT FUNCTION AS IMPLICIT;
+
+CREATE CAST (sys.BBF_VARBINARY AS sys.BBF_BINARY)
+WITHOUT FUNCTION AS IMPLICIT;
+
```
KushaalShroff, line 54: No upgrade script changes for these?

PR #1954, opened 2023-10-25
Change DB owner with ALTER AUTHORIZATION

robverschoor:
### Description

Implement `ALTER AUTHORIZATION` to change database ownership. 
Syntax: `ALTER AUTHORIZATION ON DATABASE::<dbname> TO <login-name>`
It is a common requirement for the DBA to create a database to be owned by a login. Thus far in Babelfish this was possible only by (temporarily)  making the target login a member of `sysadmin` role, which is undesirable.
To change the database owner to a different login, all that is required is to update `sys.babelfish_sysdatabases.owner` to the new owner's login name. There are some restrictions such as that the new owner cannot be a user in the database already (`guest` user does not count). a non-`sysadmin` DB owner cannot grant ownership to anyone else, including to a `sysadmin` login. (however, it is OK to grant ownership to yourself although there seems little point in that).
When the current DB owner has an active session in the database with the current DB set to that same database, and at the same time a different session change the database ownership, the now-previous owner retains access rights in the session as long as the database context remains unchanged; as soon as the database context is changed, the access rights from the DB ownership are lost: Babelfish behaves identically to T-SQL here.
To implement this feature, statement type `PLTSQL_STMT_CHANGE_DBOWNER` was introduced rather than something more generic-looking like `PLTSQL_STMT_ALTERAUTH`: other cases of `ALTER AUTHORIZATION` may affect the ownership/permissions-related artefacts that are created by Babelfish, or require different catalog modifications; therefore, the DB ownership change is more of a one-off case. 

Signed-off-by: Rob Verschoor [rcv@amazon.com](mailto:rcv@amazon.com)

### Issues Resolved

BABEL-2121: Babel does not support ALTER AUTHORIZATION syntax to change database owner

## Test Scenarios Covered

- Use case based - Yes

- Boundary conditions - Yes

- Arbitrary inputs - N/A

- Negative test cases - Yes

- Minor version upgrade tests - N/A

- Major version upgrade tests - N/A

- Performance tests - N/A

- Tooling impact - N/A

- Client tests - N/A

### Check List

- [x] Commits are signed per the DCO using --signoff 

By submitting this pull request, I confirm that my contribution is under the terms of the Apache 2.0 and PostgreSQL licenses, and grant any person obtaining a copy of the contribution permission to relicense all or a portion of my contribution to the PostgreSQL License solely to contribute all or a portion of my contribution to the PostgreSQL open source project.

[...]

PR #1962, opened 2023-10-26
Implement sp_changedbowner

robverschoor:
### Description

Implement `sp_changedbowner`.
`sp_changedbowner` is the 'old' way of changing the owner of a database, but still supported in T-SQL.
The difference with `ALTER AUTHORIZATION ON DATABASE::dbname TO login_name` [(here)](https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/1954) is that `sp_changedbowner` applies to the current database only.
`sp_changedbowner` is implemented as a wrapper around `ALTER AUTHORIZATION ON DATABASE::`, and most validations are handed off to `ALTER AUTHORIZATION`.
The test cases are mostly a subset as well.

### Issues Resolved

BABEL-4504: Babel does not support procedure "sp_changedbowner"

- Use case based - Yes

- Boundary conditions - Yes

- Arbitrary inputs - N/A

- Negative test cases - Yes

- Minor version upgrade tests - Yes

- Performance tests - N/A

- Tooling impact - N/A

- Client tests - N/A

### Check List


test/JDBC/expected/sp_changedbowner-vu-verify.out
```diff
@@ -0,0 +1,404 @@
...
+select name, suser_sname(sid) from sysdatabases where name = 'msdb'
+go
+~~START~~
+text#!#nvarchar
+msdb#!#new_OWNER_login_sp
+~~END~~
+
+~~START~~
+text#!#nvarchar
+msdb#!#jdbc_user
```
lejaokri: test could fail in our internal repo because default owner may not be `jdbc_user` anymore when we run it on internal gitfarm

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--3.3.0--3.4.0.sql
```diff
@@ -3126,4 +3126,39 @@ CALL sys.analyze_babelfish_catalogs();
...
+	DECLARE @cmd sys.NVARCHAR(300)
+	DECLARE @db  sys.sysname = DB_NAME()
+
+	-- For a NULL login name, do nothing
+	IF @loginame IS NULL
+	BEGIN
+		RETURN
+	END
+
+	IF (@db = 'master') OR (@db = 'tempdb')
```
lejaokri, line 3150: I am wondering why not add `model` and `distribution` database names in the check

robverschoor: Babelfish has not implemented model and distribution. Will adjust the error message to remove those, even though the original T-SQL error msg mentions these, but removing them is more clear for the user.

test/JDBC/expected/sp_changedbowner-vu-verify.out
robverschoor: will adjust

contrib/babelfishpg_tsql/sql/babelfishpg_tsql.sql
```diff
@@ -3617,3 +3617,40 @@ BEGIN
...
+
+	IF SUSER_ID(@loginame) IS NULL
+	BEGIN
+		RAISERROR('Cannot find the principal ''%s'', because it does not exist or you do not have permission.', 16, 1, @loginame)
+		RETURN
+	END
+
+	-- Compose the ALTER ATHORIZATION statement:
+	SET @cmd = 'ALTER AUTHORIZATION ON DATABASE::[' + @db + '] TO [' + SUSER_NAME(SUSER_ID(@loginame)) + ']'
+	EXECUTE(@cmd)
```
HarshLunagariya, line 3652: Can we avoid doing this, as it is vulnerable to SQL injection? IMO better approach would be to do it via C implementation.

HarshLunagariya: Although here I don't such risk because we are not escalating the privileges in the execution internally. But still pointing it out.

lejaokri (approved):

contrib/babelfishpg_tsql/sql/babelfishpg_tsql.sql
robverschoor: The only place where injection could happen is the parameter `@loginame`, and that gets passed through `SUSER_ID()` before being used on line 3651: first it is verified that the login actually exists, and on the line where the final statement is constructed, `@loginame` itself is not used but it is rather passed through `SUSER_NAME(SUSER_ID())`. I think this stops injection attempts.

HarshLunagariya (approved):

contrib/babelfishpg_tsql/sql/babelfishpg_tsql.sql
Deepesh125: Another perk of having it implemented in C is performance. In this impl., We access one or two catalog before we constructing and executing ALTER AUTHORIZATION equivalent syntax. I would rather implement this directly in C as a wrapper around update_db_owner (implemented in recent PR)

robverschoor: I do not disagree with the principle, but this is not a performance-sensitive operation. This procedure is typically called only once manually or in a script by a DBA when a database is created, i.e. once in the lifetime of a database.

robverschoor: Note that this proc is the 'old' way of doing things (i.e. Sybase-era), the reason for supporting it is just for completeness, i.e. for those folks who stick to their old habits. But basically I expect almost everyone will use ALTER AUTHORIZATION anyway. The rationale for implementing this in SQL is that this was just quicker to implement (less coding) since it's a very thin wrapper around ALTER AUTHORIZATION ON DATABASE::, and for a very limited use case.

PR #1970, opened 2023-11-01
Fix Crash when Instead of View Trigger calls itself

deepakshi-mittal:
### Description

This change will fix crashes when an
Instead of Trigger on View calls itself directly and indirectly
Also fix crash IO view trigger inserted is used with join (BABEL-4514)

Task: BABEL-2170

Engine PR: https://github.com/babelfish-for-postgresql/postgresql_modified_for_babelfish/pull/245

### Issues Resolved

BABEL-2170

* **Use case based -**
- IOT Trigger A on view calls Trigger A
- fix crash IO view trigger inserted is used with join (BABEL-4514)

### Check List
- [x ] Commits are signed per the DCO using --signoff 


contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -720,6 +720,28 @@ static bool
...
 	TriggerDesc *trigDesc = view->trigdesc;
+	if (triggerOids)
+	{
+		int i;
+		for (i = 0; i < trigDesc->numtriggers; i++)
+		{
+			Trigger *trigger = &trigDesc->triggers[i];
+			Oid current_tgoid = trigger->tgoid;
+			Oid prev_tgoid = InvalidOid;
+			prev_tgoid = lfirst_oid(list_tail(triggerOids));
```
forestkeeper, line 747: Where is the place add oid into triggerOids ?

deepakshi-mittal: Added in backend code
https://github.com/babelfish-for-postgresql/postgresql_modified_for_babelfish/pull/245/files
src/backend/commands/trigger.c

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -720,6 +720,28 @@ static bool
...
+			if (prev_tgoid == current_tgoid)
+			{
+				return false; /** Direct recursive trigger case*/
+			}
+			else if (list_member_oid(triggerOids, current_tgoid))
+            {
+                /** Indirect recursive trigger case*/
+                ereport(ERROR,
```
lejaokri, line 755: Test this codepath. 

Using the following
```
--setup
create table tab_base ( a int, b int )
go
create view vw_get_base as select * from tab_base
go
create table tab_base2 ( a int, b int )
go
create view vw_get_base2 as select * from tab_base2
go
create trigger tr_vw_get_base_instead_insert
on vw_get_base
INSTEAD OF INSERT
AS
INSERT INTO vw_get_base2 SELECT a, b+1000000 FROM INSERTED
go
create trigger tr_vw_get_base_instead_insert2
on vw_get_base2
INSTEAD OF INSERT
AS
INSERT INTO vw_get_base SELECT a, b+10000 FROM INSERTED
go

```

```
-- error case 

-- Next command will throw error as expected.
-- By the time we get to pltsql_bbfViewHasInsteadofTrigger, triggerOids could have more than one entry.
-- And one of the entries is current_tgoid, so we error out and do error handling
-- But the error handling in ExecCallTriggerFunc will only remove one item from the array.
insert into vw_get_base2 values (2000,1) 
GO

-- This one will run with non-emptry triggerOid because you only removed one from the previous error cleanup.
-- I dont think that's correct.
 insert into vw_get_base values (2000,1) 
GO
```

deepakshi-mittal: I checked this, the triggerOids is being cleaned up properly. The issue is not because of triggerOids variable. Its because the second query which crashes doesn't calls RewriteHandler.

lejaokri: From what I can see, on the first run which is `insert into vw_get_base2 values (2000,1) `, it goes to line 4706 in `exec_stmt_execsql` because `expr->plan == NULL. But on the second run, `expr-plan` is not NULL anymore so it skips the if block

forestkeeper (approved):

PR #1984, opened 2023-11-06
Explicit unnecessary cast should be eliminated

forestkeeper:
This commit can eliminate unnecessary explicit cast in where condition and tune the query to choose index in those conditions

Task: BABEL-4176

### Description

Previously, when using explicit casting, PG will not try to choose index scan. For example : 
```
select * from test where cast(id as bigint) = 1;
```
Even the id is defined as int, this query will not choose index scan on id index.

But for this explicit cast, it can be eliminate in certain circumstances like : 
```
        qual  in where condition

                                     \ 

                                '=' operator

                                     /        \
                cast(var( int )to bigint )       anything

```
should be optimized to 
```
        qual  in where condition

                              \ 

                                '=' operator

                               /        \

                              var       anything
```
The reason is an explicit cast from int4 to int8 for a var, won't change anything in the result, and eliminate that explicit cast would give a chance to choose the right index in optimization.

### Issues Resolved

When encountered a query structure like : 
```
        qual  in where condition

                                     \ 

                                '=' operator

                                     /        \

                cast(var( int )to bigint )       anything
```
This feature would eliminate the explicit cast : 
```
        qual  in where condition

                              \ 

                                '=' operator

                               /        \

                              var       anything
```
To make it possible to use the index scan instead of seq scan.

### Check List


contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -4453,3 +4464,54 @@ transform_pivot_clause(ParseState *pstate, SelectStmt *stmt)
...
+	{
+		OpExpr *opExpr = (OpExpr*) node;
+		Form_pg_operator form;
+		HeapTuple	tuple;
+		Node* node = optimize_explicit_cast(pstate, linitial(opExpr->args));
+		if (node != linitial(opExpr->args))
+		{
+			char	   *opname;
+
+			tuple = SearchSysCache1(OPEROID, ObjectIdGetDatum(opExpr->opno));
```
lejaokri, line 4505: memleak : `ReleaseSysCache(tuple)` before returning

forestkeeper: Thanks for reminding

lejaokri (approved):

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -4453,3 +4464,58 @@ transform_pivot_clause(ParseState *pstate, SelectStmt *stmt)
...
+static Node* optimize_explicit_cast(ParseState *pstate, Node *node)
+{
+	if (sql_dialect != SQL_DIALECT_TSQL)
+		return node;
+	if (node == NULL)
+		return NULL;
+	if (IsA(node, FuncExpr))
+	{
+		FuncExpr   *f = (FuncExpr *) node;
+		if (f->funcformat == COERCE_EXPLICIT_CAST){
```
2jungkook, line 4477: Please double check this `funcformat` is only for `CAST` function and no other functions, e.g., UDF, can have this.

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -4453,3 +4464,58 @@ transform_pivot_clause(ParseState *pstate, SelectStmt *stmt)
...
+			Node* arg = (Node*)linitial(f->args);
+			if (nodeTag(arg) == T_Var && 
+			   ((Var*) arg)->vartype == INT4OID && 
+			   f->funcresulttype == INT8OID)
```
2jungkook, line 4481: I'd like to cover all INT-related scenario in this PR, e.g., INT2 -> INT4.

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -4453,3 +4464,58 @@ transform_pivot_clause(ParseState *pstate, SelectStmt *stmt)
...
+				return optimize_explicit_cast(pstate, linitial(f->args));
```
2jungkook, line 4482: We can just return `linitial(f->args)`.

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -4453,3 +4464,58 @@ transform_pivot_clause(ParseState *pstate, SelectStmt *stmt)
...
+	}
+	else if (IsA(node, BoolExpr))
+	{
+		BoolExpr *r = (BoolExpr *) node;
+		ListCell *l;
+		foreach (l , r->args)
+		{
+			optimize_explicit_cast(pstate, (Node*)lfirst(l));
+		}
+	}
```
2jungkook, line 4493: Please add tests for this logic, e.g., `AND/OR` clause with composite indexes.

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -4453,3 +4464,58 @@ transform_pivot_clause(ParseState *pstate, SelectStmt *stmt)
...
 	MemoryContextSwitchTo(oldContext);
 }
+
+static Node* optimize_explicit_cast(ParseState *pstate, Node *node)
+{
+	if (sql_dialect != SQL_DIALECT_TSQL)
+		return node;
+	if (node == NULL)
+		return NULL;
+	if (IsA(node, FuncExpr))
```
2jungkook, line 4474: Please also add tests for nested CAST case even if we don't support this yet.

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -4453,3 +4464,58 @@ transform_pivot_clause(ParseState *pstate, SelectStmt *stmt)
...
+		{
+			optimize_explicit_cast(pstate, (Node*)lfirst(l));
+		}
+	}
+	else if (IsA(node, OpExpr))
+	{
+		OpExpr *opExpr = (OpExpr*) node;
+		Form_pg_operator form;
+		HeapTuple	tuple;
+		Node* node = optimize_explicit_cast(pstate, linitial(opExpr->args));
```
2jungkook, line 4499: We also need a test case for ` 1 = CAST(ROID AS BIGINT)`.

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -4453,3 +4464,58 @@ transform_pivot_clause(ParseState *pstate, SelectStmt *stmt)
...
+		{
+			char	   *opname;
+
+			tuple = SearchSysCache1(OPEROID, ObjectIdGetDatum(opExpr->opno));
+			if (!HeapTupleIsValid(tuple))
+				return node; // no need to error out in here, stop transform and quite, keep the original node
+			form = (Form_pg_operator) GETSTRUCT(tuple);
+
+			opname = NameStr(form->oprname);
+			if (strcmp(opname, "=") == 0)
```
2jungkook, line 4511: Let's add a comment that currently we only remove unnecessary cast for equal op.

test/JDBC/expected/cast_eliminate-vu-prepare.out
```diff
@@ -0,0 +1,6 @@
+CREATE TABLE cast_eliminate( [ROID] [int] NOT NULL PRIMARY KEY );
+GO
+
+CREATE TABLE cast_eliminate2( [ROID] [bigint] NOT NULL PRIMARY KEY );
```
2jungkook, line 4: Let's create views and see whether we can remove unnecessary casts after MVU.

test/JDBC/expected/cast_eliminate-vu-verify.out
```diff
@@ -0,0 +1,82 @@
+set babelfish_showplan_all on
+GO
+
+SELECT 1 AS [C1] FROM cast_eliminate  WHERE (CAST(ROID AS BIGINT) = 1)
+GO
+~~START~~
+text
+Query Text: SELECT 1 AS [C1] FROM cast_eliminate  WHERE (CAST(ROID AS BIGINT) = 1)
+Index Only Scan using cast_eliminate_pkey on cast_eliminate  (cost=0.15..8.17 rows=1 width=4)
```
2jungkook, line 9: Let's remove costs from output because it can vary across different versions. Ref: https://docs.aws.amazon.com/AmazonRDS/latest/AuroraUserGuide/working-with-babelfish-usage-notes-features.using.explain.html

test/JDBC/expected/cast_eliminate-vu-verify.out
```diff
@@ -0,0 +1,82 @@
+set babelfish_showplan_all on
```
2jungkook, line 1: Before this, let's do `analyze` manually because auto analyze can introduce flaky results.

test/JDBC/expected/cast_eliminate-vu-verify.out
```diff
@@ -0,0 +1,82 @@
...
+GO
+
+SELECT 1 AS [C1] FROM cast_eliminate  WHERE (CAST(ROID AS BIGINT) = 1)
```
2jungkook, line 4: To make out test suite more solid, I'd suggest to have the following test cases as well:
1. A case where cast should not be removed, e.g., BIGINT -> INT
2. A case where the right side of `=` is actually a bigint, e.g., CAST(ROID AS BIGINT) = (a real bigint number)

test/JDBC/expected/cast_eliminate-vu-verify.out
```diff
@@ -0,0 +1,82 @@
...
+
+SELECT 1 AS [C1] FROM cast_eliminate  WHERE (CAST(ROID AS BIGINT) = cast( 1 as bigint ))
+GO
+~~START~~
+text
+Query Text: SELECT 1 AS [C1] FROM cast_eliminate  WHERE (CAST(ROID AS BIGINT) = cast( 1 as bigint ))
+Index Only Scan using cast_eliminate_pkey on cast_eliminate  (cost=0.15..8.17 rows=1 width=4)
+  Index Cond: (roid = '1'::bigint)
+~~END~~
+
```
2jungkook, line 82: Not necessary, but it's better to do `reset babelfish_showplan_all`.

test/JDBC/parallel_query_jdbc_schedule
```diff
@@ -54,6 +54,7 @@ ignore#!#BABEL-4261
 ignore#!#babel_collection
 ignore#!#binary-index-vu-verify
 ignore#!#pgr_select_into
+ignore#!#cast_eliminate-vu-verify
```
2jungkook, line 57: What happens if we include this?

PR #1986, opened 2023-11-06
Support CREATE/DROP FULLTEXT INDEX for fulltext search 

roshan0708:
### Description

Currently babelfish doesn't support fulltext index creation and drop for full text search. This commit implements the logic of create and drop fulltext index and also handles the unsupported features of the CREATE/DROP FULLTEXT INDEX statements. 

Signed-off-by: Roshan Kanwar [rskanwar@amazon.com](mailto:rskanwar@amazon.com)

### Issues Resolved

Task: BABEL-4383

* **Use case based -** Yes

* **Arbitrary inputs -**

* **Performance tests -**

* **Tooling impact -**

* **Client tests -**

### Check List


contrib/babelfishpg_tsql/src/guc.c
```diff
@@ -1336,13 +1336,13 @@ define_escape_hatch_variables(void)
...
 	DefineCustomEnumVariable("babelfishpg_tsql.escape_hatch_fulltext",
-							 gettext_noop("escape hatch for fulltext"),
+							 gettext_noop("escape hatch for fulltext search"),
 							 NULL,
 							 &escape_hatch_fulltext,
 							 EH_STRICT,
 							 escape_hatch_options,
-							 PGC_USERSET,
-							 GUC_NOT_IN_SAMPLE | GUC_DISALLOW_IN_FILE | GUC_DISALLOW_IN_AUTO_FILE,
+							 PGC_SUSET,
```
surendravishnoi: what is the purpose of this change?

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -3658,6 +3661,16 @@ get_insert_bulk_kilobytes_per_batch()
 	return insert_bulk_kilobytes_per_batch;
 }
 
+static const char 
+*gen_schema_name_for_fulltext_index(char *schema_name)
```
surendravishnoi: should be check for schema name being NULL?

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -3817,3 +3830,110 @@ exec_stmt_change_dbowner(PLtsql_execstate *estate, PLtsql_stmt_change_dbowner *s
...
+	schemaOid = LookupExplicitNamespace(schema_name, true);						
+	table_name = stmt->table_name;
+	column_name = stmt->column_name;
+	index_name = stmt->index_name;
+
+	// check if schema exists
+	if (!OidIsValid(schemaOid))
+			ereport(ERROR,
+				(errcode(ERRCODE_UNDEFINED_SCHEMA),
+					errmsg("schema \"%s\" does not exist",
```
surendravishnoi: check for null schema to be on safe side

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -3817,3 +3830,110 @@ exec_stmt_change_dbowner(PLtsql_execstate *estate, PLtsql_stmt_change_dbowner *s
...
+
+	// check if table exists
+	if (!OidIsValid(relid))
+			ereport(ERROR,
+				(errcode(ERRCODE_UNDEFINED_TABLE),
+					errmsg("relation \"%s\" does not exist",
+						table_name)));
+
+	// check if fulltext index already exists in the table
+	if (!get_index_name(relid))
```
surendravishnoi: this can be combined with if else
if (get_index_name(relid)) error else get query str

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -3817,3 +3830,110 @@ exec_stmt_change_dbowner(PLtsql_execstate *estate, PLtsql_stmt_change_dbowner *s
...
+		index_exists = false;
+
+	if (!index_exists)
+		query_str = gen_createfulltextindex_cmds(table_name, schema_name, column_name, index_name);
+	else
+		ereport(ERROR,
+					(errcode(ERRCODE_WRONG_OBJECT_TYPE),
+					 errmsg("A full-text index for table or indexed view \"%s\" has already been created."
```
surendravishnoi: also print index name (mostly try to match what error SQL server will give)

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -3817,3 +3830,110 @@ exec_stmt_change_dbowner(PLtsql_execstate *estate, PLtsql_stmt_change_dbowner *s
...
+	const char		*schema_name;
+	Oid 		schemaOid;
+	Oid			relid;
+	char		*index_name;
+	char		*table_name;
+	char		*query_str;
+	bool		index_exists = true;
+
+	table_name = stmt->table_name;
+	schema_name = gen_schema_name_for_fulltext_index((char *)stmt->schema_name);
```
surendravishnoi: check for null schema

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -3817,3 +3830,110 @@ exec_stmt_change_dbowner(PLtsql_execstate *estate, PLtsql_stmt_change_dbowner *s
...
+	/* The above query will be
+	 * executed using ProcessUtility()
+	 */
+	exec_utility_cmd_helper(query_str);
+
+	return PLTSQL_RC_OK;
+}
+
+static int	
+exec_stmt_dropfulltextindex(PLtsql_execstate *estate, PLtsql_stmt_dropfulltextindex *stmt)
```
surendravishnoi: we can handle both in a single function with a boolean to distinguish create vs drop, most of the code is duplicate

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -3817,3 +3830,110 @@ exec_stmt_change_dbowner(PLtsql_execstate *estate, PLtsql_stmt_change_dbowner *s
...
+	relid = get_relname_relid((const char *) table_name, schemaOid);
+
+	// check if table exists
+	if (!OidIsValid(relid))
+			ereport(ERROR,
+				(errcode(ERRCODE_UNDEFINED_TABLE),
+					errmsg("relation \"%s\" does not exist",
+						table_name)));
+				
+	index_name = get_index_name(relid);
```
surendravishnoi: why this assumption that get_index_name will only return full text index? what about other indexes on table? If this function is only for full text index, it should be named accordingly

contrib/babelfishpg_tsql/src/pltsql_utils.c
```diff
@@ -1791,3 +1791,78 @@ List
...
+
+	relation = RelationIdGetRelation(relid);
+	indexoidlist = RelationGetIndexList(relation);
+	if (list_length(indexoidlist) == 0)
+	{
+		RelationClose(relation);
+		list_free(indexoidlist);
+		return NULL;
+	}
+	indexOid = linitial_oid(indexoidlist);
```
surendravishnoi: why are we assuming that first index is full text index only?

contrib/babelfishpg_tsql/src/pltsql_utils.c
```diff
@@ -1791,3 +1791,78 @@ List
...
+	index_name = get_rel_name(indexOid);
+	RelationClose(relation);
+	list_free(indexoidlist);
+	return index_name;
+}
+
+char
+*gen_createfulltextindex_cmds(const char *table_name, const char *schema_name, const List *column_name, const char *index_name)
```
surendravishnoi, line 1947: Please make sure column_name list and index name is not null or we handle null values

contrib/babelfishpg_tsql/src/pltsql_utils.c
```diff
@@ -1791,3 +1791,78 @@ List
...
+{
+	StringInfoData query;
+	initStringInfo(&query);
+	/*
+	 * We prepare the following query to drop a fulltext index.
+	 *
+	 * DROP INDEX <index_name>
+	 *
+	 */
+	appendStringInfo(&query, "DROP INDEX \"%s\".\"%s\"", schema_name, index_name);
```
surendravishnoi: why no check for schema name being NULL here? also, is it guaranteed that index name is NULL across database?

contrib/babelfishpg_tsql/src/tsqlIface.cpp
```diff
@@ -1802,13 +1820,11 @@ class tsqlBuilder : public tsqlCommonMutator
...
 			nop = post_process_create_type(ctx->create_type(), stmt, ctx);
-		else if (ctx->create_fulltext_index() ||
-		         ctx->alter_fulltext_index() ||
-		         ctx->drop_fulltext_index())
+		else if (ctx->alter_fulltext_index())
 		{
 			ereport(WARNING,
 					(errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
-					 errmsg("FULLTEXT related statements will be ignored.")));
+					 errmsg("ALTER CREATE FULLTEXT INDEX statement will be ignored.")));
```
surendravishnoi: ALTER FULLTEXT rather than ALTER CREATE FULLINDEX

contrib/babelfishpg_tsql/src/tsqlIface.cpp
```diff
@@ -6580,6 +6596,106 @@ post_process_alter_table(TSqlParser::Alter_tableContext *ctx, PLtsql_stmt_execsq
 	return false;
 }
 
+PLtsql_stmt*
```
surendravishnoi: Indetnantion is messed up

contrib/babelfishpg_tsql/src/tsqlIface.cpp
```diff
@@ -6580,6 +6596,106 @@ post_process_alter_table(TSqlParser::Alter_tableContext *ctx, PLtsql_stmt_execsq
...
+				char *column_name = pstrdup(downcase_truncate_identifier(column_name_str.c_str(), column_name_str.length(), true));
+				column_name_list = lappend(column_name_list, column_name);
+			}
+
+        }
+		stmt->column_name = column_name_list;
+    }
+	if (ctx->catalog_filegroup_option()) {
+		auto catalog_filegroup_option = ctx->catalog_filegroup_option();
+		if (catalog_filegroup_option->FILEGROUP() || catalog_filegroup_option->catalog_name || catalog_filegroup_option->filegroup_name)
```
surendravishnoi: which filegroup option we support?

contrib/babelfishpg_tsql/src/tsqlIface.cpp
```diff
@@ -6580,6 +6596,106 @@ post_process_alter_table(TSqlParser::Alter_tableContext *ctx, PLtsql_stmt_execsq
...
+			throw PGErrorWrapperException(ERROR, ERRCODE_FEATURE_NOT_SUPPORTED, "FILEGROUP/CATALOG option is not supported yet for CREATE FULLTEXT INDEX statement", getLineAndPos(ctx));
+	}
+	if (ctx->fulltext_with_option().size() > 0) {
```
surendravishnoi: which with options we support?

contrib/babelfishpg_tsql/src/tsqlIface.cpp
```diff
@@ -6580,6 +6596,106 @@ post_process_alter_table(TSqlParser::Alter_tableContext *ctx, PLtsql_stmt_execsq
...
+    return (PLtsql_stmt *) stmt;
+}
+
+PLtsql_stmt*
+makeDropFulltextIndexStmt(TSqlParser::Drop_fulltext_indexContext *ctx)
+{
+	PLtsql_stmt_dropfulltextindex *stmt = (PLtsql_stmt_dropfulltextindex *) palloc0(sizeof(PLtsql_stmt_dropfulltextindex));
+	stmt->cmd_type = PLTSQL_STMT_DROPFULLTEXTINDEX;
+	stmt->lineno = getLineNo(ctx);
+	if (ctx->table_name())
```
surendravishnoi, line 6763: getting table name could be a utility function shared between both create/drop

contrib/babelfishpg_tsql/src/tsqlUnsupportedFeatureHandler.cpp
```diff
@@ -1057,6 +1057,38 @@ antlrcpp::Any TsqlUnsupportedFeatureHandlerImpl::visitDdl_statement(TSqlParser::
 		if (create_user->WITHOUT())
 			handle(INSTR_UNSUPPORTED_TSQL_UNKNOWN_DDL, "CREATE USER WITHOUT LOGIN",  getLineAndPos(ctx));
 	}
+	if(ctx->create_fulltext_index())
```
surendravishnoi: why we have same error handling at two places?

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -3817,3 +3830,110 @@ exec_stmt_change_dbowner(PLtsql_execstate *estate, PLtsql_stmt_change_dbowner *s
...
+			ereport(ERROR,
+				(errcode(ERRCODE_UNDEFINED_SCHEMA),
+					errmsg("schema \"%s\" does not exist",
+						stmt->schema_name)));
+
+	relid = get_relname_relid((const char *) table_name, schemaOid);
+
+	// check if table exists
+	if (!OidIsValid(relid))
+			ereport(ERROR,
```
surendravishnoi: generic comment
1) We should have test case for each error we are throwing, in executor or ANTLR
2) We should have test case for each condition if/else we have. For example : creating full text index on one column vs multiple column

contrib/babelfishpg_tsql/src/guc.c
roshan0708: This is to only allow superuser to enable full text search, initially we were using escape hatch which the customer can use to enable fts anytime which we don’t want. After pentest, if everything goes right, we will remove this.

contrib/babelfishpg_tsql/src/pl_exec-2.c
roshan0708: That is already handled in tsqlIface file, while I’m extracting other data items.

Sairakan: Just for sanity, I think schema_name should be declared as const here, e.g. `const char *schema_name`

contrib/babelfishpg_tsql/src/tsqlUnsupportedFeatureHandler.cpp
Sairakan: +1, if we already are catching unsupported features here, it makes no sense to have the same error handling in the main handler. I would remove the handling in the main function and leave the error handling here.

contrib/babelfishpg_tsql/src/pl_exec-2.c
roshan0708: In the query statement, if we have not specified the schema name, it will be an empty string i.e. `""` else we get the proper schema name in the helper function. So the schema name will never be null. Here is the below code for reference - 
```
if (ctx->table_name())
{
	std::string table_info = ::getFullText(ctx->table_name());
	std::string table_name = "";
	std::string schema_name = "";
	size_t pos = table_info.find(".");
	if (pos != std::string::npos) {
		// Extract the schema name before the "."
		schema_name = table_info.substr(0, pos);
		// Extract the table name after the "."
		table_name = table_info.substr(pos + 1);
	} else {
		// No "." character found, set first to the entire string
		table_name = table_info;
	}
	stmt->table_name = pstrdup(downcase_truncate_identifier(table_name.c_str(), table_name.length(), true));
	stmt->schema_name = pstrdup(downcase_truncate_identifier(schema_name.c_str(), schema_name.length(), true));
}
```

contrib/babelfishpg_tsql/src/tsqlIface.cpp
roshan0708: We don't support any filegroup options now.

contrib/babelfishpg_tsql/src/tsqlIface.cpp
roshan0708: We don't support any with options now.

contrib/babelfishpg_tsql/src/tsqlUnsupportedFeatureHandler.cpp
roshan0708: Will do

contrib/babelfishpg_tsql/src/pl_exec-2.c
roshan0708: This is specifically for fulltext index, will rename it.

contrib/babelfishpg_tsql/src/pltsql_utils.c
roshan0708: Because there can be only one fulltext index per table acc. to SQL Server docs. We get a list of indexes, so we assume we get the fulltext index in the first item only.

contrib/babelfishpg_tsql/src/pl_exec-2.c
roshan0708: Yes, I've added those test cases in the JDBC test files of full text index.

contrib/babelfishpg_tsql/src/pltsql_utils.c
roshan0708: Done

contrib/babelfishpg_tsql/src/pltsql_utils.c
roshan0708: We already handle null values, so if we pass no column or NULL as index name in the query statement, ANTLR will automatically throw `syntax error`.

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -3662,7 +3661,7 @@ get_insert_bulk_kilobytes_per_batch()
 }
 
 static const char 
-*gen_schema_name_for_fulltext_index(char *schema_name)
+*gen_schema_name_for_fulltext_index(const char *schema_name)
```
Sairakan: I'm not clear on why this function exists separately from other fully qualified schema name resolutions - can you add a comment header to the function explaining why we need this? Is full text search the only case when this kind of fully qualified name is needed?

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -3832,108 +3831,75 @@ exec_stmt_change_dbowner(PLtsql_execstate *estate, PLtsql_stmt_change_dbowner *s
...
-					(errcode(ERRCODE_UNDEFINED_OBJECT),
-					 errmsg("Table or indexed view \"%s\" does not have a full-text index or user does not have permission to perform this action."
-					 , table_name))); 
+	{
+		if (old_index_name)					
+			query_str = gen_dropfulltextindex_cmds(old_index_name, schema_name);
+		else
+			ereport(ERROR,
+				(errcode(ERRCODE_UNDEFINED_OBJECT),
+					errmsg("Table or indexed view \"%s\" does not have a full-text index or user does not have permission to perform this action.",
```
Sairakan, line 3916: The error message mentions permissions - have we tested for cases of users that wouldn't have DROP FULLTEXT INDEX permissions

Sairakan (approved):
Suggestions to clean up/clarify the code, but otherwise looks good to me. I do have the question about permission checking, but that is lower priority.

contrib/babelfishpg_tsql/src/pl_exec-2.c
roshan0708: Actually, it's just a helper function that I created specifically for getting the schema name for FTS. It's basically just a combination of two functions to handle the master db schema and custom schema case.

contrib/babelfishpg_tsql/src/pl_exec-2.c
roshan0708: Actually this is a generic sql server error message which is thrown when there exist no fulltext index to be deleted from the specified table. Although according to the official docs:

> To drop a full-text index in SQL Server, you need to have ALTER permission on the table or indexed view, or be a member of the sysadmin fixed server role, or db_owner or db_ddladmin fixed database roles.

contrib/babelfishpg_tsql/src/pltsql_utils.c
```diff
@@ -1791,3 +1792,125 @@ List
...
+    {
+        Oid indexOid = lfirst_oid(cell);
+        char *name = get_rel_name(indexOid);
+
+        if (strcmp(name, index_name) == 0)
+        {
+            HeapTuple       indexTuple = SearchSysCache1(INDEXRELID, ObjectIdGetDatum(indexOid));
+            Form_pg_index   indexForm = (Form_pg_index) GETSTRUCT(indexTuple);
+
+            if (indexForm->indisunique && indexForm->indrelid == relid)
```
Sairakan: per https://github.com/MicrosoftDocs/sql-docs/blob/live/docs/t-sql/statements/create-fulltext-index-transact-sql.md#key-index-index_name, in addition to being unique, the index, must also be non-nullable and single-key. I am not sure if the pg_index tuple has metadata for all of these conditions, but if it does we might as well include them in this check as well.

contrib/babelfishpg_tsql/src/guc.c
surendravishnoi: escape hatch usually means that when set corresponding commands will be ignored so that application can work. What will we do in case of FTS feature rollback? how this change will help?

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -52,13 +52,15 @@ static int	exec_stmt_usedb(PLtsql_execstate *estate, PLtsql_stmt_usedb *stmt);
 static int	exec_stmt_usedb_explain(PLtsql_execstate *estate, PLtsql_stmt_usedb *stmt, bool shouldRestoreDb);
 static int	exec_stmt_grantdb(PLtsql_execstate *estate, PLtsql_stmt_grantdb *stmt);
 static int	exec_stmt_grantschema(PLtsql_execstate *estate, PLtsql_stmt_grantschema *stmt);
+static int  exec_stmt_fulltextindex(PLtsql_execstate *estate, PLtsql_stmt_fulltextindex *stmt);
```
surendravishnoi: indentation

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -3817,3 +3834,86 @@ exec_stmt_change_dbowner(PLtsql_execstate *estate, PLtsql_stmt_change_dbowner *s
...
+	Oid			relid;
+	List		*column_name;
+	bool		is_create;
+
+	schema_name = gen_schema_name_for_fulltext_index((char *)stmt->schema_name);
+	schemaOid = LookupExplicitNamespace(schema_name, true);						
+	table_name = stmt->table_name;
+	is_create = stmt->is_create;
+
+	if (schema_name == NULL || strlen(schema_name) == 0 || *schema_name == '\0')
```
surendravishnoi: we are checking for NULL schema here but not in gen_schema_name_for_fulltext_index? Either put an assert that schema name should not be NULL or add checks.

contrib/babelfishpg_tsql/src/pltsql_utils.c
```diff
@@ -1791,3 +1792,125 @@ List
 
 	return stmt_list;
 }
+
+/*
+ * get_fulltext_index_name
+ * Get the fulltext index name of a relation specified by OID
+ */
+char *
+get_fulltext_index_name(Oid relid, const char *table_name)
```
surendravishnoi: indentation is messed up at number of places

contrib/babelfishpg_tsql/src/tsqlIface.cpp
```diff
@@ -6580,6 +6597,102 @@ post_process_alter_table(TSqlParser::Alter_tableContext *ctx, PLtsql_stmt_execsq
...
+    if (pos != std::string::npos) {
+        // Extract the schema name before the "."
+        schema_name = table_info.substr(0, pos);
+        // Extract the table name after the "."
+        table_name = table_info.substr(pos + 1);
+    } else {
+        // No "." character found, set first to the entire string
+        table_name = table_info;
+    }
+    return std::make_pair(downcase_truncate_identifier(table_name.c_str(), table_name.length(), true),
```
surendravishnoi, line 6704: we should add test cases with mixed case schema, table and index name

contrib/babelfishpg_tsql/src/tsqlIface.cpp
```diff
@@ -6580,6 +6597,102 @@ post_process_alter_table(TSqlParser::Alter_tableContext *ctx, PLtsql_stmt_execsq
...
+                           downcase_truncate_identifier(schema_name.c_str(), schema_name.length(), true));
+}
+
+PLtsql_stmt* makeCreateFulltextIndexStmt(TSqlParser::Create_fulltext_indexContext *ctx)
```
surendravishnoi: indentation

contrib/babelfishpg_tsql/src/tsqlIface.cpp
surendravishnoi: if we do not support any, can we remove if and just give an error for having catalog_filegroup_option

contrib/babelfishpg_tsql/src/tsqlIface.cpp
surendravishnoi: same here

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -3817,3 +3834,86 @@ exec_stmt_change_dbowner(PLtsql_execstate *estate, PLtsql_stmt_change_dbowner *s
...
+					(errcode(ERRCODE_WRONG_OBJECT_TYPE),
+						errmsg("A full-text index for table or indexed view \"%s\" has already been created.",
+							table_name)));
+			else
+				query_str = gen_createfulltextindex_cmds(table_name, schema_name, column_name, ft_index_name);
+		}
+		else
+			ereport(ERROR,
+				(errcode(ERRCODE_WRONG_OBJECT_TYPE),
+					errmsg("'\"%s\"' is not a valid index to enforce a full-text search key. A full-text search key must be a unique, non-nullable, single-column index which is not offline, is not defined on a non-deterministic or imprecise nonpersisted computed column, does not have a filter, and has maximum size of 900 bytes. Choose another index for the full-text key.",
```
surendravishnoi, line 3906: there is no test for this, please make sure to add test case for each possible error case

contrib/babelfishpg_tsql/src/pl_exec-2.c
surendravishnoi: we should add some test cases for permission check as well, what level of permissions apply at schema, table level etc.

contrib/babelfishpg_tsql/src/guc.c
roshan0708: If a normal user tries to use this escape hatch to turn on fulltext search, it will throw `permission error`. This can only be turned on if the login is of super user. We have just put it like this so that let's say we roll FTS feature in prod, the customer shouldn't be able to use it rather only we, as superuser, should be able to turn it on. We'll rollback these extra changes of escape hatch once the pentest is done, that's what we have planned so that the customer is able to use it normally without turning on any escape hatch.

Sairakan (approved):

contrib/babelfishpg_tsql/src/fts.c
```diff
@@ -13,6 +14,13 @@ babelfish_fts_rewrite(PG_FUNCTION_ARGS)
     char* input_str = text_to_cstring(input_text);
     char* translated_query;
     text* result_text = NULL; // Initialize result_text to NULL
+
+    if (!pltsql_allow_fulltext_parser)
+    {
+        ereport(ERROR,
+                (errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
+                errmsg("Full Text Search is not yet supported.")));
```
forestkeeper, line 22: Are we sure about this error msg ? It's not supported ?

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -3822,3 +3825,101 @@ exec_stmt_change_dbowner(PLtsql_execstate *estate, PLtsql_stmt_change_dbowner *s
...
+			ereport(ERROR,
+				(errcode(ERRCODE_UNDEFINED_OBJECT),
+					errmsg("Table or indexed view \"%s\" does not have a full-text index or user does not have permission to perform this action.",
+						table_name))); 
+	}
+
+	/* The above query will be
+	 * executed using ProcessUtility()
+	 */
+	exec_utility_cmd_helper(query_str);
```
forestkeeper, line 3923: I'm a litte confused about this part, in some other impl, we do construct a structure, not a query string.

contrib/babelfishpg_tsql/src/pltsql_utils.c
```diff
@@ -1791,3 +1795,204 @@ List
...
+		appendStringInfo(&query, "\"%s\"", table_name);
+	else
+		appendStringInfo(&query, "\"%s\".\"%s\"", schema_name, table_name);
+
+	appendStringInfo(&query, "USING GIN(");
+	for (int i = 0; i < list_length(column_name); i++)
+	{
+		char *col_name = (char *) list_nth(column_name, i);
+		// Add column name
+		appendStringInfo(&query, "to_tsvector('fts_contains_simple', \"%s\")", col_name);
```
forestkeeper, line 1971: It's not a good idea to use string to pass a internal real cmd.

contrib/babelfishpg_tsql/src/backend_parser/gram-tsql-epilogue.y.c
```diff
@@ -1887,6 +1887,10 @@ tsql_index_nulls_order(List *indexParams)
 		if (indexElem->nulls_ordering != SORTBY_NULLS_DEFAULT)
 			continue;
 
+		/* GIN indexes don't support NULLS FIRST/LAST options */
+		if (strcmp(accessMethod, "gin") == 0)
```
forestkeeper, line 1891: Do we need to error out in here ?

contrib/babelfishpg_tsql/src/fts.c
Sairakan: This is a fail-safe, as we are waiting for appsec review. The idea is that until the parser is approved by appsec, it will be controlled by a superuser-only GUC. Customers who wish to experiment with full text search in red panda can submit a request for CP to override the GUC to allow them to use full text search features. We will remove this GUC in the next release after Red Panda assuming that it works well.

contrib/babelfishpg_tsql/src/backend_parser/gram-tsql-epilogue.y.c
Sairakan: This is in order to make this feature work with Xiaohui's PR in https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/1990 - in PG, GIN indexes don't support the NULL ordering which was implemented in that PR to apply to all index creation.

contrib/babelfishpg_tsql/src/pl_exec-2.c
Sairakan: All other cases of calling this function that I found in the code just use a StringInfoData and call the function on the char* contained within the struct, so I don't think this is that different.

contrib/babelfishpg_tsql/src/pltsql_utils.c
Sairakan: Agreed, in the next release I plan to rework this to just generate the correct parse tree within gram-tsql-epilogue.y.c

contrib/babelfishpg_tsql/src/pl_exec-2.c
forestkeeper: would you mind refer to : transformAlterTableStmt , and transformSelectIntoStmt

forestkeeper (approved):

PR #1990, opened 2023-11-07
Adjust default nulls order when creating index

xhfanhe:
### Description

Previously when creating index in Babelfish, the default nulls order is different from SQL Server. This commit changes it to the correct behavior - ASC NULLS FIRST and DESC NULLS LAST.

Task: BABEL-4312

### Check List


test/JDBC/expected/babel_ddl.out
```diff
@@ -59,10 +59,10 @@ select indexname, indexdef from pg_indexes where tablename like 't_' order by in
...
 ~~START~~
 varchar#!#text
-t1_idx1t18e881e6977bd6b8cbb78725b3a8ac988#!#CREATE INDEX t1_idx1t18e881e6977bd6b8cbb78725b3a8ac988 ON master_dbo.t1 USING btree (a)
-t1_idx2t117dbbb74ced1fe936cdf7cd7baeff266#!#CREATE INDEX t1_idx2t117dbbb74ced1fe936cdf7cd7baeff266 ON master_dbo.t1 USING btree (a)
-t1_idx3t19eceb46c036c3c1bd6895a34ec3c93f1#!#CREATE INDEX t1_idx3t19eceb46c036c3c1bd6895a34ec3c93f1 ON master_dbo.t1 USING btree (a)
-t1_idx4t1fb4b953a652720bfa47919dff09b172e#!#CREATE INDEX t1_idx4t1fb4b953a652720bfa47919dff09b172e ON master_dbo.t1 USING btree (a)
+t1_idx1t18e881e6977bd6b8cbb78725b3a8ac988#!#CREATE INDEX t1_idx1t18e881e6977bd6b8cbb78725b3a8ac988 ON master_dbo.t1 USING btree (a NULLS FIRST)
+t1_idx2t117dbbb74ced1fe936cdf7cd7baeff266#!#CREATE INDEX t1_idx2t117dbbb74ced1fe936cdf7cd7baeff266 ON master_dbo.t1 USING btree (a NULLS FIRST)
+t1_idx3t19eceb46c036c3c1bd6895a34ec3c93f1#!#CREATE INDEX t1_idx3t19eceb46c036c3c1bd6895a34ec3c93f1 ON master_dbo.t1 USING btree (a NULLS FIRST)
+t1_idx4t1fb4b953a652720bfa47919dff09b172e#!#CREATE INDEX t1_idx4t1fb4b953a652720bfa47919dff09b172e ON master_dbo.t1 USING btree (a NULLS FIRST)
```
lejaokri, line 65: are these hashes the same regardless where we run the testcase? If yes, then you're good

xhfanhe: Yes they are always the same

lejaokri (dismissed):

test/JDBC/expected/BABEL-4312.out
```diff
@@ -0,0 +1,275 @@
...
+        Index Cond: (a > 5)
+~~END~~
+
+
+SELECT TOP 1 b FROM babel_4312_tbl WHERE b <= 'sss' ORDER BY b DESC
+go
+~~START~~
+text
+Query Text: SELECT TOP 1 b FROM babel_4312_tbl WHERE b <= 'sss' ORDER BY b DESC
+Limit  (cost=0.14..2.82 rows=1 width=32)
```
forestkeeper: pls set select set_config('babelfishpg_tsql.explain_costs', 'off', false) , when explain plan, to make the test less flaky

xhfanhe: Done in the latest revision.

test/JDBC/parallel_query_jdbc_schedule
```diff
@@ -55,6 +55,7 @@ ignore#!#babel_collection
 ignore#!#binary-index-vu-verify
 ignore#!#pgr_select_into
 ignore#!#cast_eliminate-vu-verify
+ignore#!#BABEL-4312
```
forestkeeper: What's the error for parallel schedule ?

xhfanhe: Plan is the same but it’s printing an extra line of the gather node.

forestkeeper:
Can we try some test to validate reindex works ok for ealier created index?

prepare : create index ( in some older version )
verify : explain, reindex, explain
cleanup : cleanup

test/JDBC/parallel_query_jdbc_schedule
2jungkook: In that case, we can either
1. create a separate test file for parallel schedule, or
2. create a test-only function that looks up given string from explain output (it requires some effort, but will be useful)

xhfanhe: Currently parallel schedule depends on jdbc_schedule, we can only exclude extra tests in parallel schedule, but cannot run one in normal jdbc and another in parallel...
We have a group of those tests with different query plan under parallel mode, all of them need further validation and improvement in our parallel test framework. So I think this can be tackled as a whole in future work, will file a Jira for this.

forestkeeper (dismissed):

test/JDBC/upgrade/latest/schedule
```diff
@@ -486,3 +486,4 @@ pivot
 #AUTO_ANALYZE #uncomment this test when preparing for new minor version
 cast_eliminate
 TestDatatypeAggSort
+babel_index_nulls_order-after-15-5
```
2jungkook: We don't use `after-` in the latest schedule.

test/JDBC/expected/babel_index_nulls_order-after-15-5-vu-prepare.out
```diff
@@ -0,0 +1,19 @@
...
+go
+~~ROW COUNT: 17~~
+
+
+CREATE INDEX babel_index_nulls_order_after_15_5_asc_idx_a ON babel_index_nulls_order_after_15_5_tbl (a ASC)
+go
+CREATE INDEX babel_index_nulls_order_after_15_5_desc_idx_b ON babel_index_nulls_order_after_15_5_tbl (b DESC)
+go
+CREATE INDEX babel_index_nulls_order_after_15_5_default_idx_ab ON babel_index_nulls_order_after_15_5_tbl (a, b)
+go
```
2jungkook: I don't think all 3 indexes are used in the verify file. In the prepare file, please query to system tables so that readers can see the exact index names including hashed numbers. And in the verify file, please use hint to enforce indexes we want to test. Ref: https://docs.aws.amazon.com/AmazonRDS/latest/AuroraUserGuide/babelfish-tsql-hints.html

test/JDBC/expected/babel_index_nulls_order-after-15-5-vu-verify.out
```diff
@@ -0,0 +1,239 @@
+SELECT TOP 1 a FROM babel_index_nulls_order_after_15_5_tbl WHERE a <= 5 ORDER BY a DESC
```
2jungkook: In this PR, we changed nulls order. So, we should test NULLs. You need to use something like `where a<=5 or a is null`. For each indexes, please test the following queries and check both query plans and results:
1. select ... order by <column> desc
2. select ... order by <column> asc
3. select ... order by <column>

test/JDBC/expected/babel_index_nulls_order-after-15-5-vu-prepare.out
```diff
@@ -0,0 +1,19 @@
+CREATE TABLE babel_index_nulls_order_after_15_5_tbl (a INT, b VARCHAR(10))
```
2jungkook: For MVU, views behave little bit differently. It stores parse tree and re-create views during MVU. We need to create views in a prepare script, and test it in a verify script, e.g.,
```
create view ... as select top 1 a ... order by a desc
```

test/JDBC/parallel_query_jdbc_schedule
```diff
@@ -55,6 +55,9 @@ ignore#!#babel_collection
 ignore#!#binary-index-vu-verify
 ignore#!#pgr_select_into
 ignore#!#cast_eliminate-vu-verify
+ignore#!#babel_index_nulls_order-after-15-5-vu-prepare
```
2jungkook: Could you add a comment why it is ignored and what should be follow-up actions?

2jungkook (approved):

PR #2031, opened 2023-11-17
Fix bugs for GRANT/REVOKE on SCHEMA

shalinilohia50:
### Description
Bug fixes  for GRANT/REVOKE .. ON SCHEMA

1. Server crash when using empty bracketed name.  `schema::[]`
2. Server crash when using empty quoted identifier. `schema::""`
3. When granting permission to yourself, we need a different error message.
4. Error message in not-supported multi-keyword permission should (i) be in uppercase (ii) have a space between keywords
5. Not-supported object type in error message should be in uppercase
6. GRANT ON SCHEMA:: TO PUBLIC does not take effect
7. Specified collation for column names using NAME datatype

### Issues Resolved
JIRA: BABEL-4344

* **Use case based -** Added

* **Boundary conditions -**

* **Arbitrary inputs -**

* **Negative test cases -**

* **Performance tests -**

* **Tooling impact -**

* **Client tests -**

### Check List


contrib/babelfishpg_tsql/sql/ownership.sql
```diff
@@ -17,11 +17,11 @@ GRANT SELECT on sys.babelfish_sysdatabases TO PUBLIC;
...
-  schema_name NAME NOT NULL,
-  object_name NAME NOT NULL,
-  permission NAME NOT NULL,
-  grantee NAME NOT NULL,
-  object_type NAME,
+  schema_name NAME NOT NULL COLLATE "C",
+  object_name NAME NOT NULL COLLATE "C",
+  permission NAME NOT NULL COLLATE "C",
+  grantee NAME NOT NULL COLLATE "C",
+  object_type NAME COLLATE "C",
```
Deepesh125: As discussed offline, lets use sys.database_default here and ensure that we use correct collation for catalog scan/lookup (heap or index scan).

Deepesh125 (changes requested):

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -3695,43 +3694,55 @@ exec_stmt_grantschema(PLtsql_execstate *estate, PLtsql_stmt_grantschema *stmt)
...
+			if (strcmp(grantee_name, "public") != 0)
+				rolname	= get_physical_user_name(dbname, grantee_name);
+			else
+				rolname = pstrdup("public");
 			role_oid = get_role_oid(rolname, true);
 			grantee_is_db_owner = 0 == strncmp(grantee_name, get_owner_of_db(dbname), NAMEDATALEN);
 
-
-			if (role_oid == InvalidOid)
+			if (strcmp(grantee_name, "public") != 0 && role_oid == InvalidOid)
```
xhfanhe, line 3732: What if grantee_name = "public" and role_oid = InvalidOid?

contrib/babelfishpg_tsql/src/tsqlIface.cpp
```diff
@@ -5637,6 +5652,11 @@ makeGrantdbStatement(TSqlParser::Security_statementContext *ctx)
 					char *grantee_name = pstrdup(downcase_truncate_identifier(id_str.c_str(), id_str.length(), true));
 					grantee_list = lappend(grantee_list, grantee_name);
 				}
+				if (prin->PUBLIC())
+				{
+					char *grantee_name = (char *)"public";
```
xhfanhe: Please use pstrdup here

contrib/babelfishpg_tsql/src/pl_exec-2.c
shalinilohia50: In case of `grantee_name = "public"`, role_oid is InvalidOid.
We don't want to throw the following error for "PUBLIC" `Cannot find the principal 'public', because it does not exist or you do not have permission.` That's why this case is handle here.

contrib/babelfishpg_tsql/sql/ownership.sql
```diff
@@ -17,11 +17,11 @@ GRANT SELECT on sys.babelfish_sysdatabases TO PUBLIC;
...
-  permission NAME NOT NULL,
-  grantee NAME NOT NULL,
-  object_type NAME,
+  schema_name NAME NOT NULL COLLATE sys.database_default,
+  object_name NAME NOT NULL COLLATE sys.database_default,
+  permission NAME NOT NULL COLLATE sys.database_default,
+  grantee NAME NOT NULL COLLATE sys.database_default,
+  object_type NAME COLLATE sys.database_default,
   PRIMARY KEY(dbid, schema_name, object_name, permission, grantee)
 );
```
xhfanhe, line 26: Sorry just one more comment - 
I understand there may be some security concern so we are not exposing this catalog to user, but I think we should still select from this table during JDBC test to make sure all the entries are updated as expected, especially during upgrade. 
Would you please add that into the test? We can do this through psql. Otherwise code lgtm.

xhfanhe (dismissed):

xhfanhe (dismissed):

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -3695,43 +3694,55 @@ exec_stmt_grantschema(PLtsql_execstate *estate, PLtsql_stmt_grantschema *stmt)
...
 				ereport(ERROR, (errcode(ERRCODE_INSUFFICIENT_PRIVILEGE),
 					errmsg("Cannot find the principal '%s', because it does not exist or you do not have permission.", grantee_name)));
 
-			if (pg_namespace_ownercheck(schemaOid, role_oid) || is_member_of_role(role_oid, datdba) || grantee_is_db_owner)
+			if ((strcmp(rolname, user) == 0) || pg_namespace_ownercheck(schemaOid, role_oid) || is_member_of_role(role_oid, get_sysadmin_oid()) || grantee_is_db_owner)
```
Deepesh125, line 3736: should we check rolname is not null before strcmp?

Deepesh125 (changes requested):

contrib/babelfishpg_tsql/src/pl_exec-2.c
shalinilohia50: That's not needed. 
`role_oid` is extracted using `rolname` and it will be Invalid when rolname is NULL. 
We have added a condition for `role_oid == Invalid` above and it throws an error when that's true except when grantee is "public".

contrib/babelfishpg_tsql/sql/ownership.sql
```diff
@@ -17,11 +17,11 @@ GRANT SELECT on sys.babelfish_sysdatabases TO PUBLIC;
...
-  schema_name NAME NOT NULL,
-  object_name NAME NOT NULL,
-  permission NAME NOT NULL,
-  grantee NAME NOT NULL,
-  object_type NAME,
+  schema_name NAME NOT NULL COLLATE sys.database_default,
+  object_name NAME NOT NULL COLLATE sys.database_default,
+  permission NAME NOT NULL COLLATE sys.database_default,
+  grantee NAME NOT NULL COLLATE sys.database_default,
+  object_type NAME COLLATE sys.database_default,
```
KushaalShroff, line 24: We have still not addressed the MOST IMPORTANT issue with the Grant on schema code - Memory concerns on peermission/grantee being individual entries

KushaalShroff (changes requested):

contrib/babelfishpg_tsql/sql/ownership.sql
shalinilohia50: Yes, as discussed last week we will be handling that in a separate PR. 
Please read the description to check the issues handled in this PR.

Deepesh125 (dismissed):
Collations changes looks okay to me!

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -2974,41 +2998,54 @@ del_from_bbf_schema(const char *schema_name,
...
+	tuple_bbf_schema = systable_getnext(scan);
 
 	if (HeapTupleIsValid(tuple_bbf_schema))
 		CatalogTupleDelete(bbf_schema_rel, &tuple_bbf_schema->t_self);
 
-	table_endscan(scan);
+	systable_endscan(scan);
 	table_close(bbf_schema_rel, RowExclusiveLock);
 
 	CommandCounterIncrement();
```
KushaalShroff: Why are we calling CommandCounterIncrement for cases when tuple_bbf_schema is invalid?

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -2932,7 +2945,7 @@ check_bbf_schema_for_schema(const char *schema_name,
 	Relation	bbf_schema_rel;
```
KushaalShroff, line 2956: The name of this function is confusing. Also add comments on top of the function

contrib/babelfishpg_tsql/src/catalog.c
KushaalShroff: Do we even require to CommandCounterIncrement  after CatalogTupleDelete? I dont think so

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -3695,43 +3694,55 @@ exec_stmt_grantschema(PLtsql_execstate *estate, PLtsql_stmt_grantschema *stmt)
...
 		{
 			char	   *grantee_name = (char *) lfirst(lc);
+			char	*user = GetUserNameFromId(GetUserId(), false);
 			Oid	role_oid;
 			bool		grantee_is_db_owner;
-			rolname	= get_physical_user_name(dbname, grantee_name);
+			if (strcmp(grantee_name, "public") != 0)
+				rolname	= get_physical_user_name(dbname, grantee_name);
+			else
+				rolname = pstrdup("public");
```
KushaalShroff, line 3728: You are allocating new memory but not freeing it after use, why?

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -3695,43 +3694,55 @@ exec_stmt_grantschema(PLtsql_execstate *estate, PLtsql_stmt_grantschema *stmt)
 	 * sysadmin or login is not the schema owner, then it doesn't have the permission to GRANT/REVOKE.
 	 */
 	login_is_db_owner = 0 == strncmp(login, get_owner_of_db(dbname), NAMEDATALEN);
-	datdba = get_role_oid("sysadmin", false);
 	schema_name = get_physical_schema_name(dbname, stmt->schema_name);
-	schemaOid = LookupExplicitNamespace(schema_name, true);
+
+	if(schema_name)
+	{
+		schemaOid = LookupExplicitNamespace(schema_name, true);
```
KushaalShroff, line 3702: We need to free schema_name, right?

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -3695,43 +3694,55 @@ exec_stmt_grantschema(PLtsql_execstate *estate, PLtsql_stmt_grantschema *stmt)
...
+	}
+	else
+	{
+		ereport(ERROR,
+					(errcode(ERRCODE_UNDEFINED_SCHEMA),
+					 errmsg("An object or column name is missing or empty. For SELECT INTO statements, verify each column has a name. For other statements, look for empty alias names. Aliases defined as \"\" or [] are not allowed. Change the alias to a valid name.")));
```
KushaalShroff, line 3708: Error code(although right) seems to have an ambiguous relation with error message. So Why have we c/p sql server error message here?

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -3683,7 +3683,6 @@ exec_stmt_grantschema(PLtsql_execstate *estate, PLtsql_stmt_grantschema *stmt)
 	char	   *dbname = get_cur_db_name();
 	char	   *login = GetUserNameFromId(GetSessionUserId(), false);
```
KushaalShroff, line 3684: Not freeing them after use?

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -3695,43 +3694,55 @@ exec_stmt_grantschema(PLtsql_execstate *estate, PLtsql_stmt_grantschema *stmt)
...
-				(errcode(ERRCODE_INSUFFICIENT_PRIVILEGE),
-				 errmsg("Cannot find the schema \"%s\", because it does not exist or you do not have permission.", stmt->schema_name)));
 
 	foreach(lc1, stmt->privileges)
 	{
 		char	*priv_name = (char *) lfirst(lc1);
 		foreach(lc, stmt->grantees)
 		{
 			char	   *grantee_name = (char *) lfirst(lc);
+			char	*user = GetUserNameFromId(GetUserId(), false);
```
KushaalShroff: free user

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -3695,43 +3694,55 @@ exec_stmt_grantschema(PLtsql_execstate *estate, PLtsql_stmt_grantschema *stmt)
...
 			Oid	role_oid;
 			bool		grantee_is_db_owner;
-			rolname	= get_physical_user_name(dbname, grantee_name);
+			if (strcmp(grantee_name, "public") != 0)
```
KushaalShroff, line 3725: Since I wasnt part of the main review, can you tell me if grantee_name be NULL by any chance?

contrib/babelfishpg_tsql/src/pl_exec-2.c
KushaalShroff: Also, wont the user be constant for all privileges and grantees? :)

contrib/babelfishpg_tsql/src/pl_exec-2.c
KushaalShroff, line 3736: How will this work for multi-db?
`(strcmp(rolname, user) == 0)`

contrib/babelfishpg_tsql/src/tsqlIface.cpp
```diff
@@ -5527,6 +5527,11 @@ makeGrantdbStatement(TSqlParser::Security_statementContext *ctx)
 						char *grantee_name = pstrdup(downcase_truncate_identifier(id_str.c_str(), id_str.length(), true));
 						grantee_list = lappend(grantee_list, grantee_name);
 					}
+					if (prin->PUBLIC())
+					{
+						char *grantee_name = pstrdup("public");//(char *)"public";
```
KushaalShroff: Cleanup?

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -3657,10 +3663,19 @@ bbf_ProcessUtility(PlannedStmt *pstmt,
...
-											del_from_bbf_schema(logical_schema, obj, permissions[i], rol_spec->rolename);
+										if ((rol_spec->rolename != NULL) && (strcmp(rol_spec->rolename, "public") != 0))
+										{
+											/*
+											 * 1. If only schema permission exists, don't revoke any permission.
+											 * 2. If only object permission exists, delete entry from the catalog and revoke permission.
+											 * 3. If both schema and object permission exist, don't revoke any permission but delete object
+											 *    entry from the catalog.
+											 */
+											if (check_bbf_schema_for_entry(logical_schema, "ALL", permissions[i], rol_spec->rolename) && !has_schema_perms)
```
KushaalShroff: Are you saying that users wont be able to create and grant permissions to an object named as dbo.ALL?
This is severe!

KushaalShroff: Also create Macros for such hardcoded values

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -3761,8 +3792,11 @@ bbf_ProcessUtility(PlannedStmt *pstmt,
 							foreach(lc, grant->grantees)
 							{
 								RoleSpec	   *rol_spec = (RoleSpec *) lfirst(lc);
-								if ((rol_spec->rolename != NULL) && !check_bbf_schema_for_entry(logicalschema, funcname, "execute", rol_spec->rolename))
-									add_entry_to_bbf_schema(logicalschema, funcname, "execute", rol_spec->rolename, obj_type);
+								if ((rol_spec->rolename != NULL) && (strcmp(rol_spec->rolename, "public") != 0))
+								{
+									if (!check_bbf_schema_for_entry(logicalschema, funcname, "execute", rol_spec->rolename))
+										add_entry_to_bbf_schema(logicalschema, funcname, "execute", rol_spec->rolename, obj_type);
```
KushaalShroff: I am skipping the review for these and further changes in this file due to lack of context

KushaalShroff (changes requested):

contrib/babelfishpg_tsql/src/pl_handler.c
KushaalShroff: Looks like we got lucky, ALL is a reserved keyword

contrib/babelfishpg_tsql/src/catalog.c
kuntalghosh: +1. What does this function do? Why are we not considering grantee?

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -2974,41 +2998,54 @@ del_from_bbf_schema(const char *schema_name,
...
+				get_bbf_schema_perms_idx_oid(),
+				true, NULL, 5, scanKey);
 
-	scan = table_beginscan_catalog(bbf_schema_rel, 5, key);
-
-	tuple_bbf_schema = heap_getnext(scan, ForwardScanDirection);
+	tuple_bbf_schema = systable_getnext(scan);
 
 	if (HeapTupleIsValid(tuple_bbf_schema))
 		CatalogTupleDelete(bbf_schema_rel, &tuple_bbf_schema->t_self);
```
kuntalghosh, line 3061: Could you please mention the cases when the tuple will be missing? Wondering whether we should add an error or warning.

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -2942,26 +2955,37 @@ check_bbf_schema_for_schema(const char *schema_name,
...
 				CStringGetDatum(permission));
 
-	scan = table_beginscan_catalog(bbf_schema_rel, 4, key);
+	scan = systable_beginscan(bbf_schema_rel,
+				get_bbf_schema_perms_idx_oid(),
+				true, NULL, 4, key);
 
-	tuple_bbf_schema = heap_getnext(scan, ForwardScanDirection);
+	tuple_bbf_schema = systable_getnext(scan);
 	if (HeapTupleIsValid(tuple_bbf_schema))
```
kuntalghosh, line 2996: Could you please mention the cases when the tuple will be missing? Wondering whether we should add an error or warning.

contrib/babelfishpg_tsql/src/catalog.c
kuntalghosh: +1 Also, please add a comment why this is needed. May not be obvious.

kuntalghosh: We may need it if subsequent operation in the same command needs to fetch this update from catalog

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -3079,7 +3125,7 @@ grant_perms_to_objects_in_schema(const char *schema_name,
 				  const char *permission,
```
kuntalghosh, line 3144: Please add comment for the function objective

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -3095,21 +3141,31 @@ grant_perms_to_objects_in_schema(const char *schema_name,
...
+				F_NAMEEQ,
 				CStringGetDatum(grantee));
 
-	scan = table_beginscan_catalog(bbf_schema_rel, 4, scanKey);
-	tuple_bbf_schema = heap_getnext(scan, ForwardScanDirection);
+	scan = systable_beginscan(bbf_schema_rel, get_bbf_schema_perms_idx_oid(),
+							true, NULL, 4, scanKey);
+	tuple_bbf_schema = systable_getnext(scan);
 
 	while (HeapTupleIsValid(tuple_bbf_schema))
```
kuntalghosh, line 3189: Should we be worried if the code can't find any tuple?

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -3641,8 +3641,14 @@ bbf_ProcessUtility(PlannedStmt *pstmt,
 									char *permissions[] = {"select", "insert", "update", "references", "delete"};
 									for(i = 0; i < 5; i++)
 									{
-										if ((rol_spec->rolename != NULL) && !check_bbf_schema_for_entry(logical_schema, obj, permissions[i], rol_spec->rolename))
-											add_entry_to_bbf_schema(logical_schema, obj, permissions[i], rol_spec->rolename, obj_type);
+										if ((rol_spec->rolename != NULL) && (strcmp(rol_spec->rolename, "public") != 0))
```
kuntalghosh: I think we can check for the following condition at the beginning of the function
```
if ((rol_spec->rolename != NULL) && (strcmp(rol_spec->rolename, "public") != 0))
```
Then, we can skip the same check for the rest of the function and assume that the condition is true.

kuntalghosh (changes requested):

contrib/babelfishpg_tsql/src/catalog.c
shalinilohia50: If tuple is found, we form a GRANT statement using the catalog information and execute it. Otherwise, we don't have to do anything.

contrib/babelfishpg_tsql/src/catalog.c
shalinilohia50: We call this function `check_bbf_schema_for_entry` to check if the entry exists. If it exists, only then we call `del_from_bbf_schema` to delete that particular entry from the catalog depending on the use case.

contrib/babelfishpg_tsql/src/catalog.c
shalinilohia50: It checks if a particular schema has a specific SCHEMA level permission granted to any user (that's why grantee field is not needed here. If we see a row, it means the permission is granted to some user). 
If not, we don't have to do anything. We have handled that case in bbf_processutility.

contrib/babelfishpg_tsql/src/pl_exec-2.c
shalinilohia50: No, it won't be NULL. We don't allow empty string or NULL value for grantee.

contrib/babelfishpg_tsql/src/pl_exec-2.c
shalinilohia50: Shouldn't we have the same message as SQL Server?
What do you suggest?

contrib/babelfishpg_tsql/src/catalog.c
KushaalShroff: I dont think thats the use case for GRANT/REVOKE

contrib/babelfishpg_tsql/src/pl_exec-2.c
shalinilohia50: Can you suggest a case which might fail?
I have ran the test framework in both single-db and multi-db mode.

contrib/babelfishpg_tsql/src/catalog.c
kuntalghosh: In that case, there should be an assert.

contrib/babelfishpg_tsql/src/catalog.c
kuntalghosh: But, the name doesn't suggest it's a revoke.

contrib/babelfishpg_tsql/src/pl_exec-2.c
shalinilohia50: Message from Rob:

```
Same as SQL Server, if possible, would always be best.
 
Exception: when a part of the error msg makes no sense for Babelfish , like saying you cannot change ownership for 'master, model and tempdb databases' – since we don't supporet model, that was removed from the error msg. But we don't seem to have such a situation here

File: test/JDBC/expected/BABEL-GRANT.out [hunk 23]
```diff
@@ -174,7 +174,7 @@ GO
 REVOKE SELECT ON SCHEMA::scm FROM guest;
 GO
 
-GRANT SHOWPLAN ON OBJECT::t1 TO guest;  -- unsupported permission
+GRANT showplan ON OBJECT::t1 TO guest;  -- unsupported permission
```
kuntalghosh, line 177: Why do we need this change? Only the letter casing is changed here

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -3122,11 +3134,13 @@ clean_up_bbf_schema(const char *schema_name,
 }
 
 /*
+ * CONTEXT
```
KushaalShroff: Lets stick to proper commenting formats as per guidelines

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -3809,13 +3789,9 @@ bbf_ProcessUtility(PlannedStmt *pstmt,
...
 								RoleSpec	   *rol_spec = (RoleSpec *) lfirst(lc);
 								bool has_schema_perms = false;
-								if ((rol_spec->rolename != NULL) && (strcmp(rol_spec->rolename, "public") != 0))
-								{
-									if (check_bbf_schema_for_entry(logicalschema, "ALL", "execute", rol_spec->rolename) && !has_schema_perms)
-										has_schema_perms = true;
-									if (check_bbf_schema_for_entry(logicalschema, funcname, "execute", rol_spec->rolename))
-										del_from_bbf_schema(logicalschema, funcname, "execute", rol_spec->rolename);
-								}
+								if (check_bbf_schema_for_entry(logicalschema, "ALL", "execute", rol_spec->rolename) && !has_schema_perms)
```
KushaalShroff: In one of my previous comments I requested for adding a Macro for "ALL", it improves readability of code and also avoids declaring/modifying constants at multiple places

KushaalShroff (changes requested):

test/JDBC/expected/BABEL-GRANT.out
shalinilohia50: Whether the input "SHOWPLAN" is in lower/upper case, the error message should have that KEYWORD in upper case always.

contrib/babelfishpg_tsql/src/pl_handler.c
shalinilohia50: Sorry I missed that.
Addressed now.

kuntalghosh (dismissed):

contrib/babelfishpg_tsql/src/catalog.h
```diff
@@ -299,6 +299,8 @@ typedef FormData_bbf_function_ext *Form_bbf_function_ext;
 #define Anum_bbf_schema_perms_grantee 5
 #define Anum_bbf_schema_perms_object_type 6
 
+#define SCHEMA_PERMISSION_EXISTS "ALL"
```
KushaalShroff: Again the name is confusing, Isnt it for all objects?
If yes then please name accordingly like PERMISSIONS_FOR_ALL_OBJECTS_IN_SCHEMA

contrib/babelfishpg_tsql/src/pl_exec-2.c
KushaalShroff: How is user = public ever possible?

shalinilohia50: That is not possible. User will not be public.

shalinilohia50: `rolname` can be public, but not `user`.

KushaalShroff (approved):

PR #2036, opened 2023-11-20
fix typmod value in case of nvarchar(max), varchar(max), varbinary(max)

skumawat2025:
### Description

Previously, there was an issue in handling `varchar(max)` , `nvarchar(max)` and `varbinary(max)` in the code. Specifically, when dealing with `atttypmod` with a value of -8000 (representing `TSQLMaxTypmod`) in the `PrepareRowDescription` function for `nvarchar` and `varchar`, an incorrect value of `atttypmod` was being passed to the `SetColMetadataForCharType` function. This resulted in setting an incorrect `maxSize` for the column metadata entry.
Because of this if we assign a long string value to a nvarchar(max) variable with a length such that UTF16 output size is large it will throw a FATAL log ```elog(FATAL, "UTF16 output of varchar/bpchar exceeds max length");```

Additionally, This PR limits a maximum scale value for (var)char, (var)binary, and (n)(var)char types. To do that, we set a  limit of 8000 for (var)char and (var)binary, and a limit of 4000 for (n)(var)char.

Signed-off-by: Sandeep Kumawat <skumwt@amazon.com>

### Issues Resolved
BABEL-4547

### Check List


contrib/babelfishpg_tds/src/backend/tds/tdstypeio.c
```diff
@@ -2980,6 +2980,9 @@ TdsSendTypeNVarchar(FmgrInfo *finfo, Datum value, void *vMetaData)
 	}
 	else
 	{
+		/* We can store upto 2GB (2^31 - 1 bytes) for the nvarchar(max). */
+		if (unlikely(buf.len > VARCHAR_MAX))
+			elog(ERROR, "Number of bytes required for the field of nvarchar(max) exeeds 2GB");
```
KushaalShroff: We should try not throw errors inside a sender

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
```diff
@@ -1725,11 +1725,15 @@ PrepareRowDescription(TupleDesc typeinfo, List *targetlist, int16 *formats,
 												att->attcollation, (atttypmod - 4) * 2);
 				break;
 			case TDS_SEND_VARCHAR:
+				if (atttypmod == TSQLMaxTypmod)
```
KushaalShroff: Why cant we update this in the source of the typmod rather than in the protocol which is a superficial conversion

KushaalShroff (changes requested):
We dont need tests to test these changes??

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
```diff
@@ -1725,11 +1725,15 @@ PrepareRowDescription(TupleDesc typeinfo, List *targetlist, int16 *formats,
...
+					atttypmod = -1;
 				SetColMetadataForCharTypeHelper(col, TDS_TYPE_VARCHAR,
 												att->attcollation, (atttypmod == -1) ?
 												atttypmod : (atttypmod - 4));
 				break;
 			case TDS_SEND_NVARCHAR:
```
KushaalShroff, line 1734: Looks like this should be common issue with: varbinarytypmodin, bpchartypmodin and varchartypmodin
but you have only fixed n/varchar

KushaalShroff (changes requested):

KushaalShroff:
Looks like this is a common issue for varbinarytypmodin, bpchartypmodin and varchartypmodin but you have fixed only n/varchar

skumawat2025:
> We dont need tests to test these changes??

Added test cases

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
skumawat2025: We can't update the  attypmod value inside source of typmod. As this will create a problem in differentiating the below mentioned two stmts. Postgres set value -1 to typmod variable for cases where size is not specified. We handle this cases in parser [here](https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/2036/files#diff-f21f7c8218ecd0db8f17325f6de3788f4f6c8502b6b666d56b4e4a59e69d491dL4592)
So changing TSQLMaxValue to -1 in typmod source will require us to differentiate between these two cases later. 
In the latest commit I've made changes in the parser. 
```
declare @s1 nvarchar = ''
declare @s2 nvarchar(max) = ''
```

skumawat2025:
> Looks like this is a common issue for varbinarytypmodin, bpchartypmodin and varchartypmodin but you have fixed only n/varchar

Yes, in tsql these three data-types support max as a scale. So these three have this issue.
```
1. varchar(max)
2. nvarchar(max)
3. varbinary(max)
```
In the latest commit I have made changes in parser for varchar(max), nvarchar(max). 
For varbinary(max) we can not change typmod value to -1 before in parser because we are calling `resolve_varbinary_typmod_from_exp` ([see code here](https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/2036/files#diff-b2a1b31ddd2622f0d25a2cf4f427e48e80596b4a193bf4ab177a6565c91702bcL1958)) for some cases where typmod is -1 inside protocol. So I think we are left with option where we change atttypmod value in protocl for varbinary.

contrib/babelfishpg_tds/src/backend/tds/tdstypeio.c
skumawat2025: Okay. I included it as we are already doing it for varchar sender function.

Deepesh125: Ideally, we should not be hitting this error in any case. So this error makes sense here (we are doing the same for varchar also).

contrib/babelfishpg_tsql/src/pltsql_utils.c
```diff
@@ -1074,6 +1076,17 @@ is_tsql_any_char_datatype(Oid oid)
 		(*common_utility_plugin_ptr->is_tsql_nvarchar_datatype) (oid);
 }
 
+/* 
+ * tsql data-types that allow syntax varchar(max), nvarchar(max)
+ * varbinary(max) is not allowed as typmod for varbinary if fixed in tdsresponse.
```
Deepesh125: s/if/is

Why varbinary(max) is handled this way? Can we add test cases for varbinary(max) also?

test/JDBC/expected/BABEL-4547-NVARCHAR.out
```diff
@@ -0,0 +1,12 @@
+
+
+DECLARE @babel4547NVarchar NVARCHAR(max) = '';
```
Deepesh125: Do we have enough test cases with nvarchar(max) at column level? Casting to and from nvarchar(max)?

Deepesh125 (changes requested):

contrib/babelfishpg_tds/src/backend/tds/tdstypeio.c
surendravishnoi: this is not even possible, int32 can have maximum value of VARCHAR_MAX, a larger value will mean len will become negative. There could be a sanity check for negative len though.

contrib/babelfishpg_tsql/src/pltsql_utils.c
skumawat2025: Made changes to handle varbinary () in similar way in latest commit

test/JDBC/expected/BABEL-4547-NVARCHAR.out
skumawat2025: Added more test cases.

contrib/babelfishpg_tds/src/backend/tds/tdstypeio.c
Deepesh125: buf.len can not go beyond `INT_MAX/2`, check implementation of enlargeStringInfo

contrib/babelfishpg_tsql/src/pl_gram.y
```diff
@@ -7219,6 +7223,46 @@ parse_datatype(const char *string, int location)
...
+	{
+		/* in T-SQL, length-less (N)(VAR)CHAR's length is treated as 1 by default */
+		char *newString = (char *) palloc(1 + strlen(string)+ 3);
+		strcpy(newString, string);
+		strcat(newString, "(1)");
+		/* Let the main parser try to parse it under standard SQL rules */
+		typeName = typeStringToTypeName(newString);
+		typeName->names = rewrite_plain_name(typeName->names);
+		typenameTypeIdAndMod(NULL, typeName, &type_id, &typmod);
+		pfree(newString);
```
Deepesh125: Instead directly update typmod

contrib/babelfishpg_tsql/src/pl_gram.y
```diff
@@ -7219,6 +7223,46 @@ parse_datatype(const char *string, int location)
...
+			errmsg("The size '%d' exceeds the maximum allowed (8000) for '%s' datatype.",
+				typmod - VARHDRSZ, string)));
+	}
+	else if (typmod > (8000 + VARHDRSZ) && (is_tsql_binary_or_varbinary_datatype(type_id)))
+	{
+		ereport(ERROR,
+			(errcode(ERRCODE_INVALID_PARAMETER_VALUE),
+			errmsg("The size '%d' exceeds the maximum allowed (8000) for '%s' datatype.",
+				typmod - VARHDRSZ, string)));
+	}
```
Deepesh125: consolidate into one condition

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
```diff
@@ -946,7 +948,13 @@ resolve_varbinary_typmod_from_exp(Node *expr)
 			if (!con->constisnull)
 			{
 				bytea	   *source = (bytea *) con->constvalue;
-
+				actualSize = VARSIZE_ANY(source);
```
Deepesh125: Use VARSIZE_ANY_EXHDR to find actual length of the input and compare it against 8000 (define new macro just like VARCHAR_MAX)

test/JDBC/input/BABEL-4547.sql
```diff
@@ -0,0 +1,135 @@
...
+
+-- length > 57532 for varbinary
+declare @b varbinary(max) = convert(varbinary(max), replicate(0x01, 60001));
+select @b;
+go
+
+select cast(REPLICATE(0x01, 60001) as varbinary(max));
+GO
+
+CREATE TABLE babel4547t3(c varbinary(max))
```
Deepesh125, line 63: create table with varbinary(8000) / varbinary(8001)

Deepesh125 (changes requested):

test/JDBC/input/BABEL-4547.sql
skumawat2025: done

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
skumawat2025: done

contrib/babelfishpg_tsql/src/pl_gram.y
skumawat2025: done

contrib/babelfishpg_tsql/src/pl_gram.y
skumawat2025: done

contrib/babelfishpg_tsql/src/pl_gram.y
```diff
@@ -7219,9 +7224,39 @@ parse_datatype(const char *string, int location)
...
+		typmod = 1 + VARHDRSZ;
+	
+	/* for varchar/nvarchar/varbinary(MAX), set typmod back to -1 */
+	else if (typmod == TSQLMaxTypmod && is_tsql_datatype_with_max_scale_expr_allowed(type_id))
+		typmod = -1;
+	
+	else if (typmod > (8000 + VARHDRSZ) && (is_tsql_varchar_or_char_datatype(type_id) || is_tsql_binary_or_varbinary_datatype(type_id)))
+	{
+		datatype_string = strtok(datatype_string, "(");
+		datatype_string = strtok(datatype_string, " ");
```
Deepesh125: find some better way to do this.

Deepesh125 (changes requested):

Deepesh125 (dismissed):

contrib/babelfishpg_tds/src/backend/tds/tdstypeio.c
surendravishnoi: what is conclusion here?

contrib/babelfishpg_tsql/src/pl_gram.y
```diff
@@ -7218,6 +7224,31 @@ parse_datatype(const char *string, int location)
 	typeName = typeStringToTypeName(string);
 	typeName->names = rewrite_plain_name(typeName->names);
 	typenameTypeIdAndMod(NULL, typeName, &type_id, &typmod);
+	DeconstructQualifiedName(typeName->names, &schemaName, &dataTypeName);
```
surendravishnoi: do it only when we are about to throw an error

contrib/babelfishpg_tsql/src/pl_gram.y
```diff
@@ -7218,6 +7224,31 @@ parse_datatype(const char *string, int location)
...
+	/* in T-SQL, length-less (N)(VAR)CHAR's length is treated as 1 by default */
+	if (typmod == -1 && (is_tsql_varchar_or_char_datatype(type_id) || is_tsql_nchar_or_nvarchar_datatype(type_id) 
+									|| is_tsql_binary_or_varbinary_datatype(type_id)))
+		typmod = 1 + VARHDRSZ;
+	
+	/* for varchar/nvarchar/varbinary(MAX), set typmod back to -1 */
+	else if (typmod == TSQLMaxTypmod && is_tsql_datatype_with_max_scale_expr_allowed(type_id))
+		typmod = -1;
+	
+	else if (typmod > (8000 + VARHDRSZ) && (is_tsql_varchar_or_char_datatype(type_id) || is_tsql_binary_or_varbinary_datatype(type_id)))
```
surendravishnoi: did we get this limits of 8000 and 4000 from SQL server? If yes, have we validated all test cases against SQL server

test/JDBC/expected/BABEL-4547.out
surendravishnoi, line 1: this file is not opening for me, diff is too large

skumawat2025: please use this link to see this file - https://raw.githubusercontent.com/babelfish-for-postgresql/babelfish_extensions/80aca92d2c09491b9bd0fc8a700708df28b99903/test/JDBC/expected/BABEL-4547.out

contrib/babelfishpg_tds/src/backend/tds/tdstypeio.c
skumawat2025: So in postgres the maximum char string size we can store is of 1GB [reference](https://www.postgresql.org/docs/current/datatype-character.html#:~:text=In%20any%20case%2C%20the%20longest%20possible%20character%20string%20that%20can%20be%20stored%20is%20about%201%20GB). 
We can also confirm this by looking at this function. [enlargeStringInfo](https://github.com/postgres/postgres/blob/master/src/common/stringinfo.c#L346). 
So, this condition is redundant here. I kept it as we are also doing for varcharsender.

contrib/babelfishpg_tsql/src/pl_gram.y
skumawat2025: Yes these limits are from SQL server. And test cases are cross-checked with sql.

contrib/babelfishpg_tsql/src/pl_gram.y
skumawat2025: done.

contrib/babelfishpg_tds/src/backend/tds/tdstypeio.c
```diff
@@ -2782,9 +2782,6 @@ TdsSendTypeVarchar(FmgrInfo *finfo, Datum value, void *vMetaData)
 	}
 	else
 	{
-		/* We can store upto 2GB (2^31 - 1 bytes) for the varchar(max). */
```
surendravishnoi, line 2785: I am not sure if this can be removed. This is actualLen which does not seem to be coming from stringinfo length

skumawat2025: In case of varchar data-type although stringInfo is not used but we are setting a limit of 1GB over here.
https://github.com/babelfish-for-postgresql/babelfish_extensions/blob/BABEL_3_X_DEV/contrib/babelfishpg_common/src/encoding/encoding_utils.c#L151
So I think it's safe to remove this condition.

Deepesh125 (approved):

surendravishnoi (approved):

KushaalShroff (approved):

PR #2037, opened 2023-11-20
Fix typmod for sys.binary datatype

tanscorpio7:
### Description
sys.binary datatype is created as a domain over sys.bbf_binary. In babelfish we assume hex code as varbinary which must then be casted to binary. For casting we fetch rule from hashmap after coercing both source and target to base type.
So binary --> bbf_binary && varbinary-->bbf_varbinary
This cast is currently defined as binary coercible, that is NO modification needed. 
https://github.com/babelfish-for-postgresql/babelfish_extensions/blob/c57e8bc7820e1ef697dd9898b3aa433ba504a7ab/contrib/babelfishpg_tsql/src/pltsql_coerce.c#L114

We fix it by adding proper function for varbinary --> binary cast

-> The default length for local variables defined for binary and varbinary should be 1
DECLARE @a binary should be equivalent to DECLARE @a binary(1)
DECLARE @a varbinary should be equivalent to DECLARE @a varbinary(1)

->Problems during dump and restore done during upgrade.
The new function created for the cast may not exists in the source version and the ugrade scripts are run only after pg dump and restore carried out during version upgrade.
So ultimately during dump and restore no cast exists for varbinary to binary, which is a necessary requirement or upgrade will fail with error :
_"pg_restore: error: could not execute query: ERROR: argument of DEFAULT must be type sys."binary", not type sys.varbinary"_. 
if we are dumping a procedure with sys.binary arg and a default value

-> binary to varbinary
we do not need to modify binary to varbinary cast since we have a type modifer defined for base type of sys.varbinary
i.e. CAST(sys.bbf_varbiniary as sys.bbf_varbiniary). 

### Issues Resolved

[BABEL-4544]

#### Sign Off

### Check List


test/JDBC/expected/TestBinary-before-14_11-or-15_6-vu-prepare.out
```diff
@@ -0,0 +1,33 @@
+CREATE TABLE testBINARY_dt1(a BINARY(8), b VARBINARY(10));
+#inserting random values
+INSERT INTO testBINARY_dt1(a, b) values (1234, 12345);
+~~ROW COUNT: 1~~
+
+INSERT INTO testBINARY_dt1(a, b) values (NULL, NULL);
+~~ROW COUNT: 1~~
+
+#INSERT INTO testBINARY_dt1(a, b) values (0x31323334, 0x3132333435);
```
Deepesh125: insert value by casting from string data types.

test/JDBC/expected/BABEL-3166-before-14_11-or-15_6-vu-verify.out
```diff
@@ -0,0 +1,38 @@
+-- Look at function's probin for typmod information
+SELECT proname, probin FROM pg_proc WHERE proname = 'babel_3166_func';
```
Deepesh125, line 2: add a few tests for declared param,
declare @a binary = <val>
select @a 

declare @a binary(10) = <val where val length > or <= 10>
select @a


also , add tests for typmod cast,
declare @a binary = <val>
select cast(@a  as binary(100)) or binary(max) 

same for implicit casting as well as by retrieving data from table

Deepesh125 (changes requested):

contrib/babelfishpg_tsql/src/pltsql_coerce.c
```diff
@@ -638,6 +640,14 @@ init_tsql_coerce_hash_tab(PG_FUNCTION_ARGS)
...
 						if (!OidIsValid(entry->castfunc))
 						{
+						   /*
+							* varbinary to binary implicit type cast without function should be allowed
+							* during upgrade since the cast function might not exists in older version
+							*/ 
+							if (babelfish_dump_restore && ((*common_utility_plugin_ptr->is_tsql_varbinary_datatype) (castsource) 
+								&& (*common_utility_plugin_ptr->is_tsql_binary_datatype) (casttarget)))
+								entry->castfunc = 0;
+
```
Deepesh125: can you describe use case and explain why this changes are required? And what error or differences you will see if we do not make this change?

test/JDBC/expected/BABEL_4544.out
```diff
@@ -0,0 +1,222 @@
...
+SELECT DATALENGTH(CAST(@A as BINARY(50))), CAST(@A as BINARY(50)), CAST(@A as VARBINARY(60))
+GO
+~~START~~
+varbinary#!#int
+01#!#1
+~~END~~
+
+~~START~~
+int#!#binary#!#varbinary
+50#!#0100000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000#!#01
```
Deepesh125, line 130: have we verified this against expected tsql behaviour?

contrib/babelfishpg_tsql/src/pltsql_coerce.c
tanscorpio7: Until now varbinary to binary cast was implicitly done without any function call, i.e. binary coercible.
Now we are moving to cast with function. i.e. sys.varbinarybinary

If we do not make this change,
during pg_upgrade _init_tsql_coerce_hash_tab_ will not be able to find the function _sys.varbinarybinary_ for varbinary to binary cast because it may not exists in the source version and will not be created until we run our upgrade scripts (happens after dump restore in version upgrade).
Ultimately we end up with no cast for varbinary --> binary during dump and restore.

A failure case for this situation is default value for a sys.binary in a procedure
proc( id sys.binary default value 0x42) 
pre dump --> 0x42 is sys.binary
psot dump --> 0x42 is sys.varbinary (since hex code is always assumed as varbinary first and pg could not find cast to binary type)

"pg_restore: error: could not execute query: ERROR:  argument of DEFAULT must be type sys."binary", not type sys.varbinary"

test/JDBC/expected/BABEL_4544.out
tanscorpio7: Yes, 
Default length for varbinary and binary during declare is 1, data should be truncated accordingly.

.github/workflows/upgrade-test.yml
```diff
@@ -121,7 +121,7 @@ jobs:
           done
  
       - name: Upload Logs
-        if: always() && steps.test-file-rename.outcome == 'success'
+        if: always() && steps.upgrade-and-test.outcome != 'success'
```
rishabhtanwar29, line 124: Thanks for fixing this.

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--3.4.0--3.5.0.sql
```diff
@@ -37,6 +37,8 @@ LANGUAGE plpgsql;
  * So make sure that any SQL statement (DDL/DML) being added here can be executed multiple times without affecting
  * final behaviour.
  */
+ -- Initialize new cast function for varbinary to binary
+CALL babel_coercion_initializer();
```
rishabhtanwar29: IMO, this is not needed since hash tables will anyway get initialized during library load.

contrib/babelfishpg_tsql/src/pltsql_coerce.c
```diff
@@ -638,9 +640,26 @@ init_tsql_coerce_hash_tab(PG_FUNCTION_ARGS)
 
 						if (!OidIsValid(entry->castfunc))
 						{
+						   /*
```
rishabhtanwar29: Nit: Indentation seems a bit off.

contrib/babelfishpg_tsql/src/pltsql_coerce.c
```diff
@@ -638,9 +640,26 @@ init_tsql_coerce_hash_tab(PG_FUNCTION_ARGS)
...
+							* varbinary to binary implicit type cast without function should be allowed
+							* during upgrade since the cast function might not exists in older version
+							*/ 
+							if (babelfish_dump_restore && ((*common_utility_plugin_ptr->is_tsql_varbinary_datatype) (castsource) 
+								&& (*common_utility_plugin_ptr->is_tsql_binary_datatype) (casttarget)))
+							{
+								entry->castfunc = 0;
+								entry->castcontext = 'i';
+								entry->castmethod = 'b';
```
rishabhtanwar29: Suggestion: Can use MACROS here, ie.: 
1. 'i' -> COERCION_CODE_IMPLICIT
2. 'b' -> COERCION_METHOD_BINARY

test/JDBC/expected/BABEL-3166-before-14_11-or-15_6-vu-verify.out
```diff
@@ -0,0 +1,38 @@
...
+SELECT babel_3166_func(1.2, 'abc', 'abcd', 'abcdefgh', 0x12bcfe);
+go
+~~START~~
+varbinary
+12BCFE000000
+~~END~~
+
+
+
+DROP FUNCTION babel_3166_func;
```
rishabhtanwar29, line 19: Suggestion: We can move DROP statements to a `cleanup` file.

contrib/babelfishpg_tsql/src/pltsql_coerce.c
```diff
@@ -638,9 +640,26 @@ init_tsql_coerce_hash_tab(PG_FUNCTION_ARGS)
 
 						if (!OidIsValid(entry->castfunc))
 						{
+							/*
+							 * varbinary to binary implicit type cast without function should be allowed
+							 * during upgrade since the cast function might not exists in older version
```
Deepesh125: be specific here, which exact version does older version refer here? and what is the current (therefore subsequent relase) that contains the fix?

contrib/babelfishpg_tsql/src/tsqlIface.cpp
```diff
@@ -4558,7 +4559,8 @@ makeDeclareStmt(TSqlParser::Declare_statementContext *ctx, std::map<PLtsql_stmt
 		const char *name = downcase_truncate_identifier(nameStr.c_str(), nameStr.length(), true);
 		check_dup_declare(name);
 		PLtsql_type *type = parse_datatype(typeStr.c_str(), 0);
-		if (type->atttypmod == -1 && is_tsql_any_char_datatype(type->typoid))
+		if (type->atttypmod == -1 && (is_tsql_any_char_datatype(type->typoid)
+			|| is_tsql_binary_or_varbinary_datatype(type->typoid)))
 		{
 			std::string newTypeStr = typeStr + "(1)"; /* in T-SQL, length-less (N)(VAR)CHAR's length is treated as 1 */
```
Deepesh125: please update comment here

contrib/babelfishpg_tsql/src/tsqlIface.cpp
```diff
@@ -4589,7 +4591,8 @@ makeDeclareStmt(TSqlParser::Declare_statementContext *ctx, std::map<PLtsql_stmt
 			{
 				std::string typeStr = ::getFullText(local->data_type());
 				PLtsql_type *type = parse_datatype(typeStr.c_str(), 0);  // FIXME: the second arg should be 'location'
-				if (type->atttypmod == -1 && is_tsql_any_char_datatype(type->typoid))
+				if (type->atttypmod == -1 && (is_tsql_any_char_datatype(type->typoid)
+					|| is_tsql_binary_or_varbinary_datatype(type->typoid)))
 				{
 					std::string newTypeStr = typeStr + "(1)"; /* in T-SQL, length-less (N)(VAR)CHAR's length is treated as 1 */
```
Deepesh125: please update comment here

test/JDBC/expected/BABEL_4544.out
```diff
@@ -0,0 +1,222 @@
+SELECT DATALENGTH(CAST(0X61626364 AS BINARY(3)))
```
Deepesh125, line 1: where are test cases for NULL and varbinary(max) --> binary?

test/JDBC/expected/TestBinary.out
```diff
@@ -49,7 +49,7 @@ INSERT INTO BINARY_dt(a, b) values (1234, 12345);
 prepst#!# INSERT INTO BINARY_dt(a, b) values(?, ?) #!#binary|-|a|-|1234#!#varbinary|-|b|-|12345
 ~~ROW COUNT: 1~~
 
-prepst#!#exec#!#binary|-|a|-|123456789#!#varbinary|-|b|-|12345
+prepst#!#exec#!#binary|-|a|-|12345678#!#varbinary|-|b|-|12345
```
Deepesh125, line 52: why this change?

test/JDBC/expected/babel_datatype.out
```diff
@@ -2279,14 +2279,16 @@ GO
 
 insert into testing6 values (cast('ab' as varbinary));
 GO
-~~ROW COUNT: 1~~
+~~ERROR (Code: 33557097)~~
+
+~~ERROR (Message: String or binary data would be truncated.
+The statement has been terminated.)~~
```
Deepesh125, line 2285: what was the default typmod previously? 30?

Deepesh125 (changes requested):

contrib/babelfishpg_tsql/src/pltsql_coerce.c
```diff
@@ -638,9 +640,26 @@ init_tsql_coerce_hash_tab(PG_FUNCTION_ARGS)
...
+							 */ 
+							if (babelfish_dump_restore && ((*common_utility_plugin_ptr->is_tsql_varbinary_datatype) (castsource) 
+								&& (*common_utility_plugin_ptr->is_tsql_binary_datatype) (casttarget)))
+							{
+								entry->castfunc = 0;
+								entry->castcontext = COERCION_CODE_IMPLICIT;
+								entry->castmethod = COERCION_METHOD_BINARY;
+								value = hash_search(ht_tsql_cast_info, key, HASH_ENTER, NULL);
+								*(tsql_cast_info_entry_t *) value = *entry;
```
surendravishnoi, line 654: how this entry will be replaced by actual function once minor upgrade happens?

tanscorpio7: after the upgrade script has been run, we reload this hash map and then we will have the new function available.

test/JDBC/expected/TestBinary.out
tanscorpio7: Otherwise we will get error instead, since the procedure argument is defined as binary(8)
binary does not allow implicit truncation

test/JDBC/expected/babel_datatype.out
tanscorpio7: No, it was 1, which is correct. 
Typmod for declare and column should be 1.
For cast(something as binary) should be 30.
But we never called the typmod function while inserting, so we never actually got to truncation part.
Error is the right behaviour.

test/JDBC/expected/BABEL_4544.out
tanscorpio7: added

Deepesh125 (approved):

PR #2039, opened 2023-11-20
Fix the remaining issues about default keyword in proc/func call

forestkeeper:
1. The error message should be the same as sql server when there's no default define for a param called by a procedure.
2. If default keyword is used in function call and that param doesn't have default define value, then use NULL as the value.
3. Support call proc with NamedAssignExpr with a default keyword like exec proc @p1 = default

Task: BABEL-335

### Example 

case 1 : 
```
1> create proc p @p1 int as select @p1
2> go
1> exec p default
2> go
```

case 1: 
```
1> create function f (@p1 int) returns int as begin return @p1 end
2> go
1> select dbo.f(default)
2> go
----------
      NULL 
```

case 3:
```
1> create proc p @p1 int=123 as select @p1 
2> go 
1> exec p @p1=default 
2> go
-----------         
123
```

### Check List


lejaokri (approved):
adding @2jungkook to take a second look as well

2jungkook (approved):

PR #2044, opened 2023-11-21
Stored the permissions in a single column using bits

shalinilohia50:
### Description
Store the permissions in a single column using bits

### Issues Resolved

Task: BABEL-4344
Co-authored-by: Anju Bharti
Signed-off-by: Shalini Lohia

* **Use case based -**

* **Boundary conditions -**

* **Arbitrary inputs -**

* **Negative test cases -**

* **Performance tests -**

* **Tooling impact -**

* **Client tests -**

### Check List

### DIFF
--- a/contrib/babelfishpg_tsql/sql/ownership.sql
+++ b/contrib/babelfishpg_tsql/sql/ownership.sql
@@ -19,7 +19,7 @@ CREATE TABLE sys.babelfish_schema_permissions (
   dbid smallint NOT NULL,
   schema_name NAME NOT NULL COLLATE sys.database_default,
   object_name NAME NOT NULL COLLATE sys.database_default,
-  permission SMALLINT,
+  permission SMALLINT NOT NULL,
   grantee NAME NOT NULL COLLATE sys.database_default,
   object_type NAME COLLATE sys.database_default,
   PRIMARY KEY(dbid, schema_name, object_name, grantee)
--- a/contrib/babelfishpg_tsql/sql/ownership.sql
+++ b/contrib/babelfishpg_tsql/sql/ownership.sql
@@ -19,10 +19,10 @@ CREATE TABLE sys.babelfish_schema_permissions (
   dbid smallint NOT NULL,
   schema_name NAME NOT NULL,
   object_name NAME NOT NULL,
-  permission NAME NOT NULL,
+  permission SMALLINT,
   grantee NAME NOT NULL,
   object_type NAME,
-  PRIMARY KEY(dbid, schema_name, object_name, permission, grantee)
+  PRIMARY KEY(dbid, schema_name, object_name, grantee)
 );
 
 -- BABELFISH_FUNCTION_EXT
--- a/contrib/babelfishpg_tsql/src/catalog.c
+++ b/contrib/babelfishpg_tsql/src/catalog.c
@@ -2890,39 +2890,51 @@ get_bbf_schema_privilege(const char *schema_name,
 {
 	Relation	bbf_schema_rel;
 	HeapTuple	tuple_bbf_schema;
-	ScanKeyData	key[4];
-	TableScanDesc	scan;
+	ScanKeyData	scanKey[4];
+	SysScanDesc	scan;
 	int16	dbid = get_cur_db_id();
 	int16 permission = 0;
 
 	bbf_schema_rel = table_open(get_bbf_schema_perms_oid(),
 									AccessShareLock);
-	ScanKeyInit(&key[0],
+	ScanKeyInit(&scanKey[0],
 				Anum_bbf_schema_perms_dbid,
 				BTEqualStrategyNumber, F_INT2EQ,
 				Int16GetDatum(dbid));
-	ScanKeyInit(&key[1],
+	ScanKeyEntryInitialize(&scanKey[1], 0,
 				Anum_bbf_schema_perms_schema_name,
-				BTEqualStrategyNumber, F_NAMEEQ,
+				BTEqualStrategyNumber,
+				InvalidOid,
+				tsql_get_server_collation_oid_internal(false),
+				F_NAMEEQ,
 				CStringGetDatum(schema_name));
-	ScanKeyInit(&key[2],
+	ScanKeyEntryInitialize(&scanKey[2], 0,
 				Anum_bbf_schema_perms_object_name,
-				BTEqualStrategyNumber, F_NAMEEQ,
+				BTEqualStrategyNumber,
+				InvalidOid,
+				tsql_get_server_collation_oid_internal(false),
+				F_NAMEEQ,
 				CStringGetDatum(object_name));
-	ScanKeyInit(&key[3],
+	ScanKeyEntryInitialize(&scanKey[3], 0,
 				Anum_bbf_schema_perms_grantee,
-				BTEqualStrategyNumber, F_NAMEEQ,
+				BTEqualStrategyNumber,
+				InvalidOid,
+				tsql_get_server_collation_oid_internal(false),
+				F_NAMEEQ,
 				CStringGetDatum(grantee));
 
-	scan = table_beginscan_catalog(bbf_schema_rel, 4, key);
-	tuple_bbf_schema = heap_getnext(scan, ForwardScanDirection);
+	scan = systable_beginscan(bbf_schema_rel,
+				get_bbf_schema_perms_idx_oid(),
+				true, NULL, 4, scanKey);
+	tuple_bbf_schema = systable_getnext(scan);
+
 	if (HeapTupleIsValid(tuple_bbf_schema))
 	{
 		Form_bbf_schema_perms bbf_schema = (Form_bbf_schema_perms) GETSTRUCT(tuple_bbf_schema);
 		permission = bbf_schema->permission;
 	}
 
-	table_endscan(scan);
+	systable_endscan(scan);
 	table_close(bbf_schema_rel, AccessShareLock);
 	return permission;
 }
@@ -2943,7 +2955,7 @@ update_bbf_schema_privilege(const char *schema_name,
 	HeapTuple	tuple_bbf_schema;
 	TupleDesc	bbf_schema_dsc;
 	HeapTuple	new_tuple;
-	ScanKeyData key[5];
+	ScanKeyData scanKey[5];
 	TableScanDesc scan;
 	int16	dbid = get_cur_db_id();
 	int16 permission = 0;
@@ -2975,28 +2987,37 @@ update_bbf_schema_privilege(const char *schema_name,
 	bbf_schema_rel = table_open(get_bbf_schema_perms_oid(),
 							RowExclusiveLock);
 
-	ScanKeyInit(&key[0],
+	ScanKeyInit(&scanKey[0],
 				Anum_bbf_schema_perms_dbid,
 				BTEqualStrategyNumber, F_INT2EQ,
 				Int16GetDatum(dbid));
-	ScanKeyInit(&key[1],
+	ScanKeyEntryInitialize(&scanKey[1], 0,
 				Anum_bbf_schema_perms_schema_name,
-				BTEqualStrategyNumber, F_NAMEEQ,
+				BTEqualStrategyNumber,
+				InvalidOid,
+				tsql_get_server_collation_oid_internal(false),
+				F_NAMEEQ,
 				CStringGetDatum(schema_name));
-	ScanKeyInit(&key[2],
+	ScanKeyEntryInitialize(&scanKey[2], 0,
 				Anum_bbf_schema_perms_object_name,
-				BTEqualStrategyNumber, F_NAMEEQ,
+				BTEqualStrategyNumber,
+				InvalidOid,
+				tsql_get_server_collation_oid_internal(false),
+				F_NAMEEQ,
 				CStringGetDatum(object_name));
-	ScanKeyInit(&key[3],
+	ScanKeyInit(&scanKey[3],
 				Anum_bbf_schema_perms_permission,
 				BTEqualStrategyNumber, F_INT2EQ,
 				Int16GetDatum(old_priv));
-	ScanKeyInit(&key[4],
+	ScanKeyEntryInitialize(&scanKey[4], 0,
 				Anum_bbf_schema_perms_grantee,
-				BTEqualStrategyNumber, F_NAMEEQ,
+				BTEqualStrategyNumber,
+				InvalidOid,
+				tsql_get_server_collation_oid_internal(false),
+				F_NAMEEQ,
 				CStringGetDatum(grantee));
 
-	scan = table_beginscan_catalog(bbf_schema_rel, 5, key);
+	scan = table_beginscan_catalog(bbf_schema_rel, 5, scanKey);
 	tuple_bbf_schema = heap_getnext(scan, ForwardScanDirection);
 	if (HeapTupleIsValid(tuple_bbf_schema))
 	{
--- a/contrib/babelfishpg_tsql/src/catalog.c
+++ b/contrib/babelfishpg_tsql/src/catalog.c
@@ -3042,8 +3042,6 @@ update_bbf_schema_privilege(const char *schema_name,
 
 	table_endscan(scan);
 	table_close(bbf_schema_rel, RowExclusiveLock);
-
-	CommandCounterIncrement();
 }
 
 /* Check if the catalog entry exists. */
@@ -3113,10 +3111,11 @@ check_bbf_schema_for_schema(const char *schema_name,
 {
 	Relation	bbf_schema_rel;
 	HeapTuple	tuple_bbf_schema;
-	ScanKeyData key[3];
-	SysScanDesc scan;
-	int16	dbid = get_cur_db_id();
-	int16 priv = 0;
+	ScanKeyData	key[3];
+	SysScanDesc	scan;
+	int16		dbid = get_cur_db_id();
+	int16		priv = 0;
+	bool		permission_exists_on_schema = false;
 
 	bbf_schema_rel = table_open(get_bbf_schema_perms_oid(),
 									AccessShareLock);
@@ -3144,19 +3143,19 @@ check_bbf_schema_for_schema(const char *schema_name,
 				true, NULL, 3, key);
 
 	tuple_bbf_schema = systable_getnext(scan);
-	while (HeapTupleIsValid(tuple_bbf_schema))
+	if (HeapTupleIsValid(tuple_bbf_schema))
 	{
 		Form_bbf_schema_perms schemaform;
 		schemaform = (Form_bbf_schema_perms) GETSTRUCT(tuple_bbf_schema);
 		priv = schemaform->permission;
 
 		if((permission & priv) == permission)
-			return true;
+			permission_exists_on_schema = true;
 	}
 
 	systable_endscan(scan);
 	table_close(bbf_schema_rel, AccessShareLock);
-	return false;
+	return permission_exists_on_schema;
 }
 
 void
--- a/contrib/babelfishpg_tsql/src/catalog.c
+++ b/contrib/babelfishpg_tsql/src/catalog.c
@@ -3283,12 +3283,6 @@ clean_up_bbf_schema(const char *schema_name,
 	table_close(bbf_schema_rel, RowExclusiveLock);
 }
 
-/*
- * For all objects belonging to a schema which has OBJECT level permission,
- * It grants the permission explicitly when REVOKE has been executed on that
- * specific schema.
- */
-
 void
 grant_perms_to_each_obj(const char *db_name,
 					const char	*object_type,
@@ -3340,6 +3334,13 @@ grant_perms_to_each_obj(const char *db_name,
 	pfree(query.data);
 }
 
+
+/*
+ * For all objects belonging to a schema which has OBJECT level permission,
+ * It grants the permission explicitly when REVOKE has been executed on that
+ * specific schema.
+ */
+
 void
 grant_perms_to_objects_in_schema(const char *schema_name,
 				int16 priv,
@@ -3389,7 +3390,7 @@ grant_perms_to_objects_in_schema(const char *schema_name,
 		object_type = pstrdup(NameStr(schemaform->object_type));
 		permission = (priv & (schemaform->permission));
 		/* For each object, grant the permission explicitly. */
-		if (strcmp(object_name, "ALL") != 0)
+		if (strcmp(object_name, PERMISSIONS_FOR_ALL_OBJECTS_IN_SCHEMA) != 0)
 		{
 			if((permission & PRIVILEGE_BIT_FOR_EXECUTE) == PRIVILEGE_BIT_FOR_EXECUTE)
 				grant_perms_to_each_obj(db_name, object_type, schema_name, object_name, grantee, "execute");
--- a/contrib/babelfishpg_tsql/src/catalog.c
+++ b/contrib/babelfishpg_tsql/src/catalog.c
@@ -2832,7 +2832,7 @@ rename_procfunc_update_bbf_catalog(RenameStmt *stmt)
 void
 add_entry_to_bbf_schema(const char *schema_name,
 				const char *object_name,
-				const char *permission,
+				int16 permission,
 				const char *grantee,
 				const char *object_type)
 {
@@ -2855,7 +2855,7 @@ add_entry_to_bbf_schema(const char *schema_name,
 	new_record_bbf_schema[BBF_SCHEMA_PERMS_DBID] = Int16GetDatum(dbid);
 	new_record_bbf_schema[BBF_SCHEMA_PERMS_SCHEMA_NAME] = CStringGetDatum(pstrdup(schema_name));
 	new_record_bbf_schema[BBF_SCHEMA_PERMS_OBJECT_NAME] = CStringGetDatum(pstrdup(object_name));
-	new_record_bbf_schema[BBF_SCHEMA_PERMS_PERMISSION] = CStringGetDatum(pstrdup(permission));
+	new_record_bbf_schema[BBF_SCHEMA_PERMS_PERMISSION] = Int16GetDatum(permission);
 	new_record_bbf_schema[BBF_SCHEMA_PERMS_GRANTEE] = CStringGetDatum(pstrdup(grantee));
 	if (object_type != NULL)
 		new_record_bbf_schema[BBF_SCHEMA_PERMS_OBJECT_TYPE] = CStringGetDatum(pstrdup(object_type));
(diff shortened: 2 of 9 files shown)


contrib/babelfishpg_tsql/sql/ownership.sql
```diff
@@ -19,10 +19,10 @@ CREATE TABLE sys.babelfish_schema_permissions (
   dbid smallint NOT NULL,
   schema_name NAME NOT NULL COLLATE sys.database_default,
   object_name NAME NOT NULL COLLATE sys.database_default,
-  permission NAME NOT NULL COLLATE sys.database_default,
+  permission SMALLINT,
```
kuntalghosh: Should be NOT NULL, right?

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -2880,16 +2880,181 @@ add_entry_to_bbf_schema(const char *schema_name,
...
+									new_record_repl_bbf_schema);
+
+		CatalogTupleUpdate(bbf_schema_rel, &new_tuple->t_self, new_tuple);
+		heap_freetuple(new_tuple);
+	}
+
+	table_endscan(scan);
+	table_close(bbf_schema_rel, RowExclusiveLock);
+
+	CommandCounterIncrement();
```
kuntalghosh: Why is this needed?

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -2980,36 +3138,35 @@ check_bbf_schema_for_schema(const char *schema_name,
...
-	if (HeapTupleIsValid(tuple_bbf_schema))
-		catalog_entry_exists = true;
+	while (HeapTupleIsValid(tuple_bbf_schema))
+	{
+		Form_bbf_schema_perms schemaform;
+		schemaform = (Form_bbf_schema_perms) GETSTRUCT(tuple_bbf_schema);
+		priv = schemaform->permission;
+
+		if((permission & priv) == permission)
+			return true;
```
kuntalghosh: Not ending the scan and not closing the table. Have you added a test for this? Otherwise, the test should have complained.

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -3192,50 +3388,22 @@ grant_perms_to_objects_in_schema(const char *schema_name,
...
+			if((permission & PRIVILEGE_BIT_FOR_SELECT) == PRIVILEGE_BIT_FOR_SELECT)
+				grant_perms_to_each_obj(db_name, object_type, schema_name, object_name, grantee, "select");
+			if((permission & PRIVILEGE_BIT_FOR_INSERT) == PRIVILEGE_BIT_FOR_INSERT)
+				grant_perms_to_each_obj(db_name, object_type, schema_name, object_name, grantee, "insert");
+			if((permission & PRIVILEGE_BIT_FOR_UPDATE) == PRIVILEGE_BIT_FOR_UPDATE)
+				grant_perms_to_each_obj(db_name, object_type, schema_name, object_name, grantee, "update");
+			if((permission & PRIVILEGE_BIT_FOR_DELETE) == PRIVILEGE_BIT_FOR_DELETE)
+				grant_perms_to_each_obj(db_name, object_type, schema_name, object_name, grantee, "delete");
+			if((permission & PRIVILEGE_BIT_FOR_REFERENCES) == PRIVILEGE_BIT_FOR_REFERENCES)
+				grant_perms_to_each_obj(db_name, object_type, schema_name, object_name, grantee, "references");
```
kuntalghosh, line 3405: Try to execute a single GRANT statement with all the permissions at once. Otherwise, it'll execute individual GRANT statement

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -3675,20 +3676,53 @@ get_insert_bulk_kilobytes_per_batch()
...
-static int
-exec_stmt_grantschema(PLtsql_execstate *estate, PLtsql_stmt_grantschema *stmt)
+static void
+exec_gen_grantschema(char *schema_name, char *rolname,  PLtsql_stmt_grantschema *stmt, char *priv_name)
 {
 	List	   *parsetree_list;
 	ListCell   *parsetree_item;
-	char	   *dbname = get_cur_db_name();
-	char	   *login = GetUserNameFromId(GetSessionUserId(), false);
+	parsetree_list = gen_grantschema_subcmds(schema_name, rolname, stmt->is_grant, stmt->with_grant_option, priv_name);
```
kuntalghosh, line 3684: Shouldn't you free this list?

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -3714,75 +3748,63 @@ exec_stmt_grantschema(PLtsql_execstate *estate, PLtsql_stmt_grantschema *stmt)
...
 		}
+		/* Remove entry for each revoke statement. */
+		else if (!stmt->is_grant && check_bbf_schema_for_entry(stmt->schema_name, PERMISSIONS_FOR_ALL_OBJECTS_IN_SCHEMA, rolname))
+		{
+			/* If any object in the schema has the OBJECT level permission. Then, internally grant that permission back. */
+			grant_perms_to_objects_in_schema(stmt->schema_name, privilege_maskInt, rolname);
+			old_priv = get_bbf_schema_privilege(stmt->schema_name, PERMISSIONS_FOR_ALL_OBJECTS_IN_SCHEMA, rolname);
+			if(old_priv!= privilege_maskInt || ((old_priv)&(~privilege_maskInt)) != old_priv)
+				update_bbf_schema_privilege(stmt->schema_name, PERMISSIONS_FOR_ALL_OBJECTS_IN_SCHEMA, privilege_maskInt, old_priv, rolname, NULL, stmt->is_grant);
+		}
```
kuntalghosh: Including the same condition multiple times in the same if..else block is not a good coding practice. Please write it as following:
```
if (grant)
{
 if (check_bbf_schema_for_entry())
    // update the entry
 else
   // add the entry
}
else
{
 if (check_bbf_schema_for_entry())
    // update the entry
}
```

For revoke, why is this a no-op if an entry is not found?

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -3655,90 +3657,104 @@ bbf_ProcessUtility(PlannedStmt *pstmt,
...
+									continue;
+								if (check_bbf_schema_for_entry(logical_schema, obj, rol_spec->rolename))
+								{
+									int16 old_priv = get_bbf_schema_privilege(logical_schema, obj,rol_spec->rolename);
+									if(old_priv!= privilege_maskInt || ((old_priv)&(~(privilege_maskInt))) != old_priv)
+										update_bbf_schema_privilege(logical_schema, obj, privilege_maskInt, old_priv, rol_spec->rolename, obj_type, grant->is_grant);
 								}
 							}
-						}
+						}
```
kuntalghosh: whitespace

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -3779,8 +3794,14 @@ bbf_ProcessUtility(PlannedStmt *pstmt,
...
 							{
 								RoleSpec	   *rol_spec = (RoleSpec *) lfirst(lc);
-								if (!check_bbf_schema_for_entry(logicalschema, funcname, "execute", rol_spec->rolename))
-									add_entry_to_bbf_schema(logicalschema, funcname, "execute", rol_spec->rolename, obj_type);
+								if (!check_bbf_schema_for_entry(logicalschema, funcname, rol_spec->rolename))
+									add_entry_to_bbf_schema(logicalschema, funcname, 32, rol_spec->rolename, obj_type);
+								else
+								{
+									int16 old_priv = get_bbf_schema_privilege(logicalschema, funcname, rol_spec->rolename);
+									if(old_priv!= 32 || ((old_priv)|(32)) != old_priv)
```
kuntalghosh: Use a macro for 32 (should have a meaningful name)

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -3714,75 +3748,63 @@ exec_stmt_grantschema(PLtsql_execstate *estate, PLtsql_stmt_grantschema *stmt)
...
-				del_from_bbf_schema(stmt->schema_name, PERMISSIONS_FOR_ALL_OBJECTS_IN_SCHEMA, priv_name, rolname);
-			}
-			pfree(rolname);
+		if (!is_member_of_role(GetSessionUserId(), get_sysadmin_oid()) && !login_is_db_owner && !pg_namespace_ownercheck(schemaOid, GetUserId()))
+			ereport(ERROR,
+				(errcode(ERRCODE_INSUFFICIENT_PRIVILEGE),
+					errmsg("Cannot find the schema \"%s\", because it does not exist or you do not have permission.", stmt->schema_name)));
+
+		if((privilege_maskInt & PRIVILEGE_BIT_FOR_EXECUTE) == PRIVILEGE_BIT_FOR_EXECUTE)
+			exec_gen_grantschema(schema_name, rolname, stmt, "execute");
```
kuntalghosh, line 3777: Send privilege names as a list and execute a single statement

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -3623,7 +3626,6 @@ bbf_ProcessUtility(PlannedStmt *pstmt,
 						const char *logical_schema = NULL;
```
kuntalghosh: Why are you handling CREATE LOGICAL DATABASE in OBJECT_TABLE? What other subcommand can fall under this category?

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -3655,90 +3657,104 @@ bbf_ProcessUtility(PlannedStmt *pstmt,
...
+							* 1. Execute the GRANT statement.
+							* 2. Add its corresponding entry in the catalog, if doesn't exist already.
+							* 3. Don't add an entry, if the permission is granted on column list.
+							*/
+							if (prev_ProcessUtility)
+								prev_ProcessUtility(pstmt, queryString, readOnlyTree, context, params,
+									queryEnv, dest, qc);
+							else
+								standard_ProcessUtility(pstmt, queryString, readOnlyTree, context, params,
+									queryEnv, dest, qc);
```
kuntalghosh: Shouldn't this call exec_stmt_grantschema internally and update the catalog? If not, when exec_stmt_grantschema is called exactly?

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -3655,90 +3657,104 @@ bbf_ProcessUtility(PlannedStmt *pstmt,
...
 								{
-									RoleSpec	   *rol_spec = (RoleSpec *) lfirst(lc);
-									/*
-									 * Don't add an entry, if the permission is granted on column list.
-									 */
-									if (ap->cols != NULL)
+									AccessPriv *ap = (AccessPriv *) lfirst(lc1);
+									if(ap->cols != NULL)
+									{
+										flag = true;
```
kuntalghosh: What does this flag indicate?

contrib/babelfishpg_tsql/src/pltsql.h
```diff
@@ -1914,6 +1914,15 @@ extern int	insert_bulk_rows_per_batch;
...
 
+/* Privilege Bits for GRANT SCHEMA */
+#define PRIVILEGE_BIT_FOR_EXECUTE 32
+#define ALL_PRIVILEGE_EXCEPT_EXECUTE 31
+#define PRIVILEGE_BIT_FOR_SELECT 16
+#define PRIVILEGE_BIT_FOR_INSERT 8
+#define PRIVILEGE_BIT_FOR_UPDATE 4
+#define PRIVILEGE_BIT_FOR_DELETE 2
+#define PRIVILEGE_BIT_FOR_REFERENCES 1
+
```
kuntalghosh, line 1925: Please use the same ACL [defined](https://github.com/babelfish-for-postgresql/postgresql_modified_for_babelfish/blob/BABEL_3_X_DEV__PG_15_X/src/include/nodes/parsenodes.h#L82) in PG.  This will help us reusing some internal functions/macros if required. For example you can use privilege_to_string() to retrieve the values or ACLITEM_SET_PRIVS to set the privileges.

kuntalghosh (changes requested):

contrib/babelfishpg_tsql/src/pl_exec-2.c
shalinilohia50: I don't think it's needed. [References](https://github.com/babelfish-for-postgresql/babelfish_extensions/blob/a695d88e21a3b0418f2d49a1539a759e48be7d19/contrib/babelfishpg_tsql/src/dbcmds.c#L525)

contrib/babelfishpg_tsql/src/pl_exec-2.c
shalinilohia50: We have separate sub commands for below privileges:
 1. [select, insert, update, references] -> Applicable for Tables
 2. [execute] -> Applicable for Functions and Procedures
 So, we cannot execute a single statement if a grant statement specifies multiple privileges like `insert, select, update`

contrib/babelfishpg_tsql/src/catalog.c
shalinilohia50: We have separate sub commands for below privileges:

1. [select, insert, update, references] -> Applicable for Tables
2. [execute] -> Applicable for Functions and Procedures
So, we cannot execute a single statement if a grant statement specifies multiple privileges like insert, select, update

contrib/babelfishpg_tsql/src/pl_handler.c
shalinilohia50: We have defined a separate node `PLtsql_stmt_grantschema` for `GRANT/REVOKE on SCHEMA` statement which executes `exec_stmt_grantschema`

contrib/babelfishpg_tsql/src/pl_handler.c
shalinilohia50: This specifies if there is permission granted to a specific column in a table. 
Renaming it to something meaningful.

contrib/babelfishpg_tsql/sql/ownership.sql
KushaalShroff: +1

contrib/babelfishpg_tsql/src/catalog.c
KushaalShroff: +1, but that would only come into consideration if you check the logs. There would have been explicit cache leak references

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -3133,6 +3283,58 @@ clean_up_bbf_schema(const char *schema_name,
...
+					const char *schema_name,
+					const char	*object_name,
+					const char *grantee,
+					const char *permission)
+{
+	StringInfoData	query;
+	char			*schema;
+	List			*res;
+	Node			*res_stmt;
+	PlannedStmt		*wrapper;
```
KushaalShroff, line 3297: We need to free 3 of these, right?

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -3192,50 +3388,22 @@ grant_perms_to_objects_in_schema(const char *schema_name,
...
-						false,
-						PROCESS_UTILITY_SUBCOMMAND,
-						NULL,
-						NULL,
-						None_Receiver,
-						NULL);
-
-			/* make sure later steps can see the object created here */
-			CommandCounterIncrement();
+			if((permission & PRIVILEGE_BIT_FOR_EXECUTE) == PRIVILEGE_BIT_FOR_EXECUTE)
```
KushaalShroff, line 3394: Why do we need to equate???
Simple `if(permission & PRIVILEGE_BIT_FOR_EXECUTE)` should work!

contrib/babelfishpg_tsql/src/pltsql_utils.c
```diff
@@ -1999,3 +1999,35 @@ char
...
+		else if (!strcmp(priv_name,"insert"))
+			privilege_maskInt |= PRIVILEGE_BIT_FOR_INSERT;
+		else if (!strcmp(priv_name,"update"))
+			privilege_maskInt |= PRIVILEGE_BIT_FOR_UPDATE;
+		else if (!strcmp(priv_name,"delete"))
+			privilege_maskInt |= PRIVILEGE_BIT_FOR_DELETE;
+		else if (!strcmp(priv_name,"references"))
+			privilege_maskInt |= PRIVILEGE_BIT_FOR_REFERENCES;
+	}
+	return privilege_maskInt;
```
KushaalShroff, line 2032: WHY?
Why cant we simply store privilege_maskInt during parsing? You are parsing a string and storing a list of strings in antlr and then coming here to convert List of strings to int. Seems unnecessarily costly to me.

contrib/babelfishpg_tsql/src/pl_exec-2.c
shalinilohia50: For REVOKE:
1. Update the entry if it exists already. 
2. Otherwise, do nothing.

contrib/babelfishpg_tsql/src/catalog.c
KushaalShroff, line 3405: What if grant_perms_to_each_obj -> ProcessUtility throws error? It will leave the table bbf_schema_rel  open, right?
Practice safer coding techniques by using try/catch blocks

contrib/babelfishpg_tsql/src/pl_exec-2.c
KushaalShroff: I agree, we can write a single batch for all grant queries and get it executed as a batch, if thats possible. Is it? The code will be cleaner in that case

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -3714,75 +3748,63 @@ exec_stmt_grantschema(PLtsql_execstate *estate, PLtsql_stmt_grantschema *stmt)
...
-				grant_perms_to_objects_in_schema(stmt->schema_name, priv_name, rolname);
-				del_from_bbf_schema(stmt->schema_name, PERMISSIONS_FOR_ALL_OBJECTS_IN_SCHEMA, priv_name, rolname);
-			}
-			pfree(rolname);
+		if (!is_member_of_role(GetSessionUserId(), get_sysadmin_oid()) && !login_is_db_owner && !pg_namespace_ownercheck(schemaOid, GetUserId()))
+			ereport(ERROR,
+				(errcode(ERRCODE_INSUFFICIENT_PRIVILEGE),
+					errmsg("Cannot find the schema \"%s\", because it does not exist or you do not have permission.", stmt->schema_name)));
+
+		if((privilege_maskInt & PRIVILEGE_BIT_FOR_EXECUTE) == PRIVILEGE_BIT_FOR_EXECUTE)
```
KushaalShroff, line 3776: Simple if(permission & PRIVILEGE_BIT_FOR_EXECUTE) should work, right?

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -3714,75 +3748,63 @@ exec_stmt_grantschema(PLtsql_execstate *estate, PLtsql_stmt_grantschema *stmt)
...
+		if((privilege_maskInt & PRIVILEGE_BIT_FOR_REFERENCES) == PRIVILEGE_BIT_FOR_REFERENCES)
+			exec_gen_grantschema(schema_name, rolname, stmt, "references");
+
+		/* Add entry for each grant statement. */
+		if (stmt->is_grant && !check_bbf_schema_for_entry(stmt->schema_name, PERMISSIONS_FOR_ALL_OBJECTS_IN_SCHEMA, rolname))
+			add_entry_to_bbf_schema(stmt->schema_name, PERMISSIONS_FOR_ALL_OBJECTS_IN_SCHEMA,  privilege_maskInt, rolname, NULL);
+		else if(stmt->is_grant)
+		{
+			int16 old_priv = get_bbf_schema_privilege(stmt->schema_name, PERMISSIONS_FOR_ALL_OBJECTS_IN_SCHEMA, rolname);
+			if(old_priv!= privilege_maskInt || ((old_priv|privilege_maskInt) != old_priv))
```
KushaalShroff: Why do we need the first condition at all?

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -3714,75 +3748,63 @@ exec_stmt_grantschema(PLtsql_execstate *estate, PLtsql_stmt_grantschema *stmt)
...
+			exec_gen_grantschema(schema_name, rolname, stmt, "delete");
+		if((privilege_maskInt & PRIVILEGE_BIT_FOR_REFERENCES) == PRIVILEGE_BIT_FOR_REFERENCES)
+			exec_gen_grantschema(schema_name, rolname, stmt, "references");
+
+		/* Add entry for each grant statement. */
+		if (stmt->is_grant && !check_bbf_schema_for_entry(stmt->schema_name, PERMISSIONS_FOR_ALL_OBJECTS_IN_SCHEMA, rolname))
+			add_entry_to_bbf_schema(stmt->schema_name, PERMISSIONS_FOR_ALL_OBJECTS_IN_SCHEMA,  privilege_maskInt, rolname, NULL);
+		else if(stmt->is_grant)
+		{
+			int16 old_priv = get_bbf_schema_privilege(stmt->schema_name, PERMISSIONS_FOR_ALL_OBJECTS_IN_SCHEMA, rolname);
```
KushaalShroff: Isnt old_priv already defined outside the block? :)

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -3714,75 +3748,63 @@ exec_stmt_grantschema(PLtsql_execstate *estate, PLtsql_stmt_grantschema *stmt)
...
+			if(old_priv!= privilege_maskInt || ((old_priv|privilege_maskInt) != old_priv))
+				update_bbf_schema_privilege(stmt->schema_name, PERMISSIONS_FOR_ALL_OBJECTS_IN_SCHEMA, privilege_maskInt, old_priv, rolname, NULL, stmt->is_grant);
 		}
+		/* Remove entry for each revoke statement. */
+		else if (!stmt->is_grant && check_bbf_schema_for_entry(stmt->schema_name, PERMISSIONS_FOR_ALL_OBJECTS_IN_SCHEMA, rolname))
+		{
+			/* If any object in the schema has the OBJECT level permission. Then, internally grant that permission back. */
+			grant_perms_to_objects_in_schema(stmt->schema_name, privilege_maskInt, rolname);
+			old_priv = get_bbf_schema_privilege(stmt->schema_name, PERMISSIONS_FOR_ALL_OBJECTS_IN_SCHEMA, rolname);
```
KushaalShroff: Same statement in both conditions can be/should be executed outside

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -3714,75 +3748,63 @@ exec_stmt_grantschema(PLtsql_execstate *estate, PLtsql_stmt_grantschema *stmt)
...
+			if(old_priv!= privilege_maskInt || ((old_priv)&(~privilege_maskInt)) != old_priv)
```
KushaalShroff: This condition has totally confused me😅
Can you explain? Also requires comments!

KushaalShroff: Why cant we simply check `old_priv & privilege_maskInt` Is not 0?
If it is non-zero then obviously we have to revoke the intersection of permissions!

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -2880,16 +2880,181 @@ add_entry_to_bbf_schema(const char *schema_name,
...
+	{
+		/*
+		 * In case of GRANT, we add the new privilege along with the previous privilege in the column.
+		 */
+		permission = new_priv | old_priv;
+	}
+	else
+	{
+		/*
+		 * In case of REVOKE, we remove the new privilege and keep the previous privilege as it is.
```
KushaalShroff, line 2976: Why are we calling it new privilege then?

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -2880,16 +2880,181 @@ add_entry_to_bbf_schema(const char *schema_name,
...
+	ScanKeyEntryInitialize(&scanKey[4], 0,
+				Anum_bbf_schema_perms_grantee,
+				BTEqualStrategyNumber,
+				InvalidOid,
+				tsql_get_server_collation_oid_internal(false),
+				F_NAMEEQ,
+				CStringGetDatum(grantee));
+
+	scan = table_beginscan_catalog(bbf_schema_rel, 5, scanKey);
+	tuple_bbf_schema = heap_getnext(scan, ForwardScanDirection);
```
KushaalShroff, line 3021: sys, right?

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -2880,16 +2880,181 @@ add_entry_to_bbf_schema(const char *schema_name,
...
+/*
+ * Update permission column in the catalog when privilege on an object is changed.
+ */
+void
+update_bbf_schema_privilege(const char *schema_name,
+						const char *object_name,
+						int16 new_priv,
+						int16 old_priv,
+						const char *grantee,
+						const char *object_type,
```
KushaalShroff, line 2951: I cannot the use of object_type, why is it needed?
Was there no compiler warning at the time of build?

contrib/babelfishpg_tsql/src/catalog.c
shalinilohia50: Not needed here.

contrib/babelfishpg_tsql/sql/ownership.sql
```diff
@@ -19,10 +19,10 @@ CREATE TABLE sys.babelfish_schema_permissions (
...
   grantee NAME NOT NULL COLLATE sys.database_default,
   object_type NAME COLLATE sys.database_default,
-  PRIMARY KEY(dbid, schema_name, object_name, permission, grantee)
+  PRIMARY KEY(dbid, schema_name, object_name, grantee)
```
KushaalShroff, line 25: OBJECT_TYPE should be part of primary key too, right?

contrib/babelfishpg_tsql/src/catalog.c
shalinilohia50: Addressed it.

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -2880,16 +2880,181 @@ add_entry_to_bbf_schema(const char *schema_name,
...
+				BTEqualStrategyNumber, F_INT2EQ,
+				Int16GetDatum(dbid));
+	ScanKeyEntryInitialize(&scanKey[1], 0,
+				Anum_bbf_schema_perms_schema_name,
+				BTEqualStrategyNumber,
+				InvalidOid,
+				tsql_get_server_collation_oid_internal(false),
+				F_NAMEEQ,
+				CStringGetDatum(schema_name));
+	ScanKeyEntryInitialize(&scanKey[2], 0,
```
KushaalShroff, line 3001: We need to scan based on object type too!!!

How are you going to distinguish between a table and procedure?
```
1> create table t (a int)
2> go

1> create procedure t as select 1
2> go
```

contrib/babelfishpg_tsql/src/pl_handler.c
shalinilohia50: One of the CREATE DATABASE subcommands was `GRANT SELECT ON dummy.sysdatabases TO dummy;` which we avoided here. 
With this [commit](https://github.com/babelfish-for-postgresql/babelfish_extensions/commit/08c886f889461c0e9e71f081167d06ba9d56be1d), looks like it's not needed anymore.

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -2880,16 +2880,181 @@ add_entry_to_bbf_schema(const char *schema_name,
...
+				tsql_get_server_collation_oid_internal(false),
+				F_NAMEEQ,
+				CStringGetDatum(object_name));
+	ScanKeyEntryInitialize(&scanKey[3], 0,
+				Anum_bbf_schema_perms_grantee,
+				BTEqualStrategyNumber,
+				InvalidOid,
+				tsql_get_server_collation_oid_internal(false),
+				F_NAMEEQ,
+				CStringGetDatum(grantee));
```
KushaalShroff, line 2924: Based on this logic you are going to return the first object which matches irrespective of the object_type.
What if the first one was a table but user was granting for a procedure?

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -3714,75 +3748,63 @@ exec_stmt_grantschema(PLtsql_execstate *estate, PLtsql_stmt_grantschema *stmt)
...
+		if((privilege_maskInt & PRIVILEGE_BIT_FOR_UPDATE) == PRIVILEGE_BIT_FOR_UPDATE)
+			exec_gen_grantschema(schema_name, rolname, stmt, "update");
+		if((privilege_maskInt & PRIVILEGE_BIT_FOR_DELETE) == PRIVILEGE_BIT_FOR_DELETE)
+			exec_gen_grantschema(schema_name, rolname, stmt, "delete");
+		if((privilege_maskInt & PRIVILEGE_BIT_FOR_REFERENCES) == PRIVILEGE_BIT_FOR_REFERENCES)
+			exec_gen_grantschema(schema_name, rolname, stmt, "references");
+
+		/* Add entry for each grant statement. */
+		if (stmt->is_grant && !check_bbf_schema_for_entry(stmt->schema_name, PERMISSIONS_FOR_ALL_OBJECTS_IN_SCHEMA, rolname))
+			add_entry_to_bbf_schema(stmt->schema_name, PERMISSIONS_FOR_ALL_OBJECTS_IN_SCHEMA,  privilege_maskInt, rolname, NULL);
```
KushaalShroff: Why are you inserting NULL for object_type?

KushaalShroff (changes requested):
I came across a flaw with the GRANT/REVOKE on schema code/design. In the following example you see Babelfish allows us to create objects with same names and also overload procedures and functions:
```
1> create table t (a int)
2> go

1> create procedure t as select 1
2> go

1>  create procedure t(@a int) as select @a
2> go
```
1. You are not scanning/looking up  based on object_type at multiple places in the code.
a. You are going to return the first matching entry in table irrespective of the object_type.
b. You are inserting NULL for object_type at [one place](https://github.com/babelfish-for-postgresql/babelfish_extensions/blob/a695d88e21a3b0418f2d49a1539a759e48be7d19/contrib/babelfishpg_tsql/src/pl_exec-2.c#L3776), why?
2. How are you gonna support Grants/Revokes on overloaded procedures and functions?
a. In the above example One grant/revoke statement will affect unintended objects, that seems like a critical bug to me
3. We need to add these test cases

contrib/babelfishpg_tsql/src/pltsql_utils.c
anju15bharti: I tried storing bitmask during parsing but then to do this change was eventually leading change in  structures' definition of postgres, hence we need to store strings only during parsing.

contrib/babelfishpg_tsql/src/catalog.c
kuntalghosh: Yeah, in some place, it's a heap_getnext and in some place it's a sys scan.

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -3133,6 +3283,58 @@ clean_up_bbf_schema(const char *schema_name,
...
+				"(GRANT STATEMENT )",
+				false,
+				PROCESS_UTILITY_SUBCOMMAND,
+				NULL,
+				NULL,
+				None_Receiver,
+				NULL);
+
+	/* make sure later steps can see the object created here */
+	CommandCounterIncrement();
```
kuntalghosh, line 3332: Whats the next step?

contrib/babelfishpg_tsql/src/catalog.c
kuntalghosh: But, in a single iteration of the loop, the object type won't change, right?

kuntalghosh: In case of an error, the table will be closed automatically unless we catch the error explicitly.

contrib/babelfishpg_tsql/src/pl_exec-2.c
kuntalghosh: @shalinilohia50 But, for a single exec_stmt_grantschema execution, the object type won't change, right?

contrib/babelfishpg_tsql/src/pl_handler.c
kuntalghosh: Okay. But, exec_stmt_grantschema also updates the catalog. So, do you need to update it here as well?

contrib/babelfishpg_tsql/src/pltsql_utils.c
kuntalghosh: As I mentioned, use ACL [defined](https://github.com/babelfish-for-postgresql/postgresql_modified_for_babelfish/blob/BABEL_3_X_DEV__PG_15_X/src/include/nodes/parsenodes.h#L82) in PG. It provides a lot of APIs for these.

PR #2045, opened 2023-11-21
Fix bugs for translation logic for simple_term and execution logic for FULLTEXT INDEX stmts

roshan0708:
### Description

Bug fixes for `simple_term` support in CONTAINS clause and FULLTEXT INDEX creation/deletion statements:
- Fix the buffer overflow issue on translation of `simple_term`.
- Fix multiple spaces issue between words in `simple_term` which caused exception.
- Remove the util function `exec_utility_cmd_helper` and use a normal wrapper.

### Issues Resolved

JIRA: BABEL-4379, BABEL-4383
Signed-off-by: Roshan Kanwar [rskanwar@amazon.com](mailto:rskanwar@amazon.com)

* **Use case based -** Added

* **Boundary conditions -**

* **Arbitrary inputs -**

* **Negative test cases -**

* **Performance tests -**

* **Tooling impact -**

* **Client tests -**

### Check List


contrib/babelfishpg_tsql/src/fts_parser.y
```diff
@@ -176,13 +201,11 @@ static char *trim(char *s) {
     size_t end;
     size_t newLength;
 
-    if (s == NULL) {
-        return NULL;
-    }
-
     length = strlen(s);
```
Sairakan: strlen() will fail on a NULL input - you need to check for NULL before calling this

contrib/babelfishpg_tsql/src/fts_parser.y
```diff
@@ -151,21 +160,37 @@ char* translate_simple_term(const char* inputStr) {
...
                 *outputPtr++ = '>';
             } else {
                 // Copy the character
+                if (outputPtr - output + 1 > outputSize) {
+                    // Output buffer overflow
+                    pfree(trimmedInputStr);
+                    pfree(output);
+                    return NULL;
+                }
                 *outputPtr++ = *inputPtr;
```
Sairakan: When do you copy the null terminator over? Right now as far as I can tell you never set it, and just rely that the palloc() call puts a null terminator at the end, which is not correct.

contrib/babelfishpg_tsql/src/fts_parser.y
```diff
@@ -127,20 +126,30 @@ char* translate_simple_term(const char* inputStr) {
...
-    strcpy(trimmedInputStr, inputStr);
+    trimmedInputStr = pstrdup(inputStr);
 
     // removing trailing and leading spaces
     trim(trimmedInputStr);
     inputLength = strlen(trimmedInputStr);
-    
+
     // Check if the input is a phrase enclosed in double quotes
     if (trimmedInputStr[0] == '"' && trimmedInputStr[inputLength - 1] == '"') {
```
Sairakan, line 136: What if inputStr is an empty string (e.g. length == 0)? You should check for that case

contrib/babelfishpg_tsql/src/fts_parser.y
```diff
@@ -219,13 +242,11 @@ static char *trimInsideQuotes(char *s) {
     size_t newLength;
     bool insideQuotes;
 
-    if (s == NULL) {
-        return NULL;
-    }
-
     length = strlen(s);
```
Sairakan: same as above

contrib/babelfishpg_tsql/src/fts_parser.y
Sairakan: Otherwise you might try to get trimmedInputStr[-1] which is incorrect

contrib/babelfishpg_tsql/src/fts_parser.y
```diff
@@ -123,24 +122,34 @@ char* translate_simple_term(const char* inputStr) {
     char* trimmedInputStr;
 
     // Check for empty input
-    if (inputStr == NULL) {
+    if (!inputStr || !(inputLength = strlen(inputStr))) {
         return NULL;
```
Sairakan: Have you validated that the consumers of this function are capable of handling NULL values? I just checked, and it looks like SQL Server actually just returns an error on a NULL or empty input, so we should probably just copy that behavior
```
1> select * from t1 where contains(data, '')
2> go
Msg 7645, Level 15, State 1, Server EC2AMAZ-R8NSHGO, Line 1
Null or empty full-text predicate.
```

roshan0708: Will check on this.

contrib/babelfishpg_tsql/src/fts_parser.y
```diff
@@ -123,24 +122,34 @@ char* translate_simple_term(const char* inputStr) {
...
         inputLength = strlen(trimmedInputStr);
-        outputSize = inputLength - 2; // Exclude both quotes
+
+        // Calculate the maximum possible size of output
+        outputSize = inputLength * 3;
+
+        // Check for potential overflow and adjust the output size
+        if (outputSize < inputLength || outputSize < 0) {
+            pfree(trimmedInputStr);
+            return NULL; // Potential overflow
```
Sairakan: this is an exceptional situation, so I think it's better to just throw an error - especially since I don't think the callers of this function are currently designed to handle NULL return values anyways

contrib/babelfishpg_tsql/src/fts_parser.y
```diff
@@ -123,24 +122,34 @@ char* translate_simple_term(const char* inputStr) {
...
+        }
+
+        // Allocate the output buffer with the adjusted size
         output = (char*)palloc(outputSize + 1); // +1 for the null terminator
-        if (output == NULL) {   
+        if (output == NULL) {
+            pfree(trimmedInputStr);
             return NULL;
```
Sairakan: Same as above - although output failing to allocate shouldn't happen in the first place (I believe palloc would throw a FATAL before doing so anyways)

contrib/babelfishpg_tsql/src/fts_parser.y
```diff
@@ -151,21 +160,37 @@ char* translate_simple_term(const char* inputStr) {
...
                 // Replace space with "<->"
+                while (*(inputPtr + 1) == ' ') {
+                    // Handle multiples spaces between words and skip over additional spaces
+                    inputPtr++;
+                }
+                if (outputPtr - output + 3 > outputSize) {
+                    // Output buffer overflow
+                    pfree(trimmedInputStr);
+                    pfree(output);
+                    return NULL;
```
Sairakan: same as above, just throw an error

contrib/babelfishpg_tsql/src/fts_parser.y
```diff
@@ -151,21 +160,37 @@ char* translate_simple_term(const char* inputStr) {
...
                 *outputPtr++ = '<';
                 *outputPtr++ = '-';
                 *outputPtr++ = '>';
             } else {
                 // Copy the character
+                if (outputPtr - output + 1 > outputSize) {
+                    // Output buffer overflow
+                    pfree(trimmedInputStr);
+                    pfree(output);
+                    return NULL;
```
Sairakan: See above

Sairakan (approved):

contrib/babelfishpg_tsql/src/fts_parser.y
Sairakan: Unless you are absolutely sure that the consumers of this function can handle NULL, this should be replaced with an ereport()

contrib/babelfishpg_tsql/src/fts_scan.l
```diff
@@ -72,7 +80,7 @@ fts_scanner_init(const char *str)
 	memcpy(scanbuf, str, slen);
 	scanbuf[slen] = scanbuf[slen + 1] = YY_END_OF_BUFFER_CHAR;
 	scanbufhandle = yy_scan_buffer(scanbuf, slen + 2);
-
+
```
Sairakan: nit: blank space

contrib/babelfishpg_tsql/src/fts_parser.y
```diff
@@ -219,13 +248,9 @@ static char *trimInsideQuotes(char *s) {
...
     bool insideQuotes;
 
-    if (s == NULL) {
-        return NULL;
-    }
-
-    length = strlen(s);
-    if (length == 0) {
-        return s; // Empty string, nothing to trim
+     // Empty string, nothing to trim
```
Sairakan: it looks like trim() and trimInsideQuotes() are very similar functions, can they be combined somehow?

HarshLunagariya (approved):

jsudrik (approved):
Based on other approvals.

PR #2162, opened 2023-12-20
Restrict privileges from Unauthorised TSQL logins

shalinilohia50:
### Description

1. Avoid granting **CREATEROLE** and **CREATEDB** privilege to non-sysadmins logins
2. Manage **CREATEDB/CREATEROLE** privileges as part of grant/revoke membership to/from **sysadmin** via TDS Port only. Though the `grant sysadmin to user` works from psql endpoint for superuser, it will not add CREATEDB/CREATEROLE privileges. If a TSQL user wants to have the sysadmin membership and CREATEDB /CREATEROLE privileges, it should alter the server role via TDS port. 

### Issues Resolved

1. Any unprivileged Babelfish role should not grant/revoke sysadmin role or non-Babelfish roles to itself and to others from the PG port.
2. Any unprivileged Babelfish role should not drop any role via PG port.
3. Any unprivileged Babelfish role should not alter any role via PG port.
6. Any unprivileged Babelfish role should not create any role via PG port.
7. Restrict PG user to "grant sysadmin to user" to any user via PG port.

Task: [BABEL-4573](https://jira.rds.a2z.com/browse/BABEL-4573), [BABEL-4574](https://jira.rds.a2z.com/browse/BABEL-4574)

* **Use case based - Added**
Tests for Point 5:
```
dev-dsk-lshalini-1a-142a051b % ./bin/psql -U u1 -d postgres      
psql (15.5)
Type "help" for help.

postgres=> grant sysadmin to u1;
ERROR:  must have admin option on role "sysadmin"
postgres=> grant sysadmin to l1;
ERROR:  must have admin option on role "sysadmin"
postgres=> create role r1;
ERROR:  permission denied to create role
```

* **Boundary conditions -**

* **Arbitrary inputs -**

* **Negative test cases -**

* **Performance tests -**

* **Tooling impact -**

* **Client tests -**


contrib/babelfishpg_tsql/src/backend_parser/gram-tsql-rule.y
```diff
@@ -33,11 +33,11 @@ tsql_CreateLoginStmt:
 											@1)); /* Must be first */
 					n->options = lappend(n->options,
 										 makeDefElem("createdb",
-													 (Node *)makeBoolean(true),
+													 (Node *)makeBoolean(false),
 													 @1));
 					n->options = lappend(n->options,
 										 makeDefElem("createrole",
-													 (Node *)makeBoolean(true),
+													 (Node *)makeBoolean(false),
```
xhfanhe, line 40: Does this "createrole" permission control create/drop/alter role?

contrib/babelfishpg_tsql/src/backend_parser/gram-tsql-rule.y
```diff
@@ -33,11 +33,11 @@ tsql_CreateLoginStmt:
 											@1)); /* Must be first */
 					n->options = lappend(n->options,
 										 makeDefElem("createdb",
-													 (Node *)makeBoolean(true),
+													 (Node *)makeBoolean(false),
```
xhfanhe, line 36: After this change, does sysadmin still have createdb/createrole permission?

contrib/babelfishpg_tsql/src/backend_parser/gram-tsql-rule.y
shalinilohia50: Yes, now "createrole" permission will be granted to a login only via the TDS port using the below statement.
`alter server role sysadmin add member login1`

contrib/babelfishpg_tsql/src/backend_parser/gram-tsql-rule.y
shalinilohia50: sysadmin will continue having createdb/createrole permission, this change restricts any new login which is created via TDS port to have these permissions by default.

contrib/babelfishpg_tsql/src/pltsql_utils.c
```diff
@@ -1901,3 +1901,52 @@ char
...
 	return query.data;
 }
+
+/*
+ * Helper function to execute ALTER ROLE command using
+ * ProcessUtility(). Caller should make sure their
+ * inputs are sanitized to prevent unexpected behaviour.
+ */
+void
+exec_alter_role_cmd(char *query_str, RoleSpec *role)
```
rishabhtanwar29, line 1911: There is pre-existing function `exec_utility_cmd_helper` to execute the utility commands, I think we should use that instead with a little modification that it can take `Node*` as an argument. ie.
```
exec_utility_cmd_helper(char *query_str, Node *obj)
```
And then we can have switch case on Node type to update the generated parsetree depending upon the its type.

test/JDBC/expected/restricting_permission_on_role.out
```diff
@@ -0,0 +1,167 @@
+-- tsql
```
rishabhtanwar29: Although I think we should add these tests into existing `ownership_restrictions_from_pg` flle but if you really want to have a new file then I would recommend to name this as `permission_restrictions_from_pg` to keep the naming style similar.

test/JDBC/expected/restricting_permission_on_role.out
```diff
@@ -0,0 +1,167 @@
...
+create login restricting_permission_on_role_l1 with password = '123';
```
rishabhtanwar29: Let's have `tsql_login` in the above login's name and `psql_user` in psql user below for better clarity.

test/JDBC/expected/restricting_permission_on_role.out
```diff
@@ -0,0 +1,167 @@
...
+-- Grant sysadmin privilege to underprivileged T-SQL user
+grant sysadmin to restricting_permission_on_role_l1;
+go
+
+-- Grant sysadmin privilege to underprivileged PG user
+grant sysadmin to restricting_permission_on_role_u1;
+go
+
+-- psql user=restricting_permission_on_role_l1 password=123
+-- user has sysadmin membership via PG port, create user is not allowed
```
rishabhtanwar29: At here and all other places, Let's use the terms tsql login or psql user to avoid confusion about which one we talking.

contrib/babelfishpg_tsql/src/pltsql_utils.c
xhfanhe: Fyi the current `exec_utility_cmd_helper` has SQL injection risk for directly executing query string with statement number check (which is insufficient). We should always parse dummy query first and then update the Node with real value. 
But it is a good idea to integrate all query parsing/update into one function, as Rishabh mentioned.
There was a slack thread about this a month ago. I would suggest leave this change to a separate PR, but will be great to file a Jira for this.

shalinilohia50: Ack. I will create a separate JIRA for this. Because that would need changes at multiple places wherever `exec_utility_cmd_helper` is being used.

shalinilohia50: https://jira.rds.a2z.com/browse/BABEL-4649 It can be tracked here.

contrib/babelfishpg_tsql/src/rolecmds.c
```diff
@@ -1562,15 +1562,12 @@ is_alter_server_stmt(GrantRoleStmt *stmt)
...
 
-		if (strcmp(spec->rolename, "sysadmin") != 0)	/* only supported server
+		if (strcmp(spec->rolename, "sysadmin") == 0)	/* only supported server
 														 * role */
-			return false;
+			return true;
 	}
-	/* has one and only one grantee  */
-	if (list_length(stmt->grantee_roles) != 1)
-		return false;
```
HarshLunagariya: We shouldn't be removing this check.

contrib/babelfishpg_tsql/src/rolecmds.c
```diff
@@ -2342,3 +2339,48 @@ check_windows_logon_length(char *input)
...
+	rel = table_open(get_authid_login_ext_oid(), AccessShareLock);
+	scan = table_beginscan_catalog(rel, 0, NULL);
+	tuple = heap_getnext(scan, ForwardScanDirection);
+
+	while (HeapTupleIsValid(tuple))
+	{
+		Form_authid_login_ext loginform;
+		char *rolname;
+		loginform = (Form_authid_login_ext) GETSTRUCT(tuple);
+		rolname = pstrdup(NameStr(loginform->rolname));
```
HarshLunagariya, line 2363: pfree rolname

contrib/babelfishpg_tsql/src/rolecmds.c
```diff
@@ -2342,3 +2339,48 @@ check_windows_logon_length(char *input)
...
+
+		/*
+		 * For each login (except sysadmin and the member of sysadmin), remove
+		 * createrole and createdb privileges from the logins.
+		 */
+		if ((strcmp(rolname, "sysadmin") != 0) && !has_privs_of_role(get_role_oid(rolname, false), get_role_oid("sysadmin", false)))
```
HarshLunagariya: Use get_sysadmin_oid() for better performance.

contrib/babelfishpg_tds/src/backend/tds/tdsutils.c
```diff
@@ -1091,10 +1091,8 @@ handle_alter_role(AlterRoleStmt* alter_role_stmt)
 		    }
 		    else
 		    {
-			    if ((authForm && authForm->rolcanlogin) && (strcmp(defel->defname, "password") == 0 ||
-					    strcmp(defel->defname, "connectionlimit") == 0 ||
-					    strcmp(defel->defname, "validUntil") == 0))
```
HarshLunagariya: why are we not blocking this now? In the past, We allowed it because there was no way to alter these properties from TDS endpoint.

contrib/babelfishpg_tsql/src/rolecmds.c
shalinilohia50: done

contrib/babelfishpg_tds/src/backend/tds/tdsutils.c
shalinilohia50: Reverted this change as it's not needed right now.

HarshLunagariya (approved):

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--3.4.0--3.5.0.sql
```diff
@@ -118,6 +118,12 @@ AS 'babelfishpg_tsql', 'session_context'
 LANGUAGE C;
 GRANT EXECUTE ON FUNCTION sys.session_context TO PUBLIC;
 
+-- Update existing logins to remove createrole privilege
+CREATE OR REPLACE PROCEDURE sys.bbf_remove_createrole_from_logins()
```
rishabhtanwar29, line 122: We can DROP this function after use.

test/JDBC/input/permission_restrictions_from_pg-vu-prepare.mix
```diff
@@ -0,0 +1,6 @@
+-- tsql
+create login permission_restrictions_tsql_login with password = '123';
+go
+
+select rolname, rolcreaterole, rolcreatedb from pg_roles where rolname = 'permission_restrictions_tsql_login';
+go
```
rishabhtanwar29, line 6: Nit: Let's add a new line at the end of each new file.

rishabhtanwar29 (approved):

shardgupta (approved):

PR #2164, opened 2023-12-21
Unprivileged T-SQL logins should not create users and roles

shalinilohia50:
### Description

An unprivileged T-SQL login should not be allowed to create roles and users in Babelfish.

### Issues Resolved

Task: [BABEL-4646](https://jira.rds.a2z.com/browse/BABEL-4646),

* **Use case based - Added**

* **Boundary conditions -**

* **Arbitrary inputs -**

* **Negative test cases -**

* **Performance tests -**

* **Tooling impact -**

* **Client tests -**

### DIFF
--- a/contrib/babelfishpg_tsql/src/pl_handler.c
+++ b/contrib/babelfishpg_tsql/src/pl_handler.c
@@ -2653,7 +2653,10 @@ bbf_ProcessUtility(PlannedStmt *pstmt,
 					}
 					else if (isuser || isrole)
 					{
-						if (!has_privs_of_role(GetSessionUserId(), get_role_oid("sysadmin", false)))
+						const char *db_owner_name;
+
+						db_owner_name = get_db_owner_name(get_cur_db_name());
+						if (!has_privs_of_role(GetUserId(),get_role_oid(db_owner_name, false)))
 							ereport(ERROR,
 									(errcode(ERRCODE_INSUFFICIENT_PRIVILEGE),
 									 errmsg("User does not have permission to perform this action.")));
--- a/contrib/babelfishpg_tsql/src/pl_handler.c
+++ b/contrib/babelfishpg_tsql/src/pl_handler.c
@@ -2653,6 +2653,10 @@ bbf_ProcessUtility(PlannedStmt *pstmt,
 					}
 					else if (isuser || isrole)
 					{
+						if (!has_privs_of_role(GetSessionUserId(), get_role_oid("sysadmin", false)))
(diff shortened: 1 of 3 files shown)


contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -2653,6 +2653,10 @@ bbf_ProcessUtility(PlannedStmt *pstmt,
 					}
 					else if (isuser || isrole)
 					{
+						if (!has_privs_of_role(GetSessionUserId(), get_role_oid("sysadmin", false)))
+							ereport(ERROR,
+									(errcode(ERRCODE_INSUFFICIENT_PRIVILEGE),
+									 errmsg("User does not have permission to perform this action.")));
```
HarshLunagariya: Here, we are missing following scenario.
```
create login l1;
grant sysadmin to l1;
-- connect using l1
create database db1;

revoke sysadmin from l1;
-- Now login l1 is still owns db1 even if it's not sysadmin, so if we login using l1 in db1 then we will get dbo user
-- connect using l1 in db1
select current_user() -> will be dbo

and since it is dbo, we will be allowed to drop/create/alter role/users in current database db1.
```
Instead, We should verify that current user is dbo or not (OR member of db_owner or not).

HarshLunagariya (changes requested):

contrib/babelfishpg_tsql/src/pl_handler.c
shalinilohia50: What you are saying is related to drop roles/users. 
This change only affects "create role" and "create user" statement.
Problem statement is to restrict T-SQL logins to create users and roles.

HarshLunagariya: Even creating role/users will be allowed in that case.

rishabhtanwar29 (approved):

contrib/babelfishpg_tsql/src/pl_handler.c
xhfanhe: I'm all good with the change apart from this, would you please add test for this and check if it matches expected behavior?

shalinilohia50: There is an existing issue which will be tracked here. https://jira.rds.a2z.com/browse/BABEL-4652

xhfanhe (approved):

HarshLunagariya (approved):

shardgupta (approved):

PR #2183, opened 2023-12-26
Fix SPI clean up incase of errors for insert into execute with @tablevarible

tanscorpio7:
### Description

Fix crash in insert into @tablevariable execute('some error generating statement')
Problem was how we were handling SPI connections clean up.
We marked the SPI connections as internal xact only when we were not inside a sys function or non tsql procedure.
But we always call SPI finish in following terminate batch. Marking it non internal will clean up the SPI connections when we abort transaction in error handling.

As a fix we always mark the SPI connection as internal_xact for pltsql_inline_handler and pltsql_call_handler. Also clean up any leftovers SPI connections in terminate_batch. If we are less than expected spi stack depth we throw a fatal error to kill session since this state is never expected.

Also make some changes to fix static code analyzer warnings for uninitialised variables in pltsql call handler.

### Issues Resolved

[BABEL-4360]

### Sign Off

### Check List


contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -4339,34 +4351,28 @@ pltsql_call_handler(PG_FUNCTION_ARGS)
...
 
-	func->use_count--;
-
-	func->cur_estate = save_cur_estate;
-
-	pltsql_remove_current_query_env();
-	pltsql_revert_guc(save_nestlevel);
-	pltsql_revert_last_scope_identity(scope_level);
-
-	terminate_batch(false /* send_error */ , false /* compile_error */ );
```
tanscorpio7, line 4369: these changes fix static code analyzer warnings.
Uninitialized variable error is thrown for func, save_nestlevel, save_nestlevel 

for example 
contrib/babelfishpg_tsql/src/pl_handler.c:4374:2: warning: Uninitialized variable: func [uninitvar]
 func->cur_estate = save_cur_estate;
 ^
contrib/babelfishpg_tsql/src/pl_handler.c:4284:64: note: Assuming condition is false
 if ((rc = SPI_connect_ext(nonatomic ? SPI_OPT_NONATOMIC : 0)) != SPI_OK_CONNECT)
                                                               ^
contrib/babelfishpg_tsql/src/pl_handler.c:4293:2: note: Assuming condition is false
 PG_TRY();
 ^

contrib/babelfishpg_tsql/src/iterative_exec.c
```diff
@@ -1306,13 +1306,16 @@ dispatch_stmt_handle_error(PLtsql_execstate *estate,
 			{
 				if (is_part_of_pltsql_trycatch_block(estate))
 				{
+					bool 	spi_current_internal_xact_mode;
+
 					HOLD_INTERRUPTS();
 					elog(DEBUG1, "TSQL TXN PG semantics : Rollback current transaction");
 					HoldPinnedPortals();
+					SPI_getCurrentInternalTxnMode(&spi_current_internal_xact_mode);
```
surendravishnoi: do we need this given that we are unconditionally setting internal xact mode?

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -4088,15 +4088,24 @@ _PG_fini(void)
...
+	int 		current_spi_stack_depth;
 
 	HOLD_INTERRUPTS();
 
 	elog(DEBUG2, "TSQL TXN finish current batch, error : %d compilation error : %d", send_error, compile_error);
 
+	/*
+	 * Cleanup leftovers non internal SPI connections
+	 */
+	AtEOXact_SPI(false);
```
surendravishnoi: we do not want to do this. It will cleanup all SPI connections up do the point internal xact mode is set. It can leave behind SPI connections which has are internal but have depth higher than current depth. Rather, we want to call cleanup/finish up to given SPI depth.

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -4088,15 +4088,24 @@ _PG_fini(void)
...
+	SPI_get_depth(&current_spi_stack_depth);
+
+	Assert(current_spi_stack_depth == SPI_depth);
```
surendravishnoi: not sure if Asserts are enabled in production or not. Better to raise a fatal error.

contrib/babelfishpg_tsql/src/iterative_exec.c
tanscorpio7: Other wise Assert(current_spi_stack_depth == SPI_depth); fails for EXEC psql_interop_proc; (nested procedure tsql --> psql --> tsql)

This could indicate that we could enter this IF block with the top of SPI stack holding a tsql SPI connection, which is a internal xact = true, we do not want to mark this as false.

contrib/babelfishpg_tsql/src/pl_handler.c
tanscorpio7: Will make this change, although only babelfish will set internal xact = true. Postgres does not use this except inside SPI APIs, that too temporarily.

contrib/babelfishpg_tsql/src/pl_handler.c
tanscorpio7: Ok

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -4088,20 +4088,30 @@ _PG_fini(void)
...
 	 */
-	if ((rc = SPI_finish()) != SPI_OK_FINISH)
-		elog(ERROR, "SPI_finish failed: %s", SPI_result_code_string(rc));
+	SPI_get_depth(&current_spi_stack_depth);
+	
+	if (current_spi_stack_depth < SPI_depth)
+		elog(FATAL, "SPI stack corrupted");
+	
+	while (current_spi_stack_depth-- >= SPI_depth)
+		if ((rc = SPI_finish()) != SPI_OK_FINISH)
```
surendravishnoi, line 4119: please emit a WARNING message that we have found leaked SPI connection

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -4088,20 +4088,30 @@ _PG_fini(void)
...
+	 * Also cleanup remnant SPI connections
+	 * Ideally current depth should be same as 
+	 * when caller was connecting to SPI Manager
 	 */
-	if ((rc = SPI_finish()) != SPI_OK_FINISH)
-		elog(ERROR, "SPI_finish failed: %s", SPI_result_code_string(rc));
+	SPI_get_depth(&current_spi_stack_depth);
+	
+	if (current_spi_stack_depth < SPI_depth)
+		elog(FATAL, "SPI stack corrupted");
```
surendravishnoi: we should add more details to error message like expected connection count and current count. Also, message should be "SPI connection stack is inconsistent"

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -4088,20 +4088,30 @@ _PG_fini(void)
...
 
 	/*
 	 * Disconnect from SPI manager
+	 * Also cleanup remnant SPI connections
+	 * Ideally current depth should be same as 
+	 * when caller was connecting to SPI Manager
 	 */
-	if ((rc = SPI_finish()) != SPI_OK_FINISH)
-		elog(ERROR, "SPI_finish failed: %s", SPI_result_code_string(rc));
+	SPI_get_depth(&current_spi_stack_depth);
```
surendravishnoi: we should manually test both cases, fatal error and warning for leak

test/JDBC/expected/BABEL_4360.out
```diff
@@ -0,0 +1,159 @@
...
+INSERT INTO babel_4360_t VALUES (93);
+GO
+
+CREATE PROCEDURE tsql_interop_proc4
+AS
+EXEC tsql_interop_proc3;
+INSERT INTO babel_4360_t VALUES (92);
+GO
+
+EXEC tsql_interop_proc4;
```
surendravishnoi, line 88: what happens without the fix? do we have test cases for both error and leak to validate fix?

contrib/babelfishpg_tsql/src/iterative_exec.c
tanscorpio7: removed all usage of spi_current_internal_xact_mode from catch block here. Could not find any case where we would need to protect spi from deletion here.
Abort Transaction will now cleanup SPI stack as well (until it finds an internal xact), which means we should never get a SPI leak in terminate batch.

test/JDBC/expected/BABEL_4360.out
tanscorpio7: insert into @tablevariable execute ('some error condition') --> SPI stack depth expected greater than SPI stack current (error scenario)

Difficult to test leak, since abort transaction will clean up everything.

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -4339,34 +4358,28 @@ pltsql_call_handler(PG_FUNCTION_ARGS)
...
 		{
 			set_procid(prev_procid);
-			/* Decrement use-count, restore cur_estate, and propagate error */
 			pltsql_trigger_depth = save_pltsql_trigger_depth;
-			func->use_count--;
-			func->cur_estate = save_cur_estate;
-			pltsql_remove_current_query_env();
-			pltsql_revert_guc(save_nestlevel);
-			pltsql_revert_last_scope_identity(scope_level);
-			terminate_batch(true /* send_error */ , false /* compile_error */ );
```
tanscorpio7, line 4349: here we had set send_error = true
which re throws the error in terminate batch, doesn't that mean we never actually hit the next two lines in catch block ?
code coverage seems to indicate so.
cc: @surendravishnoi

contrib/babelfishpg_tsql/src/pl_handler.c
tanscorpio7: done

surendravishnoi (dismissed):

surendravishnoi (approved):

PR #2189, opened 2023-12-28
Fix valgrind warnings related to varcharvarbinary function

basasairohan:
### Description

Currently, we have defined the following cast functions which call varcharvarbinary C function with only 1 argument. But, in varcharvarbinary function, we are reading two arguments which might lead to memory related issues.

1. sys.binarysysvarchar
2. sys.binaryvarchar
3. sys.rowversionsysvarchar
4. sys.rowversionvarchar

This change modifies the above casts to have 3 arguments, one for input, one for typmod (integer) and the other for specifying isExplicit (boolean)

This change also replaces `varcharTypeName->typeOid` with `baseTypeId` in `tsql_coerce_string_literal_hook()` function as we are not setting `varcharTypeName->typeOid` at all which causes failure while trying to fetch the PG view definition as shown below.

Ex : 
```
-- T-SQL endpoint
1> create view babel_4475_v1 as select cast(cast('a' as binary(2)) as varchar(2))
2> go
```

```
-- PSQL endpoint
jdbc_testdb=# select pg_catalog.pg_get_viewdef(oid, true) from pg_class where relname = 'babel_4475_v1';
ERROR:  cache lookup failed for type 0
jdbc_testdb=# 
```

### Issues Resolved

BABEL-4475

* **Use case based -** Tests related to above cast functions have already been written in `babel_datatype.sql` and `BABEL-ROWVERSION.sql`

* **Arbitrary inputs -** N/A

* **Performance tests -** N/A

* **Tooling impact -** N/A

* **Client tests -** N/A

### Check List


contrib/babelfishpg_common/sql/upgrades/babelfishpg_common--2.5.0--2.6.0.sql
```diff
@@ -3,10 +3,189 @@
...
+-- Please have this be one of the first statements executed in this upgrade script. 
+CREATE OR REPLACE PROCEDURE babelfish_drop_deprecated_object(object_type varchar, schema_name varchar, object_name varchar) AS
+$$
+DECLARE
+    error_msg text;
+    query1 text;
+    query2 text;
+BEGIN
+
+    query1 := pg_catalog.format('alter extension babelfishpg_tsql drop %s %s.%s', object_type, schema_name, object_name);
```
Deepesh125: this should be babelfishpg_common otherwise you would get following error -- 
```
WARNING:  function binarysysvarchar_deprecated_2_6_0(bbf_binary) is not a member of extension "babelfishpg_tsql"
WARNING:  function binaryvarchar_deprecated_2_6_0(bbf_binary) is not a member of extension "babelfishpg_tsql"
WARNING:  function rowversionsysvarchar_deprecated_2_6_0(rowversion) is not a member of extension "babelfishpg_tsql"
WARNING:  function rowversionvarchar_deprecated_2_6_0(rowversion) is not a member of extension "babelfishpg_tsql"
```

contrib/babelfishpg_common/sql/upgrades/babelfishpg_common--2.5.0--2.6.0.sql
```diff
@@ -3,10 +3,189 @@
...
+$$
+LANGUAGE plpgsql;
+
 CREATE OR REPLACE FUNCTION sys.varbinarybinary (sys.BBF_VARBINARY, integer, boolean)
 RETURNS sys.BBF_BINARY
 AS 'babelfishpg_common', 'binary'
 LANGUAGE C IMMUTABLE STRICT PARALLEL SAFE;
 
+-- (sys.BBF_BINARY AS sys.VARCHAR)
+DROP CAST IF EXISTS(sys.BBF_BINARY AS sys.VARCHAR);
```
Deepesh125, line 39: what if there is dependent objects? Add test cases, 

-- prepare
create view test as select cast (some_binval as varchar)

-- upgrade to latest 

-- verify 
select * from test

Deepesh125 (changes requested):

contrib/babelfishpg_common/sql/upgrades/babelfishpg_common--2.5.0--2.6.0.sql
```diff
@@ -1,12 +1,191 @@
...
+LANGUAGE C IMMUTABLE STRICT PARALLEL SAFE;
+
+CREATE CAST (sys.BBF_BINARY AS sys.VARCHAR)
+WITH FUNCTION sys.binarysysvarchar (sys.BBF_BINARY, integer, boolean) AS ASSIGNMENT;
+
+DO $$
+DECLARE
+    exception_message text;
+BEGIN
+    CALL sys.babelfish_drop_deprecated_object('function', 'sys', 'binarysysvarchar_deprecated_2_6_0');
```
surendravishnoi: do we need to put this in DO block? babelfish_drop_deprecated_object is already fail safe

contrib/babelfishpg_common/src/varbinary.c
```diff
@@ -732,12 +731,18 @@ varbinaryvarchar(PG_FUNCTION_ARGS)
 	VarChar    *result;
 	char 	   *encoded_result;
 	size_t		len = VARSIZE_ANY_EXHDR(source);
-	int32		typmod = PG_GETARG_INT32(1);
-	int32		maxlen = typmod - VARHDRSZ;
+	int32		typmod = -1;
+	int32		maxlen = -1;
```
surendravishnoi, line 740: is this desired behavior to use typmod and maxlen as -1 by default? do we have test case which validates that case works as expected? have we found why tests were not failing with garbage typmod and maxlen?

contrib/babelfishpg_common/src/varbinary.c
```diff
@@ -52,6 +52,7 @@ PG_FUNCTION_INFO_V1(rowversionvarbinary);
 PG_FUNCTION_INFO_V1(varcharvarbinary);
 PG_FUNCTION_INFO_V1(bpcharvarbinary);
 PG_FUNCTION_INFO_V1(varbinaryvarchar);
+PG_FUNCTION_INFO_V1(varbinaryvarchar_deprecated_2_6_0);
```
surendravishnoi: where is this function defined?

test/JDBC/upgrade/14_9/schedule
```diff
@@ -412,4 +412,5 @@ smalldatetimefromparts-dep
 BABEL_4330
 BABEL-4410
 AUTO_ANALYZE-before-15-5-or-14-10
+babel-4475
```
surendravishnoi, line 415: add to latest schedule as well

test/JDBC/expected/babel-4475-vu-prepare.out
```diff
@@ -0,0 +1,11 @@
+create view babel_4475_v1 as select cast(cast('a' as binary(2)) as varchar(2))
```
surendravishnoi, line 1: we should call the cast from view, function and procedures. Also, we have 4 casts, there should be corresponding 4 tests. We should add a test which validates typmod and maxlen with cases like varchar, varchar(max), varchar(1), varchar(x) and value larger than limit

contrib/babelfishpg_common/src/varbinary.c
basasairohan: If we call this function without typmod (nargs = 1), According to the below condition, we need to call `encoding_conv_util` with the length of the source (len) as argument. Hence, -1 should work fine in this case. 

```
if (maxlen < 0 || len <= maxlen)
     encoded_result = encoding_conv_util(data, len, collInfo.enc, PG_UTF8, &encodedByteLen);
else
     encoded_result = encoding_conv_util(data, maxlen, collInfo.enc, PG_UTF8, &encodedByteLen);
```

The garbage values we were getting before were very large and hence, we will be always calling `encoding_conv_util` with the length of the source (len) as argument. We did not have any tests to test this scenario. Hence, the tests weren't failing before.

test/JDBC/upgrade/14_9/schedule
basasairohan: done

test/JDBC/expected/babel-4475-vu-prepare.out
basasairohan: Added.

contrib/babelfishpg_common/src/varbinary.c
```diff
@@ -732,12 +730,23 @@ varbinaryvarchar(PG_FUNCTION_ARGS)
 	VarChar    *result;
 	char 	   *encoded_result;
 	size_t		len = VARSIZE_ANY_EXHDR(source);
-	int32		typmod = PG_GETARG_INT32(1);
-	int32		maxlen = typmod - VARHDRSZ;
+	int32		typmod = -1;
+	int32		maxlen = -1;
 	coll_info	collInfo;
 	int			encodedByteLen;
 	MemoryContext ccxt = CurrentMemoryContext;
```
surendravishnoi, line 744: 3rd argument is not used at all? We should a comment that there are two callers to this function with 1 and 3 arguments. So, if someone parses 3rd argument tomorrow, they should make sure that number of args are checked before parsing

contrib/babelfishpg_common/src/varbinary.c
surendravishnoi: if we cast 'abcdef' with varchar(2), it should be 'ab' but what will we get if we call function with 1 argument? we can add a test case in upgrade context where we will get incorrect result ('abcdef' instead of 'abc')

contrib/babelfishpg_common/sql/upgrades/babelfishpg_common--2.5.0--2.6.0.sql
```diff
@@ -1,12 +1,147 @@
...
+LANGUAGE C IMMUTABLE STRICT PARALLEL SAFE;
+
+CREATE CAST (sys.ROWVERSION AS pg_catalog.VARCHAR)
+WITH FUNCTION sys.rowversionvarchar (sys.ROWVERSION, integer, boolean) AS ASSIGNMENT;
+
+CALL sys.babelfish_drop_deprecated_object('function', 'sys', 'rowversionvarchar_deprecated_2_6_0');
+
+-- Drops the temporary procedure used by the upgrade script.
+-- Please have this be one of the last statements executed in this upgrade script.
+DROP PROCEDURE sys.babelfish_drop_deprecated_object(varchar, varchar, varchar);
```
Deepesh125, line 144: Have we verified all the possible data type cast to varchar? apart from [var]binary and rowversion?

contrib/babelfishpg_common/src/varbinary.c
Deepesh125: I think do not have a test cases with input string length which is greater than these arbitrary large garbage values. But we should have test cases at least for varchar(max)

test/JDBC/input/babel-4475-vu-verify.sql
```diff
@@ -0,0 +1,66 @@
...
+select cast(cast('abcdef' as binary(6)) as varchar(max))
+go
+
+select cast(cast('abcdef' as binary(6)) as varchar(2))
+go
+
+select cast(cast('abcdef' as binary(6)) as varchar(10485761))
+go
+
+select cast(cast('abcdef' as binary(6)) as varchar(10485760))
```
Deepesh125, line 65: we still need some test cases whose length is greater than or less than binary(6), what about varbinary? varbinary(max)?

Deepesh125 (changes requested):

contrib/babelfishpg_common/sql/upgrades/babelfishpg_common--2.5.0--2.6.0.sql
basasairohan: Yes. The tests for other possible datatype cast to varchar already exists in babel_datatype.sql

test/JDBC/input/babel-4475-vu-verify.sql
basasairohan: Added tests for varbinary and varbinary(max).

Deepesh125 (approved):

surendravishnoi (approved):

Issue #2135, opened 2023-12-10
Error message should mention original index/constraint name


robverschoor:
Babelfish modifies certain names of indexes (and some constraints) in order to satisfy the PG requirement that these names are unique in a schema. Error messages should report the original T-SQL name instead of the mapped name.
Examples:
```
1> create table t1 (a int)
2> go
1> create index ix1 on t1(a)
2> go
1> create index ix1 on t1(a)
2> go
Msg 2714, Level 16, State 1, Server BABELFISH, Line 1
relation "ix1t1a5111d2a1767bc43a700e9f2162be019" already exists

1> drop index ix1 on t1
2> go
1> drop index ix1 on t1
2> go
Msg 67137668, Level 16, State 1, Server BABELFISH, Line 1
index "ix1t1a5111d2a1767bc43a700e9f2162be019" does not exist

1>
2> CREATE TABLE t2
3> (
4> c1 int
5> ,c2 int
6> ,CONSTRAINT CK_t2_c2_gt_c1 CHECK(c2 > c1)
7> )
8>
9> --Incorrect name of constraint in error message
10> INSERT INTO t2(c1, c2) VALUES(5, 2)
11> GO
Msg 67391682, Level 16, State 1, Server BABELFISH, Line 10 new row for relation "t2" violates check constraint "ck_t2_c2_gt_c1t1709f7347e565cadea2159827452c71d6"
```

Issue #2134, opened 2023-12-10
Mapped identifiers must be stored in an extended Babelfish catalog


robverschoor:
Babelfish modifies certain identifiers in order to map T-SQL names to PG-compatible names. Examples include index names (which get a hash appended to make them unique in a schema) and identifiers longer than 63 characters (which get truncated at 31 and have a 32-char MD5 hash of the original full name attached). These identifiers should be recorded into the Babelfish extended catalog `sys.babelfish_namespace_ext`. The purpose of recording these identifiers is to enable user-facing communication, for example in error messages as well as in the contents of the T-SQL catalogs, to report the original T-SQL name instead of the internal Babelfish name.

PR #2216, opened 2024-01-05
Support 'DROP INDEX table.index' and 'DROP INDEX ix ON schema.table'

robverschoor:
### Description

To support the syntax `DROP INDEX table.index`, the statement is rewritten in ANTLR as `DROP INDEX index ON table`, which is picked up by the backend parser.
To support the syntax `DROP INDEX schema.table.index` and `DROP INDEX index ON schema.table`, the statement is rewritten in ANTLR as `DROP INDEX index ON table SCHEMA schema`, which is picked up by the backend parser. Adding the `SCHEMA` clause for this case was necessary to avoid Bison reduce conflicts which resulted from seemingly more straightforward syntax options. In the backend parser, the trick is then to create an object reference in the `DropStmt` structure for the schema-qualified index with the schema name from the `SCHEMA` clause that was added, since PG's `DROP INDEX` only references the index, instead of its table like in T-SQL. 
Also, proper error messages are now raised when attempting DROP INDEX with a 3-part of 4-part name.

### Issues Resolved

BABEL-1483 Support DROP INDEX table.index syntax
BABEL-1652 Support DROP INDEX ix ON schema.table syntax

* **Use case based -** Yes

* **Arbitrary inputs -** N/A

* **Performance tests -** N/A

* **Tooling impact -** N/A

* **Client tests -** N/A

### Check List

### DIFF
--- a/test/JDBC/expected/drop_index-vu-verify.out
+++ b/test/JDBC/expected/drop_index-vu-verify.out
@@ -1,0 +1,532 @@
+-- This covers BABEL-1483 (Support DROP INDEX table.index syntax) and BABEL-1652 (Support DROP INDEX ix ON schema.table syntax)
+use master
+go
+
+-- tests with '[schema_name.]table_name.index_name' syntax
+create table dbo.t1_drop_index(a int)
+go
+create index ix1 on dbo.t1_drop_index(a)
+go
+select db_name(), object_name(id), object_schema_name(id), name from sysindexes where name like 'ix1%' order by 1,2,3,4
+go
+~~START~~
+nvarchar#!#varchar#!#varchar#!#varchar
+master#!#t1_drop_index#!#dbo#!#ix1t1_drop_indexa5111d2a1767bc43a700e9f2162be019
+~~END~~
+
+
+create table guest.t1_drop_index(a int)
+go
+create index ix1 on guest.t1_drop_index(a)
+go
+select db_name(), object_name(id), object_schema_name(id), name from sysindexes where name like 'ix1%' order by 1,2,3,4
+go
+~~START~~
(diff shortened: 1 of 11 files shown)


test/JDBC/expected/drop_index-vu-verify.out
```diff
@@ -0,0 +1,532 @@
...
+~~START~~
+nvarchar#!#varchar#!#varchar#!#varchar
+master#!#t1_drop_index#!#dbo#!#ix1t1_drop_indexa5111d2a1767bc43a700e9f2162be019
+~~END~~
+
+execute p1_drop_index
+go
+~~ERROR (Code: 33557097)~~
+
+~~ERROR (Message: index "ix1t1_drop_indexa5111d2a1767bc43a700e9f2162be019" does not exist)~~
```
roshan0708, line 175: In this error message, it would be more helpful to provide the actual index name the customer is trying to drop, such as "index 'ix1' does not exist." This can prevent confusion and help the customer identify the specific index causing the issue.

robverschoor: True, but that is part of a much bigger issue, namely that hashed/mapped names should be hidden from the user and the original T-SQL names should be used instead. This would only be one case where this applies. Addressing this issue should be approached holistically rather than for this particular case. See for example https://github.com/babelfish-for-postgresql/babelfish_extensions/issues/2135 and https://github.com/babelfish-for-postgresql/babelfish_extensions/issues/2134.

roshan0708: Understood. Thanks for the clarification!

Sairakan (approved):

jsudrik (approved):
Approved.

PR #2217, opened 2024-01-05
Handle operators adjacent to @@variables

robverschoor:
### Description

Operators which are directly adjacent to `@@` variables were incorrectly getting interpreted as PG operators. A space is now inserted in ANTLR prior to `@@` to ensure they are parsed correctly.
In addition, parameter defaults that use `@@` variables (`create proc p @p int=@@spid as ...`) and named parameter syntax (`exec myproc @p=@@spid`) are rewritten in the same way.

### Issues Resolved

BABEL-2888 =@@, >@@ (etc) cause syntax error 

* **Use case based -** Yes

* **Arbitrary inputs -** N/A

* **Performance tests -** N/A

* **Tooling impact -** N/A

* **Client tests -** N/A

### Check List


test/JDBC/upgrade/latest/schedule
```diff
@@ -466,6 +466,7 @@ print_null
 unquoted_string
 doublequoted_string
 operator_whitespace
+operator_atatvar
```
forestkeeper, line 469: For this Mvu test, we should put it into different upgrade test files

contrib/babelfishpg_tsql/src/tsqlIface.cpp
```diff
@@ -2130,6 +2130,63 @@ class tsqlBuilder : public tsqlCommonMutator
...
+
+		if ((ctx->comparison_operator()) && (ctx->expression().size() > 1))
+		{
+			std::string op = getFullText(ctx->comparison_operator());		
+			if ((op.back() == '=') || 
+			    (op.back() == '>') || 
+			    (op.back() == '<'))
+			{
+				// The operator must be followed immediately by the variable without any character in between
+				Assert(ctx->expression(1));
```
forestkeeper, line 2148: Would you mind assert for 
ctx->expression(1)->start
ctx->expression(1)->stop
 too ?

contrib/babelfishpg_tsql/src/tsqlIface.cpp
```diff
@@ -2130,6 +2130,63 @@ class tsqlBuilder : public tsqlCommonMutator
...
+				size_t startPosition = ctx->expression(1)->start->getStartIndex();
+				if ((startPosition - ctx->comparison_operator()->stop->getStopIndex()) == 1)
+				{
+					std::string var = getFullText(ctx->expression(1));
+					// The subsequent expression must be a variable starting with '@@'
+					if (var.front() == '@') 
+					{
```
forestkeeper, line 2155: Those two if condition can be combine into one

contrib/babelfishpg_tsql/src/tsqlIface.cpp
```diff
@@ -2130,6 +2130,63 @@ class tsqlBuilder : public tsqlCommonMutator
...
+		Assert(ctx->execute_parameter());
+		size_t startPosition = ctx->execute_parameter()->start->getStartIndex();
+		if ((startPosition - ctx->EQUAL()->getSymbol()->getStopIndex()) == 1)
+		{
+			std::string var = getFullText(ctx->execute_parameter());
+				
+			// The subsequent expression must be a variable starting with '@@'
+			if (var.front() == '@') 
+			{
+				if (var.at(1) == '@')
```
forestkeeper, line 2180: Those three if condition can combine into one.

contrib/babelfishpg_tsql/src/tsqlIface.cpp
```diff
@@ -2709,6 +2766,27 @@ class tsqlMutator : public TSqlParserBaseListener
...
+		{
+			// The '=' char must be followed immediately by the variable without any character in between
+			Assert(ctx->expression());			
+			size_t startPosition = ctx->expression()->start->getStartIndex();
+			if ((startPosition - ctx->EQUAL()->getSymbol()->getStopIndex()) == 1)
+			{
+				std::string var = getFullText(ctx->expression());
+				// The subsequent default expression must be a variable starting with '@@'
+				if (var.front() == '@') 
+				{
```
forestkeeper, line 2781: Those two if condition can be combined into one

contrib/babelfishpg_tsql/src/tsqlIface.cpp
robverschoor: start and stop always have a value for a node/token. if we cannot rely on that, we cannot rely on ANTLR

contrib/babelfishpg_tsql/src/tsqlIface.cpp
robverschoor: Indeed they can, but I am following the principle that the second test is required only when the first succeeds, so don't do the test unless it is truly required.

contrib/babelfishpg_tsql/src/tsqlIface.cpp
robverschoor: see comment above

contrib/babelfishpg_tsql/src/tsqlIface.cpp
robverschoor: see comment above

test/JDBC/upgrade/latest/schedule
robverschoor: there is no upgrade aspect since the things being fixed here all throw errors prior to the fix

contrib/babelfishpg_tsql/src/tsqlIface.cpp
robverschoor: To illustrate, start and stop are not explicitly checked for  a non-null value anywhere in the code.

contrib/babelfishpg_tsql/src/tsqlIface.cpp
forestkeeper: The runtime will automatically do this so won't be a problem. I suggested this since that'll help reading the code.

forestkeeper (approved):

PR #2242, opened 2024-01-10
Fix database  does not exists error when is_member(), schema_id() function gets called in parallel query mode.

ahmed-shameem:
### Description

This PR intends to solve the issue of parallel worker not being able to obtain correct logical database name. Whenever get_cur_db_name() is called, it returns static char current_db_name[] which is initialised as empty. Now, parallel workers are not aware of the database name. For them, it is empty whenever get_cur_db_name() gets called. Hence, we were getting the following error: 
```
database "" does not exist
```
So, to communicate the logical database name and more data related to babelfish we have introduced a struct BabelfishFixedParallelState. Currently BabelfishFixedParallelState has only 1 field, i.e., logical_db_name, we can add more fields to it whenever necessary. Then we will write the logical_db_name in DSM during initialisation (in InitializeParallelDSM) using a hook. Another hook is used while trying to fetch the logical database name in ParallelWorkerMain().

In this way we are introduction a mechanism through which we can communicate any babelfish related information to parallel workers.

Signed-off-by: Shameem Ahmed [shmeeh@amazon.com](mailto:rcv@amazon.com)
 
### Issues Resolved

BABEL-4538, BABEL-4481, BABEL-4421
 
### Check List


contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -408,6 +414,9 @@ InstallExtendedHooks(void)
 
 	pre_exec_tsql_cast_value_hook = exec_tsql_cast_value_hook;
 	exec_tsql_cast_value_hook = pltsql_exec_tsql_cast_value;
+
+    babelfixedparallelstate_insert_hook = babelfixedparallelstate_insert;
+    babelfixedparallelstate_restore_hook = babelfixedparallelstate_restore;
```

contrib/babelfishpg_tsql/src/session.c
```diff
@@ -51,6 +53,15 @@ get_cur_db_name(void)
 	return pstrdup(current_db_name);
 }
 
+void 
+set_cur_db_name_for_parallelWorker(const char* logical_db_name)
+{
+    int len = strlen(logical_db_name);
+    strncpy(current_db_name, logical_db_name, MAX_BBF_NAMEDATALEND);
+	current_db_name[len] = '\0';
+}
```
Deepesh125: Why dont we need validation on db_name? There should be Assert on the len <= MAX_BBF_NAMEDATALEND

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -4627,4 +4639,38 @@ static Node* optimize_explicit_cast(ParseState *pstate, Node *node)
...
+	{
+		/* Allow space to store the babelfish fixed-size parallel state. */
+		shm_toc_estimate_chunk(&pcxt->estimator, sizeof(BabelfishFixedParallelState));
+		shm_toc_estimate_keys(&pcxt->estimator, 1);
+	}
+	else
+	{
+		/* Initialize babelfish fixed-size state in shared memory. */
+		bfps = (BabelfishFixedParallelState *) shm_toc_allocate(pcxt->toc, sizeof(BabelfishFixedParallelState));
+		current_db_name = get_cur_db_name();
```
Deepesh125: free db_name after use?

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -4627,4 +4639,38 @@ static Node* optimize_explicit_cast(ParseState *pstate, Node *node)
...
+static void babelfixedparallelstate_insert(ParallelContext *pcxt, bool estimate)
+{
+	BabelfishFixedParallelState *bfps;
+	int len;
+	char* current_db_name;
+	if (estimate)
+	{
+		/* Allow space to store the babelfish fixed-size parallel state. */
+		shm_toc_estimate_chunk(&pcxt->estimator, sizeof(BabelfishFixedParallelState));
+		shm_toc_estimate_keys(&pcxt->estimator, 1);
```
Deepesh125: what is meaning of 1 here?

Deepesh125 (changes requested):

contrib/babelfishpg_tsql/src/session.c
ahmed-shameem: As we were setting this db name using get_cur_db_name() only where we are already checking the length, so IMO it was redundant. But, there is always a possibility for the data to get corrupted. Hence, put a check for length in latest rev.

contrib/babelfishpg_tsql/src/hooks.c
ahmed-shameem: current_db_name was not initialised with palloc, so there is no need to free it.

contrib/babelfishpg_tsql/src/hooks.c
ahmed-shameem: The second argument signified the number of keys we want to insert. In our case it is one, BABELFISH_PARALLEL_KEY_FIXED, hence 1.

contrib/babelfishpg_tsql/src/hooks.c
ahmed-shameem: Updated in latest rev.

contrib/babelfishpg_tsql/src/hooks.c
Deepesh125: get_cur_db_name() is allocating the memory internally.

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -4627,4 +4639,38 @@ static Node* optimize_explicit_cast(ParseState *pstate, Node *node)
 		}
 	}
 	return node;
+}
+
+static void babelfixedparallelstate_insert(ParallelContext *pcxt, bool estimate)
```
Deepesh125: Please add comment explaining the usage of this function and argument usage.

Deepesh125: Btw, does it make sense to move this code to session.c?

Deepesh125 (changes requested):

contrib/babelfishpg_tsql/src/session.c
```diff
@@ -51,6 +53,18 @@ get_cur_db_name(void)
 	return pstrdup(current_db_name);
 }
 
+void 
+set_cur_db_name_for_parallel_worker(const char* logical_db_name)
+{
+	int len = strlen(logical_db_name);
+
+	Assert(len <= MAX_BBF_NAMEDATALEND);
```
surendravishnoi, line 67: do we count null terminator part of MAX_BBF_NAMEDATALEND? MAX_BBF_NAMEDATALEND is max length or max available buffer?

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -4627,4 +4639,38 @@ static Node* optimize_explicit_cast(ParseState *pstate, Node *node)
...
+
+static void babelfixedparallelstate_restore(shm_toc *toc)
+{
+	BabelfishFixedParallelState *bfps;	
+
+	/* Get the babelfish fixed parallel state from DSM */
+	bfps = shm_toc_lookup(toc, BABELFISH_PARALLEL_KEY_FIXED, false);
+
+	/* Set the logcial db name for parallel workers */
+	set_cur_db_name_for_parallel_worker(bfps->logical_db_name);
```
surendravishnoi: we should handle the case of logical_db_name being NULL here.

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -4627,4 +4639,38 @@ static Node* optimize_explicit_cast(ParseState *pstate, Node *node)
...
+	char* current_db_name;
+	if (estimate)
+	{
+		/* Allow space to store the babelfish fixed-size parallel state. */
+		shm_toc_estimate_chunk(&pcxt->estimator, sizeof(BabelfishFixedParallelState));
+		shm_toc_estimate_keys(&pcxt->estimator, 1);
+	}
+	else
+	{
+		/* Initialize babelfish fixed-size state in shared memory. */
```
surendravishnoi: we rely on tds connection to call this hook. Another option is to rely on database name being NULL or not. Check can happen here as well.

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -408,6 +414,9 @@ InstallExtendedHooks(void)
 
 	pre_exec_tsql_cast_value_hook = exec_tsql_cast_value_hook;
 	exec_tsql_cast_value_hook = pltsql_exec_tsql_cast_value;
+
+	InitializeParallelDSM_hook = babelfixedparallelstate_insert;
```
surendravishnoi: let us add tsql or bbf prefix for these hook names

contrib/babelfishpg_tsql/src/hooks.c
ahmed-shameem: Freed in latest rev.

contrib/babelfishpg_tsql/src/hooks.c
ahmed-shameem: Updated in latest rev.

contrib/babelfishpg_tsql/src/session.c
ahmed-shameem: No, we assign one more extra bytes to store the NULL character.
```
static char current_db_name[MAX_BBF_NAMEDATALEND + 1] = {'\0'};
```
MAX_BBF_NAMEDATALEND is max available buffer

contrib/babelfishpg_tsql/src/hooks.c
ahmed-shameem: Updated in latest rev.

contrib/babelfishpg_tsql/src/hooks.c
ahmed-shameem: I think it is better to have the check for tds connection only, because IMO there can be other case where the db may be NULL (do not have any example in my mind right now). Can you please provide any scenario where tds connection check can cause any issue?

contrib/babelfishpg_tsql/src/hooks.c
ahmed-shameem: Updated in latest rev.

contrib/babelfishpg_tsql/src/session.c
```diff
@@ -376,3 +394,48 @@ initialize_context_table()
...
+	{
+		/* Allow space to store the babelfish fixed-size parallel state. */
+		shm_toc_estimate_chunk(&pcxt->estimator, sizeof(BabelfishFixedParallelState));
+		shm_toc_estimate_keys(&pcxt->estimator, 1);
+	}
+	else
+	{
+		/* Initialize babelfish fixed-size state in shared memory. */
+		bfps = (BabelfishFixedParallelState *) shm_toc_allocate(pcxt->toc, sizeof(BabelfishFixedParallelState));
+		current_db_name = get_cur_db_name();
```
surendravishnoi, line 428: we should handle NULL database name here even if there is no valid use case.

contrib/babelfishpg_tsql/src/session.h
```diff
@@ -1,6 +1,10 @@
 #ifndef SESSION_H
 #define SESSION_H
 #include "postgres.h"
+#include "multidb.h"
+#include "access/parallel.h"
+
+#define BABELFISH_PARALLEL_KEY_FIXED        UINT64CONST(0xBBF0000000000001)
```
surendravishnoi: how do we get this key value? can someone reuse the same key on engine side? If yes, we should keep key on engine side.

contrib/babelfishpg_tsql/src/session.c
```diff
@@ -376,3 +394,48 @@ initialize_context_table()
...
+}
+
+/* This function is responsible for restoring the babelfixedparallelstate*/
+void
+babelfixedparallelstate_restore(shm_toc *toc)
+{
+	BabelfishFixedParallelState *bfps;	
+
+	/* Get the babelfish fixed parallel state from DSM */
+	bfps = shm_toc_lookup(toc, BABELFISH_PARALLEL_KEY_FIXED, false);
```
surendravishnoi, line 455: what if no BBF key was inserted? will we get bfps as NULL? We should handle null bfps

ahmed-shameem: If no key is found, then shm_toc_lookup returns NULL if 3rd argument is true, otherwise throws elog(ERROR). As we are setting the 3rd arg as FALSE, if the key is not found, we will get an error like
```
could not find key " <key> " in shm TOC at <toc>
```

contrib/babelfishpg_tsql/src/session.c
ahmed-shameem: Updated in latest rev.

contrib/babelfishpg_tsql/src/session.h
ahmed-shameem: Yes, it is possible to use the key in engine side. We named the key as ```BBF<..>``` to indicate that it for babelfish only, but that does not guarantee that it won't be used by someone else.
Hence, moved this key into parallel.h

Deepesh125 (dismissed):

contrib/babelfishpg_tsql/src/session.c
```diff
@@ -376,3 +394,54 @@ initialize_context_table()
...
+		shm_toc_estimate_keys(&pcxt->estimator, 1);
+	}
+	else
+	{
+		/* Initialize babelfish fixed-size state in shared memory. */
+		bfps = (BabelfishFixedParallelState *) shm_toc_allocate(pcxt->toc, sizeof(BabelfishFixedParallelState));
+		current_db_name = get_cur_db_name();
+
+		if(!DbidIsValid(get_db_id(current_db_name)))
+		ereport(ERROR,
```
surendravishnoi: indentation

contrib/babelfishpg_tsql/src/session.c
surendravishnoi: we are still not handling NULL database name

surendravishnoi (dismissed):

contrib/babelfishpg_tsql/src/session.c
```diff
@@ -51,6 +52,23 @@ get_cur_db_name(void)
...
 }
 
+void 
+set_cur_db_name_for_parallel_worker(const char* logical_db_name)
+{
+	int len = strlen(logical_db_name);
+
+	Assert(len <= MAX_BBF_NAMEDATALEND);
+
+	if(logical_db_name == NULL || !DbidIsValid(get_db_id(logical_db_name)))
```
surendravishnoi: this is not helpful as you have already called strlen above. you can switch error check and assert

contrib/babelfishpg_tsql/src/session.c
```diff
@@ -51,6 +52,23 @@ get_cur_db_name(void)
...
+		ereport(ERROR,
+				(errcode(ERRCODE_UNDEFINED_DATABASE),
+				 errmsg("database \"%s\" does not exist", logical_db_name)));
```
surendravishnoi, line 72: I think %s with NULL string would be an issue. You can use "" in case of NULL

contrib/babelfishpg_tsql/src/session.c
```diff
@@ -376,3 +394,54 @@ initialize_context_table()
...
+	else
+	{
+		/* Initialize babelfish fixed-size state in shared memory. */
+		bfps = (BabelfishFixedParallelState *) shm_toc_allocate(pcxt->toc, sizeof(BabelfishFixedParallelState));
+		current_db_name = get_cur_db_name();
+
+		if(current_db_name == NULL || !DbidIsValid(get_db_id(current_db_name)))
+			ereport(ERROR,
+					(errcode(ERRCODE_UNDEFINED_DATABASE),
+					errmsg("database \"%s\" does not exist", current_db_name)));
```
surendravishnoi, line 438: I think %s with NULL string would be an issue. You can use "" in case of NULL

contrib/babelfishpg_tsql/src/session.c
ahmed-shameem: Updated in latest rev.

contrib/babelfishpg_tsql/src/session.c
ahmed-shameem: Updated in latest rev.

surendravishnoi (approved):

contrib/babelfishpg_tsql/src/session.c
ahmed-shameem: Updated.

contrib/babelfishpg_tsql/src/session.c
ahmed-shameem: Updated

contrib/babelfishpg_tsql/src/session.c
ahmed-shameem: Updated

PR #2251, opened 2024-01-14
Handle cursor erros on transaction boundaries

tanscorpio7:
### Description

On transaction boundaries, to prevent pinned portals (cursors) from being dropped, we hold them. Holding a portal internal fetches all the required tuples and stores them in the portal tuple store. This holding process could possible fail and fail our original command/error handling as well.

As a fix we put the executor run for persist portal inside try catch block and completely consume the error, if any. While for SCROLL cursors we simply unpin the portal and rethrow the error i.e. end execution.

We call HoldPinnedPortals from two places.
1. **PLTSQL Error handling:**
- No need for internal subtransaction for holding portal in this case, since we are anyway going to abort the transaction.
- We need to protect the original error in this case, since holdPinnedPortal can throw an error of its own & we may consume this error without rethrowing it.
2. **EXECSQL Stmt:** 
- If **txn command** then do hold portal before executing the txn command. (after txn command has been executed we are not in a state to create a new internal sub transaction for out hold portal step)
- If **NOT a txn command** - hold portal after executing the stmt & before commiting the transaction.

Not rethrowing the error means some clean up should be done to restore the state. (Releasing locks, clean up executor, guc level, restore user id etc)
In case of error while persisiting portals we rely on the following transaction/sub transaction abort to do the clean up. We transfer the portal->resowner to the current transaction resowner to make this happen. This also means that caller should ensures that transaction abort should be done after error in `pltsql_persist_holdable_cursor_executor`.

### Issues Resolved

[BABEL-4585]

### Check List

### DIFF
--- a/contrib/babelfishpg_tsql/src/hooks.c
+++ b/contrib/babelfishpg_tsql/src/hooks.c
@@ -4704,7 +4704,7 @@ pltsql_persist_holdable_cursor_executor(Portal portal, QueryDesc *queryDesc,
 
 			pltsql_update_cursor_error_data(portal->name);
 
-			pop_top_error_stack();
+			pop_error_stack();
 
 			InterruptHoldoffCount = saveInterruptHoldoffCount;
 			QueryCancelHoldoffCount = saveQueryCancelHoldoffCount;
--- a/contrib/babelfishpg_tsql/src/hooks.c
+++ b/contrib/babelfishpg_tsql/src/hooks.c
@@ -4689,6 +4689,9 @@ pltsql_persist_holdable_cursor_executor(Portal portal, QueryDesc *queryDesc,
 		}
 		PG_CATCH();
 		{
+			InterruptHoldoffCount = saveInterruptHoldoffCount;
+			QueryCancelHoldoffCount = saveQueryCancelHoldoffCount;
+
 			/*
 			 * Portal resources are transferred to current transaction resource
 			 * owner so transaction abort will handle the executor cleanup
@@ -4705,9 +4708,6 @@ pltsql_persist_holdable_cursor_executor(Portal portal, QueryDesc *queryDesc,
 			pltsql_update_cursor_error_data(portal->name);
 
 			pop_error_stack();
-
-			InterruptHoldoffCount = saveInterruptHoldoffCount;
-			QueryCancelHoldoffCount = saveQueryCancelHoldoffCount;
 
 			pltsql_non_tsql_proc_entry_count = 0;
 			pltsql_sys_func_entry_count = 0;
--- a/contrib/babelfishpg_tsql/src/hooks.c
+++ b/contrib/babelfishpg_tsql/src/hooks.c
@@ -20,6 +20,7 @@
 #include "catalog/pg_operator.h"
 #include "commands/copy.h"
 #include "commands/explain.h"
+#include "commands/portalcmds.h"
 #include "commands/tablecmds.h"
 #include "commands/trigger.h"
 #include "commands/view.h"
@@ -170,6 +171,9 @@ extern bool called_from_tsql_insert_exec();
 extern Datum pltsql_exec_tsql_cast_value(Datum value, bool *isnull,
 							 Oid valtype, int32 valtypmod,
 							 Oid reqtype, int32 reqtypmod);
+static void pltsql_persist_holdable_cursor_executor(Portal portal, 
+													QueryDesc *queryDesc,
+													ScanDirection direction);
 static void is_function_pg_stat_valid(FunctionCallInfo fcinfo,
 									  PgStat_FunctionCallUsage *fcu,
 									  char prokind, bool finalize);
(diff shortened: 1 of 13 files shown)


contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -4627,4 +4635,93 @@ static Node* optimize_explicit_cast(ParseState *pstate, Node *node)
...
+			ErrorData   	*error_data;
+			CmdType		operation;
+			int     	current_spi_stack_depth = SPI_get_depth();
+			bool    	sendTuples;
+			int     	rc;
+			
+			/* implicitly close SCROLL type cursors on error */
+			if (portal->cursorOptions & CURSOR_OPT_SCROLL)
+			{
+				MarkPortalFailed(portal);
```
surendravishnoi: what if we just rethrow the error?

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -4627,4 +4635,93 @@ static Node* optimize_explicit_cast(ParseState *pstate, Node *node)
...
+			else
+			{
+				ectx = MemoryContextSwitchTo(old_context);
+				error_data = CopyErrorData();
+				MemoryContextSwitchTo(ectx);
+				pltsql_update_cursor_error_message(portal->name, error_data->message, error_data->sqlerrcode);
+				FreeErrorData(error_data);
+			}
+
+			pop_top_error_stack();
```
surendravishnoi: why we need this?

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -4627,4 +4635,93 @@ static Node* optimize_explicit_cast(ParseState *pstate, Node *node)
...
+ * boundaries. Transaction commands will fail if this portal fetch
+ * fails. To handle this we silently fail and close SCROLL cursors
+ * although for NO_SCROLL cursors we make do with whatever is fetched.
+ * We also need to restore to the correct interrupt count value as we
+ * could already be inside a interrupt holdoff section. Also we do not
+ * rethrow the error here and end up in a dirty state, so some 
+ * transaction clean up should be done after this.
+ */
+static void
+pltsql_persist_holdable_cursor_executor(Portal portal, QueryDesc *queryDesc,
```
surendravishnoi, line 4666: should we do it even at lower level like execute plan? executor run can call standard executor or custom executor. We cannot do error handling of each kind of executor. May be we just focus on standard executor and try to put hook as deep as possible.

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -4627,4 +4635,93 @@ static Node* optimize_explicit_cast(ParseState *pstate, Node *node)
...
+			/* cleanup left over spi connections */
+			current_spi_stack_depth = SPI_get_depth();
+				
+			while (current_spi_stack_depth-- > prev_spi_stack_depth)
+				if ((rc = SPI_finish()) != SPI_OK_FINISH)
+					elog(ERROR, "SPI_finish failed: %s", SPI_result_code_string(rc));	
+
+			MemoryContextSwitchTo(old_context);
+
+			InterruptHoldoffCount = saveInterruptHoldoffCount;
```
surendravishnoi: how do we make sure that we are restoring all states required? these might be few which were flagged in current use case but there could be more.

surendravishnoi (dismissed):

surendravishnoi (commented):
Please ignore earlier approval

contrib/babelfishpg_tsql/src/hooks.c
tanscorpio7: 1. If we were executing a transaction command, the command will fail since we aborted midway.
2. If we were already inside our error handling, throwing the error here will pull us out of babelfish error handling.

contrib/babelfishpg_tsql/src/hooks.c
tanscorpio7: pltsql_persist_holdable_cursor_executor could be called from our error handling when we try to hold pin portals. So we already have an error on stack and the error generated inside this function is the 2nd one on the stack.
This function will clear the inside (top) error without loosing the original error.
if we flush the error, we would lose both the errors

contrib/babelfishpg_tsql/src/hooks.c
tanscorpio7: These were restored since these are specific to babelfish i.e. we are setting them.
For other states (PG local states GUC, current user etc ) we have to rely on the internal transaction command. So we have to make sure this function call is followed by some commit or abort transaction command (ones which actually do cleanup).
If we are not aborting or commiting afterwards then we do not need to hold the pin portals in the first place. Since we are trying to protect this cursors from getting cleaned up.

contrib/babelfishpg_tsql/src/hooks.c
tanscorpio7: 1. Original reason for putting the hook here
the hook only becomes relevant/executed when we are trying to hold pinned portals for pltsql cursors (which contains a future fetch error). For all other cases execution will remain same as before.

2. If we place the hook deeper we would need to figure out the proper conditions for hook being triggered. Meaning we do not alter the behaviour for any other type execution.

contrib/babelfishpg_tsql/src/hooks.c
tanscorpio7: For commit scenarios I think we might need to execute this inside a internal sub transaction to handle the clean up.
And error handling, abort transaction can handle the clean up.

PR #2253, opened 2024-01-15
Improve performance of sp_tablecollations_100 (spt_tablecollations_view)

tanscorpio7:
### Description

Make changes to sys.spt_tablecollations_view which is used by sys.sp_tablecollations_100
1.  Remove join on pg_attribute catalog and always use the sys.all_columns.name since we use it everywhere else for column name (no longer try to fetch the original name.)
2. Use pg_catalog.pg_class to fetch schema oid for given object oid instead of sys.all_objects.

### Local Performance Numbers

We first insert 40K tables with 3 columns each. We then run the procedure sys.sp_tablecollations_100 '@tablename'
All steps execute in the same session
**Step 1** EXEC sys.sp_tablecollations_100 'table_19000' (first time in the session takes longer)
**Step 2** EXEC sys.sp_tablecollations_100 'table_29000'
**Step 3** EXEC sys.sp_tablecollations_100 'table_39000' 
                 *-- Now insert another 10K new tables without terminating the current session.*
**Step 4** EXEC sys.sp_tablecollations_100 'table_44000'
**Step 5** EXEC sys.sp_tablecollations_100 'table_49000'

| Step             | After | Before |
| 1        |   745  | 14581 |
| 2     |   59  | 12294 |
| 3    |  59   | 12369 |
| 4   |  79   | 14886 |
| 5   |  72   | 14257 |

### Issues Resolved

[BABEL-3466]

### Sign-Off

### Check List

### DIFF
--- a/contrib/babelfishpg_tsql/sql/babelfishpg_tsql.sql
+++ b/contrib/babelfishpg_tsql/sql/babelfishpg_tsql.sql
@@ -709,7 +709,7 @@ CREATE OR REPLACE VIEW sys.spt_tablecollations_view AS
         c.object_id                      AS object_id,
         CAST(p.relnamespace AS int)      AS schema_id,
         c.column_id                      AS colid,
-        CAST(c.name AS sys.sysname)      AS name,
+        CAST(c.name AS sys.varchar)      AS name,
         CAST(CollationProperty(c.collation_name,'tdscollation') AS binary(5)) AS tds_collation_28,
         CAST(CollationProperty(c.collation_name,'tdscollation') AS binary(5)) AS tds_collation_90,
         CAST(CollationProperty(c.collation_name,'tdscollation') AS binary(5)) AS tds_collation_100,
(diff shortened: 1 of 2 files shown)


contrib/babelfishpg_tsql/sql/babelfishpg_tsql.sql
```diff
@@ -706,23 +706,21 @@ GRANT ALL on PROCEDURE sys.sp_describe_first_result_set TO PUBLIC;
...
     SELECT
-        o.object_id         AS object_id,
-        o.schema_id         AS schema_id,
-        c.column_id         AS colid,
-        CASE WHEN p.attoptions[1] LIKE 'bbf_original_name=%' THEN CAST(split_part(p.attoptions[1], '=', 2) AS sys.VARCHAR)
-			ELSE c.name COLLATE sys.database_default END AS name,
+        c.object_id                      AS object_id,
+        CAST(p.relnamespace AS int)      AS schema_id,
+        c.column_id                      AS colid,
+        CAST(c.name AS sys.varchar)      AS name,
```
ahmed-shameem, line 712: I would suggest you to replace all the views with pg_catalog. For example, c.object_id is same as pg_class.oid. You can check how are we obtaining the value of a particular column from the underlying view definition. IMO, if we can get rid of using all view and use only pg catalogs, that would increase the performance better.

NOTE: Make sure that we are obtaining the set of results back.

ahmed-shameem (changes requested):

contrib/babelfishpg_tsql/sql/babelfishpg_tsql.sql
tanscorpio7: I did try that, performance was same as this change.

Deepesh125 (approved):

rohit01010 (approved):

PR #2265, opened 2024-01-19
Add oid column to information_schema_tsql.columns for internal use only

tanscorpio7:
### Description

This PR fixes two things
1. Fix JOIN conditions which used `TABLE_NAME` by replacing it with `OID` instead.
2. INFORMATION_SCHEMA_TSQL.COLUMNS should return original name for `TABLE_NAME` & `COLUMN_NAME`

(Use case from DMS) Currently if we create a table with mixed case name(quoted) from PG endpoint and use same name to create another table from babelfish endpoint with certain combination of datatypes, system procedure sp_columns_100 starts failing with case not found error for this table. Root cause is the join condition used on information_schema_tsql.columns in sp_columns_100_view. We ended up picking datatype of babelfish endpoint table but typmod of PG created table.
`JOIN information_schema_tsql.columns t4 ON (pg_class.relname::sys.nvarchar(128) = t4."TABLE_NAME" AND ext.orig_name = t4."TABLE_SCHEMA")`
Here pg_class.relname can be a different table (mixed case) but will get through since we are using server default collation for comparison which is case insensitive (usually).

As a fix,
We create a new internal catalog information_schema_tsql.columns_internal which will contain the OID of table as well.
The actual information_schema_tsql.columns is just a wrapper view on this. (To mask the OID column).

### Issues Resolved

[BABEL-4553]

### Sign Off

### Check List


contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--3.4.0--3.5.0.sql
```diff
@@ -2624,6 +2943,8 @@ CALL sys.babelfish_drop_deprecated_object('view', 'sys', 'check_constraints_depr
 CALL sys.babelfish_drop_deprecated_object('view', 'sys', 'types_deprecated_3_5_0');
 CALL sys.babelfish_drop_deprecated_object('view', 'sys', 'table_types_deprecated_3_5_0');
 CALL sys.babelfish_drop_deprecated_object('view', 'sys', 'systypes_deprecated_3_5_0');
+CALL sys.babelfish_drop_deprecated_object('view', 'information_schema_tsql', 'columns');
+CALL sys.babelfish_drop_deprecated_object('view', 'sys', 'spt_columns_view_managed_3_5_0');
```
tanscorpio7, line 3313: We changed the sys.objects object_id column from OID to INT somewhere down the line but the corresponding drop & recreate for sys.spt_columns_view_managed is missing (uses sys.objects.object_id).
So after upgrade from source = PG13 we still have spt_columns_view_managed.object_id as an OID datatype and upgrade failures for this PR.
Hence, we deprecate this here.

contrib/babelfishpg_tsql/sql/information_schema_tsql.sql
```diff
@@ -234,13 +234,18 @@ $$SELECT
...
-CREATE OR REPLACE VIEW information_schema_tsql.columns AS
-	SELECT CAST(nc.dbname AS sys.nvarchar(128)) AS "TABLE_CATALOG",
+CREATE OR REPLACE VIEW information_schema_tsql.columns_internal AS
+	SELECT a.attrelid AS "TABLE_OID",
+			CAST(nc.dbname AS sys.nvarchar(128)) AS "TABLE_CATALOG",
 			CAST(ext.orig_name AS sys.nvarchar(128)) AS "TABLE_SCHEMA",
-			CAST(c.relname AS sys.nvarchar(128)) AS "TABLE_NAME",
+			CAST(
+				CASE WHEN c.reloptions[1] LIKE 'bbf_original_rel_name%' THEN substring(c.reloptions[1], 23)
+            	ELSE c.relname END
```
ahmed-shameem: Is this change really necessary? Because, pg can store the original object name if created using "", in that case the previous condition will give us correct results. 
Although, tsql returns original object name when queried whereas we are returning the downcase version of it. If we use this change in every such occurrence, we can return the original name of object from tds side. IMO, we should have a discussion regarding this.

test/JDBC/input/BABEL_4553.mix
```diff
@@ -0,0 +1,102 @@
+-- tsql
```
ahmed-shameem, line 1: Let us have tests to verify the JOIN conditions that we have modified to confirm that we were getting incorrect result set in some cases previously but with this change we are getting the correct results.

test/JDBC/input/BABEL_4553.mix
```diff
@@ -0,0 +1,102 @@
...
+CREATE DATABASE dms_test_target_db
+go
+USE dms_test_target_db
```
ahmed-shameem, line 4: Also, let us have some perf testing. IMO, we should get a bit better perf numbers as we have modified the join condition to check the oid only whereas initially there were around 4 operations involved. Anyways, let us have perf related tests as well.

contrib/babelfishpg_tsql/sql/information_schema_tsql.sql
```diff
@@ -234,13 +234,18 @@ $$SELECT
...
 
 /*
- * COLUMNS view
+ * COLUMNS view internal
  */
 
-CREATE OR REPLACE VIEW information_schema_tsql.columns AS
-	SELECT CAST(nc.dbname AS sys.nvarchar(128)) AS "TABLE_CATALOG",
+CREATE OR REPLACE VIEW information_schema_tsql.columns_internal AS
+	SELECT a.attrelid AS "TABLE_OID",
```
ahmed-shameem: IMO, we should use pg_class.oid instead of pg_attribute.attrelid. It won't affect the result but as the column name is "TABLE_OID" let us obtain it from pg_class as that makes more sense.

ahmed-shameem (changes requested):

contrib/babelfishpg_tsql/sql/information_schema_tsql.sql
sumitj824: ISC.tables returns the original name of table, we should be returning the original name here also. Lets discuss this with Dipesh.

contrib/babelfishpg_tsql/sql/information_schema_tsql.sql
```diff
@@ -329,6 +330,40 @@ CREATE OR REPLACE VIEW information_schema_tsql.columns AS
 									'SELECT, INSERT, UPDATE, REFERENCES'))
 		AND ext.dbid =sys.db_id();
 
+GRANT SELECT ON information_schema_tsql.columns_internal TO PUBLIC;
```
sumitj824: We don't need to grant perms to public if it is for internal use only.

contrib/babelfishpg_tsql/sql/information_schema_tsql.sql
```diff
@@ -607,7 +642,7 @@ CREATE OR REPLACE VIEW information_schema_tsql.COLUMN_DOMAIN_USAGE AS
            CAST(isc_col."TABLE_NAME" AS sys.sysname),
            CAST(isc_col."COLUMN_NAME" AS sys.sysname)
 
-    FROM information_schema_tsql.columns AS isc_col
+    FROM information_schema_tsql.columns_internal AS isc_col
```
sumitj824, line 651: why this change?

test/JDBC/input/BABEL_4553.mix
tanscorpio7: Added test for sys.sp_columns_100_view, sys.spt_columns_view_managed, sys.sp_pkeys_view
For sys.sp_statistics_view, the catalog query is tricky to understand.

sumitj824 (changes requested):

contrib/babelfishpg_tsql/sql/information_schema_tsql.sql
tanscorpio7: Done

contrib/babelfishpg_tsql/sql/information_schema_tsql.sql
tanscorpio7: Seems better to fetch directly from the internal view instead of the wrapper. Just for consistency.

contrib/babelfishpg_tsql/sql/information_schema_tsql.sql
tanscorpio7: Yes, just to need to decide if we want to fix this here, or take this separately.
There are some other columns which need this fix in ISC schema.

contrib/babelfishpg_tsql/sql/information_schema_tsql.sql
tanscorpio7: removed

contrib/babelfishpg_tsql/sql/information_schema_tsql.sql
sumitj824: It shouldn't matter.

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--3.4.0--3.5.0.sql
```diff
@@ -2626,6 +2606,364 @@ CREATE OR REPLACE VIEW sys.spt_tablecollations_view AS
...
+		"COLLATION_NAME",
+		"DOMAIN_CATALOG",
+		"DOMAIN_SCHEMA",
+		"DOMAIN_NAME"
+	
+	FROM information_schema_tsql.columns_internal;
+
+GRANT SELECT ON information_schema_tsql.columns TO PUBLIC;
+
+CREATE OR REPLACE VIEW information_schema_tsql.COLUMN_DOMAIN_USAGE AS
```
sumitj824, line 3061: Why there isn't grant statement for this view? Check for others as well.

tanscorpio7: Adding those, I assumed the catalogs must've the grant already.

sumitj824 (approved):

KushaalShroff (approved):

PR #2096, opened 2023-12-04
Added support in Version upgrade test framework to test upgrade with non-default-server-collation

rohit01010:
### Description
This pull request adds a support in version upgrade test framework to perform a upgrade with non-default server collation name. This pull request also add normal jdbc with non-default server collation workflow which will run all the jdbc tests with non-default server collation name.

### Upgrade paths added to test upgrade with non-default server collation
```
[14.6 --> 15.latest] with server collation name: chinese_prc_ci_as
[14.6 --> 15.latest] with server collation name: japanese_ci_as
[14.11 --> 15.latest] with server collation name: chinese_prc_ci_as
[13.6 --> 14.11 --> 15.latest] with server collation name: chinese_prc_ci_as
[14.3 --> 14.11 --> 15.latest] with server collation name: chinese_prc_ci_as
```

### JDBC Test with non-default server collation
Currently in this JDBC Test with non-default server collation, JDBC tests will run with server collation as  `chinese_prc_ci_as`. (Note: server collation can be changed by modifying `SERVER_COLLATION_NAME` env variable to desired server collation in workflow file for this test)

Signed-off-by: Rohit Bhagat [rohitbgt@amazon.com](mailto:rohitbgt@amazon.com)

### Issues Resolved

BABEL-4590

### Test Scenarios Covered ###
* **Use case based -** NA

* **Boundary conditions -** NA

* **Arbitrary inputs -** NA

* **Negative test cases -** NA

* **Minor version upgrade tests -** NA

* **Major version upgrade tests -** NA

* **Performance tests -** NA

* **Tooling impact -** NA

* **Client tests -** NA

### Check List
- [x] Commits are signed per the DCO using --signoff 

By submitting this pull request, I confirm that my contribution is under the terms of the Apache 2.0 and PostgreSQL licenses, and grant any person obtaining a copy of the contribution permission to relicense all or a portion of my contribution to the PostgreSQL License solely to contribute all or a portion of my contribution to the PostgreSQL open source project.

For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/babelfish-for-postgresql/babelfish_extensions/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #2268, opened 2024-01-19
Fix issues with datatypes for Logical Replication

rishabhtanwar29:
### Description
This commit fixes following issues:
1. (Var)binary columns are not replicated correctly since input function considers
the value as a raw ASCII string although the value is actually already in binary format.
Fixed this by using `IsLogicalWorker()` with existing `TSQLHexConstTypmod`
check in `varbinaryin` function to correctly take binary input sent by the publisher.
2. SQL_VARIANT columns are replicated correctly as far as the data contents are
concerned, but the meta data is lost in the process which is because sql_variant output
function returns the value in the respective datatype in string format which when fed
back to the sql_variant input function gets treated as a varchar value.
Fixed this by replicating sql_variant values as `bytea`. We will use `MyReplicationSlot`
and `MyWalSnd` in `sqlvariantout` function on publisher and `IsLogicalWorker()` in
`sqlvariantin` function on subscriber to identify if we are in logical replication context.
3. On the subscriber side, `IsLogicalWorker()` is sufficient for native PG applyworker
but will not work with external providers like pglogical, so will rely on SessionReplicationRole
being replica since most of the providers seem to set this GUC.

Additionally, this adds support to run logical replication tests in github action along with
JDBC framework.

Task: BABEL-3427, BABEL-4049

* **Use case based -**

* **Boundary conditions -**

* **Arbitrary inputs -**

* **Negative test cases -**

* **Performance tests -**

* **Tooling impact -**

* **Client tests -**

### Check List


.github/workflows/jdbc-tests.yml
```diff
@@ -47,58 +47,98 @@ jobs:
...
+          psql_port: 5433
+          tsql_port: 8199
+          wal_level: logical
+
+      - name: Setup Publication and Subscription
+        id: setup-pub-sub
+        if: always() && steps.start-secondary.outcome == 'success'
+        run: |
+          ~/${{env.INSTALL_DIR}}/bin/psql -d jdbc_testdb -U runner -c "CREATE PUBLICATION my_pub;"
+          ~/${{env.INSTALL_DIR}}/bin/psql -d jdbc_testdb -U runner -p 5433 -c "CREATE SUBSCRIPTION my_sub CONNECTION 'host=localhost port=5432 user=jdbc_user dbname=jdbc_testdb password=12345678' PUBLICATION my_pub;"
```
KushaalShroff, line 73: I am curious what happens when we give port as 1433?

test/JDBC/expected/datatype_test.out
```diff
@@ -0,0 +1,221 @@
...
+    , c20   SMALLINT    NOT NULL
+    , c21   SMALLMONEY  NOT NULL
+    , c22   SQL_VARIANT  NOT NULL
+    , c23   SYSNAME NOT NULL
+    , c24   TEXT    NOT NULL
+    , c25   TIME    NOT NULL
+    , c26   TINYINT NOT NULL
+    , c27   UNIQUEIDENTIFIER    NOT NULL
+    , c28   VARBINARY(20)  NOT NULL
+    , c29   VARCHAR(20)    NOT NULL
```
KushaalShroff, line 34: MAX types?

KushaalShroff (commented):
General:
1. What happens when the default collation of source/target are different?
2. Do we do any kind of compatibility checks? Like only version which compatible are allowed to replicate data

test/JDBC/replication/datatype_test.mix
```diff
@@ -0,0 +1,159 @@
...
+    , c25   TIME    NOT NULL
+    , c26   TINYINT NOT NULL
+    , c27   UNIQUEIDENTIFIER    NOT NULL
+    , c28   VARBINARY(20)  NOT NULL
+    , c29   VARCHAR(20)    NOT NULL
+    , c30   XML NOT NULL)
+GO
+
+-- tsql port=8199
+CREATE TABLE datatype_test_t1 (
```
KushaalShroff, line 54: Can we add negative tests where schema of target is a bit different.
1. Implicit casting
2. Throws error during insert maybe due to constraint issues. Whats the behaviour in that case?

.github/workflows/jdbc-tests.yml
rishabhtanwar29: Subscriber won't be able to connect to TDS port, the CREATE SUBSCRIPTION command will throw error that it can't connect to publisher server.

test/JDBC/replication/datatype_test.mix
rishabhtanwar29: 1. Added test for implicit casting.
2. Replication fails if there is a constraint failure and it will be logged in the postgres logs but we will not see any error here in this script.

test/JDBC/expected/datatype_test.out
rishabhtanwar29: Have verified MAX types locally but now I have added one more test for them.

rishabhtanwar29:
> General:
> 
>     1. What happens when the default collation of source/target are different?
> 
>     2. Do we do any kind of compatibility checks? Like only version which compatible are allowed to replicate data

1. Collation does not matter in logical replication since the transfer happens through input/output functions. I have tested with source/target having different collations locally. Automated tests can be added once this [PR](https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/2096) from Rohit gets merged into 4.x.
2. Compatibility checks are not the in the scope of this PR.

KushaalShroff (dismissed):

contrib/babelfishpg_common/src/sqlvariant.c
```diff
@@ -79,6 +83,16 @@ sqlvariantin(PG_FUNCTION_ARGS)
...
 	svhdr_5B_t *svhdr;
 
+	/*
+	 * Input as a bytea instead if it is logical replication applyworker.
+	 *    IsLogicalWorker() is sufficient for native PG applyworker but will
+	 *    not work with external providers like pglogical, so will rely on
+	 *    SessionReplicationRole being replica since most of the providers seem
+	 *    to set this GUC.
+	 */
+	if (IsLogicalWorker() || SessionReplicationRole == SESSION_REPLICATION_ROLE_REPLICA)
```
surendravishnoi: it would be cleaner to add wrapper/macro functions to identify logical workers (source and target)

rishabhtanwar29: Makes sense, added MACROS.

contrib/babelfishpg_common/src/logical.h
```diff
@@ -0,0 +1,23 @@
...
+ *    to set this GUC.
+ */
+#define IS_LOGICALREP_APPLYWORKER (IsLogicalWorker() || SessionReplicationRole == SESSION_REPLICATION_ROLE_REPLICA)
+
+/*
+ * There are two criterias for walsender:
+ * 1. MyReplicationSlot is logical.
+ * 2. This is a logical walsender process.
+ */
+#define IS_LOGICALREP_WALSENDER \
```
kuntalghosh: This should be a macro function. Otherwise, this looks like a constant macro.

```
#define MACRO_DEF()
```

Also, let's rename these macros as IS_LOGICAL_SENDER() and IS_LOGICAL_RECEIVER()?

kuntalghosh (approved):

KushaalShroff (approved):

PR #2280, opened 2024-01-24
Fix the issue datetime/datetime2 may not choose index scan

forestkeeper:
When comparing between datetime to date or datetime2 to date, it would not choose index scan because we wrongly defined the date2datetime and date2datetime2 cast function as VOLATILE. They should be defined as STABLE.

Task: BABEL-4517

### Check List


test/JDBC/expected/babel-4517-vu-verify.out
```diff
@@ -0,0 +1,107 @@
...
+
+SELECT set_config('babelfishpg_tsql.explain_costs', 'off', true)
+GO
+~~START~~
+text
+off
+~~END~~
+
+
+SELECT set_config('debug_parallel_query', '0', true)
```
2jungkook, line 12: Is this necessary? If so,
1. please reset this at the end of this file.
2. you need to rename this to `force_parallel_mode` when downporting to lower major versions

test/JDBC/upgrade/16_1/schedule
```diff
@@ -496,3 +496,4 @@ permission_restrictions_from_pg
 BABEL-4529-before-15_6-or-14_11
 BABEL-730-before-15_6-or-16_1
 babel-4475
+babel-4517
```
2jungkook, line 499: Please add this to `latest/schedule` file as well.

test/JDBC/expected/babel-4517-vu-verify.out
forestkeeper: Yes, it's for parallel query regression test, to turn  parallel query off in default, this value will be reset back after the transaction is committed ( transaction : babel_4517 ), so that's why set it with set_config('debug_parallel_query', '0', true) -> true .

Yes, I knew I'll use force_parallel_mode.

forestkeeper: I can't direct turn on/off after the test case, due to this value is not a fixed value compared to jdbc test and parallel query test

test/JDBC/upgrade/16_1/schedule
forestkeeper: OK, missed that one, thanks for the finding

2jungkook (approved):

PR #2295, opened 2024-01-25
Crash in SPExecuteSQL because of improper error handling

lejaokri:
When logging statement, SPExecuteSQL and similar function expect that there could be an error hence they surround the ereport call in PG_TRY and PG_CATCH ignoring the error in PG_CATCH and moving one without properly resetting a few things such as MemoryContext and errordata_stack_depth.

Task: BABEL-4716

### Check List

### DIFF
--- a/contrib/babelfishpg_tds/src/backend/tds/tdsrpc.c
+++ b/contrib/babelfishpg_tds/src/backend/tds/tdsrpc.c
@@ -151,6 +151,8 @@ static void SPPrepExec(TDSRequestSP req);
 static void SPCustomType(TDSRequestSP req);
 static void SPUnprepare(TDSRequestSP req);
 static void TDSLogStatementCursorHandler(TDSRequestSP req, char *stmt, int option);
+static void LogStatementNoError(const char *header, const int handle, const char *msg, const uint16 nparams);
+
 static InlineCodeBlockArgs *DeclareVariables(TDSRequestSP req, FunctionCallInfo *fcinfo, unsigned long options);
 List	   *tvp_lookup_list = NIL;
 bool		lockForFaultInjection = false;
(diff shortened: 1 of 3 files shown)


contrib/babelfishpg_tds/src/backend/tds/tdsrpc.c
```diff
@@ -600,6 +600,9 @@ SPExecuteSQL(TDSRequestSP req)
 		}
 		PG_CATCH();
 		{
+			MemoryContextSwitchTo(curr);
```
surendravishnoi: this code needs re-evaluation. We have error in try and error in cache. Why we have code after end try, control will never go there? Also, it seems errdetail_params can throws the error. It would be more maintainable to put TRY/CATCH inside it and return appropriate value. That we we can have simple ereport on caller side and need not handle it at N number of places.

contrib/babelfishpg_tds/src/backend/tds/tdsrpc.c
```diff
@@ -801,6 +804,7 @@ SPExecute(TDSRequestSP req)
 	 */
 	if (pltsql_plugin_handler_ptr->stmt_needs_logging || TDS_DEBUG_ENABLED(TDS_DEBUG2))
 	{
+		MemoryContext curr = CurrentMemoryContext;
```
surendravishnoi: it seems stmt needs logging block can be a utility function with appropriate parameter to put this code at one place.

contrib/babelfishpg_tds/src/backend/tds/tdsrpc.c
lejaokri: good point. I will make the change

lejaokri: I took a look, due to timing constraint, I will fix the crash in SPExecuteSQL as that is a regression.

The generic handling of `errdetail_params` will be handled by another Jira (will open one). It requires some more deep dive as some callers may not be safe to do a `FlusErrorState` at the time they call `errdetail_params`. This logic has been there since Day 1.

2jungkook (approved):

PR #2301, opened 2024-01-26
sp_rename should update the object name in babelfish_schema_permission catalog

shalinilohia50:
### Description

We had introduced a catalog `babelfish_schema_permission` to hold the information about objects and its permissions. With this change, `sp_rename` should update the object name in `babelfish_schema_permission` catalog so that the permission is tracked correctly in the catalogs.

### Issues Resolved

Task: [BABEL-4485](https://jira.rds.a2z.com/browse/BABEL-4485)

* **Use case based -** Added
* Rename objects with schema name
* Rename objects without schema name
* Rename objects like relations and functions

* **Boundary conditions -**

* **Arbitrary inputs -**

* **Negative test cases -**

* **Performance tests -**

* **Tooling impact -**

* **Client tests -**

### DIFF
--- a/contrib/babelfishpg_tsql/src/catalog.c
+++ b/contrib/babelfishpg_tsql/src/catalog.c
@@ -2625,6 +2625,14 @@ rename_update_bbf_catalog(RenameStmt *stmt)
 {
 	switch (stmt->renameType)
 	{
+		/*
+		 * In case of renaming a table, view, procedure and function, modify the object
+		 * names present in all the Babelfish catalogs which stores these object names to
+		 * have consistency with the new names.
+		 *
+		 * List of catalogs which are updated here:
+		 * babelfish_schema_permissions, babelfish_function_ext, babelfish_view_def
+		 */
 		case OBJECT_TABLE:
 			rename_object_update_bbf_schema_permission_catalog(stmt, stmt->renameType);
 			break;
@@ -2653,6 +2661,12 @@ rename_update_bbf_catalog(RenameStmt *stmt)
 	}
 }
 
+/*
+ * rename_object_update_bbf_schema_permission_catalog
+ *
+ * In case of renaming a table, view, procedure and function, modify the 'object_name' in
+ * 'babelfish_schema_permissions' to have consistency with the new names.
+ */
 static void
 rename_object_update_bbf_schema_permission_catalog(RenameStmt *stmt, int rename_type)
 {
@@ -2666,11 +2680,10 @@ rename_object_update_bbf_schema_permission_catalog(RenameStmt *stmt, int rename_
 	Datum		new_record_bbf_schema[BBF_SCHEMA_PERMS_NUM_OF_COLS] = {0};
 	bool		new_record_nulls_bbf_schema[BBF_SCHEMA_PERMS_NUM_OF_COLS] = {false};
 	bool		new_record_repl_bbf_schema[BBF_SCHEMA_PERMS_NUM_OF_COLS] = {false};
-	const char *logical_schema_name;
-	const char *object_name;
-	char	*object_type = NULL;
+	char		*logical_schema_name = NULL;
+	char		*object_name = NULL;
+	char		*object_type = NULL;
 	int16		dbid = get_cur_db_id();
-	Node	   *schema;
 	ObjectWithArgs *objwargs;
 
 	/* open the catalog table */
@@ -2680,7 +2693,7 @@ rename_object_update_bbf_schema_permission_catalog(RenameStmt *stmt, int rename_
 
 	if (rename_type == OBJECT_TABLE || rename_type == OBJECT_VIEW)
 	{
-		logical_schema_name = get_logical_schema_name(stmt->relation->schemaname, true);
+		logical_schema_name = (char *) get_logical_schema_name(stmt->relation->schemaname, true);
 		object_name = stmt->relation->relname;
 		object_type = OBJ_RELATION;
 	}
@@ -2691,9 +2704,7 @@ rename_object_update_bbf_schema_permission_catalog(RenameStmt *stmt, int rename_
 		else if (rename_type == OBJECT_FUNCTION)
 			object_type = OBJ_FUNCTION;
 		objwargs = (ObjectWithArgs *) stmt->object;
-		schema = (Node *) linitial(objwargs->objname);
-		logical_schema_name = strVal(schema);
-		object_name = stmt->subname;
+		DeconstructQualifiedName(objwargs->objname, &logical_schema_name, &object_name);
 	}
 
 	/* search for the row for update => build the key */
@@ -2727,6 +2738,16 @@ rename_object_update_bbf_schema_permission_catalog(RenameStmt *stmt, int rename_
 	/* get the scan result -> original tuple */
 	tuple_bbf_schema = systable_getnext(scan);
 
+	/*
+	 * If a permission on the same object is granted to multiple grantees,
+	 * there can be multiple rows in the catalog corresponding to each grantee name.
+	 * All such rows need to be updated with the new name.
+	 *
+	 * It is OK to not throw an error if an entry is not found in 'babelfish_schema_permissions'.
+	 * Explaination: An entry is added to 'babelfish_schema_permissions' only if an object has an explicit GRANT on it.
+	 * It is not necessary that each RENAME on an object has a GRANT of that object too.
+	 * Hence, there can be missing entries.
+	 */
 	while (HeapTupleIsValid(tuple_bbf_schema))
 	{
 		/* create new tuple to substitute */
--- a/contrib/babelfishpg_tsql/src/catalog.c
+++ b/contrib/babelfishpg_tsql/src/catalog.c
@@ -2732,8 +2732,8 @@ rename_object_update_bbf_schema_permission_catalog(RenameStmt *stmt, int rename_
 
 	/* scan */
 	scan = systable_beginscan(bbf_schema_rel,
-			get_bbf_schema_perms_idx_oid(),
-			true, NULL, 4, key);
+			get_bbf_schema_perms_oid(),
+			false, NULL, 4, key);
 
 	/* get the scan result -> original tuple */
 	tuple_bbf_schema = systable_getnext(scan);
--- a/contrib/babelfishpg_tsql/src/catalog.c
+++ b/contrib/babelfishpg_tsql/src/catalog.c
@@ -2727,7 +2727,7 @@ rename_object_update_bbf_schema_permission_catalog(RenameStmt *stmt, int rename_
 	/* get the scan result -> original tuple */
 	tuple_bbf_schema = systable_getnext(scan);
 
-	if (HeapTupleIsValid(tuple_bbf_schema))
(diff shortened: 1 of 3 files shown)


contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -1583,6 +1583,7 @@ static void alter_guest_schema_for_db(const char *dbname);
 /* Helper function Rename BBF catalog update*/
 static void rename_view_update_bbf_catalog(RenameStmt *stmt);
 static void rename_procfunc_update_bbf_catalog(RenameStmt *stmt);
+static void rename_object_update_bbf_schema_permission_catalog(RenameStmt *stmt, const char *obj_type);
```
xhfanhe: Nit: Let argument be (stmt, stmt->renameType) and deal with the obj_type inside fuction is a better implementation.

xhfanhe (dismissed):

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -2648,6 +2653,102 @@ rename_update_bbf_catalog(RenameStmt *stmt)
...
+	{
+		if (rename_type == OBJECT_PROCEDURE)
+			object_type = OBJ_PROCEDURE;
+		else if (rename_type == OBJECT_FUNCTION)
+			object_type = OBJ_FUNCTION;
+		objwargs = (ObjectWithArgs *) stmt->object;
+		schema = (Node *) linitial(objwargs->objname);
+		logical_schema_name = strVal(schema);
+		object_name = stmt->subname;
+	}
```
HarshLunagariya: Is it mandatory to pass schema qualified name in sp_rename? If not then we won't be getting the schema at this point.

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -2648,6 +2653,102 @@ rename_update_bbf_catalog(RenameStmt *stmt)
...
+		new_record_bbf_schema[Anum_bbf_schema_perms_object_name - 1] = CStringGetTextDatum(stmt->newname);
+		new_record_repl_bbf_schema[Anum_bbf_schema_perms_object_name - 1] = true;
+
+		new_tuple = heap_modify_tuple(tuple_bbf_schema,
+									bbf_schema_dsc,
+									new_record_bbf_schema,
+									new_record_nulls_bbf_schema,
+									new_record_repl_bbf_schema);
+
+		CatalogTupleUpdate(bbf_schema_rel, &new_tuple->t_self, new_tuple);
```
HarshLunagariya, line 2763: Will it be possible to print the updated entries in the test?

contrib/babelfishpg_tsql/src/catalog.c
shalinilohia50: Your question is not very clear to me. 
We are fetching the schema name and object name from the rename statement, based on which we search the catalog for any available entry and update it.

surendravishnoi (commented):
I think this has happened a number of times where we add a catalog but forget to align DDL code for it. Could you please open a Jira for the same and we should come up with a generic framework to catch it as part of either compilation or testing. This change can go ahead independently.

shalinilohia50:
> I think this has happened a number of times where we add a catalog but forget to align DDL code for it. Could you please open a Jira for the same and we should come up with a generic framework to catch it as part of either compilation or testing. This change can go ahead independently.

Sure.

contrib/babelfishpg_tsql/src/catalog.c
HarshLunagariya: For every test case of sp_rename, we are passing the schema qualified name with it, We should include the test w/o schema name as well.

HarshLunagariya: We should add a testcase for case sensitive behaviour as well (*only if we are storing the original case sensitive names in schema permission catalog*).

contrib/babelfishpg_tsql/src/catalog.c
shalinilohia50: Added tests to show the updated entries after rename.

contrib/babelfishpg_tsql/src/catalog.c
shalinilohia50: 1. Added tests for objects without schema name as well.
2. We are storing the object names in lower case in the catalog.

HarshLunagariya (approved):

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -2625,15 +2626,19 @@ rename_update_bbf_catalog(RenameStmt *stmt)
 	switch (stmt->renameType)
```
kuntalghosh, line 2626: Please add some comments regarding:
1. Which catalogs are being modified?
2. Why some of the objects are being modified while others are not? (for both the catalogs)

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -2648,6 +2653,102 @@ rename_update_bbf_catalog(RenameStmt *stmt)
 	}
 }
 
+static void
+rename_object_update_bbf_schema_permission_catalog(RenameStmt *stmt, int rename_type)
```
kuntalghosh, line 2671: Please add function level comment about the purpose

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -2648,6 +2653,102 @@ rename_update_bbf_catalog(RenameStmt *stmt)
...
+	}
+	else
+	{
+		if (rename_type == OBJECT_PROCEDURE)
+			object_type = OBJ_PROCEDURE;
+		else if (rename_type == OBJECT_FUNCTION)
+			object_type = OBJ_FUNCTION;
+		objwargs = (ObjectWithArgs *) stmt->object;
+		schema = (Node *) linitial(objwargs->objname);
+		logical_schema_name = strVal(schema);
```
kuntalghosh: Please use DeconstructQualifiedName for this.

contrib/babelfishpg_tsql/src/catalog.c
kuntalghosh: I think Harsh is asking to add tests without schema name.

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -2648,6 +2653,102 @@ rename_update_bbf_catalog(RenameStmt *stmt)
...
+
+	/* scan */
+	scan = systable_beginscan(bbf_schema_rel,
+			get_bbf_schema_perms_idx_oid(),
+			true, NULL, 4, key);
+
+	/* get the scan result -> original tuple */
+	tuple_bbf_schema = systable_getnext(scan);
+
+	while (HeapTupleIsValid(tuple_bbf_schema))
```
kuntalghosh, line 2751: 1. Can there be more than one entry? If not, better to use if condition for readability.
2. Let's add a comment why it's okay not to throw an error if an entry is not found.

kuntalghosh (changes requested):

contrib/babelfishpg_tsql/src/catalog.c
shalinilohia50: 1. While is needed since there can be multiple rows.
2. Added.

contrib/babelfishpg_tsql/src/catalog.c
shalinilohia50: Done.

contrib/babelfishpg_tsql/src/catalog.c
shalinilohia50: Yes, added those tests already.

contrib/babelfishpg_tsql/src/catalog.c
shalinilohia50: Added a comment.

contrib/babelfishpg_tsql/src/catalog.c
shalinilohia50: Done.

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -2648,6 +2661,115 @@ rename_update_bbf_catalog(RenameStmt *stmt)
...
+				BTEqualStrategyNumber,
+				InvalidOid,
+				tsql_get_server_collation_oid_internal(false),
+				F_TEXTEQ,
+				CStringGetTextDatum(object_type));
+
+	/* scan */
+	scan = systable_beginscan(bbf_schema_rel,
+			get_bbf_schema_perms_idx_oid(),
+			true, NULL, 4, key);
```
KushaalShroff: @shalinilohia50 I gave this comment multiple times in your last PR, we are using index to scan but the key doesnt match all of the index columns

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -2648,6 +2661,115 @@ rename_update_bbf_catalog(RenameStmt *stmt)
...
+	ScanKeyData key[4];
+	HeapTuple	tuple_bbf_schema;
+	HeapTuple	new_tuple;
+	SysScanDesc scan;
+	Datum		new_record_bbf_schema[BBF_SCHEMA_PERMS_NUM_OF_COLS] = {0};
+	bool		new_record_nulls_bbf_schema[BBF_SCHEMA_PERMS_NUM_OF_COLS] = {false};
+	bool		new_record_repl_bbf_schema[BBF_SCHEMA_PERMS_NUM_OF_COLS] = {false};
+	char		*logical_schema_name = NULL;
+	char		*object_name = NULL;
+	char		*object_type = NULL;
```
KushaalShroff: Not freeing them

KushaalShroff (changes requested):

KushaalShroff:
Can you update PR description with Testing?

contrib/babelfishpg_tsql/src/catalog.c
shalinilohia50: Based on these 4 keys, there can be multiple rows depending on grantees. We have to rename the object name in all such rows. 
At this point, we don't know the grantee names and we don't need it either.

kuntalghosh: In fact, sp_rename doesn't care about grantee name IIUC. We need to modify all the rows that match the 4 keys irrespective of the grantee name.

kuntalghosh (dismissed):

contrib/babelfishpg_tsql/src/catalog.c
shalinilohia50: Done.

contrib/babelfishpg_tsql/src/catalog.c
KushaalShroff: I think you didnt understand the issue
you are using index of 5 keys' oid to do a scan with 4 keys. Why are we not doing a scan on table's oid?
You are explaining why we need 4 keys and thats not an issue

shalinilohia50: Updated.

KushaalShroff (approved):

PR #2271, opened 2024-01-21
Support TSQL syntax for Geospatial functions

Anikait143:
### Description

- This commit contains Support for TSQL syntax for geospatial functions.

- Added Handling for STX, STY, LAT, LONG, STDistance, STAsText, and STAsBinary functions.

- Modified Test Cases to check TSQL syntax for geospatial functions

Supported Syntax - 

1. Use in a full column name with function name at the last (i.e. full_column_name DOT column_Geospatial_function)
```
select <TableName>.<Column_name>.STX
select <TableName>.<Column_name>.Lat
go
```

2. Use in a function_call statement with function name at the last (i.e. full_column_name DOT function_Geospatial_function)
```
select <TableName>.<Column_name>.STAsText()
select <TableName>.<Column_name>.STDistance(point_instance)
go
```

3. Use in a declared variable statement with function name at the last (i.e. local_id DOT Geospatial_function)
```
DECLARE @point geometry;
SET @point = geometry::STGeomFromText('POINT(-122.34900 47.65100)', 4326);
select @point.STX
select @point.STY
select @point.STAsText()
select @point.STDistance(@point)
go
```

4. Use in a complex function_call statement with function name at the last (i.e. function_call DOT Geospatial_function)
```
select geometry::STGeomFromText('POINT(-122.34900 47.65100)', 4326).STX
select geometry::STGeomFromText('POINT(-122.34900 47.65100)', 4326).STY
select geometry::STGeomFromText('POINT(-122.34900 47.65100)', 4326).STAsText()
select geometry::STGeomFromText('POINT(-122.34900 47.65100)', 4326).STDistance(point_instance)
go
```

5. Use in a complex subquery statement with function name at the last (i.e. subquery DOT Geospatial_function)
```
select ( select <geometryCol> From <geometryTable> ).STX 
select ( select <geometryCol> From <geometryTable> ).STY
select ( select <geometryCol> From <geometryTable> ).STAsText()
select ( select <geometryCol> From <geometryTable> ).STDistance(point_instance)
go
```

6. Use in a complex expression statement with function name at the last (i.e. ( expression ) DOT Geospatial_function)
```
DECLARE @point geometry;
SET @point = geometry::STGeomFromText('POINT(-122.34900 47.65100)', 4326);
select ( @point ).STX
select ( @point ).STY
select ( @point ).STAsText()
select ( @point ).STDistance(@point)
go
```

Task: BABEL- 4370
[...]

PR #2313, opened 2024-01-30
Modified Geospatial TSQL Syntax Support

Anikait143:
### Description

- This commit contains a few changes from https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/2271 :

- Moved Duplicated Code to common helper functions for Exit functions handling Geospatial functional call in both TSQLBuilder and TSQLCommonMutator.
 

- To handle local id's we are now quoting them in the geospatial handling itself rather than modifying their start positions, so as to remove the possibility of multiple rewrites within a single context.

* **Use case based -**  

* **Boundary conditions -**

* **Arbitrary inputs -**

* **Negative test cases -**

* **Performance tests -**

* **Tooling impact -**

* **Client tests -**

### Check List

### DIFF
--- a/contrib/babelfishpg_tsql/antlr/TSqlParser.g4
+++ b/contrib/babelfishpg_tsql/antlr/TSqlParser.g4
@@ -3436,10 +3436,7 @@ constant_LOCAL_ID
 // https://docs.microsoft.com/en-us/sql/t-sql/language-elements/expressions-transact-sql
 // Operator precendence: https://docs.microsoft.com/en-us/sql/t-sql/language-elements/operator-precedence-transact-sql
 expression
-    : local_id (DOT calls+=method_call)*                                        #local_id_expr
-    | subquery (DOT calls+=method_call)*                                        #subquery_expr
-    | LR_BRACKET expression RR_BRACKET (DOT calls+=method_call)*                #bracket_expr
-    | function_call (DOT calls+=method_call)*                                   #func_call_expr
+    : clr_udt_func_call                                                         #clr_udt_expr
     | expression collation                                                      #collate_expr
     | expression AT_KEYWORD TIME ZONE expression                                #time_zone_expr
     | op=(MINUS | PLUS | BIT_NOT) expression                                    #unary_op_expr
@@ -3455,6 +3452,13 @@ expression
     | odbc_literal                                                              #odbc_literal_expr
     | DOLLAR_ACTION                                                             #dollar_action_expr
     ;       
+
+clr_udt_func_call
+    : local_id (DOT calls+=method_call)*
+    | subquery (DOT calls+=method_call)*
+    | LR_BRACKET expression RR_BRACKET (DOT calls+=method_call)*
+    | function_call (DOT calls+=method_call)*
+    ;
 
 method_call
     : xml_methods
--- a/contrib/babelfishpg_tsql/src/tsqlIface.cpp
+++ b/contrib/babelfishpg_tsql/src/tsqlIface.cpp
@@ -183,6 +183,7 @@ template <class T> static void rewrite_function_call_geospatial_func_ref_no_arg(
 static void handleGeospatialFunctionsInFunctionCall(TSqlParser::Function_callContext *ctx);
 static void handleClrUdtFuncCall(TSqlParser::Clr_udt_func_callContext *ctx);
 static void handleFullColumnNameCtx(TSqlParser::Full_column_nameContext *ctx);
+template <class T> static void handleLocalIdQuotingFuncRefNoArg(T ctx, size_t geospatial_start_index, int &offset1, std::string &expr, std::vector<size_t> keysToRemove);
 static bool does_object_name_need_delimiter(TSqlParser::IdContext *id);
 static std::string delimit_identifier(TSqlParser::IdContext *id);
 static bool does_msg_exceeds_params_limit(const std::string& msg);
@@ -7910,22 +7911,8 @@ rewrite_geospatial_col_ref_query_helper(T ctx, TSqlParser::Method_callContext *m
 	keysToRemove.clear();
 	expr += ctx_str.substr(index);
 	
-	/* quoting Local ID's here so as to remove possibility of multiple rewrites in a single context */
-	for (auto &entry : local_id_positions)
-	{
-		if(entry.first >= ctx->start->getStartIndex() && entry.first <= geospatial_start_index)
-		{
-			int local_index = (int)entry.first - ctx->start->getStartIndex() + offset1;
-			if(expr.substr(local_index, entry.second.size()) ==  entry.second)
-			{
-				keysToRemove.push_back(entry.first);
-				expr = expr.substr(0, local_index) + "\"" + entry.second + "\"" + expr.substr(local_index + entry.second.size());
-				offset1 += 2;
-			}
-		}
-	}
-	for (const auto &key : keysToRemove) local_id_positions.erase(key);
-	keysToRemove.clear();
+	handleLocalIdQuotingFuncRefNoArg(ctx, geospatial_start_index, offset1, expr, keysToRemove);
+
 	std::string rewritten_exp = "(" + expr.substr(0, func_call_len + offset1 + 1) + ")." + expr.substr((int)method->start->getStartIndex() - ctx->start->getStartIndex() + offset1, method_len + 1);
 	rewritten_query_fragment.emplace(std::make_pair(ctx->start->getStartIndex(), std::make_pair(ctx_str.c_str(), rewritten_exp.c_str())));
 }
@@ -7962,22 +7949,8 @@ rewrite_geospatial_func_ref_no_arg_query_helper(T ctx, TSqlParser::Method_callCo
 	keysToRemove.clear();
 	expr += ctx_str.substr(index);
 	
-	/* quoting Local ID's here so as to remove possibility of multiple rewrites in a single context */
-	for (auto &entry : local_id_positions)
-	{
-		if(entry.first >= ctx->start->getStartIndex() && entry.first <= geospatial_start_index)
-		{
-			int local_index = (int)entry.first - ctx->start->getStartIndex() + offset1;
-			if(expr.substr(local_index, entry.second.size()) ==  entry.second)
-			{
-				keysToRemove.push_back(entry.first);
-				expr = expr.substr(0, local_index) + "\"" + entry.second + "\"" + expr.substr(local_index + entry.second.size());
-				offset1 += 2;
-			}
-		}
-	}
-	for (const auto &key : keysToRemove) local_id_positions.erase(key);
-	keysToRemove.clear();
+	handleLocalIdQuotingFuncRefNoArg(ctx, geospatial_start_index, offset1, expr, keysToRemove);
+	
 	std::string rewritten_exp = expr.substr((int)method->start->getStartIndex() - ctx->start->getStartIndex() + offset1, method_len) + expr.substr(0, func_call_len + offset1 + 1) + ")";
 	rewritten_query_fragment.emplace(std::make_pair(ctx->start->getStartIndex(), std::make_pair(ctx_str.c_str(), rewritten_exp.c_str())));
 }
@@ -8028,6 +8001,7 @@ rewrite_geospatial_func_ref_args_query_helper(T ctx, TSqlParser::Method_callCont
 	{
 		if(entry.first >= ctx->start->getStartIndex() && entry.first <= geospatial_start_index)
 		{
+			/* Here we are quoting Local Id's which are before the function name */
 			int local_index = (int)entry.first - ctx->start->getStartIndex() + offset1;
 			if(expr.substr(local_index, entry.second.size()) ==  entry.second)
 			{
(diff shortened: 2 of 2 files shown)


contrib/babelfishpg_tsql/src/tsqlIface.cpp
```diff
@@ -945,22 +926,7 @@ class tsqlCommonMutator : public TSqlParserBaseListener
 	 */
 	void exitLocal_id_expr(TSqlParser::Local_id_exprContext *ctx) override
```
KushaalShroff: Is there a better way NOT to use local id? Its the lowest level and its not optimal at all

KushaalShroff (changes requested):

contrib/babelfishpg_tsql/src/tsqlIface.cpp
Anikait143: So basically it's not local_id but it's Local_id_expr
```
expression    
       : local_id (DOT calls+=method_call)*                                        #local_id_expr
``` 
Here method_call can be spatial_methods

and yeah we cannot ignore this part as it is one important category of expression which needs to be handled for CLR_UDT methods for example:
1. @point.STX
2. @point.STAsText()
3. @point1.StDistance(@pnt)

These all expressions will fall under this category and need to be handled separetely, and it's not the lowest level as we already have handlings for same level expressions such as ```full_column_name```

contrib/babelfishpg_tsql/src/tsqlIface.cpp
```diff
@@ -8402,6 +8232,149 @@ rewrite_function_call_geospatial_func_ref_args(T ctx)
...
+		for (size_t i = 0; i < method_calls.size(); ++i)
+		{
+			TSqlParser::Method_callContext *method = method_calls[i];
+			/* rewriting the query in case of Geospatial function Call -> LR_BRACKET expression RR_BRACKET.method_call */
+			if(method->spatial_methods())
+			{
+				size_t ind;
+				std::string context = ::getFullText(ctx);
+				size_t spaces = 0;
+				for (size_t x = ctx->expression()->stop->getStopIndex() + 1 - ctx->start->getStartIndex(); x <= ctx->stop->getStopIndex() - ctx->start->getStartIndex(); x++)
```
KushaalShroff: What are we exactly doing here? Why cant we get the index of RR_BRACKET?

KushaalShroff (changes requested):

contrib/babelfishpg_tsql/src/tsqlIface.cpp
KushaalShroff: Why do we have to do it this way? Can we add a single rule directly into expression ?

KushaalShroff: Distinguish with method call keywords

contrib/babelfishpg_tsql/src/tsqlIface.cpp
```diff
@@ -180,6 +180,12 @@ template <class T> static void rewrite_geospatial_func_ref_no_arg_query_helper(T
 template <class T> static void rewrite_geospatial_func_ref_args_query_helper(T ctx, TSqlParser::Method_callContext *method, size_t geospatial_start_index);
 template <class T> static void rewrite_function_call_geospatial_func_ref_args(T ctx);
 template <class T> static void rewrite_function_call_geospatial_func_ref_no_arg(T ctx);
+static void handleGeospatialFunctionsInFunctionCall(TSqlParser::Function_callContext *ctx);
+static void handleGeospatialFunctionsInFuncCallExpr(TSqlParser::Func_call_exprContext *ctx);
+static void handleGeospatialFunctionsInLocalIdExpr(TSqlParser::Local_id_exprContext *ctx);
+static void handleGeospatialFunctionsInBracketExpr(TSqlParser::Bracket_exprContext *ctx);
+static void handleGeospatialFunctionsInSubqueryExpr(TSqlParser::Subquery_exprContext *ctx);
+static void handleFullColumnNameCtx(TSqlParser::Full_column_nameContext *ctx);
```
KushaalShroff: The question on reusable code raised by Rob, does it really help by shifting logic in another function?
I dont see a reason why The current change to introduce new and separate functions is useful

KushaalShroff (changes requested):

contrib/babelfishpg_tsql/src/tsqlIface.cpp
Anikait143: Done

contrib/babelfishpg_tsql/src/tsqlIface.cpp
Anikait143: That was exactly what Rob Pointed Out, He also used the same thing
See for example the handling of floating-point constants in exitConstant() ([here](https://github.com/babelfish-for-postgresql/babelfish_extensions/blob/BABEL_4_X_DEV/contrib/babelfishpg_tsql/src/tsqlIface.cpp#L2301) and [here](https://github.com/babelfish-for-postgresql/babelfish_extensions/blob/BABEL_4_X_DEV/contrib/babelfishpg_tsql/src/tsqlIface.cpp#L3147)).

contrib/babelfishpg_tsql/src/tsqlIface.cpp
```diff
@@ -8402,6 +8232,149 @@ rewrite_function_call_geospatial_func_ref_args(T ctx)
...
+				if (i == 0) ind = ctx->expression()->stop->getStopIndex() + 1 + spaces;
+				else ind = method_calls[i-1]->stop->getStopIndex();
+				rewrite_geospatial_query_helper(ctx, method, ind);
+			}
+		}
+	}
+}
+
+static void
+handleGeospatialFunctionsInSubqueryExpr(TSqlParser::Subquery_exprContext *ctx)
```
rishabhtanwar29: This and above 3 functions look quite similar. Can we combine these functions into one and handle the different parts using `if-else` instead?

Anikait143: Done

contrib/babelfishpg_tsql/src/tsqlIface.cpp
Anikait143: Ok got it, I have added an umbrella rule and now using this rule to distinguish between methods

contrib/babelfishpg_tsql/src/tsqlIface.cpp
```diff
@@ -8402,6 +8181,86 @@ rewrite_function_call_geospatial_func_ref_args(T ctx)
...
+	else if (ctx->geospatial_col())
+	{
+		/* Throwing error similar to TSQL as we do not allow 4-Part name for geospatial function call */
+		if(ctx->schema) throw PGErrorWrapperException(ERROR, ERRCODE_SYNTAX_ERROR, format_errmsg("The multi-part identifier \"%s\" could not be bound.", ::getFullText(ctx).c_str()), getLineAndPos(ctx));
+		
+		/* Rewriting the query as: table.col.STX -> (table.col).STX */
+		std::string rewritten_func_name = "(";
+		if(ctx->table) rewritten_func_name += stripQuoteFromId(ctx->table) + ".";
+		rewritten_func_name += stripQuoteFromId(ctx->column) + ")." + ::getFullText(ctx->geospatial_col());
+		rewritten_query_fragment.emplace(std::make_pair(ctx->start->getStartIndex(), std::make_pair(::getFullText(ctx), rewritten_func_name.c_str())));
```
rishabhtanwar29: We can probably insert the brackets `(` or `)` at the right places instead of rewriting the whole string.

contrib/babelfishpg_tsql/src/tsqlIface.cpp
```diff
@@ -8246,50 +7992,69 @@ rewrite_geospatial_func_ref_args_query_helper(T ctx, TSqlParser::Method_callCont
...
+			{
+				if((size_t)arg_offset_list[i].first < entry.first) local_index += arg_offset_list[i].second;
+			}
+			if(expr.substr(local_index, entry.second.size()) ==  entry.second)
+			{
+				keysToRemove.push_back(entry.first);
+				expr = expr.substr(0, local_index) + "\"" + entry.second + "\"" + expr.substr(local_index + entry.second.size());
+				offset2 += 2;
+				local_id_end_offset += 2;
+			}
```
rishabhtanwar29, line 8027: A comment would be helpful here to understand what we are trying to do.

contrib/babelfishpg_tsql/antlr/TSqlParser.g4
```diff
@@ -3436,10 +3436,7 @@ constant_LOCAL_ID
 // https://docs.microsoft.com/en-us/sql/t-sql/language-elements/expressions-transact-sql
 // Operator precendence: https://docs.microsoft.com/en-us/sql/t-sql/language-elements/operator-precedence-transact-sql
 expression
-    : local_id (DOT calls+=method_call)*                                        #local_id_expr
```
timchang514, line 3439: is there potential side effects from altering parser definition of expression like this, and is it necessary? IMO feels invasive

Anikait143: No there are no side effects for this change as we have just created an umbrella rule to handle CLR_UDT types, everything apart from them will work as before and it is necessary as without it we will be handling the functions with tagName which can be a concern

timchang514 (approved):

robverschoor (changes requested):
The loop `/* quoting Local ID's here so as to remove possibility of multiple rewrites in a single context */` is still duplicated 4 times in this code. Please replace with a generic function

shah-nirmit (approved):

contrib/babelfishpg_tsql/src/tsqlIface.cpp
```diff
@@ -8148,42 +7888,45 @@ rewrite_geospatial_col_ref_query_helper(T ctx, TSqlParser::Method_callContext *m
...
 			if(entry.first <= geospatial_start_index) offset1 += (int)entry.second.second.size() - entry.second.first.size();
-			else offset2 += (int)entry.second.second.size() - entry.second.first.size();
 		}
 	}
 	for (const auto &key : keysToRemove) rewritten_query_fragment.erase(key);
 	keysToRemove.clear();
+	expr += ctx_str.substr(index);
 	
-	/* shifting the local id positions to new positions after rewriting the query since they will be quoted later */
+	/* quoting Local ID's here so as to remove possibility of multiple rewrites in a single context */
```
rishabhtanwar29: Can we have a common function for quoting Local IDs? It is being done at a couple of places.

contrib/babelfishpg_tsql/src/tsqlIface.cpp
Anikait143: Done

contrib/babelfishpg_tsql/src/tsqlIface.cpp
Anikait143: Done

contrib/babelfishpg_tsql/src/tsqlIface.cpp
Anikait143: Done

contrib/babelfishpg_tsql/src/tsqlIface.cpp
```diff
@@ -8402,6 +8181,86 @@ rewrite_function_call_geospatial_func_ref_args(T ctx)
...
+		for (size_t i = 0; i < method_calls.size(); ++i)
+		{
+			TSqlParser::Method_callContext *method = method_calls[i];
+			/* rewriting the query in case of Geospatial function Call */
+			if(method->spatial_methods())
+			{
+				size_t ind = -1;
+				if (i == 0)
+				{
+					if(ctx->local_id()) ind = ctx->local_id()->stop->getStopIndex();
```
robverschoor, line 8215: Why is the test for whitespace is no longer needed?

Anikait143: We have directly used that particular index so no need to worry about the spaces in between

contrib/babelfishpg_tsql/src/tsqlIface.cpp
```diff
@@ -8402,6 +8157,108 @@ rewrite_function_call_geospatial_func_ref_args(T ctx)
...
+handleClrUdtFuncCall(TSqlParser::Clr_udt_func_callContext *ctx)
+{
+	/* checking if CLR_UDT types */
+	if(ctx != NULL && !ctx->DOT().empty())
+	{
+		std::vector<TSqlParser::Method_callContext *> method_calls = ctx->method_call();
+		for (size_t i = 0; i < method_calls.size(); ++i)
+		{
+			TSqlParser::Method_callContext *method = method_calls[i];
+			/* rewriting the query in case of Geospatial function Call */
```
robverschoor: Nit: use consistent spelling. this comment says "Geospatial function Call", the comment in the preeceding function says "geospatial function calls". Note difference in capitilization and plural form.
At some places comments mention "local_id", in others we speak of "Local ID's".

robverschoor (approved):

contrib/babelfishpg_tsql/src/tsqlIface.cpp
Anikait143: Done

robverschoor (approved):

contrib/babelfishpg_tsql/src/tsqlIface.cpp
rishabhtanwar29: I think we would be inserting two rewrites separately in the rewritten_query_fragment but I still see one. We are still rewriting the whole string.

rishabhtanwar29 (approved):

shardgupta (approved):

KushaalShroff (approved):

priyansx (commented):
LGTM

priyansx (commented):
LGTM

priyansx:
LGTM

Issue #997, opened 2023-01-03
[Bug]: sp_columns_100: Returned type character varying does not match expected type "varchar"


staticlibs:
### What happened?

I am running `BABEL_2_2_1__PG_14_5` on Fedora 36.

Executing the following T-SQL (extracted [from mssql-jdbc](https://github.com/microsoft/mssql-jdbc/blob/6b571af15233539fe8d0631a678a94115d2b5cc6/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDatabaseMetaData.java#L636)):

```
DROP TABLE IF EXISTS test1;
CREATE TABLE test1 (id INT, name VARCHAR(42));

DECLARE @mssqljdbc_temp_sp_columns_result TABLE(
    TABLE_QUALIFIER SYSNAME,
    TABLE_OWNER SYSNAME,
    TABLE_NAME SYSNAME,
    COLUMN_NAME SYSNAME,
    DATA_TYPE SMALLINT,
    TYPE_NAME SYSNAME,
    PRECISION INT,
    LENGTH INT,
    SCALE SMALLINT,
    RADIX SMALLINT,
    NULLABLE SMALLINT,
    REMARKS VARCHAR(254),
    COLUMN_DEF NVARCHAR(4000),
    SQL_DATA_TYPE SMALLINT,
    SQL_DATETIME_SUB SMALLINT,
    CHAR_OCTET_LENGTH INT,
    ORDINAL_POSITION INT,
    IS_NULLABLE VARCHAR(254),
    SS_IS_SPARSE SMALLINT,
    SS_IS_COLUMN_SET SMALLINT,
    SS_IS_COMPUTED SMALLINT,
    SS_IS_IDENTITY SMALLINT, 
    SS_UDT_CATALOG_NAME NVARCHAR(128),
    SS_UDT_SCHEMA_NAME NVARCHAR(128),
    SS_UDT_ASSEMBLY_TYPE_NAME NVARCHAR(max),
    SS_XML_SCHEMACOLLECTION_CATALOG_NAME NVARCHAR(128),
    SS_XML_SCHEMACOLLECTION_SCHEMA_NAME NVARCHAR(128),
    SS_XML_SCHEMACOLLECTION_NAME NVARCHAR(128),
    SS_DATA_TYPE TINYINT
);

INSERT INTO @mssqljdbc_temp_sp_columns_result EXEC sp_columns_100 'test1';

[...]

PR #2320, opened 2024-01-30
JTDS Driver support

### LINKED ISSUE #2137: Support JTDS driver
The JTDS driver for SQL Server is not currently supported with Babelfish. When connecting to the Babelfish endpoint an error is raised: 

Spring boot versions: 1.4.0 and 2.6.1
Java Version: 8
SQL Server JDBC drivers: MSSQL JDBC 6 and 9, JTDS 1.3.1

 
2022-05-31 15:12:13.025 [main] [INFO ] DividendApiApplication - Started DividendApiApplication in 46.102 seconds (JVM running for 50.844)
2022-05-31 15:13:13.288 [SimpleAsyncTaskExecutor-10] [ERROR] ConnectionPool - Unable to create initial connections of pool.
java.sql.SQLException: I/O Error: DB server closed connection.
       at net.sourceforge.jtds.jdbc.TdsCore.nextToken(TdsCore.java:2481)
       at net.sourceforge.jtds.jdbc.TdsCore.login(TdsCore.java:632)
       at net.sourceforge.jtds.jdbc.JtdsConnection.<init>(JtdsConnection.java:371)
       at net.sourceforge.jtds.jdbc.Driver.connect(Driver.java:184)
       at org.apache.tomcat.jdbc.pool.PooledConnection.connectUsingDriver(PooledConnection.java:310)
       at org.apache.tomcat.jdbc.pool.PooledConnection.connect(PooledConnection.java:203)

This has also been reported on GitHub: https://github.com/babelfish-for-postgresql/babelfish_extensions/issues/593

Investigation so far:
- With Babelfish, LOGIN packet is sent directly instead of PRELOGIN packet (for SSL 'ON' and 'OFF')
- With SQL Server, connection is made with SSL 'OFF' even with LOGIN packet send first. But with SLL 'ON', there is request and response from client and server but after that the client is timing out.

staticlibs:
### Description

This patch contains initial working implementation of JTDS support. With it applied, Babelfish DB can be opened and browsed from DBeaver (using "MS SQL Server / SQL Server (Old driver, jTDS)" connection setup included there). Both `ssl=off` and `ssl=require` modes are supported.

Two problems with JTDS were identified:

1. no `PRELOGIN` request sent when SSL is off

2. Babelfsh uses Partially Length-Prefixed bytes (PLP) [1] for returning `varchar` and `nvarchar` fields. PLP was introduced in TDS 7.2, but JTDS only supports TDS 7.1

Patch adds a message type check before reading `PRELOGIN` and a TDS version check to decide whether to use PLP responses.  

Known shortcomings of the patch:

1. no test runs with JTDS driver (would appreciate pointers how to introduce them, perhaps run a subset of existing JDBC tests with JTDS?)

2. not clear whether the patch covers all necessary PLP cases (should Spatial responses be changed too?)

[1] https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-tds/3f983fde-0509-485a-8c40-a9fa6679a828

### Issues Resolved

#2137

### Check List


contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
```diff
@@ -1848,12 +1848,14 @@ PrepareRowDescription(TupleDesc typeinfo, PlannedStmt *plannedstmt, List *target
...
 												att->attcollation, (atttypmod == -1) ?
-												atttypmod : (atttypmod - 4));
+												(tdsVersion >= TDS_VERSION_7_2 ? atttypmod : 0x7fff) :
+												(atttypmod - 4));
 				break;
 			case TDS_SEND_NVARCHAR:
 				SetColMetadataForCharTypeHelper(col, TDS_TYPE_NVARCHAR,
 												att->attcollation, (atttypmod == -1) ?
-												atttypmod : (atttypmod - 4) * 2);
+												(tdsVersion >= TDS_VERSION_7_2 ? atttypmod : 0x7ffe) :
```
Deepesh125: 1. what is meaning of 0x7ffe and 0x7fff? 
2. IIUC, we are still able to create table with [n]varchar(max) but we can only send 0x7fff amount of data (truncated)? Have we considered this scenario where you tried to insert the data of length > 0x7fff and try to retrieve it through JTDS driver?
3. don't we need to handle varbinary in the same way?

contrib/babelfishpg_tds/src/backend/tds/tdstypeio.c
```diff
@@ -2785,7 +2797,10 @@ TdsSendTypeVarchar(FmgrInfo *finfo, Datum value, void *vMetaData)
 	{
 		TDSInstrumentation(INSTR_TDS_DATATYPE_VARCHAR_MAX);
 
-		rc = TdsSendPlpDataHelper(destBuf, actualLen);
+		if (GetClientTDSVersion() >= TDS_VERSION_7_2)
```
Deepesh125: since we are setting 0x7fff as max_len in case of TDS v7.2, it will never come to this else block. Please correct me if I missed anything.

contrib/babelfishpg_tds/src/backend/tds/tdstypeio.c
```diff
@@ -2707,6 +2708,17 @@ TdsSendPlpDataHelper(char *data, int len)
...
+static int
+TdsSendVarlenDataHelper(char *data, int len) {
+	int			rc;
+	if ((rc = TdsPutInt16LE(len)) == 0)
+	{
+		TdsPutbytes(data, len);
+		TdsSendDone(TDS_TOKEN_DONE, TDS_DONE_COUNT, TDS_CMD_SELECT, 1);
+	}
+	return rc;
+}
```
Deepesh125: Please explain why is this needed, Please check my comment below.

Deepesh125 (changes requested):

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
staticlibs: > 1. what is meaning of 0x7ffe and 0x7fff?

Apologies, these constant values were incorrect (they are ignored by JTDS), correct value for them is `0xfffe` (max size for `BIGVARCHARTYPE` and `NVARCHARTYPE` from [2.2.5.4.3 Variable-Length Data Types](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-tds/ce3183a6-9d89-47e8-a02f-de5a1a1303de)). I've updated the constants to this value.


>     2. IIUC, we are still able to create table with [n]varchar(max) but we can only send 0x7fff amount of data (truncated)? Have we considered this scenario where you tried to insert the data of length > 0x7fff and try to retrieve it through JTDS driver?

Currently Babelfish will return an error on the attempt to insert or select `varchar(max)` value with length exceeding `0xfffe`. AFAIU the root cause here is the lack of actual support for TDS 7.1 types handling in Babelfish (that is still advertised in [its docs](https://docs.aws.amazon.com/AmazonRDS/latest/AuroraUserGuide/babelfish.html)). Babelfish uses the following TDS types for corresponding T-SQL types:

 - `varchar(max)` - `BIGVARCHARTYPE`
 - `nvarchar(max)` - `NVARCHARTYPE`
 - `varbinary(max)` - `BIGVARBINARYTYPE`

And for these types Babelfish uses PLP (Partially Length-prefixed Bytes, [2.2.5.2.3 Data Type Dependent Data Streams](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-tds/3f983fde-0509-485a-8c40-a9fa6679a828)) that was introduced in TDS 7.2 (and thus not supported in JTDS) and that allows to handle size larger than `0xfffe` for these types.

When MSSQL is run with JTDS using TDS 7.1 protocol it uses the following types mapping instead (without PLP):
 
 - `varchar(max)` - `TEXTTYPE`
 - `nvarchar(max)` - `NTEXTTYPE`
 - `varbinary(max)` - `IMAGETYPE`

And these types allow to handle size up to `0x7fffffff` without PLP.

I am not sure how to proceed with this problem - whether any changes to `babelfishpg_tds` regarding this type mapping would be acceptable or whether it is acceptable to have a limited JTDS support with max length of `varchar/nvarchar/varbinary(max)` limited to `65534`. This point it not addressed in the updated patch.

> 
>     3. don't we need to handle varbinary in the same way?

Yes, added the same max length handling for `varbinary(max)` in updated patch.

contrib/babelfishpg_tds/src/backend/tds/tdstypeio.c
staticlibs: As in previous comment, this is indeed not needed (was only needed before max_len change was added), I've removed all `tdstypeio.c` changes from the updated patch.

contrib/babelfishpg_tds/src/backend/tds/tdstypeio.c
staticlibs: You are correct, this branching is no longer needed when the max_len is set, I've removed it in the updated patch.

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
Deepesh125: For the point no 2, 

1. **Currently Babelfish will return an error on the attempt to insert or select varchar(max) value with length exceeding 0xfffe. AFAIU the root cause here is the lack of actual support for TDS 7.1 types handling in Babelfish** -- From where are getting this error? can you please some more information so that we can investigate from our side also.
2. **About handling datatypes specially for different TDS version** -- It is possible to handle [n]varchar(max) and varbinary(max) in a special way based on TDS protocol. For the reference, you can check how have we handled data types such as DATE for the TDS version < 7.3A. We conditionally set metadata[1] and call the sender function of nvarchar[2] to return the output. so we can follow the same here in this case also. 

But before doing that, lets come up with all the gaps first and address them in this pr only. 

Just small suggestion: Please try to use github link for the Microsoft documentation. 

[1]: https://github.com/babelfish-for-postgresql/babelfish_extensions/blob/BABEL_4_X_DEV/contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c#L1128
[2]: https://github.com/babelfish-for-postgresql/babelfish_extensions/blob/BABEL_4_X_DEV/contrib/babelfishpg_tds/src/backend/tds/tdstypeio.c#L3134

staticlibs:
Thanks for your comments and code pointers!

In patch update I've added support for running JDBC tests with JTDS. When `useJTDSInsteadOfMSSQLJDBC` test config options s set to `true`, then tests are read from a separate `jtds_jdbc_schedule` instead of the main `jdbc_schedule` file. And JTDS driver is used instead of mssql-jdbc one in all places in test suite where mssql-jdbc was used. It is not possible to have a mixed test that will use both JTDS and mssql-jdbc, I assumed this scenario is not necessary.

JTDS-specific tests `jtds-varchar-max-length-check.sql` and `jtds-nvarchar-max-length-check.sql` are added to check insert and select behaviour for `varchar(max)` and `nvarchar(max)`. Failures are currently commented out there, need to uncommented before running. Server handling of oversized `varchar(max)` and `nvarchar(max)` appeared to be different (added notes on this in test comments). I was not able to quickly reproduce oversized insert failures I've seen in ad-hoc tests, going to experiment more with them.

The intention here is to have a minimal working test-suite (should it be enabled as an additional GitHub action run?) with a minimal working JTDS patch, and then to iterate on the JTDS specific problems gradually extending the test suite. Please let me know whether such approach (basically to merge PR with the current backend patch and follow up in subsequent PRs) is acceptable, or whether JTDS support needs to be substantially improved (with the details from your previous comments) before merging the PR? In the latter case I can iterate on test suite in PR branch. I also can improve/adjust test harness changes as needed (tried to keep its changes to minimum).

Deepesh125:
> Thanks for your comments and code pointers!
> 
> In patch update I've added support for running JDBC tests with JTDS. When `useJTDSInsteadOfMSSQLJDBC` test config options s set to `true`, then tests are read from a separate `jtds_jdbc_schedule` instead of the main `jdbc_schedule` file. And JTDS driver is used instead of mssql-jdbc one in all places in test suite where mssql-jdbc was used. It is not possible to have a mixed test that will use both JTDS and mssql-jdbc, I assumed this scenario is not necessary.
> 
> JTDS-specific tests `jtds-varchar-max-length-check.sql` and `jtds-nvarchar-max-length-check.sql` are added to check insert and select behaviour for `varchar(max)` and `nvarchar(max)`. Failures are currently commented out there, need to uncommented before running. Server handling of oversized `varchar(max)` and `nvarchar(max)` appeared to be different (added notes on this in test comments). I was not able to quickly reproduce oversized insert failures I've seen in ad-hoc tests, going to experiment more with them.
> 
> The intention here is to have a minimal working test-suite (should it be enabled as an additional GitHub action run?) with a minimal working JTDS patch, and then to iterate on the JTDS specific problems gradually extending the test suite. Please let me know whether such approach (basically to merge PR with the current backend patch and follow up in subsequent PRs) is acceptable, or whether JTDS support needs to be substantially improved (with the details from your previous comments) before merging the PR? In the latter case I can iterate on test suite in PR branch. I also can improve/adjust test harness changes as needed (tried to keep its changes to minimum).

I agree with your suggested iterative approach. But there are still some areas where we dont have clarity yet, such as, what all is not working for JTDS?, We can file separate issues for them but who will be owner to drive that? How are we going to track all such details? 

And on side note, we should have separate GH action to test with JTDS driver.

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
Deepesh125: seems like mapping is not yet addressed in this PR? 

varchar(max) - TEXTTYPE
nvarchar(max) - NTEXTTYPE
varbinary(max) - IMAGETYPE

staticlibs: No, these TDS types usage is not implemented yet. The intention is to add a working test harness first, got the tests for the failures (some of them should be fixes by thee new types, but there are others in the same area - described in comments below) and then proceed with code changes. If the approach with JDBC  tests with `useJTDSInsteadOfMSSQLJDBC` looks like a right direction - I will proceed with actual tests and code changes one by one.

staticlibs:
> I agree with your suggested iterative approach. But there are still some areas where we dont have clarity yet, such as, what all is not working for JTDS?

I don't have the full picture yet. Basic DDL/DML/selects are working fine. Introspection also working (at least it is enough for DBeaver DB browser). But there are failures with inserts and selects for (n)varchar types, for lengths less then `0xfffe` so they seem to be not directly related to `(N)TEXTTYPE`. Known failures so far with test coverage:

1. nvarchar inserts with the length 4096 or longer [fail on backend](https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/2320/files#diff-64aa4a0bff72482d182bb2035d9db0d24e315b476ed2afba0e6508495fc33890R15)

2. nvarchar selected data with length 16384 or more is returned by backend, [but cannot be read by JTDS](https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/2320/files#diff-917e8107582659e292c85b91f3fe3223a479f7b06f503dc78ad2ec8bfe7545bdR68)

3. varchar selected data with length over 65535 is not returned by backend, instead of data [the backend error message is returned to client](https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/2320/files#diff-25a0e65be4be13f68efc9cbd8af92c2ca3105db285f254c9d14c7ba2b12a6ad5R68)

4. connection handling in JTDS is somehow different from mssql-jdbc and pgjdbc, when 2. and 3. failures above happen, JTDS cannot reopen the connection for subsequent test statements, I assume this needs to be investigated and fixed in the test harness

I intend to add more existing datatype tests to JTDS runs.

> We can file separate issues for them but who will be owner to drive that?

I am ready to continue with this task, cannot do that full time, but will be able to spare a few hours per week for it.


>  How are we going to track all such details?

I see 2 options: either commit the basic working test harness and then track each failure in a separate PR, or add more commits to this PR without merging it. I am fine with doing it either way.


> And on side note, we should have separate GH action to test with JTDS driver.

I've added [GH action run with JTDS](https://github.com/babelfish-for-postgresql/babelfish_extensions/actions/runs/8087368930?pr=2320), it is a copy of existing JDBC workflow with [useJTDSInsteadOfMSSQLJDBC variable set](https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/2320/files#diff-45c8306030991be69772d50352378e894495d117ae63eaae9026657bff118612R8).

thephantomthief:
Hi @staticlibs,

Thanks for adding support for the jTDS driver in JDBC Test Framework. You have already added tests for binary, varchar(max) and nvarchar(max) but to ensure we are fully able to support jTDS driver capabilities for Babelfish we will need to test the following:

- Datatypes
- Prepared Statements
- Callable Statements
- Transactions
- Cursors

We already have tests added for the above, and you can reuse the same tests to test Babelfish with jTDS driver. Similar to how you are reusing the TestBinary file. You can try adding the following files in your jTDS schedule file as a start:

- Datatypes
    * TestInt.txt
    * TestSmallInt.txt
    * TestTinyInt.txt
    * TestBigInt.txt
    * TestBit.txt
    * TestDecimal.txt
    * TestNumeric.txt
    * TestMoney.txt
    * TestFloat.txt
    * TestReal.txt
    * TestDatetime.txt
    * TestDate.txt
    * TestDatetime2.txt
    * TestSmallDatetime.txt
    * TestTime.txt
    * TestDateTimeOffset.txt
    * TestVarChar.txt
    * TestText.txt
    * TestChar.txt
    * TestBinary.txt
    * TestImage.txt
    * TestSqlVariant.txt
    * TestXML.txt
    * TestTableTypeFile.txt
    * TestUniqueIdentifier.txt
* SQL Batch
    * TestSQLQueries.txt
    * TestTransactionsSQLBatch.txt
    * TestSimpleErrorHandling.sql
    * TestErrorsWithTryCatch.sql
* Callable Statements
    * TestStoredProcedures.txt
* Cursors
    * TestCursorFetchNext.txt
    * TestCursorFetchPrev.txt
    * TestCursorPrepExecFetchNext.txt
    * TestCursorPrepExecFetchPrev.txt
* Transactions
    * TestTransactions.txt
    * TestTransactionSupportForProcedure.txt

Also, we need to make sure that the APIs we use in the JDBC test framework are compatible with the jTDS driver. If that is not the case, we will need to add support for the jTDSised version of the APIs in the framework. Additionally support for jTDS specific APIs (if any) should be added with tests for the same.

staticlibs:
Thanks for the details! And apologies for lack of activity on this, I was side-tracked by other Babelfish problems (prompted by usage of Cognos Analytics (#2404) and pgAgent (#2413) with Babelfish). I intend to return to this issue next week.

About the JDBC APIs: I am not too concerned about jTDS problems in the test harness (going to adjust the test harness as needed), but there is a concern about the public JDBC API (besides query execution) that is used by client apps. Foremost is the [DatabaseMetaData](https://docs.oracle.com/en/java/javase/17/docs/api/java.sql/java/sql/DatabaseMetaData.html), it can run non-trivial SQL for some of the calls and currently Babelfish is not fully compatible with it on mssql-jdbc (#997), let alone jTDS. I have a separate PR in preparation to add test coverage to these metadata calls (with new `dbmeta#!#...` commands in `.txt` tests) to be able to test these calls on both mssql-jdbc and jTDS (will add more details to that new PR when it is ready). Agree that driver-specific API coverage may be needed after the public API is covered.

staticlibs:
I've added all suggested existing tests to jTDS runs. For tests that have different output on jTDS, but behave the same on both MSSQL and Babelfish, I've created copies with `jtds-` prefix and adjusted output.

The following tests are found to be causing issues when run on jTDS:

 - TestTinyInt
 - TestDecimal
 - TestDatetime2
 - TestDatetimeoffset
 - TestVarChar
 - TestText
 - TestSQLVariant
 - TestXML
 - TestTableType
 - TestSimpleErrors
 - TestErrorsWithTryCatch
 - TestCursorFetchNext
 - TestCursorPrepExecFetchNext
 - TestTransactionSupportForProcedure

Currently test runs are disabled for these tests, I will follow-up on them one by one.

test/JDBC/src/main/java/com/sqlsamples/HandleException.java
```diff
@@ -61,6 +61,9 @@ else if(e instanceof SQLServerException) {
                     }
                     bw.write("~~ERROR (Message: "+ errorMsg + ")~~");
                 }
+                else { // JTDS throws plain SQLServerException
```
thephantomthief: Did you mean SQLException? We already handle SQLServerException in above else if condition

staticlibs: Yes, this is a typo, it throws plain `SQLException`. Will fix this and give harness changes another pass.

thephantomthief: Thanks. The test framework changes look good to me. Could you let us know how you plan on following up on the tests that are failing?

staticlibs: I've updated the following tests:

 - TestSimpleErrors
 - TestErrorsWithTryCatch
 - TestTransactionSupportForProcedure

They work correctly on jTDS, but behaviour of JDBC API methods (and the test output because of this) is slightly different. I intend to proceed with Cursors, XML and Varchar tests (that use methods unsupported on jTDS) and with remaining tests after that.

staticlibs:
Added TableType tests, they work correctly with slightly different output.

Investigated Cursor tests: cursors usage with jTDS is not supported on Babelfish. jTDS calls `sp_cursorfetch` procedure with unsupported `FETCH_INFO` option.

staticlibs:
Added TestXML, it works correctly with minor differences in metadata output.

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
staticlibs: Implemented the usage of  text/ntext/image column types for varchar(max)/nvarchar(max)/varbinary(max) results. Will write more details on remaining failures in a comment below.

staticlibs:
Remaining test failures (with comments for corresponding test groups):

 - TestDecimal
 - TestDatetime2

Fractional parts of records inserted with jTDS differ from ones inserted with mssql-jdbc.

 - TestDatetimeoffset

`select` output differs when `datetimeoffset` is being cast to `time` or `datetime2`.

 - TestVarChar
 - TestText

`insert` failure with `invalid data length or metadata length` for long input values.

 - TestSQLVariant

Multiple failures on both `insert` and `select`, have not looked deeper yet.

- TestCursorFetchNext
- TestCursorPrepExecFetchNext

These tests use cursor options unsupported on Babelfish.

staticlibs:
Fixed the problems with time and datetime2 fractional seconds and fixed the length check for incoming (n)varchar(max) long strings. Remaining failures:

 - TestDecimal
 - TestDatetime2
 - TestSQLVariant
 - TestCursorFetchNext
 - TestCursorPrepExecFetchNext

staticlibs:
Added support for returning sql_variant date values as nvarchar, it is clear now how to fix remaining TestSQLVariant failures with other date/time types.

staticlibs:
Added support for returning sql_variant time and datetime2 values as nvarchar. Also moved all "date-to-string" conversion code into `tdstimestamp.c` and used `pg_strftime` where applicable instead of manual date formatting.

Remaining failures:

 - `TestDecimal`: when `123.456` is inserted into `decimal(5,2)` column, Babelfish inserts `123.46` with mssql-jdbc and `123.456` with jTDS, root cause is currently unclear, received numeric data looks exactly the same with both mssql-jdbc and jTDS, parameter set RPC calls look different though.
 - `TestDatetime2`: when `2016-10-23 12:45:37.12` is inserted into `datetime2(6)` column, Babelfish inserts `2016-10-23 12:45:37.119` instead, root cause is currently unclear.
 - `TestCursorFetchNext`:  cursor fetch type 100 not supported
 - `TestCursorPrepExecFetchNext`: Tds `SP_CURSORPREPARE` not supported yet

staticlibs:
Added support for fetchtype `INFO` in [sp_cursorfetch](https://github.com/MicrosoftDocs/sql-docs/blob/58ab43939746c2f3879e9c3892f0a60b8abeb332/docs/relational-databases/system-stored-procedures/sp-cursorfetch-transact-sql.md) call, it fixes `TestCursorFetchNext` test for all cases except `cursor.last()` that in jTDS relies on the number of rows in cursor (that is not available in Babelfish). Excluded `cursor.last()` calls from the test.

Remaining failues:

 - `TestDecimal`
 - `TestDatetime2`
 - `TestCursorPrepExecFetchNext`

staticlibs:
Added (re-enabled existing) proof of concept support for [sp_cursorprepare](https://github.com/MicrosoftDocs/sql-docs/blob/58ab43939746c2f3879e9c3892f0a60b8abeb332/docs/relational-databases/system-stored-procedures/sp-cursorprepare-transact-sql.md) that allows to pass the following part of `TestCursorPrepExecFetchNext` on jTDS:

```
cursor#!#open#!#prepst#!#SELECT * FROM test_cursor_prep_exec_fetch_next WHERE a > @a #!#INT|-|a|-|-2#!#TYPE_SCROLL_INSENSITIVE#!#CONCUR_READ_ONLY#!#CLOSE_CURSORS_AT_COMMIT
cursor#!#fetch#!#next
```

Currently parameters count and types are hardcoded (only single INT parameter is supported), parsing for `sp_cursorprepare` second parameter needs to be implemented to support parameters properly. Also `@P0` style parameters in query body, which jTDS uses to replace `?` JDBC parameters, need to be replaced with `$1` style parameters for query body to be accepted by `SPI_prepare_cursor`.

staticlibs:
Added support for [sp_cursorprepare](https://github.com/MicrosoftDocs/sql-docs/blob/58ab43939746c2f3879e9c3892f0a60b8abeb332/docs/relational-databases/system-stored-procedures/sp-cursorprepare-transact-sql.md), this fixes `TestCursorPrepExecFetchNext` with the exception of `cursor.last()` that is excluded the same way as with `TestCursorFetchNext`.

Remaining failures (fractional parts of records inserted with jTDS differ from ones inserted with mssql-jdbc):
 - `TestDecimal`
 - `TestDatetime2`

staticlibs:
Fixed the remaining tests:

For `TestDecimal` filed #2678 and added a workaround described there (don't know if this workaround should go in - up to reviewers).

For `TestDatetime2`, added rounding up for fractional seconds conversion - there seems to be no side effects from it and MSSQL is doing it.

With all tests from the initial list passing now on jTDS - requesting a review: @Deepesh125 , @thephantomthief .

staticlibs:
Just for the record (if someone will need this for release notes in future), `useCursors` jTDS option ([all options list](https://jtds.sourceforge.net/faq.html)) is not supported on Babelfish. This option is `false` by default. When it is explicitly enabled in connection string, then jTDS sends the following query to `sp_cursoropen` RPC call when connection is opened:

```tsql
SELECT @@MAX_PRECISION
SET TRANSACTION ISOLATION LEVEL READ COMMITTED
SET IMPLICIT_TRANSACTIONS OFF
SET QUOTED_IDENTIFIER ON
SET TEXTSIZE 2147483647
```

This query is failing with `syntax error at or near "TRANSACTION"` when it is passed to `SPI_prepare_cursor`.

contrib/babelfishpg_tds/src/backend/tds/tdsrpc.c
```diff
@@ -2392,8 +2400,13 @@ HandleSPCursorOpenCommon(TDSRequestSP req)
 					pgstat_report_activity(STATE_RUNNING, activity);
 					pfree(activity);
 
+					if (req->dataParameter->isNull || req->metaDataParameterValue->len == 0)
+						paramdefstr = NULL;
+					else
+						paramdefstr = req->metaDataParameterValue->data;
```
Deepesh125, line 2406: Just for my understanding -- what is the test case? and how would this paramdefstr look?

contrib/babelfishpg_tds/src/backend/tds/tdsrpc.c
```diff
@@ -2428,7 +2441,13 @@ HandleSPCursorOpenCommon(TDSRequestSP req)
...
 
 	/* Send the response now */
-	SendCursorResponse(req);
+	if (req->spType == SP_CURSORPREPARE)
+	{
+		SendReturnValueIntInternal(1, req->cursorPreparedHandle);
+		TdsSendDone(TDS_TOKEN_DONEPROC, TDS_DONE_FINAL, 0xe0, 0);
+	}	
+	else
+		SendCursorResponse(req);
```
Deepesh125, line 2450: This basically states that we never tested this SP_CURSORPREAPRE code? how are we ensuring that we have enough test cases for this now?

contrib/babelfishpg_tds/src/backend/tds/tdslogin.c
```diff
@@ -1989,41 +1989,45 @@ TdsProcessLogin(Port *port, bool loadedSsl)
 	TdsErrorContext->phase = 0;
 	TdsErrorContext->reqType = TDS_LOGIN7;
 
-	PG_TRY();
+	if (!TdsReadNextBuffer() && TdsCheckMessageType(TDS_PRELOGIN))
```
Deepesh125, line 1992: we can have more safer check by checking whether client has ssl=off. If that's the case then only we should skip pre-login msg. Throw an FATAL error otherwise and should terminate connection

kuntalghosh, line 1992: 1. Let's add a comment why we may not expecting TDS_PRELOGIN as the first request.
2. The check should be
```
if (TdsReadNextBuffer() != EOF && TdsCheckMessageType(TDS_PRELOGIN))
```
3. If the prelogin message is not sent, what's the SSL behavior - TDS_ENCRYPT_NOT_SUP or TDS_ENCRYPT_OFF? OFF means the login will still happen in an encrypted manner.
4. According the code, the loadEncryption defaults to 0 which is TDS_ENCRYPT_OFF. So, we're freeing the ssl structure towards the end of the function without setting up the socket. Also, according to the code, the login is probably not happening in an encrypted manner.

Actually, let's not modify the SSL code at all. Something like as following (after confirming the SSL behavior):

```
/* comment why this is needed */
if (TdsReadNextBuffer() != EOF && TdsCheckMessageType(TDS_PRELOGIN))
{
        TdsErrorContext->err_text = "Parsing PreLogin Request";
        /* Pre-Login */
        rc = ParsePreLoginRequest();
        if (rc < 0)
            return rc;

        TdsErrorContext->err_text = "Make PreLogin Response";

        loadEncryption = MakePreLoginResponse(port, loadedSsl);
        TdsFlush();
}
else
{
        TDS_DEBUG(TDS_DEBUG3, "TDS7 Prelogin Message Skipped");
        loadEncryption = TDS_ENCRYPT_OFF;
}

```

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
```diff
@@ -1785,6 +1785,7 @@ PrepareRowDescription(TupleDesc typeinfo, PlannedStmt *plannedstmt, List *target
 		}
 
 		SetAttributesForColmetada(col);
+		col->atttypmod = atttypmod;
```
kuntalghosh, line 1788: Why is this needed? The actual typmod is calculated based on the datatype. Can we rely on this value?

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
```diff
@@ -2051,10 +2083,10 @@ PrepareRowDescription(TupleDesc typeinfo, PlannedStmt *plannedstmt, List *target
 					/*
 					 * If client being connected is using TDS version lower
 					 * than 7.3A then TSQL treats DATETIMEOFFSET as NVARCHAR.
-					 * Max len here would be 64('YYYY-MM-DD hh:mm:ss[.nnnnnnn]
+					 * Max len here would be 68('YYYY-MM-DD hh:mm:ss[.nnnnnnn]
 					 * [+|-]hh:mm'). and Making use of default collation Oid.
 					 */
-					SetColMetadataForCharTypeHelper(col, TDS_TYPE_NVARCHAR, serverCollationOid, 64);
+					SetColMetadataForCharTypeHelper(col, TDS_TYPE_NVARCHAR, serverCollationOid, 68);
```
kuntalghosh, line 2089: Have we added a testcase for this with TDS version lower than 7.3A?

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
```diff
@@ -2362,6 +2394,42 @@ SendReturnValueTokenInternal(ParameterToken token, uint8 status,
...
+	/* Flags */
+	TdsPutUInt16LE(0);		
+	/* TypeInfo */
+	TdsPutUInt8(TDS_TYPE_INTEGER);
+	TdsPutUInt8(0x04);
+	/* Value */
+	TdsPutUInt8(0x04);
+	TdsPutInt32LE(value);		
+}
+
```
kuntalghosh, line 2432: Why are we not using SendReturnValueTokenInternal()? We can create an empty parameter token using MakeEmptyParameterToken() and call the same function -
```
returnToken = MakeEmptyParameterToken("", TDS_TYPE_INTEGER, 4, 0);
returnToken->paramOrdinal = 0;

SendReturnValueTokenInternal(returnToken, 0x01, NULL,
                                 UInt32GetDatum(value), false, false);
```

contrib/babelfishpg_tsql/src/cursor.c
```diff
@@ -47,6 +47,8 @@ void		enable_sp_cursor_find_param_hook(void);
 void		disable_sp_cursor_find_param_hook(void);
 void		add_sp_cursor_param(char *name);
 void		reset_sp_cursor_params(void);
+InlineCodeBlockArgs *create_args(int numargs);
+void read_param_def(InlineCodeBlockArgs *args, const char *paramdefstr);
```
Deepesh125, line 51: extern?

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
```diff
@@ -2051,10 +2083,10 @@ PrepareRowDescription(TupleDesc typeinfo, PlannedStmt *plannedstmt, List *target
 					/*
 					 * If client being connected is using TDS version lower
 					 * than 7.3A then TSQL treats DATETIMEOFFSET as NVARCHAR.
-					 * Max len here would be 64('YYYY-MM-DD hh:mm:ss[.nnnnnnn]
+					 * Max len here would be 68('YYYY-MM-DD hh:mm:ss[.nnnnnnn]
```
Deepesh125, line 2086: Good catch

contrib/babelfishpg_tds/src/backend/tds/tdsrpc.c
```diff
@@ -3589,12 +3622,14 @@ ProcessRPCRequest(TDSRequest request)
...
 		case SP_CURSOROPEN:
-		case SP_CURSORPREPARE:
 		case SP_CURSORPREPEXEC:
 		case SP_CURSOREXEC:
 			GenerateBindParamsData(req);
 			HandleSPCursorOpenCommon(req);
 			break;
+		case SP_CURSORPREPARE:
+			HandleSPCursorOpenCommon(req);
+			break;
```
kuntalghosh, line 3632: GenerateBindParamsData() inititlizes some critical member of TDSRequest structure. Why are we not calling the same?

contrib/babelfishpg_tds/src/backend/tds/tdsrpc.c
kuntalghosh, line 2450: Why not modify SendCursorResponse and handle the scenarios if it's not handling a case already?

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
```diff
@@ -2362,6 +2394,42 @@ SendReturnValueTokenInternal(ParameterToken token, uint8 status,
 	(token->paramMeta.sendFunc) (finfo, datum, (void *) &token->paramMeta);
 }
 
+/*
+ * SendReturnValueIntInternal
+ *
+ * status - stored procedure (0x01) or UDF (0x02)
+ */
```
Deepesh125, line 2401: Please add comment explaining usage of this function.

Deepesh125: and should define macro for the constant

contrib/babelfishpg_tds/src/backend/tds/tdsrpc.c
```diff
@@ -2635,6 +2658,13 @@ HandleSPCursorFetchRequest(TDSRequestSP req)
 
 	TDSStatementEndCallback(NULL, NULL);
 
+	if (fetchType == SP_CURSOR_FETCH_INFO)
+	{
+		/* return rownum and nrows as output parameters */
+		SendReturnValueIntInternal(1, rownum);
+		SendReturnValueIntInternal(1, nrows);
```
kuntalghosh, line 2665: Should be
```
  if (req->cursorExtraArg2 && (req->cursorExtraArg2->flags & 0x01) == 1)
        SendReturnValueTokenInternal(req->cursorExtraArg2 0x01, NULL,
                                     UInt32GetDatum(rownum), false, false);

  if (req->cursorExtraArg3 && (req->cursorExtraArg3->flags & 0x01) == 1)
        SendReturnValueTokenInternal(req->cursorExtraArg3, 0x01, NULL,
                                     UInt32GetDatum(nrows), false, false);
```

contrib/babelfishpg_tds/src/backend/tds/tdstypeio.c
```diff
@@ -2775,9 +2776,18 @@ TdsSendTypeVarchar(FmgrInfo *finfo, Datum value, void *vMetaData)
...
 	TdsColumnMetaData *col = (TdsColumnMetaData *) vMetaData;
 
+	/*
+	* If client being connected is using TDS version lower than or equal to
+	* 7.1 then TSQL treats varchar(max) as Text.
+	*/
+	if (GetClientTDSVersion() <= TDS_VERSION_7_1_1 && 
+			col->metaEntry.type3.tdsTypeId == TDS_TYPE_TEXT)
+		return TdsSendTypeText(finfo, value, vMetaData);
+
```
kuntalghosh, line 2789: If we're handling it here, why do we need to handle in PrepareRowDescription as well? Also, this is not setting sendTableName to true although we're sending as text.

contrib/babelfishpg_tds/src/backend/tds/tdstimestamp.c
```diff
@@ -551,3 +553,197 @@ TdsGetTimestampFromDayTime(uint32 numDays, uint64 numMicro, int tz,
...
+	struct pg_tm tt,
+				*tm = &tt;
+	fsec_t		fsec;
+	char	   *buf;
+
+	memset(tm, '\0', sizeof(struct pg_tm));
+	time2tm(value, tm, &fsec);
+
+	buf = palloc0(16 + 1);
+	if (!pg_strftime(buf, 10 + 1, "%H:%M:%S", tm))
```
Deepesh125, line 642: Instead of using this number, please use with some good name.

contrib/babelfishpg_tds/src/backend/tds/tdstimestamp.c
```diff
@@ -551,3 +553,197 @@ TdsGetTimestampFromDayTime(uint32 numDays, uint64 numMicro, int tz,
...
+		ereport(ERROR,
+				(errcode(ERRCODE_DATETIME_VALUE_OUT_OF_RANGE),
+				 errmsg("date out of range")));
+
+	AppendFractionalSeconds(buf, fsec, scale);
+
+	return buf;
+}
```
Deepesh125, line 650: Please explain why is this needed instead of directly calling time_out? Worth adding comment explaining the same. Only thing I could make out this is to match the provided scale by appending extra zeros which can be very well achieved by appending zeros to output of time_out. What am I missing?

contrib/babelfishpg_tds/src/backend/tds/tdslogin.c
```diff
@@ -1989,41 +1989,45 @@ TdsProcessLogin(Port *port, bool loadedSsl)
 	TdsErrorContext->phase = 0;
```
kuntalghosh, line 1989: This is a really really large PR. We shouldn't merge this PR as a single commit as this is handling several new features together. Each feature requires its own explanation in the commit message:

Let's split the PR as per following:

1. JDBC test framework changes for jTDS driver testing
2. Prelogin changes
3. SP_cursorprepare
4. SP_cursorfetch with fetchinfo
5. Datatype changes (Individual datatypes should be addressed independently if the fixes are not related)

Each PR should be backed by respective testcases instead of putting all the test case in a single PR.

kuntalghosh (changes requested):

contrib/babelfishpg_tds/src/backend/tds/tdstimestamp.c
Deepesh125: And why is 10 being passed to pg_strftime? it should be 8, right?

contrib/babelfishpg_tds/src/backend/tds/tdstimestamp.c
```diff
@@ -551,3 +553,197 @@ TdsGetTimestampFromDayTime(uint32 numDays, uint64 numMicro, int tz,
...
+			oldval = value;
+			value /= 10;
+			remainder = oldval - value * 10;
+			buf[idx] = '0' + remainder;
+		}
+		else
+			buf[idx] = '0';
+	}
+
+	for (idx = 0; idx < scale; idx++)
```
Deepesh125, line 619: If column is defined with `TIME(7)` then `atttypmod = -1` and hence `scale` will also be `-1`. we will never ever get into this loop. But this part of code comes into picture only for TDS client with version < 7.3 for which we always need to  set scale to 7 (`DATETIMEOFFSETMAXSCALE`) if stored `attypmod` is `-1`. Our defined buf is also not able to handle this length 7. 

I am little bit more worried now in terms of how are testing this code, if existing test coverage good enough to save some seg fault. I am still inclined towards reusing `time_out` and either append zeros or truncate if required. Truncate will never be needed though.

contrib/babelfishpg_tds/src/backend/tds/tdstypeio.c
```diff
@@ -3355,6 +3411,33 @@ TdsSendTypeTime(FmgrInfo *finfo, Datum value, void *vMetaData)
 	return rc;
 }
 
+static int
+TdsSendDatetime2AsNVarcharHelper(FmgrInfo *finfo, Datum value, void *vMetaData)
```
Deepesh125, line 3415: Same as `TdsSendTimeAsNVarcharHelper`, please evaluate if we really need to do this?

contrib/babelfishpg_tds/src/backend/tds/tdstypeio.c
```diff
@@ -4155,6 +4325,33 @@ TdsRecvTypeDatetimeoffset(const char *message, const ParameterToken token)
 	return result;
 }
 
+static int
+TdsSendDatetimeoffsetAsNVarcharHelper(FmgrInfo *finfo, Datum value, void *vMetaData)
```
Deepesh125, line 4329: please reconsider if this is really needed?

contrib/babelfishpg_tds/src/backend/tds/tdstypeio.c
```diff
@@ -3772,6 +3855,48 @@ TdsTypeSqlVariantToDatum(StringInfo buf)
 	PG_RETURN_BYTEA_P(result);
 }
 
+/*
+ * dataformat: totalLen(4B) + baseType(1B) + metadatalen(1B) +
+ * encodingLen(5B) + dataLen(2B) + data(dataLen)
+ */
+static int
+TdsSendTypeSqlvariantAsNVarcharHelper(const char *st)
```
Deepesh125, line 3863: this can cause confusion with regular nvarchar sender for sql_variant datatype. Please add comment stating why is it needed

Deepesh125: Most of this code shares common code with `TdsSendTypeSqlvariant` to send string data, please consider if we can move this into some common helper code. It will help maintain code in long term

Deepesh125 (changes requested):

staticlibs:
@kuntalghosh , @Deepesh125 , thanks for the reviews! I've filed #2860 with the PRELOGIN change and copied review comments there. Going to follow up with JDBC test framework changes.

staticlibs:
Submitted test harness changes in #2861.

staticlibs:
Just FYI, I will be able to respond to review comments and follow up with cursor and datatype changes in about 2 weeks.

sftim:
@staticlibs, what support would help you best to move this forward?

staticlibs:
@sftim 

#2860 needs to be pushed (AFAIU it passed the review) to proceed with #2861 and the following parts of this PR.

PR #2352, opened 2024-02-07
Add check hook for pgstat_end_function_usage in fmgr

tanscorpio7:
### Description

In babelfish there could be cases where a procedure rollbacks/deletes it self during execution.
For example 
```
BEGIN TRAN
CREATE PROCEDURE poc1 AS ROLLBACK;
EXEC p;
```
Procedure 'proc1' does not exists after its execution since the transaction in which it was created
was rolled back by the procedure itself.
Now while executing rollback inside such procedures, the pg_stat entry for this proc is also deleted.
When we return to fmgr to end the proc invocation, we will overwrite on this freed block of memory.
In future, when alloc tries to allocate this freed block if memory to some other process,
it may hit an assert since this block was corrupted.

To avoid this, we add an additional check before writing on the pg_stat entry post invocation.
Check involves simply looking for a local cache pg_stat entry for given key.

Note* PG does not need this check since such a case is not possible there.

#### Performance Effect
Use the simplest of procedures to avoid diluting the performance changes and get the worst case scenario.
Before & after numbers are taken as average of 4 runs.
```
CREATE PROCEDURE proc1 AS SELECT 1;
CREATE PROCEDURE proc2 @in INT AS SELECT @in + 1;
```
**CASE 1**
```
DECLARE @i INT = 0;
DECLARE @dt DATETIME = getdate();
WHILE (@i < 1000000)
BEGIN
    EXEC proc1;
    SET @i = @i + 1;
END;
DECLARE @duration_ms int;
SET @duration_ms = DATEDIFF(ms, @dt, getdate());
SELECT @duration_ms
GO
```
Before Fix:
_track_functions -->_ disabled: 29189 ms enabled: 29367.75 **Change: +0.62%**
After Fix:
_track_functions -->_ disabled: 28261 ms enabled: 28266 **Change: ~0%**

**CASE 2**
```
DECLARE @i INT = 0;
DECLARE @dt DATETIME = getdate();
WHILE (@in < 1000000)
BEGIN
    EXEC proc2 1;
    SET @i = @i + 1;
END;
DECLARE @duration_ms int;
SET @duration_ms = DATEDIFF(ms, @dt, getdate());
SELECT @duration_ms
GO
```
Before Fix:
_track_functions -->_ disabled: 34730.25 ms enabled: 34732.25 **Change: ~0%**
After Fix:
_track_functions -->_ disabled: 33918.75 ms enabled: 34075.5 **Change: +0.4%**

Tests show no visible change in performance.

### Issues Resolved

[BABEL-4740]

### Sign Off

1. Enable track_function GUC for github actions otherwise execution does not even enter the pg_stat related code.
2. Different permutation and variation of example in description.
3. Existing test cases like TestTransactionSupportForProcedure & TestSimpleErrorsWithImplicitTran crashes without this fix when track function GUC is enabled.
4. Todo: Check how much the check hook deteriorates the performance.

### Check List

### DIFF
--- a/contrib/babelfishpg_tsql/src/hooks.c
+++ b/contrib/babelfishpg_tsql/src/hooks.c
@@ -27,6 +27,7 @@
 #include "commands/dbcommands.h"
 #include "commands/explain.h"
 #include "commands/tablecmds.h"
+#include "commands/trigger.h"
 #include "commands/view.h"
 #include "common/logging.h"
 #include "funcapi.h"
@@ -184,6 +185,9 @@ extern bool called_from_tsql_insert_exec();
 extern Datum pltsql_exec_tsql_cast_value(Datum value, bool *isnull,
 							 Oid valtype, int32 valtypmod,
 							 Oid reqtype, int32 reqtypmod);
+static void is_function_pg_stat_valid(FunctionCallInfo fcinfo,
+									  PgStat_FunctionCallUsage *fcu,
+									  char prokind, bool finalize);
 
 /*****************************************
  * 			Replication Hooks
@@ -253,6 +257,7 @@ static bbf_get_sysadmin_oid_hook_type prev_bbf_get_sysadmin_oid_hook = NULL;
 static transform_pivot_clause_hook_type pre_transform_pivot_clause_hook = NULL;
 static called_from_tsql_insert_exec_hook_type pre_called_from_tsql_insert_exec_hook = NULL;
 static exec_tsql_cast_value_hook_type pre_exec_tsql_cast_value_hook = NULL;
+static pltsql_pgstat_end_function_usage_hook_type prev_pltsql_pgstat_end_function_usage_hook = NULL;
 
 /*****************************************
  * 			Install / Uninstall
@@ -433,6 +438,9 @@ InstallExtendedHooks(void)
 
 	bbf_InitializeParallelDSM_hook = babelfixedparallelstate_insert;
 	bbf_ParallelWorkerMain_hook = babelfixedparallelstate_restore;
+
+	prev_pltsql_pgstat_end_function_usage_hook = pltsql_pgstat_end_function_usage_hook;
+	pltsql_pgstat_end_function_usage_hook = is_function_pg_stat_valid;
 }
 
 void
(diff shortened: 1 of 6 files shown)


contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -4920,4 +4926,20 @@ static Node* optimize_explicit_cast(ParseState *pstate, Node *node)
 		}
 	}
 	return node;
-}
\ No newline at end of file
+}
+
+/*
+ * in pltsql, pg_stat entry for a function may be dropped while
```
surendravishnoi: let us add more context about what leads to this drop

test/JDBC/expected/BABEL_4740.out
```diff
@@ -0,0 +1,350 @@
...
+GO
+EXEC babel_4740_p7
+GO
+COMMIT
+GO
+
+CREATE PROCEDURE babel_4740_p8 AS DROP PROCEDURE babel_4740_p8
+GO
+
+-- Execute this procedure 641 times
```
surendravishnoi: 64 times?

contrib/babelfishpg_tsql/src/hooks.c
tanscorpio7: added

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -4920,4 +4927,25 @@ static Node* optimize_explicit_cast(ParseState *pstate, Node *node)
 		}
 	}
 	return node;
-}
\ No newline at end of file
+}
+
+/*
+ * in pltsql, commit & rollback inside a function can drop
```
surendravishnoi: Pltsql allows rolling back parent transaction/sub-transaction while executing a procedure. The rollback will result into drop of all objects created in parent transaction/sub-transaction including the procedure itself if it is part of the active transaction. Also, PG will drop pg_stat entry for a procedure as part of rollback if the procedure is created in the same transaction. 
As a result, we can have a scenario where pg_stat entry for a procedure becomes invalid by the time it ends due to rollback inside it. This hook will validate if a pg_stat entry is valid or not before touching it.

surendravishnoi (approved):

PR #2359, opened 2024-02-13
sp_rename is not updating function/procedure name in babelfish_schema_permissions catalog

thephantomthief:
### Description

Currently, if we create procedures and functions in system databases such as master and rename those objects using sp_rename, their entries in the babelfish_schema_permissions catalog is not updated. This is because we were using the physical schema name to do the catalog lookup which would work if we were using a user defined T-SQL database in single-db mode.

In this commit, after we get the physical schema name of the procedure or function, we extract the logical schema name and then use that name to do the catalog lookup in babelfish_schema_permissions.

Task: BABEL-4768

* **Use case based -** Yes

* **Arbitrary inputs -** N/A

* **Performance tests -** N/A

* **Tooling impact -** N/A

* **Client tests -** N/A

### Check List


contrib/babelfishpg_tsql/sql/ownership.sql
```diff
@@ -28,6 +28,8 @@ CREATE TABLE sys.babelfish_schema_permissions (
   PRIMARY KEY(dbid, schema_name, object_name, grantee, object_type)
 );
 
+SELECT pg_catalog.pg_extension_config_dump('sys.babelfish_schema_permissions', '');
```
surendravishnoi: how did we come to the conclusion that dumping contents is the right thing to do during dump? what happens to corresponding DDLs (grant/revoke) during restore?

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -528,7 +528,8 @@ pltsql_pre_parse_analyze(ParseState *pstate, RawStmt *parseTree)
 				 */
 				if (relid == sysdatabases_oid ||
 					relid == namespace_ext_oid ||
-					relid == bbf_view_def_oid)
+					relid == bbf_view_def_oid ||
+					relid == bbf_schema_perms_oid)
```
surendravishnoi, line 532: is there a correlation between adding an entry here vs doing config dump for catalog table? It seems we only do it for tables which we do not config dump?

test/JDBC/upgrade/latest/schedule
```diff
@@ -509,4 +509,5 @@ BABEL-730
 babel-4475
 babel-3254
 babel-4517
-babelfish_inconsistent_metadata
\ No newline at end of file
+babelfish_inconsistent_metadata
+AAA
```
surendravishnoi, line 513: AAA?

test/JDBC/expected/AAA-vu-verify.out
```diff
@@ -0,0 +1,203 @@
...
+dbo#!#babel_4768_f2_new#!#128#!#master_babel_4768_u1#!#f#!#pg_catalog.int4
+dbo#!#babel_4768_p1_new#!#128#!#master_babel_4768_u1#!#p#!#
+dbo#!#babel_4768_p2_new#!#128#!#master_babel_4768_u1#!#p#!#sys.datetimeoffset
+dbo#!#babel_4768_t1_new#!#2#!#master_babel_4768_u1#!#r#!#<NULL>
+dbo#!#babel_4768_v1_new#!#2#!#master_babel_4768_u1#!#r#!#<NULL>
+~~END~~
+
+
+-- psql
+-- This is needed so that MVU passes
```
surendravishnoi, line 89: why this change?

test/JDBC/expected/AAA-vu-cleanup.out
```diff
@@ -0,0 +1,67 @@
+-- tsql
```
surendravishnoi, line 1: where are tests which make sure that dump/restore is working as expected?

test/JDBC/upgrade/latest/schedule
```diff
@@ -509,4 +509,5 @@ BABEL-730
 babel-4475
```
surendravishnoi, line 509: we should add it to normal JDBC run outside upgrade as well, just adding here is sufficient or we need to add to JDBC schedule as well?

contrib/babelfishpg_tsql/sql/ownership.sql
thephantomthief: right missed this. Actually this is not needed

test/JDBC/upgrade/latest/schedule
thephantomthief: This is running in normal JDBC test run. No need to explicitly add it in jdbc_schedule

contrib/babelfishpg_tsql/src/pl_handler.c
thephantomthief: This is not needed. Removing it.

PR #2377, opened 2024-02-19
Call pltsql_predicate_transformer for NOT expr argument

tanscorpio7:
### Description

in pltsql_predicate_transformer we recursively call the function for arguments for AND / OR clause.
We did not do this for NOT clause but there were cases where `LIKE` operator inside a `NOT` clause was not transformed because of this exclusion.
example: `SELECT COUNT(*) FROM babel_4566 WHERE 1=1 AND NOT OBJECT_NAME(id) LIKE '%Blah%'`
In this query the top clause is AND. So, we analyse its arguments individually. Then one of the predicates is a NOT clause and previously we directly returned it without transforming its like operator.
With this fix, the arguments of `NOT` will again be passed to pltsql_predicate_transformer for further break down and transformation.

As a fix we handle NOT clause similar to AND/OR and call the pltsql_predicate_transformer recursively for its argument.
NOT clause only has one argument, but ipltsql_predicate_transformer still needs to be called for its argument again.

#### Sign Off

### Issues Resolved

[BABEL-4566]

### Check List


ahmed-shameem (approved):
LGTM

contrib/babelfishpg_tsql/src/collation.c
```diff
@@ -519,7 +512,7 @@ pltsql_predicate_transformer(Node *expr)
 		{
 			Node	   *qual = (Node *) lfirst(lc);
 
-			if (is_andclause(qual) || is_orclause(qual))
+			if (IsA(qual, BoolExpr))
```
ahmed-shameem, line 515: Please add a comment specifying why this change.

contrib/babelfishpg_tsql/src/collation.c
```diff
@@ -485,13 +485,6 @@ pltsql_predicate_transformer(Node *expr)
 		}
 		else if (IsA(expr, BoolExpr))
 		{
-			if (boolexpr->boolop != AND_EXPR &&
```
Deepesh125, line 488: At line no 477, we have blindly assumed that predicate could be of two kinds, OpExpr and BoolExpr. Is that reasonable assumption? there can be other kind of predicate. Please check whether we need to handle any other predicate in special way?

tanscorpio7: We have considered other types of predicates [here](https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/2377/files#diff-b0c74cdd460dabbf601309526c2155b75cfbd7e5277e61fcb0f0cafe673cd710R489)

pgtsql_expression_tree_mutator also handles other types of predicates

Refactored the code and removed the boolexpr variable to avoid confusions

tanscorpio7: Looks like In the past we have added special handling for other cases (as we find out the missing cases)
[reference PR](https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/1565/files#diff-b0c74cdd460dabbf601309526c2155b75cfbd7e5277e61fcb0f0cafe673cd710R507)

Deepesh125 (approved):

PR #2392, opened 2024-02-26
Support unique constraints on nullable column with nulls not distinct

tanscorpio7:
### Description

Currently babelfish has supported UNIQUE indexes/constraints under the escape hatch _'escape_hatch_unique_constraint'_. This is because the current implementation allows duplicate NULL values in a column even when there is a UNIQUE constraint on it.
While the correct expected behaviour for pltsql UNIQUE indexes should be to treat NULL values as NOT DISTINCT. So only one NULL value should be allowed in UNIQUE index. (reject duplicate nulls in column)
We use PGs `null not distinct` option on column definition to implement this. So pltsql UNIQUE COLUMN CONSTRAINT and CREATE UNIQUE INDEX would implicitly have the `nulls not distinct` option and only single NULL value  will be allowed.

For UNIQUE COLUMN CONSTRAINT we re-write the query string in ANTLR parser while for CREATE INDEX we use the bison parser.

This PR also includes fix for NULLS ordering for UNIQUE indexes created through table or column constraints. Refer [engine PR](https://github.com/babelfish-for-postgresql/postgresql_modified_for_babelfish/pull/310) for more information.
Also INLINE INDEX is not currently supported and should throw unsupported error. Test cases added in BABEL-UNSUPPORTED
Remove the escape hatch _'escape_hatch_unique_constraint'_ since it is no longer needed.

### Issues Resolved

[BABEL-3571]

### Test Cases
ONLY TEST FILES WITH PREFIX BABEL_3571* ARE RELEVANT TO THE UNIQUE INDEXES ON NULLABLE COLUMN CHANGE. ALL OTHER TEST FILES ARE UPDATED SIMPLY TO REMOVE THE ESCAPE HATCH 'babelfishpg_tsql.escape_hatch_unique_constraint'
BABEL_3571.sql - covers all tests related to creating unique index/constraint & escape hatch behaviour
BABEL_3571-15_5-or-15_6-or-16_1-or-16_2 - upgrade tests for changes
BABEL_3571-before-15-5 - upgrade tests for changes
Created two different upgrade tests i.e. one for before 15-5 and one after since we changed the NULLS ORDERING for indexes created thorugh CREATE INDEX commands.

For all the other test cases removed the escape hatch 'babelfishpg_tsql.escape_hatch_unique_constraint' from test file
which some times required creating a new upgrade test file.

Also removed BABEL-2619 these tests checked if creating unique index/constraints on nullable columns is blocked & only allowed if escape hatch is set, which is no longer required.

#### Sign Off

### Check List


contrib/babelfishpg_tsql/src/guc.c
```diff
@@ -1517,16 +1516,6 @@ define_escape_hatch_variables(void)
 							 PGC_USERSET,
 							 GUC_NOT_IN_SAMPLE | GUC_DISALLOW_IN_FILE | GUC_DISALLOW_IN_AUTO_FILE,
 							 NULL, NULL, NULL);
-
-	DefineCustomEnumVariable("babelfishpg_tsql.escape_hatch_unique_constraint",
```
forestkeeper, line 1521: I have one question , we explicit the usage of escape_hatch_unique_constraint in doc : https://babelfishpg.org/docs/usage/escape_hatches/ , and if we are deleting this guc, we'll have to make sure it's continuous and safe for customer.

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -4673,3 +4680,23 @@ is_function_pg_stat_valid(FunctionCallInfo fcinfo, PgStat_FunctionCallUsage *fcu
...
+	{
+		switch (ordering)
+		{
+			case SORTBY_ASC:
+			case SORTBY_DEFAULT:
+				return SORTBY_NULLS_FIRST;
+			case SORTBY_DESC:
+				return SORTBY_NULLS_LAST;
+			default:
+				return SORTBY_NULLS_DEFAULT;
```
forestkeeper, line 4697: what other ordering will have in here ?

forestkeeper:
Could you show me where is the code place for : only single NULL value will be allowed

contrib/babelfishpg_tsql/src/tsqlIface.cpp
```diff
@@ -6846,6 +6848,8 @@ post_process_column_definition(TSqlParser::Column_definitionContext *ctx, PLtsql
 static void
 post_process_table_constraint(TSqlParser::Table_constraintContext *ctx, PLtsql_stmt_execsql *stmt, TSqlParser::Ddl_statementContext *baseCtx)
 {
+	if (ctx->UNIQUE())
+		rewritten_query_fragment.emplace(std::make_pair(ctx->UNIQUE()->getSymbol()->getStopIndex()+1 , std::make_pair("", " NULLS NOT DISTINCT")));
```
forestkeeper, line 6875: Is there any other places need to be replaced ?
Will this rewrite resued with other sql syntax ? ( like alter )

contrib/babelfishpg_tsql/src/hooks.c
tanscorpio7: `SORTBY_USING` is the only other type, but that option is never possible here.
I am actually reusing this bit of code from here [link](https://github.com/babelfish-for-postgresql/babelfish_extensions/blob/3fcd5ba305c108539a368f67c6a789b72e71dc57/contrib/babelfishpg_tsql/src/backend_parser/gram-tsql-epilogue.y.c#L1893),

contrib/babelfishpg_tsql/src/guc.c
tanscorpio7: Yes, changed the behaviour to send out a deprecate message instead of error [link](https://github.com/babelfish-for-postgresql/babelfish_extensions/blob/0b61d9ed5d388d482728ef1f7de33b48c2d22159/contrib/babelfishpg_tsql/sql/sys_procedures.sql#L141)
sp_babelfish_configure which is the suggested way to set escape hatches will not throw an error now. But set_config() will still error out.

contrib/babelfishpg_tsql/src/tsqlIface.cpp
tanscorpio7: `CREATE TABLE`, `ALTER TABLE ... ADD CONSTRAINT`, `CREATE INDEX` only these 3 need rewriting.
This function will handle create table and alter table rewritting.

tanscorpio7: only other SQL where unique constraint is possible is inline index, which is not supported in babelfish.
I've added proper unsupported error for inline indexes

tanscorpio7:
> Could you show me where is the code place for : only single NULL value will be allowed

tanscorpio7:
> Could you show me where is the code place for : only single NULL value will be allowed

tanscorpio7:
> Could you show me where is the code place for : only single NULL value will be allowed

@forestkeeper Where PG implements nulls not distinct behaviour ? 
looks like this is the [PR LINK](https://github.com/tanscorpio7/postgresql_modified_for_babelfish/commit/94aa7cc5f707712f592885995a28e018c7c80488#diff-ec99f9d40c4ae8c8d3df9539b022ab622368aa3b746de86aa21f12bed110bbe7R)
and at this point they are forcing a full uniqueness check [link to line](https://github.com/tanscorpio7/postgresql_modified_for_babelfish/commit/94aa7cc5f707712f592885995a28e018c7c80488#diff-ec99f9d40c4ae8c8d3df9539b022ab622368aa3b746de86aa21f12bed110bbe7R172)
If there were any nulls in the new tuple, seems they simply skip the check altogether if nulls_not_distinct is not set.

test/JDBC/expected/parallel_query/BABEL_3571.out
```diff
@@ -0,0 +1,850 @@
...
+-1#!#-a#!#0001
+0#!#<NULL>#!#<NULL>
+1#!#+a#!#1001
+9999#!#random#!#<NULL>
+~~END~~
+
+~~START~~
+text
+Query Text: SELECT * FROM babel_3571_1 ORDER BY id1
+Gather  (cost=0.15..60.90 rows=850 width=68) (actual rows=5 loops=1)
```
forestkeeper: Should turn cost off, so it'll not mess up in future releases.

test/JDBC/expected/parallel_query/BABEL_3571.out
```diff
@@ -0,0 +1,850 @@
...
+  Workers Planned: 1
+  Workers Launched: 1
+  Single Copy: true
+  ->  Index Scan using babel_3571_1_id1_key on babel_3571_1  (cost=0.15..60.90 rows=850 width=68) (actual rows=5 loops=1)
```
forestkeeper: Should turn cost off, so it'll not mess up in future releases.

contrib/babelfishpg_tsql/src/guc.c
KushaalShroff: Our release documents shall be updated with this information

contrib/babelfishpg_tsql/src/tsqlIface.cpp
```diff
@@ -6778,6 +6778,8 @@ void process_execsql_remove_unsupported_tokens(TSqlParser::Dml_statementContext
 static void
 post_process_column_constraint(TSqlParser::Column_constraintContext *ctx, PLtsql_stmt_execsql *stmt, TSqlParser::Ddl_statementContext *baseCtx)
 {
+	if (ctx->UNIQUE())
+		rewritten_query_fragment.emplace(std::make_pair(ctx->UNIQUE()->getSymbol()->getStopIndex()+1 , std::make_pair("", " NULLS NOT DISTINCT")));
```
KushaalShroff, line 6782: What are the implications on explain and DDL export?

tanscorpio7: DDL export from TDS endpoint will remain unaffected, since nulls_not_distinct is a PG feature. No bbf catalog has this information.
From psql endpoint i.e. pg_dump will will start dumping the ddl for index/constraints with the NULLS NOT DISTINCT  keyword.

confirmed with explain (simple query on table with unique index) that queries will use the index/ constraint's index whenever possible. The NULLS FRIST/LAST fix in this PR fixes this for constraints.

tanscorpio7: We also need to fix the pg_dump later for constraints in babelfish.
Currently the ordering is not dumped. So we get something like this for constraints.
`ALTER TABLE... ADD CONSTRAINT ..... (col1, col2)`
NULLS ORDERING & ASC/DESC are not included. So if user does a dump & restore or MVU then ordering information is lost which means some queries would not be able to use the constraints index.

KushaalShroff (dismissed):
Please add negative test case for alter index

jsudrik (approved):
Approved based on other reviews.

PR #2394, opened 2024-02-26
Could commpile with BABEL_4_X_DEV__PG_16_X

japinli:
Hi,

When I try compile [BABEL_4_X_DEV](https://github.com/babelfish-for-postgresql/babelfish_extensions/commit/938b6fc9f7ba5f66f9a7bc8455f1b70b6b7531c3) with [postgresql_modified_for_babelfish/BABEL_4_X_DEV__PG_16_X](https://github.com/babelfish-for-postgresql/postgresql_modified_for_babelfish/commit/65c628c3335ef0afcd637a49c5fdf24b090e68e3), here are some erors:

1. `BIO_[get/set]_data()` is replace with `BIO_[get/set]_app_data()` in [21b3a29](https://github.com/babelfish-for-postgresql/postgresql_modified_for_babelfish/commit/21b3a29565dff7baa2548fd779a0afe737b42f33).
2. Some multiple definition errors.

Here are two commits for this, please take a look!

### DIFF
--- a/contrib/babelfishpg_tds/src/backend/tds/tds-secure-openssl.c
+++ b/contrib/babelfishpg_tds/src/backend/tds/tds-secure-openssl.c
@@ -855,11 +855,6 @@ Tds_be_tls_write(Port *port, void *ptr, size_t len, int *waitfor)
  * to retry; do we need to adopt their logic for that?
  */
 
-#ifndef HAVE_BIO_GET_DATA
-#define BIO_get_data(bio) (bio->ptr)
-#define BIO_set_data(bio, data) (bio->ptr = data)
-#endif
-
 static int
 my_sock_read(BIO * h, char *buf, int size)
 {
@@ -867,7 +862,7 @@ my_sock_read(BIO * h, char *buf, int size)
 
 	if (buf != NULL)
 	{
-		res = secure_raw_read(((Port *) BIO_get_data(h)), buf, size);
+		res = secure_raw_read(((Port *) BIO_get_app_data(h)), buf, size);
 		BIO_clear_retry_flags(h);
 		if (res <= 0)
 		{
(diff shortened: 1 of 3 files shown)


contrib/babelfishpg_tsql/src/backend_parser/parser.c
```diff
@@ -29,8 +29,6 @@
 #include "src/pltsql.h"
 #include "tcop/tcopprot.h"
 
-int			pgtsql_base_yydebug;
```
thephantomthief: These changes don't look relevant to the SSL changes. Are these needed? Is it causing build failure? Because our github actions are not showing any compilation errors.

contrib/babelfishpg_tsql/src/pl_comp.c
```diff
@@ -62,9 +62,6 @@ bool		pltsql_DumpExecTree = false;
 bool		pltsql_check_syntax = false;
 
 PLtsql_function *pltsql_curr_compile;
-int			pltsql_curr_compile_body_lineno;	/* lineno of
```
thephantomthief: Same here

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -90,7 +90,7 @@
 
 #include "access/xact.h"
 
-extern bool escape_hatch_unique_constraint;
```
thephantomthief: Same here

contrib/babelfishpg_tsql/src/backend_parser/parser.c
japinli: Yeah, it is not relevant to the SSL changes, so I created a separate commit.

It seems that the multiple definitions and mismatch declaration are related to LLVM.

thephantomthief: Can you put those in a separate pull request? We squash and merge PRs so these changes would also get bundled up in a single commit in upstream.

thephantomthief (approved):

shardgupta (approved):

PR #2398, opened 2024-02-27
Skip replication of Babelfish catalog tables

rishabhtanwar29:
### Description
Babelfish has several extended catalogs in the form of normal PG tables, so it is
possible for customers to replicate them which could result in undefined behavior.
This commit addresses the above scenario by marking all such tables as catalog
tables.

Task: BABEL-4721

Engine PR: https://github.com/babelfish-for-postgresql/postgresql_modified_for_babelfish/pull/323

* **Use case based -**

* **Boundary conditions -**

* **Arbitrary inputs -**

* **Negative test cases -**

* **Performance tests -**

* **Tooling impact -**

* **Client tests -**

### Check List


contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -214,7 +239,18 @@ initTsqlSyscache()
...
+	 * catalog accesses.
+	 */
+	if (relationId == sysdatabases_oid || relationId == bbf_function_ext_oid ||
+		relationId == namespace_ext_oid || relationId == bbf_authid_login_ext_oid ||
+		relationId == bbf_authid_user_ext_oid || relationId == bbf_view_def_oid ||
+		relationId == bbf_servers_def_oid || relationId == bbf_domain_mapping_oid ||
+		relationId == bbf_extended_properties_oid || relationId == bbf_assemblies_oid ||
+		relationId == bbf_configurations_oid || relationId == bbf_helpcollation_oid ||
+		relationId == bbf_syslanguages_oid || relationId == bbf_service_settings_oid ||
+		relationId == spt_datatype_info_table_oid || relationId == bbf_versions_oid)
```
kuntalghosh: 1. What's performance impact of doing 16 condition check as part of isCatalogRelation()? For example, you can test simple heap_insert and check the perf.
2. In CacheInvalidateHeapTuple(), we're checking
```
if (!IsCatalogRelation(relation) && (!IsExtendedCatalogHook || !IsExtendedCatalogHook(relation->rd_id)))
```
and again in IsCatalogRelation(), we're checking
```
if (IsExtendedCatalogHook && IsExtendedCatalogHook(relid))
```
Please remove the first check.
3. As we increase the number, this will become unmanageable. How can we add sufficient test to easily detect if someone adds or removes a new catalog table?

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -1282,14 +1318,24 @@ clean_up_bbf_server_def()
...
+	PG_TRY();
+	{
+		allowSystemTableMods = true;
+		exec_utility_cmd_helper(query_str);
+	}
+	PG_FINALLY();
+	{
+		allowSystemTableMods = prev_allowSystemTableMods;
+	}
+	PG_END_TRY();
```
kuntalghosh: This is not the right way of doing this change (although I'm seeing this practice in our code). Please use this format:
```
int save_nestlevel = NewGUCNestLevel();
...
SetConfigOption(<set the GUC here>);
...
AtEOXact_GUC(false, save_nestlevel);
```
Also, why do you need a TRY..CATCH here? The GUC will be anyway reset at abort transaction code path.

kuntalghosh (changes requested):

contrib/babelfishpg_tsql/src/catalog.c
rishabhtanwar29: 1. I checked that it's not posing any significant impact, comparable runtime with/without the changes for 100K inserts, checked the flamegraph as well.
2. Done, raised engine side PR : https://github.com/babelfish-for-postgresql/postgresql_modified_for_babelfish/pull/323
3. I think that the number of extended catalog tables should not become too high but your point is correct regarding be able to detect it. I have added the test in a way so that it fails if an extended catalog relation gets added or removed.

contrib/babelfishpg_tsql/src/catalog.c
rishabhtanwar29: TRY..CATCH is mostly to follow the existing convention as well as to prevent any leaks in case certain error path results in GUC value not getting reset. But I have updated the code as you suggested.

kuntalghosh (dismissed):

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -230,7 +251,19 @@ initTsqlSyscache()
 bool
 IsPLtsqlExtendedCatalog(Oid relationId)
 {
-	if (relationId == sysdatabases_oid || relationId == bbf_function_ext_oid)
+	/*
+	 * The assumption of parent function is that it should not perform any
+	 * catalog accesses.
+	 */
+	if (relationId == sysdatabases_oid || relationId == bbf_function_ext_oid ||
```
surendravishnoi: will this not have impact on dump/restore, pg upgrade?

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -1298,14 +1331,19 @@ clean_up_bbf_server_def()
 {
 	char 			*query_str;
 	StringInfoData 	query;
+	int save_nestlevel;
 
 	initStringInfo(&query);
 
 	appendStringInfo(&query, "TRUNCATE TABLE sys.babelfish_server_options CASCADE");
```
surendravishnoi: when do we call this? do we have a test case for this?

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -1298,14 +1331,19 @@ clean_up_bbf_server_def()
...
 
 	query_str = query.data;
 
+	save_nestlevel = NewGUCNestLevel();
+	SetConfigOption("allow_system_table_mods", "on",
+					GUC_CONTEXT_CONFIG, PGC_S_SESSION);
 	exec_utility_cmd_helper(query_str);
```
surendravishnoi: should we add try/catch to avoid failure case which can leak the option setting? also, why not re-set the old value but use nesting level to rollback?

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--3.0.0--3.1.0.sql
```diff
@@ -2120,11 +2120,15 @@ END;
 $$ LANGUAGE 'pltsql';
 GRANT EXECUTE ON PROCEDURE sys.sp_linkedservers TO PUBLIC;
 
+SET allow_system_table_mods = on;
```
surendravishnoi, line 2123: so we are only ignoring DDLs but not DMLs? will we propagate DMLs or they will be ignored via other mechanism?

contrib/babelfishpg_tsql/src/catalog.c
rishabhtanwar29: This was initially suggested by Kuntal in the above comment to use GUC nesting level instead of try-catch, do you not agree with that?

contrib/babelfishpg_tsql/src/catalog.c
rishabhtanwar29: It is called during DROP EXTENSION tds_fdw, we already have tests for this like BABEL-4168.

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--3.0.0--3.1.0.sql
rishabhtanwar29: Not sure if I understood your question but the use of `allow_system_table_mods` is to allow DDLs on bbf catalog tables. The DML propagation(skipped) will be skipped by marking them catalog tables, see `IsPLtsqlExtendedCatalog` below.

contrib/babelfishpg_tsql/src/catalog.c
rishabhtanwar29: Ahh, I verified that this will not affect pg_upgrade as it happens through superuser but we would need to skip this hook during restore to allow DMLs on bbf catalogs. Added the babelfish_dump_restore check in the newer revision.

contrib/babelfishpg_tsql/src/catalog.c
rishabhtanwar29: I have changed the logic to directly truncate the relation, since otherwise we would be hitting permission denied error trying to truncate a system table.

surendravishnoi (approved):

PR #2415, opened 2024-03-07
Support grant/revoke on schema feature

shalinilohia50:
### Description

- Support GRANT/REVOKE .. ON SCHEMA .. in Babelfish
- Supported syntax GRANT <permission> ON SCHEMA::<schema_name> TO <user_name>
- Supported syntax REVOKE <permission> ON SCHEMA::<schema_name> TO/FROM <user_name>
- We have introduced a catalog sys.babelfish_schema_permission to hold the information about objects and its permissions.
- GRANT on schema/objects add/update a row in the catalog table if does not exist already.
- REVOKE on schema/objects removes/update the corresponding row in the catalog table if it exists already.
- Drop statement for OBJECT/SCHEMA removes all the relevant object entries from the catalog.
- sp_rename updates the object name in sys.babelfish_schema_permission catalog so that the permission is tracked correctly in the catalogs
- Added metadata check for schema permissions catalog
- GRANT/REVOKE ... ON ... SCHEMA ... WITH GRANT OPTION/CASCADE does not work as expected -> Throw unsupported error
- CREATE SCHEMA ... GRANT ... ON ... SCHEMA throws syntax error -> Throw unsupported error
- REVOKE GRANT OPTION FOR ... ON SCHEMA should be unsupported in Babelfish

- Specific to Upgrade: The catalogs will not track the permissions granted on the OBJECTs on the earlier versions.

### Issues Resolved

Task: [BABEL-4344](https://jira.rds.a2z.com/browse/BABEL-4344), [BABEL-4775](https://jira.rds.a2z.com/browse/BABEL-4775), [BABEL-4485](https://jira.rds.a2z.com/browse/BABEL-4485), [BABEL-4739](https://jira.rds.a2z.com/browse/BABEL-4739), [BABEL-4738](https://jira.rds.a2z.com/browse/BABEL-4738), [BABEL-4758](https://jira.rds.a2z.com/browse/BABEL-4758)

* **Use case based - Added in JDBC**
* **Boundary conditions -**
* **Arbitrary inputs -**
* **Negative test cases - Added in JDBC**
* **Minor version upgrade tests - Added**
* **Major version upgrade tests - Added**
* **Performance tests -N/A**
* **Tooling impact -**
* **Client tests -**


contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -2799,6 +2997,712 @@ rename_procfunc_update_bbf_catalog(RenameStmt *stmt)
...
+			}
+		}
+		tuple_bbf_schema = systable_getnext(scan);
+	}
+	systable_endscan(scan);
+	table_close(bbf_schema_rel, AccessShareLock);
+}
+
+void
+exec_internal_grant_on_function(const char *logicalschema,
```
thephantomthief, line 3635: Comment explaining what this function does would be helpful

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -2799,6 +2997,712 @@ rename_procfunc_update_bbf_catalog(RenameStmt *stmt)
...
+				InvalidOid,
+				tsql_get_server_collation_oid_internal(false),
+				F_TEXTEQ,
+				CStringGetTextDatum(PERMISSIONS_FOR_ALL_OBJECTS_IN_SCHEMA));
+
+	scan = systable_beginscan(bbf_schema_rel, get_bbf_schema_perms_idx_oid(),
+							true, NULL, 3, scanKey);
+	tuple_bbf_schema = systable_getnext(scan);
+
+	while (HeapTupleIsValid(tuple_bbf_schema))
```
thephantomthief, line 3676: is it expected that we get more than one entry?

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -2799,6 +2997,712 @@ rename_procfunc_update_bbf_catalog(RenameStmt *stmt)
...
+				InvalidOid,
+				tsql_get_server_collation_oid_internal(false),
+				F_TEXTEQ,
+				CStringGetTextDatum(grantee));
+
+	scan = systable_beginscan(bbf_schema_rel, get_bbf_schema_perms_idx_oid(),
+							true, NULL, 3, scanKey);
+	tuple_bbf_schema = systable_getnext(scan);
+
+	while (HeapTupleIsValid(tuple_bbf_schema))
```
thephantomthief, line 3565: is it expected that we get more than one entry?

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -3846,16 +3866,187 @@ bbf_ProcessUtility(PlannedStmt *pstmt,
...
+						Node *func = (Node *) lsecond(ob->objname);
+						logicalschema = get_logical_schema_name(schemaname, true);
+						funcname = strVal(func);
+					}
+
+					/* If ALL PRIVILEGES is granted/revoked. */
+					if (list_length(grant->privileges) == 0)
+					{
+						/*
+						 * Case: When ALL PRIVILEGES is revoked internally during create function.
```
thephantomthief, line 4007: We should update this comment

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -2799,6 +2997,712 @@ rename_procfunc_update_bbf_catalog(RenameStmt *stmt)
...
+			const char	*query = NULL;
+			char			*schema;
+			List			*res;
+			GrantStmt		*grant;
+			PlannedStmt		*wrapper;
+
+			schema = get_physical_schema_name((char *)db_name, logicalschema);
+
+			if (strcmp(object_type, OBJ_FUNCTION) == 0)
+				query = psprintf("GRANT ALL ON FUNCTION [%s].[%s] TO %s", schema, object_name, grantee);
```
thephantomthief: Better to make this GRANT EXECUTE ...

test/JDBC/expected/1_grant_schema.out
```diff
@@ -0,0 +1,2936 @@
...
+1
+~~END~~
+
+
+-- tsql
+revoke select on dbo.t1 from guest;
+go
+revoke execute on dbo.f1 from guest;
+go
+revoke select, execute on schema::dbo to guest;
```
thephantomthief: does it make sense to check if we still have access objects after object level revoke? If we already have tests like for this then we can ignore this

test/JDBC/expected/1_grant_schema.out
```diff
@@ -0,0 +1,2936 @@
...
+
+~~ERROR (Message: permission denied for procedure p1)~~
+
+select * from dbo.f1()
+go
+~~ERROR (Code: 33557097)~~
+
+~~ERROR (Message: permission denied for function f1)~~
+
+select * from dbo.f2() -- Needs fix, should not be accessible
```
thephantomthief: Looks like it is fixed 😄 We can remove this comment then

test/JDBC/expected/1_grant_schema.out
```diff
@@ -0,0 +1,2936 @@
...
+insert successful
+~~END~~
+
+~~ROW COUNT: 1~~
+
+insert into s1.t1 (a) values (next value for s1.sq1);
+go
+~~ERROR (Code: 33557097)~~
+
+~~ERROR (Message: permission denied for sequence sq1)~~
```
thephantomthief: Few questions:

1. This is expected right? I assume so because u1 does not have the necessary permission for sq1.
2. Do we have any way to grant permission of sq1 to u1 in BBF? Is it there in T-SQL? If so we should document if not done already.
3. In T-SQL does GRANT \<some permission\> ON SCHEMA, GRANT on sequence as well? If so and if we are unable to support this at the moment in BBF we should also document this if not already done.

test/JDBC/expected/1_grant_schema.out
```diff
@@ -0,0 +1,2936 @@
...
+-- create new objects in the schema s1
+create table s1.t2(a int);
+go
+create view s1.v2 as select 1;
+go
+create procedure s1.p2 as select 1;
+go
+create function s1.f2() returns int begin declare @a int; set @a = 1; return @a; end
+go
+
```
thephantomthief: Might be good to also create a trigger and sequence and test if they also have the correct permissions

test/JDBC/expected/1_grant_schema.out
```diff
@@ -0,0 +1,2936 @@
...
+go
+grant insert on schema::s1 to u1;
+go
+grant delete on schema::s1 to u1;
+go
+grant update on schema::s1 to u1;
+go
+
+-- tsql user=l1 password=123
+-- Test Group 3 [user u1 has select, insert, delete privilege on the schema]
```
thephantomthief: user u1 has select, insert, update and delete privilege on the schema

contrib/babelfishpg_tsql/src/catalog.c
shalinilohia50: Yes, there can be multiple rows for multiple grantees.

contrib/babelfishpg_tsql/src/catalog.c
shalinilohia50: Done.

test/JDBC/expected/1_grant_schema.out
```diff
@@ -0,0 +1,2936 @@
...
+go
+create view s1.v2 as select 1;
+go
+create procedure s1.p2 as select 1;
+go
+create function s1.f2() returns int begin declare @a int; set @a = 1; return @a; end
+go
+
+-- tsql user=l1 password=123
+-- Test Group 5[user u1 has select and execute privilege on the schema]
```
thephantomthief: user u1 has select, insert, update, delete and execute privilege on the schema

test/JDBC/expected/1_grant_schema.out
```diff
@@ -0,0 +1,2936 @@
...
+-- Test Group 6 
+-- [user u1 has execute privilege on the schema, u1 has select privilege on s1.t1 via trigger]
+select current_user;
+go
+~~START~~
+varchar
+u1
+~~END~~
+
+select * from s1.t1;
```
thephantomthief: This is expected because trigger tr1 was triggered when we did insert previously in s1.t1 which explicitly granted u1 select permission to s1.t1, right?

test/JDBC/expected/1_grant_schema.out
```diff
@@ -0,0 +1,2936 @@
...
+go
+grant select on s1.v1 to u1;
+go
+grant execute on s1.p1 to u1;
+go
+grant execute on s1.f1 to u1;
+go
+
+-- tsql user=l1 password=123
+-- Test Group 8 [user u1 has select, insert, execute privilege on these objects]
```
thephantomthief: user u1 has select, insert, delete, execute privilege on these objects

test/JDBC/expected/1_grant_schema.out
```diff
@@ -0,0 +1,2936 @@
...
+~~END~~
+
+~~ROW COUNT: 1~~
+
+insert into s1.t1 (a) values (next value for s1.sq1);
+go
+~~ERROR (Code: 33557097)~~
+
+~~ERROR (Message: permission denied for sequence sq1)~~
+
```
thephantomthief: let's check if update on s1.t1 is denied to u1

test/JDBC/expected/1_grant_schema.out
```diff
@@ -0,0 +1,2936 @@
...
+int
+~~END~~
+
+insert into s1.t1 values (1);
+go
+~~START~~
+varchar
+insert successful
+~~END~~
+
```
thephantomthief: should add test where we are now inserting using sequence, and where updates are denied for Group 9 to 17

test/JDBC/expected/1_grant_schema.out
```diff
@@ -0,0 +1,2936 @@
...
+~~ROW COUNT: 1~~
+
```
thephantomthief: we should check for update as well

test/JDBC/expected/1_grant_schema.out
```diff
@@ -0,0 +1,2936 @@
...
+-- tsql
+grant select, insert, delete, execute on schema::s1 to public;
+go
+grant select, insert, delete, execute on schema::s1 to u1;
+go
+
+-- tsql user=l1 password=123
+-- Test Group 12
+-- [select, insert, delete, execute privilege on these objects to u1]
+-- [select, insert, delete, execute privilege on schema to public]
```
thephantomthief: select, insert, delete, execute privilege on schema to public, u1

test/JDBC/expected/1_grant_schema.out
```diff
@@ -0,0 +1,2936 @@
...
+
+
+-- tsql
+revoke select, insert, delete, execute on schema::s1 from u1;
+go
+
+-- tsql user=l1 password=123
+-- Test Group 13
+-- [select, insert, delete, execute privilege on these objects to u1]
+-- [select, insert, delete, execute privilege on schema to public, u1]
```
thephantomthief: select, insert, delete, execute privilege on schema to public.

We have revoked on schema s1 from u1 above

test/JDBC/expected/1_grant_schema.out
```diff
@@ -0,0 +1,2936 @@
...
+revoke select on s1.v1 from u1;
+go
+revoke execute on s1.p1 from u1;
+go
+revoke execute on s1.f1 from u1;
+go
+
+-- tsql user=l1 password=123
+-- Test Group 14
+-- [select, insert, delete, execute privilege on schema to public, u1]
```
thephantomthief: select, insert, delete, execute privilege on schema to public

test/JDBC/expected/1_grant_schema.out
```diff
@@ -0,0 +1,2936 @@
...
+~~END~~
+
+
+-- tsql
+revoke select, insert, delete, execute on schema::s1 from u1;
+go
+
+-- tsql user=l1 password=123
+-- Test Group 15
+-- [select, insert, delete, execute privilege on schema to public]
```
thephantomthief: I think Group 14 and 15 are identical

test/JDBC/expected/1_grant_schema.out
shalinilohia50: We already have added tests for this in the section below after revoking object level access.
`-- guest user should have no permission`

test/JDBC/expected/1_grant_schema.out
shalinilohia50: 1. yes, this is expected.
2. Yes, we do have update privilege on sequences which we don't support right now in BBF.
3. [BABEL-4821] will support it later.

test/JDBC/expected/1_grant_schema.out
shalinilohia50: Yes, we have tests for triggers and sequences.
Triggers work with the same permission of object and we don't support permission on sequences yet [BABEL-4821].

test/JDBC/expected/1_grant_schema.out
shalinilohia50: Right! let me fix it.

test/JDBC/expected/1_grant_schema.out
shalinilohia50: Yes. This was added to show how GRANT would work within a trigger.

test/JDBC/expected/1_grant_schema.out
shalinilohia50: Yes, removed one.

shalinilohia50:
https://github.com/babelfish-for-postgresql/postgresql_modified_for_babelfish/pull/327

shalinilohia50:
[](https://github.com/babelfish-for-postgresql/postgresql_modified_for_babelfish/pull/327)

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -2803,6 +3001,717 @@ rename_procfunc_update_bbf_catalog(RenameStmt *stmt)
...
+				const char *object_type,
+				const char *func_args)
+{
+	Relation	bbf_schema_rel;
+	TupleDesc	bbf_schema_dsc;
+	HeapTuple	tuple_bbf_schema;
+	Datum		new_record_bbf_schema[BBF_SCHEMA_PERMS_NUM_OF_COLS];
+	bool		new_record_nulls_bbf_schema[BBF_SCHEMA_PERMS_NUM_OF_COLS];
+	int16	dbid = get_cur_db_id();
+	char	*user = GetUserNameFromId(GetUserId(), true);
```
thephantomthief: noerr should be false IMO

contrib/babelfishpg_tsql/src/catalog.c
shalinilohia50: Yes, there can be multiple rows in case of grant to multiple users.

test/JDBC/expected/1_grant_schema.out
```diff
@@ -0,0 +1,3090 @@
...
+go
+revoke execute on schema::s2 to u3;
+go
+revoke select on schema::s1 to u3;
+go
+revoke execute on schema::s1 to u3;
+go
+
+-- tsql user=l3 password=123
+-- u3 has select and execute privilege
```
thephantomthief: u3 has no privilege on s1 and s2

test/JDBC/expected/1_grant_schema.out
```diff
@@ -0,0 +1,3090 @@
...
+
+~~ERROR (Message: The server principal "l1" is not able to access the database "db1" under the current security context)~~
+
+exec db1.dbo.p1;
+go
+~~ERROR (Code: 33557097)~~
+
+~~ERROR (Message: The server principal "l1" is not able to access the database "db1" under the current security context)~~
+
+select * from db1.dbo.f1();
```
thephantomthief: Is it expected to get a different error here?

shalinilohia50: BABEL-4842 is created to track this.

thephantomthief (approved):

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -2090,6 +2140,24 @@ get_function_name(HeapTuple tuple, TupleDesc dsc)
 	return NameGetDatum(&(func->funcname));
 }
 
+static Datum
+get_perms_schema_name(HeapTuple tuple, TupleDesc dsc)
+{
+	bool		isNull;
+	Datum		schema_name = heap_getattr(tuple, Anum_bbf_schema_perms_schema_name, dsc, &isNull);
```
kuntalghosh, line 2149: We should check the null value. IIUC, it shouldn't be null. We can throw an error if it is.

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -2090,6 +2140,24 @@ get_function_name(HeapTuple tuple, TupleDesc dsc)
...
+
+	return schema_name;
+}
+
+static Datum
+get_perms_grantee_name(HeapTuple tuple, TupleDesc dsc)
+{
+	bool		isNull;
+	Datum		grantee_name = heap_getattr(tuple, Anum_bbf_schema_perms_grantee, dsc, &isNull);
```
kuntalghosh, line 2162: We should check the null value. IIUC, it shouldn't be null. We can throw an error if it is.

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -2621,6 +2701,124 @@ rename_update_bbf_catalog(RenameStmt *stmt)
...
+	{
+		if (rename_type == OBJECT_PROCEDURE)
+			object_type = OBJ_PROCEDURE;
+		else if (rename_type == OBJECT_FUNCTION)
+			object_type = OBJ_FUNCTION;
+		objwargs = (ObjectWithArgs *) stmt->object;
+		DeconstructQualifiedName(objwargs->objname, &physical_schema_name, &object_name);
+		logical_schema_name = (char *) get_logical_schema_name(physical_schema_name, true);
+	}
+
```
kuntalghosh, line 2760: As we're sending missing_ok = true to get_logical_schema_name, it means that we're expecting logical schema name is NULL for some scenarios. Have we added a test case for the same? Could you please explain when it can be NULL?

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -2803,6 +3001,717 @@ rename_procfunc_update_bbf_catalog(RenameStmt *stmt)
...
+									new_record_nulls_bbf_schema);
+
+	/* Insert new record in the bbf_authid_user_ext table */
+	CatalogTupleInsert(bbf_schema_rel, tuple_bbf_schema);
+
+	/* Close bbf_authid_user_ext, but keep lock till commit */
+	table_close(bbf_schema_rel, RowExclusiveLock);
+
+	/* Advance cmd counter to make the insert visible */
+	CommandCounterIncrement();
```
kuntalghosh, line 3068: Where are we reading this value in the same command?

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -2803,6 +3001,717 @@ rename_procfunc_update_bbf_catalog(RenameStmt *stmt)
...
+				tsql_get_server_collation_oid_internal(false),
+				F_TEXTEQ,
+				CStringGetTextDatum(grantee));
+
+	scan = systable_beginscan(bbf_schema_rel,
+				get_bbf_schema_perms_idx_oid(),
+				false, NULL, 5, scanKey);
+
+	tuple_bbf_schema = systable_getnext(scan);
+	if (HeapTupleIsValid(tuple_bbf_schema))
```
kuntalghosh, line 3165: So, we're expecting a single tuple with this combination. And, that means, none of the input parameter should be NULL. Either we should validate the input parameter (being not null) or we should put a check that there's only one tuple.

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -2803,6 +3001,717 @@ rename_procfunc_update_bbf_catalog(RenameStmt *stmt)
...
+					F_TEXTEQ,
+					CStringGetTextDatum(object_name));
+
+		scan = systable_beginscan(bbf_schema_rel,
+					get_bbf_schema_perms_idx_oid(),
+					true, NULL, 3, scanKey);
+	}
+
+	tuple_bbf_schema = systable_getnext(scan);
+	if (HeapTupleIsValid(tuple_bbf_schema))
```
kuntalghosh, line 3279: Shouldn't we throw an error if missing_ok is false? Otherwise, the argument name is correct. Should be ignore_grantee or something? In that case, we probably should just send grantee as NULL.

contrib/babelfishpg_tsql/sql/ownership.sql
```diff
@@ -14,6 +14,21 @@ CREATE TABLE sys.babelfish_sysdatabases (
...
 GRANT SELECT on sys.babelfish_sysdatabases TO PUBLIC;
 
+-- BABELFISH_SCHEMA_PERMISSIONS
+-- This catalog is implemented specially to support GRANT/REVOKE .. ON SCHEMA ..
+-- Please avoid using this catalog anywhere else.
+CREATE TABLE sys.babelfish_schema_permissions (
+  dbid smallint NOT NULL,
+  schema_name sys.NVARCHAR(128) NOT NULL COLLATE sys.database_default,
+  object_name sys.NVARCHAR(128) NOT NULL COLLATE sys.database_default,
+  permission INT NOT NULL,
```
kuntalghosh: Should we add a constraint that permission should be > 0 always?

test/JDBC/upgrade/latest/schedule
```diff
@@ -511,3 +510,5 @@ BABEL-730
 babel-4475
 babel-3254
 babel-4517
+babelfish_inconsistent_metadata
```
rishabhtanwar29: What's the point of moving this test at the end?

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -2803,6 +3001,717 @@ rename_procfunc_update_bbf_catalog(RenameStmt *stmt)
...
+	while (HeapTupleIsValid(tuple_bbf_schema))
+	{
+		bool isnull;
+		Datum datum;
+		object_name = pstrdup(TextDatumGetCString(heap_getattr(tuple_bbf_schema, Anum_bbf_schema_perms_object_name, dsc, &isnull)));
+		object_type = pstrdup(TextDatumGetCString(heap_getattr(tuple_bbf_schema, Anum_bbf_schema_perms_object_type, dsc, &isnull)));
+		current_permission = DatumGetInt32(heap_getattr(tuple_bbf_schema, Anum_bbf_schema_perms_permission, dsc, &isnull));
+		datum = heap_getattr(tuple_bbf_schema, Anum_bbf_schema_perms_function_args, dsc, &isnull);
+		if (!isnull)
+			func_args = TextDatumGetCString(datum);
```
kuntalghosh, line 3578: 1. func_args is not freed, so it might be reused in the next iteration. 
2. no need to use pstrdup as we're using the value immediately in psprintf.
3. better to define these variable inside the loop to limit the scope

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -2803,6 +3001,717 @@ rename_procfunc_update_bbf_catalog(RenameStmt *stmt)
...
+		/* For each object, grant the permission explicitly. */
+		if (strcmp(object_name, PERMISSIONS_FOR_ALL_OBJECTS_IN_SCHEMA) != 0)
```
kuntalghosh, line 3580: We should just search with PERMISSIONS_FOR_ALL_OBJECTS_IN_SCHEMA as a search key for the object name. Because, we're not doing anything for the else part.

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -2803,6 +3001,717 @@ rename_procfunc_update_bbf_catalog(RenameStmt *stmt)
...
+	while (HeapTupleIsValid(tuple_bbf_schema))
+	{
+		bool isnull;
+		Datum datum;
+		current_permission = DatumGetInt32(heap_getattr(tuple_bbf_schema, Anum_bbf_schema_perms_permission, dsc, &isnull));
+		datum = heap_getattr(tuple_bbf_schema, Anum_bbf_schema_perms_grantee, dsc, &isnull);
+		if (!isnull)
+			grantee = TextDatumGetCString(datum);
+		/* For each object, grant the permission explicitly. */
+		if (current_permission & ALL_PERMISSIONS_ON_FUNCTION)
```
kuntalghosh, line 3685: search permission in the key

test/JDBC/upgrade/latest/schedule
```diff
@@ -511,3 +510,5 @@ BABEL-730
...
+1_grant_schema
```
rishabhtanwar29: 1. Why do we need separate test file? Can't we add new test cases into the existing GRANT_SCHEMA-vu* file? We should try to minimize the number of test files unless absolutely needed.
2. What's the use of GRANT_SCHEMA.sql test? Or what's the difference b/w GRANT_SCHEMA-vu* and GRANT_SCHEMA.sql?
3. Here we just have changed the latter casing for the new test file which will make it quite difficult for future developers to understand the difference b/w GRANT_SCHEMA-vu* and 1_grant_schema-vu* test files. On a high level, it looks like what we want is to have GRANT_SCHEMA-before-15_7-16_3-vu* and GRANT_SCHEMA-vu* instead.
4. Why have we missed 15_6 and 15_5 schedule files?
5. Let's also add tests for database-level dump restore for this new catalog table as we discussed offline.

test/JDBC/expected/1_GRANT_SCHEMA-vu-verify.out
```diff
@@ -0,0 +1,189 @@
...
+DO $$
+BEGIN 
+    IF EXISTS (SELECT 1 FROM
+information_schema.tables WHERE
+table_name = 'tbl_creation_should_succeed' AND
+table_schema = 'master_dbo') THEN
+        EXECUTE 'GRANT ALL ON
+master_dbo.tbl_creation_should_succeed TO master_dbo';
+    END IF;
+END $$;
```
HarshLunagariya, line 99: Why this?

test/JDBC/upgrade/latest/schedule
shalinilohia50: 1, 2 and 3: GRANT_SCHEMA.mix has all the various scenarios mentioned in the test plan. All of it can't be written in vu-* format. We have GRANT_SCHEMA-before-15_7-16_3-vu*  and 1_GRANT_SCHEMA-vu* for upgrade testing of permissions and catalog change. We have added [1_] because we wanted this test to run in the beginning since some permissions in other test files were interfering with these tests.
4. Added.
5. Added.

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -3687,6 +3688,135 @@ get_insert_bulk_kilobytes_per_batch()
...
+				if (stmt->privileges & ACL_SELECT)
+					grant_perms_to_objects_in_schema(stmt->schema_name, ACL_SELECT, rolname);
+				if (stmt->privileges & ACL_INSERT)
+					grant_perms_to_objects_in_schema(stmt->schema_name, ACL_INSERT, rolname);
+				if (stmt->privileges & ACL_UPDATE)
+					grant_perms_to_objects_in_schema(stmt->schema_name, ACL_UPDATE, rolname);
+				if (stmt->privileges & ACL_DELETE)
+					grant_perms_to_objects_in_schema(stmt->schema_name, ACL_DELETE, rolname);
+				if (stmt->privileges & ACL_REFERENCES)
+					grant_perms_to_objects_in_schema(stmt->schema_name, ACL_REFERENCES, rolname);
```
kuntalghosh:
```
int i;
for (i = 0; i < NUM_PERMISSIONS; i++)
{
	/* Execute the GRANT SCHEMA subcommands. */
	if (stmt->privileges & permissions[i])
		exec_grantschema_subcmds(create_schema->schemaname, rolspec->rolename, true, false, permissions[i]);
}
```

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -3561,7 +3571,17 @@ bbf_ProcessUtility(PlannedStmt *pstmt,
 								   NULL);
 
 					CommandCounterIncrement();
-
+					/* Grant ALL schema privileges to the user.*/
+					if (rolspec && strcmp(queryString, "(CREATE LOGICAL DATABASE )") != 0)
+					{
+						int permissions[6] = {ACL_INSERT, ACL_SELECT, ACL_UPDATE, ACL_DELETE, ACL_REFERENCES, ACL_EXECUTE};
```
kuntalghosh: As mentioned above, declare this globally

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -3851,16 +3872,191 @@ bbf_ProcessUtility(PlannedStmt *pstmt,
...
+				const char *current_user = GetUserNameFromId(GetUserId(), false);
+				/* Ignore when GRANT statement has no specific named object. */
+				if (sql_dialect != SQL_DIALECT_TSQL || grant->targtype != ACL_TARGET_OBJECT)
+					break;
+				Assert(list_length(grant->objects) == 1);
+				if (grant->objtype == OBJECT_SCHEMA)
+						break;
+				else if (grant->objtype == OBJECT_TABLE)
+				{
+					/* Ignore CREATE database subcommands */
```
kuntalghosh: For CREATE DATABASE commands, the objtype is OBJECT_TABLE? Why is it that?

test/JDBC/expected/GRANT_SCHEMA.out
```diff
@@ -0,0 +1,4555 @@
...
+~~ROW COUNT: 1~~
+
+select * from babel_4344_t3; -- not accessible, only column privilege is granted
+go
+~~ERROR (Code: 33557097)~~
+
+~~ERROR (Message: permission denied for table babel_4344_t3)~~
+
+select * from babel_4344_s1.babel_4344_t3 -- not accessible, only column privilege is granted
+go
```
HarshLunagariya, line 429: Have we added testcases for column level access validation? After GRANT/REVOKE sequences.

kuntalghosh (changes requested):

test/JDBC/expected/GRANT_SCHEMA.out
```diff
@@ -0,0 +1,4555 @@
...
+insert successful
+~~END~~
+
+~~ROW COUNT: 1~~
+
+insert into s1.t1 (a) values (next value for s1.sq1);
+go
+~~ERROR (Code: 33557097)~~
+
+~~ERROR (Message: permission denied for sequence sq1)~~
```
HarshLunagariya, line 1537: Is it expected?

contrib/babelfishpg_tsql/src/catalog.c
shalinilohia50: We cannot use permission in the search key. Permission can be a combination of SELECT and EXECUTE, we need to check if `EXECUTE` exists on not. So just checking for `ALL_PERMISSIONS_ON_FUNCTION` which is `EXECUTE` in the key wouldn't work.

contrib/babelfishpg_tsql/src/catalog.c
shalinilohia50: We cannot use PERMISSIONS_FOR_ALL_OBJECTS_IN_SCHEMA in the search key, because we need to know all the rows where permission exists at schema level or at any object belonging to that schema. We need to find out the list of such objects and grant the permission explicitly.

contrib/babelfishpg_tsql/src/pl_handler.c
shalinilohia50: Updated the comment:
```
Ignore GRANT statements that are executed implicitly as a part of CREATE database statements. 
Refer: create_bbf_db_internal(). These GRANT statement are just executed at the end, without 
checking any schema permission or adding catalog entry.

File: contrib/babelfishpg_tsql/src/catalog.c (hunk 28 above)
Comment (shalinilohia50) [reply contrib/babelfishpg_tsql/src/catalog.c:3165]: We have added some checks for schema_name or grantee_name being NULL. We throw error in such cases.

File: contrib/babelfishpg_tsql/src/catalog.c (hunk 32 above)
Comment (shalinilohia50) [reply contrib/babelfishpg_tsql/src/catalog.c:3578]: 1. We are using const char*, it doesn't need to be freed.
2. Done.
3. Done.

File: contrib/babelfishpg_tsql/src/catalog.c (hunk 29 above)
Comment (shalinilohia50) [reply contrib/babelfishpg_tsql/src/catalog.c:3279]: If missing_ok is true, it means we are checking if any entry exists in the catalog which belong to an object. We don't care about the grantee information in that case.
`ignore_grantee` makes sense.

File: contrib/babelfishpg_tsql/src/catalog.c (hunk 26 above)
Comment (shalinilohia50) [reply contrib/babelfishpg_tsql/src/catalog.c:2760]: schema should not be NULL, changed it to false.

File: test/JDBC/expected/GRANT_SCHEMA.out (hunk 41 above)
Comment (shalinilohia50) [reply test/JDBC/expected/GRANT_SCHEMA.out:1537]: Yes, currently we are not supporting any permissions on sequence via GRANT .. ON .. SCHEMA syntax.

File: test/JDBC/expected/1_GRANT_SCHEMA-vu-verify.out (hunk 36 above)
Comment (shalinilohia50) [reply test/JDBC/expected/1_GRANT_SCHEMA-vu-verify.out:99]: We create this table from PG endpoint in one of the tests. So when we try to do a REVOKE on schema master-> dbo from TSQL endpoint, it was throwing an error in MVU. `~ERROR (Message: permission denied for table tbl_creation_should_succeed)~~` 
To fix MVU, we have added this.

File: test/JDBC/expected/GRANT_SCHEMA.out (hunk 40 above)
Comment (shalinilohia50) [reply test/JDBC/expected/GRANT_SCHEMA.out:429]: Yes, we have tests.

**Review by rishabhtanwar29 — APPROVED** [review]
Looks good from dump/restore perspective

File: contrib/babelfishpg_tsql/src/catalog.c [hunk 42]
```diff
@@ -2803,6 +3010,714 @@ rename_procfunc_update_bbf_catalog(RenameStmt *stmt)
...
+	HeapTuple	tuple_bbf_schema;
+	SysScanDesc	scan;
+	bool	catalog_entry_exists = false;
+	int16	dbid = get_cur_db_id();
+
+	/* Immediately return false, if SCHEMA name is NULL or it's a shared schema. */
+	if (schema_name == NULL || is_shared_schema(schema_name))
+		return false;
+
+	if (!ignore_grantee)
```
kuntalghosh: Doubt: When is this called with a valid grantee list, but ignore_grantee = true? Trying to see whether we even need this boolean variable.

contrib/babelfishpg_tsql/src/catalog.c
kuntalghosh: Right, I missed that this is not a equality test.

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -3687,6 +3688,122 @@ get_insert_bulk_kilobytes_per_batch()
...
+		schemaOid = LookupExplicitNamespace(schema_name, true);
+	}
+	else
+	{
+		ereport(ERROR,
+					(errcode(ERRCODE_UNDEFINED_SCHEMA),
+					 errmsg("An object or column name is missing or empty. For SELECT INTO statements, verify each column has a name. For other statements, look for empty alias names. Aliases defined as \"\" or [] are not allowed. Change the alias to a valid name.")));
+	}
+
+	if (!OidIsValid(schemaOid))
```
kuntalghosh: Instead of this check, you should send missing_ok = false to LookupExplicitNamespace

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -3851,16 +3871,196 @@ bbf_ProcessUtility(PlannedStmt *pstmt,
...
+						else
+							logical_schema = get_authid_user_ext_schema_name(dbname, current_user);
+
+						/* If ALL PRIVILEGES is granted/revoked. */
+						if (list_length(grant->privileges) == 0)
+						{
+							foreach(lc, grant->grantees)
+							{
+								RoleSpec	   *rol_spec = (RoleSpec *) lfirst(lc);
+								if (grant->is_grant)
```
kuntalghosh: As discussed, please move the if statement outside the loop

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -3851,16 +3871,196 @@ bbf_ProcessUtility(PlannedStmt *pstmt,
...
+						foreach(lc1, grant->privileges)
+						{
+							AccessPriv *ap = (AccessPriv *) lfirst(lc1);
+							AclMode privilege = string_to_privilege(ap->priv_name);
+							foreach(lc, grant->grantees)
+							{
+								RoleSpec	   *rol_spec = (RoleSpec *) lfirst(lc);
+								if (grant->is_grant)
+								{
+									call_prev_ProcessUtility(pstmt, queryString, readOnlyTree, context, params, queryEnv, dest, qc);
```
kuntalghosh: Please move this call outside the main loop and is_grant check outside the inner loop. Preferably, introduce a single variable, something like, update_pg_catalog and use the same at end.

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -3851,16 +3871,196 @@ bbf_ProcessUtility(PlannedStmt *pstmt,
...
+						{
+							call_prev_ProcessUtility(pstmt, queryString, readOnlyTree, context, params, queryEnv, dest, qc);
+							exec_internal_grant_on_function(logicalschema, funcname, obj_type);
+							return;
+						}
+						
+						foreach(lc, grant->grantees)
+						{
+							RoleSpec	   *rol_spec = (RoleSpec *) lfirst(lc);
+							if (grant->is_grant)
```
kuntalghosh: same here.

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -3851,16 +3871,196 @@ bbf_ProcessUtility(PlannedStmt *pstmt,
...
+							{
+								call_prev_ProcessUtility(pstmt, queryString, readOnlyTree, context, params, queryEnv, dest, qc);
+								/*
+								 * If it is an implicit GRANT issued by exec_internal_grant_on_function, then we should not add catalog
+								 * entry. Catalog entry is supposed to be added only by explicit GRANTs.
+								 */
+								if (strcmp("(GRANT STATEMENT )", queryString) != 0)
```
kuntalghosh: We should move this outside the loop as well. if (strcmp("(GRANT STATEMENT )", queryString) == 0), is_grant = true and in that case, we're not doing anything. 

Do we need to consider the case for ALL PRIVILEGE case as well?

contrib/babelfishpg_tsql/src/catalog.c
shalinilohia50: No, there won't be any such case. 
ignore_grantee = true only where grantee is NULL and we have to avoid look up on grantee since it is NULL and grantee_name is a NOT NULL value in the catalog.

contrib/babelfishpg_tsql/src/pl_handler.c
shalinilohia50: if (strcmp("(GRANT STATEMENT )", queryString) == 0), is_grant = true  -> We are executing the pg command by process utility but not updating the catalog.

`Do we need to consider the case for ALL PRIVILEGE case as well?` No we don't have to. There is no such case.

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -3063,6 +3063,9 @@ add_entry_to_bbf_schema_perms(const char *schema_name,
 
 	/* Close bbf_authid_user_ext, but keep lock till commit */
 	table_close(bbf_schema_rel, RowExclusiveLock);
+
+	/* make sure later steps can see the entry added here */
+	CommandCounterIncrement();
```
shalinilohia50, line 3068: For commands like `GRANT SELECT, INSERT on t1 to u1`, SELECT privilege will add an entry to the catalog and the second iteration of INSERT privilege should be able to see the first entry in the catalog so that it updates the permission column.

contrib/babelfishpg_tsql/src/catalog.c
kuntalghosh: Then, why do we need this extra boolean argument? We should just check whether grantee == NULL.

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -3851,16 +3871,215 @@ bbf_ProcessUtility(PlannedStmt *pstmt,
...
+						AccessPriv *ap = (AccessPriv *) lfirst(lc1);
+						AclMode privilege = string_to_privilege(ap->priv_name);
+						if (grant->is_grant)
+						{
+							exec_pg_command = true;
+							foreach(lc, grant->grantees)
+							{
+								RoleSpec	   *rol_spec = (RoleSpec *) lfirst(lc);
+								/* Don't add/update an entry, if the permission is granted on column list.*/
+								if (ap->cols == NULL)
```
kuntalghosh: This is a loop-invariant. Skip the loop if ap->cols != NULL. I thought you'll take care of all such occurrences.

contrib/babelfishpg_tsql/src/pl_handler.c
kuntalghosh: Again, this is a loop-invariant. Skip the loop if strcmp("(GRANT STATEMENT )", queryString)  == 0.

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -3851,16 +3871,215 @@ bbf_ProcessUtility(PlannedStmt *pstmt,
...
+						}
+						else
+						{
+							foreach(lc, grant->grantees)
+							{
+								RoleSpec	   *rol_spec = (RoleSpec *) lfirst(lc);
+								/*
+								 * If permission on schema exists, don't revoke any permission from the object.
+								 */
+								if (!privilege_exists_in_bbf_schema_permissions(logical_schema, PERMISSIONS_FOR_ALL_OBJECTS_IN_SCHEMA, rol_spec->rolename, false))
```
kuntalghosh: Skip further checks if exec_pg_command is already set to true by previous iterations. Please make this change for all such occurrences in the following code.

```
if (!exec_pg_command || !privilege_exists_in_bbf_schema_permissions(logical_schema, PERMISSIONS_FOR_ALL_OBJECTS_IN_SCHEMA, rol_spec->rolename, false))
	exec_pg_command = true;
```

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -3851,16 +3871,215 @@ bbf_ProcessUtility(PlannedStmt *pstmt,
...
+						else
+						{
+							foreach(lc, grant->grantees)
+							{
+								RoleSpec	   *rol_spec = (RoleSpec *) lfirst(lc);
+
+								/*
+								 * If permission on schema exists, don't revoke any permission from the object.
+								 */
+								if (!privilege_exists_in_bbf_schema_permissions(logicalschema, PERMISSIONS_FOR_ALL_OBJECTS_IN_SCHEMA, rol_spec->rolename, false))
```
kuntalghosh:
```
if (!exec_pg_command || !privilege_exists_in_bbf_schema_permissions(logical_schema, PERMISSIONS_FOR_ALL_OBJECTS_IN_SCHEMA, rol_spec->rolename, false))
	exec_pg_command = true;
```

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -3851,16 +3871,215 @@ bbf_ProcessUtility(PlannedStmt *pstmt,
...
+					/* If ALL PRIVILEGES is granted/revoked. */
+					if (list_length(grant->privileges) == 0)
+					{
+						/*
+						 * Case: When ALL PRIVILEGES is revoked internally during create function.
+						 * pstmt->stmt_len = 0 means it is an implicit REVOKE statement issued at the time of create function/procedure.
+						 * For more details, please refer revoke_func_permission_from_public().
+						 * If schema entry exists in the catalog, implicitly grant permission on the new object to the user.
+						 */
+						if ((pstmt->stmt_len == 0) && privilege_exists_in_bbf_schema_permissions(logicalschema, PERMISSIONS_FOR_ALL_OBJECTS_IN_SCHEMA, NULL, true))
```
kuntalghosh: Can there be a case when (pstmt->stmt_len == 0) but privilege_exists_in_bbf_schema_permissions(logicalschema, PERMISSIONS_FOR_ALL_OBJECTS_IN_SCHEMA, NULL, true) = false?

kuntalghosh (changes requested):

contrib/babelfishpg_tsql/src/pl_handler.c
shalinilohia50: Yes, it happens for a case when we create a function in a schema where there is no privilege granted by user.

contrib/babelfishpg_tsql/src/pl_handler.c
shalinilohia50: Yes, it should be &&.
```
if (!exec_pg_command && !privilege_exists_in_bbf_schema_permissions(logical_schema, PERMISSIONS_FOR_ALL_OBJECTS_IN_SCHEMA, rol_spec->rolename, false))
	exec_pg_command = true;
```

contrib/babelfishpg_tsql/src/pl_handler.c
shalinilohia50: Yes, it should be &&.
```
if (!exec_pg_command && !privilege_exists_in_bbf_schema_permissions(logical_schema, PERMISSIONS_FOR_ALL_OBJECTS_IN_SCHEMA, rol_spec->rolename, false))
	exec_pg_command = true;
```

kuntalghosh (approved):

PR #2430, opened 2024-03-14
Fix issues with TVFs which use identifiers longer than 63 characters

Sairakan:
### Description

In Table-valued functions which return data into a table variable, we create a temporary table in which to return the data. The name of this temporary table is generated as a combination of the table variable and function name. If this combination exceeds 63 characters (the limit of identifiers in postgresql), then we can run into issues trying to refer back to the temporary table to insert the data while executing the function, as the name used to access it will be different than what is stored in the postgres catalog due to truncation.

Fix this by utilizing the pre-existing `downcase_truncate_identifier()` function to ensure that the new identifier is within the postgresql character limit. Additionally, truncate the temporary table name (which normally tries to append the @@nest_level to the end) to 63 characters to prevent issues there.

### Issues Resolved

BABEL-3311

* **Use case based -**
Nested TVFs which use the same (long) table variable name

* **Boundary conditions -**
function name >63 characters
table variable name >63 characters
both function name and table variable name >63 characters
both function name and table variable name <63 characters, but combined are >63 characters

* **Arbitrary inputs -**
N/A

* **Negative test cases -**
N/A

N/A

N/A

* **Performance tests -**
N/A

* **Tooling impact -**
N/A

* **Client tests -**
N/A

### Check List


contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -1366,6 +1366,8 @@ exec_stmt_decl_table(PLtsql_execstate *estate, PLtsql_stmt_decl_table *stmt)
 		}
 
 		tblname = psprintf("%s_%d", var->refname, estate->nestlevel);
+		if (strlen(tblname) > 63)
```
timchang514: is it possible for collision here with multiple table variables longer than 63 chars?

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -2345,13 +2345,8 @@ bbf_table_var_lookup(const char *relname, Oid relnamespace)
 			continue;
 
 		tbl = (PLtsql_tbl *) estate->datums[n];
-		if (strcmp(relname, tbl->refname) == 0)
+		if (strcmp(relname, tbl->refname) == 0 && tbl->tblname)
```
timchang514, line 2357: is this necessary for the change?

Sairakan: Yeah, without this change we get an error when trying to use TVFs that have table names which are too long. I think that the FIXME has already been fixed in the ANTLR parser - I didn't see any crashes or other unexpected regressions in the full JDBC test suite after making this change

contrib/babelfishpg_tsql/src/pl_exec-2.c
Sairakan: Hmm, it might be possible. Let me test it out with multiple calls within a single transaction and see what happens.

Sairakan: Confirmed that there is a crash when nesting TVFs which use the same table name. Will fix by keeping the nest level during the truncation.

Sairakan: Also added a test for the nested function scenario.

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -1366,6 +1366,15 @@ exec_stmt_decl_table(PLtsql_execstate *estate, PLtsql_stmt_decl_table *stmt)
 		}
 
 		tblname = psprintf("%s_%d", var->refname, estate->nestlevel);
+
+		if (strlen(tblname) > 63)
+		{
+			// truncate tblname to fit the "_#" nestlevel suffix
+			tblname[63-(strlen(tblname)-63)] = '\0';
```
timchang514: instead of using 63, can express it as NAMEDATALEN (which is 64)?

also an edge case, but if estate->nestlevel is double digits, we will still be over 64 chars

contrib/babelfishpg_tsql/src/pl_exec-2.c
timchang514: let's use NAMEDATALEN here as well. it seems like the other places in the code checking for identifier length do a comparison 

`if (len >= NAMEDATALEN)` so we should match that behavior

contrib/babelfishpg_tsql/src/pl_exec-2.c
Sairakan: I'll fix this to use NAMEDATALEN.

As for the edge case, this already accounts for the possibility of having a larger nest level - I am just subtracting the difference in length between the refname and tblname, which is exactly how long the nestlevel suffix is, whether it's one, two, or more digits.

timchang514 (approved):

test/JDBC/expected/babel_table_type.out
```diff
@@ -649,7 +649,114 @@ test_name#!#1
 drop function mstvf_constraints;
 GO
 
+
+--BABEL-3311 TVF name truncation
```
2jungkook: Please use upgrade test (latest --> latest) format so that it can be covered in the dump/restore and future mVU/MVU tests, i.e., create functions in the prepare script.

Sairakan: Copied the tests into a set of new upgrade scripts

2jungkook (approved):

Issue #2431, opened 2024-03-15
[Bug]: Crash when alias is used without AS (only debug builds)


staticlibs:
### What happened?

The following query crashes the server if Babelfish is built with `-O0`:

```
select 1 a
```

It seems that a check that `std::string` is not empty is missed before calling `.front()` in [PLtsql_expr_query_mutator::add](https://github.com/babelfish-for-postgresql/babelfish_extensions/blob/b38e6e2c0261725627d9ac751721e50501af6eef/contrib/babelfishpg_tsql/src/tsqlIface.cpp#L632) (called with empty `orig_text` [from here](https://github.com/babelfish-for-postgresql/babelfish_extensions/blob/b38e6e2c0261725627d9ac751721e50501af6eef/contrib/babelfishpg_tsql/src/tsqlIface.cpp#L3060)). Will file a PR shortly.

### Version

BABEL_3_X_DEV (Default)

### Extension

babelfishpg_tsql (Default)

### Which flavor of Linux are you using when you see the bug?

Fedora

### Relevant log output

```shell
Server log:

/builddir/build/BUILD/gcc-13.2.1-20231011/obj-x86_64-redhat-linux/x86_64-redhat-linux/libstdc++-v3/include/bits/basic_string.h:1305: std::__cxx11::basic_string<_CharT, _Traits, _Alloc>::reference std::__cxx11::basic_string<_CharT, _Traits, _Alloc>::front() [with _CharT = char; _Traits = std::char_traits<char>; _Alloc = std::allocator<char>; reference = char&]: Assertion '!empty()' failed.
2024-03-15 21:20:01.869 GMT [80890] LOG:  server process (PID 80978) was terminated by signal 6: Aborted
2024-03-15 21:20:01.869 GMT [80890] DETAIL:  Failed process was running: select 1 a
2024-03-15 21:20:01.870 GMT [80890] LOG:  terminating any other active server processes

Backtrace:

libc.so.6!__pthread_kill_implementation (Unknown Source:0)
libc.so.6!raise (Unknown Source:0)
libc.so.6!abort (Unknown Source:0)
libstdc++.so.6!std::__glibcxx_assert_fail(char const*, int, char const*, char const*) (Unknown Source:0)
libstdc++.so.6!std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >::front() (Unknown Source:0)
[...]

Issue #2404, opened 2024-03-02
Too strict handling of table_type parameter in sp_tables procedure


staticlibs:
### What happened?

With `sp_tables` procedure Babelfish supports [table_type](https://github.com/MicrosoftDocs/sql-docs/blob/a0b3ff05f9802a8f85aa3c6a08eeb36b75d5070b/docs/relational-databases/system-stored-procedures/sp-tables-transact-sql.md#-table_type---table_type) parameter, for example the following call will return both tables and view:

```sql
EXEC sp_tables NULL, NULL, NULL, '''VIEW'',''TABLE'''
```

But is appeared that `table_type` parameter handling in MSSQL is more relaxed. For example, the following query is sent by IBM Cognos Analytics application, it is likely be passed through [mssql-jdbc](https://github.com/microsoft/mssql-jdbc/blob/dc191dbec9d3f75541da0549c201c79350d72dd1/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDatabaseMetaData.java#L539) impl of [JDBC getTables](https://docs.oracle.com/en/java/javase/17/docs/api/java.sql/java/sql/DatabaseMetaData.html#getTables(java.lang.String,java.lang.String,java.lang.String,java.lang.String%5B%5D)) introspection method:

```sql
EXEC sp_tables NULL, NULL, NULL, '''''''SYSTEM TABLE'''',''''TABLE'''',''''VIEW'''''''
```

This "double-escaped" query returns empty results in Babelfish.

Besides this double-quoted example, MSSQL also accepts `table_type` with unbalanced quotes like this:

```sql
EXEC sp_tables NULL, NULL, NULL, '''TABLE'''''',''''VIEW'',''SYSTEM TABLE'''
```

Current Babelfish handling of `table_type` with `unnest(string_to_array(table_type, ','))` seems to be too strict, it is suggested to relax it to accepts strings with invalid quotation that are accepted by MSSQL.

### Version

BABEL_3_X_DEV (Default)

### Extension

babelfishpg_tsql (Default)

### Which flavor of Linux are you using when you see the bug?

Fedora

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct.

staticlibs:
PR was merged, closing the issue.

PR #2433, opened 2024-03-16
Support for testing DatabaseMetaData API in JDBC tests

staticlibs:
### Description

JDBC specification includes introspection API contained in [DatabaseMetaData](https://docs.oracle.com/en/java/javase/17/docs/api/java.sql/java/sql/DatabaseMetaData.html) class. These methods are widely used by Java applications like DB browser, visualization and BI tools. From the point of view of client apps, correctness of implementation of these methods is guaranteed by JDBC lib vendor (DB vendor in most cases) and it is more reliable to use these methods instead of trying to do introspection with `INFORMATION_SCHEMA` or DB-specific SQL.

The implementation of some of these methods is done by calling system stored procedures like `sp_tables` or `sp_columns`. While these procedures in Babelfish are already covered with direct SQL tests, it is suggested to add additional test coverage by running these Java methods as part of JDBC test suite.

These methods call system procedures with a particular set of arguments (partially hardcoded in implementation lib). And, in theory, these calls and arguments can be different in newer versions of implementation lib. Currently test harness uses [very old version of mssql-jdbc lib](https://github.com/babelfish-for-postgresql/babelfish_extensions/blob/b38e6e2c0261725627d9ac751721e50501af6eef/test/JDBC/pom.xml#L51). Its update is not trivial, because error output differs in different versions (thus a large number of output files need to be updated as well). But when the time will come to update the version (or to run JDBC tests with multiple versions) it may be beneficial to have `DatabaseMetaData` API coverage in place.

Besides that, the alternative JDBC implementation in jTDS lib also implements the subset of `DatabaseMetaData` API, proposed new tests can run on jTDS as well.

Proposed change adds support for `dbmeta` keyword in `.txt` tests to `batch_run` harness, it uses the following format:

```
dbmeta#!#<method_name>#!#<arg1>|<arg2>|...|<argN>
```

Example of a `getColumns(String catalog, String schema, String table, String column)` call:

```
dbmeta#!#getColumns#!#master|dbo|tab1|col1
```

The majority of these methods return results as `ResultSet`, so existing utilities are used to print these result sets to output file.

### Related Issues

#2137 jTDS driver support
#2404 `getTables` method, unexpected double quoting done by mssql-jdbc lib
#2431 `getSchemas` method, current crashes the server (only debug builds)
#997 `getColumns` method, fixed in Babelfish 3.4

Supports calling all `DatabaseMetaData` that are using non-trivial SQL in `mssql-jdbc` lib impl.

### Check List

### DIFF
--- a/test/JDBC/expected/database_metadata.out
+++ b/test/JDBC/expected/database_metadata.out
@@ -1,6 +1,7 @@
 
 create table dbmeta_tab1(col1 int primary key, col2 nvarchar(max) not null)
 create table dbmeta_tab2(fcol1 int not null, foreign key (fcol1) references dbmeta_tab1(col1))
+create view dbmeta_view1 as select * from dbmeta_tab1
 create function dbmeta_func1() returns int as begin return 1 end
 create function dbmeta_func2(@par1 int) returns table return select cast(@par1 as int) as col1, cast ('foo' as nvarchar(max)) as col2
 create index idx1 on dbmeta_tab1(col2)
@@ -32,11 +33,15 @@ dbmeta#!#getTables#!#master|dbo|dbmeta_%
 varchar#!#varchar#!#varchar#!#varchar#!#varchar
 master#!#dbo#!#dbmeta_tab1#!#TABLE#!#<NULL>
 master#!#dbo#!#dbmeta_tab2#!#TABLE#!#<NULL>
+master#!#dbo#!#dbmeta_view1#!#VIEW#!#<NULL>
 ~~END~~
 
 dbmeta#!#getTables#!#master|dbo|dbmeta_%|TABLE|VIEW
 ~~START~~
 varchar#!#varchar#!#varchar#!#varchar#!#varchar
+master#!#dbo#!#dbmeta_tab1#!#TABLE#!#<NULL>
+master#!#dbo#!#dbmeta_tab2#!#TABLE#!#<NULL>
+master#!#dbo#!#dbmeta_view1#!#VIEW#!#<NULL>
 ~~END~~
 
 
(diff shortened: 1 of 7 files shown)


test/JDBC/src/main/java/com/sqlsamples/JDBCMetadata.java
```diff
@@ -0,0 +1,327 @@
...
+            throw new SQLException("Invalid number of parameters for 'getProcedures'," +
+                    " expected: '#!#getProcedures#!#catalog|schema[|procedure]'");
+        }
+        String catalog = parts[0];
+        String schema = parts[1];
+        String procedure = null;
+        if (parts.length == 3) {
+            procedure = parts[2];
+        }
+        ResultSet rs = dbmeta.getFunctions(catalog, schema, procedure);
```
thephantomthief: Shouldn't this be `getProcedures()`?

test/JDBC/expected/database_metadata.out
```diff
@@ -0,0 +1,276 @@
+
+create table dbmeta_tab1(col1 int primary key, col2 nvarchar(max) not null)
```
thephantomthief, line 2: We should create a view as well to see APIs like getTables() show views as well

test/JDBC/expected/database_metadata.out
```diff
@@ -0,0 +1,276 @@
...
+
+
+dbmeta#!#getTables#!#master|dbo|dbmeta_%
+~~START~~
+varchar#!#varchar#!#varchar#!#varchar#!#varchar
+master#!#dbo#!#dbmeta_tab1#!#TABLE#!#<NULL>
+master#!#dbo#!#dbmeta_tab2#!#TABLE#!#<NULL>
+~~END~~
+
+dbmeta#!#getTables#!#master|dbo|dbmeta_%|TABLE|VIEW
```
thephantomthief, line 39: is it expected we get no rows?

thephantomthief (commented):
We should add another section under https://github.com/babelfish-for-postgresql/babelfish_extensions/blob/BABEL_4_X_DEV/test/JDBC/README.md#writing-the-test-cases which explains how to add tests for DatabaseMetaData API.

test/JDBC/src/main/java/com/sqlsamples/JDBCMetadata.java
staticlibs: Yes, while output results are the same, the method name was wrong, fixed it.

test/JDBC/expected/database_metadata.out
staticlibs: Added a view.

test/JDBC/expected/database_metadata.out
staticlibs: Empty result set was caused by #2404 , it is fixed in latest 4x. Rebased the branch to latest and added corresponding output.

staticlibs:
Added a section to README. Also moved the test file out of the directory as no mo files are planned for that dir now.

thephantomthief (approved):

shardgupta (approved):

PR #2443, opened 2024-03-25
Add temp OID support for type, index, and constraint creation

timchang514:
### Description

Normal OIDs were still being consumed in some cases during temp table creation even with temp oid buffer on. To fix this, ensure that the appropriate temp table dependent objects also use temp OID. 

1. Redirect type creation functions to use temp OID when it's a TSQL temp table.
2. Redirect constraint creation to use temp OID in those cases as well. 
3. Remove workaround for indexes on temp tables. Previously, we needed to force indexes on temp tables to consume normal OIDs. With changes described in lines 1+2 above, we can now remove the workaround added in #1894  and properly allow indexes to use temp oid buffer. 

Engine PR: babelfish-for-postgresql/postgresql_modified_for_babelfish#335

### Issues Resolved

BABEL-4750

### Check List

### DIFF
--- a/contrib/babelfishpg_tsql/src/guc.c
+++ b/contrib/babelfishpg_tsql/src/guc.c
@@ -1163,7 +1163,7 @@ define_custom_variables(void)
 							gettext_noop("Temp oid buffer size"),
 							NULL,
 							&temp_oid_buffer_size,
-							65536, 0, 131072,
+							0, 0, 131072,
 							PGC_SUSET,
 							GUC_NOT_IN_SAMPLE | GUC_DISALLOW_IN_FILE | GUC_DISALLOW_IN_AUTO_FILE,
 							NULL, NULL, NULL);
--- a/contrib/babelfishpg_tsql/src/guc.c
+++ b/contrib/babelfishpg_tsql/src/guc.c
@@ -1163,7 +1163,7 @@ define_custom_variables(void)
 							gettext_noop("Temp oid buffer size"),
 							NULL,
 							&temp_oid_buffer_size,
-							0, 0, 131072,
+							65536, 0, 131072,
 							PGC_SUSET,
 							GUC_NOT_IN_SAMPLE | GUC_DISALLOW_IN_FILE | GUC_DISALLOW_IN_AUTO_FILE,
 							NULL, NULL, NULL);
--- a/test/JDBC/expected/BABEL-4752.out
+++ b/test/JDBC/expected/BABEL-4752.out
@@ -70,6 +70,38 @@ DROP LOGIN babel_object_id_login2;
 GO
 
 -- tsql
+USE master;
+GO
+
+CREATE LOGIN babel_object_id_login3 WITH PASSWORD = '12345678';
+GO
+~~ERROR (Code: 33557097)~~
+
+~~ERROR (Message: The Server principal 'babel_object_id_login3' already exists)~~
+
+
+CREATE USER babel_object_id_master_user3 FOR LOGIN babel_object_id_login1;
+GO
+~~ERROR (Code: 33557097)~~
+
+~~ERROR (Message: role "babel_object_id_login1" does not exist)~~
+
+
+CREATE LOGIN babel_object_id_login4 WITH PASSWORD = '12345678';
+GO
+~~ERROR (Code: 33557097)~~
+
+~~ERROR (Message: The Server principal 'babel_object_id_login4' already exists)~~
+
+
+CREATE USER babel_object_id_master_user4 FOR LOGIN babel_object_id_login1;
+GO
+~~ERROR (Code: 33557097)~~
+
+~~ERROR (Message: role "babel_object_id_login1" does not exist)~~
+
+
+-- tsql      user=babel_object_id_login3 password=12345678
 
 
 -- Test Table Variables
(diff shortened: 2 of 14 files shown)


contrib/babelfishpg_tsql/src/guc.c
```diff
@@ -1163,7 +1163,7 @@ define_custom_variables(void)
 							gettext_noop("Temp oid buffer size"),
 							NULL,
 							&temp_oid_buffer_size,
-							0, 0, 131072,
+							65536, 0, 131072,
```
lejaokri: need to reenable the following in `jdbc_schedule`

```
ignore#!#temp_table_jdbc
ignore#!#temp_tables_concurrent-vu-cleanup
ignore#!#temp_tables_concurrent-vu-prepare
ignore#!#temp_tables_concurrent-vu-verify
```

test/JDBC/expected/BABEL-4752.out
```diff
@@ -0,0 +1,264 @@
...
+3#!#service
+~~END~~
+
+~~ROW COUNT: 1~~
+
+~~START~~
+int#!#varchar
+1#!#product
+2#!#item
+4#!#item
```
lejaokri: im surprised this is `4#!#item`, and not `4#!#service` because of the `ROLLBACK` issue

timchang514:
Noting here: 4752 (pg_attrdef) is merged separately. Rollback issue is fixed now. We can't reenable temp OID changes until Locking and Invalidation is fixed wrt concurrency.

test/JDBC/expected/temp_tables_oid-vu-verify.out
```diff
@@ -0,0 +1,121 @@
+-- Tables typically consume three OIDs - we checked this in permanent tables.
```
lejaokri: nit: please add where the 3 oids are used for

test/JDBC/expected/temp_tables_oid-vu-verify.out
```diff
@@ -0,0 +1,121 @@
...
+0
+~~END~~
+
+
+
+-- identity and primary key should be in ENR as well
+create table #temp_tables_oid_t1(a int identity primary key, b int)
+go
+
+select relname from sys.babelfish_get_enr_list()
```
lejaokri, line 83: should we also verify the index are using temp oid?

test/JDBC/expected/temp_tables_oid-vu-verify.out
```diff
@@ -0,0 +1,121 @@
...
+GO
+~~START~~
+int
+0
+~~END~~
+
+
+
+-- identity and primary key should be in ENR as well
+create table #temp_tables_oid_t1(a int identity primary key, b int)
```
lejaokri, line 80: 1/ Is there a way to test what happens if we can assign temp oid for the identity and table but we fail to allocate temp oid for pkey because no oid available?

test/JDBC/expected/temp_tables_oid-vu-verify.out
timchang514: done

test/JDBC/expected/temp_tables_oid-vu-verify.out
timchang514: so, this is an interesting case i ran into while testing. i believe this might be the same as PG's default behavior with OID wraparound. 

here's what I happens, say if the buffer size is 4 for simplicity. 

1. I do `create table #t1...`, which consumes the first 3 oids. (call it OID 1000, 1001, 1002). 
2. I do `create table #t2...`. The table OID is the 4th oid (1004). 
3. Then, the buffer wraps around when we try to create the array type OID. In `GetNewOidWithIndex` and therefore `GetNewTempOidWithIndex`, PG calls `systable_beginscan` on `pg_type` with OID (1000), because, it wants to see if there is a collision. But it won't pickup OID 1000 because that's an entry in pg_class. So it says oh, looks like 1000 is free. And allows it to continue. 

At the very least, it means that we definitely won't have OID collisions in the same catalog. but there may be OID collisions in separate catalogs, and I'm not sure whether that may have any side effects.

I didn't run into any issues with dropping or using the tables with overlapping OIDs in my limited amount of testing. 

with the default value of 65k, it will also take a while for a customer to run into this - basically in the realm of ~20k temp tables created in an active session without dropping any.

lejaokri: > At the very least, it means that we definitely won't have OID collisions in the same catalog. 

Yes this is very importatnt

> but there may be OID collisions in separate catalogs, and I'm not sure whether that may have any side effects.

If I understand correctly, PG also allows this right?

test/JDBC/expected/temp_tables_oid-vu-verify.out
timchang514: done, and toast too

test/JDBC/expected/temp_tables_oid-vu-verify.out
timchang514: It's hard to find a verified source because OID wraparound seems quite rare, but the code implies that they do, since they're only searching the same catalog for collisions in the OID gen code.

timchang514: PG documentation specifies: 

```
The oid type is currently implemented as an unsigned four-byte integer. 
Therefore, it is not large enough to provide database-wide uniqueness 
in large databases, or even in large individual tables.
```

and in the comments of GetNewObjectId: 

```
 * OIDs are generated by a cluster-wide counter.  Since they are only 32 bits
 * wide, counter wraparound will occur eventually, and therefore it is unwise
 * to assume they are unique unless precautions are taken to make them so.
```

lejaokri (approved):

surendravishnoi (approved):

2jungkook (approved):

PR #2463, opened 2024-04-03
Free PLP token early when processing bulk load requests

### LINKED ISSUE #2462: [Bug]: BCP import memory leak with varchar(max)
### What happened?

This issue is similar to #2455, there seems to be a leak in processing Partially Length-prefixed (PLP) incoming records. Memory is allocated in `MessageContext`, so it is released after BCP import finishes. Still can be a problem when importing large files.

Excessive memory usage can be shown using BCP import with a table like this:

```tsql
create table tab1 (col1 varchar(max))
insert into tab1 values('foobar')
insert into tab1 values(null)
insert into tab1 select * from tab1
insert into tab1 select * from tab1
insert into tab1 select * from tab1
insert into tab1 select * from tab1
insert into tab1 select * from tab1
insert into tab1 select * from tab1
insert into tab1 select * from tab1
insert into tab1 select * from tab1
insert into tab1 select * from tab1
insert into tab1 select * from tab1
insert into tab1 select * from tab1
insert into tab1 select * from tab1
insert into tab1 select * from tab1
insert into tab1 select * from tab1
insert into tab1 select * from tab1
insert into tab1 select * from tab1
insert into tab1 select * from tab1
insert into tab1 select * from tab1
insert into tab1 select * from tab1
insert into tab1 select * from tab1
select count(*) from tab1
> 2097152
```

Will submit a PR.

### Version

BABEL_3_X_DEV (Default)

### Extension

babelfishpg_tds

Fedora

### Relevant log output

_No response_

### Code of Conduct

staticlibs:
### Description

`ParameterToken` [is allocated](https://github.com/babelfish-for-postgresql/babelfish_extensions/blob/4a1e451fcfd0387aa2479a1ee3605130c49e2c94/contrib/babelfishpg_tds/src/backend/tds/tdsbulkload.c#L692) to read incoming PLP  data.  While token itself is freed promptly, the [PlpData](https://github.com/babelfish-for-postgresql/babelfish_extensions/blob/4a1e451fcfd0387aa2479a1ee3605130c49e2c94/contrib/babelfishpg_tds/src/backend/tds/tdsbulkload.c#L1099) records inside it are not freed. Also the token itself can be leaked when the record is null.

Backend memory usage importing 2 million records (half of them are nulls, see details in linked issue), without the patch:

Figure_1

With patch applied:

Figure_2

### Issues Resolved

#2462

There are no functional changes, so no new tests.

### Check List

### DIFF
--- a/contrib/babelfishpg_tds/src/backend/tds/tdsbulkload.c
+++ b/contrib/babelfishpg_tds/src/backend/tds/tdsbulkload.c
@@ -34,6 +34,7 @@ void		ProcessBCPRequest(TDSRequest request);
 static void FetchMoreBcpData(StringInfo *message, int dataLenToRead, bool freeMessageData);
 static void FetchMoreBcpPlpData(StringInfo *message, int dataLenToRead);
 static int	ReadBcpPlp(ParameterToken temp, StringInfo *message, TDSRequestBulkLoad request);
+static void FreePlpToken(ParameterToken token);
 uint64_t	offset = 0;
 
 #define COLUMNMETADATA_HEADER_LEN			sizeof(uint32_t) + sizeof(uint16) + 1
@@ -698,14 +699,14 @@ SetBulkLoadRowData(TDSRequestBulkLoad request, StringInfo message)
 							{
 								rowData->isNull[i] = true;
 								i++;
-								token->isNull = false;
+								FreePlpToken(token);
 								continue;
 							}
 
 							/* Free the previously allocated temp. */
 							pfree(temp);
 							temp = TdsGetPlpStringInfoBufferFromToken(message->data, token);
-							pfree(token);
+							FreePlpToken(token);
 						}
 
 						/*
@@ -835,7 +836,7 @@ SetBulkLoadRowData(TDSRequestBulkLoad request, StringInfo message)
 						 * for the next iteration.
 						 */
 						pfree(temp->data);
-						pfree(token);
+						FreePlpToken(token);
 					}
 					break;
 				case TDS_TYPE_SQLVARIANT:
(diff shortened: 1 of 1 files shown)


contrib/babelfishpg_tds/src/backend/tds/tdsbulkload.c
```diff
@@ -698,14 +699,14 @@ SetBulkLoadRowData(TDSRequestBulkLoad request, StringInfo message)
 							{
 								rowData->isNull[i] = true;
 								i++;
-								token->isNull = false;
+								FreePlpToken(token);
```
Deepesh125, line 702: 1. why do we need to call FreePlpToken when we know certainly that it is PLP_NULL? 
2. why is `token->isNull = false` not needed anymore?

Deepesh125 (changes requested):

contrib/babelfishpg_tds/src/backend/tds/tdsbulkload.c
staticlibs: > 1. why do we need to call FreePlpToken when we know certainly that it is PLP_NULL?

`ParameterTokenData` struct itself needs to be freed in this case to not leak it. It is indeed not necessary to call `FreePlpToken` on it, `pfree(token)` would be enough. `FreePlpToken` is used here for consistency, it works correctly on empty token.

>     2. why is `token->isNull = false` not needed anymore?

AFAICS this flag (and the token struct itself) is not accessed after that point - token struct is [allocated again](https://github.com/babelfish-for-postgresql/babelfish_extensions/blob/4a1e451fcfd0387aa2479a1ee3605130c49e2c94/contrib/babelfishpg_tds/src/backend/tds/tdsbulkload.c#L692) before the next use. So we are freeing the token in this case instead of resetting `isNull` flag.

Deepesh125: Makes sense! Thanks for clarification

Deepesh125 (approved):

Deepesh125:
We should merge this fix on BABEL_2_X_DEV and BABEL_3_X_DEV

KushaalShroff (approved):

PR #2469, opened 2024-04-05
Free heap allocated Datums after the flush

### LINKED ISSUE #2468: [Bug]: BCP import records are buffered in memory
### What happened?

When BCP import is performed, fields of type `varchar` (and some other types) from incoming records are allocated in `MessageContext`. These buffered fields are kept in memory for all incoming records until the end of BCP import call.

For example, for table like this, backend memory usage on BCP import will grow linearly if we increase the number of records:

```tsql
create table tab1(col1 varchar(10))
insert into tab1 values('foobar')
insert into tab1 select col1 from tab1
insert into tab1 select col1 from tab1
insert into tab1 select col1 from tab1
insert into tab1 select col1 from tab1
insert into tab1 select col1 from tab1
insert into tab1 select col1 from tab1
insert into tab1 select col1 from tab1
insert into tab1 select col1 from tab1
insert into tab1 select col1 from tab1
insert into tab1 select col1 from tab1
insert into tab1 select col1 from tab1
insert into tab1 select col1 from tab1
insert into tab1 select col1 from tab1
insert into tab1 select col1 from tab1
insert into tab1 select col1 from tab1
insert into tab1 select col1 from tab1
insert into tab1 select col1 from tab1
insert into tab1 select col1 from tab1
insert into tab1 select col1 from tab1
insert into tab1 select col1 from tab1
select count(*) from tab1
> 1048576
```

Unlike #2455 and #2462, it is not trivial to free these allocations promptly because of two levels of batching - implicit one on protocol side and `MAX_BUFFERED_TUPLES` (1000) on executor side.

I'll file a PR with an experimental patch to track these allocations and free them as soon as possible.

### Version

BABEL_3_X_DEV (Default)

### Extension

babelfishpg_tsql (Default)

Fedora

### Relevant log output

_No response_

### Code of Conduct

staticlibs:
### Description

When BCP import is performed, `Datums` are created for all non-NULL incoming fields. Some of these `Datums` are [allocated on heap](https://github.com/babelfish-for-postgresql/babelfish_extensions/blob/b26029536061182f8b2b216d4a149922606040e9/contrib/babelfishpg_tds/src/backend/tds/tdstypeio.c#L1449). These `Datums` need to be kept in memory until the call to `table_multi_insert`.

It seems to be not possible to distinguish between heap-allocated `Datums` and passed-by-value `Datums` using only the pointer to such `Datum`. The proposed patch attempts to keep track of all heap-allocated `Datums` (in a similar manner, like `isNull` is tracked) so they can be freed immediately after the flush.

Pointers to `Datums` and `isAllocated` flags are kept in `BulkCopyStateData` in growing lists. Pointers from incoming batch are appended there before processing. And these `Datums` are freed and lists are trimmed after each flush.

There are also 2 unrelated memory cleanup changes included: freeing of `TDSRequestBulkLoadData->rowData` contents on TDS side and freeing of `attnums` list on TSQL side. `attnums` is created for every batch (I assume this is needed for error-checking), but only used for the first incoming batch.

Memory usage when importing 1 million of `varchars` (see details in linked issue), without the patch:

Figure_1

With patch applied:

Figure_2

### Issues Resolved

#2468

There are no functional changes, so no new tests.

### Check List

### DIFF
--- a/contrib/babelfishpg_tds/src/backend/tds/tdsbulkload.c
+++ b/contrib/babelfishpg_tds/src/backend/tds/tdsbulkload.c
@@ -516,6 +516,7 @@ SetBulkLoadRowData(TDSRequestBulkLoad request, StringInfo message)
 
 		rowData->columnValues = palloc0(request->colCount * sizeof(Datum));
 		rowData->isNull = palloc0(request->colCount * sizeof(bool));
+		rowData->isAllocated = palloc0(request->colCount * sizeof(bool));
 
 		offset++;
 		request->currentBatchSize++;
@@ -593,6 +594,7 @@ SetBulkLoadRowData(TDSRequestBulkLoad request, StringInfo message)
 								break;
 							case TDS_TYPE_DATETIMEOFFSET:
 								rowData->columnValues[i] = TdsTypeDatetimeoffsetToDatum(temp, colmetadata[i].scale, temp->len);
+								rowData->isAllocated[i] = true;
 								break;
 							case TDS_TYPE_MONEYN:
 								if (colmetadata[i].maxLen == TDS_MAXLEN_SMALLMONEY)
@@ -602,6 +604,7 @@ SetBulkLoadRowData(TDSRequestBulkLoad request, StringInfo message)
 								break;
 							case TDS_TYPE_UNIQUEIDENTIFIER:
 								rowData->columnValues[i] = TdsTypeUIDToDatum(temp);
+								rowData->isAllocated[i] = true;
 								break;
 						}
 
@@ -646,6 +649,7 @@ SetBulkLoadRowData(TDSRequestBulkLoad request, StringInfo message)
 						 * column.
 						 */
 						rowData->columnValues[i] = TdsTypeNumericToDatum(temp, colmetadata[i].scale);
+						rowData->isAllocated[i] = true;
 
 						offset += len;
 						request->currentBatchSize += len;
@@ -717,14 +721,17 @@ SetBulkLoadRowData(TDSRequestBulkLoad request, StringInfo message)
 							case TDS_TYPE_CHAR:
 							case TDS_TYPE_VARCHAR:
 								rowData->columnValues[i] = TdsTypeVarcharToDatum(temp, colmetadata[i].encoding, colmetadata[i].columnTdsType);
+								rowData->isAllocated[i] = true;
 								break;
 							case TDS_TYPE_NCHAR:
 							case TDS_TYPE_NVARCHAR:
 								rowData->columnValues[i] = TdsTypeNCharToDatum(temp);
+								rowData->isAllocated[i] = true;
 								break;
 							case TDS_TYPE_BINARY:
 							case TDS_TYPE_VARBINARY:
 								rowData->columnValues[i] = TdsTypeVarbinaryToDatum(temp);
+								rowData->isAllocated[i] = true;
 								break;
 						}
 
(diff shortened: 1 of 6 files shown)


contrib/babelfishpg_tsql/src/pltsql_bulkcopy.c
```diff
@@ -588,6 +629,13 @@ ExecuteBulkCopy(BulkCopyState cstate, int rowCount, int colCount,
 							RelationGetRelationName(cstate->rel))));
 	}
 
+	/* Track all incoming Datums to be able to free them after the flush */
+	for (int i = 0; i < rowCount * colCount; i++)
+	{
+		cstate->bufferedValues = lappend(cstate->bufferedValues, (void *) Values[i]);
+		cstate->bufferedValueAllocFlags = lappend_int(cstate->bufferedValueAllocFlags, ValueAllocFlags[i] ? 1 : 0);
```
KushaalShroff: Why cant we directly store booleans in cstate->bufferedValueAllocFlags instead of integers?

KushaalShroff (changes requested):

contrib/babelfishpg_tds/src/backend/tds/tdsbulkload.c
```diff
@@ -717,14 +721,17 @@ SetBulkLoadRowData(TDSRequestBulkLoad request, StringInfo message)
...
 								break;
 							case TDS_TYPE_NCHAR:
 							case TDS_TYPE_NVARCHAR:
 								rowData->columnValues[i] = TdsTypeNCharToDatum(temp);
+								rowData->isAllocated[i] = true;
 								break;
 							case TDS_TYPE_BINARY:
 							case TDS_TYPE_VARBINARY:
 								rowData->columnValues[i] = TdsTypeVarbinaryToDatum(temp);
+								rowData->isAllocated[i] = true;
```
KushaalShroff: common condition in all of the cases, can we move it out?

contrib/babelfishpg_tds/src/backend/tds/tdsbulkload.c
```diff
@@ -794,12 +801,15 @@ SetBulkLoadRowData(TDSRequestBulkLoad request, StringInfo message)
...
 								rowData->columnValues[i] = TdsTypeVarcharToDatum(temp, colmetadata[i].encoding, colmetadata[i].columnTdsType);
+								rowData->isAllocated[i] = true;
 								break;
 							case TDS_TYPE_NTEXT:
 								rowData->columnValues[i] = TdsTypeNCharToDatum(temp);
+								rowData->isAllocated[i] = true;
 								break;
 							case TDS_TYPE_IMAGE:
 								rowData->columnValues[i] = TdsTypeVarbinaryToDatum(temp);
+								rowData->isAllocated[i] = true;
```
KushaalShroff: Common condition between all cases, can we move it out?

contrib/babelfishpg_tsql/src/pltsql_bulkcopy.c
staticlibs: Just wanted to be explicit with `T_IntList` usage. Updated this line to remove int conversion.

contrib/babelfishpg_tds/src/backend/tds/tdsbulkload.c
staticlibs: Moved it out of the condition and added a comment.

contrib/babelfishpg_tds/src/backend/tds/tdsbulkload.c
staticlibs: Moved it out of the condition and added a comment.

PR #2491, opened 2024-04-11
Fix crash when trying to delete a trigger on a temp table.

Sairakan:
Instead, raise an error.

Task: BABEL-4867

### Description

Triggers do not have visibility into temp tables from other sessions. In particular, if a trigger is created on a temp table, and then a separate session attempts to drop that trigger, the reference temp table  will not be visible within the second session, and a null OID reference will be passed through the internal code. Currently this causes a segmentation fault, so fix it by detecting for this case and erroring out instead.

### Issues Resolved

BABEL-4867

* **Use case based -**
Java test which replicates the error scenario by creating a trigger on a temp table in one session and then attempting to drop the trigger in another session. Due to requiring two simultaneous sessions to be tested, this currently can only be tested in the Java file and cannot be tested in upgrade scenarios.

* **Boundary conditions -**

* **Arbitrary inputs -**

* **Negative test cases -**

* **Performance tests -**

* **Tooling impact -**

* **Client tests -**

### Check List

### DIFF
--- a/contrib/babelfishpg_tsql/src/hooks.c
+++ b/contrib/babelfishpg_tsql/src/hooks.c
@@ -2338,6 +2338,12 @@ get_tsql_trigger_oid(List *object, const char *tsql_trigger_name, bool object_fr
 			reloid = pg_trigger->tgrelid;
 			relation = RelationIdGetRelation(reloid);
 			pg_trigger_physical_schema = get_namespace_name(get_rel_namespace(pg_trigger->tgrelid));
+			if (pg_trigger_physical_schema == NULL)
+			{
+				ereport(ERROR,
+						(errcode(ERRCODE_INTERNAL_ERROR),
+						errmsg("error while trying to look up trigger")));
+			}
 			if (strcasecmp(pg_trigger_physical_schema, cur_physical_schema) == 0)
 			{
 				trigger_rel_oid = reloid;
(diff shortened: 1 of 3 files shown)


contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -2338,6 +2338,12 @@ get_tsql_trigger_oid(List *object, const char *tsql_trigger_name, bool object_fr
 			reloid = pg_trigger->tgrelid;
 			relation = RelationIdGetRelation(reloid);
 			pg_trigger_physical_schema = get_namespace_name(get_rel_namespace(pg_trigger->tgrelid));
+			if (pg_trigger_physical_schema == NULL)
+			{
+				ereport(ERROR,
+						(errcode(ERRCODE_INTERNAL_ERROR),
+						errmsg("error while trying to look up trigger")));
+			}
```
lejaokri: Do we want to throw an error or just return `NULL` and allow caller to decide whether to throw error or not?

One of the callers of `get_tsql_trigger_oid` is `pre_check_trigger_schema(List *object, bool missing_ok)`
and it has the parameter `missing_ok`. 

Consider the case when `pre_check_trigger_schema(object, missing_ok=TRUE)`, but `get_tsql_trigger_oid` errors out.

Sairakan: Oh, I wasn't aware of this consumer. I went with returning this error because if we return `InvalidOid` (which would be the equivalent of returning NULL), then the error message raised is `trigger "bar" does not exist`. Tim suggested that this might be a misleading error message, which is why I decided to go with this approach of a distinct error message if the target relation cannot be found. However, since `pre_check_trigger_schema()` doesn't call `get_tsql_trigger_oid` within a PG_TRY() block, I guess we have to just return InvalidOid or else we might get an error when we don't want to.

I suppose this is why Java has such an explicit exception syntax where functions need to be marked if they can throw exceptions (and callers of such functions must use try/catch blocks as well or else there will be a compilation error).

Sairakan: Updated to just return an InvalidOid instead of raising an error.

lejaokri (approved):

forestkeeper (approved):

PR #2506, opened 2024-04-17
FETCH from CURSOR inside a function should be disabled without a destination

shardgupta:
## Description

SELECT statements included within a function cannot return data to a client. This operation was not blocked with the FETCH CURSOR operation. This commit aim to restrict defining such function.

Issues Resolved: BABEL-4586

Cherry-pick from 4_X


thephantomthief (approved):

jsudrik (approved):
Cheerypick from 4_x. Approved.

PR #2528, opened 2024-04-24
Add support for Alter Procedure in Babelfish

Jakeowen1:
### Description

This commit adds support for Alter Procedure implementation in Babelfish. It adds a new parser rule to handle alter procedure cases, and creates a new procedure in place of the old one. All of the existing privileges are carried to the new procedure. The user can choose to modify the procedure parameters and/or body.

### Issues Resolved
BABEL-2661

* **Use case based -**

```
create procedure alter_proc1 as select 1
go

exec alter_proc1
go
~~START~~
int
1
~~END~~


alter procedure alter_proc1 as select 2
go

exec alter_proc1
go
~~START~~
int
2
~~END~~
```

```
create procedure alter_proc2 @l int as select @l
go

exec alter_proc2 @l = 200
go
~~START~~
int
200
~~END~~


alter procedure alter_proc2 
    @l int
as
    if (@l = 1)
    begin
        select 100
    END
    ELSE
    begin
        select 0
    end
go

exec alter_proc2 @l = 1
go
~~START~~
int
100
~~END~~


exec alter_proc2 @l = 200
go
~~START~~
int
0
~~END~~
```

* **Arbitrary inputs -**
```
create procedure alter_proc2 as select 1
go

create procedure alter_proc3 
as 
    exec alter_proc2
go

alter procedure alter_proc2 
    @l int
as
    select @l
go

exec alter_proc2 @l = 1
go
~~START~~
int
1
~~END~~


exec alter_proc3
go
~~ERROR (Code: 201)~~

~~ERROR (Message: procedure alter_proc2 expects parameter "@l", which was not supplied.)~~
```

* **Negative test cases -**

```
CREATE PROCEDURE alter_proc_p1 
AS
    select * from alter_proc_users
GO

-- Expect error for procedure with same name
CREATE PROCEDURE alter_proc_p1 @param1 int
AS
    select * from alter_proc_orders
GO
~~ERROR (Code: 2714)~~

~~ERROR (Message: Function 'alter_proc_p1' already exists with the same name)~~
```

Test file added to schedule

Test file added to schedule

* **Performance tests -**

* **Tooling impact -**
None

* **Client tests -**
Manual Concurrency Testing:
```
-- Initialize Procedure
-- Open two threads
-- Thread 1: Runs a while loop executing the procedure
-- Thread 2: Alter procedure

-- Run the two thread in parallel

-- output: Thread 1 prints the output of the first procedure, and after alter is run, thread 1 outputs the results of the newly created procedure
```

```
-- Initialize Procedure
-- Open 3 threads
-- Thread 1: Runs a while loop executing the procedure
-- Thread 2: Alter procedure
-- Thread 3: Alter procedure

-- Run 3 threads in parallel

-- Output: One of the alter threads will error with "tuple concurrently updated", and Thread 1 will output the result of the altered procedure that did not error
```

```
-- Initialize Procedure
-- Open 4 threads
-- Thread 1: exec proc, Thread 2: Alter proc, Thread 3: exec proc, Thread 4: Alter proc
-- Run threads 1 and 2 in parallel, then run threads 3 and 4 in parallel
-- Thread 1 will output the initial proc result, followed by the altered proc from thread 2
-- Thread 3 will output the altered proc from thread 2 result, following by the altered proc result from thread 4

-- No error for "tuple concurrently updated" like in last example, because the alter proc threads are not executed in parallel
```

alter-procedure-schema.mix file contains tests demonstrating how permissions are carried over for procedures created on schema.

### Check List


contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -2320,6 +2323,95 @@ bbf_ProcessUtility(PlannedStmt *pstmt,
...
+					cfs->options = stmt->actions;
+
+					pltsql_proc_get_oid_proname_proacl(stmt, pstate, &oldoid, &proacl, &isSameProc);
+					if (!isSameProc) /* i.e. different signature */
+						RemoveFunctionById(oldoid);
+					address = CreateFunction(pstate, cfs); // if this is the same proc, will just update the existing one
+					pg_proc_update_oid_acl(address, oldoid, *proacl);
+					/* Update function/procedure related metadata in babelfish catalog */
+					if (isSameProc)
+						pltsql_store_func_default_positions(address, cfs->parameters, queryString, origname_location);
```
Sairakan: can you add a comment here explaining why you don't need to do all of the extra catalog modifications in this case?

Jakeowen1: Added a comment to this line

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -3209,6 +3236,295 @@ pltsql_store_func_default_positions(ObjectAddress address, List *parameters, con
 	table_close(bbf_function_ext_rel, RowExclusiveLock);
 }
 
+/*
+ * Update 'function_args' in 'sys.babelfish_schema_permissions' 
+ */
+void
+alter_bbf_schema_permissions_catalog(ObjectWithArgs *owa, List *parameters, int objtypeInt, Oid oid)
```
Sairakan, line 3259: is it possible to merge this with the existing function(s) to modify the catalog? That way there will be less overhead/bugs if the catalog changes in the future

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -3209,6 +3236,295 @@ pltsql_store_func_default_positions(ObjectAddress address, List *parameters, con
...
+	}
+	table_close(bbf_schema_rel, RowExclusiveLock);
+}
+
+/*
+ * Update 'funcsignature', 'default_positions', 'flag_validity', 'flag_values', 
+ * 'modify_date', and 'definition' in 'babelfish_function_ext'
+ */
+void
+alter_bbf_function_ext(ObjectAddress address, ObjectWithArgs *owa, List *parameters, const char *queryString, int origname_location)
```
Sairakan: Same as above, can we instead modify the original function that handles babelfish_function_ext to deal with replacement instead of creating a new function that copies much of the same code?

contrib/babelfishpg_tsql/src/pl_handler.c
Sairakan: I think the comment needs more detail. In particular, we should explain that in the case that the signature is the same, then no other catalog entries are made, in pg_depend, or other places, due to the fact that CreateFunction just modifies the existing function instead of creating a new one. On the other hand, if the signature is different, then there are dangling dependencies from the original function that need to be cleaned up since new ones get created for the new function signature. Consider updating the comment in the `else` clause to help clarify the logic being used here.

contrib/babelfishpg_tsql/src/backend_parser/parser.c
```diff
@@ -314,6 +317,15 @@ pgtsql_base_yylex(YYSTYPE *lvalp, YYLTYPE * llocp, core_yyscan_t yyscanner)
 			if (next_token == TSQL_APPLY)
 				cur_token = TSQL_OUTER;
 			break;
+		case ALTER:
+			switch (next_token)
+			{
+				case PROCEDURE:
+				case TSQL_PROC:
```
forestkeeper, line 324: What's the difference between PROCEDURE/TSQL_PROC

contrib/babelfishpg_tsql/src/backend_parser/parser.c
```diff
@@ -180,6 +180,9 @@ pgtsql_base_yylex(YYSTYPE *lvalp, YYLTYPE * llocp, core_yyscan_t yyscanner)
 		case OUTER_P:
 			cur_token_length = 5;
 			break;
+		case ALTER:
```
forestkeeper, line 183: What's difference between 
TSQL_ALTER and ALTER

contrib/babelfishpg_tsql/src/backend_parser/parser.c
Jakeowen1:
```
PG_KEYWORD("proc", TSQL_PROC, UNRESERVED_KEYWORD)
PG_KEYWORD("procedure", PROCEDURE, UNRESERVED_KEYWORD)
```

TSQL_PROC is "proc" and PROCEDURE is "procedure"

contrib/babelfishpg_tsql/src/backend_parser/parser.c
Jakeowen1:
```
PG_KEYWORD("alter", ALTER, UNRESERVED_KEYWORD)
```

My understanding of TSQL_ALTER is that it's the token that represents Alter Procedure. We check the parser to see if we hit "alter", then check the next token to see if its "proc" or "procedure", in which case we set the current token to TSQL_ALTER

Sairakan: These changes were made in order to avoid parser conflicts due to native PG's interpretation of ALTER PROCEDURE

contrib/babelfishpg_tsql/src/hooks.c
Jakeowen1: Just removed this function. Using pltsql_store_func_default_positions for both cases now

contrib/babelfishpg_tsql/src/hooks.c
Jakeowen1: The update functions for bbf_schema_permissions are specific to each parameter. I think it makes sense to keep this as a separate function, but potentially move it to catalog.c where the other bff_schema_permissions functions are

test/JDBC/expected/alter-procedure-vu-verify.out
```diff
@@ -0,0 +1,118 @@
...
+exec alter_proc_p1 @param = '2020-01-02'
+GO
+~~START~~
+int#!#int#!#int#!#int#!#date
+1#!#1#!#1#!#5#!#2023-06-25
+2#!#1#!#1#!#6#!#2023-06-25
+~~END~~
+
+
+alter procedure alter_proc_p2
```
forestkeeper, line 131: Should add 
alter proc syntax case

test/JDBC/expected/alter-procedure-vu-verify.out
```diff
@@ -0,0 +1,118 @@
...
+AS
+    exec alter_proc_p1 @param = '2020-01-01'
+GO
+
+exec alter_proc_p2
+go
+~~START~~
+int#!#varchar#!#varchar#!#varchar
```
forestkeeper, line 139: Do we have a case showing alter func will failed with proper err msg ?

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -3886,6 +3980,106 @@ call_prev_ProcessUtility(PlannedStmt *pstmt,
...
+	funcOid = LookupFuncWithArgs(stmt->objtype, stmt->func, true);
+
+	*isSameFunc = OidIsValid(funcOid);
+}
+
+/*
+ * Update the oid and acl of a pg_proc entry given its address
+ */
+static void
+pg_proc_update_oid_acl(ObjectAddress address, Oid oid, Acl acl)
```
forestkeeper: This function didn't use acl ?
Why we have update acl in the function body ?

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -3840,7 +3841,13 @@ gen_sp_rename_subcmds(const char *objname, const char *newname, const char *sche
 				 errmsg("Provided objtype is not supported for sp_rename")));
 			break;
 	}
+	old_dialect = sql_dialect;
+	sql_dialect = SQL_DIALECT_PG;
+	/* this query must be run in PG dialect or we will get a syntax error due to different
+	 * ALTER behavior between PG and TSQL
```
forestkeeper, line 3847: Which test case is corresponding to this code ?

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -3886,6 +3980,106 @@ call_prev_ProcessUtility(PlannedStmt *pstmt,
...
+	if ((spi_rc = SPI_connect()) != SPI_OK_CONNECT)
+		elog(ERROR, "SPI_connect() failed in pltsql_proc_get_oid_proname_proacl with return code %d", spi_rc);
+
+	query = psprintf("SELECT oid, proacl FROM pg_catalog.pg_proc WHERE proname = '%s' AND pronamespace = %d", funcname, schemaOid);
+	SPI_execute(query, true, 0);
+
+	if (SPI_processed > 1)
+		ereport(ERROR, 
+			(errcode(ERRCODE_AMBIGUOUS_FUNCTION),
+				errmsg("Multiple procedures are defined with the same name and different parameters. Please ensure that there is only one procedure with the target name before calling ALTER PROCEDURE")));
```
forestkeeper, line 4012: Would you add a test case for this ?

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -3886,6 +3980,106 @@ call_prev_ProcessUtility(PlannedStmt *pstmt,
 								queryEnv, dest, qc);
 }
 
+/*
+ * Get the oid and acl of a TSQL proc by name. Raises an error if the proc doesn't exist, or if there are
+ * multiple procs with the same name (and different parameters). Also sets isSameFunc based on whether
+ * the found proc is the exact same proc as requested (i.e. the parameters match).
+ */
+static void 
+pltsql_proc_get_oid_proname_proacl(AlterFunctionStmt *stmt, ParseState *pstate, Oid *oid, Acl **acl, bool *isSameFunc)
```
forestkeeper, line 3991: For the **acl, I didn't see anything done in this function.

test/JDBC/input/alter/alter-procedure-vu-verify.sql
```diff
@@ -0,0 +1,66 @@
+exec alter_proc_p1
```
forestkeeper, line 1: Could you add some comments in the test cases ? Like which one is verify which part, for example : 
1. proc arg change
2. dup function
3. basic err msg
4. proc body
5. transaction support

and so on

test/JDBC/expected/alter-procedure-schema.out
```diff
@@ -0,0 +1,172 @@
...
+create schema alter_proc_schema2
+go
+
+create login alter_proc_l2 with password = '12345678'
+go
+
+create user alter_proc_u2 for login alter_proc_l2
+go
+
+create proc alter_proc_schema2.p2 @l int as select 1;
```
forestkeeper, line 139: Pls also add same schema , different name and different schema, same name test case

test/JDBC/input/alter/alter-procedure-vu-verify.sql
```diff
@@ -0,0 +1,66 @@
...
+exec alter_proc_p1 @param = 1
+GO
+
+exec alter_proc_p1 @param = 2
+GO
+
+-- expect error for no param provided
+exec alter_proc_p2
+go
+
```
forestkeeper, line 54: Pls add transaction related case, for example : 
1. begin, do alter, rollback
2. begin, do alter, that alter can't be seen in other session, commit, that alter can be seen in other session
3. begin. do alter. do other row change. commit
4. begin. do alter. do other row change. rollback
5. begin. do alter deliberate result in error. make sure that alter didn't take effect.

contrib/babelfishpg_tsql/src/pl_handler.c
Jakeowen1: We need to take the acl from the old procedure and copy it into the new procedure so that the permissions stay the same. I'll modify this function body to do so

contrib/babelfishpg_tsql/src/pl_handler.c
Jakeowen1: We set the acl pointer in this function so we can copy it into the new procedure in `pg_proc_update_oid_acl`

contrib/babelfishpg_tsql/src/procedures.c
Jakeowen1: Just spoke with Jason - this is required to avoid a syntax error when parsing alter proc. If we remove the code there is a syntax error

test/JDBC/input/alter/alter-procedure-vu-verify.sql
Jakeowen1: Going to push these test cases soon - fyi I tested case 2 manually using the python concurrency script I wrote

forestkeeper (approved):
LGTM

PR #2536, opened 2024-04-29
Fix searchpath handling for tsql udfs inside fmgr

tanscorpio7:
### Description

Currently we override the search_path for pltsql UDFs directly using the GUC variable & avoid set_config.
This leads to crash if there was another set_config('search_path... when we were in this overridden state.
Reason for crash is that the new set_config call may push the current value (the one we set in fmgr) onto
GUC stack. Now it pssoble that we exit the function call and released fcache, but fcache->prosearchpath
is still referenced in GUC stack.

To fix this we fallback to using set_config for setting the search_path & skip the check hook for search path
when setting from inside fmgr which helps with performance.
Also increase the GUC nest level since GUC_ACTION_SAVE is not possible on top of GUC_ACTION_SET
(user set), which is possible in this case. New nest level also means that we do not have to reset the search
path at the end of execution and let AtEOXact_GUC(commit=true) handle it.

We also move the set search path block in fmgr inside TRY, to ensure we reset the skip check search path
variable in case of errors.

### Issues Resolved

[BABEL-4877]

### Sign Off

### Check List


test/JDBC/input/BABEL_4877.mix
```diff
@@ -0,0 +1,178 @@
...
+
+DROP FUNCTION babel_4877_s1.babel_4877_f1
+GO
+
+EXEC dbo.babel_4877_reset_search_path
+GO
+
+-- now search path = reset val 
+
+CREATE FUNCTION babel_4877_s1.babel_4877_f1() RETURNS VARCHAR(100)
```
surendravishnoi, line 90: can you try this scenario?
begin tran();
check search path;
call f1(); -> throw an error which is stmt terminating (divide by zero or similar)
check search path;
commit;

tanscorpio7: added test cases

surendravishnoi (approved):

shardgupta (approved):

PR #2563, opened 2024-05-09
Support EXECUTE/CREATE...WITH RECOMPILE for stored procedures

robverschoor:
### Description

T-SQL supports **EXECUTE <procedure> WITH RECOMPILE** and **CREATE PROCEDURE ... WITH RECOMPILE**. 
For **EXECUTE WITH RECOMPILE**, an ad-hoc plan is generated for that particular execution.
For **CREATE PROCEDURE WITH RECOMPILE**, an ad-hoc plan is generated for every execution of the procedure, also when the **EXECUTE** statement does not specify **WITH RECOMPILE**. 
In both cases, the ad-hoc generated plan is discarded immediately afterwards.

In Babelfish, forcing ad-hoc plan generation is implemented by setting the GUC "**plan_cache_mode**" to "**force_custom_plan**" for the particular procedure execution.
For **CREATE PROCEDURE WITH RECOMPILE**, and additional status bit is used in the **babelfish_function_ext.flag_values** catalog field. 
Currently, recompilation will not apply to non-parametrized queries since PG will not generate a custom plan for such queries. 

Engine PR: https://github.com/babelfish-for-postgresql/postgresql_modified_for_babelfish/pull/360

### Issues Resolved

BABEL-346 Support WITH RECOMPILE for Transact-SQL stored procedures
 
## Test Scenarios Covered

- Use case based - Yes

- Boundary conditions - N/A

- Arbitrary inputs - N/A

- Negative test cases - N/A

- Performance tests - N/A

- Tooling impact - N/A

- Client tests - N/A

### Check List


contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -1299,6 +1325,16 @@ exec_stmt_exec(PLtsql_execstate *estate, PLtsql_stmt_exec *stmt)
 								 GUC_ACTION_SAVE, true, 0, false);
 		SetCurrentRoleId(current_user_id, false);
 	}
+	
+	/*
+	 * Restore plan_cache_mode
+	 */
+	if ((stmt->exec_with_recompile || created_with_recompile) && !old_plan_mode)
+	{
+		(void) set_config_option("plan_cache_mode", old_plan_mode,
```
Deepesh125: what would be the performance implication of this change?

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -2386,7 +2386,7 @@ bbf_ProcessUtility(PlannedStmt *pstmt,
 					address = CreateFunction(pstate, cfs); /* if this is the same proc, will just update the existing one */
 					pg_proc_update_oid_acl(address, oldoid, proacl);
 					/* Update function/procedure related metadata in babelfish catalog */
-					pltsql_store_func_default_positions(address, cfs->parameters, queryString, origname_location);
+					pltsql_store_func_default_positions(address, cfs->parameters, queryString, origname_location, false);
```
Deepesh125, line 2383: why is recompile hardcoded to be false here? can't we specify with recompile option during alter procedure?

contrib/babelfishpg_tsql/src/tsqlIface.cpp
```diff
@@ -6301,6 +6315,16 @@ makeExecuteProcedure(ParserRuleContext *ctx, std::string call_type)
...
 		ss << ::getFullText(func_proc_args);		
 	}
+	
+	if (exec_with_recompile)
+	{
+		/* 
+		 * When displaying the query plan, indicate that EXECUTE ... WITH RECOMPILE was specified.
+		 * This may be relevant for understanding the plan.
+		 */
+		ss << " /*EXECUTE WITH RECOMPILE*/";
```
Deepesh125: wouldn't spaces in front and back make it more readable? `/* EXECUTE WITH RECOMPILE */`

test/JDBC/expected/with_recompile-vu-verify.out
```diff
@@ -0,0 +1,2754 @@
...
+ * Note that when a cached plan is used, there is no good way to prove in these tests that 
+ * it was indeed a cached plan, as it could just as well be a newly generated plan that 
+ * happens to be the same as the previous plan. While developing this fix, the PG planner() 
+ * function was instrumented to prove that cached plans are indeed used when plan_cache_mode={auto|force_generic_plan},
+ * but that is outside the scope of what can be used in these automated tests.
+ * This fix for forcing plans for statements in the procedure body to be recompiled only
+ * applies to paramtrized statements. Non-parametrized stmts, like the 2nd select in tb_recomp_11, 
+ * will not be recompiled since choose_custom_plan() will always return false for those.
+ * The current fix can be extended to also force such stmts to be recompiled but that is 
+ * left for later.
```
Deepesh125, line 18: Just to confirm: we have internal ticket to track this TODO?

test/JDBC/expected/with_recompile-vu-verify.out
```diff
@@ -0,0 +1,2754 @@
...
+~~END~~
+
+set babelfish_showplan_all on
+go
+execute p_recomp_11 1
+go
+~~START~~
+text
+Query Text: EXEC    p_recomp_11 1
+  Query Text: select * from tb_recomp_11 where c11 = "@p"
```
Deepesh125, line 102: we should run analyse on all the involved table in order to stabilised the plan.

contrib/babelfishpg_tsql/src/backend_parser/kwlist.h
```diff
@@ -383,6 +383,7 @@ PG_KEYWORD("readuncommitted", TSQL_READUNCOMMITTED, UNRESERVED_KEYWORD)
 PG_KEYWORD("real", REAL, COL_NAME_KEYWORD)
 PG_KEYWORD("reassign", REASSIGN, UNRESERVED_KEYWORD)
 PG_KEYWORD("recheck", RECHECK, UNRESERVED_KEYWORD)
+PG_KEYWORD("recompile", TSQL_RECOMPILE, RESERVED_KEYWORD)
```
Deepesh125: we marked TSQL_RECOMPILE token as unreserved keyword in tsql-rule but we are marking it to reserved keyword. Either of two is placed incorrectly. Please fix that and we also need to add test for this.

Deepesh125 (changes requested):

contrib/babelfishpg_tsql/src/pl_exec-2.c
robverschoor: The impact of running `is_created_with_recompile() `is 9.8% in my environment (1 million executions go from 27337 to 30027 seconds). This is the cost of a lookup for every proc call, also when RECOMPILE is not used. This is due to the lookup of the extended BBF catalog `babelfish_function_ext` which is currently using the function signature as a key and requires composing the signature for every call. We should optimize this lookup (separately) to be more efficient.

The impact of setting/resetting the GUC in case of RECOMPILE is 2.3% in my environment (1 million executions go from 30273 to 30969 seconds). This is acceptable since RECOMPILE means the user is accepting additional overhead in order to get a better plan.

contrib/babelfishpg_tsql/src/pl_handler.c
robverschoor: ALTER PROCEDURE will be implemented in a separate fix after this fix. This `false `constant is a placeholder until then. Added ToDo comment for clarity.

contrib/babelfishpg_tsql/src/tsqlIface.cpp
robverschoor: It's a matter of taste/style I guess, but will add spaces.

test/JDBC/expected/with_recompile-vu-verify.out
robverschoor: Will create a Jira

contrib/babelfishpg_tsql/src/backend_parser/kwlist.h
robverschoor: Changed to unreserved. There is already a test case for this: BABEL-2730.sql

test/JDBC/expected/with_recompile-vu-verify.out
robverschoor: all tables are empty throughout the test; no rows are inserted during the test.
for maximum predicatability, ANALYZE has been added.

Deepesh125 (dismissed):

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -1274,6 +1290,16 @@ exec_stmt_exec(PLtsql_execstate *estate, PLtsql_stmt_exec *stmt)
...
 			SetCurrentRoleId(current_user_id, false);
 
+		/*
+		 * old_plan_mode may be NULL when an exception occurred before capturing the old mode 
+		 */
+		if ((stmt->exec_with_recompile || created_with_recompile) && !old_plan_mode)
+		{
+			(void) set_config_option("plan_cache_mode", old_plan_mode,
+							  GUC_CONTEXT_CONFIG,
+							  PGC_S_SESSION, GUC_ACTION_SAVE, true, 0, false);
```
2jungkook: If `old_plan_mode` is NULL, that means an exception occurred before setting the config, hence we don't need to restore the config. 

Also, even if `old_plan_mode` is not NULL, we may not need to restore it because we used `GUC_ACTION_SAVE`. But this needs to verified. Do we have a test case for this code path?

contrib/babelfishpg_tsql/src/pl_exec-2.c
2jungkook: I don't think this code path is being executed because this is outside of TRY/CATCH block and we re-throw inside the CATCH block.

contrib/babelfishpg_tsql/src/tsqlIface.cpp
```diff
@@ -6120,17 +6123,27 @@ makeExecuteProcedure(ParserRuleContext *ctx, std::string call_type)
...
+		func_proc_args = body->execute_statement_arg();
+		exec_options   = body->execute_option();		
 	}
 	else // execute_body_batch
 	{
 		TSqlParser::Execute_body_batchContext *ctxEBB = (TSqlParser::Execute_body_batchContext *) ctx;
 		ctx_name       = ctxEBB->func_proc_name_server_database_schema();		
-		func_proc_args = ctxEBB->execute_statement_arg();		
+		func_proc_args = ctxEBB->execute_statement_arg();
+		exec_options   = ctxEBB->execute_option();
```
2jungkook: Could you explain when this code block, i.e., else-block, is executed? Do we have a test case for this?

contrib/babelfishpg_tsql/src/tsqlIface.cpp
```diff
@@ -6120,17 +6123,27 @@ makeExecuteProcedure(ParserRuleContext *ctx, std::string call_type)
...
-	}		
-	
+	}
 
+	for (TSqlParser::Execute_optionContext *opt : exec_options)
+	{
+		if (opt->RECOMPILE())
+		{
+			exec_with_recompile = true;
+			break;
```
2jungkook, line 6164: Could you add a comment explaining this `break`? e.g.,
```
We break the for-loop because currently we do support only one exec_options. 
Later, if we support more options, then we should remove the break.
```

test/JDBC/input/with_recompile-vu-prepare.mix
```diff
@@ -0,0 +1,172 @@
...
+insert tb_recomp_51
+select * from tb_recomp_11
+go
+insert tb_recomp_61
+select * from tb_recomp_11
+go
+
+
+-- psql
+analyze master_dbo.tb_recomp_11
```
2jungkook: `analyze` should be called in the verify scripts because in MVU tests, prepare scripts are called before MVU and the verify scripts are called after MVU. In that case, this test could be flaky.

contrib/babelfishpg_tsql/src/tsqlIface.cpp
```diff
@@ -6301,6 +6315,16 @@ makeExecuteProcedure(ParserRuleContext *ctx, std::string call_type)
...
 		ss << ::getFullText(func_proc_args);		
 	}
+	
+	if (exec_with_recompile)
+	{
+		/* 
+		 * When displaying the query plan, indicate that EXECUTE ... WITH RECOMPILE was specified.
+		 * This may be relevant for understanding the plan.
+		 */
+		ss << " /* EXECUTE WITH RECOMPILE */";
```
2jungkook, line 6347: The query string is used for all processing in the engine, including parsing, planning, and execution. To me this is kind of a hack method. I can think of two alternatives:
1. Use `EXEC ... WITH RECOMPILE` and ensure the later stage can handle this appropriately, or
2. Add this information only when showing query plans.

test/JDBC/expected/parallel_query/with_recompile-vu-verify.out
```diff
@@ -0,0 +1,3450 @@
...
+ * This fix for forcing plans for statements in the procedure body to be recompiled only
+ * applies to paramtrized statements. Non-parametrized stmts, like the 2nd select in tb_recomp_11, 
+ * will not be recompiled since choose_custom_plan() will always return false for those.
+ * The current fix can be extended to also force such stmts to be recompiled but that is 
+ * left for later.
+ 
+ * Not tested:
+ * - recursive procedure calls, since Babelfish does not show plans when a conditional statement 
+ *   is involved (manual testing has shown RECOMPILE works as expected, i.e. the recursive call is
+ *   handled independently from the higher-level call)
```
2jungkook, line 23: Please add a test using recursive procedure calls without conditional statement, e.g.,
```
create procedure p_recomp_12 @p int as
select * from tb_recomp_11 where c11 = @p  -- parametrized
select * from tb_recomp_11 where c11 = 1
exec p_recomp_11 1
```

test/JDBC/expected/parallel_query/with_recompile-vu-verify.out
```diff
@@ -0,0 +1,3450 @@
...
+ * - plans inside SQL functions or triggers (not shown by PG when called from a procedure)
```
2jungkook, line 24: Could you elaborate it more? I don't get it why we cannot add this test.

test/JDBC/expected/parallel_query/with_recompile-vu-verify.out
```diff
@@ -0,0 +1,3450 @@
...
+ 
+ */
+select '==== EXEC with RECOMPILE tests with plan_cache_mode=auto ======'  
+go
+~~START~~
+varchar
+==== EXEC with RECOMPILE tests with plan_cache_mode=auto ======
+~~END~~
```
2jungkook: This is not only bloating test outputs but also causing unnecessary query executions. Let's just use comments.

test/JDBC/expected/parallel_query/with_recompile-vu-verify.out
```diff
@@ -0,0 +1,3450 @@
...
+SELECT current_setting('enable_indexscan') as enable_indexscan_after
+go
+~~START~~
+text
+on
+~~END~~
+
+
+
+select '--- p_recomp_21: no generic plan generated first ------'
```
2jungkook: In the below, 
```
Index Cond: (c21 = $1)
```
means it generates a generic plan at the first execution.

test/JDBC/expected/parallel_query/with_recompile-vu-verify.out
```diff
@@ -0,0 +1,3450 @@
...
+        Workers Planned: 1
+        Single Copy: true
+        ->  Index Scan using tb_recomp_21_pkey on tb_recomp_21
+              Index Cond: (c21 = 1)
+~~END~~
+
+set babelfish_showplan_all off
+go
+
+select '--- tb_recomp_22: generate generic plan after repeated custom plans ------'
```
2jungkook: We don't need to do this because `plan_cache_mode = force_generic_plan`.

test/JDBC/expected/parallel_query/with_recompile-vu-verify.out
```diff
@@ -0,0 +1,3450 @@
...
+        ->  Bitmap Heap Scan on tb_recomp_23
+              Recheck Cond: (c23 = 1)
+              ->  Bitmap Index Scan on tb_recomp_23_pkey
+                    Index Cond: (c23 = 1)
+~~END~~
+
+set babelfish_showplan_all off
+go
+
+-- tb_recomp_23: generate generic plan after repeated custom plans
```
2jungkook: This is unnecessary.

test/JDBC/expected/parallel_query/with_recompile-vu-verify.out
```diff
@@ -0,0 +1,3450 @@
...
+
+
+
+/*
+ * These tests validate the T-SQL RECOMPILE behaviour for stored procedures.
+ * The approach is to demonstrate the recompilation by disabling index scans, 
+ * and observing a different plan being generated only when RECOMPILE is used.
+ * Note that when a cached plan is used, there is no good way to prove in these tests that 
+ * it was indeed a cached plan, as it could just as well be a newly generated plan that 
+ * happens to be the same as the previous plan. While developing this fix, the PG planner()
```
2jungkook, line 11: Actually, for a parametrized query, we can easily check whether recompilation happened:
```
Index Cond: (c12 = $1) --> generic plan, hence not recompiled
Index Cond: (c12 = 1) --> custom plan, hence recompiled.
```
Therefore, we don't need to check different plans (index vs bitmap) in this test. Based on this, what about separating test cases for parametrized and non-parametrized queries?

test/JDBC/expected/parallel_query/with_recompile-vu-verify.out
```diff
@@ -0,0 +1,3450 @@
...
+varchar
+--- tb_recomp_12: generate generic plan after repeated custom plans -----
+~~END~~
+
+
+-- tb_recomp_12: generate generic plan after repeated custom plans
+declare @i int = 6
+while @i > 0
+begin
+exec    p_recomp_12 1
```
2jungkook: Just wondering whether there is any way not to show the results because we are not interested in the results here.

test/JDBC/expected/parallel_query/with_recompile-vu-verify.out
```diff
@@ -0,0 +1,3450 @@
...
+        Workers Planned: 1
+        Single Copy: true
+        ->  Index Scan using tb_recomp_31_pkey on tb_recomp_31
+              Index Cond: (c31 = 1)
+~~END~~
+
+set babelfish_showplan_all off
+go
+
+select '--- tb_recomp_32: generate generic plan after repeated custom plans ------'
```
2jungkook: This is wrong because PG doesn't generate generic plans in `plan_cache_mode = force_custom_plan` for parametrized queries. Also, for non-parametrized queries, PG generates generic plans from the first execution.

test/JDBC/expected/parallel_query/with_recompile-vu-verify.out
```diff
@@ -0,0 +1,3450 @@
...
+1#!#0
+~~END~~
+
+~~START~~
+int#!#int
+1#!#0
+~~END~~
+
+
+-- generic plan is now created for tb_recomp_32 --> no further replanning according to PG approach
```
2jungkook: A generic plan is not created for a parametrized query.

test/JDBC/expected/parallel_query/with_recompile-vu-verify.out
```diff
@@ -0,0 +1,3450 @@
...
+        Workers Planned: 1
+        Single Copy: true
+        ->  Index Scan using tb_recomp_32_pkey on tb_recomp_32
+              Index Cond: (c32 = 1)
+~~END~~
+
+set babelfish_showplan_all off
+go
+
+select '--- EXEC without RECOMPILE: index scan (using cached plan) ----'
```
2jungkook: Cached plan is not used.

test/JDBC/expected/parallel_query/with_recompile-vu-verify.out
```diff
@@ -0,0 +1,3450 @@
...
+text
+off
+~~END~~
+
+
+
+
+
+/*
+ * These tests validate the T-SQL RECOMPILE behaviour for stored procedures.
```
2jungkook, line 6: Since we change `plan_cache_mode` GUC during executions, the following cases must be tested:
1. What happens if a user changes `plan_cache_mode` inside a procedure?
2. Is `plan_cache_mode` restored properly after each execution?
3. Is `plan_cache_mode` restored properly even if an exception is thrown?

test/JDBC/expected/parallel_query/with_recompile-vu-verify.out
```diff
@@ -0,0 +1,3450 @@
...
+
+select current_setting('enable_indexscan') as indexscan_on
+go
+~~START~~
+text
+on
+~~END~~
+
+
+select '--- EXEC without RECOMPILE: still creating recompiled plan on every execution: index scan for both statements ----'
```
2jungkook: I'd like to learn from your experience whether there were any needs for `exec without RECOMPILE` to re-enable plan caching even though the procedure was created by `with RECOMPILE` option.

test/JDBC/expected/parallel_query/with_recompile-vu-verify.out
2jungkook: I just saw you added a recursive procedure call case. Then, please just update this comment. 

Btw, now I'm wondering what happens in the following case:
```
create procedure outer_proc @p int as
    select set_config('plan_cache_mode', 'force_generic_plan', false)
    exec inner_proc1 @p
    exec inner_proc2 @p with recompile
go

exec outer_proc 1 with recompile -- expected behavior?
go

-- It should be 'force_generic_plan' even if inner_proc2 throws an exception.
SELECT current_setting('plan_cache_mode') 
go
```

contrib/babelfishpg_tsql/src/tsqlIface.cpp
robverschoor: we are just looking to determine whether RECOMPILE was specified.

robverschoor: code comment added

contrib/babelfishpg_tsql/src/tsqlIface.cpp
robverschoor: The comment is shown only when displaying query plans. It serves no other purpose. The backend parser ignores all comments, so we should not expect an issue from this comment. Passing down the information in a different way and modifying the displayed query plan test would be more complex; in contrast this is an elegant way of doing it.

test/JDBC/expected/parallel_query/with_recompile-vu-verify.out
robverschoor: Such a test case is already included: look for `p_recomp_51`
Note that this case is not recursive however; it is only recursive when the procedure calls itself, and that always requires a condition.

test/JDBC/expected/parallel_query/with_recompile-vu-verify.out
robverschoor: A recursive call must always contain some condition that stops the recursion, otherwise it will recurse forever. Whenever a condition is present, Babelfish will not show the query plan.

test/JDBC/expected/parallel_query/with_recompile-vu-verify.out
robverschoor: removed SELECT

test/JDBC/expected/parallel_query/with_recompile-vu-verify.out
robverschoor: Removed SELECTs

test/JDBC/expected/parallel_query/with_recompile-vu-verify.out
robverschoor: T-SQL does not provide such a mechanism. When a procedure is created with RECOMPILE, the only way to stop it from doing a recompile on every exec is to recreate the proc without RECOMPILE or run ALTER PROCEDURE without RECOMPILE. 
Deciding to use RECOMPILE is done after performance tuning and query plan debugging and reflects a conscious decision about overriding the regular query plan behaviour.

test/JDBC/input/with_recompile-vu-prepare.mix
robverschoor: moved to verify script

test/JDBC/expected/parallel_query/with_recompile-vu-verify.out
robverschoor: Changed search argument from 1 to 0, so no rows are returned; query plans remain the same.

test/JDBC/expected/parallel_query/with_recompile-vu-verify.out
robverschoor: removed multiple executions

test/JDBC/expected/parallel_query/with_recompile-vu-verify.out
robverschoor: removed

test/JDBC/expected/parallel_query/with_recompile-vu-verify.out
robverschoor: cleaned up test case

test/JDBC/expected/parallel_query/with_recompile-vu-verify.out
robverschoor: cleaned up test cases

test/JDBC/expected/parallel_query/with_recompile-vu-verify.out
robverschoor: cleaned up test cases

contrib/babelfishpg_tsql/src/tsqlIface.cpp
robverschoor: This is for the first-in-batch proc execution, without using the EXEC[UTE] keyword. Added test cases and clarifying comment.

contrib/babelfishpg_tsql/src/pl_exec-2.c
robverschoor: We are following the same logic here as for setting/resetting the "search_path" GUC. 
Given that the search_path is GUC is handled this way, we should assume that this code path can be followed and we should be able to properly restore the GUC in that case.

contrib/babelfishpg_tsql/src/pl_exec-2.c
robverschoor: We are following the same logic here as for setting/resetting the "search_path" GUC.  Also note that when old_plan_mode=NULL, the GUC is not restored.  
I do not see how we can add a test case for this branch. Given that the search_path is GUC is handled this way, we should assume that this code path can be followed and we should be able to properly restore the GUC in that case.

test/JDBC/expected/parallel_query/with_recompile-vu-verify.out
robverschoor: Separating the two types makes the tests more difficult to read IMHO. Also, we want to show that these types of statements are treated the right way when they are in the same procedure body: when separating them out you could argue we are not demonstrating what happens when they are together (i.e. whether they may influence each other).

test/JDBC/expected/parallel_query/with_recompile-vu-verify.out
robverschoor: update test case and comment

test/JDBC/expected/parallel_query/with_recompile-vu-verify.out
robverschoor: A test case was added which set the `plan_cache_mode` inside the proc, and another one where an exception is triggered after that.  In both cases, the `plan_cache_mode` is correctly reset to the value before the proc was called. Look for `p_recomp_71` and `p_recomp_72`.

test/JDBC/expected/parallel_query/with_recompile-vu-verify.out
```diff
@@ -0,0 +1,2991 @@
...
+~~END~~
+
+set babelfish_showplan_all on
+go
+-- cause exception in procedure:
+exec    p_recomp_72 0 with recompile    -- sets plan_cache_mode=force_generic_plan, divides by zero
+go
+~~ERROR (Code: 8134)~~
+
+~~ERROR (Message: division by zero)~~
```
2jungkook: This is not a proper test because `showplan_all` does not execute anything inside. However, it seems like it can detect division by zero even without executing it. The test should throw an exception during run-time instead of compile-time.

test/JDBC/expected/parallel_query/with_recompile-vu-verify.out
```diff
@@ -0,0 +1,2991 @@
...
+select set_config('plan_cache_mode', 'auto', false)
+go
+~~START~~
+text
+auto
+~~END~~
+
+set babelfish_showplan_all on
+go
+exec    p_recomp_71 0 with recompile    -- sets plan_cache_mode=force_generic_plan
```
2jungkook: Same here. We should execute it without `showplan_all`.

test/JDBC/expected/parallel_query/with_recompile-vu-verify.out
```diff
@@ -0,0 +1,2991 @@
...
+text
+auto
+~~END~~
+
+SELECT current_setting('enable_indexscan') as enable_indexscan_after
+go
+~~START~~
+text
+off
+~~END~~
```
2jungkook: The default value is `on`, hence we should set it to `on`.

contrib/babelfishpg_tsql/src/pl_exec-2.c
2jungkook: This `if` block is executed only when `old_plan_mode` is NULL, and in that case `plan_cache_mode` will be set to its default value (not the previous value, please read the comment about `set_config_option()` function in PG), which is wrong. 

And let me say my point again:
> If old_plan_mode is NULL, that means an exception occurred before setting the config, hence we don't need to restore the config.

I don't see a similar logic regarding setting "search_path" GUC neither. Also, test coverage report shows that 4 lines in this file are not covered, which is not good.

test/JDBC/expected/parallel_query/with_recompile-vu-verify.out
robverschoor: removed showplan

test/JDBC/expected/parallel_query/with_recompile-vu-verify.out
robverschoor: done

test/JDBC/expected/parallel_query/with_recompile-vu-verify.out
robverschoor: removed showplan

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -1275,7 +1297,7 @@ exec_stmt_exec(PLtsql_execstate *estate, PLtsql_stmt_exec *stmt)
 
 		if (stmt->is_cross_db)
 			SetCurrentRoleId(current_user_id, false);
-
+
```
2jungkook: Let's remove unnecessary diff.

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -1333,7 +1361,7 @@ exec_stmt_exec(PLtsql_execstate *estate, PLtsql_stmt_exec *stmt)
 
 	exec_eval_cleanup(estate);
 	SPI_freetuptable(SPI_tuptable);
-
+
```
2jungkook: unnecessary diff

contrib/babelfishpg_tsql/src/pltsql_utils.c
```diff
@@ -272,6 +273,14 @@ pltsql_createFunction(ParseState *pstate, PlannedStmt *pstmt, const char *queryS
 					location_cell = option;
 					pfree(defel);
 				}
+				else if (strcmp(defel->defname, "recompile") == 0)
+				{
+					/*
+					* CREATE PROCEDURE ... WITH RECOMPILE
+					* Record RECOMPILE in catalog
+					*/
```
2jungkook: style
```
/*
 * ...
 */
```

2jungkook (approved):

PR #2566, opened 2024-05-10
Fix crash in handle_where_clause_restargets_right

ahmed-shameem:
### Description

This commit fixes the crash in handle_where_clause_restargets_right and other functions which are used by sp_describe_undeclared_parameters. The crash was happening because we were throwing WARNING instead or ERROR. But, rather than throwing error, it's better to return 0 rows for the unsupported cases to avoid unnecessary application code errors.

We keep a flag which says whether we support the specific case or not. Based on that flag we return rows.

### Issues Resolved

BABEL-4869

* **Use case based -** YES

* **Arbitrary inputs -** YES 

* **Performance tests -**

* **Tooling impact -**

* **Client tests -**

### Check List


contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -994,34 +990,45 @@ sp_describe_undeclared_parameters_internal(PG_FUNCTION_ARGS)
...
-		raw_parsetree_list = pg_parse_query(parsedbatch);
-		if (list_length(raw_parsetree_list) != 1)
+		if (is_supported_case_sp_describe_undeclared_parameters)
 		{
-			ereport(ERROR,
-					(errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
-					 errmsg("Unsupported use case in sp_describe_undeclared_parameters")));
+			if (list_length(raw_parsetree_list) != 1)
+			{
+				is_supported_case_sp_describe_undeclared_parameters = false;
```
timchang514, line 1020: Can we move this up to where raw_parsetree_list is set? It would simplify the logic here

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -994,34 +990,45 @@ sp_describe_undeclared_parameters_internal(PG_FUNCTION_ARGS)
...
 		}
-		list_item = list_head(raw_parsetree_list);
-		parsetree = lfirst_node(RawStmt, list_item);
 
 		/*
 		 * Analyze the parsed statement to suggest types for undeclared
 		 * parameters
 		 */
-		switch (nodeTag(parsetree->stmt))
+		if (is_supported_case_sp_describe_undeclared_parameters)
```
timchang514: This can be combined with the above if statement?

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -1107,171 +1114,174 @@ sp_describe_undeclared_parameters_internal(PG_FUNCTION_ARGS)
...
-			case T_DeleteStmt:
-				values_list = list_make1(handle_where_clause_restargets_right(pstate, delete_stmt->whereClause, NIL));
-				break;
-			default:
-				ereport(ERROR,
-						(errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
-						 errmsg("Unsupported use case in sp_describe_undeclared_parameters")));
-				break;
-		}
+			relation_close(r, AccessShareLock);
```
timchang514, line 1156: Do we have guarantee here that relation was opened? What if is_supported_case_sp_describe_undeclared_parameters = false (ie nodeTag is default).

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -980,7 +969,14 @@ sp_describe_undeclared_parameters_internal(PG_FUNCTION_ARGS)
 		params = PG_ARGISNULL(1) ? NULL : text_to_cstring(PG_GETARG_TEXT_PP(1));
 
 		/* First, pass the batch to the ANTLR parser */
-		result = antlr_parser_cpp(batch);
+		if (batch)
+			result = antlr_parser_cpp(batch);
+		else
+		{
+			ereport(ERROR,
+					(errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
```
rishabhtanwar29: Probably `ERRCODE_UNDEFINED_FUNCTION` would be better.

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -107,6 +107,7 @@ char	   *sp_describe_first_result_set_view_name = NULL;
 
 bool		sp_describe_first_result_set_inprogress = false;
 char	   *orig_proc_funcname = NULL;
+static bool		is_supported_case_sp_describe_undeclared_parameters = true;
```
rishabhtanwar29: Nit: indentation seems a bit off.

test/JDBC/upgrade/latest/schedule
```diff
@@ -529,3 +529,4 @@ babel_test_int8_numeric_oper
 babel_test_int2_numeric_oper
 login_token-dep
 test_like_for_AI
+BABEL-4869
```
rishabhtanwar29, line 532: Why only latest schedule? I don't see anything in prepare file which is not supported in the previous versions.

ahmed-shameem: Updated

contrib/babelfishpg_tsql/src/procedures.c
ahmed-shameem: Updated

contrib/babelfishpg_tsql/src/procedures.c
ahmed-shameem: Updated

contrib/babelfishpg_tsql/src/procedures.c
ahmed-shameem: Updated

contrib/babelfishpg_tsql/src/procedures.c
ahmed-shameem: Updated

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -602,9 +603,8 @@ handle_bool_expr_rec(BoolExpr *expr, List *list)
```
rishabhtanwar29, line 603: Are we sure function `handle_bool_expr_rec` and other helper functions like this are only used by sp_describe_undeclared_parameter_internal function and not by any other? The problem is that if these helper functions are being used by other functions then we might unnecessarily set the value of `is_supported_case_sp_describe_undeclared_parameters` and then use an incorrect value for the rest of queries of that session.
At least, add a comment in the top of each such function to warn others about use of such functions.

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -1620,6 +1619,7 @@ sp_describe_undeclared_parameters_internal(PG_FUNCTION_ARGS)
 	}
 	else
 	{
+		is_supported_case_sp_describe_undeclared_parameters = true;
```
rishabhtanwar29, line 1656: How are we making sure that we don't leak the value of this static variable in case of an error? I don't see a try/catch block which would prevent this so be careful.

test/JDBC/expected/BABEL-4869-vu-verify.out
```diff
@@ -0,0 +1,253 @@
...
+
+-- Join Conditions Case: JOIN with a parameter in the condition
+EXEC sp_describe_undeclared_parameters 'SELECT * FROM babel_4869_vu_prepare_t1 JOIN babel_4869_vu_prepare_t2 ON babel_4869_vu_prepare_t1.Column2 = babel_4869_vu_prepare_t2.Column2 WHERE babel_4869_vu_prepare_t1.Column1 = @Param'
+GO
+~~START~~
+int#!#varchar#!#int#!#nvarchar#!#smallint#!#tinyint#!#tinyint#!#int#!#varchar#!#varchar#!#varchar#!#nvarchar#!#int#!#varchar#!#varchar#!#varchar#!#bit#!#bit#!#bit#!#bit#!#bit#!#varchar#!#int#!#int
+~~END~~
+
+
+-- Error Handling Case: Syntax error
```
rishabhtanwar29, line 41: Let's also add few happy scenarios at the end in which we correctly call this procedure and get correct result. This will also help in verifying if we are leaking the value of static variable we introduced.

ahmed-shameem: Few scenarios which returns rows are already added. Please check line 62, 73, 155

contrib/babelfishpg_tsql/src/procedures.c
ahmed-shameem: Updated these helper functions by adding one more parameter so that we know the call is indeed from sp_describe_undeclared_parameters_internal.

contrib/babelfishpg_tsql/src/procedures.c
ahmed-shameem: If we put the code block inside a TRY/CATCH block then we won't get some desired ERROR message hence did not take that approach. Rather set the value of is_supported_case_sp_describe_undeclared_parameters in the SRF_IS_FIRSTCALL for any new query. This means, even if something goes wrong, for the next query the parameter's value will be set again to true at the very beginning. So I think this should be able to handle any such abrupt error cases.

ahmed-shameem: Added TRY/CATCH block in latest rev.

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -933,695 +945,718 @@ sp_describe_undeclared_parameters_internal(PG_FUNCTION_ARGS)
...
+		}
+		else
+		{
+			/* Set the value again so that it is not left as false */
+			is_supported_case_sp_describe_undeclared_parameters = true;
+			SRF_RETURN_DONE(funcctx);
+		}
 	}
-	else
+	PG_CATCH();
```
rishabhtanwar29: Nit: this can actually be `PG_FINALLY()`.

rishabhtanwar29 (approved):
LGTM

contrib/babelfishpg_tsql/src/procedures.c
ahmed-shameem: updated in latest rev.

timchang514 (approved):
LGTM

shardgupta (approved):

PR #2570, opened 2024-05-13
Add typmod validation in [VAR]BINARY input function

rishabhtanwar29:
### Description
(Var)binary input function does not check if input string is bigger than the
desired typmod. This is problematic as it is possible to insert a binary value
larger than typmod into a column. Although, the issue is not there for normal
inserts into a table as there exists typmod coercion functions for binary and
varbinary datatypes which are always called along with input function to adjust
the value length according to given typmod.

But the issue exists for logical replication and dump/restore in which the value
is coming from a remote server possibly with a different typmod. In these cases
typmod coercion functions are not called so no checks are performed on size of
input value.
To fix the issue, this commit adds a check in the (var)binary input function to
validate input string with typmod and throw error if the value is larger than typmod
similar to (var)char datatype.

Task: BABEL-4956

* **Use case based -**

* **Boundary conditions -**

* **Arbitrary inputs -**

* **Negative test cases -**

* **Performance tests -**

* **Tooling impact -**

* **Client tests -**

### Check List

### DIFF
--- a/contrib/babelfishpg_common/src/varbinary.c
+++ b/contrib/babelfishpg_common/src/varbinary.c
@@ -196,6 +196,14 @@ varbinaryin(PG_FUNCTION_ARGS)
 		 */
 		int			bc = (len - 1) / 2 + VARHDRSZ;	/* maximum possible length */
 
+		if (typmod >= (int32) VARHDRSZ && bc > typmod)
+		{
+			ereport(ERROR,
+					(errcode(ERRCODE_STRING_DATA_RIGHT_TRUNCATION),
+					 errmsg("String or binary data would be truncated.\n"
+							"The statement has been terminated.")));
+		}
+
 		result = palloc(bc);
 		bc = babelfish_hex_decode_allow_odd_digits(inputText + 2, len - 2, VARDATA(result));
 		SET_VARSIZE(result, bc + VARHDRSZ); /* actual length */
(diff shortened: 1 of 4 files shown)


contrib/babelfishpg_common/src/varbinary.c
```diff
@@ -196,6 +196,28 @@ varbinaryin(PG_FUNCTION_ARGS)
...
 		int			bc = (len - 1) / 2 + VARHDRSZ;	/* maximum possible length */
 
+		if (typmod >= (int32) VARHDRSZ && bc > typmod)
+		{
+			/* Verify that extra bytes are zeros, and clip them off */
+			char	*temp_result;
+			size_t	i;
+
+			temp_result = palloc0(bc);
+			bc = babelfish_hex_decode_allow_odd_digits(inputText + 2, len - 2, temp_result);
```
Deepesh125, line 206: If input length is less than typmod then we are doing the computation again on line 221.

rishabhtanwar29: Actually this is intentional as the first call is just for the verification purpose here in which we have to transform the whole input string to bytea to determine if we can truncate it. Then the second call will do the actual transformation with the correct length. This is what a typmod cast function will also do, which is that first the input function will convert the whole string to bytea and then typmod cast function will re-do it with a different length.

Deepesh125 (approved):

PR #2584, opened 2024-05-16
Wrong column type in sys.sql_modules

shalinilohia50:
### Description

Currently, `definition` column in the view sys.sql_modules stores upto 4000 characters. SSMS scripting uses sys.sql_modules view to show the definition of the procedure, and the procedure definition is truncated to 4000 characters if the actual procedure definition is more than 4K. 

To avoid this incorrect definition, We can directly use `sys.nvarchar` and there is no need to specify the tymod of 4000 in case of sys.nvarchar for the definition column i.e., `sys.nvarchar(4000)`.

Fixed the definition column for the views which should use nvarchar(max): sql_modules, system_sql_modules, all_sql_modules.

### Issues Resolved

Task: [BABEL-4870](https://jira.rds.a2z.com/browse/BABEL-4870)

* **Use case based -** Added test in DDL test framework. Also, testing using SSMS.

* **Boundary conditions -**

* **Arbitrary inputs -**

* **Negative test cases -**

* **Performance tests -**

* **Tooling impact -**

* **Client tests -**


test/JDBC/input/views/sys-sql_modules-vu-verify.mix
```diff
@@ -62,6 +62,20 @@ FROM sys.sql_modules
...
+    definition,
+    uses_ansi_nulls,
+    uses_quoted_identifier,
+    is_schema_bound,
+    uses_database_collation,
+    is_recompiled,
+    null_on_null_input,
+    execute_as_principal_id,
+    uses_native_compilation
+FROM sys.all_sql_modules
```
roshan0708: Shouldn't this SELECT query target `sys.sql_modules` instead of `sys.all_sql_modules`?

basasairohan (approved):

test/JDBC/input/views/sys-sql_modules-vu-verify.mix
shalinilohia50: Updated.

roshan0708 (approved):

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--3.5.0--3.6.0.sql
```diff
@@ -493,7 +493,107 @@ GRANT EXECUTE ON PROCEDURE sys.sp_babelfish_configure(
...
+    ,CAST(t1.is_recompiled as sys.bit)
+    ,CAST(t1.null_on_null_input as sys.bit)
+    ,CAST(t1.execute_as_principal_id as int)
+    ,CAST(t1.uses_native_compilation as sys.bit)
+FROM sys.all_sql_modules_internal t1;
+GRANT SELECT ON sys.all_sql_modules TO PUBLIC;
+
+ALTER VIEW sys.system_sql_modules RENAME TO system_sql_modules_deprecated_3_6_0;
+
+CREATE OR REPLACE VIEW sys.system_sql_modules AS
```
thephantomthief, line 557: We should add tests for sys.system_sql_modules too

shalinilohia50: We don't store the definition of system objects, so we can't really have tests for it now.

thephantomthief (approved):

test/JDBC/expected/sys-all_sql_modules-vu-prepare.out
```diff
@@ -38,6 +38,13 @@ CREATE PROC sys_all_sql_modules_vu_prepare_p1 AS
 SELECT 1
 GO
 
+CREATE PROCEDURE sys_all_sql_modules_more_than_4k
```
Deepesh125, line 41: Is there any existing dependency test cases for these views? sql_modules, system_sql_modules, all_sql_modules and all_sql_modules_internal?

Deepesh125 (changes requested):
Have we tested object explorer with these changes?

shalinilohia50:
> Have we tested object explorer with these changes?

Yes, I have tested with SSMS

test/JDBC/expected/sys-all_sql_modules-vu-prepare.out
shalinilohia50: Yes, we have that in place.

shalinilohia50: sys-sql_modules-dep-*, sys-all_sql_modules-dep-* etc.

Deepesh125: None of these files depends on "definition" column, one being updated by us. We should have dependency object such that it uses "definition" column

shalinilohia50: Added such tests which use definition column and uses more than 4k characters

test/JDBC/expected/sys-all_sql_modules-before-14_6-dep-vu-verify.out
```diff
@@ -0,0 +1,31 @@
...
+int
+1
+~~END~~
+
+
+SELECT * FROM sys_all_sql_modules_dep_vu_prepare_v1;
+GO
+~~START~~
+int#!#bit#!#nvarchar
+<NULL>#!#0#!#
```
Deepesh125, line 37: why is definition empty?

test/JDBC/expected/sys-all_sql_modules-dep-vu-prepare.out
```diff
@@ -6,18 +6,36 @@ BEGIN
...
+GO
+
 CREATE VIEW sys_all_sql_modules_dep_vu_prepare_v1 AS
 SELECT
     execute_as_principal_id,
-    uses_native_compilation
+    uses_native_compilation,
+    definition
 FROM sys.all_sql_modules
 WHERE object_id = OBJECT_ID('sys_all_sql_modules_dep_vu_prepare_f1')
```
Deepesh125, line 40: we should also select definition of sys_all_sql_modules_dep_more_than_4k

test/JDBC/expected/sys-all_sql_modules-vu-prepare.out
```diff
@@ -38,6 +38,13 @@ CREATE PROC sys_all_sql_modules_vu_prepare_p1 AS
...
+END
+GO
+
```
Deepesh125, line 47: there's no dependent objects

test/JDBC/expected/sys-sql_modules-dep-vu-prepare.out
```diff
@@ -6,18 +6,36 @@ BEGIN
...
+GO
+
 CREATE VIEW sys_sql_modules_dep_vu_prepare_v1 AS
 SELECT
     execute_as_principal_id,
-    uses_native_compilation
+    uses_native_compilation,
+    definition
 FROM sys.sql_modules
 WHERE object_id = OBJECT_ID('sys_sql_modules_dep_vu_prepare_f1')
```
Deepesh125, line 40: select sys_sql_modules_dep_more_than_4k as well

test/JDBC/expected/sys-sql_modules-vu-prepare.out
```diff
@@ -38,6 +38,13 @@ CREATE PROC sys_sql_modules_vu_prepare_p1 AS
...
+END
+GO
+
```
Deepesh125, line 47: no dependency on sys.sql_modules

test/JDBC/expected/sys-system_sql_modules-dep-vu-prepare.out
```diff
@@ -17,5 +17,5 @@ END
 GO
 
 CREATE VIEW sys_system_sql_modules_dep_vu_prepare_v1 AS
-    SELECT uses_ansi_nulls FROM sys.system_sql_modules WHERE object_id = OBJECT_ID('sys.fn_listextendedproperty')
+    SELECT uses_ansi_nulls, definition FROM sys.system_sql_modules WHERE object_id = OBJECT_ID('sys.fn_listextendedproperty')
```
Deepesh125, line 20: include object whosse body is of length more than 4000

test/JDBC/expected/sys-system_sql_modules-dep-vu-verify.out
```diff
@@ -4,8 +4,8 @@ GO
 EXEC sys_system_sql_modules_dep_vu_prepare_p1
 GO
 ~~START~~
-bit
-1
+bit#!#nvarchar
+1#!#
```
Deepesh125, line 8: why is definition empty here? same for sys_system_sql_modules_dep_vu_prepare_v1?

Deepesh125 (changes requested):

test/JDBC/expected/sys-all_sql_modules-before-14_6-dep-vu-verify.out
shalinilohia50: Before 14_6, the definition column used to be empty.

test/JDBC/expected/sys-system_sql_modules-dep-vu-prepare.out
shalinilohia50: We don't support object definition of system objects. That would be NULL.

test/JDBC/expected/sys-system_sql_modules-dep-vu-verify.out
shalinilohia50: We don't support definition of system objects. That would be NULL.

test/JDBC/expected/sys-all_sql_modules-vu-prepare.out
shalinilohia50: Dependency tests are added in `sys-all_sql_modules-dep*` files.

test/JDBC/expected/sys-sql_modules-dep-vu-prepare.out
shalinilohia50: When we execute a view with object_id(sys_all_sql_modules_dep_more_than_4k), we get the truncated definition.
It has to do with the fact that while creation, the definition column in `sys.sql_modules` has `nvarchar(4000)` and that is not changed after upgrade in a view. So, we get the truncated definition only. In case of procedures, datatype of column depends on the execution time and we don't see the problem.

test/JDBC/expected/sys-all_sql_modules-dep-vu-prepare.out
shalinilohia50: When we execute this view with object_id(sys_all_sql_modules_dep_more_than_4k), we get the truncated definition.
It has to do with the fact that while creation, the definition column in `sys.sql_modules` has `nvarchar(4000)` and that is not changed after upgrade in a view. So, we get the truncated definition only. In case of procedures, datatype of column depends on the execution time and we don't see the problem.

Deepesh125 (approved):

PR #2587, opened 2024-05-20
Support Partition Function and Partition Scheme in Babelfish

sumitj824:
### Description

This commit introduces support for Partition Functions and Partition Schemes in Babelfish. To manage the metadata related to these features, the following system catalogs have been added:

1. sys.babelfish_partition_function: This catalog stores the metadata for Partition Functions, which define the partitioning logic based on boundary values and option.

2. sys.babelfish_partition_scheme: This catalog stores the metadata for Partition Schemes, which define the partition function mapping for a partitioned table.

3. sys.babelfish_partition_depend: This catalog tracks the dependencies between Partition Schemes and the partitioned tables created using those schemes.

Additionally, this commit adds support for the following system views related to partition functions and schemes:
1. sys.partition_schemes: This view provides information about the defined partition schemes in the database.
2. sys.partition_functions: This view provides information about the defined partition functions in the database.
3. sys.partition_range_values: This view provides the boundary values defined for each range partition function.
4. sys.partition_parameters: This view provides information about the parameters used by partition functions.

Engine PR: https://github.com/babelfish-for-postgresql/postgresql_modified_for_babelfish/pull/379

### Issues Resolved

Task: BABEL-4910, BABEL-4967

* **Use case based -** Yes

* **Arbitrary inputs -** Yes

* **Performance tests -** NA

* **Tooling impact -** NA

* **Client tests -** NA

### Check List


contrib/babelfishpg_tsql/antlr/TSqlParser.g4
```diff
@@ -1172,7 +1172,12 @@ alter_partition_scheme
     ;
 
 create_partition_scheme
-    : CREATE PARTITION SCHEME partition_scheme_name=id AS PARTITION partition_function_name=id ALL? TO LR_BRACKET id (COMMA id)* RR_BRACKET
+    : CREATE PARTITION SCHEME partition_scheme_name=id AS PARTITION partition_function_name=id ALL? TO LR_BRACKET filegroups (COMMA filegroups)* RR_BRACKET
+    ;
+
+filegroups
```
Deepesh125: This is confusing, filegroups and id is separate thing. Can we update the above grammar to something like `(COMMA filegroups | id)*`?

contrib/babelfishpg_tsql/sql/object_definition_tsql.sql
```diff
@@ -11,4 +11,51 @@ CREATE TABLE sys.babelfish_view_def (
 );
 GRANT SELECT ON sys.babelfish_view_def TO PUBLIC;
 
-SELECT pg_catalog.pg_extension_config_dump('sys.babelfish_view_def', '');
\ No newline at end of file
+-- BBF_PARTITION_FUNCTION
+-- This catalog stores the metadata of partition funtions.
+CREATE TABLE sys.babelfish_partition_function
+(
+  dbid SMALLINT NOT NULL,
```
Deepesh125: Can we add comment on each field just like we did in LLD

contrib/babelfishpg_tsql/sql/object_definition_tsql.sql
```diff
@@ -11,4 +11,51 @@ CREATE TABLE sys.babelfish_view_def (
...
+CREATE SEQUENCE sys.babelfish_partition_scheme_seq START 2 MAXVALUE 2147483647 CYCLE;
+
+-- BBF_PARTITION_DEPEND
+-- This catalog tracks the dependecy b/w partition scheme and partitioned tables created using that.
+CREATE TABLE sys.babelfish_partition_depend
+(
+  dbid SMALLINT NOT NULL,
+  scheme_id INT NOT NULL,
+  schema_name sys.NVARCHAR(128) NOT NULL COLLATE sys.database_default,
+  table_name sys.NVARCHAR(128) NOT NULL COLLATE sys.database_default,
```
Deepesh125: NVARCHAR(128) might not be enough to store schema qualified table name. schema name and table name could be of the length 64. And total length combined could be 64 + 64 + 1 (for the separator)

contrib/babelfishpg_tsql/sql/sys_views.sql
```diff
@@ -2479,6 +2479,69 @@ SELECT
 WHERE FALSE;
 GRANT SELECT ON sys.dm_hadr_database_replica_states TO PUBLIC;
 
+CREATE OR REPLACE VIEW sys.partition_functions AS
+SELECT
+  CAST(partition_function_name as sys.sysname) as name,
```
Deepesh125: In Babelfish, sysname is the alias of varchar(128) instead of nvarchar. So let's use nvarchar(128) instead of sysname to avoid rework in future

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -4029,3 +4035,310 @@ exec_stmt_fulltextindex(PLtsql_execstate *estate, PLtsql_stmt_fulltextindex *stm
...
+	return 0;
+}
+
+/*
+ * Allows only login that is either db owner or member of sysadmin.
+ */
+static void 
+perform_permission_check(char *name, bool is_create, bool is_function)
+{
+	char		*dbname = get_cur_db_name();
```
Deepesh125, line 4100: should free dbname

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -4802,6 +4805,44 @@ fill_missing_values_in_copyfrom(Relation rel, Datum *values, bool *nulls)
...
+		AttrNumber	attnum;
+
+		attnum = (AttrNumber) attnameAttNum(rel, "scheme_id", false);
+		Assert(attnum != InvalidAttrNumber);
+
+		if (nulls[attnum - 1])
+		{
+			int32 scheme_id = get_available_partition_scheme_id();
+			values[attnum - 1] = Int32GetDatum(scheme_id);
+			nulls[attnum - 1] = false;
```
Deepesh125: Please add a comment to explain why is this needed

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -4029,3 +4035,309 @@ exec_stmt_fulltextindex(PLtsql_execstate *estate, PLtsql_stmt_fulltextindex *stm
...
+				ereport(ERROR,
+					(errcode(ERRCODE_INVALID_PARAMETER_VALUE),
+						errmsg("Could not implicitly convert range values type specified at ordinal %d to partition function parameter type.",
+							i+1)));
+			}
+			PG_END_TRY();
+		}
+
+		/* find comparator operator for input type, which will be used during the sorting */
+		operator = compatible_oper(NULL, list_make1(makeString("<")), typ->typoid, typ->typoid, false, -1);
```
Deepesh125: 1. Every data type will have in built comparator, we should try to use that and should update logic of tsql_compare_values accordingly. 
2. what if this operator is not available for given data type? what if partition is created over the user defined data types? we should find base data type of typ->typoid and should use it to find the operator

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -4029,3 +4035,309 @@ exec_stmt_fulltextindex(PLtsql_execstate *estate, PLtsql_stmt_fulltextindex *stm
...
+		MemoryContext		cur_ctxt = CurrentMemoryContext;
+		LOCAL_FCINFO(fcinfo, 1);
+
+		/* check if given name is exceeding the allowed limit */
+		if(strlen(partition_function_name) > 128)
+		{
+			ereport(ERROR, 
+				(errcode(ERRCODE_INVALID_PARAMETER_VALUE),
+					errmsg("The identifier that starts with '%.128s' is too long. Maximum length is 128.", partition_function_name)));
+		}
```
Deepesh125: I dont think we will ever hit this here. Have we added any test which executes this code?

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -4453,3 +4608,691 @@ rename_tsql_db(char *old_db_name, char *new_db_name)
...
+				errmsg("Partition function '%s' is being used by one or more partition schemes.", partition_function_name)));
+	}
+}
+
+/*
+ * partition_function_exists
+ * 		checks if provided partition function name exists in database
+ */
+bool
+partition_function_exists(char *partition_function_name)
```
Deepesh125: const char *

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -4453,3 +4608,691 @@ rename_tsql_db(char *old_db_name, char *new_db_name)
...
+	ScanKeyInit(&scanKey[0],
+				Anum_bbf_partition_function_dbid,
+				BTEqualStrategyNumber, F_INT2EQ,
+				Int16GetDatum(dbid));
+
+	ScanKeyEntryInitialize(&scanKey[1], 0,
+				Anum_bbf_partition_function_name,
+				BTEqualStrategyNumber, InvalidOid,
+				tsql_get_server_collation_oid_internal(false),
+				F_TEXTEQ, CStringGetTextDatum(partition_function_name));
```
Deepesh125, line 4983: partition_function_name is of sys.nvarchar(128), are we sure that TEXTEQ is right choice here?

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -4453,3 +4608,691 @@ rename_tsql_db(char *old_db_name, char *new_db_name)
...
+	
+	/* scan using index */
+	scan = systable_beginscan(rel, get_bbf_partition_function_idx_oid(),
+					false, NULL, 2, scanKey);
+	
+	tuple = systable_getnext(scan);
+	if (HeapTupleIsValid(tuple))
+	{
+		bool isnull;
+		partition_function_id = DatumGetInt32(heap_getattr(tuple, Anum_bbf_partition_function_id,
```
Deepesh125: just add safety check that isnull should not be NULL

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -4029,3 +4035,309 @@ exec_stmt_fulltextindex(PLtsql_execstate *estate, PLtsql_stmt_fulltextindex *stm
...
+			tsql_typename = text_to_cstring(DatumGetTextPP(tsql_type_datum));
+		}
+
+		/*
+		 * check if datatype is supported or not, if tsql_typename is NULL
+		 * then it implies that typename corresponds to UDT
+		 */
+		if(!tsql_typename || (strcmp(tsql_typename, "text") == 0) || (strcmp(tsql_typename, "ntext") == 0)
+			|| (strcmp(tsql_typename, "image") == 0) || (strcmp(tsql_typename, "xml") == 0) || (strcmp(tsql_typename, "geometry") == 0)
+			|| (strcmp(tsql_typename, "geography") == 0) || (strcmp(tsql_typename, "timestamp") == 0) )
```
Deepesh125: we already have some helper function such as is_tsql_ntext_datatype, we should try to use that

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -4029,3 +4035,309 @@ exec_stmt_fulltextindex(PLtsql_execstate *estate, PLtsql_stmt_fulltextindex *stm
...
+			PG_TRY();
+			{
+				values[i] = exec_cast_value(estate, val, &isnull,
+								valtype, valtypmod,
+								typ->typoid, typ->atttypmod);
+			}
+			PG_CATCH();
+			{
+				MemoryContext err_ctx;
+				err_ctx = MemoryContextSwitchTo(cur_ctxt);
```
Deepesh125: why memory context changes are required?

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -4029,3 +4035,309 @@ exec_stmt_fulltextindex(PLtsql_execstate *estate, PLtsql_stmt_fulltextindex *stm
...
+
+		/* find comparator operator for input type, which will be used during the sorting */
+		operator = compatible_oper(NULL, list_make1(makeString("<")), typ->typoid, typ->typoid, false, -1);
+		oprcode = oprfuncid(operator);
+		ReleaseSysCache(operator);
+
+		/* set the function oid of operator in tsql comparator context */
+		cxt.function_oid = oprcode;
+		cxt.colloid = tsql_get_server_collation_oid_internal(false);
+		cxt.contains_duplicate = false;
```
Deepesh125: where are we validating this? do we have test cases for the same?

contrib/babelfishpg_tsql/sql/object_definition_tsql.sql
```diff
@@ -11,4 +11,51 @@ CREATE TABLE sys.babelfish_view_def (
...
+-- BBF_PARTITION_FUNCTION
+-- This catalog stores the metadata of partition funtions.
+CREATE TABLE sys.babelfish_partition_function
+(
+  dbid SMALLINT NOT NULL,
+  function_id INT NOT NULL UNIQUE,
+  partition_function_name sys.NVARCHAR(128) NOT NULL COLLATE sys.database_default,
+  input_type sys.sysname,
+  partition_option sys.bit,
+  range_values sys.sql_variant[],
```
Deepesh125: IIUC, there could be at max 15000 partitions, can we have some check constraint here to ensure the same?

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -4453,3 +4608,691 @@ rename_tsql_db(char *old_db_name, char *new_db_name)
...
+	table_close(rel, AccessShareLock);
+	return is_used;
+}
+
+/*
+ * add_entry_to_bbf_partition_function
+ *		Add a new entry to the sys.babelfish_partition_function catalog table.
+ */
+void
+add_entry_to_bbf_partition_function(char *partition_function_name, char *typname,
```
Deepesh125: const char * everywhere

Deepesh125 (changes requested):

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -4453,3 +4608,691 @@ rename_tsql_db(char *old_db_name, char *new_db_name)
...
+
+	/* Build a tuple to insert */
+	new_record[Anum_bbf_partition_function_dbid - 1] = Int16GetDatum(dbid);
+	new_record[Anum_bbf_partition_function_id - 1] = Int32GetDatum(partition_function_id);
+	new_record[Anum_bbf_partition_function_name - 1] = CStringGetTextDatum(partition_function_name);
+	new_record[Anum_bbf_partition_function_input_type - 1] =  CStringGetTextDatum(typname);
+	new_record[Anum_bbf_partition_function_partition_option - 1] = BoolGetDatum(partition_option);
+	new_record[Anum_bbf_partition_function_range_values - 1] = PointerGetDatum(values);
+	new_record[6] = TimestampGetDatum(GetSQLLocalTimestamp(3));
+	new_record[7] = TimestampGetDatum(GetSQLLocalTimestamp(3));
```
Deepesh125: will these two calls to GetSQLLocalTimestamp return the same value? We should call  GetSQLLocalTimestamp once and use the same value at both places

contrib/babelfishpg_tsql/sql/object_definition_tsql.sql
sumitj824: We are storing schema name and table name in separate column instead of schema qualified table names.

contrib/babelfishpg_tsql/src/pl_exec-2.c
sumitj824: We will hit this here only, because we are not truncating the identifier name of partition specifier during ANTLR parsing. I will add the testcase for the same.

contrib/babelfishpg_tsql/src/pl_exec-2.c
sumitj824: We are changing the context to free the allocated memory in previous context and raise approariate error message.

contrib/babelfishpg_tsql/src/pl_exec-2.c
Deepesh125: Ignore this

contrib/babelfishpg_tsql/antlr/TSqlParser.g4
sumitj824: This mean that filegroup name can be either PRIMARY or user specified. I will add comment to clear the confusion.

contrib/babelfishpg_tsql/sql/object_definition_tsql.sql
sumitj824: We are already checking the partition limit before inserting into catalog so check will be redundant here but yes to avoid user direct manipulation of these catalogs, I will add that check here also.

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -4453,3 +4608,691 @@ rename_tsql_db(char *old_db_name, char *new_db_name)
...
+
+	/* Build a tuple to insert */
+	MemSet(new_record, 0, sizeof(new_record));
+	MemSet(new_record_nulls, false, sizeof(new_record_nulls));
+
+	new_record[Anum_bbf_partition_scheme_dbid - 1] = Int16GetDatum(dbid);
+	new_record[Anum_bbf_partition_scheme_id - 1] = Int32GetDatum(partition_scheme_id);
+	new_record[Anum_bbf_partition_scheme_name - 1] = CStringGetTextDatum(partition_scheme_name);
+	new_record[Anum_bbf_partition_scheme_func_name - 1] = CStringGetTextDatum(partition_function_name);
+	new_record[4] = BoolGetDatum(next_used);
```
Deepesh125: Let's use macro here also just like other fields

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -4453,3 +4608,691 @@ rename_tsql_db(char *old_db_name, char *new_db_name)
...
+	/* raise error if it doesn't exists in database */
+	if (!partition_scheme_exists)
+	{
+		ereport(ERROR, 
+			(errcode(ERRCODE_UNDEFINED_OBJECT), 
+				errmsg("Cannot drop the partition scheme '%s', because it does not exist or you do not have permission.", partition_scheme_name)));
+	}
+
+	/* raise error if there are dependent tables on it */
+	if(has_dependent_objects)
```
Deepesh125: Need to fix indentation, at other places also

contrib/babelfishpg_tsql/src/pl_exec-2.c
sumitj824: 1. yes, that can be other approach where we don't rely on `<` operator for comparison and direct use cmp function defined for each type. I will incorporate the same.
2. Only TSQL types are supported so there operator will always exists for given datatype. UDTs are not supported in TQL, please check the datatype check done in start.

contrib/babelfishpg_tsql/src/tsqlNodes.h
```diff
@@ -52,8 +52,10 @@ typedef enum pltsql_stmt_type
 	PLTSQL_STMT_CHANGE_DBOWNER,
 	PLTSQL_STMT_DBCC,
 	PLTSQL_STMT_ALTER_DB,
-	PLTSQL_STMT_FULLTEXTINDEX
-	PLTSQL_STMT_GRANTSCHEMA
+	PLTSQL_STMT_FULLTEXTINDEX,
+	PLTSQL_STMT_GRANTSCHEMA,
+	PLTSQL_STMT_PARTITION_FUNCTION,
+	PLTSQL_STMT_PARTITION_SCHEME
```
Deepesh125: why is this change required? we can remove this whole enum because it is wrapped under `#if 0 ... #endif`

contrib/babelfishpg_tsql/src/tsqlIface.cpp
```diff
@@ -1842,6 +1862,12 @@ class tsqlBuilder : public tsqlCommonMutator
 			clear_rewritten_query_fragment();
 			return;
 		}
+		if (ctx->create_partition_function() || ctx->drop_partition_function()
+			 || ctx->create_partition_scheme() || ctx->drop_partition_scheme())
+		{
+			clear_rewritten_query_fragment();
+			return;
```
Deepesh125: can partition function or scheme be created through function body?

contrib/babelfishpg_tsql/src/tsqlIface.cpp
```diff
@@ -7071,6 +7097,80 @@ getDatabaseSchemaAndTableName(TSqlParser::Table_nameContext* tctx)
...
 }
 
+PLtsql_stmt *
+makeCreatePartitionFunction(TSqlParser::Create_partition_functionContext *ctx)
+{
+	PLtsql_stmt_partition_function *stmt = (PLtsql_stmt_partition_function *) palloc0(sizeof(PLtsql_stmt_partition_function));
+	std::string typeStr = ::getFullText(ctx->data_type());
+	PLtsql_type *type = parse_datatype(typeStr.c_str(), 0);
+	
+	stmt->is_right = true;
```
Deepesh125: we should add TODO comment here to support left boundary option

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -4453,3 +4608,691 @@ rename_tsql_db(char *old_db_name, char *new_db_name)
...
+	{
+		id = nextval_internal(get_bbf_partition_function_seq_oid(), false);
+		if (start == 0)
+			start = id;
+		else if (start == id)
+		{
+			loop_completed = true;
+			break;
+		}
+	} while (partition_function_exists_with_id(id));
```
Deepesh125: can we come up with better solution here? something using lock may be

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -4453,3 +4608,691 @@ rename_tsql_db(char *old_db_name, char *new_db_name)
...
+ * get_available_partition_function_id
+ * 		Generates new ids for partition scheme.
+ * 		1. To guard against race conditions for IDs, we check if the 
+ * 		   ID generated by sequence is used by an existing partition function.
+ * 		2. To guard rare case where all possible sequence values have been exhausted
+ * 		   and the sequence wraps around, we will loop through the entire range of
+ * 		   sequence values and on loop completion, we should bail out.
+ */
+int32
+get_available_partition_function_id(void)
```
KushaalShroff, line 4716: Can we use LOCKTAG to avoid race condition?

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -4453,3 +4608,691 @@ rename_tsql_db(char *old_db_name, char *new_db_name)
...
+		if (start == 0)
+			start = id;
+		else if (start == id)
+		{
+			loop_completed = true;
+			break;
+		}
+	} while (partition_function_exists_with_id(id));
+
+	if (loop_completed)
```
KushaalShroff: This is a race condition, you can try by attach gdb on one process here and in another process you will fetch the same id since partition_function_exists_with_id will return false (process 1 has not been committed yet)

KushaalShroff: I think we can use LOCKTAG to avoid

KushaalShroff (changes requested):

Deepesh125 (changes requested):

contrib/babelfishpg_tsql/src/catalog.c
sumitj824: yes, it will work even for length greater than 64.

contrib/babelfishpg_tsql/src/catalog.c
sumitj824: these columns are defined as NOT NULL in the catalog, safety check will be redundant.

contrib/babelfishpg_tsql/src/catalog.c
sumitj824: yes, `GetSQLLocalTimestamp` will return the same value for each calls because it returns the transaction start timestamp which will remain same.

contrib/babelfishpg_tsql/src/tsqlIface.cpp
sumitj824: No, similar to other create commands it is not allowed inside procedure.

contrib/babelfishpg_tsql/src/catalog.c
sumitj824: `nextval_internal` itself handles this, internally it uses lock on sequence to generate unique number for each function call, even if it is called during parallel mode. Also, if the transaction is not committed the sequence value will be counted as committed.

contrib/babelfishpg_common/src/typecode.c
```diff
@@ -298,6 +301,20 @@ lookup_tsql_datatype_oid(const char *typename)
...
 
+static Oid
+lookup_pg_datatype_oid(const char *typename)
+{
+	Oid			nspoid;
+	Oid			typoid;
+
+	nspoid = get_namespace_oid("pg_catalog", true);
+	if (nspoid == InvalidOid)
+		return InvalidOid;
```
rishabhtanwar29: Can directly use PG_CATALOG_NAMESPACE

contrib/babelfishpg_common/src/typecode.c
```diff
@@ -400,6 +433,17 @@ is_tsql_timestamp_datatype(Oid oid)
...
+bool
+is_tsql_xml_datatype(Oid oid)
+{
+	if (tsql_xml_oid == InvalidOid)
+	{
+		/* lookup for xml datatype in pg_catalog namespace */
+		tsql_xml_oid = lookup_pg_datatype_oid("xml");
+	}
+	return tsql_xml_oid == oid;
+}
```
rishabhtanwar29: Are you just looking for `XMLOID`? ;-)
Also, why to call it `tsql_xml` if it is pg_catalog.xml we are searching for.

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -4453,3 +4610,691 @@ rename_tsql_db(char *old_db_name, char *new_db_name)
...
 		CommitTransactionCommand();
 }
+
+/*
+ * partition_function_exists_with_id
+ * 		Helper function to generate new ids for partition function,
+ * 		it check if provided id is already in use.
+ */
+static bool
+partition_function_exists_with_id(int32 id)
```
rishabhtanwar29: `partition_function_id_exists` seems a better name.

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -4453,3 +4610,691 @@ rename_tsql_db(char *old_db_name, char *new_db_name)
 	if (!xactStarted)
 		CommitTransactionCommand();
 }
+
+/*
+ * partition_function_exists_with_id
+ * 		Helper function to generate new ids for partition function,
+ * 		it check if provided id is already in use.
```
rishabhtanwar29: 1. Please remove the first line, this function doesn't generate the ids, second line is enough.
2. Nit: Returns true if provided function id is in use, false otherwise.

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -4453,3 +4610,691 @@ rename_tsql_db(char *old_db_name, char *new_db_name)
...
+	return id;
+}
+
+/*
+ * partition_scheme_exists_with_id
+ * 		Helper function to generate new ids for partition scheme, 
+ * 		it check if provided id is already in use.
+ */
+static bool
+partition_scheme_exists_with_id(int32 id)
```
rishabhtanwar29: `partition_scheme_id_exists` seems a better name.

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -4453,3 +4610,691 @@ rename_tsql_db(char *old_db_name, char *new_db_name)
...
+				(errcode(ERRCODE_INTERNAL_ERROR),
+				 errmsg("Cannot find an available ID for new partition function.")));
+	}
+	return id;
+}
+
+/*
+ * partition_scheme_exists_with_id
+ * 		Helper function to generate new ids for partition scheme, 
+ * 		it check if provided id is already in use.
```
rishabhtanwar29: Same as above, please fix the description.

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -4453,3 +4610,691 @@ rename_tsql_db(char *old_db_name, char *new_db_name)
...
+	return count;
+}
+
+/*
+ * is_partition_scheme_used
+ * 		checks if partition scheme is used by any partitioned table or not
+ * 		by looking up sys.babelfish_partition_depend catalog
+ */
+static 
+bool is_partition_scheme_used(const char *partition_scheme_name)
```
rishabhtanwar29: Nit: return type and function name should be in different lines.

contrib/babelfishpg_tsql/src/catalog.c
sumitj824: please check reply on Kushaal's comment.

contrib/babelfishpg_tsql/src/pltsql_utils.c
```diff
@@ -1170,6 +1165,24 @@ is_tsql_text_ntext_or_image_datatype(Oid oid)
...
+
+bool is_tsql_rowversion_or_timestamp_datatype(Oid oid)
+{
+	return (*common_utility_plugin_ptr->is_tsql_rowversion_or_timestamp_datatype) (oid);
+}
+
+bool is_tsql_xml_datatype(Oid oid)
+{
+	return (*common_utility_plugin_ptr->is_tsql_xml_datatype) (oid);
+}
```
rishabhtanwar29: Why do we need yet another wrapper function for this? I would prefer to directly make use of the `common_utility_plugin_ptr`

contrib/babelfishpg_tsql/src/tsqlNodes.h
```diff
@@ -2,109 +2,6 @@
 #include "executor/spi.h"
 #include "pltsql.h"
 
-#if 0
```
rishabhtanwar29, line 5: Thanks for removing this unused code.

test/JDBC/upgrade/latest/schedule
```diff
@@ -538,3 +538,4 @@ test_conv_string_to_datetimeoffset
 test_conv_string_to_smalldatetime
 test_conv_string_to_time
 BABEL-3401
+PARTITION
```
rishabhtanwar29: (Small nit): Both Here and at other places, it is always a good practice to add newline at the end of a file.

contrib/babelfishpg_common/src/typecode.c
sumitj824: yes, removed this unnecessary change.

contrib/babelfishpg_tsql/src/pltsql_utils.c
sumitj824: I had created these wrapper function similar to other function to make code cleaner. Removed these in next commit.

contrib/babelfishpg_common/src/typecode.c
```diff
@@ -392,6 +394,22 @@ is_tsql_rowversion_datatype(Oid oid)
 	return tsql_rowversion_oid == oid;
 }
 
+bool
+is_tsql_geometry_datatype(Oid oid)
+{
+	if (tsql_geometry_oid == InvalidOid)
+		tsql_geometry_oid = lookup_tsql_datatype_oid("geometry");
```
Deepesh125, line 401: lookup_tsql_datatype_oid may return InvalidOid. we should safe guard against it.

contrib/babelfishpg_tsql/src/tsqlIface.cpp
Deepesh125: then how is it different from other DDLs and why are we handling it along with other operations such as CREATE TABLE handled on line 1881

contrib/babelfishpg_tsql/src/pltsql.h
```diff
@@ -1828,6 +1858,17 @@ typedef struct tsql_identity_insert_fields
 	Oid			schema_oid;
 } tsql_identity_insert_fields;
 
+/* 
+ * used to provide extra arg during sort,
```
Deepesh125: add little more info on this sort, add example function name for the reference

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -4029,3 +4036,317 @@ exec_stmt_fulltextindex(PLtsql_execstate *estate, PLtsql_stmt_fulltextindex *stm
...
+		if (!tsql_typename || is_tsql_text_ntext_or_image_datatype(typ->typoid) ||
+			(*common_utility_plugin_ptr->is_tsql_geometry_datatype) (typ->typoid) ||
+			(*common_utility_plugin_ptr->is_tsql_geography_datatype) (typ->typoid) ||
+			(*common_utility_plugin_ptr->is_tsql_rowversion_or_timestamp_datatype) (typ->typoid) ||
+			typ->typoid == XMLOID) /* we don't have XML type specific to TSQL */
+		{
+			ereport(ERROR, 
+				(errcode(ERRCODE_INVALID_PARAMETER_VALUE),
+					errmsg("The type '%s' is not valid for this operation.", typ->typname)));
+		}
```
Deepesh125: will this code work if there is user defined data type based on one of these data types?

contrib/babelfishpg_tsql/src/catalog.c
Deepesh125: why do we want to do unnecessary processing?

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -4453,3 +4610,704 @@ rename_tsql_db(char *old_db_name, char *new_db_name)
...
+			(errcode(ERRCODE_UNDEFINED_OBJECT), 
+				errmsg("Cannot drop the partition scheme '%s', because it does not exist or you do not have permission.", partition_scheme_name)));
+	}
+
+	/* raise error if there are dependent tables on it */
+	if (has_dependent_objects) 
+	{
+		ereport(ERROR, 
+			(errcode(ERRCODE_UNDEFINED_OBJECT),
+				errmsg("The partition scheme \"%s\" is currently being used to partition one or more tables.", partition_scheme_name)));
```
Deepesh125, line 5184: do we have test cases for this?

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -4453,3 +4610,704 @@ rename_tsql_db(char *old_db_name, char *new_db_name)
...
+	
+	ScanKeyInit(&scanKey,
+					Anum_bbf_partition_function_id,
+					BTEqualStrategyNumber, F_INT2EQ,
+					Int32GetDatum(id));
+
+	/* scan using index */
+	scan = systable_beginscan(rel,
+			get_bbf_partition_function_id_idx_oid(),
+			false, NULL, 1, &scanKey);
```
Deepesh125, line 4694: Is this possible due to unique key thing on the column?

Deepesh125 (changes requested):

contrib/babelfishpg_tsql/src/pl_exec-2.c
sumitj824: yes, for all user defined types `tsql_typename` will be NULL and it will throw an error similar to TSQL.

contrib/babelfishpg_tsql/src/catalog.c
sumitj824: Not yet, it will be added with the support of partitioned table.

contrib/babelfishpg_tsql/src/catalog.c
sumitj824: yes, unique constraint creates index internally.

contrib/babelfishpg_tsql/src/catalog.c
sumitj824: removed duplicate calls.

contrib/babelfishpg_tsql/src/tsqlIface.cpp
sumitj824: we don't need to invoke the mutator to rewrite these statements.

contrib/babelfishpg_common/src/typecode.c
sumitj824: safe guarding is not needed here.

Deepesh125 (dismissed):
Approving all the code except changes related to dump/restore. Get it reviewed by @rishabhtanwar29

rishabhtanwar29 (approved):
Dump/restore changes look good.

Deepesh125 (dismissed):

contrib/babelfishpg_tsql/sql/object_definition_tsql.sql
```diff
@@ -11,4 +11,53 @@ CREATE TABLE sys.babelfish_view_def (
 );
 GRANT SELECT ON sys.babelfish_view_def TO PUBLIC;
 
-SELECT pg_catalog.pg_extension_config_dump('sys.babelfish_view_def', '');
\ No newline at end of file
+-- BBF_PARTITION_FUNCTION
```
KushaalShroff: why is this part of the object_definition_tsql.sql file?

This file is only for the catalogs specifically added for the object definition purposes

contrib/babelfishpg_tsql/sql/object_definition_tsql.sql
```diff
@@ -11,4 +11,53 @@ CREATE TABLE sys.babelfish_view_def (
...
+(
+  dbid SMALLINT NOT NULL, -- to maintain separation b/w databases
+  function_id INT NOT NULL UNIQUE, -- globally unique ID of partition function
+  partition_function_name sys.NVARCHAR(128) NOT NULL COLLATE sys.database_default,
+  input_parameter_type sys.sysname, -- data type of the column which will be used for partitioning
+  partition_option sys.bit, -- whether boundary option is LEFT or RIGHT
+  range_values sys.sql_variant[] CHECK (array_length(range_values, 1) < 15000), -- boundary values
+  create_date SYS.DATETIME NOT NULL,
+  modify_date SYS.DATETIME NOT NULL,
+  PRIMARY KEY(dbid, partition_function_name)
```
KushaalShroff: is it worth to add function_id as well in the primary key?

contrib/babelfishpg_tsql/src/catalog.h
```diff
@@ -386,6 +386,89 @@ extern Oid	bbf_extended_properties_idx_oid;
 extern Oid	get_bbf_extended_properties_oid(void);
 extern Oid	get_bbf_extended_properties_idx_oid(void);
 
+/*****************************************
+ *			PARTITION_FUNCTION
+ *****************************************/
+#define BBF_PARTITION_FUNCTION_TABLE_NAME "babelfish_partition_function"
+#define BBF_PARTITION_FUNCTION_IDX_NAME "babelfish_partition_function_pkey"
+#define BBF_PARTITION_FUNCTION_ID_IDX_NAME "babelfish_partition_function_function_id_key"
```
KushaalShroff: Since we have 2 indexes then it would be better to name them consistently so that one can clearly make out the difference

contrib/babelfishpg_tsql/sql/object_definition_tsql.sql
```diff
@@ -11,4 +11,53 @@ CREATE TABLE sys.babelfish_view_def (
...
+  dbid SMALLINT NOT NULL, -- to maintain separation between databases
+  scheme_id INT NOT NULL UNIQUE, -- globally unique ID of partition scheme
+  partition_scheme_name sys.NVARCHAR(128) NOT NULL COLLATE sys.database_default,
+  partition_function_name sys.NVARCHAR(128) NOT NULL COLLATE sys.database_default,
+  next_used sys.bit, -- next used filegroup is specified or not during creation
+  PRIMARY KEY(dbid, partition_scheme_name)
+);
+
+-- SEQUENCE to maintain the ID of partition scheme.
+CREATE SEQUENCE sys.babelfish_partition_scheme_seq START 2 MAXVALUE 2147483647 CYCLE;
```
KushaalShroff: Can you please add a comment why this starts with 2?

contrib/babelfishpg_tsql/sql/object_definition_tsql.sql
```diff
@@ -11,4 +11,53 @@ CREATE TABLE sys.babelfish_view_def (
...
+  input_parameter_type sys.sysname, -- data type of the column which will be used for partitioning
+  partition_option sys.bit, -- whether boundary option is LEFT or RIGHT
+  range_values sys.sql_variant[] CHECK (array_length(range_values, 1) < 15000), -- boundary values
+  create_date SYS.DATETIME NOT NULL,
+  modify_date SYS.DATETIME NOT NULL,
+  PRIMARY KEY(dbid, partition_function_name)
+);
+
+-- SEQUENCE to maintain the ID of partition function.
+CREATE SEQUENCE sys.babelfish_partition_function_seq START 2 MAXVALUE 2147483647 CYCLE;
```
KushaalShroff: Can you please add a comment why this starts with 2?

contrib/babelfishpg_tsql/src/tsqlIface.cpp
```diff
@@ -1839,6 +1859,12 @@ class tsqlBuilder : public tsqlCommonMutator
 			clear_rewritten_query_fragment();
 			return;
 		}
+		if (ctx->create_partition_function() || ctx->drop_partition_function()
+			 || ctx->create_partition_scheme() || ctx->drop_partition_scheme())
+		{
+			clear_rewritten_query_fragment();
```
KushaalShroff: Why are we calling clear_rewritten_query_fragment?

contrib/babelfishpg_tsql/src/tsqlIface.cpp
```diff
@@ -7091,6 +7117,81 @@ getDatabaseSchemaAndTableName(TSqlParser::Table_nameContext* tctx)
 	                        downcase_truncate_identifier(table_name.c_str(), table_name.length(), true));
 }
 
+PLtsql_stmt *
+makeCreatePartitionFunction(TSqlParser::Create_partition_functionContext *ctx)
+{
+	PLtsql_stmt_partition_function *stmt = (PLtsql_stmt_partition_function *) palloc0(sizeof(PLtsql_stmt_partition_function));
```
KushaalShroff: why are we doing palloc0 instead of palloc

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -4057,3 +4064,317 @@ exec_stmt_fulltextindex(PLtsql_execstate *estate, PLtsql_stmt_fulltextindex *stm
...
+		pfree(arr_value);
+	}
+	else /* drop */
+	{
+		/* delete entry from the sys.babelfish_partition_scheme catalog */
+		remove_entry_from_bbf_partition_function(partition_function_name);
+	}
+
+	/* make sure later steps can see the updated catalog entry */
+	CommandCounterIncrement();
```
KushaalShroff, line 4318: can you explain in detail why we need this?

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -4057,3 +4064,317 @@ exec_stmt_fulltextindex(PLtsql_execstate *estate, PLtsql_stmt_fulltextindex *stm
...
+
+		pfree(tsql_typename);
+		pfree(values);
+		pfree(arr_value);
+	}
+	else /* drop */
+	{
+		/* delete entry from the sys.babelfish_partition_scheme catalog */
+		remove_entry_from_bbf_partition_function(partition_function_name);
+	}
```
KushaalShroff: Can we move this up and return quickly, this will just make the code look cleaner and avoid extra tab spacing for the above if block

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -4057,3 +4064,317 @@ exec_stmt_fulltextindex(PLtsql_execstate *estate, PLtsql_stmt_fulltextindex *stm
...
+	if (result == 0)
+		cxt->contains_duplicate = true;
+	return result;
+}
+
+/*
+ * Allows only login that is either db owner or member of sysadmin.
+ */
+static void
+perform_permission_check(const char *name, bool is_create, bool is_function)
```
KushaalShroff: Needs better naming to indicate the specific usecase

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -4057,3 +4064,317 @@ exec_stmt_fulltextindex(PLtsql_execstate *estate, PLtsql_stmt_fulltextindex *stm
...
+
+		/* check if the given number of boundaries are exceeding allowed limit */
+		if (nargs >= 15000)
+		{
+			ereport(ERROR, 
+				(errcode(ERRCODE_INVALID_PARAMETER_VALUE),
+					errmsg("CREATE/ALTER partition function failed as only a maximum of 15000 partitions can be created.")));
+		}
+
+		values = palloc0(nargs * sizeof(Datum));
```
KushaalShroff: why are we using palloc0 again?

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -4057,3 +4064,317 @@ exec_stmt_fulltextindex(PLtsql_execstate *estate, PLtsql_stmt_fulltextindex *stm
...
+			(*common_utility_plugin_ptr->is_tsql_rowversion_or_timestamp_datatype) (typ->typoid) ||
+			typ->typoid == XMLOID) /* we don't have XML type specific to TSQL */
+		{
+			ereport(ERROR, 
+				(errcode(ERRCODE_INVALID_PARAMETER_VALUE),
+					errmsg("The type '%s' is not valid for this operation.", typ->typname)));
+		}
+
+		/* check if the given number of boundaries are exceeding allowed limit */
+		if (nargs >= 15000)
```
KushaalShroff: 15000 can be a macro

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -4057,3 +4064,317 @@ exec_stmt_fulltextindex(PLtsql_execstate *estate, PLtsql_stmt_fulltextindex *stm
...
+		}
+
+		values = palloc0(nargs * sizeof(Datum));
+
+		for (int i = 0; i < nargs; i++)
+		{
+			Datum val;
+
+			/* evaluate the value from the expr */
+			val = exec_eval_expr(estate, list_nth(arg, i), &isnull, &valtype, &valtypmod);
```
KushaalShroff: Do we have enough test cases to cover these scenarios?

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -4057,3 +4064,317 @@ exec_stmt_fulltextindex(PLtsql_execstate *estate, PLtsql_stmt_fulltextindex *stm
...
+				values[i] = exec_cast_value(estate, val, &isnull,
+								valtype, valtypmod,
+								typ->typoid, typ->atttypmod);
+			}
+			PG_CATCH();
+			{
+				MemoryContext err_ctx;
+				err_ctx = MemoryContextSwitchTo(cur_ctxt);
+				pfree(tsql_typename);
+				pfree(values);
```
KushaalShroff: You dont need to free in error cycle, hence the memoryContext switch is also not required

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -4057,3 +4064,317 @@ exec_stmt_fulltextindex(PLtsql_execstate *estate, PLtsql_stmt_fulltextindex *stm
...
+			ereport(ERROR,
+				(errcode(ERRCODE_INVALID_PARAMETER_VALUE),
+					errmsg("Duplicate values are not allowed in partition function boundary values list.")));
+		}
+
+		sql_variant_oid = (*common_utility_plugin_ptr->get_tsql_datatype_oid) ("sql_variant");
+		/* cast each value to sql_variant datatype */
+		for (int i = 0; i < nargs; i++)
+		{
+			values[i] = exec_cast_value(estate, values[i], &isnull,
```
KushaalShroff: Are we overriding the previously allocated memory here?

KushaalShroff (changes requested):

contrib/babelfishpg_tsql/sql/object_definition_tsql.sql
sumitj824: moved this to new file name `babelfish_partitions.sql`.

contrib/babelfishpg_tsql/sql/object_definition_tsql.sql
sumitj824: We have already index on function_id using UNIQUE.

contrib/babelfishpg_tsql/sql/object_definition_tsql.sql
sumitj824: added comment for partition scheme, changed the start value to 1 for partition function as it not needed.

contrib/babelfishpg_tsql/src/tsqlIface.cpp
sumitj824: removed it.

contrib/babelfishpg_tsql/src/pl_exec-2.c
sumitj824: we have testcase using local variables, i think that is sufficient here.

contrib/babelfishpg_tsql/src/pl_exec-2.c
sumitj824: Creation of partition scheme after create of partition function in same batch will fail because it will not be able see the updated catalog entry.

contrib/babelfishpg_tsql/src/pl_exec-2.c
sumitj824: addressed it.

shah-nirmit:
Test Coverage LGTM.

contrib/babelfishpg_tsql/sql/object_definition_tsql.sql
KushaalShroff: Thanks for making the change

contrib/babelfishpg_tsql/src/pl_exec-2.c
KushaalShroff: Can you please mention the test case that is failing?

contrib/babelfishpg_tsql/src/pl_exec-2.c
KushaalShroff: Lets add more test cases where local variables are getting values from volatile functions

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -4507,3 +4664,699 @@ rename_tsql_db(char *old_db_name, char *new_db_name)
...
+	bool		loop_completed = false;
+
+	do
+	{
+		id = nextval_internal(get_bbf_partition_function_seq_oid(), false);
+		if (start == 0)
+			start = id;
+		else if (start == id)
+		{
+			loop_completed = true;
```
KushaalShroff: Why cant we directly throw error here?

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -4507,3 +4664,699 @@ rename_tsql_db(char *old_db_name, char *new_db_name)
...
+ */
+static bool
+is_partition_function_used(const char *partition_function_name)
+{
+	Relation	rel;
+	HeapTuple	tuple;
+	SysScanDesc	scan;
+	ScanKeyData	scanKey[2];
+	bool		is_used = false;
+	int16		dbid = get_cur_db_id();
```
KushaalShroff: As discussed offline, cross db queries do not work for partitions in TSQL. Can we add these test cases for 4 part names.

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -4507,3 +4664,699 @@ rename_tsql_db(char *old_db_name, char *new_db_name)
...
+			has_dependent_objects = false;
+			CatalogTupleDelete(rel, &tuple->t_self);
+		}
+	}
+
+	systable_endscan(scan);
+	table_close(rel, RowExclusiveLock);
+
+	/* raise error if it doesn't exists in database */
+	if (!partition_function_exists)
```
KushaalShroff: Quick thought: will this resolve to function being defined and thus always resolving to false or will it resolve to local variable since the names are same "partition_function_exists".
It will be local variable, but it would be good to rename the variables since we have seen some weird scenarios with static functions' call stack before

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -4507,3 +4664,699 @@ rename_tsql_db(char *old_db_name, char *new_db_name)
...
+	table_close(rel, AccessShareLock);
+	return exists;
+}
+
+/*
+ * get_partition_function
+ * 		Returns the partition function name for the given partition scheme name.
+ */
+char*
+get_partition_function(const char *partition_scheme_name)
```
KushaalShroff: Where is this getting used?

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -4507,3 +4664,699 @@ rename_tsql_db(char *old_db_name, char *new_db_name)
...
+
+	while ((tuple = systable_getnext(scan)) != NULL)
+	{
+		if (HeapTupleIsValid(tuple))
+			CatalogTupleDelete(rel, &tuple->t_self);
+	}
+
+	systable_endscan(scan);
+	table_close(rel, RowExclusiveLock);
+}
```
KushaalShroff, line 5339: Are we also taking care of metadata consistency checks

test/JDBC/expected/PARTITION-vu-verify.out
```diff
@@ -0,0 +1,954 @@
...
+
+CREATE PARTITION FUNCTION DateTime2PartitionFunctionContainsDuplicateValue (datetime2(3))
+AS RANGE RIGHT FOR VALUES ('2022-01-01 00:00:00.000', '2022-07-01 00:00:00.000', '2019-01-01 00:00:00.000', '2022-07-01 00:00:00.000');
+GO
+~~ERROR (Code: 33557097)~~
+
+~~ERROR (Message: Duplicate values are not allowed in partition function boundary values list.)~~
+
+
+CREATE PARTITION FUNCTION SmallDateTimePartitionFunctionContainsDuplicateValue (smalldatetime)
```
KushaalShroff, line 299: "Where is the money" test case🤑

test/JDBC/expected/PARTITION-vu-verify.out
```diff
@@ -0,0 +1,954 @@
...
+
+CREATE PARTITION FUNCTION CharPartitionFunctionContainsDuplicateValue (char(5))
+AS RANGE RIGHT FOR VALUES ('A', 'F', 'K', 'P', 'U', 'Z', 'D', 'K');
+GO
+~~ERROR (Code: 33557097)~~
+
+~~ERROR (Message: Duplicate values are not allowed in partition function boundary values list.)~~
+
+
+CREATE PARTITION FUNCTION VarCharPartitionFunctionContainsDuplicateValue (varchar(10))
```
KushaalShroff, line 315: Can we add test cases for n/varchar(MAX)

test/JDBC/expected/PARTITION-vu-verify.out
```diff
@@ -0,0 +1,954 @@
...
+
+CREATE PARTITION FUNCTION NCharPartitionFunctionContainsDuplicateValue (nchar(5))
+AS RANGE RIGHT FOR VALUES (N'A', N'F', N'K', N'P', N'U', N'Z', N'B',  N'K');
+GO
+~~ERROR (Code: 33557097)~~
+
+~~ERROR (Message: Duplicate values are not allowed in partition function boundary values list.)~~
+
+
+CREATE PARTITION FUNCTION BinaryPartitionFunctionContainsDuplicateValue (binary(4))
```
KushaalShroff, line 339: Can you please add test cases for var/binary(MAX)

test/JDBC/expected/PARTITION-vu-verify.out
```diff
@@ -0,0 +1,954 @@
...
+CREATE PARTITION FUNCTION UniqueIdentifierPartitionFunctionContainsDuplicateValue (uniqueidentifier)
+AS RANGE RIGHT FOR VALUES ('00000000-0000-0000-0000-000000000000', '6F9619FF-8B86-D011-B42D-00C04FC964FF', 'FFFFFFFF-FFFF-FFFF-FFFF-FFFFFFFFFFFF', '6F9619FF-8B86-D011-B42D-00C04FC964FF');
+GO
+~~ERROR (Code: 33557097)~~
+
+~~ERROR (Message: Duplicate values are not allowed in partition function boundary values list.)~~
+
+
+CREATE PARTITION FUNCTION SqlVariantPartitionFunctionContainsDuplicateValue (sql_variant)
+AS RANGE RIGHT FOR VALUES (1, 'abc', CONVERT(DATETIME, '2023-05-01'), 'abc');
```
KushaalShroff, line 364: what if abc is of different base type? can we also add different collations?

test/JDBC/expected/PARTITION-vu-verify.out
```diff
@@ -0,0 +1,954 @@
...
+SmallIntPartitionFunction#!#R #!#RANGE#!#4#!#1#!#0
+SqlVariantPartitionFunction#!#R #!#RANGE#!#6#!#1#!#0
+TinyIntPartitionFunction#!#R #!#RANGE#!#4#!#1#!#0
+UniqueIdentifierPartitionFunction#!#R #!#RANGE#!#4#!#1#!#0
+VarBinaryPartitionFunction#!#R #!#RANGE#!#5#!#1#!#0
+VarCharPartitionFunction#!#R #!#RANGE#!#6#!#1#!#0
+~~END~~
+
+
+SELECT * FROM GetRangeMetadataView
```
KushaalShroff, line 579: No order by?

test/JDBC/input/PARTITION-vu-prepare.mix
```diff
@@ -0,0 +1,385 @@
...
+    SELECT 
+        name, type, type_desc, fanout, boundary_value_on_right, is_system
+    FROM 
+        sys.partition_functions
+    ORDER BY 
+        name
+)
+GO
+
+CREATE VIEW GetRangeMetadataVIEW
```
KushaalShroff, line 18: Lets add order by here as well

test/JDBC/input/PARTITION-vu-prepare.mix
```diff
@@ -0,0 +1,385 @@
...
+    FROM 
+        sys.partition_range_values pr
+    LEFT JOIN 
+        sys.partition_functions pf on (pr.function_id = pf.function_id)
+    WHERE
+        name not in ('BinaryPartitionFunction', 'VarBinaryPartitionFunction')
+)
+GO
+
+CREATE VIEW GetRangeMetadataVarBinaryFunctionView
```
KushaalShroff, line 34: Lets add order by here as well

test/JDBC/expected/PARTITION-vu-verify.out
```diff
@@ -0,0 +1,954 @@
...
+
+
+
+
+
+
+-- create when arg > max allowed limit should throw error
+DECLARE @partition_count INT = 14999;
+DECLARE @partition_interval INT = 1000;
+DECLARE @partition_function_definition VARCHAR(MAX) = 'CREATE PARTITION FUNCTION PartitionFunctionWith15000Arg (int) AS RANGE RIGHT FOR VALUES (';
```
KushaalShroff, line 835: Is create partition allowed inside a procedure/function?

Can you please add test cases?

KushaalShroff (changes requested):

test/JDBC/expected/PARTITION-vu-verify.out
sumitj824: It is allowed inside procedure while not allowed inside function. Added testcase.

test/JDBC/expected/PARTITION-vu-verify.out
sumitj824: added tests.

test/JDBC/expected/PARTITION-vu-verify.out
sumitj824: Thanks for pointing it out, (MAX) is not allowed. I have done the handling in latest commit and added testcases.

test/JDBC/expected/PARTITION-vu-verify.out
sumitj824: added.

contrib/babelfishpg_tsql/src/catalog.c
sumitj824: We will add the metadata consistency checks later as part of Task:BABEL-5024.

contrib/babelfishpg_tsql/src/catalog.c
sumitj824: It will get used in incoming PR to support partitioned tables.

contrib/babelfishpg_tsql/src/catalog.c
sumitj824: added testcase.

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -4057,3 +4064,343 @@ exec_stmt_fulltextindex(PLtsql_execstate *estate, PLtsql_stmt_fulltextindex *stm
...
+		typ->typoid == XMLOID) /* we don't have XML type specific to TSQL */
+	{
+		ereport(ERROR, 
+			(errcode(ERRCODE_INVALID_PARAMETER_VALUE),
+				errmsg("The type '%s' is not valid for this operation.", typ->typname)));
+	}
+	/*
+	 * Types varchar(max), nvarchar(max), varbinary(max) are also not supported.
+	 */
+	else if (typ->atttypmod == -1 && is_tsql_datatype_with_max_scale_expr_allowed(typ->typoid))
```
KushaalShroff, line 4213: Thanks for adding this

test/JDBC/expected/PARTITION-vu-prepare.out
```diff
@@ -0,0 +1,458 @@
...
+BEGIN
+DROP PARTITION SCHEME PartitionSchemeInsideProc1 
+END;
+GO
+
+CREATE PROCEDURE PartitionProcedureDropPartitionFunctionAndScheme AS
+BEGIN
+DROP PARTITION SCHEME PartitionSchemeInsideProc2;
+DROP PARTITION FUNCTION PartitionFunctionInsideProc2;
+END;
```
KushaalShroff, line 457: Have we made sure that ddl export works?

KushaalShroff (approved):

test/JDBC/expected/PARTITION-vu-prepare.out
sumitj824: I have validated this manually and we will add requirred DDL scripting test in separate task.

contrib/babelfishpg_tsql/src/pl_exec-2.c
sumitj824: It was failing for procedure which creates partition scheme after creating partition function.

Deepesh125 (approved):

PR #2591, opened 2024-05-20
Fix differences in T-SQL COALESCE function while calling the function with variables and constants together

basasairohan:
### Description

This commit will fix the following issues with T-SQL COALESCE function : 

1. When the arguments contain the mix of variables and constants as shown below, the below query should return 4 as it is the first non-null value. But BBF throws error as we try to evaluate the constant ('dfidhfgi') at the compile time in this case.
```
DECLARE @abc int = 4 ;
select COALESCE(@abc, 'dfidhfgi')
```
2. When the arguments are of `varbinary` and `nvarchar` type, BBF throws error as we have defined the CAST between `varbinary` and `nvarchar` as ASSIGNMENT, but in this case IMPLICIT casting is used.
```
select coalesce(CAST('x'AS VARBINARY), CAST('x' AS NVARCHAR(4000)), 'x')
```

### Issues Resolved

BABEL-726

* **Use case based -** babel_726-vu-*.sql

* **Boundary conditions -** babel_726-vu-*.sql

* **Arbitrary inputs -** babel_726-vu-*.sql

* **Negative test cases -** babel_726-vu-*.sql

* **Minor version upgrade tests -** babel_726-vu-*.sql

* **Major version upgrade tests -** babel_726-vu-*.sql

* **Performance tests -** N/A

* **Tooling impact -** N/A

* **Client tests -** N/A

### Check List


test/JDBC/expected/babel_726-vu-verify.out
```diff
@@ -0,0 +1,575 @@
...
+b              
+~~END~~
+
+
+-- char and varbinary
+select coalesce(a25, a26) from babel_726_t1
+GO
+~~ERROR (Code: 33557097)~~
+
+~~ERROR (Message: COALESCE could not convert type varbinary to bpchar)~~
```
thephantomthief, line 456: is this expected?

test/JDBC/expected/babel_726-vu-verify.out
```diff
@@ -0,0 +1,575 @@
...
+
+exec babel_726_vu_prepare_p1
+go
+~~START~~
+nvarchar
+x
+~~END~~
+
+
+select babel_726_vu_prepare_f1()
```
thephantomthief, line 561: So this means that if this procedure was executed on prior versions, it would give an error but starting this version, we will get the correct output?

test/JDBC/expected/babel_726-vu-prepare.out
```diff
@@ -0,0 +1,98 @@
...
+go
+~~ROW COUNT: 1~~
+
+~~ROW COUNT: 1~~
+
+~~ROW COUNT: 1~~
+
+
+-- dependency testing
+CREATE FUNCTION babel_726_vu_prepare_f1() RETURNS NVARCHAR(4000) AS
```
thephantomthief, line 88: we cannot have a view dependent on `COALESCE(CAST('x'AS VARBINARY), CAST('x' AS NVARCHAR(4000)), 'x')` on previous versions? Because it will throw error?

contrib/babelfishpg_common/sql/upgrades/babelfish_common_helper--4.1.0--4.2.0.sql
```diff
@@ -862,5 +862,10 @@ CREATE OPERATOR CLASS sys.numeric_int8 FOR TYPE int8
 END IF;
 END $$;
 
+DROP CAST IF EXISTS(sys.BBF_VARBINARY AS sys.VARCHAR);
```
surendravishnoi: this is problematic as dropping case could fail due to dependent objects and result in upgrade failure. Please talk to Dipesh regarding this.

test/JDBC/expected/babel_726-vu-prepare.out
basasairohan: Yes

test/JDBC/expected/babel_726-vu-verify.out
basasairohan: Yes

test/JDBC/expected/babel_726-vu-verify.out
basasairohan: We do not have any IMPLICIT CAST between varbinary and bpchar. That is the reason this is failing. IMO, This should be tracked as a separate issue.

contrib/babelfishpg_common/sql/upgrades/babelfish_common_helper--4.1.0--4.2.0.sql
```diff
@@ -862,10 +862,17 @@ CREATE OPERATOR CLASS sys.numeric_int8 FOR TYPE int8
...
-WITH FUNCTION sys.varbinarysysvarchar (sys.BBF_VARBINARY, integer, boolean) AS IMPLICIT;
+DO $$
+DECLARE 
+    schema_oid oid;
+    cast_source oid;
+    cast_target oid;
+BEGIN
+    select oid INTO schema_oid from pg_namespace where nspname='sys';
+    select oid into cast_source from pg_type where typname='bbf_varbinary' and typnamespace=schema_oid;
+    select oid into cast_target from pg_type where typname='varbinary' and typnamespace=schema_oid;
```
thephantomthief: Shouldn't this be varchar?

thephantomthief (approved):

shardgupta (approved):

PR #2608, opened 2024-05-23
pg_upgrade fails with an error "ERROR:  cannot change data type of view column "definition" from nvarchar to nvarchar(4000)"

shalinilohia50:
### Description
pg_upgrade fails with an error "ERROR: cannot change data type of view column "definition" from nvarchar to nvarchar(4000)"

```
ALTER EXTENSION
ERROR:  cannot change data type of view column "definition" from nvarchar to nvarchar(4000)
```
We have changed the datatype of definition column of all_sql_modules_internal from nvarchar(4000) to nvarchar in version 2.9. In upgrade script 3.0 to 3.1, we have again defined the view  all_sql_modules_internal where definition column is of nvarchar(4000) datatype which is incorrect and causing the upgrade failure. Reference commit: https://github.com/babelfish-for-postgresql/babelfish_extensions/commit/508180ff77ad5820342aa14b834aaf013a9991ed 

### Issues Resolved

* **Use case based -**

* **Boundary conditions -**

* **Arbitrary inputs -**

* **Negative test cases -**

* **Performance tests -**

* **Tooling impact -**

* **Client tests -**

### Check List


xhfanhe:
I'm upset that we are removing lines from a previous version upgrade script. 

Are we planning to backport this change to previous minor versions? What if cx performs MVU 2.8->3.1, and then 3.1->3.2? IMO this will not work unless we backport this to 3.1. 

And even if we are backporting this, we should use the deprecate -> recreate process instead of simply removing this piece.

Or, we can consider to only support `nvarchar(max)` in latest major version 4.2 and not bother with previous version support.

test/JDBC/upgrade/14_3/schedule
```diff
@@ -187,7 +187,6 @@ sys_all_objects
 sys_all_objects-dep
 sys-all_parameters
 sys-all_sql_modules_before-14_5
-sys-all_sql_modules-before-14_6-dep
```
2jungkook, line 190: Why do you remove tests in this schedule file?

test/JDBC/upgrade/14_6/schedule
```diff
@@ -323,7 +323,7 @@ sys-syslanguages
 sys-sysobjects
 sys-system_objects
 sys-system_sql_modules
-sys-system_sql_modules-dep
+sys-system_sql_modules-before-14_6-dep
```
2jungkook: This is 14_6 schedule file so we should not use "before-14_6" here.

2jungkook:
> I'm upset that we are removing lines from a previous version upgrade script.
> 
> Are we planning to backport this change to previous minor versions? What if cx performs MVU 2.8->3.1, and then 3.1->3.2? IMO this will not work unless we backport this to 3.1.
> 
> And even if we are backporting this, we should use the deprecate -> recreate process instead of simply removing this piece.
> 
> Or, we can consider to only support `nvarchar(max)` in latest major version 4.2 and not bother with previous version support.

2.8->3.1 and 3.1->3.2 work fine because during 2.8->3.1 we use 3.0-->3.1 upgrade script in 3_1_STABLE branch. However, my concern here is the following case:
- 2.3/2.4/2.5/2.6/2.7/2.8 --> 3.6

In this case, 2.8-->2.9 upgrade script is not used (hence `sys.all_sql_modules_internal` still has `nvarchar(4000)`) and we use 3.0-->3.1 upgrade script modified by this PR. That means, when we execute 3.1 --> 3.2 --> ... --> 3.5 upgrade scripts, we are using an older version of `sys.all_sql_modules_internal` which we didn't expect at the time of writing 3.1 --> 3.2 --> ... --> 3.5 upgrade scripts.

Deepesh125:
> > I'm upset that we are removing lines from a previous version upgrade script.
> > Are we planning to backport this change to previous minor versions? What if cx performs MVU 2.8->3.1, and then 3.1->3.2? IMO this will not work unless we backport this to 3.1.
> > And even if we are backporting this, we should use the deprecate -> recreate process instead of simply removing this piece.
> > Or, we can consider to only support `nvarchar(max)` in latest major version 4.2 and not bother with previous version support.
> 
> 2.8->3.1 and 3.1->3.2 work fine because during 2.8->3.1 we use 3.0-->3.1 upgrade script in 3_1_STABLE branch. However, my concern here is the following case:
> 
> * 2.3/2.4/2.5/2.6/2.7/2.8 --> 3.6
> 
> In this case, 2.8-->2.9 upgrade script is not used (hence `sys.all_sql_modules_internal` still has `nvarchar(4000)`) and we use 3.0-->3.1 upgrade script modified by this PR. That means, when we execute 3.1 --> 3.2 --> ... --> 3.5 upgrade scripts, we are using an older version of `sys.all_sql_modules_internal` which we didn't expect at the time of writing 3.1 --> 3.2 --> ... --> 3.5 upgrade scripts.

@2jungkook 3.5--3.6 script will have updated version of the definition.Hence, as far as target version is 3.6 no matter from what path is customer coming from, they will get the latest definition of the view.

2jungkook:
> > > I'm upset that we are removing lines from a previous version upgrade script.
> > > Are we planning to backport this change to previous minor versions? What if cx performs MVU 2.8->3.1, and then 3.1->3.2? IMO this will not work unless we backport this to 3.1.
> > > And even if we are backporting this, we should use the deprecate -> recreate process instead of simply removing this piece.
> > > Or, we can consider to only support `nvarchar(max)` in latest major version 4.2 and not bother with previous version support.
> > 
> > 
> > 2.8->3.1 and 3.1->3.2 work fine because during 2.8->3.1 we use 3.0-->3.1 upgrade script in 3_1_STABLE branch. However, my concern here is the following case:
> > 
> > * 2.3/2.4/2.5/2.6/2.7/2.8 --> 3.6
> > 
> > In this case, 2.8-->2.9 upgrade script is not used (hence `sys.all_sql_modules_internal` still has `nvarchar(4000)`) and we use 3.0-->3.1 upgrade script modified by this PR. That means, when we execute 3.1 --> 3.2 --> ... --> 3.5 upgrade scripts, we are using an older version of `sys.all_sql_modules_internal` which we didn't expect at the time of writing 3.1 --> 3.2 --> ... --> 3.5 upgrade scripts.
> 
> @2jungkook 3.5--3.6 script will have updated version of the definition.Hence, as far as target version is 3.6 no matter from what path is customer coming from, they will get the latest definition of the view.

@Deepesh125 I'm talking about 3.1 --> 3.2 --> ... --> 3.5 upgrade script, not 3.5 --> 3.6 upgrade script. When we wrote 3.1 --> 3.2 --> ... --> 3.5 upgrade script, our assumption for `sys.all_sql_modules_internal` was based on 3.0 --> 3.1 upgrade script. This PR breaks that assumption. 

In the 3.0 --> 3.1 upgrade script, we should rename and deprecate `sys.all_sql_modules_internal`, i.e., we fallback to original `nvarchar(4000)`. Based on that we execute 3.1 --> 3.2 --> ... --> 3.5 upgrade script. Finally, in 3.5 --> 3.6 upgrade script, `sys.all_sql_modules_internal` will be back to `nvarchar` again.

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.2.0--2.3.0.sql
```diff
@@ -4174,46 +4174,6 @@ CALL sys.babelfish_update_collation_to_default('sys', 'objects', 'type_desc');
 CALL sys.babelfish_update_collation_to_default('sys', 'sysobjects', 'xtype');
 CALL sys.babelfish_update_collation_to_default('sys', 'sysobjects', 'type');
 
--- TODO: BABEL-3127
-CREATE OR REPLACE VIEW sys.all_sql_modules_internal AS
```
2jungkook, line 4178: Please do not remove anything from previous upgrade scripts. It will cause more unpredictable problems. Also, you are modifying old schedule files, that is already a bad sign.

2jungkook (changes requested):

PR #2620, opened 2024-05-26
[OSS-ONLY] Remove extra check for helper functions for sp_describe_undeclared_parameters_internal

ahmed-shameem:
### Description
This commit removes the extra check (if condition) for helper functions for sp_describe_undeclared_parameters_internal which was introduced in this [commit](https://github.com/babelfish-for-postgresql/babelfish_extensions/commit/eda23eac1e40e8d6fd4453eb2d0d59c833008373). 

### Issues Resolved

BABEL-4869

* **Use case based -**

* **Boundary conditions -**

* **Arbitrary inputs -**

* **Negative test cases -**

* **Performance tests -**

* **Tooling impact -**

* **Client tests -**

### Check List

### DIFF
--- a/contrib/babelfishpg_tsql/src/procedures.c
+++ b/contrib/babelfishpg_tsql/src/procedures.c
@@ -630,10 +630,10 @@ handle_where_clause_attnums(ParseState *pstate, Node *w_clause, List *target_att
 					if (attrno == InvalidAttrNumber)
 					{
 						ereport(ERROR,
-								(errcode(ERRCODE_UNDEFINED_COLUMN),
-									errmsg("column \"%s\" of relation \"%s\" does not exist",
-										name,
-										RelationGetRelationName(pstate->p_target_relation))));
+							(errcode(ERRCODE_UNDEFINED_COLUMN),
+				 			 errmsg("column \"%s\" of relation \"%s\" does not exist",
+								name,
+								RelationGetRelationName(pstate->p_target_relation))));
 					}
 					target_attnums = lappend_int(target_attnums, attrno);
 					break;
@@ -731,17 +731,16 @@ handle_where_clause_restargets_left(ParseState *pstate, Node *w_clause, List *ex
 					if (attrno == InvalidAttrNumber)
 					{
 						ereport(ERROR,
-								(errcode(ERRCODE_UNDEFINED_COLUMN),
-									errmsg("column \"%s\" of relation \"%s\" does not exist",
-										name,
-										RelationGetRelationName(pstate->p_target_relation))));
+						(errcode(ERRCODE_UNDEFINED_COLUMN),
+						 errmsg("column \"%s\" of relation \"%s\" does not exist",
+							name,
+							RelationGetRelationName(pstate->p_target_relation))));
 					}
 					res = (ResTarget *) palloc(sizeof(ResTarget));
 					res->type = ref->type;
 					res->name = field->val.str;
 					res->indirection = NIL; /* Unused for now */
-					res->val = (Node *) ref;	/* Store the ColumnRef
-													* here if needed */
+					res->val = (Node *) ref;	/* Store the ColumnRef here if needed */
 					res->location = ref->location;
 
 					extra_restargets = lappend(extra_restargets, res);
@@ -826,8 +825,7 @@ handle_where_clause_restargets_right(ParseState *pstate, Node *w_clause, List *e
 					res->type = ref->type;
 					res->name = field->val.str;
 					res->indirection = NIL; /* Unused for now */
-					res->val = (Node *) ref;	/* Store the ColumnRef
-													* here if needed */
+					res->val = (Node *) ref;	/* Store the ColumnRef here if needed */
 					res->location = ref->location;
 
 					extra_restargets = lappend(extra_restargets, res);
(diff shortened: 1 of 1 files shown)


contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -859,36 +814,20 @@ handle_where_clause_restargets_right(ParseState *pstate, Node *w_clause, List *e
...
-						break;
 					}
 					ref = (ColumnRef *) xpr->rexpr;
 					field = linitial(ref->fields);
 					res = (ResTarget *) palloc(sizeof(ResTarget));
 					res->type = ref->type;
 					res->name = field->val.str;
 					res->indirection = NIL; /* Unused for now */
-					res->val = (Node *) ref; /* Store the ColumnRef here if needed */
+					res->val = (Node *) ref;	/* Store the ColumnRef
```
timchang514: nit: unnecessary diff

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -859,36 +814,20 @@ handle_where_clause_restargets_right(ParseState *pstate, Node *w_clause, List *e
 
 					if (nodeTag(xpr->rexpr) != T_ColumnRef)
 					{
-						xpr = (A_Expr *) arg;
-
-						if (nodeTag(xpr->rexpr) != T_ColumnRef)
+						if (is_sp_describe_undeclared_parameters)
```
timchang514, line 816: does this need to be two separate if statements?

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -777,16 +731,17 @@ handle_where_clause_restargets_left(ParseState *pstate, Node *w_clause, List *ex
 					if (attrno == InvalidAttrNumber)
 					{
 						ereport(ERROR,
-						(errcode(ERRCODE_UNDEFINED_COLUMN),
-						 errmsg("column \"%s\" of relation \"%s\" does not exist",
-								name,
-								RelationGetRelationName(pstate->p_target_relation))));
+								(errcode(ERRCODE_UNDEFINED_COLUMN),
```
timchang514: nit: unnecessary diff. 

Also, is this case different than the others? ie throwing an error rather than returning null result set?

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -737,38 +718,11 @@ handle_where_clause_restargets_left(ParseState *pstate, Node *w_clause, List *ex
...
-								is_supported_case_sp_describe_undeclared_parameters = false;
-								return extra_restargets;
-							}
-						}
-						ref = (ColumnRef *) xpr->lexpr;
-						field = linitial(ref->fields);
-						name = field->val.str;
-						attrno = attnameAttNum(pstate->p_target_relation, name, false);
-						if (attrno == InvalidAttrNumber)
+						if (is_sp_describe_undeclared_parameters)
```
timchang514, line 721: same as below, does it need to be two if statements?

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -649,10 +630,10 @@ handle_where_clause_attnums(ParseState *pstate, Node *w_clause, List *target_att
 					if (attrno == InvalidAttrNumber)
 					{
 						ereport(ERROR,
-						(errcode(ERRCODE_UNDEFINED_COLUMN),
-						 errmsg("column \"%s\" of relation \"%s\" does not exist",
-								name,
-								RelationGetRelationName(pstate->p_target_relation))));
+								(errcode(ERRCODE_UNDEFINED_COLUMN),
```
timchang514: nit: unnecessary diff.

Also, is this case different than the others? ie throwing an error rather than returning null result set?

contrib/babelfishpg_tsql/src/procedures.c
ahmed-shameem: Updated in latest rev.

contrib/babelfishpg_tsql/src/procedures.c
ahmed-shameem: Kept it in two separate if intentionally as this is a helper function hence it can be called by some other function. Then when the outer if satisfies, we do not know what will be the behaviour at that time. So, better to put extra check for sp_describe_undeclared_parameters.

contrib/babelfishpg_tsql/src/procedures.c
ahmed-shameem: Updated the unnecessary change in latest rev
This is a different error and expected. This will be encountered when user tries to perform sp_descirbe_undeclared_parameters on some non-existent column.

contrib/babelfishpg_tsql/src/procedures.c
ahmed-shameem: Kept it in two separate if intentionally as this is a helper function hence it can be called by some other function. Then when the outer if satisfies, we do not know what will be the behaviour at that time. So, better to put extra check for sp_describe_undeclared_parameters.

contrib/babelfishpg_tsql/src/procedures.c
ahmed-shameem: Updated the unnecessary change in latest rev
This is a different error and expected. This will be encountered when user tries to perform sp_descirbe_undeclared_parameters on some non-existent column.

timchang514 (approved):

shardgupta (approved):

PR #2613, opened 2024-05-24
Fix issue with convert function when converting string to date, datetimeoffset, datetime2, datetime, smalldatetime or time

rohit01010:
### Description
This PR will add fix for following issues:
- For converting string to datetime2, datetimeoffset and smalldatetime using convert function, we are directly using cast function which will convert the string to given type with default style. Added support to use corresponding SQL helper functions defined which will fix this issue.
- For conversion of string to date, datetime, smalldatetime and time using convert function, there are helper functions defined but are unable to handle all the supported datetime string literals. Refactored these definitions and fixed this issue.
- For conversion of string to datetime2 and datetimeoffset using convert function, there are no helper functions defined which handles this conversion as there are for conversion of string to date, datetime, smalldatetime and time. Added definitions to handle these datatypes as well.
- Function sys.babelfish_get_microsecs_from_fractsecs() is unable to handle the edge case where fractional seconds get overflowed after rounding it to given scale.
- Function babelfish_get_timeunit_from_string only returns HOURS, MINUTES, SECONDS and FRACTSECONDS from a given time string. Extended this function to return OFFHOURS, OFFMINUTES and OFFSIGN to get details of timeoffset from time string.

Authored-by: Rohit Bhagat [rohitbgt@amazon.com](mailto:rohitbgt@amazon.com)
Signed-off-by: Rohit Bhagat [rohitbgt@amazon.com](mailto:rohitbgt@amazon.com)

### Issues Resolved
BABEL-4896

### Test Scenarios Covered ###
* **Use case based -** YES

* **Boundary conditions -** YES

* **Arbitrary inputs -** YES

* **Negative test cases -** YES

* **Minor version upgrade tests -** YES

* **Major version upgrade tests -** YES

* **Performance tests -** NA

* **Tooling impact -** NA

* **Client tests -** YES

### Check List
- [x] Commits are signed per the DCO using --signoff 

By submitting this pull request, I confirm that my contribution is under the terms of the Apache 2.0 and PostgreSQL licenses, and grant any person obtaining a copy of the contribution permission to relicense all or a portion of my contribution to the PostgreSQL License solely to contribute all or a portion of my contribution to the PostgreSQL open source project.

For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/babelfish-for-postgresql/babelfish_extensions/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #2622, opened 2024-05-27
Fix issue with convert function when converting string to date, datetimeoffset, datetime2, datetime, smalldatetime or time

rohit01010:
### Description
This PR will add fix for following issues:
- For converting string to datetime2, datetimeoffset and smalldatetime using convert function, we are directly using cast function which will convert the string to given type with default style. Added support to use corresponding SQL helper functions defined which will fix this issue.
- For conversion of string to date, datetime, smalldatetime and time using convert function, there are helper functions defined but are unable to handle all the supported datetime string literals. Refactored these definitions and fixed this issue.
- For conversion of string to datetime2 and datetimeoffset using convert function, there are no helper functions defined which handles this conversion as there are for conversion of string to date, datetime, smalldatetime and time. Added definitions to handle these datatypes as well.
- Function sys.babelfish_get_microsecs_from_fractsecs() is unable to handle the edge case where fractional seconds get overflowed after rounding it to given scale.
- Function babelfish_get_timeunit_from_string only returns HOURS, MINUTES, SECONDS and FRACTSECONDS from a given time string. Extended this function to return OFFHOURS, OFFMINUTES and OFFSIGN to get details of timeoffset from time string.

First commit changes cherry-picked from: https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/2613

Authored-by: Rohit Bhagat [rohitbgt@amazon.com](mailto:rohitbgt@amazon.com)
Signed-off-by: Rohit Bhagat [rohitbgt@amazon.com](mailto:rohitbgt@amazon.com)

### Issues Resolved
BABEL-4896

* **Use case based -** YES

* **Arbitrary inputs -** YES

* **Performance tests -** NA

* **Tooling impact -** NA

* **Client tests -** YES

### Check List


contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.9.0--3.0.0.sql
```diff
@@ -99,5 +99,1292 @@ $$
 STRICT
 LANGUAGE plpgsql IMMUTABLE;
 
+CREATE OR REPLACE FUNCTION sys.babelfish_conv_string_to_date(IN p_datestring TEXT,
```
rishabhtanwar29: 1. Looks like we are re-creating these objects which we permanently dropped in 14.12 which doesn't seem right. I think we should rather have a check before deprecating the object to see if exists. Dropping a non-existing object would be fine since `babelfish_drop_deprecated_object` ignores non-existing functions.
2. These inter-version upgrade scripts are only meant for those objects which needs to updated for the compatibility with the new major version.

test/JDBC/expected/test_conv_string_to_date-before-15_7-vu-prepare.out
rishabhtanwar29, line 1: Could be `-before-14_12-or-15_7`

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.9.0--3.0.0.sql
rohit01010: Here we are re-creating these objects because upgrade script 3.0.0 to 3.1.0 is using these dropped/deprecated objects and will fail if these objects does not exists, causing upgrade to fail.

test/JDBC/expected/test_conv_string_to_date-before-15_7-vu-prepare.out
rohit01010: There is this issue: https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/2581, with JDBC test framework where long file name causes JDBC testframework to crash. Hence to avoid this changed test file names like this.

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.9.0--3.0.0.sql
rishabhtanwar29: Then let's fix 3.0.0--3.1.0 upgrade script (or any other upgrade script) instead.

test/JDBC/expected/test_conv_string_to_date-before-15_7-vu-prepare.out
rishabhtanwar29: Then it's fine to keep the names, although let's try to get https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/2581 merged ASAP.

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.9.0--3.0.0.sql
rohit01010: ok

rishabhtanwar29 (approved):

shardgupta (approved):

PR #2637, opened 2024-05-31
Fixed implementation of remove_accents_internal to better manage the memory

Deepesh125:
### Description

In previous implementation of remove_accents_internal, we were passing input string to utrans_transUChars which will modify the string in place in order to remove the accents. And memory for the input string was also not well allocated in order to accommodate this expansion of string by  utrans_transUChars causing the function to write out of allocated heap. This may cause various issues, for example, 
1. Error like "pfree called with invalid pointer 0x149fb5b3c050 (header 0x0069032703270327)"
2. warnings like "problem in alloc set ExprContext: detected write past chunk end in block 0x35c6770, chunk 0x35c67e8" and "detected write past chunk end in ExprContext 0x2b24f48"
3. even result in server crash while allocating or freeing the memory

This commit fixes the implementation of remove_accents_internal to rightly calculate the capacity and allocate the enough memory to input string to handle possible expansion of the string by utrans_transUChars. 

Task: BABEL-5006

### Check List


contrib/babelfishpg_tsql/src/collation.c
```diff
@@ -493,7 +494,7 @@ Datum remove_accents_internal(PG_FUNCTION_ARGS)
 
 		// Open transliterator
 		cached_transliterator = utrans_openU(rules, len_uchar, UTRANS_FORWARD, NULL, 0, NULL, &status);
-		if (U_FAILURE(status))
+		if (U_FAILURE(status) || !cached_transliterator)
```
ahmed-shameem, line 498: Good catch!

contrib/babelfishpg_tsql/src/collation.c
```diff
@@ -505,40 +506,63 @@ Datum remove_accents_internal(PG_FUNCTION_ARGS)
...
+	/*
+	 * utrans_transUChars will modify input string in place so ensure that it has enough capacity to store
+	 * transformed string.
+	 */
+	utf16_res = (UChar *) palloc0(capacity * sizeof(UChar));
+	/*
+	 * utf16_input would have one NULL terminator at the end. Copy that too. Limiting memory copy to min of
+	 * (len_uinput + 1) * sizeof(UChar) and capacity * sizeof(UChar) in order to avoid buffer overwriting.
+	 */
+	memcpy(utf16_res, utf16_input, Min((len_uinput + 1) * sizeof(UChar), capacity * sizeof(UChar)));
```
ahmed-shameem, line 542: I think this can lead to a potential issue when the (len_uinput + 1) * sizeof(UChar) < capacity * sizeof(UChar) but we need more buffer than what we estimated. What I mean is let say the input is 'æææ', will that work? Ideally it should return 'aeaeae' after removing accents. But is that what we are getting? I think this can handle a single such character, but when buffer is not available it won't return us the expected results. I looked into the tests, did not observed any such test.

test/JDBC/expected/non_default_server_collation/japanese_ci_as/test_like_for_AI-vu-verify.out
```diff
@@ -2815,6 +2815,7 @@ SELECT * FROM test_like_for_AI_prepare_t1_cs WHERE col_c LIKE 'cafe';
 GO
 ~~START~~
 nvarchar#!#varchar#!#text#!#ntext#!#char#!#nchar
+café#!#café#!#café#!#café#!#café                                              #!#café
```
ahmed-shameem, line 2818: This should not be the case as per our expectations. As char adds extra padding based on the typmod, in PG, they are not same. We so not return anything is case of CI_AS and CS_AS. So, are we okay with this?

test/JDBC/expected/non_default_server_collation/japanese_ci_as/test_like_for_AI-vu-verify.out
```diff
@@ -2847,6 +2848,7 @@ SELECT * FROM test_like_for_AI_prepare_t1_cs WHERE col_nchar LIKE 'cafe';
...
```
ahmed-shameem, line 2851: Same as above.

ahmed-shameem (changes requested):

contrib/babelfishpg_tsql/src/collation.c
kuntalghosh: Besides, this is unsafe coding.
```
size = Min((len_uinput + 1) * sizeof(UChar), capacity * sizeof(UChar);
utf16_res = (UChar *) palloc0(size);
memcpy(utf16_res, utf16_input, size);
```

kuntalghosh (changes requested):

contrib/babelfishpg_tsql/src/collation.c
Deepesh125: @ahmed-shameem 
> (len_uinput + 1) * sizeof(UChar) < capacity * sizeof(UChar)


Mathematically this will always be true ` (len_uinput + 1) * sizeof(UChar) < capacity * sizeof(UChar)` ( except one case where len_uinput = INT32_MAX which will not occur in postgres) 

For the example you are referring,  'æææ', len_uinput will be 3 and capacity will be 12. And we don't have exact that test case but we have similar test cases. In fact all the strings in new tests will be expanded upon removal of accents.

Deepesh125: @kuntalghosh we don't want to do that, utf16_res should have enough space to accommodate any expansion of the string needed. This expansion will not go beyond `len_uinput  * MAX_BYTES_PER_CHAR`. 

So, we need to allocate enough space required `len_uinput  * MAX_BYTES_PER_CHAR` or INT32_MAX (whatever is max). But we only want to copy `len_uinput + 1 * sizeof(UChar)` otherwise it may cause buffer overrun.

tanscorpio7: So 1 UTF-16 character can at the most be broken down into 4 base characters ?

contrib/babelfishpg_tsql/src/collation.c
```diff
@@ -493,7 +494,7 @@ Datum remove_accents_internal(PG_FUNCTION_ARGS)
 
 		// Open transliterator
 		cached_transliterator = utrans_openU(rules, len_uchar, UTRANS_FORWARD, NULL, 0, NULL, &status);
-		if (U_FAILURE(status))
+		if (U_FAILURE(status) || !cached_transliterator)
 		{
 			ereport(ERROR,
```
tanscorpio7, line 500: Should we free `rules` before error, since it is in topmemory context

contrib/babelfishpg_tsql/src/collation.c
```diff
@@ -505,40 +506,63 @@ Datum remove_accents_internal(PG_FUNCTION_ARGS)
...
+	 * value which would be equivalent to 2GB UChar points and 2GB * sizeof(UChar) in byte terms.
+	 * XXX: It is assumed that this capacity should handle almost all the general input strings.
 	 */
 	capacity = (limit < (PG_INT32_MAX / MAX_BYTES_PER_CHAR)) ? (limit * MAX_BYTES_PER_CHAR) : PG_INT32_MAX;
 
+	/*
+	 * utrans_transUChars will modify input string in place so ensure that it has enough capacity to store
+	 * transformed string.
+	 */
+	utf16_res = (UChar *) palloc0(capacity * sizeof(UChar));
```
tanscorpio7, line 537: palloc limit seems to be
```
#define MaxAllocSize	((Size) 0x3fffffff) /* 1 gigabyte - 1 */

#define AllocSizeIsValid(size)	((Size) (size) <= MaxAllocSize)
```
capacity * sizeof(UChar) can go upto 4GB.

tanscorpio7: We can also probably change this ```Min((len_uinput + 1) * sizeof(UChar), capacity * sizeof(UChar))``` to ```(len_uinput + 1) * sizeof(UChar)```
Because ```len_uinput + 1``` will never exceed the limit of palloc, since ```icu_to_uchar``` tries to palloc the same length.

contrib/babelfishpg_tsql/src/collation.c
Deepesh125: @tanscorpio7 yes!

contrib/babelfishpg_tsql/src/collation.c
Deepesh125: we want to reuse rules, that's why we are caching it in top memory context

Deepesh125: Good catch! But we dont need to free it as it would not have initialised in case of error.

Deepesh125: [Doc](https://unicode-org.github.io/icu-docs/apidoc/dev/icu4c/utrans_8h.html#a2fc501ae07ce3ce56b164db8636e9f12) of utrans_openU says 
```
Returns
    a transliterator pointer that may be passed to other utrans_xxx() functions, or NULL if the open call fails.
````

test/JDBC/expected/non_default_server_collation/japanese_ci_as/test_like_for_AI-vu-verify.out
Deepesh125: As discussed, this is intended T-SQL behaviour so we will go with this.

contrib/babelfishpg_tsql/src/collation.c
Deepesh125: This is good point. With the latest revision, we are limiting the size of input that this function can process. For longer term, we will try to chunk the input string into smaller parts, process it and concat back the final string.

tanscorpio7 (approved):

ahmed-shameem (approved):

kuntalghosh (approved):

PR #2649, opened 2024-06-10
Fix valgrind warnings related to source and destination overlap in strcpy in datetime, datetime2 and datetimeoffset related functions

basasairohan:
### Description

In datetime_in_str function, we are unnecessarily using strcpy function which is also causing issues in valgrind as the source and the destination overlap. 

This commit removes this unnecessary strcpy call in datetime_in_str function. Also, we can directly send the original input string in clean_input_str() function rather than sending modified_str as we were doing before.

### Issues Resolved

BABEL-5034

Tested locally with Valgrind
* **Use case based -** N/A

* **Arbitrary inputs -** N/A

* **Performance tests -** N/A

* **Tooling impact -** N/A

* **Client tests -** N/A

### Check List

### DIFF
--- a/contrib/babelfishpg_common/src/datetime.c
+++ b/contrib/babelfishpg_common/src/datetime.c
@@ -174,7 +174,7 @@ check_regex_for_text_month(char *str, DateTimeContext context)
 char*
 clean_input_str(char *str, bool *contains_extra_spaces, DateTimeContext context)
 {
-	char *result = (char *) palloc(MAXDATELEN);
+	char *result = (char *) palloc(MAXDATELEN + 1);
 	int i = 0, j = 0;
 	int last_non_space = -1;
 	int num_colons = 0;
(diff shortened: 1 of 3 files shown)


contrib/babelfishpg_common/src/datetime.c
```diff
@@ -196,7 +196,7 @@ clean_input_str(char *str, bool *contains_extra_spaces, DateTimeContext context)
 		if (str[i] == '\0')
 			break;
 
-		if (j > MAXDATELEN)
+		if (j >= MAXDATELEN)
```
tanscorpio7, line 199: We are also doing j+=2 below.
Is this condition enough to make sure we do not overflow there ?

tanscorpio7: We are also adding a null byte at the end after the while loop ends. So should this be be MAXDATALEN - 1.
i.e. not writing on the last byte inside the while loop.

basasairohan: > We are also doing j+=2 below. Is this condition enough to make sure we do not overflow there ?

Yes, this condition should be sufficient to make sure that there is no overflow.

> We are also adding a null byte at the end after the while loop ends. So should this be be MAXDATALEN - 1.
i.e. not writing on the last byte inside the while loop.

We can update the size of result to MAXDATALEN + 1 instead.

tanscorpio7 (approved):

shardgupta (approved):

PR #2677, opened 2024-06-20
Optimise lookup for babelfish_function_ext catalog

rishabhtanwar29:
### Description
Extended catalog `sys.babelfish_function_ext` currently uses as its key the schema name plus the function signature. This dates back to the time when Babelfish did not yet block duplicate functions with different signatures, but that is blocked now.
When doing a lookup for this catalog, the function signature must first be constructed in `get_pltsql_function_signature_internal()`. This uses a GUC to ensure that non-system names are double-quoted (required for the lookup), which appears to be a relatively expensive operation.

Since Babelfish no longer supports function overloading, we can perform a lookup directly using the schema name + function name. This requires an updated primary key on the catalog as (funcname + nspname + funcsignature). With this, we now will first perform lookup with partial key (funcname + nspname) and return the tuple if there is a unique match which will save us from constructing function signature. In case of multiple matches, we will then perform the lookup using complete key (funcname + nspname + funcsignature).

This optimization improves the performance which was previously impacted by WITH RECOMPILE which was added as part of https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/2563 . It introduced a lookup into babelfish_function_ext catalog using get_bbf_function_tuple_from_proctuple which performs the signature-based lookup for every stored procedure call.

### Performance data
Below table shows the time taken (in ms) for 100000 executions of the a procedure,  randomly picking a procedure each time from 10000 existing procedures in the database:

| S.N. | Scenario                           | Sample 1 | Sample 2 | Sample 3  | Sample 4 | Sample 5 | Sample 6 | Average  |
| :--- | :----------------------- | :-------- | :-------- | :--------- | :-------- | :-------- | :-------- | :-------- |
| 1      | Without current changes | 34785    | 32008      | 32079       | 32018      | 31957      | 32239     | 32514.33 |
| 2     | With current changes     | 31205      | 28574      | 28744       | 28749      | 28433     | 28598      | 29050.5  |
| 3     | Without RECOMPILE       | 30723     | 27920      | 27998       | 27876      | 28089     | 27938      | 28424     |

Performance gain with current changes (b/w SN1 and SN2): 10.65%
% perf loss with RECOMPILE (b/w SN2 and SN3): 2.2%

Task: BABEL-4976

Engine PR: https://github.com/babelfish-for-postgresql/postgresql_modified_for_babelfish/pull/393

* **Use case based -**

* **Boundary conditions -**

* **Arbitrary inputs -**

* **Negative test cases -**

* **Performance tests -**

* **Tooling impact -**

* **Client tests -**

### Check List


contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--4.2.0--4.3.0.sql
```diff
@@ -37,6 +37,9 @@ LANGUAGE plpgsql;
  * So make sure that any SQL statement (DDL/DML) being added here can be executed multiple times without affecting
  * final behaviour.
  */
+SET allow_system_table_mods = ON;
```
surendravishnoi, line 43: add a comment explaining why this is required

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--4.2.0--4.3.0.sql
```diff
@@ -37,6 +37,10 @@ LANGUAGE plpgsql;
...
+ALTER TABLE sys.babelfish_function_ext DROP CONSTRAINT babelfish_function_ext_pkey;
+ALTER TABLE sys.babelfish_function_ext ADD PRIMARY KEY (funcname, nspname, funcsignature);
```
Deepesh125: Let's use name constraint just to be on safer side

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--4.2.0--4.3.0.sql
```diff
@@ -37,6 +37,10 @@ LANGUAGE plpgsql;
  * So make sure that any SQL statement (DDL/DML) being added here can be executed multiple times without affecting
  * final behaviour.
  */
+SET allow_system_table_mods = ON;
+ALTER TABLE sys.babelfish_function_ext DROP CONSTRAINT babelfish_function_ext_pkey;
```
surendravishnoi, line 44: do we have upgrade test with entries in babelfish_function_ext which covers this use case?

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--4.2.0--4.3.0.sql
rishabhtanwar29: Done!

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--4.2.0--4.3.0.sql
rishabhtanwar29: Yeah several tests in which we are creating functions/procedure, in-fact function/procedures are quite abundant in our upgrade tests so the scenario of catalog dropping primary key when catalog is not empty is being covered.

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--4.2.0--4.3.0.sql
rishabhtanwar29: Primary key constraint name is always fixed.

robverschoor: This works:
`alter table mytable add constraint my_pk primary key (mycol)`

surendravishnoi (approved):

PR #2679, opened 2024-06-24
Cross-db reference of sys* catalogs with dbo schema

robverschoor:
### Description

Cross-database references to the classic `sys*` catalogs (e.g. `sysobjects`, `syscolumns`, `syslogins`, etc.), while specifying the `dbo` schema or omitting the schema (e.g. `db..sysobjects` notation), should not raise an error. For these 'classic' catalogs, referencing the object in `dbo` or `sys` schema should be equivalent. 
This fix addresses the following notations which either raised an error or could return an incorrect value:

```
- mydb.dbo.sysobjects
- .mydb.dbo.sysobjects
- mydb..sysobjects
- .mydb..sysobjects
- .mydb.sys.sysobjects

- .mydb.dbo.mytable (for non-catalog objects)
- .object (for any object in case the executing user has a non-dbo default schema)
```

Various existing test cases, which were raising an error previously, had to be adjusted for this fix.

Note: when the executing user has a non-`dbo` default schema, the case of `dbname..object` still ignores the default schema, but that is outside the scope of this fix (see BABEL-5050).

### Issues Resolved

BABEL-5041 Cross-DB system view query is not currently supported in Babelfish

- Use case based - Yes

- Boundary conditions - N/A

- Arbitrary inputs - N/A

- Negative test cases - Yes

- Performance tests - N/A

- Tooling impact - N/A

- Client tests - N/A

### Check List


kuntalghosh (dismissed):

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -1579,6 +1579,87 @@ is_created_with_recompile(Oid objectId)
...
+	    (pg_strncasecmp(name, "syslanguages", 12) == 0) ||
+	    (pg_strncasecmp(name, "syslogins", 9) == 0) ||
+	    (pg_strncasecmp(name, "sysprocesses", 12) == 0) ||
+
+	    // DB-specific classic catalogs
+	    (pg_strncasecmp(name, "syscolumns", 10) == 0) ||
+	    (pg_strncasecmp(name, "sysforeignkeys", 14) == 0) ||
+	    (pg_strncasecmp(name, "sysindexes", 10) == 0) ||
+	    (pg_strncasecmp(name, "sysobjects", 10) == 0) ||
+	    (pg_strncasecmp(name, "systypes", 8) == 0) ||
```
HarshLunagariya: Let's include length comparison as well. Current implementation will return true for "valid_view_name + any_prefix" as well.

contrib/babelfishpg_tsql/src/tsqlIface.cpp
```diff
@@ -2345,13 +2370,27 @@ class tsqlBuilder : public tsqlCommonMutator
...
+			{
+				db_name = stripQuoteFromId(ctx->database);
 
-			if (!string_matches(db_name.c_str(), get_cur_db_name()))
-				is_cross_db = true;
+				if (!string_matches(db_name.c_str(), get_cur_db_name()))
+					is_cross_db = true;
+
+				if (setSysSchema)
+					schema_name = "sys";
```
HarshLunagariya, line 2392: We are now rewriting the query at parser level(earlier it was at analyser/rewriter phase). Will this affect any object definition stored? For example, If we create view `create view v1 as select * from dbo.sysdatabases`, will it store the rewritten definition as `create view v1 as select * from sys.sysdatabases` ?

contrib/babelfishpg_tsql/src/tsqlIface.cpp
```diff
@@ -8043,48 +8082,98 @@ rewrite_information_schema_to_information_schema_tsql(T ctx, GetCtxFunc<T> getSc
...
+	bool catalog_need_sys_schema = false;
+	bool must_rewrite = false; // Handles case of .dbname..sometable for a non-catalog table
+
+	if (getObject != nullptr)
+	{
+		std::string objNameStripped = "";
+		objName = ::getFullText(getObject(ctx));
+		objNameStripped = stripQuoteFromId(objName);
+		schema_is_dbo_or_sys = (pg_strncasecmp(stripQuoteFromId(::getFullText(schema)).c_str(), "dbo", 3) == 0) ||
+		                       (pg_strncasecmp(stripQuoteFromId(::getFullText(schema)).c_str(), "sys", 3) == 0) ||
```
HarshLunagariya: Here also length comparison is required along with pg_strncasecmp().

contrib/babelfishpg_tsql/src/tsqlIface.cpp
```diff
@@ -8043,48 +8082,98 @@ rewrite_information_schema_to_information_schema_tsql(T ctx, GetCtxFunc<T> getSc
...
+	std::string objName = "";
+	bool schema_is_dbo_or_sys = false;
+	bool catalog_need_sys_schema = false;
+	bool must_rewrite = false; // Handles case of .dbname..sometable for a non-catalog table
+
+	if (getObject != nullptr)
+	{
+		std::string objNameStripped = "";
+		objName = ::getFullText(getObject(ctx));
+		objNameStripped = stripQuoteFromId(objName);
```
HarshLunagariya, line 8099: Why are we not using `std::string stripQuoteFromId(TSqlParser::IdContext *ctx)` i.e. `objNameStripped = stripQuoteFromId(getObject(ctx));` ?

test/JDBC/expected/BABEL-2455.out
```diff
@@ -261,16 +261,8 @@ go
 -- trigger
 CREATE TRIGGER .tr2455 on t2455_base AFTER INSERT AS print 'triggered';
 go
-~~ERROR (Code: 33557097)~~
```
HarshLunagariya, line 264: Here, was it considering whole "dbo.tr2455" as trigger name?

test/JDBC/expected/catalogs_dbo_sys_schema-upgrade-vu-verify.out
```diff
@@ -0,0 +1,6872 @@
...
+
+select name from .myschematest_mydb..syslogins where name like 'login_myschematest_%' order by 1
+go
+~~START~~
+varchar
+login_myschematest_master
+login_myschematest_mydb
+login_myschematest_tempdb
+~~END~~
+
```
HarshLunagariya, line 1255: sysprocesses test cases are missing.

test/JDBC/expected/catalogs_dbo_sys_schema-upgrade-vu-verify.out
```diff
@@ -0,0 +1,6872 @@
...
+~~START~~
+smallint
+~~END~~
+
+select cast(0 as smallint) from master.sys.sysforeignkeys where 0 = 1 order by 1
+go
+~~START~~
+smallint
+~~END~~
+
```
HarshLunagariya: Why are we returning 0 rows? Was there no way to add deterministic test?

HarshLunagariya (changes requested):

contrib/babelfishpg_tsql/src/tsqlIface.cpp
robverschoor: For view definitions, the stored statement is indeed rewritten (note that other aspects of views are also stored rewritten, for example `select *` is expanded to the full column list, but this is unrelated to this fix).
For procedures, the body is stored unchanged; only when the procedure is executed are the individual statements processed and the rewriting takes places.

contrib/babelfishpg_tsql/src/catalog.c
robverschoor: The name passed as an argument here does not contain a schema prefix.

contrib/babelfishpg_tsql/src/tsqlIface.cpp
robverschoor: The length is specified

contrib/babelfishpg_tsql/src/tsqlIface.cpp
robverschoor: The `GetCtxFunc<T> getObject` argument (which was added in this fix) resolves to different object types depending on which parser rule this is invoked from. That does not match with the original declaration of stripQuoteFromId(), so something had to be changed. Adding another copy of this proc with a string argument was the least amount of change since stripQuoteFromId() is already invoked from 30+ locations.

test/JDBC/expected/BABEL-2455.out
robverschoor: In T-SQL it is valid to specify leading dots, e.g.  `.myproc` or `..myproc` which are both equivalent to `myproc` . This form was one of the syntax variations that was not handled previously and led to an error, but is handled correctly now (see PR summary at the top of this page).

test/JDBC/expected/catalogs_dbo_sys_schema-upgrade-vu-verify.out
robverschoor: sysprocesses was removed because the contents are not predictable: sometimes process keep hanging around during the JDBC tests, but this varies. Will add test similar to sysforeignkeys where we avoid hitting any rows and just test for name resolution.

test/JDBC/expected/catalogs_dbo_sys_schema-upgrade-vu-verify.out
robverschoor: It appears that other test cases are leaving schema elements behind. So while this test case is cleaning things up, it appears that in github, rows show up which originate from other tests.

contrib/babelfishpg_tsql/src/catalog.c
HarshLunagariya: I meant to say that this function will return true for 'syscolumnsxyz' as well which is incorrect.

robverschoor: fixed

contrib/babelfishpg_tsql/src/tsqlIface.cpp
robverschoor: fixed

test/JDBC/expected/catalogs_dbo_sys_schema-upgrade-vu-verify.out
robverschoor: added test cases for syprocesses, similar to sysforeignkeys

contrib/babelfishpg_tsql/src/tsqlIface.cpp
robverschoor: added testcases for procedures and views

contrib/babelfishpg_tsql/src/catalog.c
HarshLunagariya: Current implementation won't work if supplied name is some prefix of valid system view name. For example, For name 'syscolu'  condition `(pg_strncasecmp(name, "syscolumns", strlen(name)) == 0)` will return true.

Ideally we can modify the check as following:
`strlen(name) == 10 && pg_strncasecmp(name, "syscolumns", strlen(name)) == 0`

HarshLunagariya (approved):

shardgupta (approved):

PR #2686, opened 2024-06-25
Fix concurrency issues with ENR temp tables.

Sairakan:
### Description

With the temp OID buffer introduced in [5a69710](https://github.com/babelfish-for-postgresql/postgresql_modified_for_babelfish/commit/5a69710c6dd38b58e2860a1f5b89edb5f4611e77), it is possible for temp tables from different sessions to share the same OID. Bypass locking for ENR temp tables in order to prevent these sessions from interfering with each other unnecessarily - they should not need to wait on each other to perform operations on their own temp tables.

Also, prevent invalidation messages for local temp tables from being added to the shared invalidation (SI) queue. Otherwise, under high concurrency we may see errors such as `relation XXXXX deleted while still in use` due to one backend receiving an invalidation message from a different backend for an OID which it is trying to use for its own temp table, when that table should still be valid.
 
### Issues Resolved

BABEL-4913, BABEL-4892

* **Use case based -**
lock contention testing via a test function in JDBCTempTable.java

* **Boundary conditions -**

* **Arbitrary inputs -**

* **Negative test cases -**

* **Performance tests -**

* **Tooling impact -**

* **Client tests -**

### Check List


contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -4659,6 +4671,19 @@ static bool set_and_persist_temp_oid_buffer_start(Oid new_oid)
...
 }
 
+static bool
+pltsql_is_local_only_inval_msg(const SharedInvalidationMessage *msg)
+{
+	return (msg->id == SHAREDINVALRELCACHE_ID && msg->rc.local_only);
+}
+
+static bool
+pltsql_is_enr_locktag(const LOCKTAG *locktag)
```
surendravishnoi: instead of this introduce a hook for get_ENR_withoid, locktag type can be checked on engine itself after hook null check

test/JDBC/expected/temp_oid.out
```diff
@@ -3,7 +3,7 @@ show babelfishpg_tsql.temp_oid_buffer_size
 GO
 ~~START~~
 text
-0
+65536
```
surendravishnoi, line 6: where do we have any tests for this?

Sairakan: I have been manually testing to validate that there is no lock contention, using the following test:

session 1:
```
begin tran
go
create table #t1 (a int)
go
```
session 2:
```
begin tran
go
create table #t1 (a int)
go
```
Without the engine changes enabled, the second session will hang until the first session either commits or aborts its transaction. With the engine enabled, the second session will successfully create the table.

contrib/babelfishpg_tsql/src/guc.c
```diff
@@ -1163,7 +1163,7 @@ define_custom_variables(void)
 							gettext_noop("Temp oid buffer size"),
 							NULL,
 							&temp_oid_buffer_size,
-							0, 0, 131072,
+							65536, 0, 131072,
```
lejaokri, line 1166: this seems out of place for this change. Is this expected?

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -4670,6 +4682,18 @@ static bool set_and_persist_temp_oid_buffer_start(Oid new_oid)
...
+static bool
+pltsql_is_local_only_inval_msg(const SharedInvalidationMessage *msg)
+{
+	return (msg->id == SHAREDINVALRELCACHE_ID && msg->rc.local_only);
+}
+
+static EphemeralNamedRelation
+pltsql_get_tsql_enr_from_oid(const Oid oid)
+{
+	return get_ENR_withoid(currentQueryEnv, oid, ENR_TSQL_TEMP);
```
lejaokri: nit: lets return right away if oid is invalid, no need to search for the entire array.

contrib/babelfishpg_tsql/src/guc.c
Sairakan: this is for testing purposes only. This will be reverted before the PR gets merged.

contrib/babelfishpg_tsql/src/hooks.c
Sairakan: I added a check for InvalidOid in get_ENR_withoid() on the engine side

test/JDBC/expected/temp_oid.out
Sairakan: Added a test in JDBCTempTable.java which does the above actions. I confirmed that the test fails when the hooks above are disabled, and succeeds when they are enabled.

contrib/babelfishpg_tsql/src/guc.c
timchang514: won't it cause the tests to fail? can we write the tests in a way that is independent of the actual GUC value set?

Sairakan: Reverting this won't cause the tests to fail, since there won't be any lock contention anyways.

contrib/babelfishpg_tsql/src/hooks.c
timchang514: we can also invoke this hook across the engine, wherever we're calling get_ENR_withoid. doesn't need to be in this change, just a followup to clean up code

timchang514 (approved):

lejaokri (dismissed):
remember to disable temp oid buffer size

test/JDBC/expected/temp_oid.out
surendravishnoi: can we have some tests based on pg catalog queries which cal validate locking assumption. Same for rel caches.

lejaokri (dismissed):

test/JDBC/src/main/java/com/sqlsamples/JDBCTempTable.java
```diff
@@ -334,6 +357,49 @@ private static void test_trigger_on_temp_table(BufferedWriter bw, Logger logger)
...
     }
+
+    private static void test_lock_contention(BufferedWriter bw, Logger logger) throws Exception {
+        String connectionString = initializeConnectionString();
+
+        /* do a sanity check of pg_locks catalog */
+        Connection c = DriverManager.getConnection(connectionString);
+        Statement s = c.createStatement();
+        s.execute("BEGIN TRAN");
+        int oid = create_table_and_report_oid(c, "CREATE TABLE #t1 (a int)", "#t1");
```
surendravishnoi: I think it should be regular JDBC sql test for lock validation. Also, we should add most DDLs which can impact temp tables like drop, alter etc.

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -360,6 +366,12 @@ InstallExtendedHooks(void)
 	prev_GetNewTempOidWithIndex_hook = GetNewTempOidWithIndex_hook;
 	GetNewTempOidWithIndex_hook = pltsql_GetNewTempOidWithIndex;
 
+	prev_pltsql_is_local_only_inval_msg_hook = pltsql_is_local_only_inval_msg_hook;
```
timchang514, line 369: The matching lines need to be added in UninstallExtendedHooks as well

surendravishnoi (approved):

contrib/babelfishpg_tsql/src/guc.c
timchang514: leaving this in the merge bc with this fix, we can reenable temp oid for testing

PR #2688, opened 2024-06-26
Include ordering for constraints on babelfish tables

rishabhtanwar29:
### Description
Babelfish supports column ordering and NULLs ordering in table constraints but PG
does not. It was implemented through special grammar rule and few dialect T-SQL dialect
dependent logic in engine. Currently this is not handled in pg_dump so the information
about column/nulls ordering gets lost during dump. Although, this information is available
in the underlying index for a UNIQUE/PRIMARY KEY constraint but this information anyway
gets lost since the underlying index does not get dumped for a UNIQUE/PRIMARY KEY constraint,
instead it is dumped as a constraint.

To fix this issue, we will now dump the underlying index of a UNIQUE/PRIMARY KEY constraint
and make use of ALTER TABLE ADD CONSTRAINT USING snytax to add that as constraint.
Postgres disallows creation of an constraint using index like this for non default sort ordering as
native pg_dump does not handle such constraints. So we will need to bypass that check for
Babelfish dump/restore with this special handling in bbf_dump.

Task: BABEL-4888

Engine PR: https://github.com/babelfish-for-postgresql/postgresql_modified_for_babelfish/pull/395

* **Use case based -**

* **Boundary conditions -**

* **Arbitrary inputs -**

* **Negative test cases -**

* **Performance tests -**

* **Tooling impact -**

* **Client tests -**

### Check List


test/JDBC/expected/BABEL-4888-vu-verify.out
```diff
@@ -0,0 +1,145 @@
...
+~~ERROR (Code: 515)~~
+
+~~ERROR (Message: null value in column "id1" of relation "babel_4888_t4" violates not-null constraint)~~
+
+
+INSERT INTO babel_4888_t4 VALUES (NULL, NULL, 0x9999)
+GO
+~~ERROR (Code: 515)~~
+
+~~ERROR (Message: null value in column "id1" of relation "babel_4888_t4" violates not-null constraint)~~
```
surendravishnoi, line 129: we should also validate both asc and desc ordering, null first and last ordering

rishabhtanwar29: Actually these combinations are already being tested. If you check the prepare file, table `babel_4888_t1` for example, has `id1` in `DESC` order and `id2` in `ASC` order.

surendravishnoi: but how are we validating that NULLs are actually at right places and index have right ordering?

rishabhtanwar29: We are verifying it with the index definition, check the SELECT command below.

sumitj824 (approved):

shardgupta (approved):

PR #2696, opened 2024-06-28
Add tests for crash due to uninitialized rte->relid in temp table with trigger

timchang514:
### Description

Fixing a crash in temp tables with trigger. `rte->relid` was uninitialized in `addRangeTableEntryForENR` when dealing with TSQL temp tables, causing an assert to fail. Adding tests for this change.

### Issues Resolved

BABEL-5071
BABEL-4992

### Check List

### DIFF
--- a/test/JDBC/input/BABEL-5071-vu-prepare.sql
+++ b/test/JDBC/input/BABEL-5071-vu-prepare.sql
@@ -1,0 +1,33 @@
+create table tab5071_1 (a int)
+go
+
+create trigger trig5071_1 on tab5071_1 for insert as begin
+create table #temp_tst (a int)
+insert #temp_tst select a*-1 from inserted;
+select * from #temp_tst;
+drop table #temp_tst
+end
+go
+
+CREATE TABLE tab4992_1 (c1 int, c2 int)
+GO
+
+CREATE TABLE tab4992_2 (c1 int, c2 int)
+GO
+
+INSERT INTO tab4992_1 VALUES(1, 2)
+GO
+
+CREATE TRIGGER tab4992_1deleteOperation on [dbo].[tab4992_1]
+after delete AS
+BEGIN
+            CREATE TABLE #tmp
+            (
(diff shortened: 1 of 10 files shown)


lejaokri (commented):
You might want to add this to some of the MVU schedule files to verify that a trigger with temp tables before the fix isnt broken after the upgrade. Otherwise this looks good! THanks for the quick fix!

lejaokri (dismissed):

test/JDBC/upgrade/latest/schedule
```diff
@@ -538,3 +538,4 @@ babel_4328_datetimeoffset
 babel_726
 BABEL-3401
 PARTITION
+BABEL-5071
```
2jungkook, line 541: Could you add this to other schedule files? For example, the prepare script is executed in a previous version and the verify script is executed in the latest version.

test/JDBC/input/BABEL-5071-vu-prepare.sql
```diff
@@ -0,0 +1,33 @@
...
+after delete AS
+BEGIN
+            CREATE TABLE #tmp
+            (
+                c1 INT PRIMARY KEY,
+                c2 INT,
+                c3 VARCHAR(64) NOT NULL DEFAULT ''
+            )
+            INSERT INTO #tmp VALUES (1, 2, '')
+            INSERT INTO tab4992_2 SELECT c1,c2 FROM #tmp --> Crash
```
2jungkook, line 31: Prepare script should not contain "Crash" as it should be able to be executed in previous versions.

timchang514: It won't crash in the prepare, it would only crash once the delete is executed (see verify.sql)

test/JDBC/upgrade/latest/schedule
timchang514: done

2jungkook (approved):

PR #2702, opened 2024-07-01
Support Table and Index Partitioning in Babelfish

sumitj824:
### Description
This commit introduces Support for Table and Index Partitioning in Babelfish, enabling users to partition large tables and indexes using partition schemes.

Key Changes:

1. Partitioned Table Support:
   - Added syntax support for creating partitioned tables using the `CREATE TABLE ... on PARTITION_SCHEME (column)` statement.
   - Implemented support for partitioned table using partition scheme, where child partitions are implicitly created by looking up the metadata of the partition function and partitioning column values.
   - Restricted InsertBulk operations for partitioned tables as they are not supported yet.
   - Restricted users from modifying Babelfish partitioned tables to attach, detach, or modify partitions of these tables to maintain data consistency in the Babelfish catalog. Existing users who have created partitioned tables from the PostgreSQL endpoint can continue to modify, attach, and detach partitions as usual.
   - Introduced sys.babelfish_partition_depend catalog to track the dependency between partitioned tables and the partition schemes used to create them. The catalog maintains consistent table name even after RENAME or DROP TABLE operations.

2. Partitioned Index Support:
   - Added syntax support for creating partitioned indexes using the `CREATE INDEX ... ON PARTITION_SCHEME (column)` statement.
   - Implemented support for creating aligned partitioned indexes, where the index partitioning scheme aligns with the underlying table partitioning scheme.
   - Restricted the ability to create non-aligned partitioned indexes with custom partition schemes.

3. System Catalog Views:
   - Modified system catalog views to exclude listing child partitions for partitioned tables and indexes.
   - Modified system catalog views to list the partitioned table along with regular tables.
   - These changes ensures that only the parent partitioned object is listed in the relevant system views

Signed-off-by: Sumit Jaiswal [sumiji@amazon.com](mailto:sumiji@amazon.com)

Engine PR: https://github.com/babelfish-for-postgresql/postgresql_modified_for_babelfish/pull/399

### Issues Resolved

Task: BABEL-4968, BABEL-4969

* **Use case based -** Yes

* **Arbitrary inputs -** Yes

* **Performance tests -** NA

* **Tooling impact -** NA

* **Client tests -** NA

### Check List


contrib/babelfishpg_tsql/src/backend_parser/gram-tsql-rule.y
```diff
@@ -2574,6 +2575,70 @@ tsql_stmt :
...
+tsql_partition_scheme_name:
+			IDENT
+				{
+					/*
+					 * If the length of the identifier ($1) is equal to NAMEDATALEN - 1,
+					 * which is the maximum allowed length for an identifier in PostgreSQL,
+					 * it means the identifier might have been truncated by the scanner
+					 * due to the length limit.
+					 */
+					if (strlen($1) == NAMEDATALEN - 1)
```
Deepesh125: When you have multi byte chars (Chinese chars) then this logic may not hold true.

contrib/babelfishpg_tsql/src/backend_parser/gram-tsql-rule.y
```diff
@@ -2574,6 +2575,70 @@ tsql_stmt :
...
+					n->tsql_partition_scheme = $2;
+					n->strategy = PARTITION_STRATEGY_RANGE;
+					n->partParams = $4;
+					n->location = @1;
+					$$ = n;
+				}
+		;
+
+tsql_partition_scheme_name:
+			IDENT
```
Deepesh125, line 2630: is it possible to define TSQL specific IDENT where we can support iedntifier of length >64 and < 128

contrib/babelfishpg_tsql/src/backend_parser/gram-tsql-rule.y
```diff
@@ -2574,6 +2575,70 @@ tsql_stmt :
...
+					n->options = $11;
+					n->oncommit = ONCOMMIT_NOOP;
+					n->tablespacename = NULL;
+					n->if_not_exists = false;
+					$$ = (Node *) n;
+				}
+		;
+
+tsql_PartitionSpec:
+			ON tsql_partition_scheme_name '(' part_params ')'
```
Deepesh125: "ON" should part of parent statement?

contrib/babelfishpg_tsql/src/backend_parser/gram-tsql-rule.y
```diff
@@ -2574,6 +2575,70 @@ tsql_stmt :
...
+					n->options = $11;
+					n->oncommit = ONCOMMIT_NOOP;
+					n->tablespacename = NULL;
+					n->if_not_exists = false;
+					$$ = (Node *) n;
+				}
+		;
+
+tsql_PartitionSpec:
+			ON tsql_untruncated_IDENT '(' part_params ')'
```
Deepesh125, line 2607: Shouldn't this "ON" part of parent statement?

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -3083,8 +3083,6 @@ rename_object_update_bbf_schema_permission_catalog(RenameStmt *stmt, int rename_
 		pfree(physical_schema_name);
 	if (logical_schema_name != NULL)
 		pfree(logical_schema_name);
-	if (object_name != NULL)
-		pfree(object_name);
```
Deepesh125, line 3199: why this change?

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -4236,6 +4259,12 @@ exec_stmt_partition_function(PLtsql_execstate *estate, PLtsql_stmt_partition_fun
 		/* evaluate the value from the expr */
 		val = exec_eval_expr(estate, list_nth(arg, i), &isnull, &valtype, &valtypmod);
 
+		/* raise error for null value */
+		if (isnull)
```
Deepesh125, line 4264: how did we detect this? is there any test cases added?

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -2527,6 +2543,97 @@ bbf_ProcessUtility(PlannedStmt *pstmt,
 						}
 					}
 				}
+				
+				if (!babelfish_dump_restore && atstmt->objtype == OBJECT_TABLE)
```
Deepesh125: need comment explaining this part of code

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -2527,6 +2543,97 @@ bbf_ProcessUtility(PlannedStmt *pstmt,
...
+							char		*db_name = get_cur_db_name();
+							const char	*user = get_user_for_database(db_name);
+				
+							logical_schema_name = get_authid_user_ext_schema_name(db_name, user);
+							pfree(db_name);
+						}
+						else /* Schema is explicitly specified for TDS client. */
+						{
+							logical_schema_name = (char *) get_logical_schema_name(atstmt->relation->schemaname, true);
+							if (!logical_schema_name) /* pg_temp schema */
```
Deepesh125: If it could be only pg_temp then should we directly do string compare? and get logical schema name for others by setting missing_ok = true. it will be better handling

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -2527,6 +2543,97 @@ bbf_ProcessUtility(PlannedStmt *pstmt,
...
+					}
+					else
+						break;
+					
+					/*
+					 * For babelfish partitioned table, user should not be
+					 * allowed to attach/detach of table directly.
+					 */
+					if (cmd->subtype == AT_AttachPartition || cmd->subtype == AT_DetachPartition 
+						|| cmd->subtype == AT_DetachPartitionFinalize)
```
Deepesh125: Just curios -- what is difference b/w AT_DetachPartition and AT_DetachPartitionFinalize?

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -2527,6 +2543,97 @@ bbf_ProcessUtility(PlannedStmt *pstmt,
...
+							ReleaseSysCache(tuple);
+						}
+
+						if (!is_partition_table)
+						{
+							pfree(logical_schema_name);
+							break;
+						}
+
+						parent_table_name = get_rel_name(get_partition_parent(relid, true));
```
Deepesh125: In Babelfish, will we ever have partitioned table detached from parent table? And why are we passing even_if_detached = true here?

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -6391,6 +6579,690 @@ bbf_ExecDropStmt(DropStmt *stmt)
...
+				ereport(ERROR,
+					(errcode(ERRCODE_UNDEFINED_OBJECT),
+						errmsg("Cannot drop the babelfish partition table '%s'.", relname)));
+			}
+		}
+		else if (form->relkind == RELKIND_PARTITIONED_TABLE) /* relation is partitioned table */
+		{
+			/* In case of babelfish partitioned table then remove the entry from bbf_partition_depend table. */
+			if (is_bbf_partitioned_table(dbid, logical_schemaname, relname))
+				remove_entry_from_bbf_partition_depend(dbid, logical_schemaname, relname);
```
Deepesh125: What will happen if we remove the entry here and then it runs into some other other during standard_processUtility?

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -6391,6 +6579,690 @@ bbf_ExecDropStmt(DropStmt *stmt)
...
+			relation_close(relation, AccessShareLock);
+			continue;
+		}
+
+		if (form->relispartition) /* relation is partition of table */
+		{
+			/* Raise error when parent table of partition is babelfish partitioned table. */
+			if (is_bbf_partitioned_table(dbid, logical_schemaname, get_rel_name(get_partition_parent(form->oid, true))))
+			{
+				ereport(ERROR,
```
Deepesh125: relation_close(relation, AccessShareLock);

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -6391,6 +6579,690 @@ bbf_ExecDropStmt(DropStmt *stmt)
...
+
+/*
+ * bbf_drop_partitioned_table
+ * 	When the table is being dropped is:
+ * 	1. babelfish partitioned table, then it removes the entry from bbf_partition_depend catalog.
+ * 	2. partition of babelfish partitioned table, then it throws an error.
+ * 	3. other than above, then it does nothing.
+ */
+static void
+bbf_drop_partitioned_table(DropStmt *stmt)
```
Deepesh125: better naming would be handle_drop_babelfish_partitioned_table or something like that

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -6391,6 +6579,690 @@ bbf_ExecDropStmt(DropStmt *stmt)
...
+	Oid		inhrelid;
+	HeapTuple	tuple;
+	const char	*old_dialect;
+	StringInfoData	query;
+	char		*partition_name;
+	char		*new_partition_name;
+
+	logical_schema_name = (char *) get_logical_schema_name(physical_schema_name, true);
+
+	if (!logical_schema_name) /* not a tsql schema */
```
Deepesh125: use T-SQL here and everywhere else

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -6391,6 +6579,690 @@ bbf_ExecDropStmt(DropStmt *stmt)
...
+		set_config_option("babelfishpg_tsql.sql_dialect", old_dialect, GUC_CONTEXT_CONFIG,
+					PGC_S_SESSION, GUC_ACTION_SAVE, true, 0, false);
+	}
+	PG_END_TRY();
+
+	/* Need to make a wrapper PlannedStmt. */
+	wrapper = makeNode(PlannedStmt);
+	wrapper->commandType = CMD_UTILITY;
+	wrapper->canSetTag = false;
+	wrapper->utilityStmt = (Node *) rename_partition_stmt;
```
Deepesh125: It is possible that rename_partition_stmt might be NULL here, we should detect such cases

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -6391,6 +6579,690 @@ bbf_ExecDropStmt(DropStmt *stmt)
...
+			BTEqualStrategyNumber, F_OIDEQ,
+			ObjectIdGetDatum(parentrelid));
+
+	scan = systable_beginscan(relation, InheritsParentIndexId, true,
+					NULL, 1, &key);
+
+	while ((tuple = systable_getnext(scan)) != NULL)
+	{
+		inhrelid = ((Form_pg_inherits) GETSTRUCT(tuple))->inhrelid;
+		partition_names = lappend(partition_names, get_rel_name(inhrelid));
```
Deepesh125: get_rel_name returns palloc'd copy so we should clean the list using list_free_deep api

Deepesh125 (changes requested):

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -6391,6 +6579,690 @@ bbf_ExecDropStmt(DropStmt *stmt)
...
+	char		*logical_schema_name;
+	ArrayType	*values;
+	bool		isnull;
+	int		i;
+	LOCAL_FCINFO(fcinfo, 1);
+
+	/* Strip trailing whitespaces from provided partition scheme. */
+	i = strlen(partition_scheme_name);
+	while (i > 0 && isspace((unsigned char) partition_scheme_name[i - 1]))
+		partition_scheme_name[--i] = '\0';
```
Deepesh125: is "valid  " valid scheme name?

contrib/babelfishpg_tsql/src/catalog.c
sumitj824: The `object_name` variable is reference of `stmt->relation->relname` and it shouldn't be freed. This was causing subsequent operations involving `stmt->relation->relname` to fail.

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -6391,6 +6579,690 @@ bbf_ExecDropStmt(DropStmt *stmt)
...
+	{
+		const char *user = get_user_for_database(db_name);
+		logical_schema_name = get_authid_user_ext_schema_name(db_name, user);
+		physical_schema_name = get_physical_schema_name(db_name, logical_schema_name);
+	}
+	else
+	{
+		physical_schema_name = pstrdup(stmt->relation->schemaname);
+		logical_schema_name = (char *) get_logical_schema_name(physical_schema_name, false);
+	}
```
Deepesh125: This code is being repeated some other places also, can we have helper function to do this?

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -6391,6 +6579,690 @@ bbf_ExecDropStmt(DropStmt *stmt)
...
+
+	partition_stmt = create_partition_stmt(physical_schema_name, relname);
+
+	/* Make a wrapper PlannedStmt. */
+	wrapper = makeNode(PlannedStmt);
+	wrapper->commandType = CMD_UTILITY;
+	wrapper->canSetTag = false;
+	wrapper->utilityStmt = (Node *) partition_stmt;
```
Deepesh125: handle case when partition_stmt is NULL

contrib/babelfishpg_tsql/src/backend_parser/gram-tsql-rule.y
sumitj824: It is based on PG node `PartitionSpec` where we have substituted the keyword `PARTITION BY` with `ON`. I think combining the `ON` keyword with the scheme name in `tsql_PartitionSpec` seems to provide a more coherent and understandable representation.

contrib/babelfishpg_tsql/src/backend_parser/gram-tsql-rule.y
sumitj824: `tsql_untruncated_IDENT` is genric TSQL specific IDENT and it can be used to find untruncated identifiers for any length. We cannot restrict it only to length >64 and < 128.

contrib/babelfishpg_tsql/src/backend_parser/gram-tsql-rule.y
sumitj824: Ohh right, for multil byte chars the truncated length can be less than max allowed length(NAMEDATALEN). Will remove this check in next commit.

contrib/babelfishpg_tsql/src/pl_exec-2.c
sumitj824: Ideally, `NULL` values should be allowed. However, due to the challenges we encountered in determining the appropriate boundary for storing the null partition, we made the decision to block NULL values from being included in the boundary list. Note that: this limitation does not imply that the partition column itself cannot contain NULL values; it only restricts NULL values from appearing in the boundary list definition. Yes, we do have testcases for the same.

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -6391,6 +6579,690 @@ bbf_ExecDropStmt(DropStmt *stmt)
...
+	wrapper->stmt_location = 0;
+	wrapper->stmt_len = 0;
+
+	/* Construct hash based on partitioned table name. */
+	unique_hash = construct_unique_hash(relname);
+
+	for (i = 0; i < nelems + 1; i++)
+	{
+		char *partition_name =  psprintf("%s_partition_%d", unique_hash, i);
```
Deepesh125: Make it macro

contrib/babelfishpg_tsql/src/pl_handler.c
sumitj824: We do have other cases like non-existing schema and we are only interested in existing TSQL schema. I will update the comment which reflect the same.

contrib/babelfishpg_tsql/src/pl_handler.c
sumitj824: Since, here we are blocking the modify operation on partition of babelfish partitioned so we are interested for all the partition even if it deatched. However, you are right that we will not have such case in babelfish since detach is blocked. I will fix this in next commit.

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -6391,6 +6579,690 @@ bbf_ExecDropStmt(DropStmt *stmt)
...
+
+	/* Search for the column specified with partition scheme in table's columns. */
+	tuple = SearchSysCacheAttName(relid, colname);
+
+	/* Raise an error if column specified with partition scheme doesn't exists in table. */
+	if (!HeapTupleIsValid(tuple))
+	{
+		ereport(ERROR,
+				(errcode(ERRCODE_UNDEFINED_COLUMN),
+					errmsg("column '%s' does not exist", colname)));
```
Deepesh125: Need to add details of table in error message

contrib/babelfishpg_tsql/src/pl_handler.c
sumitj824: The changes made to the catalog will be roll-backed if we enocountered any issue while executing the statment.

contrib/babelfishpg_tsql/src/pl_handler.c
sumitj824: Do you mean to use DIALECT check here? If so, then the changes are dialect independent. 
OR
Do you mean to use `tsql_schema_name` instead of `logical_schema_name`? If so, then I think everywhere we follow to use `logical_schema_name`.

contrib/babelfishpg_tsql/src/pl_handler.c
sumitj824: No, it is not possible `rename_partition_stmt` to be NULL. The `parsetree_nth_stmt` will lead assertion error in case of any such failure.

contrib/babelfishpg_tsql/src/pl_handler.c
sumitj824: We have already freed each of names individually in next for loop.

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -6391,6 +6579,690 @@ bbf_ExecDropStmt(DropStmt *stmt)
...
+					errmsg("Invalid object name '%s'.", partition_scheme_name)));
+	}
+
+	/* Find the partition scheme used to create partitioned table. */
+	partition_scheme_used_for_table = get_partition_scheme_for_partitioned_table(dbid, logical_schema_name, relname);
+	pfree(logical_schema_name);
+
+	/* partition_scheme_used_for_table will be null for non-partitioned table */
+	if (!partition_scheme_used_for_table ||
+			strcmp(partition_scheme_name, partition_scheme_used_for_table) != 0)
```
Deepesh125: pg_strncasecmp needed?

Deepesh125 (changes requested):

contrib/babelfishpg_tsql/src/pl_handler.c
sumitj824: as discussed earlier, it is not possible partition_stmt to be NULL. The parsetree_nth_stmt will lead assertion error in case of any such failure.

contrib/babelfishpg_tsql/src/pl_handler.c
sumitj824: At one place, we are interested in logical_schema_name and physical_schema_name with missing false while at second place we are interested only in logical_schema_name with missing true. So, we cannot combine both into a single function.

contrib/babelfishpg_tsql/src/pl_handler.c
sumitj824: It is valid in TSQL, I have handled the same and have added testcases as well.

contrib/babelfishpg_tsql/src/pl_handler.c
sumitj824: It is as per TSQL similar to error in simple create index when column is not present in table.

contrib/babelfishpg_tsql/src/catalog.c
shah-nirmit: Thanks for the fixing the bug

contrib/babelfishpg_tsql/src/pl_handler.c
sumitj824: DETACH PARTITION separates the partition from its parent table and any constraints, indexes, or triggers that were automatically created for the partition when it was part of the partitioned table are automatically dropped. While, DETACH PARTITION FINALIZE only the detach the partition and doesn't drop anything so that the detached table can be used as a standalone table.

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -3500,6 +3612,9 @@ bbf_ProcessUtility(PlannedStmt *pstmt,
 			{
 				DropStmt   *drop_stmt = (DropStmt *) parsetree;
 
+				if (drop_stmt->removeType == OBJECT_TABLE)
```
shah-nirmit, line 3513: Do we need to add a new function for OBJECT_TABLE, can we use/enhance bbf_ExecDropStmt?

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -6391,6 +6584,681 @@ bbf_ExecDropStmt(DropStmt *stmt)
...
+	 * if provided partition scheme exists in current database.
+	 */
+	partition_function_name = get_partition_function_name(dbid, partition_scheme_name);
+	
+	/* Raise error if provided partition scheme doesn't exists in current database. */
+	if (!partition_function_name)
+	{
+		ereport(ERROR, 
+			(errcode(ERRCODE_UNDEFINED_OBJECT), 
+				errmsg("Invalid partition scheme '%s' specifed.", partition_scheme_name)));
```
shah-nirmit: partition_function_name?

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -6391,6 +6584,681 @@ bbf_ExecDropStmt(DropStmt *stmt)
...
+
+	md5 = (char *) palloc(MD5_HASH_LEN + 1);
+
+	success = pg_md5_hash(relation_name, strlen(relation_name), md5, &errstr);
+	
+	if (unlikely(!success)) /* Out of memory. */
+	{
+		ereport(ERROR,
+				(errcode(ERRCODE_INTERNAL_ERROR),
+				 errmsg("constructing unique hash name failed for relation \"%s\": %s", relation_name, errstr)));
```
shah-nirmit: Nit: "Constructing unique partition name"

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -6391,6 +6584,681 @@ bbf_ExecDropStmt(DropStmt *stmt)
...
+		partbound->lowerdatums = list_make1(node);
+	}
+
+	/* Set upper bound of partition. */
+	if (idx == total_partitions - 1) /* last partition */
+	{
+		ColumnRef *node = makeNode(ColumnRef);
+		node->fields = list_make1(makeString("maxvalue"));
+		partbound->location = -1;
+		partbound->is_default = false;
```
shah-nirmit: Nit: Do we need to explicitly set location/is_default. It was one time handled in lower bound setting right?

contrib/babelfishpg_tsql/src/pltsql_bulkcopy.c
```diff
@@ -561,7 +561,6 @@ ExecuteBulkCopy(BulkCopyState cstate, int rowCount, int colCount,
 	 */
```
shah-nirmit, line 592: Nit: Please update the comment as well

test/JDBC/input/PARTITION-vu-verify.mix
shah-nirmit, line 1: Can we add few negative test for not being able to list child partitions in metadata views, for the changes you added to views

shah-nirmit (changes requested):

contrib/babelfishpg_tsql/src/pl_handler.c
sumitj824: `bbf_ExecDropStmt` function lacks the capability to handle drop operations from PSQL dialects, which may result in inconsistencies in the metadata and it will need a separate effort to fix that.

contrib/babelfishpg_tsql/src/pl_handler.c
sumitj824: The API `get_partition_function_name` returns partition function name corresponding to partition scheme and if partition scheme doesn't exists it will return NULL. Hence, the error message it correct.

contrib/babelfishpg_tsql/src/pl_handler.c
sumitj824: is_default is needed, removed location in next commit.

shah-nirmit (approved):

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -6838,7 +6838,7 @@ create_partition_stmt(char *physical_schema_name, char *relname)
 	 * This will be executed using ProcessUtility().
 	 */
 	initStringInfo(&query);
-	appendStringInfo(&query, "CREATE TABLE %s.dummy PARTITION OF %s.%s"
+	appendStringInfo(&query, "CREATE TABLE \"%s\".dummy PARTITION OF \"%s\".\"%s\""
```
Deepesh125: Nice catch

test/JDBC/input/PARTITION-vu-prepare.mix
```diff
@@ -138,6 +138,10 @@ GO
 --------------------------------------------------
 --- Supported Datatypes for Partition Function 
 --------------------------------------------------
+-- all datatypes should sort and store the values in ascending order by itself
```
Deepesh125, line 141: can we add following test case 
CREATE TABLE dbo.PartitionTable (col1 datetime2(0) PRIMARY KEY, col2 char(10))  
    ON myRangePS1--this is the comment
    (col1) ;

contrib/babelfishpg_tsql/src/pl_handler.c
Deepesh125: Assertion does not get run in release build. And it will lead to some segmentation fault, lets make sure that we handle NULL cases

Deepesh125 (changes requested):

contrib/babelfishpg_tsql/src/hooks.h
```diff
@@ -17,7 +17,6 @@ void pre_wrapper_pgstat_init_function_usage(const char *);
 pg_locale_t *collation_cache_entry_hook_function(Oid ,pg_locale_t *);
 extern bool output_update_transformation;
 extern bool output_into_insert_transformation;
-extern char *extract_identifier(const char *start);
```
Deepesh125, line 20: why this change?

sumitj824: moved the declaration of `extract_identifier` to `pltsql.h`  header file, so that it can be accessed in backend parser.

Deepesh125 (dismissed):

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -4182,17 +4184,38 @@ exec_stmt_partition_function(PLtsql_execstate *estate, PLtsql_stmt_partition_fun
...
 	tsql_type_datum = (*common_utility_plugin_ptr->translate_pg_type_to_tsql) (fcinfo);
-
 	if (tsql_type_datum)
 	{
 		tsql_typename = text_to_cstring(DatumGetTextPP(tsql_type_datum));
 	}
-
+	else
+	{
+		tuple = SearchSysCache1(TYPEOID, ObjectIdGetDatum(typ->typoid));
```
forestkeeper, line 4203: Should add check for tuple : 
```
		if (!HeapTupleIsValid(tup))
			??
```

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -4182,17 +4184,38 @@ exec_stmt_partition_function(PLtsql_execstate *estate, PLtsql_stmt_partition_fun
...
+		typform = (Form_pg_type) GETSTRUCT(tuple);
+		if (OidIsValid(typform->typbasetype) && typform->typnamespace == get_namespace_oid("sys", false))
+		{
+			/* Input type is DOMAIN type created in sys schema. */
```
forestkeeper, line 4207: Comment style is not correct.

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -4236,6 +4259,12 @@ exec_stmt_partition_function(PLtsql_execstate *estate, PLtsql_stmt_partition_fun
 		/* evaluate the value from the expr */
 		val = exec_eval_expr(estate, list_nth(arg, i), &isnull, &valtype, &valtypmod);
 
+		/* raise error for null value */
```
forestkeeper, line 4263: comment style

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -5375,6 +5373,264 @@ get_partition_function(int16 dbid, const char *partition_scheme_name)
...
+
+	/* Update the table name of the found row. */
+	if (HeapTupleIsValid(tuple))
+	{
+		new_record[Anum_bbf_partition_depend_table_name - 1] = CStringGetTextDatum(stmt->newname);
+		new_record_replace[Anum_bbf_partition_depend_table_name - 1] = true;
+		new_tuple = heap_modify_tuple(tuple, dsc, new_record, new_record_nulls, new_record_replace);
+		CatalogTupleUpdate(rel, &new_tuple->t_self, new_tuple);
+		heap_freetuple(new_tuple);
+	}
```
forestkeeper, line 5531: What's the else ops expected in here ?

contrib/babelfishpg_tsql/src/pl_exec-2.c
sumitj824: typ->typoid will always be valid OID and will always have entry in pg_type. Please check the definition of parse_datatype for more details.

contrib/babelfishpg_tsql/src/catalog.c
sumitj824: no ops, if there is no entry then do nothing.

contrib/babelfishpg_tsql/src/pl_exec-2.c
sumitj824: I think we can use both single or multi-line comment.

contrib/babelfishpg_tsql/src/backend_parser/gram-tsql-rule.y
```diff
@@ -2574,6 +2575,68 @@ tsql_stmt :
 				{ $$ = NULL; }
 		;
 
+tsql_CreatePartitionStmt:
+			CREATE OptTemp TABLE qualified_name '(' OptTableElementList ')'
```
KushaalShroff, line 2586: Can you explain why we need OptTemp here?
Also worthy of a comment since we are not even using the TEMP clause here

contrib/babelfishpg_tsql/src/backend_parser/gram-tsql-rule.y
```diff
@@ -2574,6 +2575,68 @@ tsql_stmt :
...
+			OptInherit OptPartitionSpec table_access_method_clause OptWith
```
KushaalShroff, line 2587: Same for other Opts

contrib/babelfishpg_tsql/src/backend_parser/gram-tsql-rule.y
```diff
@@ -2574,6 +2575,68 @@ tsql_stmt :
...
+			tsql_PartitionSpec
+				{
+					CreateStmt *n = makeNode(CreateStmt);
+					n->relation = $4;
+					n->tableElts = $6;
+					n->inhRelations = $8;
```
KushaalShroff: Why do we need this? At least we can set it to NULL directly here.
Also think of a scenario where we run such a query from PG endpoint with TSQL Dialect

contrib/babelfishpg_tsql/src/tsqlIface.cpp
```diff
@@ -7085,8 +7085,12 @@ post_process_create_table(TSqlParser::Create_tableContext *ctx, PLtsql_stmt_exec
 		if (cctx->ON())
 		{
 			Assert(cctx->storage_partition_clause());
-			removeTokenStringFromQuery(stmt->sqlstmt, cctx->ON(), baseCtx);
-			removeCtxStringFromQuery(stmt->sqlstmt, cctx->storage_partition_clause(), baseCtx);
+			/* remove storage_partition_clause only if it's not partitioning clause */
```
KushaalShroff, line 7151: why do we need this for create table and not for create index?

contrib/babelfishpg_tsql/src/pltsql_bulkcopy.c
```diff
@@ -580,6 +579,11 @@ ExecuteBulkCopy(BulkCopyState cstate, int rowCount, int colCount,
 					(errcode(ERRCODE_WRONG_OBJECT_TYPE),
 					 errmsg("cannot bulk copy to sequence \"%s\"",
 							RelationGetRelationName(cstate->rel))));
+		else if (cstate->rel->rd_rel->relkind == RELKIND_PARTITIONED_TABLE)
+			ereport(ERROR,
+					(errcode(ERRCODE_WRONG_OBJECT_TYPE),
+					 errmsg("cannot bulk copy to partitioned-table \"%s\"",
```
KushaalShroff: "IS NOT SUPPORTED YET IN BABELFISH"

test/JDBC/input/PARTITION-vu-verify.mix
```diff
@@ -566,13 +669,745 @@ GO
...
+--- Unsupported Option: Insert Bulk To Partitioned Table
+--------------------------------------------------------------------
+CREATE TABLE TestPartitionSourceTable(Id BIGINT, Value sys.varchar(50));
+GO
+
+INSERT INTO TestPartitionSourceTable (Id, Value) VALUES (1, 'Value 1');
+GO
+
+-- should throw error
+insertbulk#!#TestPartitionSourceTable#!#partition_vu_prepare_bigint_table
```
KushaalShroff: In JDBC lets avoid adding the whole insert bulk API invoke, can we please append this same test in the dotnet test file?

We had an issue with the scoping of sp_executesql because of which Tim had ignored these files.

contrib/babelfishpg_tsql/src/pltsql.h
```diff
@@ -64,9 +64,29 @@
...
+ * name of partitioned table.
+ */
+#define BBF_GET_PARTITION_NAME(hash, partition_number)  \
+	psprintf("%s_partition_%d", hash, partition_number)
+
+/*
+ * This macro constructs a new partition name by replacing the
+ * hash portion of an existing partition name with a new hash value.
+ */
+#define BBF_GET_NEW_PARTITION_NAME(partition_name, hash)  \
```
KushaalShroff: Worthy of .h?
Can these macros be part of the .c file?

contrib/babelfishpg_tsql/src/pltsql.h
```diff
@@ -64,9 +64,29 @@
...
- /* Max number of partitions allowed for babelfish partitioned tables. */
+/* Max number of partitions allowed for babelfish partitioned tables. */
 #define MAX_PARTITIONS_LIMIT 15000
 
+/*
+ * This macro constructs a partition name with the provided hash
+ * and partition number. The hash will be unique based on the
+ * name of partitioned table.
+ */
+#define BBF_GET_PARTITION_NAME(hash, partition_number)  \
```
KushaalShroff: We are creating macro for psprintf which seems unnecessary (also used only once), rather you should make a MACRO for the format of "%s_partition_%d" if we intend to use it again

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -17,14 +17,17 @@
 
 #include "access/attnum.h"
 #include "access/relation.h"
+#include "access/genam.h"
```
KushaalShroff: Lets refactor the helper/exec functions to pl_exec/pl_exec-2 if possible

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -2527,6 +2543,102 @@ bbf_ProcessUtility(PlannedStmt *pstmt,
...
+						break;
+					
+					/*
+					 * For babelfish partitioned table, user should not be
+					 * allowed to attach/detach of table directly.
+					 */
+					if (cmd->subtype == AT_AttachPartition || cmd->subtype == AT_DetachPartition 
+						|| cmd->subtype == AT_DetachPartitionFinalize)
+					{
+						if (is_bbf_partitioned_table(dbid, logical_schema_name, atstmt->relation->relname))
```
KushaalShroff: This will be blocked by ANTLR parser and if we are doing this specifically for PG endpoint then can we please add appropriate comments?

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -2527,6 +2543,102 @@ bbf_ProcessUtility(PlannedStmt *pstmt,
...
+
+						tuple = SearchSysCache1(RELOID, ObjectIdGetDatum(relid));
+						if (HeapTupleIsValid(tuple))
+						{
+							Form_pg_class classForm = (Form_pg_class) GETSTRUCT(tuple);
+							is_partition_table = classForm->relispartition;
+							ReleaseSysCache(tuple);
+						}
+
+						if (!is_partition_table)
```
KushaalShroff: We are doing this very late in the code flow, even ALTER TABLE for non partitioned tables we are waiting until this stage to break.
My suggestion is to create an API is_partition_table(relid) and use it at the start of this block in
```
if (!babelfish_dump_restore && atstmt->objtype == OBJECT_TABLE && !superuser())
```
Lets avoid extra processing for normal tables

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -2527,6 +2543,102 @@ bbf_ProcessUtility(PlannedStmt *pstmt,
...
+					else if (atstmt->relation->schemaname) /* Schema is explicitly specified for non TDS client. */
+					{
+						logical_schema_name = (char *) get_logical_schema_name(atstmt->relation->schemaname, true);
+						if (!logical_schema_name) /* not a TSQL schema */
+							break;
+						
+						/* Find dbid from physical schema name for a TSQL schema. */
+						dbid = get_dbid_from_physical_schema_name(atstmt->relation->schemaname, false);
+					}
+					else
```
KushaalShroff: We are allowing to continue for cases where we do not have explicit schema names, there could be a (very unlikely)scenario where we have the physical schema in the search_path and then we are leaking the alter to go through

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -2527,6 +2543,102 @@ bbf_ProcessUtility(PlannedStmt *pstmt,
...
+							pfree(logical_schema_name);
+							break;
+						}
+
+						parent_table_name = get_rel_name(get_partition_parent(relid, false));
+
+						if (is_bbf_partitioned_table(dbid, logical_schema_name, parent_table_name))
+							ereport(ERROR,
+									(errcode(ERRCODE_FEATURE_NOT_SUPPORTED), 
+									errmsg("Cannot modify the partition '%s' of babelfish partitioned table '%s'.",
```
KushaalShroff: Does this error message make sense to a TSQL user?
The partitioned and partitions are something implicit to Postges and Babelfish users do not have that transparency, right?

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -3723,13 +3844,45 @@ bbf_ProcessUtility(PlannedStmt *pstmt,
 				if (restore_tsql_tabletype)
 					create_stmt->tsql_tabletype = true;
 
+				/*
+				 * For babelfish partitioned table, non-superusers should not be
+				 * allowed to create new partitions of a table directly.
+				 */
+				if (!babelfish_dump_restore && create_stmt->partbound && !superuser())
```
KushaalShroff: This seems like a very grey area of interoperability, It makes sense of what you did for alter(and drop) so that we continue to support altering partitions and partitioned(desperately need comments regarding this) but for creating partitioned tables, why do we want to still allow this?

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -3723,13 +3844,45 @@ bbf_ProcessUtility(PlannedStmt *pstmt,
...
+				{
+					rel = linitial(create_stmt->inhRelations);
```
KushaalShroff: Why is this a list and do we need to iterate?

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -6391,6 +6584,691 @@ bbf_ExecDropStmt(DropStmt *stmt)
...
+	 *
+	 */
+	if (OidIsValid(partition_column_basetypoid))
+	{
+		Datum		tsql_typname;
+		/* Get the TSQL typename from partition column type. */
+		InitFunctionCallInfoData(*fcinfo, NULL, 0, InvalidOid, NULL, NULL);
+		fcinfo->args[0].value = ObjectIdGetDatum(partition_column_typoid);
+		fcinfo->args[0].isnull = false;
+		tsql_typname = (*common_utility_plugin_ptr->translate_pg_type_to_tsql) (fcinfo);
```
KushaalShroff: Easier to use (*common_utility_plugin_ptr->get_tsql_datatype_oid) since we already have type name?
And then we can check for invalid oid

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -6391,6 +6584,691 @@ bbf_ExecDropStmt(DropStmt *stmt)
...
+			partition_column_typoid = pg_type->oid;
+			partition_column_basetypoid = pg_type->typbasetype;
+			partition_column_typname = pstrdup(NameStr(pg_type->typname));
+			ReleaseSysCache(ctype);
+			break;
+		}
+	}
+
+	/* Get OID of sql_variant type. */
+	sql_variant_type_oid = (*common_utility_plugin_ptr->get_tsql_datatype_oid) ("sql_variant");
```
KushaalShroff: We dont have a utility directly for this?

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -6391,6 +6584,691 @@ bbf_ExecDropStmt(DropStmt *stmt)
...
+	ScanKeyData	key;
+	Oid		parentrelid;
+	Oid		inhrelid;
+	HeapTuple	tuple;
+	const char	*old_dialect;
+	StringInfoData	query;
+	char		*partition_name;
+	char		*new_partition_name;
+
+	logical_schema_name = (char *) get_logical_schema_name(physical_schema_name, true);
```
KushaalShroff: Are we sure that physical_schema_name will always be not null here?

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -5193,3 +5197,32 @@ unique_constraint_nulls_ordering(ConstrType constraint_type, SortByDir ordering)
...
+/*
+ * is_partitioned_table_reloptions_allowed
+ * 	This function checks if the given reloptions are allowed or not for partitioned tables.
+ * 	Returns true if allowed, false otherwise.
+ * 
+ * 	Only bbf_rel_create_date and bbf_original_rel_name reloptions
+ * 	are allowed in TSQL dialect and while restoring babelfish database.
+ */
+static bool
+is_partitioned_table_reloptions_allowed(Datum reloptions)
```
KushaalShroff, line 5240: This hook seems to be in a hot path in Native PG, lets discuss if we can do it in another way!

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -5375,6 +5373,264 @@ get_partition_function(int16 dbid, const char *partition_scheme_name)
...
+
+	/* Find the logical schema name and dbid from physical schema name. */
+	logical_schema_name = (char *) get_logical_schema_name(stmt->relation->schemaname, true);
+
+	if (!logical_schema_name) /* not a tsql schema */
+		return;
+
+	dbid = get_dbid_from_physical_schema_name(stmt->relation->schemaname, false);
+
+	if (!is_bbf_partitioned_table(dbid, logical_schema_name, table_name))
```
KushaalShroff: Inside is_bbf_partitioned_table we are looking up in bbf_partition_depend and then later in this function we are again looking up this table. Kindly change to lookup this table only once.

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -5375,6 +5373,264 @@ get_partition_function(int16 dbid, const char *partition_scheme_name)
...
+bool
+is_bbf_partitioned_table(int16 dbid, char *schema_name, char *table_name)
+{
+	Relation	rel;
+	HeapTuple	tuple;
+	ScanKeyData	scanKey[3];
+	SysScanDesc	scan;
+	bool		sucess = false;
+
+	rel = table_open(get_bbf_partition_depend_oid(), AccessShareLock);
```
KushaalShroff, line 5557: NoLock?

KushaalShroff (changes requested):

contrib/babelfishpg_tsql/src/pltsql.h
sumitj824: removed these macros because it was getting used only once.

contrib/babelfishpg_tsql/src/pl_handler.c
sumitj824: Good catch!! changed the implementation to take search_path into account.

contrib/babelfishpg_tsql/src/pl_handler.c
sumitj824: Removed the changes to restrict create partition for babelfish partitioned table, as creating new partitions in our scenario would require dropping existing partitions which is already blocked.

contrib/babelfishpg_tsql/src/pl_handler.c
sumitj824: no, we don't have any such utility.

contrib/babelfishpg_tsql/src/pl_handler.c
sumitj824: yes, here it will always be not null value.

contrib/babelfishpg_tsql/src/tsqlIface.cpp
sumitj824: We have TSQL specific rule for create index, while for create table we rely in PG rules. If we don't want to remove filegroups for create table as well in ANTLR then we will need to make our own TSQL rules.

contrib/babelfishpg_tsql/src/backend_parser/gram-tsql-rule.y
sumitj824: right, these are not needed. Fixed this.

contrib/babelfishpg_tsql/src/backend_parser/gram-tsql-rule.y
sumitj824: If we don't include optional nodes (like OptTemp) in the tsql_CreatePartitionStmt rule, the parser was encountering  shift-reduce conflict because it couldn't determine whether to shift (for tsql_CreatePartitionStmt) or reduce (for CreateStmt) when encountering the optional nodes due to similarity between the grammar rules.

When parser will encounter a statement starting with CREATE TABLE, it couldn't reliably decide whether to apply the CreateStmt rule or the tsql_CreatePartitionStmt rule because the latter didn't include the OptTemp clause.

contrib/babelfishpg_tsql/src/catalog.c
sumitj824: AccessShareLock seems to be safe here.

KushaalShroff: Why?

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -6391,6 +6476,796 @@ bbf_ExecDropStmt(DropStmt *stmt)
...
+	pfree(query.data);
+}
+
+/*
+ * For Babelfish partitioned tables, non-superusers should not be permitted
+ * to attach or detach partitions from the partitioned table, and they
+ * should also be restricted from modifying the partitions from both
+ * TSQL as well as PG endpoint.
+ * 
+ * NOTE: We are only blocking operation on Babelfish partitioned tables i.e.
```
KushaalShroff: Lets speak to DBEs and get the CREATE PARTITIONED TABLES to be blocked from psql

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -6391,6 +6476,796 @@ bbf_ExecDropStmt(DropStmt *stmt)
...
+	{
+		if (is_bbf_partitioned_table(dbid, logical_schemaname, stmt->relation->relname))
+			ereport(ERROR,
+					(errcode(ERRCODE_FEATURE_NOT_SUPPORTED), 
+					errmsg("Cannot %s babelfish partitioned table '%s'.",
+							cmd->subtype == AT_AttachPartition ? "attach partition to" : "detach partition from",
+							stmt->relation->relname)));
+	}
+	/*
+	 * For partition of babelfish partitioned table, user should not
```
KushaalShroff: Seems convoluted to understand which is being blocked

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -6391,6 +6476,796 @@ bbf_ExecDropStmt(DropStmt *stmt)
...
+	 * be allowed to modify the partition.
+	 * These will blocked from both TSQL and PG endpoint.
+	 */
+	else if (is_partition_table)
```
KushaalShroff: Lets do same for create but with partitioned this time

contrib/babelfishpg_tsql/src/pl_handler.c
sumitj824: refactored it to new file

contrib/babelfishpg_tsql/src/pl_handler.c
sumitj824: updated the comment.

contrib/babelfishpg_tsql/src/pl_handler.c
sumitj824: sure, I will check with Alex and block it in separate PR if he agrees.

contrib/babelfishpg_tsql/src/pl_handler.c
sumitj824: Will check with Alex and block it in separate PR if he agrees.

contrib/babelfishpg_tsql/src/catalog.c
sumitj824: Since we are accessing the data of `sys.babelfish_partition_depend` catalog, we would require AccessShareLock so the accessed does not get modified concurrently. I think we can use NOLOCK only if we are interested in metadata of table.

Deepesh125 (approved):
Code wise this PR looks good to me.

shah-nirmit:
Added Tests Looks Good. LGTM

KushaalShroff (approved):

contrib/babelfishpg_tsql/src/pl_exec-2.c
forestkeeper: What will happen if customer delete that from pg_type manually.

sumitj824: If type is deleted, parse_datatype API in ANTLR itself will raise error.

PR #2708, opened 2024-07-02
Block dropping of internal extended stored procedures

roshan0708:
### Description
This commit adds a check to prevent dropping of specific extended stored procedures:
- xp_qv
- xp_instance_regread
- sp_addlinkedsrvlogin
- sp_droplinkedsrvlogin
- sp_dropserver
- sp_enum_oledb_providers
- sp_testlinkedserver

These procedures are critical and should not be dropped accidentally. This change also fixes the logic of dropping these procedures for TSQL dialect.

### Issues Resolved

Task: BABEL-5080, BABEL-4390

* **Use case based -** yes

* **Arbitrary inputs -**

* **Performance tests -**

* **Tooling impact -**

* **Client tests -**

### Check List


contrib/babelfishpg_tsql/src/pl_funcs-2.c
```diff
@@ -350,6 +351,67 @@ coalesce_typmod_hook_impl(const CoalesceExpr *cexpr)
...
+				is_restricted = true;
+				break;
+			}
+		}
+	}
+
+	ReleaseSysCache(proctup);
+	if (is_restricted)
+	{
+		ereport(ERROR,
```
shah-nirmit, line 402: Nit: we are throwing an error saying that the procedure doesn't exist , which doesn't make sense to me as we verified that procedure exist in the above loop.

contrib/babelfishpg_tsql/src/pl_funcs-2.c
```diff
@@ -350,6 +351,67 @@ coalesce_typmod_hook_impl(const CoalesceExpr *cexpr)
...
+	const char		*procname;
+	Oid		schema_oid = InvalidOid;
+	Oid		sys_oid = InvalidOid;
+	Oid		dbo_oid = InvalidOid;
+	bool		is_restricted = false;
+
+	proctup = SearchSysCache1(PROCOID, ObjectIdGetDatum(proc_id));
+	if (!HeapTupleIsValid(proctup))
+	{
+		elog(ERROR, "cache lookup failed for procedure %u", proc_id);
```
shah-nirmit: Nit: Doesn't seem like a very useful error message from user POV.

contrib/babelfishpg_tsql/src/pl_funcs-2.c
```diff
@@ -350,6 +351,67 @@ coalesce_typmod_hook_impl(const CoalesceExpr *cexpr)
...
+		return;
+	}
+
+	procform = (Form_pg_proc) GETSTRUCT(proctup);
+	procname = pstrdup(NameStr(procform->proname));
+	schema_oid = procform->pronamespace;
+	sys_oid = get_namespace_oid("sys", true);
```
shah-nirmit: Question: Should this have `missing_ok = false` , with true set the function can return an InvalidOid and we are not checking if its valid in next logic

contrib/babelfishpg_tsql/src/pl_funcs-2.c
```diff
@@ -350,6 +351,67 @@ coalesce_typmod_hook_impl(const CoalesceExpr *cexpr)
 	return exprTypmod((Node *) linitial(cexpr->args));
 }
 
+void
+check_restricted_stored_procedure_hook_impl(Oid proc_id)
```
Deepesh125: Previous implementation was good in a way that we didn't require new hook. Anyway, bbf_processUtility gets called before standard processUtility so we do not need hook unless I am missing something

contrib/babelfishpg_tsql/src/pl_funcs-2.c
```diff
@@ -350,6 +351,67 @@ coalesce_typmod_hook_impl(const CoalesceExpr *cexpr)
...
+{
+	/*
+	 * BABEL-4390
```
Deepesh125: Please refrain from adding internal jira id to OSS, we can check with @shardgupta if it is allowed

contrib/babelfishpg_tsql/src/pl_funcs-2.c
```diff
@@ -350,6 +351,67 @@ coalesce_typmod_hook_impl(const CoalesceExpr *cexpr)
...
+	 * List of procedure names that are not allowed to be dropped. These procedures 
+	 * are considered essential or restricted due to security or operational reasons.
+	 */
+	const char		*restricted_procedures[] = {
```
Deepesh125: will it make sense if we make this local static?

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -6358,22 +6343,6 @@ bbf_ExecDropStmt(DropStmt *stmt)
 			if (schema_name != NULL)
 				logicalschema = get_logical_schema_name(schema_name, true);
 
-			/* Restrict dropping of extended stored procedures */
```
Deepesh125, line 6453: Do we know why do we pass missingOk = true on line 6326? and why are skipping all the logic on line 6332?

contrib/babelfishpg_tsql/src/pl_funcs-2.c
```diff
@@ -350,6 +351,67 @@ coalesce_typmod_hook_impl(const CoalesceExpr *cexpr)
...
+	Oid		sys_oid = InvalidOid;
+	Oid		dbo_oid = InvalidOid;
+	bool		is_restricted = false;
+
+	proctup = SearchSysCache1(PROCOID, ObjectIdGetDatum(proc_id));
+	if (!HeapTupleIsValid(proctup))
+	{
+		elog(ERROR, "cache lookup failed for procedure %u", proc_id);
+		return;
+	}
```
Deepesh125: As far as I can understand current changes, proc_id will always be valid and hence we should not hit this error ideally. Am I missing anything? 

Btw, we should use new way of error reporting - https://github.com/babelfish-for-postgresql/postgresql_modified_for_babelfish/blob/BABEL_4_X_DEV__PG_16_X/src/include/utils/elog.h#L100

contrib/babelfishpg_tsql/src/pl_funcs-2.c
```diff
@@ -350,6 +351,67 @@ coalesce_typmod_hook_impl(const CoalesceExpr *cexpr)
...
+
+	procform = (Form_pg_proc) GETSTRUCT(proctup);
+	procname = pstrdup(NameStr(procform->proname));
+	schema_oid = procform->pronamespace;
+	sys_oid = get_namespace_oid("sys", true);
+	dbo_oid = get_namespace_oid("master_dbo", true);
```
Deepesh125, line 384: will this work in both the migration_mode?

contrib/babelfishpg_tsql/src/pl_funcs-2.c
Deepesh125: we have extern variable sys_schema_oid which can be used now. 

Just curious -- can end user drop sys objects? If they can not then we probably do not need to check it.

test/JDBC/expected/BABEL-4390-vu-verify.out
```diff
@@ -82,3 +82,28 @@ GO
 
 ~~ERROR (Message: Cannot drop the procedure 'sp_testlinkedserver', because it does not exist or you do not have permission.)~~
 
+
+-- psql
+-- should fail
+DROP PROCEDURE dbo.xp_qv;
```
Deepesh125: are we adding hook because we want to block this use case? And shouldn't this be master_dbo?

test/JDBC/input/BABEL-4390-vu-prepare.sql
```diff
@@ -56,4 +56,19 @@ CREATE PROCEDURE babel_4390_prepare_p9 AS
...
+CREATE PROCEDURE babel_4390_test_schema.xp_qv AS
+BEGIN
+    SELECT 1;
+END
+GO
+
+CREATE PROCEDURE babel_4390_test_schema.sp_addlinkedsrvlogin AS
+BEGIN
+    SELECT 1;
+END
```
Deepesh125: can we add positive test where we show that we are able to dro such procedures in user defined schema

Deepesh125 (changes requested):

contrib/babelfishpg_tsql/src/pl_funcs-2.c
Deepesh125: On OSS side, in theory user can drop sys objects so let this check in place.

roshan0708: End user cannot drop sys objects, so yes we can rule out this check for sys oid.

contrib/babelfishpg_tsql/src/pl_funcs-2.c
roshan0708: Yes, it'll work with `bbf_processUtility` as well, I'll remove the custom hook and in the next commit.

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -3504,6 +3504,34 @@ bbf_ProcessUtility(PlannedStmt *pstmt,
 				{
 					if (sql_dialect == SQL_DIALECT_TSQL)
 						bbf_ExecDropStmt(drop_stmt);
+					/*
+					 * Prevent dropping of restricted stored procedures from PSQL dialect
+					 */
+					else if (drop_stmt->removeType == OBJECT_PROCEDURE)
```
shah-nirmit: Blocking :We should handle Procedure or any object irrepective of DIALECT in single function. Adding additional conditional logic complicates things for future contributors. We dont have a single convention. Current approach doesnt seem extendible. If not possible in current effort create a ticket to track the same.

contrib/babelfishpg_tsql/src/pl_funcs-2.c
roshan0708: Yes, since we are only creating extra procedures in `master_dbo` schema.

contrib/babelfishpg_tsql/src/pl_handler.c
Deepesh125: I agree here, we need to have some kind of pre-processing where we want to implement some generic function with logic to handle restriction of various objects. We may also want to let superuser to drop such procedures. So let's try to think from this lines. You can also study existing restriction if there is any and see if we can handle it in the same generic function.

roshan0708: @shah-nirmit I did try using a single function for both dialects although I'm getting error while initializing Babelfish, some `DROP PROCEDURE` is failing with `Could not find relation with OID 0`, so I believe @Deepesh125 is correct here, we will need to do some pre-processing specially in case of PSQL DIALECT.

contrib/babelfishpg_tsql/src/pl_funcs-2.c
```diff
@@ -350,6 +351,60 @@ coalesce_typmod_hook_impl(const CoalesceExpr *cexpr)
...
+	const char		*procname;
+	Oid		schema_oid = InvalidOid;
+	Oid		dbo_oid = InvalidOid;
+	bool		is_restricted = false;
+
+	/*
+	 * List of procedure names that are not allowed to be dropped. These procedures 
+	 * are considered essential or restricted due to security or operational reasons.
+	 */
+	static const char	*restricted_procedures[] = {
```
thephantomthief, line 368: Do we need to make these variables static if these are only needed for the context of this one specific function?

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -3502,8 +3502,7 @@ bbf_ProcessUtility(PlannedStmt *pstmt,
 
 				if (drop_stmt->removeType != OBJECT_SCHEMA)
 				{
-					if (sql_dialect == SQL_DIALECT_TSQL)
```
thephantomthief, line 3541: Why is this change needed?

roshan0708: This change is a part of discussion here - https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/2708#discussion_r1677347072

During the initial stage of this PR, I was handling this drop restriction separately for both dialects and we came to a conclusion to unify the logic for both dialects.

thephantomthief (approved):

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -6284,34 +6268,37 @@ bbf_ExecDropStmt(DropStmt *stmt)
 			schema_oid = get_object_namespace(&address);
 			if (OidIsValid(schema_oid))
 				schema_name = get_namespace_name(schema_oid);
-			if (schema_name != NULL)
-				logicalschema = get_logical_schema_name(schema_name, true);
-
-			if (schema_name && major_name)
+			if (sql_dialect == SQL_DIALECT_TSQL)
 			{
```
shah-nirmit, line 6364: So we clear extended property only when dialect is TSQL??? , is this expected behaviour? Please check with harsh/shalini.

HarshLunagariya: Ideally, we should be cleaning up the catalogs independent to dialects. That's not happening for schema and extended properties at least - which will make our catalog bloated with stale entries (when creating object from TDS endpoint and dropping it from PG endpoint). This should be tracked as separate issue and fixed.

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -6419,10 +6406,14 @@ bbf_ExecDropStmt(DropStmt *stmt)
...
 										 &relation,
 										 AccessShareLock,
-										 true);
+										 is_missing);
 			Assert(relation == NULL);
 			if (!OidIsValid(address.objectId))
 				continue;
+				
+			/* Restrict dropping of extended stored procedures for non-sysadmin roles */
+			if (stmt->removeType == OBJECT_PROCEDURE && !is_member_of_role(GetSessionUserId(), get_sysadmin_oid()))
```
HarshLunagariya: 1. Wanted to clear out that -- 'Even w/o current code change, Only sysadmin-member were able to drop these procedures.' Is this true? If it's true then there is no point of this change.

2. Please verify whether this check blocks dropping procedures for superuser or not? Ideally it should not block it.

roshan0708: Correct, even without current change, only sysadmin member will be able to drop these. I'll update the logic to throw error accordingly and allow only superuser to drop it.

test/JDBC/upgrade/13_4/schedule
```diff
@@ -20,7 +20,6 @@ BABEL-1444
 BABEL-1465
 BABEL-1466
 BABEL-1475
-BABEL-4390
```
HarshLunagariya, line 23: Why are we not adding the new files?

HarshLunagariya (approved):

test/JDBC/expected/restricted_procedures-vu-verify.out
```diff
@@ -0,0 +1,99 @@
+-- tsql
+-- reset the login password
+ALTER LOGIN babel_4390_user_l1 WITH PASSWORD = 'abc';
+GO
```
HarshLunagariya, line 4: Is there any testcase where we are checking the superuser behaviour that it should be able to drop it?

roshan0708: Didn't add the test for this case, as we'll need to recreate that dropped procedure else version upgrade will break. Tested this case locally:
```
jdbc_testdb=# show is_superuser;
 is_superuser 
--------------
 on
(1 row)

jdbc_testdb=# drop procedure master_dbo.xp_qv;
DROP PROCEDURE
```

shah-nirmit (approved):
LGTM

contrib/babelfishpg_tsql/src/pl_funcs-2.c
Deepesh125: By defining it as a static, we would saving it from reinitialising it from every call to this function. Initialisation scope would be til the end of backend but variable scope would be local within the function it defined.

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -6376,34 +6360,37 @@ bbf_ExecDropStmt(DropStmt *stmt)
...
-				}
-				else if (stmt->removeType == OBJECT_VIEW)
-				{
-					type = ExtendedPropertyTypeNames[EXTENDED_PROPERTY_VIEW];
-					delete_extended_property(db_id, type, schema_name,
-											 major_name, NULL);
-				}
-				else if (stmt->removeType == OBJECT_SEQUENCE)
+				if (schema_name != NULL)
+					logicalschema = get_logical_schema_name(schema_name, true);
```
Deepesh125, line 6366: logicalscheam can be null here? is it possible?

Deepesh125 (dismissed):

contrib/babelfishpg_tsql/src/pl_handler.c
shalinilohia50: If schema_name is not NULL, logicalschema should not be NULL. 
We can set the missingOk flag to false to throw an error in case of any exception.

roshan0708: Changed the default value of missingOk flag to `false` in the latest commit.

roshan0708: @Deepesh125 `logical_schema` being `NULL` case is already being handled inside `clean_up_bbf_schema_permissions`, setting it to `false` in the function call is breaking alot of existing test cases.

shalinilohia50 (approved):

Deepesh125 (approved):

PR #2713, opened 2024-07-08
[OSS-ONLY] Cache mappings for sys.remove_accents_internal

tanscorpio7:
### Description

To support like operator for AI collations, we strip the accents from the strings before doing the matching.
To do so we used the ICU transliterate function `utrans_transUChars` which is quite slow.
So as an improvement we cache the mappings for it locally and use it if the ICU version on host matches to
the one we used to generate the mapping, otherwise we fallback to the ICU function `utrans_transUChars`

### Issues Resolved

[BABEL-5008]

### Sign Off

### Check List


contrib/babelfishpg_tsql/src/utils/removeaccent.map
```diff
@@ -0,0 +1,816 @@
+/* TRANSFORMATION_RULE "[[:Latin:][:Nd:]]; Latin-ASCII" */
+
+static const remove_accent_map_pair pltsql_remove_accent_map[811] =
```
ahmed-shameem: Let's have a MACRO for the number of entries rather than hardcoding it

contrib/babelfishpg_tsql/Makefile
```diff
@@ -93,7 +93,7 @@ PG_CFLAGS += -g -Werror -Wfloat-conversion
 PG_CFLAGS += -fstack-protector-strong
 PG_CPPFLAGS += -I$(TSQLSRC) -I$(PG_SRC) -DFAULT_INJECTOR -Wfloat-conversion
 
-SHLIB_LINK += -L$(ANTLR4_RUNTIME_LIB_DIR) $(ANTLR4_RUNTIME_LIB) -lcrypto $(ICU_LIBS)
```
ahmed-shameem: IMO, we should keep this. This picks up the correct ICU version.

contrib/babelfishpg_tsql/src/utils/removeaccent.map
```diff
@@ -0,0 +1,816 @@
+/* TRANSFORMATION_RULE "[[:Latin:][:Nd:]]; Latin-ASCII" */
```
Deepesh125, line 1: Let's create readme file on how to generate this mapping. Plus, add comment in each row to show the character. Also, let's commit script we are using to generate this mapping.

contrib/babelfishpg_tsql/src/collation.c
```diff
@@ -460,125 +446,129 @@ get_remove_accents_internal_oid()
...
+{
+	unsigned char *input_str = (unsigned char *)text_to_cstring(PG_GETARG_TEXT_PP(0)),
+	              *result,
+	              *result_itr;
+	text          *return_result;
+	int           len = strlen((char *)input_str),
+	              l;
+
+	/* normalized value never takes more bytes than original char */
+	result = palloc0(len+1);
```
Deepesh125: (char *) palloc0(len + 1)

contrib/babelfishpg_tsql/src/collation.c
```diff
@@ -460,125 +446,129 @@ get_remove_accents_internal_oid()
...
+		remove_accent_map_pair *pr;
+
+		/* "break" cases all represent errors */
+		if (*input_str == '\0')
+			break;
+
+		l = pg_utf_mblen(input_str);
+		
+		if (len < l)
+			break;
```
Deepesh125: will we ever hit this case? or any example would help understand it better

contrib/babelfishpg_tsql/src/collation.c
```diff
@@ -460,125 +446,129 @@ get_remove_accents_internal_oid()
...
+
+		if (!pg_utf8_islegal(input_str, l))
+			break;
```
Deepesh125: Ideally, it will not hit this case so will it make sense to throw an error?

contrib/babelfishpg_tsql/src/collation.c
```diff
@@ -460,125 +446,129 @@ get_remove_accents_internal_oid()
...
+
+		if (l == 1)
 		{
-			ereport(ERROR,
-					(errcode(ERRCODE_EXTERNAL_ROUTINE_EXCEPTION),
-						errmsg("Error opening transliterator: %s", u_errorName(status))));
+			*result_itr++ = *input_str++;
```
Deepesh125: single byte char does not require any normalisation?

Deepesh125 (changes requested):

contrib/babelfishpg_tsql/src/collation.c
tanscorpio7: We shouldn't be hitting this, just a safety check.

contrib/babelfishpg_tsql/src/collation.c
tanscorpio7: No, our mapping does not have any single character as source character.
Quick exit here means we are not unnecessarily doing a btree search below.

contrib/babelfishpg_tsql/src/collation.c
tanscorpio7: All break cases are errors, thrown outside the loop.

contrib/babelfishpg_tsql/src/utils/removeaccent.map
```diff
@@ -0,0 +1,816 @@
...
+
+static const remove_accent_map_pair pltsql_remove_accent_map[811] = 
+{
+    {0x0000c380,0x00000041},
```
ahmed-shameem: We should also have automated tests in place to catch any changes made in Unicode entries. By change I mean any new additions, or even change in unaccented form with our rule above.

ahmed-shameem (changes requested):

contrib/babelfishpg_tsql/src/utils/removeaccent.map
tanscorpio7: We can use ```lengthof(pltsql_remove_accent_map)``` for length of array.

contrib/babelfishpg_tsql/src/collation.c
tanscorpio7: switching to string info instead.

contrib/babelfishpg_tsql/src/utils/removeaccent.map
Deepesh125: Since this is static array, we do not need to supply array lenght

contrib/babelfishpg_tsql/src/collation.c
```diff
@@ -552,13 +692,6 @@ Datum remove_accents_internal(PG_FUNCTION_ARGS)
 						&status);
 
 	/* Allocated capacity may not be enough to hold un-accented string. This shouldn't occur ideally but still defensive code. */
-	if (status == U_BUFFER_OVERFLOW_ERROR)
-	{
```
Deepesh125, line 567: why this change?

contrib/babelfishpg_tsql/src/collation.c
```diff
@@ -457,9 +458,148 @@ get_remove_accents_internal_oid()
...
 
+#ifdef USE_ICU
+	if (U_ICU_VERSION_MAJOR_NUM == 60 && U_ICU_VERSION_MINOR_NUM == 2)
+	{
+		elog(LOG, "Using cached mappings to remove accents");
+		remove_accents_internal_oid = LookupFuncName(list_make2(makeString("sys"), makeString("remove_accents_internal_using_cache")), -1, funcargtypes, true);
+		return;
+	}
+#endif
+	elog(LOG, "Using ICU function to remove accents");
```
Deepesh125: Let's add comment here explaining when and why will we use remove_accents_internal_using_cache and when and why will be use remove_accents_internal. 

readme file is still missing on explaining how to generate mapping? and if we generate new mapping using new version of ICU then what all changes might be required

contrib/babelfishpg_tsql/src/collation.c
```diff
@@ -457,9 +458,148 @@ get_remove_accents_internal_oid()
...
+
+		/* Use the mapping if availaible or else the character */
+		if (pr && pr->normalized_char)
+			utf8_normalized_str = pr->normalized_char;
+		else
+			utf8_normalized_str = utf8_char;
+
+		store_coded_char(c, utf8_normalized_str);
+
+		appendBinaryStringInfo(&result, c, strlen((const char *) c));
```
Deepesh125: why const char *? and not unsigned char * simply?

contrib/babelfishpg_tsql/src/collation.c
```diff
@@ -457,9 +458,148 @@ get_remove_accents_internal_oid()
...
+	}
+
+	if (len > 0)
+		ereport(ERROR,
+			(errcode(ERRCODE_CHARACTER_NOT_IN_REPERTOIRE),
+			 errmsg("error while removing accents")));
```
Deepesh125: Ambiguous error message

contrib/babelfishpg_tsql/src/collation.c
ahmed-shameem: Yes, I think we should not remove it as there is a possibility that the end-user is not using the cached mappings. In that case we may get buffer overflow. Can you please explain this?

tanscorpio7: we are checking for error below `if (U_FAILURE(status))`
This check is redundant.

contrib/babelfishpg_tsql/src/collation.c
tanscorpio7: strlen expects arg of type (const char *)

contrib/babelfishpg_tsql/src/collation.c
tanscorpio7: changed

contrib/babelfishpg_tsql/src/utils/removeaccent.map
tanscorpio7: PG declares length in their map which is why keeping our implementation as close to it as possible.
[ref](https://github.com/tanscorpio7/postgresql_modified_for_babelfish/blob/BABEL_4_X_DEV__PG_16_X/src/backend/utils/mb/Unicode/gbk_to_utf8.map#L40)

contrib/babelfishpg_tsql/src/collation.c
tanscorpio7: added

test/JDBC/input/collations/verify_mapping_for_remove_accents.mix
```diff
@@ -0,0 +1,36 @@
...
+DECLARE
+    erro TEXT;
+    cnt INT:= 1;
+    var TEXT;
+BEGIN
+    WHILE cnt <= 1114112 LOOP
+        BEGIN
+            INSERT INTO master_dbo.all_utf8_chars VALUES (cast(chr(cnt) as bytea), chr(cnt));
+        EXCEPTION
+        WHEN others THEN
```
surendravishnoi: why we ignore error here?

contrib/babelfishpg_tsql/src/utils/removeaccent.map
```diff
@@ -0,0 +1,816 @@
+/* TRANSFORMATION_RULE "[[:Latin:][:Nd:]]; Latin-ASCII" */
+
+static const remove_accent_map_pair pltsql_remove_accent_map[811] = 
+{
```
surendravishnoi, line 7: why .map instead of .h?

contrib/babelfishpg_tsql/src/collation.c
```diff
@@ -450,16 +451,160 @@ transform_from_ci_as_for_likenode(Node *node, OpExpr *op, like_ilike_info_t like
...
+ * the cache. Otherwise we fallback on the ICU function
+ */
 static void
 get_remove_accents_internal_oid()
 {
 	const Oid funcargtypes[1] = {TEXTOID};
 	if (OidIsValid(remove_accents_internal_oid))
 		return;
 
+#ifdef USE_ICU
```
surendravishnoi, line 477: why USE_ICU check?

contrib/babelfishpg_tsql/src/collation.c
```diff
@@ -450,16 +451,160 @@ transform_from_ci_as_for_likenode(Node *node, OpExpr *op, like_ilike_info_t like
...
+	if (U_ICU_VERSION_MAJOR_NUM == 60 && U_ICU_VERSION_MINOR_NUM == 2)
+	{
+		elog(LOG, "Using cached mappings to remove accents");
+		remove_accents_internal_oid = LookupFuncName(list_make2(makeString("sys"), makeString("remove_accents_internal_using_cache")), -1, funcargtypes, true);
+		return;
+	}
+#endif
+	elog(LOG, "Using ICU function to remove accents");
```
surendravishnoi, line 485: this should be a warning

contrib/babelfishpg_tsql/src/collation.c
```diff
@@ -450,16 +451,160 @@ transform_from_ci_as_for_likenode(Node *node, OpExpr *op, like_ilike_info_t like
...
+ */
 static void
 get_remove_accents_internal_oid()
 {
 	const Oid funcargtypes[1] = {TEXTOID};
 	if (OidIsValid(remove_accents_internal_oid))
 		return;
 
+#ifdef USE_ICU
+	if (U_ICU_VERSION_MAJOR_NUM == 60 && U_ICU_VERSION_MINOR_NUM == 2)
```
surendravishnoi: can we also add ICU version to generated map file? In that way, we always check against map file rather than using fixed values here

contrib/babelfishpg_tsql/src/collation.c
```diff
@@ -450,16 +451,160 @@ transform_from_ci_as_for_likenode(Node *node, OpExpr *op, like_ilike_info_t like
...
+}
+
+PG_FUNCTION_INFO_V1(remove_accents_internal_using_cache);
+Datum remove_accents_internal_using_cache(PG_FUNCTION_ARGS)
+{
+	unsigned char *input_str = (unsigned char *) text_to_cstring(PG_GETARG_TEXT_PP(0));
+	unsigned char *input_str_start = input_str;
+	text          *return_result;
+	int           len = strlen((char *)input_str),
+	              l;
```
surendravishnoi: let use vars more descriptive than l, c

contrib/babelfishpg_tsql/src/collation.c
```diff
@@ -450,16 +451,160 @@ transform_from_ci_as_for_likenode(Node *node, OpExpr *op, like_ilike_info_t like
...
+
+	v1 = *(const uint32 *) p1;
+	v2 = ((const remove_accent_map_pair *) p2)->original_char;
+	return (v1 > v2) ? 1 : ((v1 == v2) ? 0 : -1);
+}
+
+PG_FUNCTION_INFO_V1(remove_accents_internal_using_cache);
+Datum remove_accents_internal_using_cache(PG_FUNCTION_ARGS)
+{
+	unsigned char *input_str = (unsigned char *) text_to_cstring(PG_GETARG_TEXT_PP(0));
```
surendravishnoi: check for input being NULL

contrib/babelfishpg_tsql/src/collation.c
```diff
@@ -450,16 +451,160 @@ transform_from_ci_as_for_likenode(Node *node, OpExpr *op, like_ilike_info_t like
...
+		else if (l == 4)
+		{
+			b1 = *input_str++;
+			b2 = *input_str++;
+			b3 = *input_str++;
+			b4 = *input_str++;
+		}
+		else
+		{
+			elog(ERROR, "unsupported character length %d", l);
```
surendravishnoi: add more context to error

contrib/babelfishpg_tsql/src/collation.c
```diff
@@ -450,16 +451,160 @@ transform_from_ci_as_for_likenode(Node *node, OpExpr *op, like_ilike_info_t like
...
+
+		l = pg_utf_mblen(input_str);
+		
+		if (len < l)
+			break;
+
+		if (!pg_utf8_islegal(input_str, l))
+			break;
+
+		if (l == 1)
```
surendravishnoi: this can be done in a single loop with l as length. Get value from input_str and place it at right place in utf8_char using a local variable

surendravishnoi: this can be done in a single loop with l as length. Get value from input_str and place it at right place in utf8_char using a local variable

contrib/babelfishpg_tsql/src/collation.c
```diff
@@ -450,16 +451,160 @@ transform_from_ci_as_for_likenode(Node *node, OpExpr *op, like_ilike_info_t like
...
+
+		/* Use the mapping if availaible or else the character */
+		if (pr && pr->normalized_char)
+			utf8_normalized_str = pr->normalized_char;
+		else
+			utf8_normalized_str = utf8_char;
+
+		store_coded_char(c, utf8_normalized_str);
+
+		appendBinaryStringInfo(&result, c, strlen((const char *) c));
```
surendravishnoi: length of c is fixed, unnecessary strlen

contrib/babelfishpg_tsql/src/collation.c
```diff
@@ -450,16 +451,160 @@ transform_from_ci_as_for_likenode(Node *node, OpExpr *op, like_ilike_info_t like
...
+	}
+
+	if (len > 0)
+		ereport(ERROR,
+			(errcode(ERRCODE_CHARACTER_NOT_IN_REPERTOIRE),
+			 errmsg("invalid byte sequence for encoding UTF-8 while removing accents using cache")));
```
surendravishnoi: let us not talk about cache, it is internal. Try to given operation context in which this error can happen

contrib/babelfishpg_tsql/src/collation.c
tanscorpio7: The macros in below if condition are only available when ICU is available.

test/JDBC/input/collations/verify_mapping_for_remove_accents.mix
tanscorpio7: Some integer are converted to hex value which is not valid in that encoding, so we get an error like
```ERROR:  requested character not valid for encoding: 57330```
We want to skip these errors.

File: contrib/babelfishpg_tsql/src/collation.c (hunk 23 above)
Comment (tanscorpio7) [reply contrib/babelfishpg_tsql/src/collation.c]: fixed

File: contrib/babelfishpg_tsql/src/collation.c (hunk 19 above)
Comment (tanscorpio7) [reply contrib/babelfishpg_tsql/src/collation.c]: done

File: contrib/babelfishpg_tsql/src/collation.c (hunk 16 above)
Comment (tanscorpio7) [reply contrib/babelfishpg_tsql/src/collation.c:485]: this gets executed when we are loading cache which does not necessarily happen during like with AI collations.

File: contrib/babelfishpg_tsql/src/collation.c (hunk 17 above)
Comment (tanscorpio7) [reply contrib/babelfishpg_tsql/src/collation.c]: done

File: contrib/babelfishpg_tsql/src/collation.c (hunk 18 above)
Comment (tanscorpio7) [reply contrib/babelfishpg_tsql/src/collation.c]: done

File: contrib/babelfishpg_tsql/src/collation.c (hunk 22 above)
Comment (tanscorpio7) [reply contrib/babelfishpg_tsql/src/collation.c]: we convert the normalised str from `unit32` to `C String` in store_coded_char. The length will be <= 5 bytes (including null terminator)

File: contrib/babelfishpg_tsql/src/utils/removeaccent.map (hunk 14 above)
Comment (tanscorpio7) [reply contrib/babelfishpg_tsql/src/utils/removeaccent.map:7]: PG seems to use .map for their mappings, following the same, not sure of the difference in .h and .map

File: contrib/babelfishpg_tsql/src/collation.c (hunk 16 above)
Comment (tanscorpio7) [reply contrib/babelfishpg_tsql/src/collation.c:485]: We would not want to throw a warning when user isn't running like

File: contrib/babelfishpg_tsql/src/collation.c (hunk 20 above)
Comment (tanscorpio7) [reply contrib/babelfishpg_tsql/src/collation.c]: done

File: contrib/babelfishpg_tsql/src/collation.c (hunk 21 above)
Comment (tanscorpio7) [reply contrib/babelfishpg_tsql/src/collation.c]: Done

Comment (tanscorpio7) [reply contrib/babelfishpg_tsql/src/collation.c]: done

File: contrib/babelfishpg_tsql/src/collation.c (hunk 15 above)
Comment (surendravishnoi) [reply contrib/babelfishpg_tsql/src/collation.c:477]: can our code run without ICU? what library we use in that case?

File: contrib/babelfishpg_tsql/src/collation.c (hunk 21 above)
Comment (Deepesh125) [reply contrib/babelfishpg_tsql/src/collation.c]: previous impl was better in terms of readability and PG also does that thing in the same way. Check https://github.com/babelfish-for-postgresql/postgresql_modified_for_babelfish/blob/BABEL_4_X_DEV__PG_16_X/src/backend/utils/mb/conv.c#L507

File: contrib/babelfishpg_tsql/src/utils/removeaccent.map [hunk 24]
```diff
@@ -1,6 +1,9 @@
 /* TRANSFORMATION_RULE "[[:Latin:][:Nd:]]; Latin-ASCII" */
 
-static const remove_accent_map_pair pltsql_remove_accent_map[811] = 
+#define pltsql_remove_accent_map_icu_major_version 60
+#define pltsql_remove_accent_map_icu_min_version 2
```
Deepesh125, line 4: Instead, let's create internal function which will help us getting version of ICU using which this map is created

Deepesh125 (dismissed):

Deepesh125 (dismissed):

contrib/babelfishpg_tsql/src/collation.c
```diff
@@ -461,16 +462,168 @@ transform_from_ci_as_for_likenode(Node *node, OpExpr *op, like_ilike_info_t like
...
+		else if (char_len == 4)
+		{
+			b1 = *input_str++;
+			b2 = *input_str++;
+			b3 = *input_str++;
+			b4 = *input_str++;
+		}
+		else
+		{
+			elog(ERROR, "unsupported character length %d", char_len);
```
surendravishnoi, line 595: we should also print input string?

test/JDBC/expected/verify_mapping_for_remove_accents.out
```diff
@@ -0,0 +1,40 @@
...
+DECLARE
+    erro TEXT;
+    cnt INT:= 1;
+    var TEXT;
+BEGIN
+    WHILE cnt <= 1114112 LOOP
+        BEGIN
+            INSERT INTO master_dbo.all_utf8_chars VALUES (cast(chr(cnt) as bytea), chr(cnt));
+        EXCEPTION
+        WHEN others THEN
```
surendravishnoi: handle other errors than what we already expect

surendravishnoi (dismissed):

Deepesh125 (approved):

PR #2495, opened 2024-04-15
Added wrapper function for TRIM, LTRIM, RTRIM, LEFT and RIGHT

rohit01010:
### Description
Currently, for string functions TRIM, LTRIM, RTRIM, LEFT and RIGHT underlying Postgres function is being used which leads to incorrect output behaviour. This PR will fix this issue by adding wrapper functions for these string functions.

Changes included in this PR are as follows:-
- Added wrapper functions for TRIM, LTRIM, RTRIM, LEFT and RIGHT.
- Added hook `func_select_candidate_for_special_hook` for selecting best candidate among list of functions definitions based on return type for string functions which are in special function list. 
- Updated definition of function `pltsql_report_proc_not_found_error ` to include `input_typeids` of function.
- Added `PG_CATALOG` namespace before TRIM, LTRIM, RTRIM, LEFT or RIGHT function invocations in some of the system functions as with these wrapper definitions of TRIM, LTRIM, RTRIM, LEFT and RIGHT functions in `sys` namespace, any function invocation of TRIM, LTRIM, RTRIM, LEFT or RIGHT function without namespace, by default it will call the function in `sys` namespace.
- Added new test files and updated existing test files expected output as the output behaviour of these functions will change with this fix.
- In case of Inline Table Valued Function(ITVF), made changes such that if the return type of body is different than return type stored for during creation of ITVF, an implicit cast will occur and return type of body will be casted to return type stored.
- fixed composite actions to avoid replacing partial word for test files present in schedule file of upgrades.

Engine PR: https://github.com/babelfish-for-postgresql/postgresql_modified_for_babelfish/pull/344

Authored-by: Rohit Bhagat [rohitbgt@amazon.com](mailto:rohitbgt@amazon.com)
Signed-off-by: Rohit Bhagat [rohitbgt@amazon.com](mailto:rohitbgt@amazon.com)

### Issues Resolved

BABEL-4489

### Test Scenarios Covered ###
* **Use case based -** YES

* **Boundary conditions -** YES

* **Arbitrary inputs -** NA

* **Negative test cases -** YES

* **Minor version upgrade tests -** YES

* **Major version upgrade tests -** YES

* **Performance tests -** NA

* **Tooling impact -** NA

* **Client tests -** YES

### Check List
- [x] Commits are signed per the DCO using --signoff 

[...]

PR #2747, opened 2024-07-17
Added wrapper function for Substring to handle binary datatypes and multibyte character

Anikait143:
## Description

### 1. Issues
- SUBSTRING() function with last argument missing raises strange error message.
- SUBSTRING() should work on [VAR]BINARY, Image argument.

- Added error message handling in ANTLR parser to throw appropriate error message for different no. of arguments.
- Added wrapper function handling for SUBSTRING() function to handle binary datatypes and multibyte characters.
```
# Input
1> SELECT substring('abc',1)
2> GO
```
```

Msg 33557097, Level 16, State 1, Server persist-rcv-17aprb-15-3-0-6, Line 1
"a" is not a valid binary digit 
```
```

The substring function requires 3 argument(s).                                                                                                                                                                                                                                             
```

```
1> select substring(0x0a0b0c, 1,2), substring(0x0a0b0c,2,1)
2> GO
substring        substring
---------------- ----------------
0x               x
```
```
1> select substring(0x0a0b0c, 1,2), substring(0x0a0b0c,2,1)
2> GO
substring substring
--------- ---------
0x0A0B    0x0B
```
- Added Test Cases for these issues


Engine PR: https://github.com/babelfish-for-postgresql/postgresql_modified_for_babelfish/pull/415

Task: BABEL- 4118, BABEL- 3658



* **Use case based -**   substring-*


* **Boundary conditions -**


* **Arbitrary inputs -**


* **Negative test cases -**


* **Performance tests -**


* **Tooling impact -**

* **Memory tests -**


* **Client tests -**



### Check List


contrib/babelfishpg_tsql/src/string.c
```diff
@@ -504,6 +505,21 @@ tsql_varchar_substr(PG_FUNCTION_ARGS)
...
 }
 
+/*
+ * tsql_varbinary_substr()
+ * Return a substring starting at the specified position.
+ */
+Datum
+tsql_varbinary_substr(PG_FUNCTION_ARGS)
+{
+	if (PG_ARGISNULL(0) || PG_ARGISNULL(1) || PG_ARGISNULL(2))
```
sumitj824: Why don't we define SQL function as STRICT instead of NULL check here?

contrib/babelfishpg_tsql/sql/sys_functions.sql
```diff
@@ -3073,6 +3077,35 @@ CREATE OR REPLACE FUNCTION sys.substring(string sys.NCHAR, i INTEGER, j INTEGER)
 RETURNS sys.NVARCHAR
 AS 'babelfishpg_tsql', 'tsql_varchar_substr' LANGUAGE C IMMUTABLE PARALLEL SAFE;
 
+CREATE OR REPLACE FUNCTION sys.substring(string sys.VARBINARY, i INTEGER, j INTEGER)
```
sumitj824, line 3164: Please use meaningful variable name instead of i and j everywhere.

contrib/babelfishpg_tsql/sql/sys_functions.sql
```diff
@@ -3073,6 +3077,35 @@ CREATE OR REPLACE FUNCTION sys.substring(string sys.NCHAR, i INTEGER, j INTEGER)
...
+DECLARE
+    string_arg_datatype text;
+    string_basetype oid;
+BEGIN
+    string_arg_datatype := sys.translate_pg_type_to_tsql(pg_typeof(string)::oid);
+    IF string_arg_datatype IS NULL THEN
+        -- for User Defined Datatype, use immediate base type to check for argument datatype validation
+        string_basetype := (SELECT typbasetype FROM pg_type WHERE oid = pg_typeof(string)::oid);
+        string_arg_datatype := sys.translate_pg_type_to_tsql(string_basetype);
+    END IF;
```
sumitj824: IIRC, Rohit is planning to create a SQL wrapper function for this operation, Can you please check with him?

contrib/babelfishpg_tsql/sql/sys_functions.sql
```diff
@@ -3073,6 +3077,35 @@ CREATE OR REPLACE FUNCTION sys.substring(string sys.NCHAR, i INTEGER, j INTEGER)
 RETURNS sys.NVARCHAR
 AS 'babelfishpg_tsql', 'tsql_varchar_substr' LANGUAGE C IMMUTABLE PARALLEL SAFE;
 
+CREATE OR REPLACE FUNCTION sys.substring(string sys.VARBINARY, i INTEGER, j INTEGER)
+RETURNS sys.VARBINARY
+AS 'babelfishpg_tsql', 'tsql_varbinary_substr' LANGUAGE C IMMUTABLE PARALLEL SAFE;
+
+CREATE OR REPLACE FUNCTION sys.substring(string ANYELEMENT, i INTEGER, j INTEGER)
```
sumitj824, line 3168: Do we have handled the cases for NULL values when this function will be used? Please add testcases when NULL value is casted to different datatypes.

contrib/babelfishpg_tsql/sql/sys_functions.sql
```diff
@@ -3073,6 +3077,35 @@ CREATE OR REPLACE FUNCTION sys.substring(string sys.NCHAR, i INTEGER, j INTEGER)
...
+BEGIN
+    string_arg_datatype := sys.translate_pg_type_to_tsql(pg_typeof(string)::oid);
+    IF string_arg_datatype IS NULL THEN
+        -- for User Defined Datatype, use immediate base type to check for argument datatype validation
+        string_basetype := (SELECT typbasetype FROM pg_type WHERE oid = pg_typeof(string)::oid);
+        string_arg_datatype := sys.translate_pg_type_to_tsql(string_basetype);
+    END IF;
+
+    -- restricting arguments with invalid datatypes for substring function
+    IF string_arg_datatype NOT IN ('varchar', 'char', 'nchar', 'nvarchar', 'binary', 'varbinary', 'image') THEN
```
sumitj824: Here list should be only ('binary', 'varbinary', 'image'), right?

contrib/babelfishpg_tsql/sql/sys_functions.sql
```diff
@@ -3073,6 +3077,35 @@ CREATE OR REPLACE FUNCTION sys.substring(string sys.NCHAR, i INTEGER, j INTEGER)
...
+RETURNS sys.VARBINARY
+AS
+$BODY$
+DECLARE
+    string_arg_datatype text;
+    string_basetype oid;
+BEGIN
+    string_arg_datatype := sys.translate_pg_type_to_tsql(pg_typeof(string)::oid);
+    IF string_arg_datatype IS NULL THEN
+        -- for User Defined Datatype, use immediate base type to check for argument datatype validation
```
sumitj824, line 3180: Have we validated that UDT should be allowed as per TSQL?

contrib/babelfishpg_tsql/sql/sys_functions.sql
```diff
@@ -3057,11 +3057,15 @@ CREATE OR REPLACE FUNCTION sys.substring(string TEXT, i INTEGER, j INTEGER)
...
+CREATE OR REPLACE FUNCTION sys.substring(string NTEXT, i INTEGER, j INTEGER)
+RETURNS sys.NVARCHAR
+AS 'babelfishpg_tsql', 'tsql_varchar_substr' LANGUAGE C IMMUTABLE PARALLEL SAFE;
+
 CREATE OR REPLACE FUNCTION sys.substring(string sys.VARCHAR, i INTEGER, j INTEGER)
 RETURNS sys.VARCHAR
 AS 'babelfishpg_tsql', 'tsql_varchar_substr' LANGUAGE C IMMUTABLE PARALLEL SAFE;
 
-CREATE OR REPLACE FUNCTION sys.substring(string sys.VARCHAR, i INTEGER, j INTEGER)
+CREATE OR REPLACE FUNCTION sys.substring(string sys.BPCHAR, i INTEGER, j INTEGER)
```
sumitj824, line 3152: Why this change? Are we handling the upgrade if we are changing datatype of existing function?

sumitj824 (changes requested):

contrib/babelfishpg_tsql/src/string.c
Anikait143: Done

contrib/babelfishpg_tsql/sql/sys_functions.sql
Anikait143: Some of these functions were already defined in previous releases with argument names as i and j, so just for this little change it would require to deprecate previous definitions in upgrade scripts, so I think it's better to keep these names as it is for now.

contrib/babelfishpg_tsql/sql/sys_functions.sql
Anikait143: Included

contrib/babelfishpg_tsql/sql/sys_functions.sql
Anikait143: Needed special handling, did that and also added tests

contrib/babelfishpg_tsql/sql/sys_functions.sql
Anikait143: Actually it should only be ('binary', 'image'), modified!

contrib/babelfishpg_tsql/sql/sys_functions.sql
Anikait143: Yes, already have added test for the same.

contrib/babelfishpg_tsql/sql/sys_functions.sql
Anikait143: There were 2 exactly identical functions earlier, so just modified one of the definitions, so we just need to add the new definition in upgrade file which we have already done.

contrib/babelfishpg_tsql/sql/sys_functions.sql
```diff
@@ -3055,23 +3055,56 @@ IS 'This function returns column or parameter information. Currently only works
...
+
+CREATE OR REPLACE FUNCTION sys.substring(string ANYELEMENT, i INTEGER, j INTEGER)
+RETURNS sys.VARBINARY
+AS
+$BODY$
+DECLARE
+    string_arg_datatype text;
+    string_basetype oid;
+BEGIN
+    string_arg_datatype := sys.translate_pg_type_to_tsql(pg_typeof(string)::oid);
```
Deepesh125: you may want to store the value of pg_typeof(string)::oid and reuse it

contrib/babelfishpg_tsql/src/tsqlIface.cpp
```diff
@@ -2882,6 +2882,9 @@ class tsqlMutator : public TSqlParserBaseListener
 				{
 					if (ctx->function_arg_list() && !ctx->function_arg_list()->expression().empty())
 					{
+						if (ctx->function_arg_list()->expression().size() != 3)
+							ereport(ERROR, (errcode(ERRCODE_UNDEFINED_FUNCTION), errmsg("The substring function requires 3 argument(s).")));
```
Deepesh125: why do we need to handle this specially here?

test/JDBC/expected/BABEL-2875.out
```diff
@@ -7,8 +7,8 @@ CREATE TABLE t_sksql(
 )
 go
 
-SELECT SUBSTRING( T.name, 1, 20 ) AS "UDDT name",
-SUBSTRING( T2.name, 1, 20 ) AS "system dt name",
+SELECT pg_catalog.SUBSTRING( T.name, 1, 20 ) AS "UDDT name",
+pg_catalog.SUBSTRING( T2.name, 1, 20 ) AS "system dt name",
```
Deepesh125: why do we need to make these changes?

contrib/babelfishpg_tsql/sql/sys_functions.sql
Anikait143: Done

contrib/babelfishpg_tsql/src/tsqlIface.cpp
Anikait143: Modified the handling to rewrite `substring` to `sys.substring`

test/JDBC/expected/BABEL-2875.out
Anikait143: Don't need these changes anymore, reverted them

contrib/babelfishpg_tsql/sql/sys_functions.sql
```diff
@@ -3136,23 +3136,58 @@ IS 'This function returns column or parameter information. Currently only works
...
 RETURNS sys.NVARCHAR
-AS 'babelfishpg_tsql', 'tsql_varchar_substr' LANGUAGE C IMMUTABLE PARALLEL SAFE;
+AS 'babelfishpg_tsql', 'tsql_varchar_substr' LANGUAGE C IMMUTABLE STRICT PARALLEL SAFE;
 
 CREATE OR REPLACE FUNCTION sys.substring(string sys.NCHAR, i INTEGER, j INTEGER)
 RETURNS sys.NVARCHAR
-AS 'babelfishpg_tsql', 'tsql_varchar_substr' LANGUAGE C IMMUTABLE PARALLEL SAFE;
+AS 'babelfishpg_tsql', 'tsql_varchar_substr' LANGUAGE C IMMUTABLE STRICT PARALLEL SAFE;
+
+CREATE OR REPLACE FUNCTION sys.substring(string sys.VARBINARY, i INTEGER, j INTEGER)
```
sumitj824, line 3164: Use meaningful arg names.

contrib/babelfishpg_tsql/src/pltsql_coerce.c
```diff
@@ -1031,15 +1031,15 @@ init_special_function_list()
...
+validate_special_function(char *func_nsname, char *func_name, List* fargs, int nargs, Oid *input_typeids)
 {
 	tsql_special_function_t    *special_func;
 	bool                        type_match;
 	Oid                         input_type_id, valid_type_id, base_type_id;
 	Oid                         sys_varcharoid;
 
 	/* Sanity checks */
-	if (func_name == NULL || (nargs != 0 && input_typeids == NULL))
+	if (func_name == NULL || (nargs != 0 && input_typeids == NULL) || fargs == NULL)
```
sumitj824: NIL?

contrib/babelfishpg_tsql/src/pltsql_coerce.c
```diff
@@ -1070,15 +1070,55 @@ validate_special_function(char *func_nsname, char *func_name, int nargs, Oid *in
 
 	/* If function is not a special function no additional handling required */
 	if (special_func == NULL)
```
sumitj824, line 1072: Is substring is part of special function?

contrib/babelfishpg_tsql/src/pltsql_coerce.c
```diff
@@ -1070,15 +1070,55 @@ validate_special_function(char *func_nsname, char *func_name, int nargs, Oid *in
...
+				if (input_typeids[i] != UNKNOWNOID)
+				{
+					Node *arg = (Node *) lfirst(list_nth_cell(fargs, i));
+					/* Throw error when input is constant and NULL */
+					if (IsA(arg, Const) && ((Const *)arg)->constisnull)
+					{
+						if (common_utility_plugin_ptr == NULL)
+							ereport(ERROR,
+									(errcode(ERRCODE_INTERNAL_ERROR),
+										errmsg("Failed to find common utility plugin.")));
```
sumitj824: this is not needed, common_utility_plugin_ptr will always be present.

contrib/babelfishpg_tsql/src/pltsql_coerce.c
```diff
@@ -1070,15 +1070,55 @@ validate_special_function(char *func_nsname, char *func_name, int nargs, Oid *in
...
+
+						if ((*common_utility_plugin_ptr->get_tsql_datatype_oid)("int") != input_typeids[i]
+							&& (*common_utility_plugin_ptr->get_tsql_datatype_oid)("bigint") != input_typeids[i]
+							&& (*common_utility_plugin_ptr->get_tsql_datatype_oid)("smallint") != input_typeids[i]
+							&& (*common_utility_plugin_ptr->get_tsql_datatype_oid)("tinyint") != input_typeids[i])
```
sumitj824: Suggestion: Finding tsql typname by oid and then string comparsion will be more efficient here.

contrib/babelfishpg_tsql/src/pltsql_coerce.c
```diff
@@ -1070,15 +1070,55 @@ validate_special_function(char *func_nsname, char *func_name, int nargs, Oid *in
...
+	}		
 
 	sys_varcharoid = get_sys_varcharoid();
 
 	/* Report error in case of invalid argument datatype */
 	for (int i = 0; i < special_func->nargs; i++)
 	{
-		/* for unknown literals consider its type as sys.VARCHAR */
-		input_type_id = (input_typeids[i] == UNKNOWNOID) ? sys_varcharoid : input_typeids[i];
+		/* if argument is NULL then keep its typeId as UNKNOWN and skip the report error handling
```
sumitj824: comment style is not consistent.

contrib/babelfishpg_tsql/src/tsqlIface.cpp
```diff
@@ -925,6 +925,11 @@ class tsqlCommonMutator : public TSqlParserBaseListener
 					size_t startPosition = id->keyword()->TRANSLATE()->getSymbol()->getStartIndex();
 					rewritten_query_fragment.emplace(std::make_pair(startPosition, std::make_pair("", "sys.")));
 				}
+				if (id->keyword()->SUBSTRING())
+				{
+					size_t startPosition = id->keyword()->SUBSTRING()->getSymbol()->getStartIndex();
+					rewritten_query_fragment.emplace(std::make_pair(startPosition, std::make_pair("", "sys.")));
+				}
```
sumitj824, line 952: Instead of adding explicit for each built-in functions, can't we do something we handle for all of the built-in function?

contrib/babelfishpg_tsql/src/tsqlIface.cpp
```diff
@@ -2472,6 +2477,11 @@ class tsqlBuilder : public tsqlCommonMutator
 					size_t startPosition = id->keyword()->TRANSLATE()->getSymbol()->getStartIndex();
 					rewritten_query_fragment.emplace(std::make_pair(startPosition, std::make_pair("", "sys.")));
 				}
+				if (id->keyword()->SUBSTRING()) /* SUBSTRING */
+				{
+					size_t startPosition = id->keyword()->SUBSTRING()->getSymbol()->getStartIndex();
+					rewritten_query_fragment.emplace(std::make_pair(startPosition, std::make_pair("", "sys.")));
+				}
```
sumitj824: I believe this is not needed in tsqlBuilder because it is inherited from tsqlCommonMutator. Can you explain why we are making same changes at multiple places?

contrib/babelfishpg_tsql/sql/sys_functions.sql
Anikait143: please refer to https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/2747#discussion_r1690378053

contrib/babelfishpg_tsql/src/pltsql_coerce.c
Anikait143: Done

contrib/babelfishpg_tsql/src/pltsql_coerce.c
Anikait143: No, substring does not need special handling just added check to handle NULL arguments

contrib/babelfishpg_tsql/src/pltsql_coerce.c
Anikait143: please refer to https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/2495#discussion_r1643784228

contrib/babelfishpg_tsql/src/pltsql_coerce.c
Anikait143: `Oid` comparison is faster and less prone to vulnerability than string comparison, hence decided to use Oid comparision here. reference: https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/2495#discussion_r1643795450

contrib/babelfishpg_tsql/src/pltsql_coerce.c
Anikait143: Done

contrib/babelfishpg_tsql/src/tsqlIface.cpp
Anikait143: Rohit is working on this in https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/2790

contrib/babelfishpg_tsql/src/tsqlIface.cpp
Anikait143: `TSqlParserBaseListener` handles rewrite in case of Views and `tsqlBuilder` handles rewrite for other statements, we have already done rewrites in both mutators for previous string functions as well as Geosptatial functions

contrib/babelfishpg_tsql/src/pltsql_coerce.c
sumitj824: nack, here get_tsql_datatype_oid works on string comparison so it not "faster" instead what i suggested will be faster.

> less prone to vulnerability than string comparison

Can you explain how that can lead to vulnerability?

contrib/babelfishpg_tsql/src/tsqlIface.cpp
sumitj824: it is redundant, the whole `exitFunction_call` or any other exitcall is only needed in tsqlCommonMutator. And since tsqlBuilder is subclass of tsqlCommonMutator, the function definition will be inherited.

Anikait143: if the exitFunction_call was not present in tsqlBuilder class then there won't have been an issue we could have put the change only in tsqlCommonMutator, but since the function exitFunction_call is present in tsqlBuilder, during query execution, tsqlBuilder::exitFunction_call will be use as we are overriding the definition of exitFunctioncall, So that's the reason we have to add the logic to both the mutators.

sumitj824: Ideally exitFunction_call should have been present only tsqlCommonMutator because we are doing function rewrite in those which should work inside views/proc/func also.
>  but since the function exitFunction_call is present in tsqlBuilder, during query execution, tsqlBuilder::exitFunction_call will be use as we are overriding the definition of exitFunctioncall, So that's the reason we have to add the logic to both the mutators.

why can't we do following in tsqlBuilder instead of copying which will increase the maintainence
```
void exitFunction_call(ctx) override
{
	   tsqlCommonMutator::exitFunction_call(ctx);
           ...
           existing_rewrites
}
```

Anikait143: Done

contrib/babelfishpg_tsql/src/pltsql_coerce.c
Anikait143: Got it, Modified

contrib/babelfishpg_tsql/src/pltsql_coerce.c
```diff
@@ -1070,15 +1070,70 @@ validate_special_function(char *func_nsname, char *func_name, int nargs, Oid *in
...
+						if (common_utility_plugin_ptr == NULL)
+							ereport(ERROR,
+									(errcode(ERRCODE_INTERNAL_ERROR),
+										errmsg("Failed to find common utility plugin.")));
+
+						/* if tsql_typename is NULL it implies that inputTypId corresponds to UDT */
+						InitFunctionCallInfoData(*fcinfo, NULL, 0, InvalidOid, NULL, NULL);
+						fcinfo->args[0].value = ObjectIdGetDatum(input_typeids[i]);
+						fcinfo->args[0].isnull = false;
+						tsql_typename = (*common_utility_plugin_ptr->translate_pg_type_to_tsql) (fcinfo);
```
sumitj824: I would suggest to create direct API of translate_pg_type_to_tsql instead of using SQL variant of it, because invoking SQL function using fcinfo is costly operation.

contrib/babelfishpg_tsql/src/pltsql_coerce.c
```diff
@@ -1070,15 +1070,70 @@ validate_special_function(char *func_nsname, char *func_name, int nargs, Oid *in
...
+						if(tsql_typename)
```
sumitj824: What will be behaviour for UDT? Have we added testcase for the same?

contrib/babelfishpg_tsql/src/tsqlIface.cpp
```diff
@@ -2444,53 +2469,11 @@ class tsqlBuilder : public tsqlCommonMutator
...
-                                      }
-                              }
-				
-				if (id->keyword()->TRIM())
-				{
-					rewritten_query_fragment.emplace(std::make_pair(id->keyword()->TRIM()->getSymbol()->getStartIndex(), std::make_pair(::getFullText(id->keyword()->TRIM()), "sys.trim")));
-				}
-			}
-
 			if (ctx->func_proc_name_server_database_schema()->procedure)
```
sumitj824: Is there any specific reason to not moving this in tsqlCommonMutator?

contrib/babelfishpg_tsql/src/pltsql_coerce.c
Anikait143: Done

contrib/babelfishpg_tsql/src/pltsql_coerce.c
Anikait143: Yes we already have tests for UDTs

contrib/babelfishpg_tsql/src/tsqlIface.cpp
Anikait143: Moved this as well to tsqlCommonMutator

contrib/babelfishpg_common/src/typecode.c
```diff
@@ -300,6 +300,21 @@ lookup_tsql_datatype_oid(const char *typename)
...
+	ht_oid2typecode_entry_t *entry;
+
+	if (OidIsValid(oid))
+	{
+		entry = hash_search(ht_oid2typecode, &oid, HASH_FIND, NULL);
+
+		if (entry && entry->persist_id < TOTAL_TYPECODE_COUNT)
+			PG_RETURN_TEXT_P(CStringGetTextDatum(type_infos[entry->persist_id].tsql_typname));
+	}
+	return (Datum) 0;
```
sumitj824: Why are returning datum by converting the cstring and later converting datum to cstring? Why can't we return cstring if we are interested in that only?

contrib/babelfishpg_tsql/src/pltsql_coerce.c
```diff
@@ -1070,15 +1070,63 @@ validate_special_function(char *func_nsname, char *func_name, int nargs, Oid *in
 
 	/* If function is not a special function no additional handling required */
 	if (special_func == NULL)
+	{
+		/* report error for case when NULL casted to different datatypes and passed as 2nd or 3rd argument of SUBSTRING() function */
+		if (strlen(func_name) == 9 && strncmp(func_name, "substring", 9) == 0)
+		{
+			for (int i = 1; i < nargs; i++)
+			{
+				if (input_typeids[i] != UNKNOWNOID)
```
sumitj824: NIT: 
instead of
```
if (input_typeids[i] != UNKNOWNOID)
{
 do something
}

```
you can do 
```
if (input_typeids[i] == UNKNOWNOID)
      continue;

do something

```

Anikait143: Done

contrib/babelfishpg_common/src/typecode.c
Anikait143: Changed return type to Void*, no need to convert anymore.

contrib/babelfishpg_common/src/babelfishpg_common.h
```diff
@@ -77,6 +77,7 @@ typedef struct common_utility_plugin
 										  bool *isBaseDec, bool *isBaseBin,
 										  bool *isBaseDate, int *variantHeaderLen);
 	Oid			(*lookup_tsql_datatype_oid) (const char *typestr);
+	void	   	*(*resolve_pg_type_to_tsql) (Oid oid);
```
sumitj824: If we are unable to return char* then instead of returning char* to void *, lets use argument to retrieve the name. Something like get_tsql_datatype_name(Oid oid, char **typname).

Anikait143: Done

contrib/babelfishpg_tsql/src/pltsql_coerce.c
```diff
@@ -1070,15 +1070,65 @@ validate_special_function(char *func_nsname, char *func_name, int nargs, Oid *in
...
+						if (!((strlen(typ_name) == 3 && strncmp(typ_name,"int", 3) == 0) ||
+							(strlen(typ_name) == 7 && strncmp(typ_name,"tinyint", 7) == 0) ||
+							(strlen(typ_name) == 8 && strncmp(typ_name,"smallint", 8) == 0) ||
+							(strlen(typ_name) == 6 && strncmp(typ_name,"bigint", 6) == 0)))
+							ereport(ERROR,
+								(errcode(ERRCODE_UNDEFINED_FUNCTION),
+									errmsg("Argument data type %s is invalid for argument %d of substring function.", 
+											format_type_be(input_typeids[i]), i+1)));
+					}
+					pfree(typ_name);
```
sumitj824: we should free typname inside if block.

contrib/babelfishpg_tsql/src/pltsql_coerce.c
```diff
@@ -1070,15 +1070,65 @@ validate_special_function(char *func_nsname, char *func_name, int nargs, Oid *in
...
+
+					if (common_utility_plugin_ptr == NULL)
+						ereport(ERROR,
+								(errcode(ERRCODE_INTERNAL_ERROR),
+									errmsg("Failed to find common utility plugin.")));
+
+					(*common_utility_plugin_ptr->resolve_pg_type_to_tsql) (input_typeids[i], &typ_name);
+					if(typ_name)
+					{
+						if (!((strlen(typ_name) == 3 && strncmp(typ_name,"int", 3) == 0) ||
```
sumitj824: Lets store the strlen(typ_name) in variable and reuse it instead of calling it again and again.

sumitj824 (approved):

contrib/babelfishpg_common/src/typecode.c
```diff
@@ -300,6 +300,20 @@ lookup_tsql_datatype_oid(const char *typename)
...
+
+	if (OidIsValid(oid))
+	{
+		entry = hash_search(ht_oid2typecode, &oid, HASH_FIND, NULL);
+
+		if (entry && entry->persist_id < TOTAL_TYPECODE_COUNT)
+			*typ_name = pstrdup(type_infos[entry->persist_id].tsql_typname);
+	}
+}
+
```
Deepesh125: Non need to return duplicate copy of tsql_typname because tsql_typname is already declared in read only memory and we can return it directly.

Also add comment here -- type_infos will return const char * so called should not attempt to modify it.

contrib/babelfishpg_tsql/src/pltsql_coerce.c
```diff
@@ -1070,15 +1070,69 @@ validate_special_function(char *func_nsname, char *func_name, int nargs, Oid *in
...
 
 	/* Report error in case of invalid argument datatype */
 	for (int i = 0; i < special_func->nargs; i++)
 	{
-		/* for unknown literals consider its type as sys.VARCHAR */
-		input_type_id = (input_typeids[i] == UNKNOWNOID) ? sys_varcharoid : input_typeids[i];
+		/* 
+		 * if argument is NULL then keep its typeId as UNKNOWN and skip the report error handling 
+		 * otherwise consider it as sys.VARCHAR
+		 */
```
Deepesh125, line 1123: which exact test is affecting due to this change?

Deepesh125 (changes requested):

contrib/babelfishpg_tsql/src/pltsql_coerce.c
Anikait143: Consider the EXAMPLE:
```
Select substring('abc', NULL, 2)
```
in this, 2nd Arguments should be of type int, but according to current workflow, we will consider it as Varchar,
So ideally it should return `NULL` but it will return Error that varchar is not accepted for argument 2, so to handle this we added this change.

contrib/babelfishpg_common/src/typecode.c
Anikait143: Done

contrib/babelfishpg_common/src/typecode.c
```diff
@@ -300,6 +300,21 @@ lookup_tsql_datatype_oid(const char *typename)
 	return typoid;
 }
 
+/* type_infos will return const char * so called, should not attempt to modify it */
```
KushaalShroff: -called
+caller

contrib/babelfishpg_common/src/typecode.c
```diff
@@ -300,6 +300,21 @@ lookup_tsql_datatype_oid(const char *typename)
...
+void
+resolve_pg_type_to_tsql(Oid oid, char **typ_name)
```
KushaalShroff: Why cant we return const char * instead of using a double pointer?
Seems inefficient for a caller to make sure its const

contrib/babelfishpg_tsql/src/pltsql_coerce.c
```diff
@@ -1070,15 +1070,69 @@ validate_special_function(char *func_nsname, char *func_name, int nargs, Oid *in
...
+				{
+					char	*typ_name = NULL;
+					int		len;
+
+					if (common_utility_plugin_ptr == NULL)
+						ereport(ERROR,
+								(errcode(ERRCODE_INTERNAL_ERROR),
+									errmsg("Failed to find common utility plugin.")));
+
+					(*common_utility_plugin_ptr->resolve_pg_type_to_tsql) (input_typeids[i], &typ_name);
```
KushaalShroff: should there be an assert in this function that nargs not more than input_typeids len?

KushaalShroff (changes requested):

contrib/babelfishpg_common/src/typecode.c
Anikait143: It was crashing when return type was char*, crashing in https://github.com/amazon-aurora/babelfish_extensions/blob/57c3f503362b450b943610e68b09696cfdade278/contrib/babelfishpg_common/src/varchar.c#L104

we have used Void* for functions to return types

contrib/babelfishpg_tsql/src/pltsql_coerce.c
Anikait143: that are already done in engine. Reference: https://github.com/babelfish-for-postgresql/postgresql_modified_for_babelfish/blob/81e345dcb8e8e24e3b1d3161a39dc219dd24cd7a/src/backend/parser/parse_func.c#L158

contrib/babelfishpg_common/src/typecode.c
Deepesh125: why would it crash if we return const char *?

KushaalShroff: +1, not clear on why its crashing

Anikait143: Made that change, crash was because earlier it wasn't defined as `Const`, worked now

contrib/babelfishpg_common/src/typecode.c
Anikait143: Done

Deepesh125 (approved):

Deepesh125 (approved):

KushaalShroff (approved):

PR #2750, opened 2024-07-17
Fix crash issue in pre transforming column name

forestkeeper:
Fix the crash happen when column ref is followed by a comment without space like : 'SELECT table1.c2--table1.REPGETTEXT('.

Previously, the column transform didn't consider '--' will directly followed with column ref and will wrongly regard the comment part as the column ref, thus will cause crash later in finding the column name.

This fix add the check for '--' at the end of column def in the sql string to avoid this crash.

Task: BABEL-5070

### Check List

### DIFF
--- a/contrib/babelfishpg_tsql/src/hooks.c
+++ b/contrib/babelfishpg_tsql/src/hooks.c
@@ -1952,7 +1952,17 @@ pre_transform_target_entry(ResTarget *res, ParseState *pstate,
 				colname_start = pstate->p_sourcetext + res->location;
 				last_dot = colname_start;
 				while(*colname_start != '\0')
-				{	
+				{
+					/*
+					 * comment follow up with column like : 
+					 *
+					 * 'SELECT table1.c2--table1.REPGETTEXT('
+					 * 
+					 * will cause crash if we don't break the searching
+					 * for the last_dot position
+					 */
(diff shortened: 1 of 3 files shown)


lejaokri (approved):

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -1952,7 +1952,17 @@ pre_transform_target_entry(ResTarget *res, ParseState *pstate,
...
+				{
+					/*
+					 * comment follow up with column like : 
+					 *
+					 * 'SELECT table1.c2--table1.REPGETTEXT('
+					 * 
+					 * will cause crash if we don't break the searching
+					 * for the last_dot position
+					 */
+					if (*colname_start == '-' && *(colname_start+1) == '-')
```
Deepesh125, line 1964: 1. colname_start + 1 may not be valid
2. we should check `open_square_bracket == 0 && double_quotes % 2 == 0` along with existing condition

forestkeeper: colname_start + 1 concern is fine, since the while condition is : 
```
*colname_start != '\0'
```

meaning the current colname_start is not '\0' and there's always a '\0' at the end of string, so colname_start +1 won't be invalid , at least it'll be '\0'

PR #2753, opened 2024-07-18
Test changes for 'Fix error when trying to drop columns from temp tables.'

Sairakan:
### Description

Tests for https://github.com/babelfish-for-postgresql/postgresql_modified_for_babelfish/pull/410

### Issues Resolved

BABEL-4912

* **Use case based -**

* **Boundary conditions -**

* **Arbitrary inputs -**

* **Negative test cases -**

* **Performance tests -**

* **Tooling impact -**

* **Client tests -**

### Check List


test/JDBC/expected/temp_table.out
```diff
@@ -0,0 +1,9 @@
+-- BABEL-4912 test ALTER TABLE DROP COLUMN
+CREATE TABLE #t1 (a INT, b INT)
+GO
+
+ALTER TABLE #t1 DROP COLUMN b
```
timchang514, line 18: can we add more tests, such as

- dropping then recreating the same column multiple times 
- make sure that dependencies aren't broken (so if the column depends on another object, or an object depends on the column that the proper behavior happens)
- more complicated dependencies such as with toast/index/identity etc
- ensure that the table also works (i can select, update, drop from it normally)

Sairakan: Added more test cases. Note that foreign key constraints are not allowed for temp tables, so I can't test that.

test/JDBC/expected/temp_table_rollback-vu-verify.out
```diff
@@ -153,37 +153,21 @@ BEGIN TRAN
 ALTER TABLE #temp_table_rollback_t1 DROP COLUMN b
 ROLLBACK
 GO
-~~ERROR (Code: 3726)~~
-
-~~ERROR (Message: cannot drop column b of table "#temp_table_rollback_t1" because other objects depend on it)~~
-
```
lejaokri, line 159: thanks for handling this.
Do we have tests for scenarios where it's supposed to fail with this error?

Sairakan: I just added an extra test case, that while not reproducing this exact error, produces a similar error due to a dependency.

lejaokri (approved):

PR #2770, opened 2024-07-23
Disable Shared Catcache Invalidation Messages for Temp Tables

timchang514:
### Description

Engine PR: babelfish-for-postgresql/postgresql_modified_for_babelfish/411

Alter the hook checks slightly to skip catcache as well as relcache messages. This can be extended to all types of cache invalidations if necessary. 

### Issues Resolved

BABEL-5012

### Check List

### DIFF
--- a/test/JDBC/expected/temp_catalog_inval-vu-verify.out
+++ b/test/JDBC/expected/temp_catalog_inval-vu-verify.out
@@ -22,12 +22,25 @@ int
 drop table #t1
 go
 
+select * from #t1
+go
+~~ERROR (Code: 33557097)~~
+
+~~ERROR (Message: relation "#t1" does not exist)~~
+
+
 
 -- constraint, depend, index, sequence, attrdefault
-create table #t2(a int identity primary key, b char(15) default 'column default')
+create table #t2(a int identity primary key, b int default 0)
 go
 
 alter table #t2 add c int
+go
+
+ALTER TABLE #t2 ALTER COLUMN a bigint
+go
+
+alter table #t2 alter column b bigint
 go
 
 insert into #t2(c) values (2)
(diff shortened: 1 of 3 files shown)


timchang514:
Testing the change is a bit difficult, since catcache invalidations weren't causing collisions with the OID change enabled. 

In addition, the invalidation messages haven't yet been disabled in XLOG so they would still appear there. 

What I have done so far is set a breakpoint on SIMessageIsForTempTable and checking the values with the ones listed in `pg_waldump`.

test/JDBC/expected/temp_catalog_inval-vu-verify.out
```diff
@@ -0,0 +1,59 @@
...
+1
+~~END~~
+
+
+drop table #t1
+go
+
+
+-- constraint, depend, index, sequence, attrdefault
+create table #t2(a int identity primary key, b char(15) default 'column default')
```
surendravishnoi: this is not clear. Ideally we want to test use case where an existing entry is invalidated. create table will add new entries. How do we validate that existing entries are invalidated?
drop table is one use case where we try to do select after drop. Can we find some alter commands which invalidate existing cat cache entries and later do a select

timchang514: ack. added some alter table and drop calls for the above.

test/JDBC/expected/temp_catalog_inval-vu-verify.out
```diff
@@ -0,0 +1,79 @@
...
+go
+~~ROW COUNT: 1~~
+
+
+insert into #t2(b, c) values (3, 2)
+go
+~~ROW COUNT: 1~~
+
+
+select * from #t2
```
surendravishnoi, line 61: is this sufficient to cover all catalogs we put in ENR? pg_depend, pg_shdepend?

timchang514: pg_depend tracks relationships between the objects (such as between table and index). and pg_shdepend will track the relationship between the role and the temp table. so both are already tested here.

surendravishnoi (approved):

PR #2748, opened 2024-07-17
Improve logic to extract identifier names

rishabhtanwar29:
### Description
Previously, the logic to extract original identifier names from query string was
not proper and could not handle the identifiers with special characters or object
name being prefix of schema name. This logic was there to extract original table
name and original function/procedure name from the query and due to this issue,
we were storing wrong original names.

Implemented helper function `extract_multipart_identifier_name` which improves
the logic to extract original name from a multi-part identifier in a query string taking
care of spaces or possibly quoted identifier. It internally uses `extract_identifier`
function to extract individual parts of an identifier.

Additional fixes:
* Fixed `sp_rename_internal` to properly handle rename of a table so that it also updates the
  underlying original table name (`bbf_original_rel_name`).
* Fixed helper function `babelfish_sp_rename_word_parse` which is used by `sp_rename`
  to split multi-part identifier names. Previously, it was using `string_split()` for the splitting
  which could not handle the quotes or special characters. Implemented helper SQL function
 `babelfish_split_identifier` which is mostly a wrapper over `split_object_name` `C` function
  to correctly split multi-part identifiers.
* Updated logic to fetch original table_name/column_name in sql views `information_schema_tsql.columns_internal`
  and `information_schema_tsql.tables` so that they don't assume that `bbf_original_rel_name`
  or `bbf_original_name` is always the first option in `reloptions` and `attoptions` list respectively,
  instead now the logic is to iterate over the whole list and search of the desired option.

Task: BABEL-5089
Signed-off-by: Rishabh Tanwar <ritanwar@amazon.com>

### Test Scenarios Covered ###
* **Use case based -**

* **Boundary conditions -**

* **Arbitrary inputs -**

* **Negative test cases -**

* **Minor version upgrade tests -**

* **Major version upgrade tests -**

* **Performance tests -**

* **Tooling impact -**

* **Client tests -**

### Check List
- [x] Commits are signed per the DCO using --signoff 

By submitting this pull request, I confirm that my contribution is under the terms of the Apache 2.0 and PostgreSQL licenses, and grant any person obtaining a copy of the contribution permission to relicense all or a portion of my contribution to the PostgreSQL License solely to contribute all or a portion of my contribution to the PostgreSQL open source project.

For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/babelfish-for-postgresql/babelfish_extensions/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #2781, opened 2024-07-25
Fix the index creation error on table created through SELECT INTO

pranavJ23:
### Description
Issue - 
If a table is formed via SELECT INTO command then the column name does not get converted to lowercase.
Due to this, indexation over that column fails which is the issue we want to resolve.
This happens as the SELECT INTO command does not convert the column name in lowercase(save it as it is) whereas the index creation command search for the column name(over which indexation is done) in lower case.
### Issues Resolved
Conversion of column name to lower case(if column name contains uppercase character/s) in Select Into query.

Note: if the table is pre-existing then user needs to drop the table and create it again to enable this feature.
### Approach
1. Down-case the column name using downcase_identifier.
2. check via filter that resjunk should be false and the column name should not be NULL.

### Task
[BABEL-4926]
Signed-off-by: Pranav Jain pranavjc@amazon.com

1. Null case i.e if the column name is null.
3. Case when multiple column names are given in SELECT INTO query
4. Case when column name is already in lowercase 
5. Case when column name is passed with a different name in SELECT INTO statement(using as).
6. Case when column is passed via some function.
7. Case when column/s is/are passed via *(select all) instead of specifying names.
8. Case when column with special letters like $,& is passed into square brackets.
9. Case when identity function is used with select into statement.
10. Added mvu test when the select into is executed in prepare step and the create index is executed in verify step
11. Case when column is modified via ALTER TABLE or ALTER COLUMN 
12. Creating UNIQUE INDEX on the column after SELECT INTO
13. Dropping column via ALTER TABLE...DROP COLUMN after SELECT INTO
14. Case when table is passed to SELECT INTO via Table-Returning Function 
15. Multibyte Column Name (Chinese)
16. Multibyte Column Name (Chinese) more then 64 characters
17. Table with multi-part names
18. Table with multi-part names and column name length > 64

### Check List
- [Pranav Jain] Commits are signed per the DCO using --signoff 


contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -6090,11 +6096,36 @@ transformSelectIntoStmt(CreateTableAsStmt *stmt)
...
+				{
+					tle->resname = downcase_identifier(tle->resname, strlen(tle->resname), false, false);
+				}	
+
 				current_resno += 1;
 				tle->resno = current_resno;
 				modifiedTargetList = lappend(modifiedTargetList, tle);
+
+				if (original_name!=NULL&&strcmp(original_name,tle->resname))
+				{
```
tanscorpio7: nit: indentation, spaces between operator, variables

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -6090,11 +6096,36 @@ transformSelectIntoStmt(CreateTableAsStmt *stmt)
...
+					newstmt = makeNode(AlterTableStmt);
+					newstmt->relation = into->rel;
+					newstmt->cmds = NIL;
+					newstmt->objtype = OBJECT_TABLE;
+					newstmt->cmds = lappend(newstmt->cmds, cmd);
+					tempStore = lappend(tempStore,newstmt);
+				}
 			}
 		}
+
```
tanscorpio7: lets remove not required changes

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -6090,11 +6096,36 @@ transformSelectIntoStmt(CreateTableAsStmt *stmt)
 			}
 			else
 			{
+				char* original_name = tle->resname;
```
tanscorpio7: only set original_name when `tle->resname!=NULL`

tanscorpio7 (changes requested):

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -6122,9 +6153,13 @@ transformSelectIntoStmt(CreateTableAsStmt *stmt)
 	}
 
 	result = lappend(result, stmt);
+	
 	if (altstmt)
 		result = lappend(result, altstmt);
 
+	if (tempStore!=NIL)
```
tanscorpio7: lets use a more descriptive variable name

test/JDBC/expected/BABEL-4926.out
```diff
@@ -0,0 +1,187 @@
...
+~~END~~
+
+-- create the BABEL_4926_REPO and select the column from BABEL_4926 
+select numb,Numb1,ColuM into BABEL_4926_REPO from BABEL_4926;
+go
+-- indexing over the columns of BABEL_4926_REPO
+create index IDX_REPRODUCTION on BABEL_4926_REPO(numb,Numb1,ColuM);
+go
+-- output is the lowercase name,original_name
+select attname,attoptions from pg_attribute where attrelid = (select oid from pg_class where relname = 'babel_4926_repo');
```
tanscorpio7: also prefix column name with jira name. So that we can only select only those column here.

tanscorpio7 (changes requested):

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -6057,6 +6058,7 @@ static List *
 transformSelectIntoStmt(CreateTableAsStmt *stmt)
 {
 	List *result;
+	List *babel_4926_original_name_list;
```
roshan0708: Let's not use jira numbers in the actual code, use some other generic name for the variables.

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -6066,6 +6068,7 @@ transformSelectIntoStmt(CreateTableAsStmt *stmt)
 	into = stmt->into;
 	result = NIL;
 	altstmt = NULL;
+	babel_4926_original_name_list = NIL;
```
roshan0708: +1

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -6159,11 +6162,40 @@ transformSelectIntoStmt(CreateTableAsStmt *stmt)
 			}
 			else
 			{
+				AlterTableStmt *newstmt;
+				AlterTableCmd *cmd;
+				char* original_name = NULL;
+				newstmt = NULL;
```
roshan0708: Try to stick to either CamelCase or SnakeCase when naming variables or functions. And try to make variable names meaningful in the current context.

test/JDBC/jdbc_schedule
```diff
@@ -7,9 +7,7 @@
 # 5. To add a test, add test name (without extension Eg. BABEL-TEST) on a
```
roshan0708: Fix: remove this file change since you're not adding anything new in the schedule file.

test/JDBC/input/BABEL-4926.sql
```diff
@@ -0,0 +1,158 @@
+-- create a new table BABEL_4926 1
```
roshan0708: Let's change the test file name. Use a name that is more suited to the fix.

test/JDBC/input/BABEL-4926.sql
```diff
@@ -0,0 +1,158 @@
...
+go
+-- dropping the table BABEL_4926
+drop table BABEL_4926;
+go
+-- dropping the table BABEL_4926_REPO
+drop table BABEL_4926_REPO;
+go
+
+
+
```
roshan0708: Refactor: Try to group tests as per the case, this makes it easier to review and check the edge cases.

test/JDBC/jdbc_schedule
```diff
@@ -7,9 +7,7 @@
...
 #    new line
 # 6. If you want the framework to not run certain files, use: ignore#!#<test name>
-
 all
-
```
rishabhtanwar29: Unnecessary change.

roshan0708:
Fix: add test file name in `upgrade/latest/schedule` file.

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -98,6 +98,7 @@ extern bool pltsql_recursive_triggers;
 extern bool restore_tsql_tabletype;
```
rishabhtanwar29: [General] Please follow commit description guidelines, we are missing `Task` and `signoff` fields. Use description of https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/2748 as an example.

roshan0708 (changes requested):
Fix: add test file name in `upgrade/latest/schedule` file.

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -6191,9 +6223,13 @@ transformSelectIntoStmt(CreateTableAsStmt *stmt)
 	}
 
 	result = lappend(result, stmt);
+	
 	if (altstmt)
 		result = lappend(result, altstmt);
 
+	if (babel_4926_original_name_list != NIL)
```
rishabhtanwar29: I think this additional variable is not needed, we can directly append into `result` if needed at line number `6195`.
[Bonus] In fact we can take this opportunity to even remove the redundant `altstmt` variable as well and directly append into `result` whenever needed.

test/JDBC/expected/BABEL-4926.out
```diff
@@ -0,0 +1,225 @@
+-- create a new table BABEL_4926 1
```
rishabhtanwar29: 1. Let's have the meaningful test filenames which reflects the purpose instead of plain Task/jira numbers, something like `TEST-SELECT_INTO`, although feel free to have even a better name.
2. Let's have test cases in upgrade format: <testname>-vu-prepare/verify/cleanup.
3. In fact, we can use the existing test files for select-into (BABEL_539-vu-*) to add new test scenarios instead of creating separate test files. See https://github.com/babelfish-for-postgresql/babelfish_extensions/commit/1ece7f885dba6379f775ee15213a86f570d88b72

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -6159,11 +6162,40 @@ transformSelectIntoStmt(CreateTableAsStmt *stmt)
...
 			{
+				AlterTableStmt *newstmt;
+				AlterTableCmd *cmd;
+				char* original_name = NULL;
+				newstmt = NULL;
+
+				if (tle->resname != NULL)
+				{
+					original_name = tle->resname;
+					tle->resname = downcase_identifier(tle->resname, strlen(tle->resname), false, false);
```
rishabhtanwar29: Can this `tle-resname` be truncated if the name is larger than 64 chars? If yes then we are going to store truncated original name which is not desirable. Let's check it once, otherwise we will need to fetch the original_name from query string same as how we generally do it.

test/JDBC/expected/BABEL-4926.out
```diff
@@ -0,0 +1,225 @@
...
+~~END~~
+
+-- Create a new table BABEL_4926_REPO with column babel_4926_COL 
+select babel_4926_COL into BABEL_4926_REPO from BABEL_4926;
+go
+-- Indexing over column babel_4926_COL
+create index IDX_REPRODUCTION on BABEL_4926_REPO(babel_4926_COL);
+go
+-- output is the lowercase name,original_name
+select attname,attoptions from pg_attribute where attname like '%babel_4926%' and attrelid = (select oid from pg_class where relname = 'babel_4926_repo') order by attname asc;
```
rishabhtanwar29: Not fancy enough ;-), let's come up with a query having join between pg_attribute and pg_class. Moreover, we don't want full `attoptions` list here, instead the query can be modified to just print the `bbf_original_name` element of that list.

[BONUS] Let's see if we can use `information_schema.columns` view instead for this purpose in which case you can ignore the suggestion above.

test/JDBC/expected/BABEL-4926.out
```diff
@@ -0,0 +1,225 @@
...
+
+-- create a new table BABEL_4926 7
+create table BABEL_4926([babel_4926_COL$] int);
+go
+select [babel_4926_COL$] from BABEL_4926; 
+go
+~~START~~
+int
+~~END~~
+
```
rishabhtanwar29: Let's also add a tests with column names having length larger than 64 chars, special characters, quoted with `[]` and `""` etc.

test/JDBC/input/BABEL-4926.sql
```diff
@@ -0,0 +1,158 @@
...
+go
+-- output is the lowercase name,original_name
+select attname,attoptions from pg_attribute where attname like '%babel_4926%' and attrelid = (select oid from pg_class where relname = 'babel_4926_repo') order by attname asc;
+go
+-- dropping the table BABEL_4926
+drop table BABEL_4926;
+go
+-- dropping the table BABEL_4926_REPO
+drop table BABEL_4926_REPO;
+go
```
rishabhtanwar29: Nit: please add a newline at end of each new file.

contrib/babelfishpg_tsql/src/pl_handler.c
forestkeeper: This naming looks confusing

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -6159,11 +6162,40 @@ transformSelectIntoStmt(CreateTableAsStmt *stmt)
 			}
 			else
 			{
+				AlterTableStmt *newstmt;
```
forestkeeper: We should add else if () ?? some condition to enter this block here, we can't make all the else conditions into this block actually.

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -98,6 +98,7 @@ extern bool pltsql_recursive_triggers;
 extern bool restore_tsql_tabletype;
 extern bool babelfish_dump_restore;
 extern bool pltsql_nocount;
+extern const char *ATTOPTION_BBF_ORIGINAL_NAME;
```
forestkeeper: Can you show me a a const define case using Upper naming only ?

contrib/babelfishpg_tsql/src/pl_handler.c
pranavJ23: I tried this but It was throwing an error, we are first appending the SelectInto statement commands and after that appending the altstmt or newstmt. if we directly append it to result then the newstmt will be in queue before selectInto commands and it will throw an error while executing(it won't find the new table that we have created using the select into statement)

forestkeeper:
Please polish the description and make those points clear : 
1. what's the root cause ( Upper case/Lower case issue ? But more details pls )
2. How do we fix this ( for example : change the behavior of select into in certain step)
3. Test cases design ( we already have that, that's good )
4. What if the table is already created by previous wrong behavior, what will happen when create index on those tables ? ( Or we already covered it ? )

forestkeeper (changes requested):
Pls add mvu test when the select into is executed in prepare step and the create index is executed in verify step

contrib/babelfishpg_tsql/src/pl_handler.c
pranavJ23: This is an existing variable I am using

test/JDBC/expected/BABEL-4926.out
pranavJ23: done

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -6181,11 +6184,40 @@ transformSelectIntoStmt(CreateTableAsStmt *stmt)
...
+					original_name = tle->resname;
+					tle->resname = downcase_identifier(tle->resname, strlen(tle->resname), false, false);
+				}	
+
 				current_resno += 1;
 				tle->resno = current_resno;
 				modifiedTargetList = lappend(modifiedTargetList, tle);
+
+				if (original_name != NULL && strcmp(original_name, tle->resname) && strlen(original_name) == NAMEDATALEN)
+				{
```
tanscorpio7: `strlen(original_name) < NAMEDATALEN`
only then we save the original name.

tanscorpio7 (changes requested):

contrib/babelfishpg_tsql/src/pl_handler.c
tanscorpio7: actually truncation makes the name < NAMEDATALEN. so we may not be able to tell if a name is truncated or not just by looking at the length.

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -6181,11 +6184,40 @@ transformSelectIntoStmt(CreateTableAsStmt *stmt)
...
+					original_name = tle->resname;
+					tle->resname = downcase_identifier(tle->resname, strlen(tle->resname), false, false);
+				}	
+
 				current_resno += 1;
 				tle->resno = current_resno;
 				modifiedTargetList = lappend(modifiedTargetList, tle);
+
+				if (original_name != NULL && strcmp(original_name, tle->resname))
```
forestkeeper: Pls add comments about what this if block trying to do?

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -6180,6 +6185,38 @@ transformSelectIntoStmt(CreateTableAsStmt *stmt)
...
+				original_name = tle->resname;
+				tle->resname = downcase_identifier(tle->resname, strlen(tle->resname), false, false);
+
+				// check if the original_name is already in lowercase
+				if (original_name != NULL && strcmp(original_name, tle->resname))
+				{
+					AlterTableStmt *newstmt;
+					AlterTableCmd *cmd;
+
+					newstmt = NULL;
```
tanscorpio7: We can remove this line.

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -98,6 +98,8 @@ extern bool pltsql_recursive_triggers;
 extern bool restore_tsql_tabletype;
 extern bool babelfish_dump_restore;
 extern bool pltsql_nocount;
+extern const char *ATTOPTION_BBF_ORIGINAL_NAME;
+extern TargetEntry *get_tle_by_resno(List *tlist,AttrNumber resno);
```
tanscorpio7: we can remove this line

tanscorpio7 (changes requested):

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -6261,11 +6294,14 @@ transformSelectIntoStmt(CreateTableAsStmt *stmt)
...
 			else
 			{
+				
 				current_resno += 1;
 				tle->resno = current_resno;
 				modifiedTargetList = lappend(modifiedTargetList, tle);
+				
 			}
 		}
+
```
tanscorpio7: nit: not needed change.

tanscorpio7 (changes requested):

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -6180,6 +6183,36 @@ transformSelectIntoStmt(CreateTableAsStmt *stmt)
 		foreach (elements, q->targetList)
 		{
 			TargetEntry *tle = (TargetEntry *)lfirst(elements);
+			char* original_name = NULL;
+
+			if (tle->resname != NULL && !tle->resjunk && strlen(tle->resname) < NAMEDATALEN -1)
```
tanscorpio7: Nit: spacing `char *original_name` &  `NAMEDATALEN - 1`

tanscorpio7 (approved):

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -6180,6 +6183,36 @@ transformSelectIntoStmt(CreateTableAsStmt *stmt)
 		foreach (elements, q->targetList)
 		{
 			TargetEntry *tle = (TargetEntry *)lfirst(elements);
+			char *original_name = NULL;
+
+			if (tle->resname != NULL && !tle->resjunk && strlen(tle->resname) < NAMEDATALEN - 1)
+			{	
+
+				original_name = tle->resname;
+				tle->resname = downcase_identifier(tle->resname, strlen(tle->resname), false, false);
```
rishabhtanwar29, line 6251: resname should be downcased irrespective of whether we can find original_name or not or in other words, down-casing should not get affect by whether or not `strlen(tle->resname) < NAMEDATALEN - 1`

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -6180,6 +6183,36 @@ transformSelectIntoStmt(CreateTableAsStmt *stmt)
...
+
+				// check if the original_name is already in lowercase
```
rishabhtanwar29: Use PG-style comments: /* XXX */

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -6293,9 +6328,13 @@ transformSelectIntoStmt(CreateTableAsStmt *stmt)
 	}
 
 	result = lappend(result, stmt);
```
rishabhtanwar29: My initial comment was to move this line up in the top and directly append everything else into `result` instead of creating new lists.

test/JDBC/expected/SELECT_INTO_TEST-vu-verify.out
```diff
@@ -0,0 +1,304 @@
...
+~~END~~
+
+-- Create a new table select_into_REPO with column select_into_COL 
+select select_into_COL into select_into_REPO from select_into;
+go
+-- Indexing over column select_into_COL
+create index IDX_REPRODUCTION on select_into_REPO(select_into_COL);
+go
+-- output is the lowercase name,original_name
+select attname,attoptions from pg_attribute join pg_class on pg_attribute.attrelid=pg_class.oid where pg_class.relname='select_into_repo' and attname like '%select_into%'order by attname asc;
```
rishabhtanwar29: Have you considered using `information_schema.columns` instead?

test/JDBC/expected/SELECT_INTO_TEST-vu-verify.out
```diff
@@ -0,0 +1,304 @@
...
+
+-- Create a new table select_into_REPO with column select_into_COL 
+select select_into_COL_select_into_COL_select_into_COL_select_into_COL_ into select_into_REPO from select_into;
+go
+-- Indexing over column select_into_COL
+create index IDX_REPRODUCTION on select_into_REPO(select_into_COL_select_into_COL_select_into_COL_select_into_COL_);
+go
+~~ERROR (Code: 33557097)~~
+
+~~ERROR (Message: column "select_into_col_select_into_colb68a72f6d624cf699ece5bf8f6446ce4" does not exist)~~
```
rishabhtanwar29: This shouldn't be failing.

pranavJ23: here the column name length > 63, so we are not lowercasing the column name for this case and not storing the original name

tanscorpio7: We want to lowercase the name for greater than 64 as well 
but not store the original name.

pranavJ23: Done

contrib/babelfishpg_tsql/src/pl_handler.c
pranavJ23: it was not work, as altstmt should be appended before our column list. Altstmt create the identity column and our alterTablestmt(newstmt, which stores the original name) should be appended after Altstmt (the alterTablestmt which creates the identity column).

test/JDBC/expected/SELECT_INTO_TEST-vu-verify.out
pranavJ23: We want to test both downcased name and original name for the column whereas we will only get the original name from information_schema.columns

rishabhtanwar29: information_schema.columns returns either original_name or stored name depending upon whether or not the original_name exists.

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -6180,6 +6183,36 @@ transformSelectIntoStmt(CreateTableAsStmt *stmt)
...
+			char *original_name = NULL;
+
+			if (tle->resname != NULL && !tle->resjunk)
+			{	
+
+				original_name = tle->resname;
+				tle->resname = downcase_identifier(tle->resname, strlen(tle->resname), false, false);
+
+				/* check if the original_name is already in lowercase */
+				if (original_name != NULL && strcmp(original_name, tle->resname) && strlen(tle->resname) < NAMEDATALEN - 1)
```
rishabhtanwar29: I suspect `strlen(tle->resname) < NAMEDATALEN - 1` might differ for multi-bytes characters, let's check with @Deepesh125 regarding it.

rishabhtanwar29 (approved):

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -6294,7 +6295,7 @@ transformSelectIntoStmt(CreateTableAsStmt *stmt)
 
 	result = lappend(result, stmt);
 	if (altstmt)
-		result = lappend(result, altstmt);
+		result = lappend(result, altstmt);
```
rishabhtanwar29: Nit: unnecessary change.

rishabhtanwar29 (dismissed):

test/JDBC/expected/SELECT_INTO_TEST-vu-verify.out
```diff
@@ -0,0 +1,300 @@
...
+go
+
+--dependent objects
+create index IDX_REPRODUCTION on select_into_pre_exist_repro(select_into_pre_exist_COL);
+go
+drop table if exists select_into_pre_exist;
+go
+drop table if exists select_into_pre_exist_repro;
+go
+-- column length >=64 todo
```
Deepesh125: with multibyte characters also, from Japanese, Chinese, etc

test/JDBC/expected/SELECT_INTO_TEST-vu-verify.out
```diff
@@ -0,0 +1,300 @@
...
+go
+~~START~~
+int
+~~END~~
+
+-- Create a new table select_into_REPO with column select_into_COL 
+select select_into_COL into select_into_REPO from select_into;
+go
+-- Indexing over column select_into_COL
+create index IDX_REPRODUCTION on select_into_REPO(select_into_COL);
```
Deepesh125, line 15: Let's also add test cases to test 
1. ALTER TABLE...ALTER COLUMN works fine on this column
2. Add new constraint through ALTER TABLE ... ADD CONSTRAINT
3. CREATE UNIQUE INDEX
4. ALTER TABLE ... DROP COLUMN

test/JDBC/input/SELECT_INTO_TEST-vu-prepare.sql
```diff
@@ -0,0 +1,12 @@
+-- create a new table select_into
+create table select_into_pre_exist(select_into_pre_exist_COL int);
+go
+-- create a new function to double the value
+CREATE FUNCTION dbo.select_into_double_function (@in INT) RETURNS INT
```
Deepesh125, line 5: Test with table returning function also

test/JDBC/expected/SELECT_INTO_TEST-vu-verify.out
```diff
@@ -0,0 +1,300 @@
+
```
Deepesh125, line 1: why aren't we testing any sys or information_schema_tsql views anywhere? How are we making sure that sys.column (for example) is showing correct details

test/JDBC/expected/SELECT_INTO_TEST-vu-verify.out
```diff
@@ -0,0 +1,300 @@
...
+go
+-- Indexing over column select_into_COL
+create index IDX_REPRODUCTION on select_into_REPO(select_into_COL_select_into_COL_select_into_COL_select_into_COL_);
+go
+-- output is the lowercase name
+select attname from pg_attribute join pg_class on pg_attribute.attrelid=pg_class.oid where pg_class.relname='select_into_repo' and attname like '%select_into%'order by attname asc;
+go
+~~START~~
+varchar
+select_into_col_select_into_colb68a72f6d624cf699ece5bf8f6446ce4
```
Deepesh125, line 247: just trying to understand, where is this truncation (ident + MD5) is happening when we are passing `truncate = false` to `downcase_identifier`?

Deepesh125 (changes requested):

test/JDBC/expected/SELECT_INTO_TEST-vu-verify.out
pranavJ23: tle->resname is already truncated and MD5 file is already attached to it

test/JDBC/expected/SELECT_INTO_TEST-vu-verify.out
```diff
@@ -298,3 +298,242 @@ go
...
+~~START~~
+int
+1
+2
+3
+~~END~~
+
+
+-- Indexing over column 中文_COL
+create index IDX_REPRODUCTION on select_into_REPO(中文_COL);
```
Deepesh125, line 522: with length > 64 also

Deepesh125 (changes requested):

Deepesh125 (dismissed):

test/JDBC/input/SELECT_INTO_TEST-vu-verify.sql
```diff
@@ -0,0 +1,399 @@
+
```
Deepesh125, line 1: Just realised that we missed adding tests cases for following scenarios - 
1. When source table's column length is >=64 and/or contains multi-byte characters
2. When source table's column is specified as multi-part, say, `select master.dbo.table.xyz into ...` , `select dbo.table.xyz into ...` , etc

Deepesh125 (dismissed):

roshan0708 (approved):

test/JDBC/upgrade/latest/schedule
```diff
@@ -561,5 +561,6 @@ reverse
 sys_sequences
 replicate
 space
+SELECT_INTO_TEST
```
forestkeeper, line 564: why we only add this into latest ?

forestkeeper (approved):

shardgupta (approved):

PR #2784, opened 2024-07-25
Support $PARTITION Function in Babelfish

sumitj824:
### Description
This commit adds support for the $PARTITION function in Babelfish, which allows users to retrieve the partition number for a given value and partition function.

### Issues Resolved
Task: BABEL-5021

* **Use case based -** Yes

* **Arbitrary inputs -** Yes

* **Performance tests -** NA

* **Tooling impact -** NA

* **Client tests -** NA

### Check List


contrib/babelfishpg_tsql/src/tsqlIface.cpp
```diff
@@ -1234,6 +1234,41 @@ class tsqlCommonMutator : public TSqlParserBaseListener
...
+		 * to sys.search_partition(partition_func_name, exp, db_name)
+		 * where db_name is optional.
+		 */
+
+		/* Remove dot after the $PARTITION and partition function name from query. */
+		rewritten_query_fragment.emplace(std::make_pair(ctx->DOT().back()->getSymbol()->getStartIndex(), std::make_pair(::getFullText(ctx->DOT().back()), "")));
+		rewritten_query_fragment.emplace(std::make_pair(ctx->id().back()->start->getStartIndex(), std::make_pair(::getFullText(ctx->id().back()), "")));
+		
+		/* Replace "$PARTITION" with "sys.search_partition". */
+		rewritten_query_fragment.emplace(std::make_pair(ctx->DOLLAR_PARTITION()->getSymbol()->getStartIndex(), std::make_pair(::getFullText(ctx->DOLLAR_PARTITION()),
```
KushaalShroff: Updated comment after speaking to Sumit: So we can use this $Partition inside a query, which is analogous to spatial datatype functions like STX. This implementation looks good to me but can you check why we followed re-writting STX at time of pre-process hook and not in ANTLR?

KushaalShroff (changes requested):

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--4.2.0--4.3.0.sql
```diff
@@ -6749,6 +6749,17 @@ END;
 $$
 LANGUAGE 'pltsql';
 
+CREATE OR REPLACE FUNCTION sys.search_partition(IN func_name sys.NVARCHAR(128), IN arg anyelement, IN db_name sys.NVARCHAR(128) DEFAULT NULL)
```
KushaalShroff: Lets remove redundant object for NULL arg, we can add explicit cast for NULL::TEXT

contrib/babelfishpg_tsql/src/tsqlIface.cpp
```diff
@@ -1234,6 +1234,41 @@ class tsqlCommonMutator : public TSqlParserBaseListener
...
+	{
+		/*
+		 * Re-write db_name.$PARTITION.partition_func_name(exp)
+		 * to sys.search_partition(partition_func_name, exp, db_name)
+		 * where db_name is optional.
+		 */
+
+		/* Remove dot after the $PARTITION and partition function name from query. */
+		rewritten_query_fragment.emplace(std::make_pair(ctx->DOT().back()->getSymbol()->getStartIndex(), std::make_pair(::getFullText(ctx->DOT().back()), "")));
+		rewritten_query_fragment.emplace(std::make_pair(ctx->id().back()->start->getStartIndex(), std::make_pair(::getFullText(ctx->id().back()), "")));
```
KushaalShroff, line 1257: Why are we rewriting index by index when we have to replace the entire Partition_function_callContext string??

Lets make it more cleaner by first fetching all the data we want and build a temp string and then emplace the entire ctx string with this temp one

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -1916,6 +1920,186 @@ checksum(PG_FUNCTION_ARGS)
...
+	table_close(rel, AccessShareLock);
+
+	/* Raise error if provided partition function doesn't exist in the provided database. */
+	if (!func_param_typname)
+		ereport(ERROR,
+				(errcode(ERRCODE_INVALID_PARAMETER_VALUE),
+					errmsg("Invalid object name '%s'.", partition_func_name)));
+	
+	/* NULL values will always belong to first partition. */
+	if (PG_ARGISNULL(1))
```
KushaalShroff, line 2054: I see, need to add comment about the chronology of the error/behaviour in TSQL at the top of the function

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -1916,6 +1920,186 @@ checksum(PG_FUNCTION_ARGS)
...
+	 * Find oid of comparator function for input type, which will be used during the sorting.
+	 * Here, we are first finding the default operator class for the input type then using that
+	 * we are finding the operator family for that operator class and finally using that we are
+	 * finding the defined comparator function for that operator family.
+	 */
+	basetype_oid = getBaseType(sqlvariant_typoid);
+	opclass_oid = GetDefaultOpClass(basetype_oid, BTREE_AM_OID);
+	opfamily_oid = get_opclass_family(opclass_oid);
+	cmpfunction_oid = get_opfamily_proc(opfamily_oid, basetype_oid, basetype_oid,
+						BTORDER_PROC);
```
KushaalShroff: Lets use the sql_variant_cmp oid directly

KushaalShroff (changes requested):

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--4.2.0--4.3.0.sql
sumitj824: discussed it offline, it will be needed to handle const NULL and string literals.

contrib/babelfishpg_tsql/runtime/functions.c
sumitj824: addressed.

contrib/babelfishpg_tsql/runtime/functions.c
sumitj824: addressed.

contrib/babelfishpg_tsql/src/tsqlIface.cpp
sumitj824: you're right, Ideally we should be replacing whole query by query but there are following challenges due to which we we replacing one by one:

1. If we try to replace whole query including exp by new one, then for local_id in exp we will not get correct rewritten query.
2. If we want replace subpart of query, then we aren't not able find proper original subpart neither by concatenating each of nodes one by one(we will miss space and comment) nor by using substr (will fail for multibytes).

KushaalShroff (approved):

PR #2795, opened 2024-07-26
Fix error when OUTPUT INTO table variable statement fires trigger 

tanscorpio7:
### Description

Cherry Picked: https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/2639

Fix error portal snapshots (1) did not account for all active snapshots (2) when OUTPUT INTO table variable statement fires triggers.

pltsql statements that could fire trigger must always run inside a transaction block. If a txn block is not active, we create one. We exclude some types of statements from this, one of them being DML on table variables since they can never fire a trigger. We detect this using the variable mod_stmt_tablevar.

Now an edge case here is OUTPUT into table variable stmt. Which we internally transform into two DML nodes. TOP INSERT NODE (signifies the output into clause) Another DELETE / UPDATE / INSERT node inside the top insert node -> with clause (specifies the first part of the output into clause ) As a fix we simply skip transaction commands inside pltsql UDFs and not rely on mod_stmt_tablevar on anymore.

Also blocked certain cases of OUTPUT clause which should not be allowed inside pltsql UDFs.

    OUTPUT clause flushes results to client (INTO clause is missing)
    OUTPUT INTO clause has a non local object as its target. (UDFs only allow DML on local objects)

### Issues Resolved

[BABEL-4859]

### Check List

### DIFF
--- a/contrib/babelfishpg_tsql/src/pl_exec.c
+++ b/contrib/babelfishpg_tsql/src/pl_exec.c
@@ -4630,6 +4630,9 @@ exec_stmt_execsql(PLtsql_execstate *estate,
 
 	/* fetch current search_path */
 	char	   *old_search_path = NULL;
+	bool		ro_func = (estate->func->fn_prokind == PROKIND_FUNCTION) &&
+						  (estate->func->fn_is_trigger == PLTSQL_NOT_TRIGGER) &&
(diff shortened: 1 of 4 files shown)


jsudrik (approved):
Approved.

PR #2810, opened 2024-07-30
Fix behaviour of error message for GRANT

aswiniip:
### Description
Currently on executing queries like `GRANT EXECUTE ON PROCEDURE::test TO public;`, babelfish throws error with error message `GRANT ON PROCEDURE is not currently supported in Babelfish` which is incorrect as the above query has an incorrect syntax. With this change it now shows `syntax error` for such cases and `not currently supported` for allowed object types (according to GRANT documentation) which are not supported by us currently.

The issue arrised because the GRANT query is being successfully parsed by our antlr parser and then the check for supported/unsupported occurs resulting in the error being thrown in the `TsqlUnsupportedFeatureHandler()` rather than the antlr parser.

### Changes Made:
In the rule definition of GRANT/REVOKE/DENY in `TSqlParser.g4`, `object_type` is used which contains some types like PROCEDURE, TABLE which are not syntactically correct according to the GRANT/REVOKE/DENY documentation. In this PR, I have created a new `permission_object_type` which contains only those types which are syntactically correct for GRANT/REVOKE/DENY and added relevant tests for the same. After this change, babelfish throws error during the parsing and gives syntax error.

### Issues Resolved
BABEL-5079

* **Use case based -** Yes

* **Arbitrary inputs -** Yes

* **Performance tests -**

* **Tooling impact -**

* **Client tests -**

### Check List

Signed-off-by: P Aswini Kumar [aswinii@amazon.com](mailto:aswinii@amazon.com)


HarshLunagariya (approved):

contrib/babelfishpg_tsql/antlr/TSqlParser.g4
```diff
@@ -2935,6 +2935,32 @@ single_permission
     | UNSAFE ASSEMBLY
     ;
 
+permission_object_type
+    : APPLICATION ROLE
```
forestkeeper: We should write this into format like : 
(USING_SYMBOL | TYPE_SYMBOL| .. | )

forestkeeper: That format the antlr performance would be better

forestkeeper (changes requested):
Pls refer to this blog and rewrite some of the antlr rules : https://tomassetti.me/improving-the-performance-of-an-antlr-parser/

forestkeeper (approved):

PR #2811, opened 2024-07-30
Procedure body is stored in incorrect syntax in information_schema.routines after ALTER PROC

Jakeowen1:
### Description
Currently when procedures are modified using `alter procedure` in Babelfish, the `information_schema.routines` catalog updates the definition to be `alter procedure ...` which is incorrect behavior. This commit fixes the information_schema entry to store the alter procedure as `create procedure ...`

### Issues Resolved
BABEL-4971

* **Use case based -**
```
ALTER PROCEDURE alter_proc_p1
AS
    select * from alter_proc_orders
GO

select ROUTINE_NAME, ROUTINE_BODY, ROUTINE_DEFINITION from information_schema.routines where SPECIFIC_NAME LIKE 'alter_proc_p1';
go
~~START~~
nvarchar#!#nvarchar#!#nvarchar
alter_proc_p1#!#SQL#!#CREATE PROCEDURE alter_proc_p1<newline>AS<newline>    select * from alter_proc_orders
~~END~~
```

### Check List


contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -3477,6 +3477,11 @@ pltsql_store_func_default_positions(ObjectAddress address, List *parameters, con
 
 	if (HeapTupleIsValid(oldtup))
 	{
+		if(original_query && strncasecmp("alter", original_query, 5) == 0) {
```
forestkeeper: Is it possible the original_sql = " alter xxx" , ( starts with a space or \r\n )

Jakeowen1: No it gets parsed so even if there are spaces before `ALTER` original_sql will not have those spaces / extra characters in front

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -3477,6 +3477,15 @@ pltsql_store_func_default_positions(ObjectAddress address, List *parameters, con
 
 	if (HeapTupleIsValid(oldtup))
 	{
+		char* alter_str = strtok(original_query, " ");
+		if(alter_str && strncasecmp("alter", alter_str, 5) == 0) {
```
2jungkook: Please follow PG's coding style:
```
if (...)
{
  ...
}
```

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -3477,6 +3477,15 @@ pltsql_store_func_default_positions(ObjectAddress address, List *parameters, con
...
+			char* proc_str = strtok(NULL, " ");
```
2jungkook: What if there are `\n, \t, \r, \f` in the original query? Please add such test cases.

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -3477,6 +3477,15 @@ pltsql_store_func_default_positions(ObjectAddress address, List *parameters, con
...
+			if(proc_str && strncasecmp("proc", proc_str, 4) == 0) {
+				/* Modify the original string definiton for information_schema.routines */
+				char* newQuery = psprintf("CREATE %s", (queryString + 6));
```
2jungkook: I don't think this is the right approach. It will wipe out some comments in the original `create` query, e.g.,
```
create /* some comments */   procedure alter_proc_p1 as select 1;
```
, and it may not be able to handle comments in the `alter` query, e.g.,
```
alter -- another comment
    procedure alter_proc_p1 as select 2;
```
```

**Review by 2jungkook — CHANGES_REQUESTED** [review]

File: test/JDBC/expected/alter-procedure-vu-verify.out [hunk 5]
```diff
@@ -51,11 +52,20 @@ int#!#int#!#int#!#int#!#date
...
+alter_proc_p1#!#SQL#!#CREATE       -- test comment<newline>    PROCEDURE alter_proc_p1<newline>AS<newline>    select * from alter_proc_orders
+~~END~~
+
 
 
 -- Test Case: Modify the procedure body, add a parameter, use "proc"
---            instead of "procedure"              
-ALTER PROC alter_proc_p1
+--            instead of "procedure"
+       ALTER /* TEST COMMENT */
```
forestkeeper, line 67: Is there test case like : 
```
alter 

/*
test comment
*/

-- another comment

proc ddd

...
```

test/JDBC/expected/alter-procedure-vu-verify.out
```diff
@@ -27,8 +27,9 @@ GO
 ~~ERROR (Message: No existing procedure found with the name for ALTER PROCEDURE)~~
 
 
--- Test Case: Modify the procedure body
-ALTER PROCEDURE alter_proc_p1
+-- Test Case: Modify the procedure body, and check information_schema updated with spaces
+ALTER       -- test comment
```
forestkeeper: Can we also try 
```
aLTER

-- test comment 

proc
```

forestkeeper:
The last concern we had is: is it always reliable that replacing the alter with create will be a legal sql ?

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -3477,6 +3477,50 @@ pltsql_store_func_default_positions(ObjectAddress address, List *parameters, con
...
+				while(original_query[i] != '*' && original_query[i+1] != '/')
+				{
+					appendStringInfoChar(&infoSchemaStr, original_query[i]);
+					i++;
+				}
+				appendStringInfoChar(&infoSchemaStr, original_query[i]);
+				appendStringInfoChar(&infoSchemaStr, original_query[i+1]);
+				i++;
+			}
+			else if(strncasecmp(original_query + i, "create", 6) == 0)
```
forestkeeper: Will it be wrong is original_query pointer has less then 6 char left before it's end ?

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -3477,6 +3477,50 @@ pltsql_store_func_default_positions(ObjectAddress address, List *parameters, con
...
+			{
+				break;
+			}
+			else if(strncasecmp(original_query + i, "alter", 5) == 0)
```
forestkeeper: Will it be wrong is original_query pointer has less then 5 char left before it's end ?

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -3477,6 +3477,50 @@ pltsql_store_func_default_positions(ObjectAddress address, List *parameters, con
...
+			{
+				// Change alter to create, add rest of characters, and update
+				appendStringInfoString(&infoSchemaStr, "CREATE");
+				appendStringInfoString(&infoSchemaStr, original_query + i + 5);
```
forestkeeper: Let's add a check to the follow up token should only be 'proc'/'procedure', although it maybe redundant, but will secure safe, otherwise one missing will lead to a customer impact result.

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -4235,6 +4239,88 @@ pg_proc_update_oid_acl(ObjectAddress address, Oid oid, Acl *acl)
...
+			if(original_query[i] == '-')
+			{
+				while(original_query[i] != '\n' && i < strlen(original_query))
+				{
+					appendStringInfoChar(&infoSchemaStr, original_query[i]);
+					i++;
+				}
+				appendStringInfoChar(&infoSchemaStr, original_query[i]);
+			}
+			else if(original_query[i] == '/' && original_query[i + 1] == '*' && i < strlen(original_query))
```
forestkeeper: before check original_query[i + 1] == '*', should also check i+1 < strlen(original_query), otherwise it will have the risk of segment fault.

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -4235,6 +4239,88 @@ pg_proc_update_oid_acl(ObjectAddress address, Oid oid, Acl *acl)
...
+			{
+				while(original_query[i] != '*' && original_query[i+1] != '/')
```
forestkeeper: Should add check i+1<strlen() before checking original_query[i+1]

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -4235,6 +4239,88 @@ pg_proc_update_oid_acl(ObjectAddress address, Oid oid, Acl *acl)
...
+				{
+					appendStringInfoChar(&infoSchemaStr, original_query[i]);
+					i++;
+				}
+				appendStringInfoChar(&infoSchemaStr, original_query[i]);
+				appendStringInfoChar(&infoSchemaStr, original_query[i+1]);
+				i++;
+			}
+			else if(strncasecmp(original_query + i, "alter", 5) == 0)
```
forestkeeper: strncasecmp didn't guarantee it'll check the end of string, before the if , should also make sure it didn't reach the boundary, like i+5 < strlen()

forestkeeper (dismissed):
LGTM

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -4235,6 +4239,64 @@ pg_proc_update_oid_acl(ObjectAddress address, Oid oid, Acl *acl)
...
+ * "ALTER ..." to "CREATE ..." in bbf_function_ext
+ */
+static void bbf_func_ext_update_proc_definition(Oid oid)
+{
+	Relation	bbf_function_ext_rel;
+	TupleDesc	bbf_function_ext_rel_dsc;
+	Datum 		new_record[BBF_FUNCTION_EXT_NUM_COLS];
+	bool		new_record_replaces[BBF_FUNCTION_EXT_NUM_COLS];
+	bool		new_record_nulls[BBF_FUNCTION_EXT_NUM_COLS];
+	char 		*original_query = get_original_query_string();
```
2jungkook, line 4253: Please add a null check.

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -4235,6 +4239,64 @@ pg_proc_update_oid_acl(ObjectAddress address, Oid oid, Acl *acl)
...
+	HeapTuple	tuple,
+				proctup,
+				oldtup;
+	
+	
+	proctup = SearchSysCache1(PROCOID, ObjectIdGetDatum(oid));
+	if (!HeapTupleIsValid(proctup)) // throw error
+		return;
```
2jungkook: We are not throwing an error here. Please update the comment or logic.

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -4235,6 +4239,64 @@ pg_proc_update_oid_acl(ObjectAddress address, Oid oid, Acl *acl)
...
+		tuple = heap_modify_tuple(oldtup, bbf_function_ext_rel_dsc,
+								  new_record, new_record_nulls,
+								  new_record_replaces);
+		CatalogTupleUpdate(bbf_function_ext_rel, &tuple->t_self, tuple);
+		heap_freetuple(oldtup);
+
+		/* Clean up */
+		ReleaseSysCache(proctup);
+		heap_freetuple(tuple);
+		table_close(bbf_function_ext_rel, RowExclusiveLock);
```
2jungkook: This does not happen when `HeapTupleIsValid(oldtup)` is false.

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -4235,6 +4239,64 @@ pg_proc_update_oid_acl(ObjectAddress address, Oid oid, Acl *acl)
...
+			appendStringInfoString(&infoSchemaStr, original_query + 5);
+			new_record[Anum_bbf_function_ext_definition - 1] = CStringGetTextDatum(infoSchemaStr.data);
+			new_record_replaces[Anum_bbf_function_ext_definition - 1] = true;
+		}
+
+		tuple = heap_modify_tuple(oldtup, bbf_function_ext_rel_dsc,
+								  new_record, new_record_nulls,
+								  new_record_replaces);
+		CatalogTupleUpdate(bbf_function_ext_rel, &tuple->t_self, tuple);
+		heap_freetuple(oldtup);
```
2jungkook: This won't be free'd when an exception happens in `table_open()`.

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -4235,6 +4239,64 @@ pg_proc_update_oid_acl(ObjectAddress address, Oid oid, Acl *acl)
...
+			/* Change alter to create, add rest of characters, and update */
+			appendStringInfoString(&infoSchemaStr, "CREATE");
+			appendStringInfoString(&infoSchemaStr, original_query + 5);
+			new_record[Anum_bbf_function_ext_definition - 1] = CStringGetTextDatum(infoSchemaStr.data);
+			new_record_replaces[Anum_bbf_function_ext_definition - 1] = true;
+		}
+
+		tuple = heap_modify_tuple(oldtup, bbf_function_ext_rel_dsc,
+								  new_record, new_record_nulls,
+								  new_record_replaces);
```
2jungkook: If the above `if` statement is false, we may call `heap_modify_tuple()` with a garbage `new_record` value. Is this an expected behavior? Do we have a such test case?

test/JDBC/expected/alter-procedure-vu-prepare.out
```diff
@@ -22,6 +22,9 @@ go
 create procedure alter_proc_p3 as select 1
 go
 
+create procedure alter_proc_p4 as select 1
```
2jungkook, line 25: I don't see a test case where we do `alter procedure` in a prepare script and verify it in a verify script. Please add this case, if there's no such test already.

contrib/babelfishpg_tsql/src/pl_handler.c
Jakeowen1: new_record_replaces is initialized to false and only is true when the if statement is hit. Therefore if this function were run with an original query that does not start with "alter" nothing would update. This case should never happen though because bbf_func_ext_update_proc_definition is called from the alter procedure case

2jungkook: If you think this case should never happen then please change the if condition to `Assert()`. In that way, you give a clear context and assumption for future developers.

contrib/babelfishpg_tsql/src/pl_handler.c
2jungkook: Could you check this?

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -4235,6 +4239,72 @@ pg_proc_update_oid_acl(ObjectAddress address, Oid oid, Acl *acl)
...
+		heap_freetuple(oldtup);
+
+		/* Clean up */
+		ReleaseSysCache(proctup);
+		heap_freetuple(tuple);
+		table_close(bbf_function_ext_rel, RowExclusiveLock);
+	}
+	else
+	{
+		heap_freetuple(oldtup);
```
2jungkook: We should not free invalid tuple.

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -4235,6 +4239,72 @@ pg_proc_update_oid_acl(ObjectAddress address, Oid oid, Acl *acl)
...
+								  new_record_replaces);
+		CatalogTupleUpdate(bbf_function_ext_rel, &tuple->t_self, tuple);
+		heap_freetuple(oldtup);
+
+		/* Clean up */
+		ReleaseSysCache(proctup);
+		heap_freetuple(tuple);
+		table_close(bbf_function_ext_rel, RowExclusiveLock);
+	}
+	else
```
2jungkook: Reducing unnecessary indentation is recommended for code readability, e.g.,
```
if (!HeapTupleIsValid(oldtup))
{
   ReleaseSysCache(proctup);
   elog(ERROR, "cache lookup failed for function %u", oid);
}

// now we can assume oldtup is valid
```

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -4235,6 +4239,84 @@ pg_proc_update_oid_acl(ObjectAddress address, Oid oid, Acl *acl)
...
+	}
+
+	/*
+	 * This solution only works because original_query does not contain
+	 * any leading characters or comments before "ALTER". When BABEL-5140
+	 * is resolved we will need to refactor this code
+	 */
+	if(!(strlen(original_query) >= 5 && strncasecmp(original_query, "alter", 5) == 0))
+	{
+		elog(ERROR, "original query: %s, is improperly formatted", original_query);
```
2jungkook, line 4276: table close

2jungkook (approved):

PR #2829, opened 2024-08-02
Fix difference in behaviour for STUFF() function with multibyte characters

Anikait143:
## Description

### 1. Issues
Arguments datatype and return datatype of STUFF() string function is not defined correctly in babelfish, which is causing  output difference between TSQL and Babelfish.

- Added wrapper function handling for STUFF() function to ensure correct input arguments datatype and return datatype.
```
# Input
1> DECLARE @string1 NVARCHAR(30) = N'比尔·拉', @string2 VARCHAR(30) = '比'
2> SELECT STUFF(@string1, 1, 1, @string2)
3> SELECT STUFF(@string1, 2, 1, @string2)
4> SELECT STUFF(@string1, 3, 1, @string2)
5> go
```
```

stuff 
??·? 

(1 rows affected)
stuff 
??·? 

(1 rows affected)
stuff 
???? 

(1 rows affected)
```
```
# current output (correct output)  (NOTE: there is still difference from TSQL but that is because the way we store multi-byte characters in VARCHAR, issues related to STUFF() function are fixed)

stuff                                                                                                                                                                                                                                                           
比尔·拉                                                                                                                                                                                                                                                            

(1 rows affected)
stuff                                                                                                                                                                                                                                                           
比比·拉                                                                                                                                                                                                                                                            

(1 rows affected)
stuff                                                                                                                                                                                                                                                           
比尔比拉                                                                                                                                                                                                                                       
```

```
1> DECLARE @string1 VARCHAR(30) = '比尔·拉', @string2 NVARCHAR(30) = N'比'
2> SELECT STUFF(@string1, 1, 1, @string2)
3> SELECT STUFF(@string1, 2, 1, @string2)
4> SELECT STUFF(@string1, 3, 1, @string2)
5> go
stuff 
??·? 

(1 rows affected)
stuff 
??·? 

(1 rows affected)
stuff 
???? 

(1 rows affected) 
```
```
# current output (correct output) (NOTE: there is still difference from TSQL but that is because the way we store multi-byte characters in VARCHAR, issues related to STUFF() function are fixed)
1> DECLARE @string1 VARCHAR(30) = '比尔·拉', @string2 NVARCHAR(30) = N'比'
2> SELECT STUFF(@string1, 1, 1, @string2)
3> SELECT STUFF(@string1, 2, 1, @string2)
4> SELECT STUFF(@string1, 3, 1, @string2)
5> go
stuff                                                                                                                                                                                                                                                           
比尔·拉                                                                                                                                                                                                                                                            

(1 rows affected)
stuff                                                                                                                                                                                                                                                           
比比·拉                                                                                                                                                                                                                                                            

(1 rows affected)
stuff                                                                                                                                                                                                                                                           
比尔比拉                                                                                                                                                                                                                                                            

(1 rows affected)
```
- Added Test Cases for these issues

### Disabled special handling for DUMP RESTORE
1. Upgrade from Previous version to Current version
-   In this case, Dump would take place according to previous definitions, and during restore PG handling will first try to match argument datatypes, but in this case since one of the earlier definitions is defined with ANYELEMENT, so it will not be able to match the exact definitions and will hit our handling, but we don't need that rather PG handling will take care.

2. Upgrade from Current version to future versions
-   In this case, DDLs will be dumped according to current definitions and during restore PG handling will take early exit as it would be able to match with exact definitions, so anyway it will not hit our additional handling

Task: BABEL- 4838



* **Use case based -**   stuff-*


* **Boundary conditions -**


* **Arbitrary inputs -**


* **Negative test cases -**


* **Performance tests -**


* **Tooling impact -**

* **Memory tests -**


* **Client tests -**



### Check List


contrib/babelfishpg_tsql/src/pltsql_coerce.c
```diff
@@ -1307,7 +1332,15 @@ tsql_func_select_candidate(List *names,
 	bool			  candidates_are_opers = false;
 
 	if (is_special)
+	{
+		/*
+		 * In case of dump restore we don't require special handling as PG handling will be sufficient
```
shah-nirmit, line 1337: Why are we adding this change now and not with previous ones?

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--3.6.0--3.7.0.sql
```diff
@@ -10321,6 +10321,80 @@ $$
 STRICT
 LANGUAGE plpgsql;
 
+DO $$
+DECLARE
+    exception_message text;
+BEGIN
+    ALTER FUNCTION sys.stuff(TEXT, INTEGER, INTEGER, TEXT) RENAME TO stuff_deprecated_3_7;
+    ALTER FUNCTION sys.stuff(ANYELEMENT, INTEGER, INTEGER, ANYELEMENT) RENAME TO stuff_any_deprecated_3_7;
```
shah-nirmit: Why do we need the whole thing in do block?

shah-nirmit (approved):

contrib/babelfishpg_tsql/src/pltsql_coerce.c
Anikait143: ### 1. Upgrade from Previous version to Current version
-   In this case, Dump would take place according to previous definitions, and during restore PG handling will first try to match argument datatypes, but in this case since one of the earlier definitions is defined with ANYELEMENT, so it will not be able to match the exact definitions and will hit our handling, but we don't need that rather PG handling will take care.

### 2. Upgrade from Current version to future versions
-   In this case, DDLs will be dumped according to current definitions and during restore PG handling will take early exit as it would be able to match with exact definitions, so anyway it will not hit our additional handling

Also updated the description

contrib/babelfishpg_tsql/sql/sys_functions.sql
```diff
@@ -1424,36 +1424,63 @@ $BODY$
...
-	overlay (expr placing replace_expr from start for length)
+BEGIN
+    IF start IS NULL OR expr IS NULL OR length IS NULL THEN
+        RETURN NULL;
+    END IF;
+    IF start <= 0 OR start > sys.len(expr) OR length < 0 THEN
+        RETURN NULL;
+    END IF;
+    IF replace_expr IS NULL THEN
+        RETURN (SELECT (overlay (expr::sys.VARCHAR placing '' from start for length))::sys.VARCHAR)::VARBINARY;
```
rishabhtanwar29: Let's be consistent with the use of `VARBINARY` and `sys.VARBINARY`. We should should use only one i.e. `sys.VARBINARY`.

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--3.6.0--3.7.0.sql
Anikait143: 1. Suppose the Rename fails for some reason it would indicate that the function doesn't exist, which implies that the older definitions are dropped and new definitions are there, So, in that case we will not require to create the new functions.
2. If the rename passes only then we will have to create the new definitions.
3. Moreover the babelfish_drop_deprecated_object function does not handle the case when the function doesnot exist so we need that in do warning block

rishabhtanwar29: We don't need this in 3.x

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--3.6.0--3.7.0.sql
```diff
@@ -10367,6 +10367,80 @@ $$
...
+        END IF;
+        RETURN (SELECT overlay (expr placing replace_expr from start for length));
+    END;
+    $BODY$
+    LANGUAGE plpgsql IMMUTABLE PARALLEL SAFE;
+
+    CALL sys.babelfish_drop_deprecated_object('function', 'sys', 'stuff_deprecated_3_7');
+    CALL sys.babelfish_drop_deprecated_object('function', 'sys', 'stuff_any_deprecated_3_7');
+
+EXCEPTION WHEN OTHERS THEN
```
rishabhtanwar29: We should catch specific error code instead of all.

contrib/babelfishpg_tsql/sql/sys_functions.sql
Anikait143: Done

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--3.6.0--3.7.0.sql
Anikait143: Removed `DO` block

rishabhtanwar29 (approved):

Deepesh125 (approved):

PR #2830, opened 2024-08-04
Update views sys.triggers, sys.objects, sys.all_objects to show the correct object_id of trigger

shalinilohia50:
### Description

1. Show object_id of `pg_trigger` instead of `pg_proc` in the views sys.triggers, sys.events, sys.objects, sys.all_objects etc. for triggers. It removes inconsistency of object_id  between OBJECT_ID() and SYS views.
2. If the object type is specified as 'p' and it's actually a trigger, then object_id() should return NULL.

### Issues Resolved

Task: BABEL-5108, BABEL-3927

NOTE: Updated DDL export tests output where ANSI_NULLS and QUOTED_IDENTIFIER are set to OFF in DDL scripting of triggers. It doesn't affect the scripting, we can take this up later.

* **Use case based -** Tested locally
```

select OBJECT_ID([name]), object_id from sys.objects where [name] = 't_trigger'
go
object_id   object_id  
----------- -----------
      19747       19747

(1 rows affected)
select OBJECT_ID([name]), object_id from sys.triggers where [name] = 't_trigger'
go
object_id   object_id  
----------- -----------
      19747       19747

(1 rows affected)
select OBJECT_ID([name]), object_id from sys.all_objects where [name] = 't_trigger'
go
object_id   object_id  
----------- -----------
      19747       19747

(1 rows affected)

select OBJECT_ID('t_trigger', 'p');
go
object_id  
-----------
      NULL

(1 rows affected)
select OBJECT_ID('t_trigger', 'fn');
go
object_id  
-----------
      NULL

(1 rows affected)
select OBJECT_ID('t_trigger', 'tr');
go
object_id  
-----------
      19747

(1 rows affected)
select OBJECT_ID('t_trigger');
go
object_id  
-----------
      19747

(1 rows affected)
```

* **Boundary conditions -**

* **Arbitrary inputs -**

* **Negative test cases -**

* **Performance tests -**

* **Tooling impact -**

* **Client tests -**


contrib/babelfishpg_tsql/sql/sys_views.sql
```diff
@@ -1518,7 +1518,9 @@ union all
 -- details of user defined procedures
 select
     p.proname::sys.sysname as name 
-  , p.oid as object_id
+  , case
```
thephantomthief, line 1523: So OBJECT_ID() for a trigger will return NULL but in sys.all_objects it can show a non-NULL object id?

shalinilohia50: object_id() for trigger will return NULL only if we pass the object_type as 'p' which means a procedure. Even object_id() returns non-NULL value if the object type is specified as 'tr'. Added those manual tests in the PR description.
`sys.all_objects` should show a non-NULL object id of trigger if it exists.

thephantomthief (approved):

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -2128,6 +2128,17 @@ object_id(PG_FUNCTION_ARGS)
 					 !strcmp(object_type, "pc") || !strcmp(object_type, "tf") || !strcmp(object_type, "rf") ||
 					 !strcmp(object_type, "x"))
 			{
+				/*
+				 * If the object type is specified as 'p' and it's actually a trigger,
+				 * then object_id() should return NULL.
+				 */
+				if (!strcmp(object_type, "p") && tsql_get_trigger_oid(object_name, schema_oid, user_id))
```
tanscorpio7: I think the first condition ```!strcmp(object_type, "p")``` is not required. Can you please check it.

tanscorpio7 (changes requested):

contrib/babelfishpg_tsql/runtime/functions.c
shalinilohia50: If we don't add this condition `!strcmp(object_type, "p")`, then  OID look up for trigger will take place for other object types as well by `tsql_get_trigger_oid`. Just saving that extra look up for any object type other than procedure.

contrib/babelfishpg_tsql/sql/sys_views.sql
```diff
@@ -1518,7 +1518,9 @@ union all
...
+      when t.typname = 'trigger' then tr.oid else p.oid
+    end as object_id
```
tanscorpio7, line 1525: do we need a similar change for the previous subquery as well ?

contrib/babelfishpg_tsql/runtime/functions.c
tanscorpio7: ```select object_id('trig','fn')``` -- this returns null in SQL server even though trigger exists.
I think the check should be done for all the object types.

contrib/babelfishpg_tsql/sql/sys_views.sql
shalinilohia50: The previous subquery is for system defined triggers/procedures, but there is no system defined trigger in TSQL. IMO we can skip that.

contrib/babelfishpg_tsql/runtime/functions.c
shalinilohia50: Updated.

tanscorpio7 (approved):

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -2128,6 +2128,17 @@ object_id(PG_FUNCTION_ARGS)
 					 !strcmp(object_type, "pc") || !strcmp(object_type, "tf") || !strcmp(object_type, "rf") ||
 					 !strcmp(object_type, "x"))
 			{
+				/*
+				 * If the object type is not specified as 'tr' and it's actually a trigger,
+				 * then object_id() should return NULL.
+				 */
+				if (tsql_get_trigger_oid(object_name, schema_oid, user_id))
```
rishabhtanwar29: Let's use OidIsValid()

test/JDBC/src/main/resources/config.txt
```diff
@@ -49,4 +49,4 @@ testFileRoot = ./
 isUpgradeTestMode = false
 
 # Default SLA in milliseconds
-defaultSLA = 40000
+defaultSLA = 60000
```
rishabhtanwar29: Do we really need to increase baseline SLA for all the tests?

test/python/expected/pyodbc/ddl_all_objects.out
```diff
@@ -151,8 +151,8 @@ ALTER TABLE [dbo].[test_tsql_const] ADD  CONSTRAINT [test_tsql_const_pkey] PRIMA
 )WITH (PAD_INDEX = OFF, STATISTICS_NORECOMPUTE = OFF, SORT_IN_TEMPDB = OFF, IGNORE_DUP_KEY = OFF, ONLINE = OFF, ALLOW_ROW_LOCKS = ON, ALLOW_PAGE_LOCKS = ON)
 GO
 
-SET ANSI_NULLS ON
-SET QUOTED_IDENTIFIER ON
+SET ANSI_NULLS OFF
+SET QUOTED_IDENTIFIER OFF
```
HarshLunagariya, line 155: Why are we changing this?

test/JDBC/src/main/resources/config.txt
HarshLunagariya: Do we have an agreement(with dev-team) to increase it?

contrib/babelfishpg_tsql/sql/sys_views.sql
```diff
@@ -2007,7 +2011,27 @@ LEFT OUTER JOIN sys.babelfish_view_def bvd
...
+  , CAST(0 as sys.bit)  AS is_schema_bound
+  , CAST(0 as sys.bit)  AS uses_database_collation
+  , CAST(0 as sys.bit)  AS is_recompiled
+  , CAST(0 AS sys.bit) as null_on_null_input
+  , null::integer as execute_as_principal_id
+  , CAST(0 as sys.bit) as uses_native_compilation
+  , CAST(ao.is_ms_shipped as INT) as is_ms_shipped
+FROM sys.all_objects ao
+LEFT OUTER JOIN sys.pg_namespace_ext nmext on ao.schema_id = nmext.oid
+LEFT OUTER JOIN sys.babelfish_namespace_ext ext ON nmext.nspname = ext.nspname
```
HarshLunagariya: I think we don't require 'babelfish_namespace_ext' join here.

test/python/expected/pyodbc/ddl_all_objects.out
shalinilohia50: As mentioned in the description, Updated DDL export tests output where ANSI_NULLS and QUOTED_IDENTIFIER are set to OFF in DDL scripting of triggers. It doesn't affect the scripting, we can take this up later.

shardgupta (approved):

test/JDBC/src/main/resources/config.txt
shalinilohia50: No, this was a temp change. 
Increased the SLA for a particular test file.

shalinilohia50: No, this was a temp change. 
Increased the SLA for a particular test file.

PR #2849, opened 2024-08-13
fix infinite loop in replace with AI collations

tanscorpio7:
### Description

ICU usearch_next() goes into infinite loop when pattern to search starts with a surrogate pair.
To get around this we check if output of usearch_next() is stuck and not proceeding forwards 
and set the offset for next search ourselves.
The next offset is simply the next character after the current char in source string.

```
SRC STRING - 'abc🙂defghi🙂🙂'    PATTERN TO FIND = '🙂def'

usearch_next() gets stuck on "🙂" idx = 3 and repeatedly returns this index.
We will intervene and set the offset to "d" idx = 4. 
So that usearch_next only starts looking from this character.
```
### Issues Resolved

[BABEL-5169]

### Sign Off

### Check List

### DIFF
--- a/contrib/babelfishpg_tsql/src/collation.c
+++ b/contrib/babelfishpg_tsql/src/collation.c
@@ -52,7 +52,7 @@
 #define UCHAR_IS_SURROGATE(c) ((c & 0xF800) == 0xD800)
 
 /* Find length of given Uchar */
-#define UCHAR_LENGTH(c) ( UCHAR_IS_SURROGATE(c) ? 2 : 1)
+#define UCHAR_LENGTH(c) (UCHAR_IS_SURROGATE(c) ? 2 : 1)
 
 Oid			server_collation_oid = InvalidOid;
 collation_callbacks *collation_callbacks_ptr = NULL;
--- a/contrib/babelfishpg_tsql/src/collation.c
+++ b/contrib/babelfishpg_tsql/src/collation.c
@@ -1475,12 +1475,13 @@ pltsql_strpos_non_determinstic(text *src_text, text *substr_text, Oid collid, in
 		int32_t src_len_utf8 = VARSIZE_ANY_EXHDR(src_text);
 		int32_t substr_len_utf8 = VARSIZE_ANY_EXHDR(substr_text);
 		int32_t src_ulen, substr_ulen;
-		int32_t u8_pos = -1;
+		int32_t u8_pos = -1, pos_prev_loop = -1;
 		UErrorCode	status = U_ZERO_ERROR;
 		UStringSearch *usearch;
 		UChar *src_uchar, *substr_uchar;
 		coll_info_t coll_info_of_inputcollid = tsql_lookup_collation_table_internal(collid);
 		bool is_CS_AI = false;
+		bool is_substr_starts_with_surrogate;
 
 		if (OidIsValid(coll_info_of_inputcollid.oid) &&
 		    coll_info_of_inputcollid.collateflags == 0x000e /* CS_AI  */ )
@@ -1490,6 +1491,8 @@ pltsql_strpos_non_determinstic(text *src_text, text *substr_text, Oid collid, in
 
 		src_ulen = icu_to_uchar(&src_uchar, VARDATA_ANY(src_text), src_len_utf8);
 		substr_ulen = icu_to_uchar(&substr_uchar, VARDATA_ANY(substr_text), substr_len_utf8);
+
+		is_substr_starts_with_surrogate = U16_IS_SURROGATE(substr_uchar[0]);
 
 		usearch = usearch_openFromCollator(substr_uchar,
 										substr_ulen,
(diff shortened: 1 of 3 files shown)


contrib/babelfishpg_tsql/src/collation.c
```diff
@@ -1511,12 +1511,28 @@ pltsql_strpos_non_determinstic(text *src_text, text *substr_text, Oid collid, in
...
 			if (U_FAILURE(status))
 				ereport(ERROR,
 					(errcode(ERRCODE_INTERNAL_ERROR),
 					 errmsg("failed to perform ICU search: %s",
 							u_errorName(status))));
 
+			/* ICU bug, When pattern start with a surrogate pair ICU usearch_next stops moving forward entering an infinite loop */
+			if (u16_pos == pos_prev_loop)
+			{
+				if (U16_IS_SURROGATE(src_uchar[matched_idx]) && U16_IS_SURROGATE(substr_uchar[0]) && matched_idx + 2 < src_ulen)
```
ahmed-shameem: Does the problem arises only when the first char is surrogate pair? I think better to use U16_IS_SURROGATE(substr_uchar[u16_pos]).

contrib/babelfishpg_tsql/src/collation.c
```diff
@@ -1602,13 +1618,28 @@ pltsql_replace_non_determinstic(text *src_text, text *from_text, text *to_text,
...
 			if (U_FAILURE(status))
 				ereport(ERROR,
 					(errcode(ERRCODE_INTERNAL_ERROR),
 					 errmsg("failed to perform ICU search: %s",
 							u_errorName(status))));
 
+			/* ICU bug, When pattern start with a surrogate pair ICU usearch_next stops moving forward entering an infinite loop */
+			if (pos == pos_prev_loop)
+			{
+				if ( U16_IS_SURROGATE(src_uchar[matched_idx]) && U16_IS_SURROGATE(from_uchar[0]) && matched_idx + 2 < src_ulen)
```
ahmed-shameem: Same as above

contrib/babelfishpg_tsql/src/collation.c
tanscorpio7: done

contrib/babelfishpg_tsql/src/collation.c
tanscorpio7: done

contrib/babelfishpg_tsql/src/collation.c
tanscorpio7: Yes, it only happens when substring to find start with a surrogate pair.
In this case ICU doesn't seem to set the offset for the next search correctly.

contrib/babelfishpg_tsql/src/collation.c
```diff
@@ -1491,6 +1492,8 @@ pltsql_strpos_non_determinstic(text *src_text, text *substr_text, Oid collid, in
 		src_ulen = icu_to_uchar(&src_uchar, VARDATA_ANY(src_text), src_len_utf8);
 		substr_ulen = icu_to_uchar(&substr_uchar, VARDATA_ANY(substr_text), substr_len_utf8);
 
+		is_substr_starts_with_surrogate = U16_IS_SURROGATE(substr_uchar[0]);
```
Deepesh125: What if search str is empty?

contrib/babelfishpg_tsql/src/collation.c
```diff
@@ -1517,6 +1520,23 @@ pltsql_strpos_non_determinstic(text *src_text, text *substr_text, Oid collid, in
...
 
+			/* ICU bug, When pattern start with a surrogate pair ICU usearch_next stops moving forward entering an infinite loop */
+			if (u16_pos == pos_prev_loop)
+			{
+				/* IF UTF16 code point is in the range D800 - DBFF, then it is a surrogate pair */
+				int32_t next_char_idx = u16_pos + (src_uchar[u16_pos] & 0xD800 ? 2 : 1);
+
+				if (is_substr_starts_with_surrogate && next_char_idx < src_ulen)
+				{
+					usearch_setOffset(usearch, next_char_idx, &status);
```
Deepesh125, line 1539: Do we need error checking?

Deepesh125 (changes requested):

contrib/babelfishpg_tsql/src/collation.c
tanscorpio7: empty string never reaches pltsql_strpos_non_determinstic.
Even if it did we will error out before we reach here. Confirmed by setting values using GDB.

tanscorpio7: We also have test cases for null inputs/empty string inputs in charindex, replace and patindex.

contrib/babelfishpg_tsql/src/collation.c
tanscorpio7: Done

Deepesh125 (approved):

PR #2857, opened 2024-08-14
Run JDBC test framework in multi-db by default and add separate workflow for single-db mode

sumitj824:
### Description
This commit makes the following changes:
- Modifies the JDBC test workflow to run in multi-db migration mode by default
- Adds a new workflow to run the JDBC tests in single-db migration mode
- Introduces a new flag `single_db_mode_expected` to indicate different expected output for single-db migration mode
- Fixes test files having different output in multi-db and single-db migration mode by either adding separate expected output files with the `single_db_mode_expected` flag or making changes in test files itself to have same output.

### Issues Resolved

Task: BABEL-5081

### Check List


rishabhtanwar29 (approved):

test/JDBC/src/test/java/com/sqlsamples/TestQueryFile.java
```diff
@@ -424,6 +425,7 @@ public void TestQueryBatch(String inputFileName) throws SQLException, ClassNotFo
         BufferedWriter bw = new BufferedWriter(fw);
         curr_exec_time = 0L;
         checkParallelQueryExpected = false;
+        checkSingleDbModeExpected = false;
```
ahmed-shameem, line 427: I think rather than having this flag in test file, better to check whether there is any expected file for single_db mode or not just like we did for non-default server collation.
Because we want to run single-db mode upgrade tests as well, if we follow this approach..i think it will be more difficult to do that

.github/workflows/jdbc-tests-single-db-mode.yml
```diff
@@ -0,0 +1,177 @@
...
+          name: jdbc-output-diff.diff
+          path: |
+            test/JDBC/Info/output-diff.diff
+            test/JDBC/Info/replication-output-diff.diff
+
+      - name: Check and upload coredumps
+        if: always() && (steps.jdbc.outcome == 'failure' || steps.replication.outcome == 'failure')
+        uses: ./.github/composite-actions/upload-coredump
+
+      - name: Generate Code Coverage
```
ahmed-shameem: If we want to run code-coverage for single-db, then we should update the pr-code-coverage which will invoke this workflow on [workflow_call]. Or else we should remove these.

ahmed-shameem (changes requested):

test/JDBC/src/test/java/com/sqlsamples/TestQueryFile.java
sumitj824: Can you explain how it will be difficult to run upgrade in single-db mode with this approach? I think searching of file only if required (i.e. flag is specified) is better approach than always searching for file in folder.

.github/workflows/jdbc-tests-single-db-mode.yml
sumitj824: removed it.

test/JDBC/src/test/java/com/sqlsamples/TestQueryFile.java
```diff
@@ -439,12 +441,14 @@ public void TestQueryBatch(String inputFileName) throws SQLException, ClassNotFo
...
-            nonDefaultServerCollationExpectedFile = new File(parallelQueryGeneratedFilesDirectoryPath + "non_default_server_collation/" + serverCollationName + "/" + outputFileName + ".out");
+        }
+        else if (isSingleDbMode && checkSingleDbModeExpected){
+            expectedFile = new File(generatedFilesDirectoryPath + "single_db/" + outputFileName + ".out");
         }
         else{
             expectedFile = new File(generatedFilesDirectoryPath + outputFileName + ".out");
-            nonDefaultServerCollationExpectedFile = new File(generatedFilesDirectoryPath + "non_default_server_collation/" + serverCollationName + "/" + outputFileName + ".out");
         }
+        nonDefaultServerCollationExpectedFile = new File(generatedFilesDirectoryPath + "non_default_server_collation/" + serverCollationName + "/" + outputFileName + ".out");
```
rohit01010: why are we making this change related to `nonDefaultServerCollationExpectedFile` ?

test/JDBC/src/main/java/com/sqlsamples/batch_run.java
```diff
@@ -10,6 +10,7 @@
 import static com.sqlsamples.Statistics.curr_exec_time;
 import static com.sqlsamples.Statistics.sla;
 import static com.sqlsamples.Config.checkParallelQueryExpected;
+import static com.sqlsamples.Config.checkSingleDbModeExpected;
```
rohit01010: NIT: we are already importing `com.sqlsamples.Config.*` on line number 8. No need of importing `checkSingleDbModeExpected` again.

test/JDBC/src/test/java/com/sqlsamples/TestQueryFile.java
```diff
@@ -20,6 +20,7 @@
 import static com.sqlsamples.Statistics.curr_exec_time;
 import static com.sqlsamples.Statistics.sla;
 import static com.sqlsamples.Config.checkParallelQueryExpected;
+import static com.sqlsamples.Config.checkSingleDbModeExpected;
```
rohit01010: NIT: we are already importing `com.sqlsamples.Config.*` on line number 16. No need of importing `checkSingleDbModeExpected` again.

.github/composite-actions/install-extensions/action.yml
```diff
@@ -8,7 +8,7 @@ inputs:
   migration_mode:
     description: 'Database migration mode'
     required: no
-    default: "single-db"
+    default: "multi-db"
```
rohit01010, line 11: In other workflows we use `install-extensions` with default value of `migration_mode`, changing the `migration_mode` to `multi-db` will cause all those workflows to run with `migration_mode` as `multi-db` and not just JDBC test workflow. Is this expected ?

.github/workflows/jdbc-tests-single-db-mode.yml
```diff
@@ -0,0 +1,143 @@
...
+
+      - name: Start secondary server
+        id: start-secondary
+        if: always() && steps.jdbc.outcome == 'success'
+        uses: ./.github/composite-actions/install-extensions
+        with:
+          psql_port: 5433
+          tsql_port: 8199
+          wal_level: logical
+
```
rohit01010: Since we had set the default value of `migration_mode` as `multi-db` in composite action `install-extensions`, here the secondary server will be installed with `multi-db` migration mode, whereas the primary server is with migration mode `single-db`. I believe here we need to pass the migration mode as `single-db`

.github/workflows/jdbc-tests-single-db-mode.yml
```diff
@@ -0,0 +1,143 @@
...
+          ~/${{env.INSTALL_DIR}}/bin/psql -v ON_ERROR_STOP=1 -d jdbc_testdb -U runner -p 5433 -c "CREATE SUBSCRIPTION my_sub CONNECTION 'host=localhost port=5432 user=jdbc_user dbname=jdbc_testdb password=12345678' PUBLICATION my_pub;"
+
+      - name: Run Replication Tests
+        id: replication
+        if: always() && steps.setup-pub-sub.outcome == 'success'
+        timeout-minutes: 60
+        uses: ./.github/composite-actions/run-jdbc-tests
+        with:
+          input_dir: 'replication'
+
```
rohit01010: Here as well we need to pass `migration_mode` as `single-db`.

rohit01010 (changes requested):

.github/composite-actions/install-extensions/action.yml
sumitj824: yes this is expected, we want to run all the workflows in "multi-db" mode.

test/JDBC/src/test/java/com/sqlsamples/TestQueryFile.java
sumitj824: It was needed because we are always looking for presence of expected file in non_default_server_collation folder.

rohit01010 (approved):

.github/composite-actions/install-extensions/action.yml
rohit01010: Kindly update this in the PR description.

test/JDBC/src/main/resources/config.txt
```diff
@@ -39,6 +39,9 @@ scheduleFile = ./jdbc_schedule
 # WHETHER TEST MODE IS PARALLEL QUERY MODE ON
 isParallelQueryMode = false
 
+# WHETHER DATABASE MIGRATION MODE IS SINGLE-DB
```
tanscorpio7, line 42: Should we also maintain a new schedule file for single-db-mode which would be appended over the existing schedule.

test/JDBC/src/test/java/com/sqlsamples/TestQueryFile.java
```diff
@@ -441,11 +442,16 @@ public void TestQueryBatch(String inputFileName) throws SQLException, ClassNotFo
             expectedFile = new File(parallelQueryGeneratedFilesDirectoryPath + outputFileName + ".out");
             nonDefaultServerCollationExpectedFile = new File(parallelQueryGeneratedFilesDirectoryPath + "non_default_server_collation/" + serverCollationName + "/" + outputFileName + ".out");
         }
+        else if (isSingleDbMode && checkSingleDbModeExpected){
+            expectedFile = new File(generatedFilesDirectoryPath + "single_db/" + outputFileName + ".out");
+            nonDefaultServerCollationExpectedFile = new File(generatedFilesDirectoryPath + "single_db/" + "non_default_server_collation/" + serverCollationName + "/" + outputFileName + ".out");
+        }
```
tanscorpio7, line 448: What happens when input file has both `-- parallel_query_expected` & `-- single_db_mode_expected`

tanscorpio7: We will not run single-db mode in parallel query ?

sumitj824: No, there is no plan to run single-db mode in parallel query.

test/JDBC/src/main/resources/config.txt
sumitj824: There is no need to maintain separate schedule file for single-db mode as there are not many test files which are supposed to be run in only single-db mode.

jsudrik (approved):
Approved.

PR #2858, opened 2024-08-14
Add 'user connections' and 'network packet size (B)' configuration to sys.configurations view

sharathbp:
### Description

Add 'user connections' and 'network packet size (B)' configuration to sys.configurations view

1. At the moment babelfish doesn't show "user connections" and "network packet size (B)" in sys.configurations. With this change we show the above configurations.
2. This change is required for supporting RDS Proxy for babelfish. RDS Proxy fetches user connection and network packet size properties from sys.configurations for connection pooling.

### Issues Resolved

BABEL-4825

* **Use case based -** Yes

* **Arbitrary inputs -** Yes

* **Performance tests -** N/A

* **Tooling impact -** N/A

* **Client tests -** N/A

### Check List


thephantomthief (changes requested):
Please add upgrade script changes and tests

contrib/babelfishpg_tsql/sql/sys_views.sql
```diff
@@ -2460,7 +2460,31 @@ SELECT  configuration_id,
...
-FROM sys.babelfish_configurations;
+FROM sys.babelfish_configurations
+UNION ALL
+SELECT 
+CAST(103 as INT) AS configuration_id,
+CAST('user connections' AS SYS.NVARCHAR(35)) AS name,
+CAST(s.setting AS sys.sql_variant) AS value,
+CAST(s.min_val AS sys.sql_variant) AS minimum,
+CAST(s.max_val AS sys.sql_variant) AS maximum,
+CAST(s.setting AS sys.sql_variant) AS value_in_use,
```
sharathbp: Actually this is incorrect, s.setting in pg_catalog stores current value and doesnt provide previous value. Need to check how to get the previous value.

thephantomthief: Why do we need previous value?

sharathbp: Because for 'user connections' the config value is not dynamic, it gets applied when the engine is restarted. So value_in_use will be previous value until the engine is restarted and that value will be used by Proxy.

In pg_catalog we have is_restart value to determine if the value is applied or not.

thephantomthief: Makes sense

kuntalghosh: Instead of adding multiple union all like this, I'll prefer to use as following:
```
SELECT * from babelfish_configurations
union all
SELECT get_config_id(name),
CAST('user connections' AS SYS.NVARCHAR(35)) AS name,
CAST(s.setting AS sys.sql_variant) AS value,
CAST(s.min_val AS sys.sql_variant) AS minimum,
CAST(s.max_val AS sys.sql_variant) AS maximum,
CAST(s.setting AS sys.sql_variant) AS value_in_use,
CAST(s.short_desc AS sys.nvarchar(255)) AS description,
CAST(0 AS sys.BIT) AS is_dynamic,
CAST(1 AS sys.BIT) AS is_advanced
FROM pg_catalog.pg_settings s where name in
('sdsd','sd',...);
```
Also, I think you need to mention collate 'C' while doing the string comparison. Please check with Dipesh.

kuntalghosh (dismissed):

contrib/babelfishpg_tsql/sql/sys_views.sql
sharathbp: But how to set the configuration_id to 103 for max_connection and 505 for packet size and also the names? we can use if condition but that will have multiple if conditions.

tanscorpio7:
I think database_metadata.txt test case needs some update in the expected file. JDBC framework is breaking because it cannot handle the output diff for some reason.

test/JDBC/expected/database_metadata.out
```diff
@@ -149,7 +149,7 @@ master#!#dbo#!#dbmeta_tab2#!#<NULL>#!#<NULL>#!#<NULL>#!#0#!#<NULL>#!#<NULL>#!#<N
 
 dbmeta#!#getMaxConnections
 ~~START~~
-0
+262143
```
Deepesh125, line 152: This could become flaky test because in APG we calculate max_connection using following formula - `LEAST({DBInstanceClassMemory/9531392},5000).`

contrib/babelfishpg_tsql/sql/sys_views.sql
```diff
@@ -2452,43 +2452,68 @@ create or replace view sys.dm_exec_connections
...
+        name,
+        value,
+        minimum,
+        maximum,
+        value_in_use,
+        description,
         is_dynamic, 
         is_advanced 
-FROM sys.babelfish_configurations;
+FROM sys.babelfish_configurations
```
Deepesh125, line 2463: we should use union with `pg_catalog.pg_settings in ('max_connection', 'babelfishpg_tds.tds_default_packet_size')` in predicate

test/JDBC/expected/sys-configurations.out
```diff
@@ -2,13 +2,34 @@ SELECT * FROM sys.configurations;
 GO
```
Deepesh125, line 2: since sys.configuations uses UNINON internally, we really need to add order by clause here in the query to avoid flakiness

Deepesh125: Applicable at all the places where we are testing sys.configuations

Deepesh125 (changes requested):

test/JDBC/expected/sys-configurations-vu-verify.out
```diff
@@ -5,35 +5,39 @@ SELECT * FROM sys_configurations_vu_prepare_v1
 GO
 ~~START~~
 int#!#nvarchar#!#sql_variant#!#sql_variant#!#sql_variant#!#sql_variant#!#nvarchar#!#bit#!#bit
-16387#!#SMO and DMO XPs#!#1#!#0#!#1#!#1#!#Enable or disable SMO and DMO XPs#!#1#!#1
-1534#!#user options#!#0#!#0#!#32767#!#0#!#user options#!#1#!#0
+103#!#user connections#!#100#!#1#!#262143#!#100#!#Sets the maximum number of concurrent connections.#!#0#!#1
 115#!#nested triggers#!#1#!#0#!#1#!#1#!#Allow triggers to be invoked within triggers#!#1#!#0
 124#!#default language#!#0#!#0#!#9999#!#0#!#default language#!#1#!#0
+505#!#network packet size (B)#!#4096#!#512#!#32767#!#4096#!#Sets the default packet size forall the clients being connected#!#1#!#1
```
thephantomthief: typo: `forall` -> `for all`

test/JDBC/expected/sys-configurations.out
```diff
@@ -220,6 +261,8 @@ sql_variant#!#int#!#nvarchar#!#smallint
 1033#!#1126#!#default full-text language#!#3
 2049#!#1127#!#two digit year cutoff#!#3
 0#!#1555#!#Transform noise words for full-text query#!#3
+4096#!#505#!#Sets the default packet size forall the clients being connected#!#3
+100#!#103#!#Sets the maximum number of concurrent connections.#!#2
```
Deepesh125: this can be still flaky

Deepesh125 (changes requested):

Deepesh125 (approved):

thephantomthief (approved):

PR #2860, opened 2024-08-14
jTDS: support logins without PRELOGIN message

### LINKED ISSUE #2137: Support JTDS driver

staticlibs:
### Description

This PR is a part of a change originally implemented in #2320.

This change modifies the TDS login allowing to receive `LOGIN7` as the first client message instead of `PRELOGIN`. It allows to successfully open a connection to Babelfish using jTDS driver without SSL. When the first message is `PRELOGIN` - then login logic is unchanged.

Copying review comments from #2320:

> 1. Let's add a comment why we may not expecting TDS_PRELOGIN as the first request.

Added clarifying comment.

>     2. The check should be
> 
> 
> ```
> if (TdsReadNextBuffer() != EOF && TdsCheckMessageType(TDS_PRELOGIN))
> ```

Modified the check.

>     3. If the prelogin message is not sent, what's the SSL behavior - TDS_ENCRYPT_NOT_SUP or TDS_ENCRYPT_OFF? OFF means the login will still happen in an encrypted manner.

If the first message is `LOGIN7` then SSL behaviour is `TDS_ENCRYPT_OFF (0x00)`. Login does not happen in encrypted manner in this case - password is sent by client in plain text in `LOGIN7` message.

>     4. According the code, the loadEncryption defaults to 0 which is TDS_ENCRYPT_OFF. So, we're freeing the ssl structure towards the end of the function without setting up the socket. Also, according to the code, the login is probably not happening in an encrypted manner.

This is correct, `loadEncryption` was `0` by default, added a branch to explicitly set it to `TDS_ENCRYPT_OFF (0x00)`.
 
> Actually, let's not modify the SSL code at all. Something like as following (after confirming the SSL behavior):

SSL behaviour with jTDS is the following (assuming `babelfishpg_tds.tds_ssl_encrypt = 'false'`):

1. `ssl=off` (in jTDS URL, also default): `PRELOGIN` is not sent, `loadEncryption = TDS_ENCRYPT_OFF (0x00)`

2. `ssl=request`: `PRELOGIN` is sent with `TDS_ENCRYPT_OFF (0x00)`, `loadEncryption = TDS_ENCRYPT_OFF (0x00)`

3. `ssl=require` or `ssl=authenticate`: `PRELOGIN` is sent with `TDS_ENCRYPT_ON (0x01)`, `loadEncryption = TDS_ENCRYPT_ON (0x01)`

> we can have more safer check by checking whether client has ssl=off. If that's the case then only we should skip pre-login msg. Throw an FATAL error otherwise and should terminate connection

M, it seems that we don't have such information at the point of the check. First message (either `PRELOGIN` or `LOGIN7`) is always sent without SSL. Client SSL-request value is included in the `PRELOGIN` message itself.

### Issues Resolved

#2137

JDBC test suite changes to support jTDS will follow in a separate PR.

### Check List

### DIFF
--- a/contrib/babelfishpg_tds/src/backend/tds/tdslogin.c
+++ b/contrib/babelfishpg_tds/src/backend/tds/tdslogin.c
@@ -1991,41 +1991,54 @@ TdsProcessLogin(Port *port, bool loadedSsl)
 	TdsErrorContext->phase = 0;
 	TdsErrorContext->reqType = TDS_LOGIN7;
 
-	PG_TRY();
+	/*
+	 * Check whether the first client message is indeed PRELOGIN.
+	 * 7.1 clients may send LOGIN7 instead, for example
+	 * jTDS sends PRELOGIN only when SSL is enabled.
+	 */
+	if (TdsReadNextBuffer() != EOF && TdsCheckMessageType(TDS_PRELOGIN))
 	{
-		TdsErrorContext->err_text = "Parsing PreLogin Request";
-		/* Pre-Login */
-		rc = ParsePreLoginRequest();
+		PG_TRY();
+		{
+			TdsErrorContext->err_text = "Parsing PreLogin Request";
+			/* Pre-Login */
+			rc = ParsePreLoginRequest();
+			if (rc < 0)
+				return rc;
+
+			TdsErrorContext->err_text = "Make PreLogin Response";
+
+			loadEncryption = MakePreLoginResponse(port, loadedSsl);
+			TdsFlush();
+
+			TdsErrorContext->err_text = "Setup SSL Handshake";
+			/* Setup the SSL handshake */
+			if (loadEncryption == TDS_ENCRYPT_ON ||
+				loadEncryption == TDS_ENCRYPT_OFF ||
+				loadEncryption == TDS_ENCRYPT_REQ)
+				rc = SecureOpenServer(port);
+		}
+		PG_CATCH();
+		{
+			PG_RE_THROW();
+		}
+		PG_END_TRY();
+
+		/*
+		 * If SSL handshake failure has occurred then no need to go ahead with
+		 * login, Just return from here.
+		 */
 		if (rc < 0)
 			return rc;
 
-		TdsErrorContext->err_text = "Make PreLogin Response";
-
-		loadEncryption = MakePreLoginResponse(port, loadedSsl);
-		TdsFlush();
-
-		TdsErrorContext->err_text = "Setup SSL Handshake";
-		/* Setup the SSL handshake */
-		if (loadEncryption == TDS_ENCRYPT_ON ||
-			loadEncryption == TDS_ENCRYPT_OFF ||
-			loadEncryption == TDS_ENCRYPT_REQ)
-			rc = SecureOpenServer(port);
+		if (loadEncryption == TDS_ENCRYPT_ON)
+			TDSInstrumentation(INSTR_TDS_LOGIN_END_TO_END_ENCRYPT);
 	}
-	PG_CATCH();
+	else
 	{
-		PG_RE_THROW();
+		TDS_DEBUG(TDS_DEBUG3, "TDS7 PreLogin Message Skipped");
+		loadEncryption = TDS_ENCRYPT_OFF;
 	}
-	PG_END_TRY();
-
-	/*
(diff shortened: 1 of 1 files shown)


contrib/babelfishpg_tds/src/backend/tds/tdslogin.c
```diff
@@ -1989,41 +1989,54 @@ TdsProcessLogin(Port *port, bool loadedSsl)
 	TdsErrorContext->phase = 0;
 	TdsErrorContext->reqType = TDS_LOGIN7;
 
-	PG_TRY();
+	/*
+	 * Check whether the first client message is indeed PRELOGIN.
+	 * 7.1 clients may send LOGIN7 instead, for example
+	 * jTDS sends PRELOGIN only when SSL is enabled.
+	 */
+	if (TdsReadNextBuffer() != EOF && TdsCheckMessageType(TDS_PRELOGIN))
```
Deepesh125, line 1999: I am still concerned with this check. We are only checking message type to be PRELOGIN and proceeding ahead if it isn't assuming that first message would always be LOGIN request in case tds version = 7.1. There is no safety check validating that message is actually LOGIN request if it isn't PRELOGIN. We have [Assert here](https://github.com/babelfish-for-postgresql/babelfish_extensions/blob/4217dbf92ee2c8f61a8840b090dcfd34d9114936/contrib/babelfishpg_tds/src/backend/tds/tdslogin.c#L763) but it may not work if engine is not configured with `--case-assert` flag. 

Can we check and throw an error upfront if `!TdsCheckMessageType(TDS_PRELOGIN) && !TdsCheckMessageType(TDS_LOGIN7)`

Any malicious user should not be able get passed with corrupted message.

kuntalghosh: Please note that the same scenario can tried today as well -  a valid prelogin followed by a corrupted login message.

We actually have the check in FetchLoginRequest, but as an assert.
```
Assert(TdsCheckMessageType(TDS_LOGIN7));
```
We should change this to an error message.

We're also validating the login request with all the length checks. So, it's unlikely to hit any issue even with corrupted packet.

staticlibs: Thanks for the reviews! As I understand, the check for `LOGIN7` was just integrated a few hours ago in #2888. I've rebased the PR branch to the latest to re-run the tests.

sftim:
I see this is failing tests. Would you be willing to rebase it against the appropriate target branch @staticlibs ?

staticlibs:
@sftim 

I've rebased the PR to the latest `BABEL_4_X_DEV` and the tests are passing now (previous failure was unrelated to this change).

PR #2877, opened 2024-08-23
Fixed Cast from sys.varchar to TIME shows unexpected result due to undefined typmod

Yvinayak07:
### 1. Issue:

While casting from varchar to TIME, shows unexpected result due to undefined typmod. 
It was discovered that some test related to TIME fails randomly on arch64 machine and passed on 32 bit machine.
For query `select cast(cast('12:45:37.123' as varchar) as time);`
Output :

- arch32: 12:45:37.1230000
- arch64: 12:45:37.0000000

Expected result : 12:45:37.1230000

Reason: Junk value for arch32 is set to a value greater then 7. So it will read real value till 7, Whereas in arch64 junk value is set to 0. It won't read any value after '.' setting it to zero.
 
time_in function was called from varchar2time function with one argument whereas time_in function expects 3 arguments , 3rd one beign the typmod. As this argument was not passed it was picking junk value. This is fixed by calculating tymod in varchar2time and passing correct value of typmod to time_in function.
Also added test cases for this purpose.

Task: BABEL-5179
Authored-by: yashneet vinayak [yashneet@amazon.com](mailto:yashneet@amazon.com)

### Issues Resolved
Task: BABEL-5179

* **Use case based -** yes

* **Arbitrary inputs -** NA

* **Minor version upgrade tests -** yes

* **Major version upgrade tests -** yes

* **Performance tests -** NA

* **Tooling impact -** NA

* **Client tests -** NA

### Check List


contrib/babelfishpg_common/src/varchar.c
```diff
@@ -888,11 +888,12 @@ Datum
 varchar2time(PG_FUNCTION_ARGS)
 {
 	VarChar    *source = PG_GETARG_VARCHAR_PP(0);
+	int32		typmod = PG_GETARG_INT32(1);
 	char	   *str;
 	TimeADT		time;
 
 	str = varchar2cstring(source);
-	time = DatumGetTimeADT(DirectFunctionCall1(time_in, CStringGetDatum(str)));
+	time = DatumGetTimeADT(DirectFunctionCall3(time_in, CStringGetDatum(str),0,typmod));
```
Anikait143: We can use `invalidOID` macro for oid 0, and also please use proper spacing between the arguments.

contrib/babelfishpg_common/src/varchar.c
```diff
@@ -888,11 +888,12 @@ Datum
 varchar2time(PG_FUNCTION_ARGS)
 {
 	VarChar    *source = PG_GETARG_VARCHAR_PP(0);
+	int32		typmod = PG_GETARG_INT32(1);
```
Anikait143: Instead of directly assigning typmod with the argument, we should first check if there are actually 2 arguments passed, else it will again result in Junk values.
```suggestion
	int32		typmod = -1;
	if (PG_NARGS() > 1)
		typmod = PG_GETARG_INT32(1);
```

Anikait143 (commented):
We should add upgrade tests around the concerned CAST

contrib/babelfishpg_common/src/varchar.c
sumitj824: +1

contrib/babelfishpg_common/sql/upgrades/babelfish_common_helper--4.3.0--4.4.0.sql
```diff
@@ -0,0 +1,79 @@
...
+        raise warning '%', error_msg;
+    when dependent_objects_still_exist then --if 'drop view' statement fails
+        GET STACKED DIAGNOSTICS error_msg = MESSAGE_TEXT;
+        raise warning '%', error_msg;
+end
+$$
+LANGUAGE plpgsql;
+
+-- (sys.VARCHAR AS pg_catalog.TIME)
+DROP CAST IF EXISTS(sys.VARCHAR AS pg_catalog.TIME);
```
sumitj824: `IF EXISTS` is redundant here, defined cast will always exists at this point.

contrib/babelfishpg_common/sql/upgrades/babelfish_common_helper--4.3.0--4.4.0.sql
```diff
@@ -0,0 +1,79 @@
...
+
+DO $$
+DECLARE
+    exception_message text;
+BEGIN
+    ALTER FUNCTION sys.varchar2time(sys.VARCHAR) RENAME TO varchar2time_deprecated_4_4_0;
```
sumitj824: We can directly remove the previous definition, we don't need to mark the function as deprecated because we have already dropped the dependent cast.

contrib/babelfishpg_common/src/varchar.c
sumitj824: Instead of above suggestion, lets directly use `DEFAULT` in SQL definition of function itself.

contrib/babelfishpg_common/sql/upgrades/babelfish_common_helper--4.3.0--4.4.0.sql
```diff
@@ -0,0 +1,79 @@
...
+    exception_message text;
+BEGIN
+    CALL sys.babelfish_drop_deprecated_object('function', 'sys', 'varchar2time_deprecated_4_4_0');
+
+EXCEPTION WHEN OTHERS THEN
+    GET STACKED DIAGNOSTICS
+    exception_message = MESSAGE_TEXT;
+    RAISE WARNING '%', exception_message;
+END;
+$$;
```
sumitj824: No need of this, please check previous comments.

sumitj824 (commented):
Please add testcase and follow the guidelines for good PR title and description.

contrib/babelfishpg_common/sql/upgrades/babelfish_common_helper--4.3.0--4.4.0.sql
tanscorpio7: I think we could still rename it. It would clearly mark the function as deprecated in users existing views or other objects.

tanscorpio7: Apparently the view definition does not replace the cast with cast function when it is stored.
Still just to be safe we could rename these sql functions.

contrib/babelfishpg_common/src/varchar.c
Anikait143: I think the above suggestion is backward compatible as it ensures that dumps with the older version (having only one argument) won't break after an upgrade(and avoid junk values during upgrades). [reference](https://github.com/babelfish-for-postgresql/babelfish_extensions/blob/f33f75f7d07327660d39492745a384db95893993/contrib/babelfishpg_common/src/varbinary.c#L784)

contrib/babelfishpg_common/sql/upgrades/babelfish_common_helper--4.3.0--4.4.0.sql
sumitj824: >Apparently the view definition does not replace the cast with cast function when it is stored.

yes, that is why we can directly drop and re-create it and we don't need to mark it as deprecated.

> Still just to be safe we could rename these sql functions.

We can have sufficient dependency upgrade test to validate that.

contrib/babelfishpg_common/sql/upgrades/babelfish_common_helper--4.3.0--4.4.0.sql
shah-nirmit: +1

contrib/babelfishpg_common/src/varchar.c
sumitj824: nack, just adding `DEFAULT` in SQL definition won't break any upgrades. Also, Cast only gets resolved during runtime so there will not be any issue with  dumps of older version. And the reference is irrelevant here.

@Yvinayak07 Please add upgrade  dependency test to validate the same.

contrib/babelfishpg_common/sql/upgrades/babelfish_common_helper--4.3.0--4.4.0.sql
Yvinayak07: The cast is dependent on the function (varchar2time) .Dropping the function directly results in failure of upgrade tests. 

verification:
1) created a view that used this cast.
`create view babel_5179_v1 as select cast(cast('12:45:37.123' as varchar) as time)`

2) Finding the oid of the view.
`select oid from pg_class where relname = 'babel_5179_v1';`

3) Finding the query using this oid
`select * from pg_rewrite where ev_class=19764;`

4) Finding the funcid from the query:
  = 17090

5) Finding Which function is dependent on this cast:
`select proname from pg_proc where oid=17090;`

output: 
    proname    
--------------
 varchar2time

conclusion: This view is dependent on varchar2time function. So we are deprecating this function instead of droppiing it directly.

cc: @Deepesh125

contrib/babelfishpg_common/src/varchar.c
Yvinayak07: Resolved

contrib/babelfishpg_common/src/varchar.c
Yvinayak07: Added default value and upgrade dependency tests

contrib/babelfishpg_common/sql/upgrades/babelfish_common_helper--4.3.0--4.4.0.sql
Yvinayak07: resolved

contrib/babelfishpg_common/sql/upgrades/babelfish_common_helper--4.3.0--4.4.0.sql
Yvinayak07: mentioned comments above

test/JDBC/expected/babel-5179-vu-prepare.out
```diff
@@ -0,0 +1,57 @@
...
+)
+go
+
+create procedure babel_5179_p1 as select cast(cast('12:45:37.123' as varchar) as time)
+go
+
+create view babel_5179_v2 as select cast(cast('12:45:37.123' as varchar) as time(2))
+go
+
+create table babel_5179_t2(col1 varchar(50), col2 as cast(col1 as time), constraint timecmp check(col1 >= '12:45:37.1'));
```
Anikait143: The definition is same as `babel_5179_t1`, lets use different typmod here

test/JDBC/expected/babel-5179-vu-prepare.out
```diff
@@ -0,0 +1,57 @@
...
+insert into babel_5179_t1 values ('12:45:37.1');
+go
+~~ROW COUNT: 1~~
+
+~~ROW COUNT: 1~~
+
+~~ROW COUNT: 1~~
+
+
+create view babel_5179_v11 as select col1, col2 from babel_5179_t1;
```
Anikait143: We should use the prepared objects in verify file as well to verify the output. Same for `babel_5179_v22`

Anikait143 (commented):
Please use more generalized test file names rather than Jira numbers

contrib/babelfishpg_common/sql/upgrades/babelfish_common_helper--4.3.0--4.4.0.sql
sumitj824: Function `sys.babelfish_drop_deprecated_object` itself raises the warning in case of exception, we can directly invoke it instead of using DO blocks.

test/JDBC/input/babel-5179-vu-prepare.sql
```diff
@@ -0,0 +1,45 @@
+create view babel_5179_v1 as select cast(cast('12:45:37.123' as varchar) as time)
```
sumitj824: please use meaningful testfile name instead of task number.

contrib/babelfishpg_common/sql/upgrades/babelfish_common_helper--4.3.0--4.4.0.sql
sumitj824: Can you explain where the above gets used? If you mean the original views will still use the original definition then we should not add DEFAULT in sql definition suggested below, in that case we should check if the arg is present then only use it.

test/JDBC/expected/babel-5179-vu-prepare.out
Yvinayak07: added with typmod = 2

test/JDBC/expected/babel-5179-vu-prepare.out
Yvinayak07: used in verify file

test/JDBC/input/babel-5179-vu-prepare.sql
Yvinayak07: changed name of file to cast-varchar-to-time-vu-*

contrib/babelfishpg_common/sql/upgrades/babelfish_common_helper--4.3.0--4.4.0.sql
Anikait143: Yes, that's the case where the original views will still use the original definition, also stated [above](https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/2877#discussion_r1733309961).

@Yvinayak07 please refer to the [suggestion](https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/2877#discussion_r1732485566).

shah-nirmit (approved):
LGTM

Anikait143 (approved):

contrib/babelfishpg_common/sql/upgrades/babelfish_common_helper--4.3.0--4.4.0.sql
```diff
@@ -0,0 +1,68 @@
...
+        GET STACKED DIAGNOSTICS error_msg = MESSAGE_TEXT;
+        raise warning '%', error_msg;
+end
+$$
+LANGUAGE plpgsql;
+
+-- (sys.VARCHAR AS pg_catalog.TIME)
+DROP CAST (sys.VARCHAR AS pg_catalog.TIME);
+
+DO $$
```
sumitj824: Sorry I missed this earlier, why do we need DO block here? Rename can be done directly here, please avoid doing unnecessary exception handling.

sumitj824 (approved):
LGTM, just address the last comment.

contrib/babelfishpg_common/sql/upgrades/babelfish_common_helper--4.3.0--4.4.0.sql
Yvinayak07: removed unnecessary exception handling

test/JDBC/input/cast-varchar-to-time-vu-prepare.sql
```diff
@@ -0,0 +1,48 @@
...
+create procedure babel_5179_p1 as select cast(cast('12:45:37.123' as varchar) as time)
+go
+
+create view babel_5179_v2 as select cast(cast('12:45:37.123' as varchar) as time(2))
+go
+
+create table babel_5179_t2(col1 varchar(50), col2 as cast(col1 as time(2)), constraint timecmp check(col1 >= '12:45:37.1'));
+go
+
+insert into babel_5179_t2 values ('12:45:37.123');
```
Deepesh125, line 30: just have similar insert script in verify also to make sure that table is restored expectedly after MVU

Deepesh125 (dismissed):

test/JDBC/input/cast-varchar-to-time-vu-prepare.sql
Yvinayak07: added insert statement tests in verify file

contrib/babelfishpg_common/sql/upgrades/babelfish_common_helper--4.3.0--4.4.0.sql
Deepesh125: +1

Deepesh125 (changes requested):

Deepesh125 (approved):

PR #2878, opened 2024-08-23
enable replace tests in version upgrade

tanscorpio7:
### Description

Enable replace tests in version upgrade framework with source major version = 13.
This test was earlier hanging in Github actions due to BABEL-5169 which has been fixed now with https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/2849.

Also removed replace-before-14_3-vu-* test files since these were/are not running in any framework.

### Issues Resolved

[No-Jira]

### Check List


test/JDBC/expected/replace-before-14_3-vu-verify.out
```diff
@@ -1,2855 +0,0 @@
...
-~~ERROR (Message: nondeterministic collations are not supported for substring searches)~~
-
-
-SELECT * FROM babel_4836_replace_dep_view1
-GO
-~~START~~
-text
-~~ERROR (Code: 33557097)~~
-
-~~ERROR (Message: nondeterministic collations are not supported for substring searches)~~
```
rohit01010, line 545: when we were testing these queries in upgrades coming from version before 14.3, to 15.latest, we were getting these errors which was not there for upgrades from version 14.3 and above to 15.latest. And this output difference was expected as before 14.3 we were using PG replace function for replace, which will throw error for non-deterministic collations. Hence we added these tests. Can you explain why we are not getting this error for upgrades from version before 14.3, to 16.latest ?

Deepesh125 (approved):

test/JDBC/expected/replace-before-14_3-vu-verify.out
tanscorpio7: We supported replace and charindex for non deterministic collation by putting hooks in postgres strpos() and replace() functions. We did not add a dialect check here because 
1. Dialect is set to PSQL when executing non pltsql functions. So dialect == TSQL would not work here.
2. We used ICU functions to implement this, hence replace() and strpos() will work even for postgres user defined non deterministic collation correctly.

rohit01010 (approved):

PR #2885, opened 2024-08-26
Support increment of identity sequence in TSQL Bulk Copy when in 'keep identity values' mode

Anikait143:
### 1. Issues
- Bcp / SqlBulkCopy / insert bulk should increment the identity sequence value when in "keep identity values" mode
- Dotnet Github workflow doesnot write output in Console

- Supported increment of identity sequence in TSQL Bulk Copy when in 'keep identity values' mode by seeding the identity sequence correctly during BulkCopy.
-  Enabled `VSTEST_DISABLE_STANDARD_OUTPUT_CAPTURING` flag to print console output in dotnet github action workflow.
- Added Test Cases for the issue.

Task: BABEL- 4865, BABEL- 4997

* **Use case based -**  YES

* **Arbitrary inputs -** YES

* **Performance tests -**

* **Tooling impact -**

* **Memory tests -**

* **Client tests -**

### Check List


test/dotnet/input/bcpOptions.txt
```diff
@@ -230,6 +230,131 @@ Select * from destinationTable
 drop table sourceTable
 drop table destinationTable
 
+Create table sourceTable(a int, b int identity)
```
KushaalShroff, line 234: Lets add test case with primary key + identity

Must raise error for 1st row and 1001th row. To test cleanup just rerun the insertbulk multiple times

KushaalShroff (changes requested):

contrib/babelfishpg_tsql/src/pltsql_bulkcopy.c
```diff
@@ -979,6 +1013,48 @@ EndBulkCopy(BulkCopyState cstate, bool aborted)
...
+				if (strcmp(defel->defname, "start") == 0)
+				{
+					init_identity_value = defGetInt64(defel);
+					break;
+				}
+			}
+
+			PG_TRY();
+			{
+				init_identity_value = DirectFunctionCall1(pg_sequence_last_value,
```
KushaalShroff, line 1041: Our implementation in get_identity_current is the other way around - if we get an error returning from pg_sequence_last_value then only we set ``init_identity_value = defGetInt64(defel);`` 

Which seems to be better, we only check for sequence_options when pg_sequence_last_value throws an error.

The comment also describes it:
```
 * Given a table name with an identity column, fetch the last identity sequence
 * value stored in the pg_sequences catalog. If not set, return the seed value
 * instead. Return NULL on error.
 ```

File: test/dotnet/utils/TestUtils.cs [hunk 3]
```diff
@@ -100,6 +100,38 @@ public bool insertBulkCopyWithTransaction(DbConnection bblCnn, DbCommand bblCmd,
...
+			DataTable dataTable = new DataTable();
+			try
+			{
+				reader = bblCmd.ExecuteReader();
+				dataTable.Load(reader);
+				reader.Close();
+
+				/* Set KeepIdentity default for this API */
+				SqlBulkCopy bulkCopy = new SqlBulkCopy((SqlConnection)bblCnn, SqlBulkCopyOptions.KeepIdentity, null);
+				bulkCopy.DestinationTableName = destinationTable;
```
KushaalShroff, line 91: Why have we introduced a new function? We need to add this option in transinsertbulk and reuse

test/dotnet/src/BatchRun.cs
```diff
@@ -226,6 +226,15 @@ bool BatchRunner(DbConnection bblCnn, string queryFilePath, Serilog.Core.Logger
 							string destinationTable = result[2];
 							testFlag &= testUtils.insertBulkCopyWithTransaction(bblCnn, bblCmd, sourceTable, destinationTable, bblTransaction, logger, ref stCount);
 						}
+						else if (strLine.ToLowerInvariant().StartsWith("keepidentityinsertbulk"))
```
KushaalShroff: Not needed

contrib/babelfishpg_tsql/src/pltsql_bulkcopy.c
```diff
@@ -703,6 +703,15 @@ ExecuteBulkCopy(BulkCopyState cstate, int rowCount, int colCount,
 					else
 					{
 						myslot->tts_values[col_index_to_insert] = Values[col_index_to_fetch];
+
+						/* In case of identity column, store the updated identity sequence value. */
+						if (cstate->seq_index == i)
```
KushaalShroff, line 717: What will happen if in a concurrent session we do a simple insert in this table, how can we handle such scenarios? what is TSQL behaviour?

test/dotnet/input/bcpOptions.txt
```diff
@@ -230,6 +230,203 @@ Select * from destinationTable
...
+drop table sourceTable
+drop table sourceTable1
+drop table destinationTable
+
+create table sourceTable(c1 int identity, c2 CHAR(1024), check (c1 < 1000))
+ set IDENTITY_INSERT sourceTable ON
+INSERT INTO sourceTable (c1, c2) SELECT generate_series(1, 1001, 1), 'Foo';
+ set IDENTITY_INSERT sourceTable OFF
+create table destinationTable(c1 int identity, c2 CHAR(1024), check (c1 < 500))
+ set IDENTITY_INSERT destinationTable ON
```
KushaalShroff: What is TSQL behaviour if you set IDENTITY_INSERT destinationTable ON inside the same connection which does the insert bulk?
you can use transinsertbulk keyword in .net to check

test/dotnet/input/InsertBulk/insertBulk.txt
```diff
@@ -296,6 +296,110 @@ Select * from destinationTable
 drop table sourceTable
 drop table destinationTable
 
+Create table sourceTable(a int, b int identity)
```
KushaalShroff, line 351: Add comments on top of the tests and explain their purpose

test/dotnet/input/InsertBulk/insertBulkErrors.txt
```diff
@@ -555,4 +555,77 @@ drop table sourceTable
 drop table sourceTable1
 drop table destinationTable
 
- SET implicit_transactions OFF
\ No newline at end of file
+ SET implicit_transactions OFF
+
+ # identity
```
KushaalShroff: Add a comment about identity errors testing at the top of this file as well

test/dotnet/input/InsertBulk/insertBulkErrors.txt
```diff
@@ -555,4 +555,77 @@ drop table sourceTable
...
+drop table sourceTable1
+drop table destinationTable
+
+create table sourceTable(c1 int identity, c2 CHAR(1024), check (c1 < 1000))
+ set IDENTITY_INSERT sourceTable ON
+INSERT INTO sourceTable (c1, c2) SELECT generate_series(1, 1001, 1), 'Foo';
+ set IDENTITY_INSERT sourceTable OFF
+create table destinationTable(c1 int identity, c2 CHAR(1024), check (c1 < 500))
+ set IDENTITY_INSERT destinationTable ON
+INSERT INTO destinationTable (c1, c2) VALUES(1001, 'Foo')
```
KushaalShroff: Use transinsertbulk since we enforce check constraints.

What is the behaviour of TSQL in case of check constraints off and KeepIdentity is on, I am guessing user can go on inserting c1>1000. 

Can we check that? These tests dont really check that

test/dotnet/input/InsertBulk/insertBulkErrors.txt
```diff
@@ -555,4 +555,77 @@ drop table sourceTable
...
+select Ident_CURRENT('sourceTable')
+select Ident_CURRENT('sourceTable1')
+select Ident_CURRENT('destinationTable')
+drop table sourceTable
+drop table sourceTable1
+drop table destinationTable
+
+create table sourceTable(c1 int identity, c2 CHAR(1024), check (c1 < 1000))
+ set IDENTITY_INSERT sourceTable ON
+INSERT INTO sourceTable (c1, c2) SELECT generate_series(1, 1001, 1), 'Foo';
```
KushaalShroff, line 594: How will this work? we need to remove the check constraint from the sourceTable-s and then it becomes a test case for insertBulk with check constraints

KushaalShroff (changes requested):

contrib/babelfishpg_tsql/src/pltsql_bulkcopy.c
Anikait143: In the scenario, where performing a bulk insert from a source table to a destination table, and there is a concurrent insert happening in the source table, the impact on the identity values in the destination table will depend on the order of operations.

Here's what could happen:

1. `Bulk Insert Happens First`: If the bulk insert operation from the source table to the destination table completes before the concurrent insert in the source table, the identity values in the destination table will be consistent with the source table at the time of the bulk insert.
In this case, the concurrent insert in the source table will not affect the identity values in the destination table because the bulk insert has already copied the data, including the identity values, from the source table.

2. `Concurrent Insert Happens First`: If the concurrent insert in the source table happens before the bulk insert operation, the identity values in the destination table may be affected.
When the bulk insert operation copies data from the source table, it will include the newly inserted row with the new identity value from the concurrent insert. As a result, the identity value sequence in the destination table will have a gap or a duplicate value, depending on how the bulk insert process handles identity values.

Here's what we have done to handle these scenarios:

1. For Scenario-1, if Bulk insert happens first then obviously the flow will continue and there will be no implications.

2. However for Scenario-2, if Insert happens before seeding the identity value, then we will consider the inserted row to calculate the new_identity_value during bulk insert and then while seeding the identity value we will impose lock on the table to prevent further inserts until the end of bulk insert process.

test/dotnet/input/InsertBulk/insertBulkErrors.txt
Anikait143: yes, in TSQL if check constraint is off and KeepIdentity is on, we can insert c1>=500. Added test for the same.

contrib/babelfishpg_tsql/src/pltsql_bulkcopy.c
Anikait143: Makes sense, reversed the order

test/dotnet/utils/TestUtils.cs
Anikait143: Done

test/dotnet/src/BatchRun.cs
Anikait143: Done

test/dotnet/input/bcpOptions.txt
Anikait143: Added tests

test/dotnet/input/InsertBulk/insertBulk.txt
Anikait143: Added comments

test/dotnet/input/InsertBulk/insertBulkErrors.txt
Anikait143: Done

contrib/babelfishpg_tsql/src/pltsql_bulkcopy.c
KushaalShroff: > If the concurrent insert in the source table happens

The question was about concurrent insert in destination table. What will happen if I do concurrent BCP on same destination table? Our logic today for identity does not take care of this.

Anikait143: In case of concurrent insert in destination table:
As per our current implementation, we seed the identity sequence value at the end of BCP execution.
So, having concurrent insert will insert the row with identity sequence value incremented from last seed value and BCP will insert rows with given identity values (keep identity is ON), so there might be difference in Identity sequence value as compared to TSQL which will hold Row lock in each row inserted by BCP, seeding the new value each time, so the concurrent insert will have identity sequence value incremented from new seed value.
Consider this scenario:
Session 1 executes:
```
SET IDENTITY_INSERT MyTable ON;
INSERT INTO MyTable (Id, Name) VALUES (10, 'A'), (11, 'B');
SET IDENTITY_INSERT MyTable OFF;
```
Session 2 executes concurrently:
```
INSERT INTO MyTable (Name) VALUES ('C');
```
Outcome: (TSQL)

- Session 1 inserts records with explicit identity values (10, 11).
- Session 2 inserts a record without specifying an identity value. TSQL will generate the next value in the sequence for the identity column (e.g., if the next value should be 12, then 12 will be used).

whereas in BBF for session since we haven't yet seeded the new value from session-1 i.e. 11, so session will insert value with old seed value(e.g., if the next value should be 1, then 1 will be used).

To resolve this we can have seed value updated on each row copied, but it will have it's performance overhead.

KushaalShroff: Need to update this thread about our discussion. We will keep working on this and let this PR get merged for now

KushaalShroff (approved):

test/dotnet/ExpectedOutput/insertBulk.out
```diff
@@ -379,6 +379,244 @@ hello#!#jello
...
+#Q# set IDENTITY_INSERT sourceTable1 OFF
+#Q#Select count(c1) from sourceTable
+#D#int
+1001
+#Q#select count(c1) from sourceTable1
+#D#int
+1001
+#Q#select count(c1) from destinationTable
+#D#int
+2002
```
KushaalShroff, line 400: The exact same test case written in insertBulkErrors.out has this output:
```
#Q#select count(c1) from sourceTable1
#D#int
1001
#Q#select count(c1) from destinationTable
#D#int
0
#Q#select Ident_CURRENT('sourceTable')
#D#decimal
1001
#Q#select Ident_CURRENT('sourceTable1')
#D#decimal
1001
#Q#select Ident_CURRENT('destinationTable')
#D#decimal
1
```
Here we seem to have copied rows for destinationTable but not in insertBulkErrors.out
Please explain and fix it

PR #2887, opened 2024-08-28
Fix Error Handling and Cleanup during Insert Bulk Process

KushaalShroff:
### Description
This commit fixes an issue with the error handling and cleanup phase of the Insert Bulk Process.
1. For Error Handling there was a scenario where bulk_load_callback(0, 0, NULL, NULL) call for cleanup would flush the remaining rows during EndBulkCopy and could result in an error. To fix this, we move the transaction rollback logic from TSQL extension to TDS. We also improved it by using Savepoints in case of active transaction which is aligned with TSQL Behaviour
2. In this case, if we have a reset-connection after the Bulk Load TDS packet we werent cleaning up the Bulk Load state. To do so we reset the offset.
3. During Reset Connection TDS is not resetting any TSQL transaction semantic. To resolve this we introduce a wrapper of AbortOutOfAnyTransaction to reset NestedTranCount.

Issues Resolved
BABEL-5200, BABEL-5199, BABEL-5220

Signed-off-by: Kushaal Shroff [kushaal@amazon.com](mailto:kushaal@amazon.com)

1. Testing explicit transaction (error case handled in 5.)
a. Commit without error
b. Rollback without error
2. Index with without transaction
4. Primary Key error case
5. Unique constraint with error case
6. Check constraint with error case
a. transaction testing during error scenarios
b. @@trancount test - error should not terminate transaction
c. Test CheckConstraint BCP Option Enabled
d. Test Reusing the same connection for BCP even after error scenarios
7. Reset-connection testing with Primary Key error
The above tests test the seq and index.

For Reset-connection, Although the tests in 6. do reset but for validation through logs I added a temp log to debug TdsResetConnection and with this log its evident that pid 21384 is being reset and reused for Bulk Load
```
2024-08-27 10:52:28.688 UTC [21384] CONTEXT:  TDS Protocol: Message Type: SQL BATCH, Phase: TDS_REQUEST_PHASE_FETCH. Resetting the TDS connection
2024-08-27 10:52:28.696 UTC [21384] ERROR:  duplicate key value violates unique constraint "destinationtable_pkey"
2024-08-27 10:52:28.696 UTC [21384] DETAIL:  Key (c1)=(2) already exists.
2024-08-27 10:52:28.696 UTC [21384] CONTEXT:  TDS Protocol: Message Type: Bulk Load, Phase: TDS_REQUEST_PHASE_PROCESS. Processing Bulk Load Request
2024-08-27 10:52:28.697 UTC [21384] LOG:  BULK LOAD RESET NOW
2024-08-27 10:52:28.697 UTC [21384] CONTEXT:  TDS Protocol: Message Type: SQL BATCH, Phase: TDS_REQUEST_PHASE_FETCH. Resetting the TDS connection
2024-08-27 10:52:28.706 UTC [21384] ERROR:  duplicate key value violates unique constraint "destinationtable_pkey"
2024-08-27 10:52:28.706 UTC [21384] DETAIL:  Key (c1)=(2) already exists.
2024-08-27 10:52:28.706 UTC [21384] CONTEXT:  TDS Protocol: Message Type: Bulk Load, Phase: TDS_REQUEST_PHASE_PROCESS. Processing Bulk Load Request
2024-08-27 10:52:28.707 UTC [21384] LOG:  BULK LOAD RESET NOW
2024-08-27 10:52:28.707 UTC [21384] CONTEXT:  TDS Protocol: Message Type: SQL BATCH, Phase: TDS_REQUEST_PHASE_FETCH. Resetting the TDS connection
2024-08-27 10:52:28.712 UTC [21384] ERROR:  duplicate key value violates unique constraint "destinationtable_pkey"
2024-08-27 10:52:28.712 UTC [21384] DETAIL:  Key (c1)=(2) already exists.
2024-08-27 10:52:28.712 UTC [21384] CONTEXT:  TDS Protocol: Message Type: Bulk Load, Phase: TDS_REQUEST_PHASE_PROCESS. Processing Bulk Load Request
2024-08-27 10:52:28.714 UTC [21384] LOG:  BULK LOAD RESET NOW
2024-08-27 10:52:28.714 UTC [21384] CONTEXT:  TDS Protocol: Message Type: SQL BATCH, Phase: TDS_REQUEST_PHASE_FETCH. Resetting the TDS connection
2024-08-27 10:52:28.719 UTC [21384] ERROR:  duplicate key value violates unique constraint "destinationtable_pkey"
2024-08-27 10:52:28.719 UTC [21384] DETAIL:  Key (c1)=(2) already exists.
2024-08-27 10:52:28.719 UTC [21384] CONTEXT:  TDS Protocol: Message Type: Bulk Load, Phase: TDS_REQUEST_PHASE_PROCESS. Processing Bulk Load Request
2024-08-27 10:52:28.720 UTC [21384] LOG:  BULK LOAD RESET NOW
2024-08-27 10:52:28.720 UTC [21384] CONTEXT:  TDS Protocol: Message Type: SQL BATCH, Phase: TDS_REQUEST_PHASE_FETCH. Resetting the TDS connection
```

### Check List

### DIFF
--- a/contrib/babelfishpg_tds/src/backend/tds/tdsbulkload.c
+++ b/contrib/babelfishpg_tds/src/backend/tds/tdsbulkload.c
@@ -36,7 +36,7 @@ static void FetchMoreBcpData(StringInfo *message, int dataLenToRead, bool freeMe
 static void FetchMoreBcpPlpData(StringInfo *message, int dataLenToRead);
 static int	ReadBcpPlp(ParameterToken temp, StringInfo *message, TDSRequestBulkLoad request);
 static void FreePlpToken(ParameterToken token);
-uint64_t	offset = 0;
+uint64_t	volatile bcpOffset = 0;
 
 #define COLUMNMETADATA_HEADER_LEN			sizeof(uint32_t) + sizeof(uint16) + 1
 #define FIXED_LEN_TYPE_COLUMNMETADATA_LEN	1
@@ -80,7 +80,7 @@ do \
 #define CheckMessageHasEnoughBytesToReadColMetadata(message, dataLen) \
 do \
 { \
-	if ((*message)->len - offset < dataLen) \
+	if ((*message)->len - bcpOffset < dataLen) \
 		FetchMoreBcpData(message, dataLen, false); \
 } while(0)
 
@@ -91,7 +91,7 @@ do \
 #define CheckMessageHasEnoughBytesToReadRows(message, dataLen) \
 do \
 { \
-	if ((*message)->len - offset < dataLen) \
+	if ((*message)->len - bcpOffset < dataLen) \
 		FetchMoreBcpData(message, dataLen, true); \
 } while(0)
 
@@ -99,14 +99,14 @@ do \
 #define CheckPlpMessageHasEnoughBytesToRead(message, dataLen) \
 do \
 { \
-	if ((*message)->len - offset < dataLen) \
+	if ((*message)->len - bcpOffset < dataLen) \
 		FetchMoreBcpPlpData(message, dataLen); \
 } while(0)
 
 void
 TdsResetBcpOffset()
 {
-	offset = 0;
+	bcpOffset = 0;
 }
 
 static void
@@ -140,12 +140,12 @@ FetchMoreBcpData(StringInfo *message, int dataLenToRead, bool freeMessageData)
 	if (freeMessageData)
 	{
 		temp = makeStringInfo();
-		appendBinaryStringInfo(temp, (*message)->data + offset, (*message)->len - offset);
+		appendBinaryStringInfo(temp, (*message)->data + bcpOffset, (*message)->len - bcpOffset);
 
 		if ((*message)->data)
 			pfree((*message)->data);
 		pfree((*message));
-		offset = 0;
+		bcpOffset = 0;
 	}
 	else
 		temp = *message;
@@ -153,7 +153,7 @@ FetchMoreBcpData(StringInfo *message, int dataLenToRead, bool freeMessageData)
 	/*
 	 * Keep fetching for additional packets until we have enough data to read.
 	 */
-	while (dataLenToRead + offset > temp->len)
+	while (dataLenToRead + bcpOffset > temp->len)
 	{
 		/*
 		 * We should hold the interrupts until we read the next request frame.
@@ -180,7 +180,7 @@ FetchMoreBcpData(StringInfo *message, int dataLenToRead, bool freeMessageData)
 
 /*
  * Incase of PLP data we should not discard the previous packet since we
- * first store the offset of the PLP Chunks first and then read the data later.
+ * first store the bcpOffset of the PLP Chunks first and then read the data later.
  */
 static void
 FetchMoreBcpPlpData(StringInfo *message, int dataLenToRead)
@@ -205,7 +205,7 @@ FetchMoreBcpPlpData(StringInfo *message, int dataLenToRead)
 	/*
 	 * Keep fetching for additional packets until we have enough data to read.
 	 */
-	while (dataLenToRead + offset > (*message)->len)
+	while (dataLenToRead + bcpOffset > (*message)->len)
 	{
 		/*
 		 * We should hold the interrupts until we read the next request frame.
@@ -247,33 +247,34 @@ GetBulkLoadRequest(StringInfo message)
 	request->rowData = NIL;
 	request->reqType = TDS_REQUEST_BULK_LOAD;
 
-	if (unlikely((uint8_t) message->data[offset] != TDS_TOKEN_COLMETADATA))
+	TdsResetBcpOffset();
+	if (unlikely((uint8_t) message->data[bcpOffset] != TDS_TOKEN_COLMETADATA))
 		ereport(ERROR,
 				(errcode(ERRCODE_PROTOCOL_VIOLATION),
 				 errmsg("The incoming tabular data stream (TDS) Bulk Load Request (BulkLoadBCP) protocol stream is incorrect. "
 						"unexpected token encountered processing the request.")));
 
-	offset++;
+	bcpOffset++;
 
-	memcpy(&colCount, &message->data[offset], sizeof(uint16));
+	memcpy(&colCount, &message->data[bcpOffset], sizeof(uint16));
 	colmetadata = palloc0(colCount * sizeof(BulkLoadColMetaData));
 	request->colCount = colCount;
 	request->colMetaData = colmetadata;
-	offset += sizeof(uint16);
+	bcpOffset += sizeof(uint16);
 
 	for (int currentColumn = 0; currentColumn < colCount; currentColumn++)
 	{
 		CheckMessageHasEnoughBytesToReadColMetadata(&message, COLUMNMETADATA_HEADER_LEN);
 		/* UserType */
-		memcpy(&colmetadata[currentColumn].userType, &message->data[offset], sizeof(uint32_t));
-		offset += sizeof(uint32_t);
+		memcpy(&colmetadata[currentColumn].userType, &message->data[bcpOffset], sizeof(uint32_t));
+		bcpOffset += sizeof(uint32_t);
 
 		/* Flags */
-		memcpy(&colmetadata[currentColumn].flags, &message->data[offset], sizeof(uint16));
-		offset += sizeof(uint16);
+		memcpy(&colmetadata[currentColumn].flags, &message->data[bcpOffset], sizeof(uint16));
+		bcpOffset += sizeof(uint16);
 
 		/* TYPE_INFO */
-		colmetadata[currentColumn].columnTdsType = message->data[offset++];
+		colmetadata[currentColumn].columnTdsType = message->data[bcpOffset++];
 
 		/* Datatype specific Column Metadata. */
 		switch (colmetadata[currentColumn].columnTdsType)
@@ -285,14 +286,14 @@ GetBulkLoadRequest(StringInfo message)
 			case TDS_TYPE_DATETIMEN:
 			case TDS_TYPE_UNIQUEIDENTIFIER:
 				CheckMessageHasEnoughBytesToReadColMetadata(&message, FIXED_LEN_TYPE_COLUMNMETADATA_LEN);
-				colmetadata[currentColumn].maxLen = message->data[offset++];
+				colmetadata[currentColumn].maxLen = message->data[bcpOffset++];
 				break;
 			case TDS_TYPE_DECIMALN:
 			case TDS_TYPE_NUMERICN:
 				CheckMessageHasEnoughBytesToReadColMetadata(&message, NUMERIC_COLUMNMETADATA_LEN);
-				colmetadata[currentColumn].maxLen = message->data[offset++];
-				colmetadata[currentColumn].precision = message->data[offset++];
-				colmetadata[currentColumn].scale = message->data[offset++];
+				colmetadata[currentColumn].maxLen = message->data[bcpOffset++];
+				colmetadata[currentColumn].precision = message->data[bcpOffset++];
+				colmetadata[currentColumn].scale = message->data[bcpOffset++];
 				break;
 			case TDS_TYPE_CHAR:
 			case TDS_TYPE_VARCHAR:
@@ -300,12 +301,12 @@ GetBulkLoadRequest(StringInfo message)
 			case TDS_TYPE_NVARCHAR:
 				{
 					CheckMessageHasEnoughBytesToReadColMetadata(&message, STRING_COLUMNMETADATA_LEN);
-					memcpy(&colmetadata[currentColumn].maxLen, &message->data[offset], sizeof(uint16));
-					offset += sizeof(uint16);
+					memcpy(&colmetadata[currentColumn].maxLen, &message->data[bcpOffset], sizeof(uint16));
+					bcpOffset += sizeof(uint16);
 
-					memcpy(&collation, &message->data[offset], sizeof(uint32_t));
-					offset += sizeof(uint32_t);
-					colmetadata[currentColumn].sortId = message->data[offset++];
+					memcpy(&collation, &message->data[bcpOffset], sizeof(uint32_t));
+					bcpOffset += sizeof(uint32_t);
+					colmetadata[currentColumn].sortId = message->data[bcpOffset++];
 					colmetadata[currentColumn].encoding = TdsGetEncoding(collation);
 				}
 				break;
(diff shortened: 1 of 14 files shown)


contrib/babelfishpg_tds/src/backend/tds/tdsbulkload.c
```diff
@@ -102,6 +103,12 @@ do \
 		FetchMoreBcpPlpData(message, dataLen); \
 } while(0)
 
+void
+TdsResetBcpOffset()
+{
+	offset = 0;
```
surendravishnoi: it is a global variable, should be named with context like bcp_offset. Also, can we reset it when we detect a new BCP request?

contrib/babelfishpg_tds/src/backend/tds/tdsbulkload.c
```diff
@@ -899,6 +906,60 @@ SetBulkLoadRowData(TDSRequestBulkLoad request, StringInfo message)
...
+
+	RESUME_CANCEL_INTERRUPTS();
+
+	if (ret < 0)
+		TdsErrorContext->err_text = "EOF on TDS socket while fetching For Bulk Load Request";
+
+	if (internal_sp_started && before_lxid == MyProc->lxid && before_subtxn_id == GetCurrentSubTransactionId())
+	{
+		if (TDS_DEBUG_ENABLED(TDS_DEBUG2))
+			elog(LOG, "TSQL TXN PG semantics : Rollback internal savepoint");
```
surendravishnoi, line 943: use BCP context in logging, BCP TXN semantics

contrib/babelfishpg_tds/src/backend/tds/tdsbulkload.c
```diff
@@ -899,6 +906,60 @@ SetBulkLoadRowData(TDSRequestBulkLoad request, StringInfo message)
...
+		 * did not match, then we shall rollback the current transaction and let the
+		 * the actual error be relayed to the customer.
+		 */
+		elog(LOG, "The current transaction is rolled back since it "
+				"was in inconsistent state during Bulk Copy");
+		pltsql_plugin_handler_ptr->pltsql_rollback_txn_callback();
+	}
+	MemoryContextSwitchTo(oldcontext);
+	RESUME_INTERRUPTS();
+	TdsResetBcpOffset();
```
surendravishnoi: should it be the first thing, that way we do not leak offset if there is an error during above commands

contrib/babelfishpg_tds/src/backend/tds/tdsbulkload.c
```diff
@@ -910,11 +971,31 @@ ProcessBCPRequest(TDSRequest request)
...
 	pgstat_report_activity(STATE_RUNNING, "Processing Bulk Load Request");
 
+	/*
+	 * If a transaction is active then start a Savepoint to rollback
+	 * later in case of error.
+	 */
+	if (IsTransactionBlockActive())
+	{
+		if (TDS_DEBUG_ENABLED(TDS_DEBUG2))
+			elog(LOG, "TSQL TXN Start internal savepoint");
```
surendravishnoi, line 997: BCP context

contrib/babelfishpg_tds/src/backend/tds/tdsbulkload.c
```diff
@@ -955,79 +1020,75 @@ ProcessBCPRequest(TDSRequest request)
...
+						 req->rowCount, req->colCount),
+						 errhidestmt(true)));
+
+			CleanupBCPDuringError(internal_sp_started, before_subtxn_id,
+					 before_lxid, oldowner, oldcontext);
+			PG_RE_THROW();
+		}
+		PG_END_TRY();
+		/* Free the List of Rows. */
+		if (req->rowData)
```
surendravishnoi: this should go inside TRY block

contrib/babelfishpg_tds/src/backend/tds/tdsbulkload.c
```diff
@@ -955,79 +1020,75 @@ ProcessBCPRequest(TDSRequest request)
...
-			if (values)
-				pfree(values);
-			if (nulls)
-				pfree(nulls);
 		}
+		/* If there we no rows then we have reached the end of the loop. */
+		else
+			break;
+
+		if (values)
```
surendravishnoi: go inside try block

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -3632,27 +3632,11 @@ execute_bulk_load_insert(int ncol, int nrow,
 		 * In an error condition, the caller calls the function again to do
 		 * the cleanup.
 		 */
-		MemoryContext oldcontext;
```
surendravishnoi, line 3635: should we reset offset here?

contrib/babelfishpg_tds/src/backend/tds/tdsbulkload.c
```diff
@@ -899,6 +906,60 @@ SetBulkLoadRowData(TDSRequestBulkLoad request, StringInfo message)
...
+	else
+	{
+		/*
+		 * In the case of an error and transaction is active but the earlier savepoint
+		 * did not match, then we shall rollback the current transaction and let the
+		 * the actual error be relayed to the customer.
+		 */
+		elog(LOG, "The current transaction is rolled back since it "
+				"was in inconsistent state during Bulk Copy");
+		pltsql_plugin_handler_ptr->pltsql_rollback_txn_callback();
```
surendravishnoi, line 961: we should do commit, start transaction after it. Have we tested this scenario manually?

contrib/babelfishpg_tds/src/backend/tds/tdsbulkload.c
```diff
@@ -899,6 +906,60 @@ SetBulkLoadRowData(TDSRequestBulkLoad request, StringInfo message)
...
+		StartTransactionCommand();
+	}
+	else
+	{
+		/*
+		 * In the case of an error and transaction is active but the earlier savepoint
+		 * did not match, then we shall rollback the current transaction and let the
+		 * the actual error be relayed to the customer.
+		 */
+		elog(LOG, "The current transaction is rolled back since it "
```
surendravishnoi, line 959: we should give error after rollback

contrib/babelfishpg_tds/src/backend/tds/tdsbulkload.c
```diff
@@ -899,6 +906,60 @@ SetBulkLoadRowData(TDSRequestBulkLoad request, StringInfo message)
...
+	HOLD_INTERRUPTS();
+
+	/*
+	 * Discard remaining TDS_BULK_LOAD packets only if End of
+	 * Message has not been reached for the current request.
+	 * Otherwise we have no TDS_BULK_LOAD packets left for the
+	 * current request that need to be discarded.
+	 */
+	if (!TdsGetRecvPacketEomStatus())
+		ret = TdsDiscardAllPendingBcpRequest();
```
kuntalghosh, line 933: Let's be explicit about cancelling interrupts
```
HOLD_CANCEL_INTERRUPTS();
ret = TdsDiscardAllPendingBcpRequest();
RESUME_CANCEL_INTERRUPTS();
```

contrib/babelfishpg_tds/src/backend/tds/tdsbulkload.c
```diff
@@ -899,6 +906,60 @@ SetBulkLoadRowData(TDSRequestBulkLoad request, StringInfo message)
...
+static void
+CleanupBCPDuringError(bool internal_sp_started,
+					 volatile int before_subtxn_id,
+					 volatile int before_lxid,
+					 ResourceOwner oldowner,
+					 MemoryContext oldcontext)
+{
+	int			ret = 0;
+
+	HOLD_CANCEL_INTERRUPTS();
```
kuntalghosh: Let's move this to its parent function add a comment
```
hold interrupts during cleanup
```

test/dotnet/utils/TestUtils.cs
```diff
@@ -39,10 +39,50 @@ public bool insertBulkCopy(DbConnection bblCnn, DbCommand bblCmd, String sourceT
 			DbDataReader reader = null;
 			try
 			{
+				/* To Enforce Reset Connection. */
 				reader = bblCmd.ExecuteReader();
-				SqlBulkCopy bulkCopy = new SqlBulkCopy(ConfigSetup.BblConnectionString);
+				using (SqlConnection destinationConnection =
```
surendravishnoi, line 44: how does this reset connection?

test/dotnet/input/InsertBulk/insertBulkErrors.txt
```diff
@@ -0,0 +1,322 @@
+##########################################################
+#################### TEST DETAILS ########################
+### 1. Testing explicit transaction (error case handled in 5.)
+###    a. Commit without error
+###    b. Rollback without error
+### 2. Index with without transaction
+### 3. Primary Key error case
```
surendravishnoi, line 7: we should test each error case without transaction, with transaction followed by commit and rollback
we should also add tests for savepoints and rollback to savepoint, without error and with error
we should manually test both error cases where savepoint state is not as expected
if supported, test with implicit transaction on/off, xact abort on/off
I am not clear how we are testing reset connection with pooling
how are we testing out of order packet error?

test/dotnet/ExpectedOutput/insertBulkErrors.out
```diff
@@ -0,0 +1,376 @@
+#Q#Create table sourceTable(a int, b int not null)
```
surendravishnoi, line 1: can we integrate heap AM check in test framework?

contrib/babelfishpg_tds/src/backend/tds/tdsbulkload.c
KushaalShroff: This is local to this file, we have not done anything unique for other TDS packet types. Yes we can reset it when reading new request as well

contrib/babelfishpg_tds/src/backend/tds/tdsbulkload.c
KushaalShroff: Please check TdsErrorContext->err_text, there is already context added to every log statement in Bulk Load

contrib/babelfishpg_tds/src/backend/tds/tdsbulkload.c
KushaalShroff: Okay, what could go wrong during pfree?

contrib/babelfishpg_tds/src/backend/tds/tdsbulkload.c
kuntalghosh: Can't we reset this offset whenever we call GetBulkLoadRequest()?

Do we expect a non-zero offset across two GetBulkLoadRequest() call?

Besides, if you're using this variable in catch block, better to declare it as volatile.

contrib/babelfishpg_tds/src/backend/tds/tdsbulkload.c
```diff
@@ -929,24 +1010,8 @@ ProcessBCPRequest(TDSRequest request)
...
-			 */
-			if (!TdsGetRecvPacketEomStatus())
-				ret = TdsDiscardAllPendingBcpRequest();
-
-			RESUME_CANCEL_INTERRUPTS();
-
-			if (ret < 0)
-				TdsErrorContext->err_text = "EOF on TDS socket while fetching For Bulk Load Request";
-
+			CleanupBCPDuringError(internal_sp_started, before_subtxn_id,
```
kuntalghosh, line 1019: I think it's possible to combine the two TRY..CATCH block into a single one. The cleanup is working in the same way. 

```
while(1)
{
   PG_TRY();
   {
        message = SetBulkLoadRowData(req, message);

        /*
         * If the row-count is 0 then there are no rows left to be inserted.
         * We should begin with cleanup.
         */
        if (req->rowCount == 0)
        {
            /* Using Same callback function to do the clean-up. */
            pltsql_plugin_handler_ptr->bulk_load_callback(0, 0, NULL, NULL);
            break;
        }
    
        /* should be valid as rowCount > 0 */
        Assert(req->rowData);

        nargs = req->colCount * req->rowCount;
        values = palloc0(nargs * sizeof(Datum));
        nulls = palloc0(nargs * sizeof(bool));
    
        /* Flaten and create a 1-D array of Value & Datums */
        foreach(lc, req->rowData)
        {
            BulkLoadRowData *row = (BulkLoadRowData *) lfirst(lc);
    
            for (int currentColumn = 0; currentColumn < req->colCount; currentColumn++)
            {
                if (row->isNull[currentColumn]) /* null */
                    nulls[count] = row->isNull[currentColumn];
                else
                    values[count] = row->columnValues[currentColumn];
                count++;
            }
        }
    
        retValue += pltsql_plugin_handler_ptr->bulk_load_callback(req->colCount,
                                                                    req->rowCount, values, nulls);
    
        /* Free the List of Rows. */
        list_free_deep(req->rowData);
        req->rowData = NIL;

        if (values)
            pfree(values);
        if (nulls)
            pfree(nulls);
    }
    PG_CATCH();
    {
        int         ret = 0;

        /* hold interrupts during cleanup */
        HOLD_INTERRUPTS();
        CleanupBCPDuringError(..);
        RESUME_INTERRUPTS();

        PG_RE_THROW();
    }
    PG_END_TRY();
}
```

test/dotnet/utils/TestUtils.cs
KushaalShroff: pooling=false is removed from connection string and .net driver does not reset connection if we only provide this connectionString to the SqlBulkCopy API, but if we manually create and open this connection object then it will reset connection. Check PR description for validation through logging

contrib/babelfishpg_tds/src/backend/tds/tdsbulkload.c
```diff
@@ -899,6 +906,60 @@ SetBulkLoadRowData(TDSRequestBulkLoad request, StringInfo message)
...
+	 */
+	if (!TdsGetRecvPacketEomStatus())
+		ret = TdsDiscardAllPendingBcpRequest();
+
+	RESUME_CANCEL_INTERRUPTS();
+
+	if (ret < 0)
+		TdsErrorContext->err_text = "EOF on TDS socket while fetching For Bulk Load Request";
+
+	if (internal_sp_started && before_lxid == MyProc->lxid && before_subtxn_id == GetCurrentSubTransactionId())
```
kuntalghosh, line 940: This is not checking pltsql_disable_internal_savepoint GUC, is that expected?

contrib/babelfishpg_tds/src/backend/tds/tdsbulkload.c
```diff
@@ -955,79 +1020,75 @@ ProcessBCPRequest(TDSRequest request)
...
+	/* If we Started an internal savepoint then release it. */
+	if (internal_sp_started && before_subtxn_id == GetCurrentSubTransactionId())
+	{
+		elog(DEBUG5, "TSQL TXN Release internal savepoint");
+		ReleaseCurrentSubTransaction();
+		CurrentResourceOwner = oldowner;
+		MemoryContextSwitchTo(oldcontext);
 	}
+	/* Unlikely case where Transaction is active but the savepoints do not match. */
+	else if (internal_sp_started && before_subtxn_id != GetCurrentSubTransactionId())
```
kuntalghosh: Then, let's put unlikely hint.

kuntalghosh (changes requested):

test/dotnet/input/InsertBulk/insertBulkErrors.txt
KushaalShroff: > we should test each error case without transaction, with transaction followed by commit and rollback

Thats already done

> we should also add tests for savepoints and rollback to savepoint, without error and with error

I will add tests for manual savepoints as well

> we should manually test both error cases where savepoint state is not as expected

That cannot be tested from InsertBulk Copy point of view, right?
Update: We will test with GDB and see the rollback/cleanup

> if supported, test with implicit transaction on/off, xact abort on/off

Let me check

> I am not clear how we are testing reset connection with pooling

These are the test cases:
https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/2887/files#diff-b1b7adfb8ac542c63531c143322672a0ec670b83ace04a84a53eec81e0e562c1R293
Also check PR description for more validation on these tests through logs

> how are we testing out of order packet error?

.NET driver sends attention to interrupt any Bulk Load Error in a rare case where the error has already been deducted but there are rows the driver has still not sent in the request. For example, This we encounter in the NOT NULL case where first row is NULL - the driver gets error and it sends an out of band attention packet for which it does not wait for a response to close the connection

test/dotnet/ExpectedOutput/insertBulkErrors.out
KushaalShroff: The tests to enable seq/enable index scan do the same check right? If we need it, going ahead in the next release I think we can do that as well

contrib/babelfishpg_tsql/src/pl_exec-2.c
KushaalShroff: No, TDS catches it and resets it. No need to expose offset via plugin_ptr

contrib/babelfishpg_tds/src/backend/tds/tdsbulkload.c
KushaalShroff: But the original error wont be relayed to customer, will that be a good Customer experience?

contrib/babelfishpg_tds/src/backend/tds/tdsbulkload.c
KushaalShroff: Spoken offline that this is not required and it would be ease of context to debug if it remains separate

contrib/babelfishpg_tds/src/backend/tds/tdsbulkload.c
KushaalShroff: Doing same thing:
```
void
pltsql_rollback_txn(void)
{
	PLTsqlRollbackTransaction(NULL, NULL, false);
	CommitTransactionCommand();
	StartTransactionCommand();
}

File: contrib/babelfishpg_tds/src/backend/tds/tdsbulkload.c (hunk 16 above)
Comment (KushaalShroff) [reply contrib/babelfishpg_tds/src/backend/tds/tdsbulkload.c:940]: Correct me if I am wrong, That was done for performance improvement factor and we are doing it here from a necessity angle

File: contrib/babelfishpg_tds/src/backend/tds/tdsbulkload.c (hunk 11 above)
Comment (KushaalShroff) [reply contrib/babelfishpg_tds/src/backend/tds/tdsbulkload.c]: What difference does it make though? For both cases of cleanup we need to hold interrupts so it will merely avoid redundancy in code lines

File: test/dotnet/input/InsertBulk/insertBulkErrors.txt (hunk 13 above)
Comment (KushaalShroff) [reply test/dotnet/input/InsertBulk/insertBulkErrors.txt:7]: There is a .net driver level option to useInternalTransaction: https://learn.microsoft.com/en-us/dotnet/api/system.data.sqlclient.sqlbulkcopyoptions?view=netframework-4.8.1&redirectedfrom=MSDN

But in that .net driver itself starts an explicit transaction 

And implicit transactions have no role to play for the Insert Bulk but I have added tests

File: contrib/babelfishpg_tds/src/backend/tds/tdsbulkload.c (hunk 2 above)
Comment (KushaalShroff) [reply contrib/babelfishpg_tds/src/backend/tds/tdsbulkload.c:943]: Please check TdsErrorContext->err_text, there is already context added to every log statement in Bulk Load

**Review by surendravishnoi — DISMISSED** [review]

File: contrib/babelfishpg_tds/src/backend/tds/tdsbulkload.c (hunk 11 above)
Comment (kuntalghosh) [reply contrib/babelfishpg_tds/src/backend/tds/tdsbulkload.c]: The cleanup function doesn't care whether you're pausing the interrupts or not. It's rather from where you're calling the function, it matters. In this case, since you're calling in catch block, you want to hold the interrupts. 
Yeah, the comment part looks redundant. Please ignore.

File: contrib/babelfishpg_tds/src/backend/tds/tdsbulkload.c [hunk 18]
```diff
@@ -1054,7 +1123,7 @@ ProcessBCPRequest(TDSRequest request)
 		pltsql_plugin_handler_ptr->stmt_needs_logging = false;
 		error_context_stack = plerrcontext;
 	}
-	offset = 0;
+	bcpOffset = 0;
```
kuntalghosh: It's wrong to reset it here. What if we hit any of the previous two errors. We need to reset it immediately after the while loop. 
We should move the while loop inside the try catch and reset bcpOffset in both catch and end of try.
I don't see a need of introducing the reset function in that case.

```
PG_TRY();
{
    while() {}
}
PG_CATCH();
{
    bcpOffset = 0;
}
PG_END_TRY();
bcpOffset = 0;
```

contrib/babelfishpg_tds/src/backend/tds/tdsbulkload.c
kuntalghosh: I think there's some misunderstanding. I probably mentioned that context can be set beforehand.
```
volatile char *tmpContext = NULL;

tmpContext = "foo";
SetBulkLoadRowData();

tmpContext = "bar";
pltsql_plugin_handler_ptr->bulk_load_callback();

CATCH()
TdsErrorContext->err_text = tmpContext;
```

kuntalghosh (changes requested):

kuntalghosh (approved):
Let's address the pending review comments in the next PR.

surendravishnoi (approved):

contrib/babelfishpg_tds/src/backend/tds/tdsbulkload.c
```diff
@@ -955,79 +1026,79 @@ ProcessBCPRequest(TDSRequest request)
...
-			list_free_deep(req->rowData);
-			req->rowData = NIL;
 			if (values)
 				pfree(values);
 			if (nulls)
 				pfree(nulls);
 		}
+		PG_CATCH();
+		{
+			if (TDS_DEBUG_ENABLED(TDS_DEBUG2))
```
tanscorpio7, line 1073: We should hold interrupts before doing any logging in catch blocks.

PR #2893, opened 2024-08-29
Fix the index creation error on table created through SELECT INTO 

pranavJ23:
### Description
Issue - 
If a table is formed via SELECT INTO command then the column name does not get converted to lowercase.
Due to this, indexation over that column fails which is the issue we want to resolve.
This happens as the SELECT INTO command does not convert the column name in lowercase(save it as it is) whereas the index creation command search for the column name(over which indexation is done) in lower case.
cherry picked from - https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/2781
### Issues Resolved
Conversion of column name to lower case(if column name contains uppercase character/s) in Select Into query.

Note: if the table is pre-existing then user needs to drop the table and create it again to enable this feature.
### Approach
1. Down-case the column name using downcase_identifier.
2. check via filter that resjunk should be false and the column name should not be NULL.

### Task
[BABEL-4926]
Signed-off-by: Pranav Jain pranavjc@amazon.com

1. Null case i.e if the column name is null.
3. Case when multiple column names are given in SELECT INTO query
4. Case when column name is already in lowercase 
5. Case when column name is passed with a different name in SELECT INTO statement(using as).
6. Case when column is passed via some function.
7. Case when column/s is/are passed via *(select all) instead of specifying names.
8. Case when column with special letters like $,& is passed into square brackets.
9. Case when identity function is used with select into statement.
10. Added mvu test when the select into is executed in prepare step and the create index is executed in verify step
11. Case when column is modified via ALTER TABLE or ALTER COLUMN 
12. Creating UNIQUE INDEX on the column after SELECT INTO
13. Dropping column via ALTER TABLE...DROP COLUMN after SELECT INTO
14. Case when table is passed to SELECT INTO via Table-Returning Function 
15. Multibyte Column Name (Chinese)
16. Multibyte Column Name (Chinese) more then 64 characters
17. Table with multi-part names
18. Table with multi-part names and column name length > 64

### Check List
- [Pranav Jain] Commits are signed per the DCO using --signoff 


tanscorpio7 (approved):

shardgupta (approved):

andrevb-croesus:
When can we expect this bugfix to be available on AWS RDS instances ? 
I understand this has been merged in 4_X_DEV branch but what's the timeline until this usually gets pushed to RDS ?

shardgupta:
> When can we expect this bugfix to be available on AWS RDS instances ? I understand this has been merged in 4_X_DEV branch but what's the timeline until this usually gets pushed to RDS ?

This will be available on RDS in Jan 2025 tentatively.

PR #2899, opened 2024-08-30
Support Cross-database references in views

rishabhtanwar29:
### Description
Support execution of views which references objects (tables/views/functions) from
across the databases. Here we are talking about Babelfish logical database (T-SQL database)
which is different from a physical Postgres database.
To support this, perform permission checks for cross database objects using session user (login)
instead of current_user (user of current T-SQL database). The reason login can be used for
permission check is since login is member of all it’s users, so it inherits all their permissions so it
will be able execute any cross database objects owned by its users.

This commit handles functions and tables/views separately for cross database permission
checks. For functions/procedures, a new hook `ExecFuncProc_AclCheck_hook` and for tables/views
existing `ExecutorStart_hook` will be used to decide whether to use session user or current_user
for permission check depending upon whether the object is from same or different database.
We will be using `is_schema_from_db` function to identify if the object is from different database
which performs a lookup into `babelfish_namespace_ext` catalog table which can be expensive as will
be doing it pretty frequently. So, added this table into SYSCACHE for better performance.
Tables/views permissions are handled slightly different than functions as we do not blindly want to check
the permissions against session user (current login) since permissions of RTEs inside a view are checked
against that view's owner which can very well be a user of some different database. So if we blindly check
permission against session user instead of view's owner then it would break view's ownership chaining.
Instead, we will replace `checkAsUser` with it's corresponding mapped login if present and only in cases
where `checkAsUser` is not set, we will replace it with session user (login). We are using login to allow cross
database queries since login can access all its objects across the databases. Getting mapped login to a
user require lookup into sys.babelfish_authid_user_ext catalog table using its primary key column (rolname) so
added this table is also into SYSCACHE.

Additionally, remove previous code to globally set current user to session user since newer logic
takes care of the permission check now.

Task: BABEL-5206

Engine PR: https://github.com/babelfish-for-postgresql/postgresql_modified_for_babelfish/pull/434

* **Use case based -**
BABEL-CROSS-DB-vu-*

* **Boundary conditions -**

* **Arbitrary inputs -**

* **Negative test cases -**

* **Performance tests -**

* **Tooling impact -**

* **Client tests -**

### Check List


contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -829,6 +836,36 @@ pltsql_GetNewTempOidWithIndex(Relation relation, Oid indexId, AttrNumber oidcolu
...
+			 */
+			if (nspname != NULL && !is_shared_schema(nspname) &&
+				!is_schema_from_db(schema_id, get_cur_db_id()))
+				userid = GetSessionUserId();
+			if (nspname)
+				pfree(nspname);
+		}
+	}
+
+	return object_aclcheck(ProcedureRelationId, funcid, userid, ACL_EXECUTE);
```
Deepesh125, line 880: we should also call prev hook

test/JDBC/input/ownership/BABEL-CROSS-DB-vu-prepare.mix
```diff
@@ -61,5 +67,112 @@ GO
...
+GO
+
+-- View containing cross-db table and function on which login babel_cross_db_vu_prepare_l1 has privilege
+CREATE VIEW babel_cross_db_vu_prepare_s1.babel_cross_db_vu_prepare_v1 AS
+    SELECT x.id,
+        babel_cross_db_vu_prepare_db2.babel_cross_db_vu_prepare_s2.babel_cross_db_vu_prepare_f2(x.id),
+        x.name AS t1_name,
+        y.name AS t2_name
+    FROM babel_cross_db_vu_prepare_s1.babel_cross_db_vu_prepare_t1 x INNER JOIN
+    babel_cross_db_vu_prepare_db2.babel_cross_db_vu_prepare_s2.babel_cross_db_vu_prepare_t2 y
```
Deepesh125: can we add similar test where session user dont have permission on either of joined tables

Deepesh125 (dismissed):

test/JDBC/input/ownership/BABEL-CROSS-DB-vu-prepare.mix
rishabhtanwar29: Done!

test/JDBC/input/ownership/BABEL-CROSS-DB-vu-verify.mix
```diff
@@ -246,3 +246,54 @@ GO
...
+SELECT * FROM babel_cross_db_vu_prepare_s1.babel_cross_db_vu_prepare_v4;
+GO
+
+-- Execute procedure from db2
+EXEC babel_cross_db_vu_prepare_db2.babel_cross_db_vu_prepare_s2.babel_cross_db_vu_prepare_p2;
+GO
+
+-- Execute procedure in db1 which accesses cross database objects
+EXEC babel_cross_db_vu_prepare_s1.babel_cross_db_vu_prepare_p1;
+GO
```
HarshLunagariya, line 317: Let's call this procedure from a different db.

test/JDBC/expected/BABEL-CROSS-DB-vu-prepare.out
```diff
@@ -61,5 +67,130 @@ GO
 CREATE TABLE dbo.babel_cross_db_vu_prepare_db1_t3 (id int identity, a int);
 GO
```
HarshLunagariya: Let's add nested tests (with views, procedures, function calls) along with cross-db calls internally.

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -866,6 +905,39 @@ pltsql_ExecutorStart(QueryDesc *queryDesc, int eflags)
...
+					 */
+					if (nspname != NULL && !is_shared_schema(nspname) &&
+						!is_schema_from_db(schema_id, get_cur_db_id()))
+						perminfo->checkAsUser = GetSessionUserId();
+					if (nspname)
+						pfree(nspname);
+				}
+			}
+		}
+	}
```
HarshLunagariya: After this changes, Can we remove any redundant logic from `pl_exec.c` or `pl_exec-2.c`?

rishabhtanwar29: Yep, removed it.

test/JDBC/expected/BABEL-CROSS-DB-vu-prepare.out
rishabhtanwar29: Done!

test/JDBC/input/ownership/BABEL-CROSS-DB-vu-verify.mix
rishabhtanwar29: Done!

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -866,6 +905,58 @@ pltsql_ExecutorStart(QueryDesc *queryDesc, int eflags)
...
+					 * login if present and only in cases where checkAsUser is not set, we will replace it
+					 * with session user (login). We are using login to allow cross database queries since login
+					 * can access all its objects across the databases.
+					 */
+					if (nspname != NULL && !is_shared_schema(nspname))
+					{
+						if (OidIsValid(perminfo->checkAsUser))
+						{
+							Oid loginId = get_login_for_user(perminfo->checkAsUser, nspname);
+							if (OidIsValid(loginId))
```
KushaalShroff, line 947: In case of invalid loginId I think we should throw an error like "could not find mapped login for db: %s"

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -866,6 +905,58 @@ pltsql_ExecutorStart(QueryDesc *queryDesc, int eflags)
...
+					 * ownership chaining. Instead, we will replace checkAsUser with it's corresponding mapped
+					 * login if present and only in cases where checkAsUser is not set, we will replace it
+					 * with session user (login). We are using login to allow cross database queries since login
+					 * can access all its objects across the databases.
+					 */
+					if (nspname != NULL && !is_shared_schema(nspname))
+					{
+						if (OidIsValid(perminfo->checkAsUser))
+						{
+							Oid loginId = get_login_for_user(perminfo->checkAsUser, nspname);
```
KushaalShroff, line 946: get_login_for_user will give invalid for guests, lets add tests for guest as owner if possible

test/JDBC/singledb_jdbc_schedule
```diff
@@ -6,6 +6,9 @@ ignore#!#test_db_collation-vu-prepare
 ignore#!#test_db_collation-vu-verify
 ignore#!#test_db_collation-vu-cleanup
 ignore#!#BABEL-CROSS-DB
+ignore#!#BABEL-CROSS-DB-vu-prepare
```
KushaalShroff: Lets add test case for single db as well between master and db1

contrib/babelfishpg_tsql/src/pl_exec.c
```diff
@@ -4639,23 +4638,11 @@ exec_stmt_execsql(PLtsql_execstate *estate,
 
 	if (stmt->is_cross_db)
 	{
-		char	   *login = GetUserNameFromId(GetSessionUserId(), false);
-		char	   *user = get_user_for_database(stmt->db_name);
```
KushaalShroff, line 4643: ++ for this change.
We had Identified that this function takes 30% of the time while executing a view which selects from a table

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -860,59 +851,35 @@ get_authid_login_ext_idx_oid(void)
...
 	{
-		Datum		datum;
-		bool		isnull;
-		TupleDesc	dsc;
+		BpChar type = ((Form_authid_user_ext) GETSTRUCT(tuple))->type;
+		char *type_str = bpchar_to_cstring(&type);
 
-		dsc = RelationGetDescr(relation);
-		datum = heap_getattr(tuple, USER_EXT_TYPE + 1, dsc, &isnull);
-		if (!isnull)
```
KushaalShroff, line 902: We dont need to check for nullable type? Is null even possible?

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -3095,34 +3043,68 @@ guest_role_exists_for_db(const char *dbname)
...
+		loginId = get_role_oid((DatumGetName(datum)->data), true);
 
-	return role_exists;
+		if (!OidIsValid(loginId))
+		{
+			char *orig_username = TextDatumGetCString(SysCacheGetAttr(AUTHIDUSEREXTROLENAME,
+								tuple, Anum_bbf_authid_user_ext_orig_username, &isnull));
+
+			Assert(!isnull);
+			if (strlen(orig_username) == 3 && pg_strcasecmp(orig_username, "dbo") == 0)
```
KushaalShroff, line 3097: What about guest?

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -3095,34 +3043,68 @@ guest_role_exists_for_db(const char *dbname)
...
+			{
+				int16 dbid = get_dbid_from_physical_schema_name(physical_schema_name, true);
+
+				if (DbidIsValid(dbid))
```
KushaalShroff: Lets add unlikely error for invalid error case

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -841,6 +848,38 @@ pltsql_GetNewTempOidWithIndex(Relation relation, Oid indexId, AttrNumber oidcolu
...
+			 * to allow cross database access.
+			 */
+			if (nspname != NULL && !is_shared_schema(nspname) &&
+				!is_schema_from_db(schema_id, get_cur_db_id()))
+				userid = GetSessionUserId();
+			if (nspname)
+				pfree(nspname);
+		}
+	}
+	else if (prev_ExecFuncProc_AclCheck_hook)
```
KushaalShroff, line 877: If we have system objects which have not been granted to PUBLIC, what happens in that case?

What if I execute this function with 4 part name, hypothetically is PG language (or C), it gets executed in PG dialect, then we will be using the PG acl check?

KushaalShroff (changes requested):

contrib/babelfishpg_tsql/src/catalog.c
rishabhtanwar29: This is a non-nullable column and can't be NULL so NULL check was redundant.

contrib/babelfishpg_tsql/src/catalog.c
rishabhtanwar29: There is one to many relation between guest user and login so we can't find the mapped login for user. So we will not handle guest user here and instead send invalid dbid.

contrib/babelfishpg_tsql/src/hooks.c
rishabhtanwar29: We skip user switching the user for shared schemas so all the system objects in sys,information_schema,pg_catalog etc will not be affected with the above change anyways. There are few  system objects in master_dbo schema but all of them are publicly executable.
Yes, any function/procedure which gets executed in PG dialect will use PG acl check.

contrib/babelfishpg_tsql/src/hooks.c
rishabhtanwar29: Right, as I mentioned in earlier comment as well that since guest will have one to many relation with logins, we can't determine the right login for guest user so `get_login_for_user` will return InvalidOid for guest. This means that we will fallback to existing behavior (not allowing cross-db access) for guest owned relations and not change checkAsUser for them.

contrib/babelfishpg_tsql/src/hooks.c
rishabhtanwar29: As mentioned in comment above, we will fallback to existing behavior and not change `checkAsUser` instead of raising error since in this way we will atleast be allowing non-cross db accesses for guest owned objects.

contrib/babelfishpg_tsql/src/catalog.c
rishabhtanwar29: Added missingOk = false instead.

test/JDBC/singledb_jdbc_schedule
rishabhtanwar29: Updated test so that it works for both single and multi db modes.

contrib/babelfishpg_tsql/src/hooks.c
KushaalShroff: Can there be any scenario now, and in future where the current user we fall back to, can have access and does not throw error?

contrib/babelfishpg_tsql/src/hooks.c
KushaalShroff: How are we thinking of matching TSQL behaviour for guest user for cross database owned objects?

contrib/babelfishpg_tsql/src/hooks.c
KushaalShroff: Lets think about it from this perspective: there are few queries that implicitly can execute inside a PG dialect block, will the dialect check in this case fallback to using PGs ACL Check?

contrib/babelfishpg_tsql/src/hooks.c
rishabhtanwar29: But in case the current user has permission then it should be able to access the object so it seems more intuitive to not throw error.

contrib/babelfishpg_tsql/src/hooks.c
rishabhtanwar29: There are few ways to make it work for guest user, one is to switch to login which does not have a mapped user in respective logical database which would require an extra catalog lookup into user_ext table another way is to create a generic login which does not have a user mapped to any database and switch to that login for guest owned objects. But cross-db access for guest owned objects do not seem a priority right now so that support can be added later as current implementation does not create a one way door.

contrib/babelfishpg_tsql/src/hooks.c
rishabhtanwar29: Right, I guess there can be a few such scenarios but currently one scenario I can think of is when a PG function in one logical db refers objects from other logical db. In that case the logic for switching the user will be skipped as it will happen in PG dialect. I tested such scenario and figured out that checking for TDS_client is better than dialect check. Updated the code accordingly and added a test as well.

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -841,6 +848,38 @@ pltsql_GetNewTempOidWithIndex(Relation relation, Oid indexId, AttrNumber oidcolu
...
+			char *nspname = get_namespace_name(schema_id);
+
+			/*
+			 * Check if function's schema is from a different logical database and
+			 * it is not a shared schema. If yes, then set userid to session user
+			 * to allow cross database access.
+			 */
+			if (nspname != NULL && !is_shared_schema(nspname) &&
+				!is_schema_from_db(schema_id, get_cur_db_id()))
+				userid = GetSessionUserId();
```
HarshLunagariya, line 872: We will be switching the userid even for PG created schema as well? Should we restrict that or not?

Before this change:
```
- User creates schema from PG endpoint named as db1_sch1 and a table t1 in it.
- If it tries to access it from TDS endpoint
- Normal Query: USE master; SELECT * FROM db1.sch1.t1; -> Allowed
- Inside Func/Proc/View: USE master; CREATE VIEW v1 AS SELECT * FROM db1.sch1.t1; SELECT * FROM v1; -> permission denied 
```

After this change:
```
All first 3 pointers will behave but the inside func/proc/view it will not throw permission denied now.
```
To me it looks like we are matching the normal query behaviour. Kindly verify once.

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -866,6 +905,58 @@ pltsql_ExecutorStart(QueryDesc *queryDesc, int eflags)
...
+					 * Check if relation's schema is valid and is not a shared schema. If yes,
+					 * then replace checkAsUser to its mapped login if present otherwise replace
+					 * with session user (current login).
+					 * We do not blindly want to check the permissions against session user (current login)
+					 * since permissions of RTEs inside a view are checked against that view's owner
+					 * which can very well be a user of some different database. So if we blindly check
+					 * permission against session user instead of view's owner thenit would break view's
+					 * ownership chaining. Instead, we will replace checkAsUser with it's corresponding mapped
+					 * login if present and only in cases where checkAsUser is not set, we will replace it
+					 * with session user (login). We are using login to allow cross database queries since login
```
HarshLunagariya: Let's not mention that ```would break view's ownership chaining``` since from T-SQL POV it is not really ownership chaining to avoid confusion.

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -866,6 +905,58 @@ pltsql_ExecutorStart(QueryDesc *queryDesc, int eflags)
...
+					 * can access all its objects across the databases.
+					 */
+					if (nspname != NULL && !is_shared_schema(nspname))
```
HarshLunagariya, line 942: Earlier comment about PG endpoint schema applies here as well.

test/JDBC/expected/single_db/TestSpatialPoint-vu-verify.out
```diff
@@ -3469,9 +3469,9 @@ SET @sql =
 -- Execute the dynamic SQL
 EXEC sp_executesql @sql;
 GO
-~~ERROR (Code: 911)~~
+~~ERROR (Code: 33557097)~~
 
-~~ERROR (Message: database "db2" does not exist)~~
+~~ERROR (Message: database "db2" does not exist. Make sure that the name is entered correctly.)~~
```
HarshLunagariya, line 3474: Why is it being reflected here?

test/JDBC/upgrade/13_9/schedule
```diff
@@ -97,7 +97,6 @@ BABEL-889
 BABEL-APPLOCK
 babel_char
 BABEL-CHECK-CONSTRAINT-before-14_6
-BABEL-CROSS-DB
```
HarshLunagariya, line 100: Any reasons to particularly removing it from 13.x schedules only?

contrib/babelfishpg_tsql/src/hooks.c
rishabhtanwar29: Yep we are matching the normal query behavior which logically looks right. Actually it is the main motive of this PR to support cross db queries even in a view.

contrib/babelfishpg_tsql/src/hooks.c
rishabhtanwar29: Updated!

contrib/babelfishpg_tsql/src/hooks.c
rishabhtanwar29: Replied above.

test/JDBC/expected/single_db/TestSpatialPoint-vu-verify.out
rishabhtanwar29: Previously `get_user_for_database` in `exec_stmt_execsql` used to throw database does not exist error but now I have removed that piece of code so we are getting the same error during schema rewrite later.

test/JDBC/upgrade/13_9/schedule
rishabhtanwar29: Yeah cause cross database references was not supported in 13.x so prepare file has become incompatible for those versions so I removed the test from 13.x schedules files since we anyway don't support a lot of things their.

HarshLunagariya (approved):

KushaalShroff (approved):

PR #2907, opened 2024-09-03
Support Securityadmin fixed server role

anju15bharti:
### Description

Added support for new fixed server role securityadmin

This role has following privileges with it:

1) Members of the securityadmin fixed server role can manage logins and their properties. 
2) They can GRANT, DENY, and REVOKE server-level permissions. 
3) securityadmin can also GRANT, DENY, and REVOKE database-level permissions if they have access to a database.

### Issues Resolved

* **Use case based -**

* **Boundary conditions -**

* **Arbitrary inputs -**

* **Negative test cases -**

* **Performance tests -**

* **Tooling impact -**

* **Client tests -**

### Check List


contrib/babelfishpg_tds/src/backend/tds/tdsutils.c
```diff
@@ -1229,18 +1231,18 @@ handle_grant_role(GrantRoleStmt *grant_stmt)
 			continue;
 
 		roleid = get_role_oid(rolename, false);
-		if (OidIsValid(roleid) && roleid == bbf_role_admin_oid)
```
HarshLunagariya, line 1232: This means we only want to allow `GRANT securityadmin TO ...` and `GRANT ... TO securityadmin` when we are in bbf_role_admin's context. Right?

contrib/babelfishpg_tsql/sql/ownership.sql
```diff
@@ -285,11 +285,14 @@ BEGIN
 		RAISE E'Could not initialize babelfish with given role name: % is not the DB owner of current database.', sa_name;
 	END IF;
 
+	EXECUTE format('CREATE ROLE securityadmin CREATEROLE INHERIT PASSWORD NULL');
```
HarshLunagariya, line 288: Do we need CREATEROLE privileges? Because internally we are doing it via the privileges of bbf_role_admin. Right now, I don't see a point of allowing CREATEROLE property because even if we make any login member of it, then also this property can't be inherited to it's members.

contrib/babelfishpg_tsql/sql/ownership.sql
```diff
@@ -285,11 +285,14 @@ BEGIN
...
+	EXECUTE format('GRANT CONNECT ON DATABASE %s TO securityadmin WITH GRANT OPTION', CURRENT_DATABASE());
```
HarshLunagariya: 1. Why granting connect permission is required? 
2. Acc. ideal T-SQL behaviour, Does all the members of securityadmin can login to database always -- even if some login is disabled? Otherwise it doesn't make sense to add this permission.

contrib/babelfishpg_tsql/sql/ownership.sql
```diff
@@ -365,7 +371,8 @@ CAST(CASE WHEN Ext.type = 'R' THEN NULL ELSE Ext.credential_id END AS INT) AS cr
 CAST(CASE WHEN Ext.type = 'R' THEN 1 ELSE Ext.owning_principal_id END AS INT) AS owning_principal_id,
 CAST(CASE WHEN Ext.type = 'R' THEN 1 ELSE Ext.is_fixed_role END AS sys.BIT) AS is_fixed_role
 FROM pg_catalog.pg_roles AS Base INNER JOIN sys.babelfish_authid_login_ext AS Ext ON Base.rolname = Ext.rolname
-WHERE (pg_has_role(suser_id(), 'sysadmin'::TEXT, 'MEMBER')
+WHERE (bbf_is_member_of_role_nosuper(suser_id(), 'sysadmin') 
+  OR bbf_is_member_of_role_nosuper(suser_id(), 'securityadmin')
```
HarshLunagariya: Should we use pg_has_role? Because anyway superuser will have all privileges/memberships.

contrib/babelfishpg_tsql/sql/ownership.sql
```diff
@@ -501,8 +508,20 @@ CAST(Ext.orig_loginname AS sys.nvarchar(128)) AS name,
...
+SELECT
+CAST(Base.oid As INT) AS principal_id,
+CAST(CAST(Base.oid as INT) as sys.varbinary(85)) AS sid,
+CAST(Ext.orig_loginname AS sys.nvarchar(128)) AS name,
+CAST('SERVER ROLE' AS sys.nvarchar(128)) AS type,
+CAST ('GRANT OR DENY' as sys.nvarchar(128)) as usage
+FROM pg_catalog.pg_roles AS Base INNER JOIN sys.babelfish_authid_login_ext AS Ext ON Base.rolname = Ext.rolname
+WHERE Ext.orig_loginname = 'securityadmin'
+AND Ext.type = 'R' AND
+bbf_is_member_of_role_nosuper(sys.suser_id(), 'securityadmin');
```
HarshLunagariya: Do we really need a separate union?

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--4.3.0--4.4.0.sql
```diff
@@ -11,6 +11,224 @@ SELECT set_config('search_path', 'sys, '||current_setting('search_path'), false)
...
+do
+LANGUAGE plpgsql
+$$
+DECLARE securityadmin TEXT;
+BEGIN
+    IF EXISTS (
+               SELECT FROM pg_catalog.pg_roles
+               WHERE  rolname = 'securityadmin') 
+        THEN
+            RAISE EXCEPTION 'Role "securityadmin" already exists.';
```
HarshLunagariya, line 82: This exception will fail Latest-latest upgrade.

contrib/babelfishpg_tsql/src/backend_parser/gram-tsql-rule.y
```diff
@@ -3145,11 +3158,24 @@ tsql_alter_server_role:
...
+												parser_errposition(@4)));
+			}
+			else if (strcmp($4, "sysadmin") != 0 && strcmp($4, "securityadmin") != 0)
+			{
 				ereport(ERROR, (errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
-                               errmsg("only sysadmin role is supported in ALTER SERVER ROLE statement"),
-                               parser_errposition(@4)));
+						errmsg("Only fixed server role is supported in ALTER SERVER ROLE statement"),
+									parser_errposition(@4)));
+			}
```
HarshLunagariya: Let's avoid this duplicate code.

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -2864,7 +2864,12 @@ bbf_ProcessUtility(PlannedStmt *pstmt,
 
 					if (islogin)
 					{
-						if (!has_privs_of_role(GetSessionUserId(), get_role_oid("sysadmin", false)))
+						/*
+						 * Check if the current login has privileges to create
+						 * login.
+						 */
+						if (!has_privs_of_role(GetSessionUserId(), get_role_oid("sysadmin", false)) &&
+																	!has_privs_of_role(GetSessionUserId(), get_securityadmin_oid()))
```
HarshLunagariya: Please maintain 80 char limit for better readability.

test/JDBC/expected/BABEL-2403.out
```diff
@@ -81,6 +81,8 @@ name#!#pg_catalog#!#rolname#!#{"Rule": "<rolname> in babelfish_authid_login_ext
 text#!#sys#!#name#!#{"Rule": "<default_database_name> in babelfish_authid_login_ext must also exist in babelfish_sysdatabases"}
 name#!#pg_catalog#!#rolname#!#{"Rule": "<rolname> in babelfish_authid_login_ext must also exist in pg_authid"}
 text#!#sys#!#name#!#{"Rule": "<default_database_name> in babelfish_authid_login_ext must also exist in babelfish_sysdatabases"}
+name#!#pg_catalog#!#rolname#!#{"Rule": "<rolname> in babelfish_authid_login_ext must also exist in pg_authid"}
+text#!#sys#!#name#!#{"Rule": "<default_database_name> in babelfish_authid_login_ext must also exist in babelfish_sysdatabases"}
```
HarshLunagariya, line 85: Why are these 2 entries appearing?

contrib/babelfishpg_tsql/src/rolecmds.c
```diff
@@ -2430,3 +2455,25 @@ remove_createrole_from_logins(PG_FUNCTION_ARGS)
...
+	char	*dc_role;
+	bool	result;
+
+	if (PG_ARGISNULL(0) || PG_ARGISNULL(1))
+		PG_RETURN_NULL();
+
+	member = PG_GETARG_OID(0);
+	rolename = text_to_cstring(PG_GETARG_TEXT_P(1));
+	dc_role = downcase_identifier(rolename, strlen(rolename), false, false);
+	role = get_role_oid(dc_role, false);
```
HarshLunagariya: Can we pass the oid for rolename as well? To avoid down-casing and fetching the oid here.

contrib/babelfishpg_tsql/src/rolecmds.c
```diff
@@ -1628,8 +1649,12 @@ check_alter_server_stmt(GrantRoleStmt *stmt)
...
 				 errmsg("%s is not a login", grantee_name)));
 
-	/* only sysadmin role is assumed below */
-	if (!has_privs_of_role(GetSessionUserId(), sysadmin))
+	/* 
+	 * check if it has sysadmin privileges or
+	 * if server role is securityadmin and it has privileges of securityadmin
+	 */
+	if (!has_privs_of_role(GetSessionUserId(), sysadmin) && ((strcmp(granted_name, "securityadmin") != 0)
+										|| !has_privs_of_role(GetSessionUserId(), securityadmin_oid)))
```
HarshLunagariya: What happens when we alter try to alter sysadmin role? For that case as well we should be throwing this kind of error. But from the current implementation, I don't think we will be throwing this error. Can you please validate and modify this check to include sysadmin check as well if required.

HarshLunagariya (changes requested):

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--4.3.0--4.4.0.sql
```diff
@@ -11,6 +11,224 @@ SELECT set_config('search_path', 'sys, '||current_setting('search_path'), false)
  * final behaviour.
  */
 
+do
```
tanscorpio7: casing

contrib/babelfishpg_tsql/src/backend_parser/gram-tsql-rule.y
```diff
@@ -3127,12 +3127,25 @@ tsql_alter_server_role:
...
 			
+			if (strcmp($4, "serveradmin") == 0
+				|| strcmp($4, "setupadmin") == 0
+				|| strcmp($4, "processadmin") == 0
+				|| strcmp($4, "dbcreator") == 0
+				|| strcmp($4, "diskadmin") == 0
+				|| strcmp($4, "bulkadmin") == 0) 
+			{
+					ereport(ERROR, (errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
+							errmsg("Server role '%s' is currently not supported in Babelfish", $4),
```
tanscorpio7: Server role --> Fixed server role ?

contrib/babelfishpg_tsql/src/backend_parser/gram-tsql-rule.y
```diff
@@ -3127,12 +3127,25 @@ tsql_alter_server_role:
...
+												parser_errposition(@4)));
+			}
+			else if (strcmp($4, "sysadmin") != 0 && strcmp($4, "securityadmin") != 0)
```
tanscorpio7: Can we move the check for supported server roles to epilogue file ?
This would make it reusable and we could also use MACROs instead of hard coded strings.

contrib/babelfishpg_tsql/src/rolecmds.c
```diff
@@ -615,6 +618,16 @@ get_bbf_role_admin_oid(void)
 	return bbf_admin_oid;
 }
 
+
+/* Returns OID of securityadmin server role */
+Oid
+get_securityadmin_oid(void)
+{
+	if (!OidIsValid(securityadmin_oid))
+		securityadmin_oid = get_role_oid("securityadmin", false);
```
tanscorpio7: Let us try to use MACRO everywhere

contrib/babelfishpg_tsql/src/rolecmds.c
```diff
@@ -1574,20 +1587,26 @@ bool
...
-														 * role */
+		/* only supported server roles */
+		if (is_sysadmin || (rolename_len == 13 && strncmp(spec->rolename, "securityadmin", 13) == 0))
 			return true;
 	}
-	/* has one and only one grantee  */
-	if (list_length(stmt->grantee_roles) != 1)
+	/* if granted role is sysadmin and has one and only one grantee  */
+	if (is_sysadmin && list_length(stmt->grantee_roles) != 1)
 		return false;
```
tanscorpio7: We can remove these lines. We will anyway return false below.

contrib/babelfishpg_tsql/src/rolecmds.c
```diff
@@ -1574,20 +1587,26 @@ bool
...
 		RoleSpec   *spec = (RoleSpec *) linitial(stmt->granted_roles);
+		int     	rolename_len = strlen(spec->rolename);
+
+		if (rolename_len == 8 && strncmp(spec->rolename, "sysadmin", 8) == 0)
+			is_sysadmin = true;
 
-		if (strcmp(spec->rolename, "sysadmin") == 0)	/* only supported server
-														 * role */
+		/* only supported server roles */
+		if (is_sysadmin || (rolename_len == 13 && strncmp(spec->rolename, "securityadmin", 13) == 0))
```
tanscorpio7: Why are we using strncmp() ? I think we can simply use strcmp().

contrib/babelfishpg_tsql/src/rolecmds.c
```diff
@@ -1604,9 +1623,11 @@ check_alter_server_stmt(GrantRoleStmt *stmt)
 	CatCList   *memlist;
 	Oid			sysadmin;
 	char	   *db_name;
+	Oid       	securityadmin_oid;
 
 	spec = (RoleSpec *) linitial(stmt->grantee_roles);
 	sysadmin = get_role_oid("sysadmin", false);
+	securityadmin_oid = get_securityadmin_oid();
```
tanscorpio7: We can directly call get_securityadmin_oid() when needed instead of keeping a local copy.

tanscorpio7 (changes requested):

contrib/babelfishpg_tsql/sql/ownership.sql
tanscorpio7: +1

contrib/babelfishpg_tsql/sql/ownership.sql
tanscorpio7: +1

anju15bharti: Right, but isn't it better to give the permission to the required role only?

contrib/babelfishpg_tsql/src/backend_parser/gram-tsql-rule.y
anju15bharti: Since we only support fixed server, this means the same but will make it more specific if required.

test/JDBC/expected/BABEL-2403.out
anju15bharti: Because we have added one additonal login for securityadmin in the catalog.

contrib/babelfishpg_tsql/src/rolecmds.c
anju15bharti: Yes it does include all the required check.

contrib/babelfishpg_tsql/src/rolecmds.c
anju15bharti: Yes we can but since it's existing line, I didn't prefer to remove it.

contrib/babelfishpg_tsql/src/rolecmds.c
anju15bharti: Keeping in mind that the alter role stmt also goes through this path, we should keep this check with strncmp instead strcmp since the rolename can be something different.

contrib/babelfishpg_tds/src/backend/tds/tdsutils.c
anju15bharti: Right or to be more specific, we restrict grant/revoke from pg side.

contrib/babelfishpg_tsql/sql/ownership.sql
anju15bharti: But a role can only grant this attribute if it itself has that attribute, hence we need to keep this attribute same as sysadmin.

contrib/babelfishpg_tsql/sql/ownership.sql
anju15bharti: Yes it can be optimized.

contrib/babelfishpg_tsql/src/rolecmds.c
tanscorpio7:
```
if (strcmp(spec->rolename, BABELFISH_SYSADMIN) == 0 || strcmp(spec->rolename, BABELFISH_SECURITYADMIN) == 0)
```
I meant using this check instead. Just to improve readability.

contrib/babelfishpg_tsql/sql/ownership.sql
tanscorpio7: I think we will get the same behaviour with `pg_has_role`.

contrib/babelfishpg_tsql/sql/ownership.sql
anju15bharti: It's to grant connect permission on the babelfish physical database

contrib/babelfishpg_tsql/sql/ownership.sql
anju15bharti: will get same behaviour but the check will pass since it's superuser but here we don't want to check this, we want to check if login is member of sysadmin/securityadmin or not which will get not actually get tested since it will satisfy the superuser condition first.

contrib/babelfishpg_tsql/src/rolecmds.c
anju15bharti: But strncmp is more preferrable over strcmp since strcmp is more vulnerable compared to strncmp.

contrib/babelfishpg_tsql/src/rolecmds.c
```diff
@@ -1591,20 +1604,26 @@ bool
...
 
+	bool is_sysadmin = 0;
+
 	if (list_length(stmt->granted_roles) == 1)
 	{
 		RoleSpec   *spec = (RoleSpec *) linitial(stmt->granted_roles);
+		int     	rolename_len = strlen(spec->rolename);
+
+		if (rolename_len == 8 && strncmp(spec->rolename, BABELFISH_SYSADMIN, 8) == 0)
+			is_sysadmin = true;
```
tanscorpio7: We can return from here itslelf, no need for using `is_sysadmin`.

contrib/babelfishpg_tsql/src/rolecmds.c
tanscorpio7: I'd vote for removing it. Since we really want to maintain readability. Plus there would be lots of changes coming around this code soon.

contrib/babelfishpg_tsql/src/rolecmds.c
```diff
@@ -1645,8 +1664,12 @@ check_alter_server_stmt(GrantRoleStmt *stmt)
...
 				 errmsg("%s is not a login", grantee_name)));
 
-	/* only sysadmin role is assumed below */
-	if (!has_privs_of_role(GetSessionUserId(), sysadmin))
+	/* 
+	 * check if it has sysadmin privileges or
+	 * if server role is securityadmin and it has privileges of securityadmin
+	 */
+	if (!has_privs_of_role(GetSessionUserId(), sysadmin) && ((strcmp(granted_name, BABELFISH_SECURITYADMIN) != 0)
+										|| !has_privs_of_role(GetSessionUserId(), get_securityadmin_oid())))
```
tanscorpio7, line 1734: 2nd condition is a bit confusing, can we refactor it a bit ?

contrib/babelfishpg_tsql/sql/ownership.sql
tanscorpio7: Why do we need connect to babelfish physical database permission ?

contrib/babelfishpg_tsql/sql/ownership.sql
tanscorpio7: But we are internally switching to bbf_role_admin which already has create role. No ?

contrib/babelfishpg_tsql/sql/ownership.sql
```diff
@@ -285,11 +285,14 @@ BEGIN
...
 	END IF;
 
+	EXECUTE format('CREATE ROLE securityadmin CREATEROLE INHERIT PASSWORD NULL');
+	EXECUTE format('GRANT CONNECT ON DATABASE %s TO securityadmin WITH GRANT OPTION', CURRENT_DATABASE());
 	EXECUTE format('CREATE ROLE bbf_role_admin CREATEDB CREATEROLE INHERIT PASSWORD NULL');
 	EXECUTE format('GRANT CREATE ON DATABASE %s TO bbf_role_admin WITH GRANT OPTION', CURRENT_DATABASE());
 	EXECUTE format('GRANT %I to bbf_role_admin WITH ADMIN TRUE;', sa_name);
 	EXECUTE format('CREATE ROLE sysadmin CREATEDB CREATEROLE INHERIT ROLE %I', sa_name);
 	EXECUTE format('GRANT sysadmin TO bbf_role_admin WITH ADMIN TRUE');
+	EXECUTE format('GRANT securityadmin TO bbf_role_admin WITH ADMIN TRUE');
```
tanscorpio7, line 294: We do not need a new statement, we can use ADMIN option to grant admin membership to bbf_role_admin when creating securityadmin.

contrib/babelfishpg_tds/src/backend/tds/tdsutils.c
tanscorpio7: Do we also need to add these checks in is_babelfish_role() ?

tanscorpio7 (changes requested):

contrib/babelfishpg_tsql/sql/ownership.sql
anju15bharti: We need this role to have atleast by default connect access on this babelfish physical database similar to our internal role, in fact for sysadmin, we provided create priv on this database which I think is necessarily not needed for our role we can just provide connect access to it.

contrib/babelfishpg_tsql/sql/ownership.sql
anju15bharti: I preferred creating the role clean and then granting required priv to required role as per the requirement.

contrib/babelfishpg_tsql/src/rolecmds.c
anju15bharti: Yes if we all agree to remove that existing line of code.

contrib/babelfishpg_tsql/src/rolecmds.c
anju15bharti: are we expecting to just give the negation of the second condition instead of expanding the negation?

contrib/babelfishpg_tsql/sql/ownership.sql
anju15bharti: Currently we are switching it to internal role but if in future we decide to use directly this, it should have the required privilege and also to explicitly reflect which priv it has and also to keep privilege wise similar to sysadmin and for future reference I have created this role with required attribute as discussed in design.

contrib/babelfishpg_tds/src/backend/tds/tdsutils.c
anju15bharti: yes

contrib/babelfishpg_tds/src/backend/tds/tdsutils.c
```diff
@@ -939,14 +939,17 @@ is_babelfish_role(const char *role)
 	Oid			bbf_master_guest_oid;
 	Oid			bbf_tempdb_guest_oid;
 	Oid			bbf_msdb_guest_oid;
+	Oid			securityadmin_oid;
 
 	sysadmin_oid = get_role_oid(BABELFISH_SYSADMIN, true);	/* missing OK */
 	role_oid = get_role_oid(role, true);	/* missing OK */
+	securityadmin_oid = get_role_oid(BABELFISH_SECURITYADMIN, true);  /* missing OK */
```
Deepesh125, line 946: Lets use `get_sysadmin_oid` to get oid of sysadmin and also introduce similar function to get oid of security_admin

contrib/babelfishpg_tsql/sql/ownership.sql
```diff
@@ -501,8 +508,11 @@ CAST(Ext.orig_loginname AS sys.nvarchar(128)) AS name,
 CAST('SERVER ROLE' AS sys.nvarchar(128)) AS type,
 CAST ('GRANT OR DENY' as sys.nvarchar(128)) as usage
 FROM pg_catalog.pg_roles AS Base INNER JOIN sys.babelfish_authid_login_ext AS Ext ON Base.rolname = Ext.rolname
-WHERE Ext.type = 'R' AND
-(pg_has_role(sys.suser_id(), 'sysadmin'::TEXT, 'MEMBER'));
+WHERE Ext.type = 'R'
+AND Ext.orig_loginname = 'sysadmin'
+AND bbf_is_member_of_role_nosuper(sys.suser_id(), sys.suser_id('sysadmin'))
+OR Ext.orig_loginname = 'securityadmin'
+AND bbf_is_member_of_role_nosuper(sys.suser_id(), sys.suser_id('securityadmin'));
```
Deepesh125: Can you explain this change for my understanding?

contrib/babelfishpg_tsql/src/rolecmds.c
```diff
@@ -2472,3 +2495,21 @@ remove_createrole_from_logins(PG_FUNCTION_ARGS)
 	table_close(rel, AccessShareLock);
 	PG_RETURN_INT32(0);
 }
+
+PG_FUNCTION_INFO_V1(bbf_is_member_of_role_nosuper);
+Datum
+bbf_is_member_of_role_nosuper(PG_FUNCTION_ARGS)
```
Deepesh125, line 2565: is there any alternate way of doing this? defining C function just to check membership may result in performance penalty. @sumitj824 can give more insights on this

contrib/babelfishpg_tsql/src/backend_parser/gram-tsql-epilogue.y.c
```diff
@@ -1967,3 +1967,24 @@ tsql_index_nulls_order(List *indexParams, const char *accessMethod)
 		}
 	}
 }
+
+static void
+is_server_role_supported (const char* serverrole, int position, core_yyscan_t yyscanner)
```
Deepesh125: nitpick: const char * serverrole

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -3140,7 +3147,8 @@ bbf_ProcessUtility(PlannedStmt *pstmt,
 
 							if (strcmp(defel->defname, "password") == 0)
 							{
-								if (get_role_oid(stmt->role->rolename, true) != GetSessionUserId() && !is_member_of_role(GetSessionUserId(), datdba))
+								if (get_role_oid(stmt->role->rolename, true) != GetSessionUserId() && (!is_member_of_role(GetSessionUserId(), datdba)
+																					&& !is_member_of_role(GetSessionUserId(), securityadm_oid)))
 									ereport(ERROR,(errcode(ERRCODE_INSUFFICIENT_PRIVILEGE),
 											 errmsg("Cannot alter the login '%s', because it does not exist or you do not have permission.", stmt->role->rolename)));
```
Deepesh125: doesnt this seem like redundant?

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -3705,13 +3727,30 @@ bbf_ProcessUtility(PlannedStmt *pstmt,
...
+					if (strlen(rolspec->rolename) == 8 && strncmp(rolspec->rolename, BABELFISH_SYSADMIN, 8) == 0)
+					{
+						if (grant_role->is_grant)
+							appendStringInfo(&query, "ALTER ROLE dummy WITH createrole createdb; ");
+						else
+							appendStringInfo(&query, "ALTER ROLE dummy WITH nocreaterole nocreatedb; ");
+					}
+
+					/* Otherwise, provide attribute for role priv */
 					else
```
Deepesh125: just completeness, check `else if rolspec->rolename == "security_admin" then .... else throw error`

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -3018,6 +3020,16 @@ sp_addlinkedsrvlogin_internal(PG_FUNCTION_ARGS)
 
 	initStringInfo(&query);
 
+	/*
+	 * check privileses for login
```
Deepesh125: please fix spelling mistake

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -3018,6 +3020,16 @@ sp_addlinkedsrvlogin_internal(PG_FUNCTION_ARGS)
...
+	 * allow if has privileges of sysadmin or securityadmin.
+	 */
+	if (!has_privs_of_role(GetSessionUserId(), get_sysadmin_oid()) &&
+				!has_privs_of_role(GetSessionUserId(), get_securityadmin_oid()))
+		ereport(ERROR,
+				(errcode(ERRCODE_INSUFFICIENT_PRIVILEGE),
+					errmsg("User does not have permission to perform this action.")));
```
Deepesh125, line 3036: Please point out additional testing being done for this change

test/JDBC/expected/is_srvrolemember.out
```diff
@@ -122,7 +122,7 @@ SELECT is_srvrolemember('securityadmin')
 GO
 ~~START~~
 int
-0
+1
```
Deepesh125, line 125: where are we granting this permission?

test/JDBC/expected/is_srvrolemember-vu-verify.out
```diff
@@ -122,7 +122,7 @@ SELECT is_srvrolemember('securityadmin')
 GO
 ~~START~~
 int
-0
+1
```
Deepesh125, line 125: where are we granting this permission?

test/JDBC/expected/securityadmin_role-vu-verify.out
```diff
@@ -0,0 +1,1268 @@
+-- tsql
```
Deepesh125, line 1: We should add test cases for following scenarios - 
1. Test cases in conjunction with GRANT ON SCHEMA support 
2. add test cases where user is member of both `sysadmin` and `securityadmin`, it should be able to connect to newly created database
3. we should also add negative tests cases where we try to add built in users to `securityadmin`
4. `create login securityadmin with password = '123'` should be error'd out
5. `alter server role securityadmin add member <db-role>` -- this should be error'd out also
6.

Deepesh125 (changes requested):

contrib/babelfishpg_tsql/sql/ownership.sql
anju15bharti: login_token should return rows for fixed server role of which the current login is member of.

contrib/babelfishpg_tsql/src/pl_handler.c
anju15bharti: we have blocked alter server role for any other roles other than sysadmin and securityadmin in the parsing phase itself so this else condition which you are taking about will actually never hit.

test/JDBC/expected/is_srvrolemember-vu-verify.out
anju15bharti: contrib/babelfishpg_tsql/sql/babelfishpg_tsql.sql:2029

test/JDBC/expected/securityadmin_role-vu-verify.out
anju15bharti: Added.

contrib/babelfishpg_tsql/src/procedures.c
anju15bharti: test/JDBC/expected/securityadmin_role-vu-verify.out:222

contrib/babelfishpg_tds/src/backend/tds/tdsutils.c
anju15bharti: we don't have get_sysadmin_oid in tds, we have get_securityadmin_oid and get_sysadmin_oid in tsql

contrib/babelfishpg_tsql/src/procedures.c
anju15bharti: corrected

contrib/babelfishpg_tsql/src/backend_parser/gram-tsql-epilogue.y.c
anju15bharti: changed

contrib/babelfishpg_tsql/sql/ownership.sql
HarshLunagariya: Then why do we need `Ext.orig_loginname = 'xyz'` condition. Don't we just need to check whether it is a fixed server role i.e. `type = 'R' AND is_fixed_role =1`?

contrib/babelfishpg_tsql/sql/ownership.sql
HarshLunagariya: Re-iterating again, But why does it actually need CONNECT? What are the internal operations are being blocked when we don't provide it?

contrib/babelfishpg_tsql/src/backend_parser/gram-tsql-epilogue.y.c
```diff
@@ -1967,3 +1967,24 @@ tsql_index_nulls_order(List *indexParams, const char *accessMethod)
 		}
 	}
 }
+
+static void
+is_server_role_supported (const char *serverrole, int position, core_yyscan_t yyscanner)
```
HarshLunagariya: better naming required. It should indicate that it is throwing error.

contrib/babelfishpg_tsql/src/rolecmds.c
HarshLunagariya: let's remove it.

contrib/babelfishpg_tsql/src/rolecmds.c
HarshLunagariya: Let's remove it.

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -3715,14 +3762,16 @@ bbf_ProcessUtility(PlannedStmt *pstmt,
 						else
 							standard_ProcessUtility(pstmt, queryString, readOnlyTree, context, params,
 													queryEnv, dest, qc);
-						exec_alter_role_cmd(query.data, spec);
+						if (*query.data)
```
HarshLunagariya: Why `*query.data`?

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -3692,13 +3714,38 @@ bbf_ProcessUtility(PlannedStmt *pstmt,
...
 					check_alter_server_stmt(grant_role);
 					spec = (RoleSpec *) linitial(grant_role->grantee_roles);
+					rolspec = (RoleSpec *) linitial(grant_role->granted_roles);
+					grantee_oid = get_role_oid(spec->rolename, false);
 					initStringInfo(&query);
-					if (grant_role->is_grant)
-						appendStringInfo(&query, "ALTER ROLE dummy WITH createrole createdb; ");
+					
+					/* If sysadmin, provide attribute for role and database priv */
+					if (strlen(rolspec->rolename) == 8 && strncmp(rolspec->rolename, BABELFISH_SYSADMIN, 8) == 0)
```
HarshLunagariya: Should we use some macro/function for `strlen(rolspec->rolename) == 8 && strncmp(rolspec->rolename, BABELFISH_SYSADMIN, 8) == 0` it is at multiple places.

contrib/babelfishpg_tsql/src/pl_handler.c
HarshLunagariya: @anju15bharti Agree. But when you support dbcreator role in your next PR then anyway you're modify this check , right? Let's do it here only.

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--4.3.0--4.4.0.sql
tanscorpio7: fixed roles are only targeted for PG 16, so should not be problem right ?

contrib/babelfishpg_tsql/sql/ownership.sql
tanscorpio7: A superuser will anyway have select privileges on underlying pg_catalogs. I do not think we need to restrict super users here.

contrib/babelfishpg_tds/src/backend/tds/tdsutils.c
tanscorpio7: We should revisit this function `is_babelfish_role`.
I am curious why we cannot simply check if bbf_admin_role is member of the role to verify if it is a babelfish role.

contrib/babelfishpg_tsql/src/rolecmds.c
tanscorpio7: Can `granted_name` be `sysadmin`. Should this be allowed for `BABELFISH_SECURITYADMIN ` members ?

contrib/babelfishpg_tsql/src/rolecmds.c
tanscorpio7: We do not need to restrict superusers from viewing `sys` catalogs. They anyway have access to everything in pg_catalog.

test/JDBC/expected/securityadmin_role-vu-verify.out
```diff
@@ -0,0 +1,1520 @@
...
+Alter server role securityadmin add member securityadmin_login1
+go
+
+select * from securityadmin_show_role_mem
+go
+~~START~~
+varchar#!#varchar
+sysadmin#!#jdbc_user
+securityadmin#!#securityadmin_login1
+~~END~~
```
HarshLunagariya: Does it shows all the logins? If yes, then let's show only the logins which we want to validate. Because stale logins from other tests can affect it's output in future.

test/JDBC/expected/securityadmin_role-vu-verify.out
```diff
@@ -0,0 +1,1520 @@
...
+~~ERROR (Code: 33557097)~~
+
+~~ERROR (Message: 'GRANT Database' is not currently supported in Babelfish)~~
+
+
+GRANT ALTER ANY EVENT NOTIFICATION TO no_securityadmin_login1 WITH GRANT OPTION; 
+go
+~~ERROR (Code: 33557097)~~
+
+~~ERROR (Message: 'GRANT Database' is not currently supported in Babelfish)~~
```
HarshLunagariya, line 142: This message this misleading, But it can be fixed separately if required in separate effort.

test/JDBC/expected/securityadmin_role-vu-verify.out
```diff
@@ -0,0 +1,1520 @@
...
+
+SELECT name, type, type_desc, default_database_name, default_language_name, credential_id, owning_principal_id, is_fixed_role
+FROM sys.server_principals 
+WHERE name =  'public';
+GO
+~~START~~
+varchar#!#char#!#nvarchar#!#varchar#!#varchar#!#int#!#int#!#bit
+public#!#R#!#SERVER_ROLE#!#<NULL>#!#<NULL>#!#<NULL>#!#1#!#0
+~~END~~
+
```
HarshLunagariya: nit: There are multiple such queries starting from 'jdbc_user' to '%securityadmin%'. Let's just combine them to save some test execution time.

test/JDBC/expected/securityadmin_role-vu-verify.out
```diff
@@ -0,0 +1,1520 @@
...
+~~START~~
+int
+1
+~~END~~
+
+
+-- terminate-tsql-conn user=securityadmin_login1 password=123
+
+-- psql
+SET client_min_messages = 'error';
```
HarshLunagariya: Why this?

test/JDBC/expected/securityadmin_role-vu-verify.out
```diff
@@ -0,0 +1,1520 @@
...
+-- terminate-tsql-conn
+
+-- tsql user=securityadmin_login1 password=123
+EXEC sp_addlinkedsrvlogin 'Accounts', 'False'
+GO
+
+EXEC sp_droplinkedsrvlogin 'Accounts', NULL
+GO
+
+-- inside procedure
```
HarshLunagariya, line 402: Let's update the test plan to include these cases as well.

test/JDBC/expected/securityadmin_role-vu-verify.out
```diff
@@ -0,0 +1,1520 @@
...
+exec securityadmin_drop_mem_p1
+go
+
+exec securityadmin_drop_login_p1
+go
+
+-- terminate-tsql-conn user=securityadmin_login1 password=123
+
+-- tsql
+-- login is member of both securityadmin as well as sysadmin
```
HarshLunagariya, line 421: Not mentioned in test plan, let's include this family of test cases as well.

test/JDBC/expected/securityadmin_role-vu-verify.out
```diff
@@ -0,0 +1,1520 @@
...
+
+-- tsql user=securityadmin_login1 password=123
+
+-- both attribute should be true
+select rolname, rolcreaterole, rolcreatedb from pg_roles where rolname = 'securityadmin_login1'
+go
+~~START~~
+varchar#!#bit#!#bit
+securityadmin_login1#!#1#!#1
+~~END~~
```
HarshLunagariya, line 509: can we test the `is_srvrolemember` for this case as well where it is member of sysadmin but not securityadmin? If we haven't add already at anywhere else.

test/JDBC/expected/securityadmin_role-vu-verify.out
```diff
@@ -0,0 +1,1520 @@
...
+0
+~~END~~
+
+
+select bbf_is_member_of_role_nosuper(suser_id(), suser_id('sysadmin'))
+go
+~~START~~
+bit
+1
+~~END~~
```
HarshLunagariya, line 1228: Execute `is_srvrolemember()` as well.

test/JDBC/expected/securityadmin_role-vu-verify.out
```diff
@@ -0,0 +1,1520 @@
...
+
+ALTER ROLE securityadmin WITH CONNECTION LIMIT 1;
+GO
+~~ERROR (Code: 0)~~
+
+~~ERROR (Message: ERROR: Babelfish-created logins/users/roles cannot be altered outside of a Babelfish session
+    Server SQLState: 42501)~~
+
+
+GRANT securityadmin TO securityadmin_restrict_new_login;
```
HarshLunagariya, line 1318: Try granting it to PG created role.

test/JDBC/expected/securityadmin_role-vu-verify.out
```diff
@@ -0,0 +1,1520 @@
...
+~~ERROR (Message: ERROR: permission denied to set session authorization
+    Server SQLState: 42501)~~
+
+
+SET ROLE securityadmin;
+GO
+~~ERROR (Code: 0)~~
+
+~~ERROR (Message: ERROR: permission denied to set role "securityadmin"
+    Server SQLState: 42501)~~
```
HarshLunagariya, line 1395: Which all roles can set it? Should we block it explicitly (w/o consulting permissions)?

test/JDBC/expected/securityadmin_role-vu-verify.out
```diff
@@ -0,0 +1,1520 @@
...
+
```
HarshLunagariya, line 1554: Add testcases where you are trying to assign object ownership to securityadmin.

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--4.3.0--4.4.0.sql
anju15bharti: yes

contrib/babelfishpg_tsql/src/pl_handler.c
anju15bharti: will modify this check in dbcreator but there also will not add any else condition because that according to me is irrelevant since it will never be the case. On next PR we will just add additional else check to handle another case.

contrib/babelfishpg_tsql/src/pl_handler.c
anju15bharti: because the *query.data can be null if it's member of sysadmin then we do not need to execute the internal revoking command.

contrib/babelfishpg_tsql/src/rolecmds.c
anju15bharti: This condition will not allow BABELFISH_SECURITYADMIN  members to alter sysadmin since it will pass the check and will throw error. I have testcases added as well.

test/JDBC/expected/securityadmin_role-vu-verify.out
anju15bharti: It's added below.

test/JDBC/expected/securityadmin_role-vu-verify.out
anju15bharti: From pg side, no babelfish role should be allowed to set if user is not superuser.

contrib/babelfishpg_tsql/sql/ownership.sql
anju15bharti: @tanscorpio7 We intentionally added the API which should not consider superuser priv because if superuser is also considers then in BBF for master user if it is only member of sysadmin then also it will bypass the check for securityadmin membership as well and we'll see row for securityadmin as well which is not expected in sql server.I will confirm with dbes.

contrib/babelfishpg_tsql/sql/ownership.sql
anju15bharti: It's only to have connect permission on the babelfish physical database or to be able to grant connect on physical database to other login but without explicitly granting connect to securityadmin it has connect permission. So need not to grant it.

HarshLunagariya (approved):

contrib/babelfishpg_tsql/src/pl_handler.c
tanscorpio7: +1, we already check for member ship just before

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -3505,8 +3526,9 @@ bbf_ProcessUtility(PlannedStmt *pstmt,
 						int			role_oid = get_role_oid(role_name, true);
 
 						if (!OidIsValid(role_oid) ||
-							!is_member_of_role(GetSessionUserId(), get_sysadmin_oid()) ||
-							role_oid == get_bbf_role_admin_oid())
+							(!is_member_of_role(GetSessionUserId(), get_sysadmin_oid()) && 
+							!is_member_of_role(GetSessionUserId(), securityadmin_oid)) ||
```
tanscorpio7: We already check for membership before at line 3507

contrib/babelfishpg_tsql/src/pl_handler.c
tanscorpio7: @anju15bharti ^

contrib/babelfishpg_tsql/src/pl_handler.c
tanscorpio7: We should be checking for length here `query.len`

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -3723,14 +3770,16 @@ bbf_ProcessUtility(PlannedStmt *pstmt,
...
+							exec_alter_role_cmd(query.data, spec);
 
 					}
 					PG_FINALLY();
 					{
 						/* Clean up. Restore previous state. */
 						SetUserIdAndSecContext(save_userid, save_sec_context);
-						pfree(query.data);
+						if (query.data)
+							pfree(query.data);
```
tanscorpio7: This will not free the memory if query.data is empty string. But we should unconditionally free it, since `initStringInfo` always allocates memory.

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -2912,7 +2912,12 @@ bbf_ProcessUtility(PlannedStmt *pstmt,
 
 					if (islogin)
 					{
-						if (!has_privs_of_role(GetSessionUserId(), get_role_oid("sysadmin", false)))
+						/*
+						 * Check if the current login has privileges to create
+						 * login.
+						 */
+						if (!has_privs_of_role(GetSessionUserId(), get_sysadmin_oid()) &&
+								!has_privs_of_role(GetSessionUserId(), get_securityadmin_oid()))
```
tanscorpio7, line 2991: Alter login should not work if current user is member of security admin and the login being altered is member of sysadmin

test/JDBC/input/securityadmin_role-vu-prepare.mix
```diff
@@ -0,0 +1,90 @@
...
+
+create user no_securityadmin_user1 for login no_securityadmin_login1
+go
+
+-- tsql
+create role securityadmin_rol
+go
+
+create database securityadmin_db1
+go
```
tanscorpio7, line 19: Lets also add a test case where securityadmin tries to drop a database it does not own and fails.

tanscorpio7 (changes requested):

contrib/babelfishpg_tsql/src/pl_handler.c
shalinilohia50: +1. We should have a better check with explanation.

test/JDBC/expected/securityadmin_role-vu-verify.out
```diff
@@ -0,0 +1,1557 @@
+-- tsql
+alter login securityadmin_login1 with password='123'
+go
+
+alter login no_securityadmin_login1 with password='123'
+go
+
+-- make login member of securityadmin
+Alter server role securityadmin add member securityadmin_login1
```
shalinilohia50, line 9: If a login already has a user/no user, can it be added as a member of `securityadmin`? 
Let's add a test for the same.

shalinilohia50 (changes requested):

contrib/babelfishpg_tsql/src/pl_handler.c
HarshLunagariya: Please validate and add test for the same as well.

HarshLunagariya (changes requested):

test/JDBC/expected/securityadmin_role-vu-verify.out
anju15bharti: In sysadmin, we additionally added check to not allow user creation or block making member if has user mapped but here we haven't even added these additional check then why is it needed in the file? Although I have already tested it locally.

test/JDBC/input/securityadmin_role-vu-prepare.mix
anju15bharti: we have testcase added in test/JDBC/expected/securityadmin_role-vu-verify.out:625

contrib/babelfishpg_tsql/src/pl_handler.c
tanscorpio7: Instead let us add check for sysadmin and restrict its drop here. Currently we rely on postgres on restricting dropping sysadmin.
I could run this on psql endpoint
```
drop owned by sysadmin;
```
and then drop sysadmin from TDS endpoint.

contrib/babelfishpg_tsql/sql/ownership.sql
Deepesh125: @anju15bharti IMHO, let's not over engineer this feature and lets keep it simple for now. If such requirements comes up in future then we will revisit this. I am making this point especially in this feature it is sensitive feature in terms of security.

contrib/babelfishpg_tsql/src/pl_handler.c
Deepesh125: good catch! we need some common logic to prevent alteration of pre-defined roles

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -3103,33 +3131,59 @@ sp_droplinkedsrvlogin_internal(PG_FUNCTION_ARGS)
...
+		*
+		* Linked logins were first implemented as PG USER MAPPINGs for the CURRENT_USER which
+		* was not entirely correct because T-SQL linked logins are not user or login specific.
+		* To address this we now create user mapping for the PG PUBLIC role internally.
+		*
+		* To ensure sp_droplinkedsrvlogin works in accordance with both the older and newer
+		* implementation of linked logins, we try to drop USER MAPPINGs for both the CURRENT_USER
+		* and PUBLIC PG roles.
+		*/
+		appendStringInfo(&query, "DROP USER MAPPING IF EXISTS FOR CURRENT_USER SERVER \"%s\"", servername);
```
Deepesh125, line 3173: make sure to free this

Deepesh125 (changes requested):

contrib/babelfishpg_tsql/src/pl_handler.c
anju15bharti: It is not actually redundant it's just that password based login case is handled separately and before the general case, below that we are not checking if it's password based login.

contrib/babelfishpg_tsql/src/pl_handler.c
anju15bharti: we have already blocked for securityadmin and bbf_role_admin role to get dropped, will add the same for sysadmin as well.

contrib/babelfishpg_tsql/sql/ownership.sql
anju15bharti: Agree, but we also don't need these attributes for sysadmin but we have kept it, so following that trend I have kept this attribute to this securityadmin role, Also giving this attribute to securityadmin is not escalating any privilege because even if we consider this role alone it should have this attribute.

tanscorpio7 (approved):

test/JDBC/expected/securityadmin_role-vu-verify.out
```diff
@@ -0,0 +1,1609 @@
...
+
+~~ERROR (Message: Current login securityadmin_login1 does not have permission to alter server role)~~
+
+
+-- alter server role securityadmin add member db roles should error out
+Alter server role securityadmin add member guest
+go
+~~ERROR (Code: 33557097)~~
+
+~~ERROR (Message: role "guest" does not exist)~~
```
Deepesh125, line 573: This error is misleading

anju15bharti: since alter server role is server level and on server level guest is not present, so it is expected.

anju15bharti:
Closing this PR since the target branch is moved to amazon aurora repo
New PR https://github.com/amazon-aurora/babelfish_extensions/pull/67

PR #2911, opened 2024-09-04
Fixing the default and max typmod arg cases for various datatypes in procedure/function

tanyagupta17:
### Description
Currently, end user is able to specify nvarchar(max) for output param while defining the procedure. Procedure will be created without any error but it will throw an error upon executing the procedure. This PR will handle the default and max params for procedure and functions. We have handled and tested the following datatypes :

```
NVARCHAR [ ( n | max ) ]
VARCHAR [ ( n | max ) ]
VARBINARY [ ( n | max ) ]
NCHAR [ ( n ) ]
SMALLDATETIME
DECIMAL [ ( p [ , s ] ) ]
BINARY [ ( n ) ]
BPCHAR [ ( n ) ]
```

### Issues Resolved

BABEL-5059

* **Use case based -** YES

* **Arbitrary inputs -** YES

* **Performance tests -** NA

* **Tooling impact -** NA

* **Client tests -** NA

### Check List


contrib/babelfishpg_tsql/src/pltsql_function_probin_handler.c
```diff
@@ -62,14 +62,20 @@ pltsql_function_probin_writer(CreateFunctionStmt *stmt, Oid languageOid, char **
 	char	   *langname;
 	int			probin_len;
 	Jsonb	   *jb;
+	bool		dump_restore = false;
+	const char *babelfish_dump_restore = GetConfigOption("babelfishpg_tsql.dump_restore", true, false);
```
sumitj824: Why do we need additional handling during dump/restore? As discussed offline earlier, we only need to handle it in upgrade script.

tanyagupta17: In upgrade script we are only able to handle minor version upgrade not major version upgrade so currently I am testing ways for major version upgrade. Looks like that would require C changes afterall

sumitj824: I can see that changes in SQL upgrade script are sufficient for almost all MVU/mVU paths. The failures are only in some upgrade path which involves 14.6/13.6 version. So I would suggest to narrow down the issue, instead of making changes in dump/restore.

tanyagupta17: As the SQL upgrade script was working for all the upgrade paths from 14.6 and ahead. After discussion with Rishabh, we narrowed down the issue, and found that in the upgrade query we are using `sys.babelfish_function_ext` to filter out the bbf defined functions. But this catalog was introduced in 14.6, so the following upgrade paths were failing as the procedure information was not stored for them in the `sys.babelfish_function_ext` catalog and hence was not getting updated:
13_6-14_6-target_latest
13_9-14_6-target_latest
14_3-14_6_target-latest
14_5-14_6-target_latest
So we concluded to use pg_language catalog and using the pltsql filter to narrow down the bbf defined procedures and functions.

contrib/babelfishpg_tsql/src/pltsql.h
```diff
@@ -2175,7 +2175,7 @@ extern void update_ViewStmt(Node *n, const char *view_schema);
 extern void update_AlterDefaultPrivilegesStmt(Node *n, const char *schema, const char *role1, const char *role2, const char *grantee, const char *priv);
 extern AccessPriv *make_accesspriv_node(const char *priv_name);
 extern RoleSpec   *make_rolespec_node(const char *rolename);
-extern void pltsql_check_or_set_default_typmod(TypeName *typeName, int32 *typmod, bool is_cast);
+extern void pltsql_check_or_set_default_typmod(TypeName *typeName, int32 *typmod, bool is_cast, bool is_procedure_or_func);
```
KushaalShroff: So we are setting the 3 arg true only in extension and false only in engine?

Also the naming is ambiguous, is it procedure or function which is true? The naming must indicate that

KushaalShroff (changes requested):

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--4.3.0--4.4.0.sql
```diff
@@ -10566,6 +10566,78 @@ CALL sys.babelfish_alter_default_privilege_on_schema();
...
+            )
+        )::text
+    )
+)
+FROM (
+    SELECT p.oid, p.proallargtypes, p.proargtypes
+    FROM pg_proc p
+    INNER JOIN sys.babelfish_namespace_ext sch ON sch.nspname = p.pronamespace::regnamespace::name
+    WHERE p.prolang = (SELECT oid FROM pg_language WHERE lanname = 'pltsql') and p.prokind = 'p'
+) p2
```
sumitj824: Please avoid using unnecessary subquery. I think it can be optimize further.

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--4.3.0--4.4.0.sql
```diff
@@ -10566,6 +10566,78 @@ CALL sys.babelfish_alter_default_privilege_on_schema();
...
+            )
+        )::text
+    )
+)
+FROM (
+    SELECT p.oid, p.proargtypes, p.prorettype, p.proallargtypes
+    FROM pg_proc p
+    INNER JOIN sys.babelfish_namespace_ext sch ON sch.nspname = p.pronamespace::regnamespace::name
+    WHERE p.prolang = (SELECT oid FROM pg_language WHERE lanname = 'pltsql') and p.prokind = 'f'
+) p2
```
sumitj824: Same here, avoid using unnecessary subquery.

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--4.3.0--4.4.0.sql
```diff
@@ -10566,6 +10566,78 @@ CALL sys.babelfish_alter_default_privilege_on_schema();
...
+                                END
+                            WHEN p2.proallargtypes is null AND p2.prorettype is not null THEN
+                                CASE
+                                    WHEN a::text = '-8000' AND (p2.prorettype = ('sys.varchar'::regtype)::oid OR p2.prorettype = ('sys.nvarchar'::regtype)::oid OR p2.prorettype = ('sys.varbinary'::regtype)::oid) THEN '-1'
+                                    WHEN a::text = '-1' AND p2.prorettype = ('sys.smalldatetime'::regtype)::oid THEN '0'
+                                    WHEN a::text = '-1' AND (p2.prorettype = ('sys.varchar'::regtype)::oid OR p2.prorettype = ('sys.nvarchar'::regtype)::oid OR p2.prorettype = ('sys.varbinary'::regtype)::oid OR p2.prorettype = ('sys.nchar'::regtype)::oid OR p2.prorettype = ('sys.binary'::regtype)::oid OR p2.prorettype = ('sys.bpchar'::regtype)::oid) THEN '1'
+                                    ELSE a::text
+                                END
+                            ELSE a::text
+                        END
```
sumitj824: Instead of having separate UPDATE statment, I think we should use Single UPDATE Statement with a CASE Expression based on prokind.

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--4.3.0--4.4.0.sql
```diff
@@ -10566,6 +10566,78 @@ CALL sys.babelfish_alter_default_privilege_on_schema();
 -- Drop this procedure after it gets executed once.
 DROP PROCEDURE sys.babelfish_alter_default_privilege_on_schema();
 
+UPDATE pg_proc p1
+SET probin = (
+    SELECT (
+        jsonb_set(
+            probin::jsonb,
+            '{typmod_array}',
```
sumitj824: I think we should also update the `{version_num}` to indicate that value has been modified along with this.

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--4.3.0--4.4.0.sql
```diff
@@ -10566,6 +10566,78 @@ CALL sys.babelfish_alter_default_privilege_on_schema();
 -- Drop this procedure after it gets executed once.
 DROP PROCEDURE sys.babelfish_alter_default_privilege_on_schema();
 
+UPDATE pg_proc p1
```
sumitj824, line 10570: Where are we handling the upgrade scenarios of user already having these changes(i.e. user coming from 3.8.0)?

contrib/babelfishpg_tsql/src/pltsql_function_probin_handler.c
```diff
@@ -418,6 +418,9 @@ buildTypmodArray(CreateFunctionStmt *stmt, int **typmod_array_p, int *array_len_
 		if (!arg_typmod)
 		{
 			(*typmod_array_p)[i] = -1;
+
+			if (sql_dialect == SQL_DIALECT_TSQL)
+				pltsql_check_or_set_default_typmod(fp->argType, &(*typmod_array_p)[i], false, true);
```
Deepesh125: can we please add comment why is dialect check needed here?

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--4.3.0--4.4.0.sql
```diff
@@ -10566,6 +10566,78 @@ CALL sys.babelfish_alter_default_privilege_on_schema();
...
+            probin::jsonb,
+            '{typmod_array}',
+            to_jsonb(
+                (
+                    SELECT jsonb_agg(
+                        CASE
+                            WHEN a::text = '-8000' AND (p2.proargtypes[b-1] = ('sys.varchar'::regtype)::oid OR p2.proargtypes[b-1] = ('sys.nvarchar'::regtype)::oid OR p2.proargtypes[b-1] = ('sys.varbinary'::regtype)::oid) THEN '-1'
+                            WHEN a::text = '-1' AND p2.proargtypes[b-1] = ('sys.smalldatetime'::regtype)::oid THEN '0'
+                            WHEN a::text = '-1' AND (p2.proargtypes[b-1] = ('sys.varchar'::regtype)::oid OR p2.proargtypes[b-1] = ('sys.nvarchar'::regtype)::oid OR p2.proargtypes[b-1] = ('sys.varbinary'::regtype)::oid OR p2.proargtypes[b-1] = ('sys.nchar'::regtype)::oid OR p2.proargtypes[b-1] = ('sys.binary'::regtype)::oid OR p2.proargtypes[b-1] = ('sys.bpchar'::regtype)::oid) THEN '1'
+                            ELSE a::text
```
sumitj824: Do we need to handle cases for UDT defined on these types and later used in procedures or functions?

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--4.3.0--4.4.0.sql
```diff
@@ -10566,6 +10566,78 @@ CALL sys.babelfish_alter_default_privilege_on_schema();
...
+        jsonb_set(
+            probin::jsonb,
+            '{typmod_array}',
+            to_jsonb(
+                (
+                    SELECT jsonb_agg(
+                        CASE
+                            WHEN a::text = '-8000' AND (p2.proargtypes[b-1] = ('sys.varchar'::regtype)::oid OR p2.proargtypes[b-1] = ('sys.nvarchar'::regtype)::oid OR p2.proargtypes[b-1] = ('sys.varbinary'::regtype)::oid) THEN '-1'
+                            WHEN a::text = '-1' AND p2.proargtypes[b-1] = ('sys.smalldatetime'::regtype)::oid THEN '0'
+                            WHEN a::text = '-1' AND (p2.proargtypes[b-1] = ('sys.varchar'::regtype)::oid OR p2.proargtypes[b-1] = ('sys.nvarchar'::regtype)::oid OR p2.proargtypes[b-1] = ('sys.varbinary'::regtype)::oid OR p2.proargtypes[b-1] = ('sys.nchar'::regtype)::oid OR p2.proargtypes[b-1] = ('sys.binary'::regtype)::oid OR p2.proargtypes[b-1] = ('sys.bpchar'::regtype)::oid) THEN '1'
```
sumitj824: We can simply do `p2.proargtypes[b-1]::regtype in ('sys.varchar', 'sys.nvarchar' ...)`.

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--4.3.0--4.4.0.sql
```diff
@@ -10566,6 +10566,78 @@ CALL sys.babelfish_alter_default_privilege_on_schema();
...
+                    FROM jsonb_array_elements_text(probin::jsonb->'typmod_array')  WITH ORDINALITY AS elem(a,b)
+                )
+            )
+        )::text
+    )
+)
+FROM (
+    SELECT p.oid, p.proallargtypes, p.proargtypes
+    FROM pg_proc p
+    INNER JOIN sys.babelfish_namespace_ext sch ON sch.nspname = p.pronamespace::regnamespace::name
```
sumitj824: Here we aren't updating the typmod for procedure/functions in sys schema. Can you please check that if there is any system functions defined in PLTSQL language for which we need to update?

tanyagupta17: I checked system defined functions and procedures defined in PLTSQL language in sys schema. Our changes don't have any affect on them and we don't need to update them.As their typmod value is either a non-default and non-max value or from the datatypes that we are not updating(eg: int4, sysname,text,bit etc).

contrib/babelfishpg_tsql/src/pltsql_function_probin_handler.c
tanyagupta17: Added the comment

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--4.3.0--4.4.0.sql
tanyagupta17: Done

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--4.3.0--4.4.0.sql
tanyagupta17: As adding this change would bring in additional complexity, we have decided to introduce this change only for 16x and hence this won't be required anymore

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--4.3.0--4.4.0.sql
tanyagupta17: As adding this change would bring in additional complexity, we have decided to introduce this change only for 16x and hence this won't be required anymore

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--4.3.0--4.4.0.sql
tanyagupta17: As per the testing the UDT cases are already handled.

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--4.3.0--4.4.0.sql
tanyagupta17: Done

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--4.3.0--4.4.0.sql
tanyagupta17: Done

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--4.3.0--4.4.0.sql
tanyagupta17: Done

contrib/babelfishpg_tsql/src/pltsql.h
tanyagupta17: From engine side we are handling functionalities like create/alter table , create type, etc. using the hook check_or_set_default_typmod_hook which internally calls the function pltsql_check_or_set_default_typmod. By keeping the value for is_procedure_or_func as false we make sure to keep these functionalities same as before, as we target to only update proc/func typmod values through this.

I have named the variable to be is_procedure_or_func because it handles both the procedure and function arguments together as both expect the same behaviour from pltsql_check_or_set_default_typmod function. Thus it could be either a procedure or function to be true.

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--4.3.0--4.4.0.sql
```diff
@@ -10566,6 +10566,50 @@ CALL sys.babelfish_alter_default_privilege_on_schema();
...
+                            FROM jsonb_array_elements_text(p1.probin::jsonb->'typmod_array') WITH ORDINALITY AS elem(a,b)
+                        )
+                    )
+                )::text
+            )
+        )
+    END
+FROM pg_proc p2
+INNER JOIN sys.babelfish_namespace_ext sch ON sch.nspname = p2.pronamespace::regnamespace::name
+WHERE p2.prolang = (SELECT oid FROM pg_language WHERE lanname = 'pltsql')
```
sumitj824: Please use inner join instead of subquery.

test/JDBC/input/storedProcedures/BABEL-5059-vu-prepare.sql
```diff
@@ -0,0 +1,151 @@
...
+GO
+
+CREATE TYPE nv_2_new from nvarchar(2)
+GO
+
+CREATE TYPE vchar_new from varchar
+GO
+
+CREATE TYPE binr_new from binary 
+GO
```
sumitj824: Please add testcases for UDT created from other types as well which we are handling in the PR.

test/JDBC/input/forjson/forjsonauto-vu-prepare.sql
```diff
@@ -139,7 +139,7 @@ END
 GO
 
 CREATE FUNCTION forjson_vu_f_1()
-RETURNS sys.NVARCHAR(5000) AS
+RETURNS sys.NVARCHAR(4000) AS
```
sumitj824, line 149: What will happen if user has already created with 5000 length then upgrade?

test/JDBC/input/storedProcedures/BABEL-5059-vu-prepare.sql
```diff
@@ -0,0 +1,151 @@
+-- nvarchar
+CREATE PROCEDURE proc_test_1 (@a NVARCHAR, @b NVARCHAR OUTPUT) AS BEGIN SET @b=@a; SELECT @b; END;
```
sumitj824: Please follow the naming convention for objects. To ensure the uniqueness of object names across different testfiles, we prefix them with the corresponding file name, names should be like `babel_5050_vu_prepare_proc_test_1`.

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--4.3.0--4.4.0.sql
```diff
@@ -10566,6 +10566,50 @@ CALL sys.babelfish_alter_default_privilege_on_schema();
...
+                                        CASE
+                                            WHEN a::text = '-8000' AND p2.prorettype::regtype in ('sys.varchar', 'sys.nvarchar', 'sys.varbinary') THEN '-1'
+                                            WHEN a::text = '-1' AND p2.prorettype::regtype::text = 'sys.smalldatetime' THEN '0'
+                                            WHEN a::text = '-1' AND p2.prorettype::regtype in ('sys.varchar', 'sys.nvarchar', 'sys.varbinary', 'sys.nchar','sys.binary','sys.bpchar') THEN '1'
+                                            ELSE a::text
+                                        END
+                                    ELSE a::text
+                                END
+                            )
+                            FROM jsonb_array_elements_text(p1.probin::jsonb->'typmod_array') WITH ORDINALITY AS elem(a,b)
```
sumitj824: Instead of (a,b), please use meaningful names like (typmod, position/index).

test/JDBC/input/storedProcedures/BABEL-5059-vu-prepare.sql
```diff
@@ -0,0 +1,151 @@
...
+-- bpchar
+CREATE PROCEDURE proc_test_28 (@a bpchar, @b bpchar OUTPUT) AS BEGIN SET @b=@a; SELECT len(@b); END;
+GO
+
+CREATE PROCEDURE proc_test_29 (@a bpchar(10), @b bpchar(10) OUTPUT) AS BEGIN SET @b=@a; SELECT @b; END;
+GO
+
+CREATE PROCEDURE proc_test_30 (@a bpchar(8000), @b bpchar(8000) OUTPUT) AS BEGIN SET @b=@a; SELECT len(@b); END;
+GO
+
```
sumitj824: Please use char instead of bpchar.

contrib/babelfishpg_tsql/src/pltsql_function_probin_handler.c
```diff
@@ -418,6 +418,13 @@ buildTypmodArray(CreateFunctionStmt *stmt, int **typmod_array_p, int *array_len_
 		if (!arg_typmod)
 		{
 			(*typmod_array_p)[i] = -1;
+
+			/**
+			 * For TSQL procedures and functions, default typmod for sys.(N)(VAR)(BP)CHAR, sys.(VAR)BINARY datatypes
+			 * should be handled differently 
+			 */
```
sumitj824: NIT: comment style should be like:
```
/*
 * This is multiline
 * comment.
 */
```

contrib/babelfishpg_tsql/src/pltsql_function_probin_handler.c
sumitj824: Are we sure that it should be restricted to TSQL dialect only? In future, we can create our system procs/funcs in PLTSQL language, are we saying typmod will not needed for them?

contrib/babelfishpg_tsql/src/pltsql_utils.c
```diff
@@ -419,7 +419,7 @@ pltsql_createFunction(ParseState *pstate, PlannedStmt *pstmt, const char *queryS
  * And length should be restricted to 4000 for sys.varchar and sys.char datatypes
  */
 void
-pltsql_check_or_set_default_typmod(TypeName *typeName, int32 *typmod, bool is_cast)
+pltsql_check_or_set_default_typmod(TypeName *typeName, int32 *typmod, bool is_cast, bool is_procedure_or_func)
```
sumitj824: Please update above comment to reflect the extra arg we are adding.

contrib/babelfishpg_tsql/src/pltsql_function_probin_handler.c
```diff
@@ -434,6 +441,12 @@ buildTypmodArray(CreateFunctionStmt *stmt, int **typmod_array_p, int *array_len_
 				else
 				{
 					(*typmod_array_p)[i] = ptr->val.ival.ival;
+				/**
+			 	* For TSQL procedures and functions, MAX typmod for sys.(N)(VAR)CHAR, sys.VARBINARY datatypes
```
sumitj824: Can we add comment what dataypes previous call will cover and what this will cover? Instead of having genric comment.

test/JDBC/input/forjson/forjsonauto-vu-prepare.sql
tanyagupta17: If a user has an already created function with typmod value as 5000 it would still be executed as before.
But if the user tries to create a new function now, it won't be able to create such a function with a typmod value >4000 for nvarchar. This change handles the limit of typmod for various datatypes such as 'sys.varchar', 'sys.nvarchar', 'sys.varbinary', 'sys.nchar','sys.binary','sys.bpchar'.

test/JDBC/input/storedProcedures/BABEL-5059-vu-prepare.sql
tanyagupta17: Done

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--4.3.0--4.4.0.sql
tanyagupta17: Done

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--4.3.0--4.4.0.sql
tanyagupta17: Done

contrib/babelfishpg_tsql/src/pltsql_function_probin_handler.c
tanyagupta17: Had seen instances in code base where we had used the previous comment style. Updated to new one.

test/JDBC/input/storedProcedures/BABEL-5059-vu-prepare.sql
tanyagupta17: Done

test/JDBC/input/storedProcedures/BABEL-5059-vu-prepare.sql
tanyagupta17: Done

contrib/babelfishpg_tsql/src/pltsql_function_probin_handler.c
tanyagupta17: Had added the specific datatypes and typmod typ (max/default) along with proc/func specification. 
Updated the comment as per our discussion.

contrib/babelfishpg_tsql/src/pltsql_utils.c
tanyagupta17: Done

contrib/babelfishpg_tsql/src/pltsql_function_probin_handler.c
tanyagupta17: When a proc/function is created from PG side using PLTSQL language we don't have "MAX" typmode for any datatype. The behavior for pltsql func/proc when executed from TSQL side is still same as before (for default and specific typmod) so we should restrict to TSQL dialect only.

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--4.3.0--4.4.0.sql
```diff
@@ -10566,6 +10566,50 @@ CALL sys.babelfish_alter_default_privilege_on_schema();
...
+                jsonb_set(
+                    p1.probin::jsonb,
+                    '{typmod_array}',
+                    to_jsonb(
+                        (
+                            SELECT jsonb_agg(
+                                CASE
+                                    WHEN p2.prokind = 'p' OR (p2.prokind = 'f' AND p2.proallargtypes IS NULL AND p2.proargtypes[typ_index-1] IS NOT NULL) THEN
+                                        CASE
+                                            WHEN typmod::text = '-8000' AND p2.proargtypes[typ_index-1]::regtype in ('sys.varchar', 'sys.nvarchar', 'sys.varbinary') THEN '-1'
```
sumitj824: Do we really need to explicitly cast typmod to text?

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--4.3.0--4.4.0.sql
```diff
@@ -10566,6 +10566,50 @@ CALL sys.babelfish_alter_default_privilege_on_schema();
 -- Drop this procedure after it gets executed once.
 DROP PROCEDURE sys.babelfish_alter_default_privilege_on_schema();
 
+-- To handle the typmod values for procedure and function
+UPDATE pg_proc p1
+SET probin = 
+    CASE 
+        WHEN p1.probin::jsonb->'typmod_array' = '[]'::jsonb THEN p1.probin
```
sumitj824: Instead of this lets add filter in where clause using proargtypes and prorettype.

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--4.3.0--4.4.0.sql
```diff
@@ -10566,6 +10566,50 @@ CALL sys.babelfish_alter_default_privilege_on_schema();
...
+                            SELECT jsonb_agg(
+                                CASE
+                                    WHEN p2.prokind = 'p' OR (p2.prokind = 'f' AND p2.proallargtypes IS NULL AND p2.proargtypes[typ_index-1] IS NOT NULL) THEN
+                                        CASE
+                                            WHEN typmod::text = '-8000' AND p2.proargtypes[typ_index-1]::regtype in ('sys.varchar', 'sys.nvarchar', 'sys.varbinary') THEN '-1'
+                                            WHEN typmod::text = '-1' AND p2.proargtypes[typ_index-1]::regtype::text = 'sys.smalldatetime' THEN '0'
+                                            WHEN typmod::text = '-1' AND p2.proargtypes[typ_index-1]::regtype in ('sys.varchar', 'sys.nvarchar', 'sys.varbinary', 'sys.nchar','sys.binary','sys.bpchar') THEN '1'
+                                            ELSE typmod::text
+                                        END
+                                    WHEN p2.prokind = 'f' AND p2.proallargtypes IS NULL AND p2.prorettype IS NOT NULL THEN
```
sumitj824: Can you explain the usecase of this? Is it possible for PLTSQL function having proargtypes and prorettype both not null?

test/JDBC/input/storedProcedures/BABEL-5059-vu-prepare.sql
```diff
@@ -0,0 +1,247 @@
+-- nvarchar
+CREATE PROCEDURE babel_5059_proc_test_1 (@a NVARCHAR, @b NVARCHAR OUTPUT) AS BEGIN SET @b=@a; SELECT @b; END;
```
sumitj824, line 3: I only see tests for procedure, Please add tests for functions as well.

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--4.3.0--4.4.0.sql
tanyagupta17: Removed

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--4.3.0--4.4.0.sql
tanyagupta17: Updating the filter with the following two changes and resoning:

1. In procedure we only rely on proargtypes, if its not there then typmod_array will be [] , so exclude this in where clause.

2. In function we will always have a return type thus there will always be some value in typmod_array. In function when its ITVF or MSTVF, then we have values in proallargtypes and not in proargtypes , but we dont need to update these kinds of functions therefore we can exclude this as well in where clause

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--4.3.0--4.4.0.sql
tanyagupta17: Yes, in case of functions that have input arguments(let's say n arguments) and a return value, we store input argument types in proargtypes and return type in prorettype.
In this case, the size of the typmod array will be n+1 (input + return types). Hence, the first n values, we are picking up from proargtypes and the last one from prorettype

test/JDBC/input/storedProcedures/BABEL-5059-vu-prepare.sql
tanyagupta17: Added for functions with all the datatypes that we are updating typmod value for, UDT testing ,  ITVF and MSTVF

sumitj824 (approved):
Upgrade Script and Tests LGTM.

contrib/babelfishpg_tsql/src/pltsql.h
Deepesh125: we do not need to introduce new field for that. You can refactor pltsql_check_or_set_default_typmod using some helper. We can directly call this directly from extension with is_procedure_or_func = true and it will be false if helper gets called through hook.

test/JDBC/input/storedProcedures/BABEL-5059-vu-verify.sql
```diff
@@ -0,0 +1,469 @@
+-- Procedure
```
Deepesh125, line 1: can you also print function signature using `select definition from information_schema.routines  where xyz `? this will ensure that max are handled correctly during DDL export also

contrib/babelfishpg_tsql/src/pltsql_function_probin_handler.c
```diff
@@ -463,13 +476,27 @@ buildTypmodArray(CreateFunctionStmt *stmt, int **typmod_array_p, int *array_len_
...
 	{
-		(*typmod_array_p)[i++] = -1;
+		(*typmod_array_p)[i] = -1;
+		/*
+		 * Handling default typmod value for sys.(N)(VAR)(BP)CHAR, sys.(VAR)BINARY return datatypes 
+		 * for TSQL functions
+		 */
+        if (sql_dialect == SQL_DIALECT_TSQL)
+            pltsql_check_or_set_default_typmod(ret, &(*typmod_array_p)[i], false, true);
+		i++;
```
Deepesh125: ident

test/JDBC/input/storedProcedures/BABEL-5059-vu-verify.sql
tanyagupta17: Done

contrib/babelfishpg_tsql/src/pltsql_function_probin_handler.c
tanyagupta17: Done

contrib/babelfishpg_tsql/src/pltsql.h
tanyagupta17: Done

Deepesh125 (approved):

sumitj824 (approved):

KushaalShroff (approved):

PR #2918, opened 2024-09-09
Fix date functions to take into account the timezone setting

Anikait143:
### 1. Issues
- When the timezone parameter group setting is set to something other than the default UTC, Babelfish date functions(SYSDATETIME(), SYSDATETIMEOFFSET(), GETDATE(), CURRENT_TIMESTAMP) don't take it into account, while they should. The same issue occurs with a session-level timezone parameter.
- SYSDATETIMEOFFSET() function returns datetimeoffset which does not store timezone due to incorrect casting function.

- Added appropriate CAST from TIMESTAMP to respective return types of date functions to ensure proper accounting for timezone setting.
- Added timestamptz_datetimeoffset function for Converting timestamp with time zone to datetimeoffset. It ensures to store timezone property similar to timestamptz_datetime2 and timestamptz_datetime functions.

```
1> select set_config('timezone', 'US/Pacific', false);
2> go

1> select cast(current_setting('timezone') as varchar(11))  as tz ,cast(now() as varchar(30)) as now ,SYSDATETIME()     ,SYSDATETIMEOFFSET()     ,SYSUTCDATETIME()     ,CURRENT_TIMESTAMP     ,GETDATE()     ,GETUTCDATE();
2> go 

tz          now                            sysdatetime                            sysdatetimeoffset                             sysutcdatetime                         getdate                 getdate                 getutcdate
----------- ------------------------------ -------------------------------------- --------------------------------------------- -------------------------------------- ----------------------- ----------------------- -----------------------
US/Pacific  2024-09-09 11:24:54.88749-07              2024-09-09 18:26:19.3272810            2024-09-09 18:26:19.3272810 +00:00            2024-09-09 18:25:06.3912410 2024-09-09 18:26:19.327 2024-09-09 18:26:19.327 2024-09-09 18:25:06.390
```
```
1> select set_config('timezone', 'US/Pacific', false);
2> go

1> select cast(current_setting('timezone') as varchar(11))  as tz ,cast(now() as varchar(30)) as now ,SYSDATETIME()     ,SYSDATETIMEOFFSET()     ,SYSUTCDATETIME()     ,CURRENT_TIMESTAMP     ,GETDATE()     ,GETUTCDATE();
2> go 

tz          now                            sysdatetime                            sysdatetimeoffset                             sysutcdatetime                         getdate                 getdate                 getutcdate
----------- ------------------------------ -------------------------------------- --------------------------------------------- -------------------------------------- ----------------------- ----------------------- -----------------------
US/Pacific  2024-09-09 11:24:54.88749-07              2024-09-09 11:25:06.3912410            2024-09-09 11:25:06.3912410 -07:00            2024-09-09 18:25:06.3912410 2024-09-09 11:25:06.390 2024-09-09 11:25:06.390 2024-09-09 18:25:06.390
```
- Added Test Cases for these date functions

Overall execution time(in ms) for 1000 direct function call when parallel query is enabled:
```
Functions             Before   After

getdate	                251	259
sysdatetime	        286	292
sysdatetimeoffset	329	332

```

Task: BABEL-5069

* **Use case based -**   YES

* **Arbitrary inputs -** YES

* **Performance tests -** NA

* **Tooling impact -** NA

* **Memory tests -** NA

* **Client tests -** NA

### Check List


contrib/babelfishpg_common/src/datetimeoffset.c
```diff
@@ -639,6 +640,43 @@ timestamp_datetimeoffset(PG_FUNCTION_ARGS)
...
+		if (tm2timestamp(tm, fsec, NULL, &time) != 0)
+			ereport(ERROR,
+					(errcode(ERRCODE_DATETIME_VALUE_OUT_OF_RANGE),
+					 errmsg("data out of range for datetimeoffset")));
+	}
+
+	result = (tsql_datetimeoffset *) palloc(DATETIMEOFFSET_LEN);
+	result->tsql_ts = time;
+	result->tsql_tz = (int16) tz / 60;
+	CheckDatetimeoffsetRange(result);
```
sumitj824, line 674: Can you explain why are we checking for out of range two times here, first in else block and then by invoking CheckDatetimeoffsetRange?

contrib/babelfishpg_common/src/datetimeoffset.c
```diff
@@ -639,6 +640,43 @@ timestamp_datetimeoffset(PG_FUNCTION_ARGS)
 	PG_RETURN_DATETIMEOFFSET(result);
 }
 
+/* timestamptz_datetimeoffset()
+ * Convert timestamp with time zone to datetimeoffset
+ */
+Datum
+timestamptz_datetimeoffset(PG_FUNCTION_ARGS)
```
sumitj824: Suggestion: I would suggest to not define these function with `PG_FUNCTION_ARGS` until it is supposed to be invoked by SQL function. Invoking these functions from C using `DirectFunctionCall1` involves building `FCINFO` and will have downside of it in terms of perfomance.

sumitj824 (commented):
Lets also get the performance numbers of these functions with enforced parallel query.

contrib/babelfishpg_common/src/datetimeoffset.c
Anikait143: **First check**: The check within timestamp2tm() and tm2timestamp() ensures that the timestamp can be correctly converted to and from the pg_tm structure without any issues. If the conversion fails (e.g., due to an invalid date like February 30), an error is thrown at this point.

**Second check**: The CheckDatetime2Range() function ensures that the final timestamp is valid within the specific range constraints of datetimeoffset as defined by the PostgreSQL system. This is a safeguard to ensure that the value meets all the data type's criteria before it is returned.

This double-check ensures that no invalid or out-of-range values are processed or returned, maintaining data integrity. We have done the same for **timestamptz_datetime2** and **timestamptz_datetime** functions as well.

contrib/babelfishpg_common/src/datetimeoffset.c
Anikait143: Done

test/JDBC/expected/getdate-vu-verify.out
```diff
@@ -77,3 +77,604 @@ int
...
+
+
+declare @x datetime2 = getdate() AT TIME ZONE 'Pacific Standard Time'
+declare @y datetime2 = getdate() AT TIME ZONE 'UTC'
+select DATEDIFF(MINUTE, @y, @x)
+go
+~~START~~
+int
+0
+~~END~~
```
tanscorpio7, line 196: What does this code block test ?

test/JDBC/expected/getdate-vu-prepare.out
```diff
@@ -137,3 +137,128 @@ AS
...
 GO
+
+CREATE PROCEDURE dbo.GetSysDatetimeDiff
+AS
+BEGIN
+    DECLARE @x datetime2 = SYSDATETIME();
+    select set_config('timezone', 'US/Pacific', false);
+    DECLARE @y datetime2 = SYSDATETIME()
+    select set_config('timezone', 'UTC', false);
+    SELECT DATEDIFF(MINUTE, @y, @x);
```
tanscorpio7, line 148: should this be set time zone -> declare datetime2 -> set time zone --> declare datetime2 ?

tanscorpio7: Or are we running the first line in default time zone ?

test/JDBC/input/getdate-before-15_9-or-16_5-vu-prepare.sql
```diff
@@ -0,0 +1,139 @@
...
+
+Create view sysdatetime_dep_view
+AS 
+    WITH
+        Pass0 as (select sys.sysdatetime() as C union all select sys.sysdatetime()), --2 rows
+        Pass1 as (select sys.sysdatetime() as C from Pass0 as A, Pass0 as B),--4 rows
+        Pass2 as (select sys.sysdatetime() as C from Pass1 as A, Pass1 as B),--16 rows
+        Pass3 as (select sys.sysdatetime() as C from Pass2 as A, Pass2 as B),--256 rows
+        Pass4 as (select sys.sysdatetime() as C from Pass3 as A, Pass3 as B),--65536 rows
+        Tally as (select row_number() over(order by C) as Number, min(C) over () as min_getdate from Pass4)
```
tanscorpio7, line 79: Could you explain this view definition

Anikait143: Pass0:
`select sys.sysdatetime() as C union all select sys.sysdatetime()`
This creates 2 rows of sys.sysdatetime().
Pass1:
`select sys.sysdatetime() as C from Pass0 as A, Pass0 as B`
The FROM Pass0 as A, Pass0 as B cross joins the 2 rows from Pass0 with itself, resulting in 4 rows.
Pass2:
`select sys.sysdatetime() as C from Pass1 as A, Pass1 as B`
This cross joins the 4 rows from Pass1 with itself, producing 16 rows.
Pass3:
`select sys.sysdatetime() as C from Pass2 as A, Pass2 as B`
Similarly, this cross joins the 16 rows from Pass2 with itself, producing 256 rows.
Pass4:
`select sys.sysdatetime() as C from Pass3 as A, Pass3 as B`
Here, the cross join of 256 rows results in 65,536 rows.
Tally:
`select row_number() over(order by C) as Number, min(C) over () as min_getdate from Pass4`
This query adds a row number for each of the 65,536 rows using the ROW_NUMBER() function, and it also computes the minimum sys.sysdatetime() value from the entire set using the MIN() window function.

Purpose:
The view tests the consistency or behavior of sys.sysdatetime() across a large number of rows generated via cross joins. By comparing the current sys.sysdatetime() with the minimum datetime value from the set, the query assess how consistent sys.sysdatetime() is during query execution.

This test previously existed in getdate tests just renamed them to run before this version.

test/JDBC/expected/getdate-vu-prepare.out
Anikait143: Yes, first line is running in default time zone -> set time zone -> declare @y -> reset time zone -> Take DateDiff

test/JDBC/expected/getdate-vu-verify.out
Anikait143: Both @x and @y will contain GETDATE() but adjusted to different time zones. However, this doesn't affect the DATE component. It will store the same time, just tagged with different time zones, so the DATEDIFF(MINUTE, @y, @x) will calculate the difference in minutes, but since the timestamps are essentially still the same, the difference is 0. So, we are just testing the integrity of timestamps here.

tanscorpio7 (approved):

contrib/babelfishpg_common/src/datetimeoffset.c
sumitj824: We don't have any check like first one in `timestamp_datetimeoffset`. Can you find any testcase where the first check will fail?

1. If yes, then Should we add similar checks in `timestamp_datetimeoffset` as well?
2. and if no then this check is redundant.

Anikait143: This is a sanity check required as internally used pg_localtime may produce invalid date due to inconsistencies in time zone. 
Time zone data issues: If there are issues or inconsistencies in the time zone data used by PostgreSQL, it can lead to invalid dates or times being produced by pg_localtime. 
So, this is only required when we are working with timezones and we donot require this in `timestamp_datetimeoffset`. Even PG itself uses this sanity checks, See: timestamp_trunc function of postgresql, So we should not avoid this check.

sumitj824: So, you mean PG's `statement_timestamp` function may produce invalid date/time due to inconsistencies in time zone. I don't agree with that, these sanity check are only required when we are modifying the time like `timestamp_trunc` or processing the user input. Here, it seems redundant to me.

sumitj824: Wait, these are not checks, we are finding the time and timezone using `timestamp2tm` and `tm2timestamp` function call so it seems good to me. 

The explanation provided earlier by you is completely wrong.

sumitj824 (approved):

contrib/babelfishpg_common/src/datetimeoffset.c
Anikait143: Nack, **That was pretty obvious from the function call itself that we are calculating time and timezone** from their, I just specified the need to throw the error when these functions return non-zero result and the scenario where they can return non-zero result, that was what was asked as a question in the first place. And yes, there can be inconsistencies while calculating time from time zone. For example:
`
Daylight Saving Time (DST) transitions: During the transition periods between Daylight Saving Time (DST) and standard time, there can be ambiguous or non-existent times. For example, when transitioning to DST, there is a "missing" hour, and when transitioning back to standard time, there is a "repeated" hour. 
`
In these cases, the pg_localtime function may produce an invalid date or time. So, that's why the error message is also important, we can't just rely on the function call only, which may result in wrong results.

tanscorpio7: `timestamp2tm` and `tm2timestamp` are converters but returns -1 on failure (value out of range), I think Anikait was referring to this as the first check.

Anikait143: Correct!!

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -936,22 +936,21 @@ Datum getutcdate(PG_FUNCTION_ARGS)
 
 Datum getdate_internal(PG_FUNCTION_ARGS)
 {
-	PG_RETURN_TIMESTAMP(DirectFunctionCall2(timestamp_trunc,CStringGetTextDatum("millisecond"),
-						PointerGetDatum(GetCurrentStatementStartTimestamp())));
+	PG_RETURN_DATUM(DirectFunctionCall1(common_utility_plugin_ptr->timestamptz_datetime,
```
Deepesh125, line 939: instead of calling it through DirectFunctionCall1, we should refactor original code and then call helper function directly from here. That will save lot of stuff related to security definer while invoking the function

Deepesh125 (changes requested):

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -936,22 +936,21 @@ Datum getutcdate(PG_FUNCTION_ARGS)
 
 Datum getdate_internal(PG_FUNCTION_ARGS)
 {
-	PG_RETURN_TIMESTAMP(DirectFunctionCall2(timestamp_trunc,CStringGetTextDatum("millisecond"),
-						PointerGetDatum(GetCurrentStatementStartTimestamp())));
+	PG_RETURN_DATUM(DirectFunctionCall1(common_utility_plugin_ptr->timestamptz_datetime, 
+						DirectFunctionCall2(timestamp_trunc,CStringGetTextDatum("millisecond"),
+											TimestampTzGetDatum(GetCurrentStatementStartTimestamp()))));
```
Deepesh125: `timestamp_trunc` returns `timestamp` whereas `timestamptz_datetime` is expecting `TimestampTz`, how is both compliant and working?

Anikait143: The compatibility between timestamp_trunc and timestamptz_datetime is achieved through implicit casting in PostgreSQL.

This implicit casting behavior is defined in the PostgreSQL source code and is part of the built-in type conversion rules. It allows for seamless interoperability between timestamp and TimestampTz values in function calls and expressions.
The implicit cast from timestamp to TimestampTz preserves the original time value but adds or removes the time zone offset based on the current time zone setting. This ensures that the timestamptz_datetime function receives a valid TimestampTz value, even though the timestamp_trunc function returns a timestamp value.

But yes, we shouldn't rely on implicit casting, Used `timestamptz_trunc` instead of `timestamp_trunc` to correct this.

Deepesh125 (approved):

PR #2930, opened 2024-09-12
Fix dangling file storage when aborting a transaction with a temp table.

Sairakan:
### Description

There were some cases when a transaction or procedure was aborted, and 
the transaction/procedure created a temp table, where the physical on-disk file 
associated with the temp table was not properly being deleted. This commit fixes the
behavior by triggering a manual call to smgrDoPendingDeletes() (with the
appropriate isCommit flag) when necessary during transaction aborts.

### Issues Resolved

BABEL-4751

* **Use case based -**
The issue is due to orphaned files, so it cannot be tested through SQL tests. Instead, these changes were manually verified by running tests which previously would leave orphaned files on disk, and verified that after these changes there are no longer orphaned files left on disk. There are also some additional tests where temp and permanent tables are both used within an aborted transaction to confirm that this change does not cause regressions for those use cases.

* **Boundary conditions -**

* **Arbitrary inputs -**

* **Negative test cases -**

* **Performance tests -**

* **Tooling impact -**

* **Client tests -**

### Check List

### DIFF
--- a/contrib/babelfishpg_tsql/src/pl_exec-2.c
+++ b/contrib/babelfishpg_tsql/src/pl_exec-2.c
@@ -1686,14 +1686,14 @@ execute_batch(PLtsql_execstate *estate, char *batch, InlineCodeBlockArgs *args, 
 	PG_CATCH();
 	{
 		/* Delete temporary tables as ENR */
-		pltsql_remove_current_query_env();
+		pltsql_remove_current_query_env(true);
 
 		PG_RE_THROW();
 	}
 	PG_END_TRY();
 
 	/* Delete temporary tables as ENR */
-	pltsql_remove_current_query_env();
+	pltsql_remove_current_query_env(false);
 
 	after_lxid = MyProc->lxid;
 
--- a/contrib/babelfishpg_tsql/src/pl_handler.c
+++ b/contrib/babelfishpg_tsql/src/pl_handler.c
@@ -30,6 +30,7 @@
 #include "catalog/pg_type.h"
 #include "catalog/pg_default_acl.h"
 #include "catalog/pg_shdepend.h"
+#include "catalog/storage.h"
 #include "commands/createas.h"
 #include "commands/dbcommands.h"
 #include "commands/defrem.h"
@@ -5224,7 +5225,7 @@ pltsql_call_handler(PG_FUNCTION_ARGS)
 
 		func->cur_estate = save_cur_estate;
 
-		pltsql_remove_current_query_env();
+		pltsql_remove_current_query_env(send_error);
 		pltsql_revert_guc(save_nestlevel);
 		pltsql_revert_last_scope_identity(scope_level);
 	}
@@ -6498,7 +6499,7 @@ set_current_query_is_create_tbl_check_constraint(Node *expr)
 }
 
 void
-pltsql_remove_current_query_env(void)
+pltsql_remove_current_query_env(bool is_abort)
 {
 	bool old_abort_curr_txn = AbortCurTransaction;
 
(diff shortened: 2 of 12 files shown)


contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -6508,6 +6509,8 @@ pltsql_remove_current_query_env(void)
 		AbortCurTransaction = false;
 
 		ENRDropTempTables(currentQueryEnv);
+		if (is_abort)
+			smgrDoPendingDeletes(!old_abort_curr_txn);
```
timchang514, line 6513: Just making sure that this doesn't interfere with the custom code added for TV handling in RelationDropStorage?

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -1686,14 +1686,14 @@ execute_batch(PLtsql_execstate *estate, char *batch, InlineCodeBlockArgs *args,
 	PG_CATCH();
 	{
 		/* Delete temporary tables as ENR */
-		pltsql_remove_current_query_env();
+		pltsql_remove_current_query_env(true);
```
timchang514, line 1689: Is this guaranteed to be ABORT case?

contrib/babelfishpg_tsql/src/pl_handler.c
Sairakan: No, this should not have any effect. Our TV handling just makes sure that TVs always get dropped no matter what, whether we abort or commit the transaction - this change won't affect that behavior.

contrib/babelfishpg_tsql/src/pl_exec-2.c
Sairakan: To be honest, I am not 100% sure, but I would assume that any code path within `pltsql_inline_handler()` which throws an error would be trying to abort the transaction.

contrib/babelfishpg_tsql/src/pl_handler.c
lejaokri: Hmm. `smgrDoPendingDeletes()` would handle non-temp tables also, correct? Did we test the case where `smgrDoPendingDeletes` could potentially drop files for non-temporary tables? My main concern is that the tables we're dropping in `ENRDropTempTables` may not match exactly with what `smgrDoPendingDeletes` is deleting.

Sairakan: Yes, it is technically possible - however, this code path is only hit at the end of either a stored procedure call or a batch. Therefore, it will follow the normal transaction semantics under most circumstances, and only need to do this special cleanup at transaction boundaries. I believe that this is backed up by the fact the fact that our full JDBC test suite succeeded. I'll try some ad-hoc testing to exercise this further, but I am confident in this solution.

Sairakan: I did some testing mixing temp and permanent tables, and they all behaved normally:
```
1> create table t1 (a int)
2> go
1> create proc proc1 as begin; declare @tv table (a int); create table #t1 (a int); insert into t1 values (1); insert into @tv values (1); insert into #t1 values (1); end;
2> go
1> begin tran
2> insert into t1 values (2)
3> exec proc1
4> select * from t1
5> drop table t1
6> rollback
7> go

(1 rows affected)

(1 rows affected)

(1 rows affected)

(1 rows affected)
a          
-----------
          2
          1

(2 rows affected)
1> select * from t1
2> go
a          
-----------

(0 rows affected)
1> begin tran
2> insert into t1 values (2)
3> create table #outer_table (a int)
4> insert into #outer_table values (1)
5> exec proc1
6> select * from t1
7> drop table t1
8> rollback
9> go

(1 rows affected)

(1 rows affected)

(1 rows affected)

(1 rows affected)

(1 rows affected)
a          
-----------
          2
          1

(2 rows affected)
1> select * from t1
2> go
a          
-----------

(0 rows affected)
1> select * from #outer_table
2> go
relation "#outer_table" does not exist
```

lejaokri (approved):

2jungkook (approved):

PR #2935, opened 2024-09-16
Add '\' as ESCAPE character for sp_columns_100 for LIKE operator

ahmed-shameem:
### Description
There are few occurrences where we use procedure  sp_columns_100  to fetch column metadata. The issue with this procedure is, when the table name contains _ character, we may obtain some unexpected output . This is because in some places we are using using sp_columns_100 incorrectly with a backslash escaping underscores (e.g. “\_“). Underscores are wildcards in sp_columns_100 (and in LIKE in general), but sp_columns_100 does not use a backslash as an escape character.
Till now, sp_columns_100 , was invoking a plpgsql function sp_column_100_internal which used PG’s SIMILAR TO  to operator, which does use a backslash as the default escape character. 

With BABEL-4953, we have updated the definition of sp_columns_100 where we removed redundant function call and implemented the procedure by using TSQL LIKE operator on column name, which will create some problems in other places.

The change in the new version is correct. However, since currently we rely on the incorrect functionality of the escape characters, we are making this temporary change for now. Once, we have fixed the dependencies on this procedure, we need to revert it.

Authored-by: Shameem Ahmed [shmeeh@amazon.com](mailto:shmeeh@amazon.com)

* **Use case based -**

* **Boundary conditions -**

* **Arbitrary inputs -**

* **Negative test cases -**

* **Performance tests -**

* **Tooling impact -**

* **Client tests -**

### Check List

### DIFF
--- a/contrib/babelfishpg_tsql/sql/babelfishpg_tsql.sql
+++ b/contrib/babelfishpg_tsql/sql/babelfishpg_tsql.sql
@@ -500,7 +500,7 @@ BEGIN
 				END
 			) as SS_DATA_TYPE
 		from sys.sp_columns_100_view
-		where pg_catalog.lower(table_name) like pg_catalog.lower(sys.babelfish_truncate_identifier(@table_name)) COLLATE database_default
+		where pg_catalog.lower(table_name) like pg_catalog.lower(sys.babelfish_truncate_identifier(@table_name)) COLLATE database_default ESCAPE '\'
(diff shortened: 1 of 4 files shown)


Deepesh125 (dismissed):

farrest (commented):
Workaround for DMS requiring the '\' escape character.

farrest (commented):
This looks good to me. Satisfied the tests should cover the use case of the '\' escape character

Deepesh125 (approved):

PR #2939, opened 2024-09-17
Fix hex string to varchar conversion

rishabhtanwar29:
### Description
Varbinary to varchar conversion results in error if the binary input contains non-printable
UTF8 characters or NULL character. PG natively supports such behavior of throwing error
for untranslatable characters but this is not true for T-SQL. T-SQL just keeps all unrecognized
characters bytes into the resultant varchar string instead of throwing error so that it is upon
client to handle them, most of the clients replace unrecognized characters with a`?`.

This commit fixes this issue by handling non-printable and NULL characters in a string instead
of throwing error. This is achieved by keeping the unrecognized bytes as it is in the output
string so that it is upon client to handle such characters. This is important to note that the PG's
internal logic to handle NULL (0x00) character remains same so any string operations on strings
containing NULL character might generate different result compared to T-SQL standard.

Task: BABEL-4562

* **Use case based -**

* **Boundary conditions -**

* **Arbitrary inputs -**

* **Negative test cases -**

* **Performance tests -**

* **Tooling impact -**

* **Client tests -**

### Check List


contrib/babelfishpg_common/src/encoding/mb/conv.c
```diff
@@ -225,17 +225,23 @@ TsqlUtfToLocal(const unsigned char *utf, int len,
 		unsigned char b2 = 0;
 		unsigned char b3 = 0;
 		unsigned char b4 = 0;
+		const unsigned char *cur;
 
-		/* "break" cases all represent errors */
+		/* Handle NULL character */
 		if (*utf == '\0')
-			break;
+		{
```
Deepesh125, line 232: These routines are being used to send the data of string data type with different collation to end client, how are we making sure that this change will not regress anything existing?

contrib/babelfishpg_common/src/encoding/mb/conv.c
```diff
@@ -271,6 +277,12 @@ TsqlUtfToLocal(const unsigned char *utf, int len,
 		}
 		iutf = (b1 << 24 | b2 << 16 | b3 << 8 | b4);
 
+		if (!pg_utf8_islegal(cur, l))
```
Deepesh125, line 280: Just curious, how are we making sure that our test cases covers case which hits "break" previously?

contrib/babelfishpg_tds/src/backend/tds/tdstypeio.c
```diff
@@ -2778,7 +2778,19 @@ TdsSendTypeVarchar(FmgrInfo *finfo, Datum value, void *vMetaData)
 			   *buf = OutputFunctionCall(finfo, value);
 	TdsColumnMetaData *col = (TdsColumnMetaData *) vMetaData;
 
-	len = strlen(buf);
+	/*
+	 * Get the actual length of varchar field from Bytea struct and use that length
+	 * instead of plain strlen since strlen calculates string upto first null character (\0)
+	 * even though the input string might contain more characters.
+	 */
+	if (pltsql_plugin_handler_ptr->is_tsql_varchar_or_char_datatype(col->pgTypeOid))
```
Deepesh125, line 2786: why do we need this check?

contrib/babelfishpg_common/src/varbinary.c
```diff
@@ -788,17 +788,6 @@ varbinaryvarchar(PG_FUNCTION_ARGS)
...
-
-	/*
-	 * Allow trailing null bytes 
-	 * Its safe since multi byte UTF-8 does not contain 0x00 
-	 * This is needed since we implicity add trailing zeroes to 
-	 * binary type if input is less than binary(n)
-	 * ex: CAST(CAST('a' AS BINARY(10)) AS VARCHAR) should work
-	 * and not fail because of null byte
-	 */
-	while(len>0 && data[len-1] == '\0')
```
Deepesh125, line 800: it makes sense if we dont remove trailing zeroes for binary, but we should remove them for varbinary, right? what is expected tsql behaviour?

test/JDBC/upgrade/15_2/schedule
```diff
@@ -437,7 +437,6 @@ BABEL-4672
 permission_restrictions_from_pg
 BABEL-730-before-15_6-or-16_1
 GRANT_SCHEMA-before-15_7-16_3
-babel-4475
```
Deepesh125, line 440: why?

test/JDBC/input/functions/string_functions/replace-before-15_8-or-16_4-vu-verify.sql
```diff
@@ -198,6 +198,9 @@ DECLARE @inputString BINARY(10) = 0x61626364656667, @pattern BINARY(10) = 0x6364
 SELECT replace(@inputString, @pattern, @replacement)
 GO
 
+UPDATE babel_4836_replace_t4 SET a = 0x6162636465, b = 0x6263, c = 0x747576;
```
Deepesh125, line 201: why is this needed?

rishabhtanwar29:
Varbinary to varchar conversion results in error for non-printable UTF8 characters or NULL character. PG natively supports such behavior of throwing error for untranslatable characters but this is not true for T-SQL. T-SQL just keeps all unrecognized characters bytes into the resultant varchar string instead of throwing error. I modified the PG behavior to not throw error in such cases to match the T-SQL behavior but looks like just not throwing the error is not sufficient as we also need to fix the logic of how PG handles strings since PG's internal modules doesn't expect these untranslatable characters in the input string as well treats NULL character as the end of the string. This will require tremendous amount of changes in the native PG code wherever string related handling is involved which doesn't seem feasible and straightforward for now.

PR #2949, opened 2024-09-19
Fix difference in behaviour for CONCAT() and CONCAT_WS() functions with multibyte characters

Anikait143:
## Description

### 1. Issues
- Arguments datatype and return datatype of CONCAT() and CONCAT_WS() string function is not defined correctly in babelfish, which is causing  output difference between TSQL and Babelfish.

- CONCAT() and CONCAT_WS() functions does not raise error when called with < 2 or < 3 arguments respectively.

- BABEL-4299 missed to add SQL changes to upgrade script, as a result after upgrade to red panda or later, the definition will remain same as the previous one and there would be no performance improvement.

- Added wrapper function handling for CONCAT() and CONCAT_WS() function to ensure correct input arguments datatype and return datatype.
- Also ensured to throw appropriate error when CONCAT() and CONCAT_WS() functions are called with < 2 or < 3 arguments respectively.

```
# table setup
1> create table a([AccName] nvarchar(60) NOT NULL)
2> insert into a values ('资产'), ('固定资产'), ('流动资产'), ('负债'), ('流动负债')
3> go
```
- CONCAT()
```
1> select concat(accname, '-') as con_str2 from a
2> go
con_str2                                                                                                                                                                                                                                                        
??-                                                                                                                                                                                                                                                             
????-                                                                                                                                                                                                                                                           
????-                                                                                                                                                                                                                                                           
??-                                                                                                                                                                                                                                                             
????-                                                                                                                                                                                                                                                           

(5 rows affected)
```
```
1> select concat(accname, '-') as con_str2 from a
2> go
con_str2                                                                                                                                                                                                                                                        
资产-                                                                                                                                                                                                                                                             
固定资产-                                                                                                                                                                                                                                                           
流动资产-                                                                                                                                                                                                                                                           
负债-                                                                                                                                                                                                                                                             
流动负债-                                                                                                                                                                                                                                                           

(5 rows affected)
```
- CONCAT_WS()
```
1> select concat_ws('-', accname, '-') as con_str2 from a
2> go
con_str2                                                                                                                                                                                                                                                        
??--                                                                                                                                                                                                                                                             
????--                                                                                                                                                                                                                                                           
????--                                                                                                                                                                                                                                                           
??--                                                                                                                                                                                                                                                             
????--                                                                                                                                                                                                                                                           

(5 rows affected)
```
```
1> select concat_ws('-', accname, '-') as con_str2 from a
2> go
con_str2                                                                                                                                                                                                                                                        
资产--                                                                                                                                                                                                                                                             
固定资产--                                                                                                                                                                                                                                                           
流动资产--                                                                                                                                                                                                                                                           
负债--                                                                                                                                                                                                                                                             
流动负债--                                                                                                                                                                                                                                                           

(5 rows affected)
```
- Error message for invalid no. of arguments
```
1> select concat('a')
2> go
concat
----------------
a

1> select concat()
2> go
Msg 33557097, Level 16, State 1, Server persist-rcv-25oct-14-6-0-7-b, Line 1
function concat() does not exist

1> select concat_ws(' ', 'a')
2> go
concat_ws
----------------
a

1> select concat_ws()
2> go
Msg 33557097, Level 16, State 1, Server persist-rcv-25oct-14-6-0-7-b, Line 1
function concat_ws() does not exist
```
```
1> select concat('a')
2> go
The concat function requires 2 to 254 arguments.

1> select concat()
2> go
The concat function requires 2 to 254 arguments.

1> select concat_ws(' ', 'a')
2> go
The concat_ws function requires 3 to 254 arguments.

1> select concat_ws()
2> go
The concat_ws function requires 3 to 254 arguments.
```
- Added missed upgrade changes of BABEL-4299 to upgrade script.
- Added Test Cases for the issues


Task: BABEL- 3409, BABEL- 3648


* **Use case based -**   YES


* **Arbitrary inputs -** YES


* **Performance tests -** NO


* **Tooling impact -** NO

* **Memory tests -** NO


* **Client tests -** NO



### Check List


contrib/babelfishpg_tsql/sql/sys_functions.sql
```diff
@@ -4001,6 +4001,84 @@ BEGIN
 END;
 $$ LANGUAGE plpgsql IMMUTABLE STRICT PARALLEL SAFE;
 
+-- wrapper functions for concat --
+CREATE OR REPLACE FUNCTION sys.concat(VARIADIC args sys.VARCHAR[] DEFAULT '{}')
```
ahmed-shameem, line 4005: We have define the input datatypes as varchar and nvarchar, have we tested with other datatypes like date, datetime, time, smalldatetime, datetime2, datetimeoffset. Will these datatype work?

contrib/babelfishpg_tsql/sql/sys_functions.sql
```diff
@@ -4001,6 +4001,84 @@ BEGIN
...
+RETURNS sys.VARCHAR
+AS $$
+DECLARE
+    arr_len INTEGER;
+BEGIN
+    arr_len := array_length(args, 1);
+
+    -- Defined as per TSQL definition but PG has limitation for max number of args = 100
+    IF arr_len IS NULL OR arr_len < 2 OR arr_len > 254 THEN
```
ahmed-shameem: So if I supply more than 100 args to sys.concat, it will work as we are using ARRAY_TO_STRING function instead of concat of PG, right?

contrib/babelfishpg_tsql/sql/sys_functions.sql
```diff
@@ -4001,6 +4001,84 @@ BEGIN
...
+    IF arr_len < 2 OR arr_len > 254 THEN
+        RAISE EXCEPTION 'The concat function requires 2 to 254 arguments.';
+    END IF;
+
+    RETURN (PG_CATALOG.ARRAY_TO_STRING(args, ''));
+END;
+$$ LANGUAGE plpgsql STABLE PARALLEL SAFE;
+
+-- wrapper functions for concat_ws --
+CREATE OR REPLACE FUNCTION sys.concat_ws(seperator sys.VARCHAR DEFAULT '', VARIADIC args sys.VARCHAR[] DEFAULT '{}')
```
ahmed-shameem, line 4040: Same applies here as well.

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--3.7.0--3.8.0.sql
```diff
@@ -68,6 +68,7324 @@ BEGIN
 END;
 $$ LANGUAGE plpgsql;
 
+CREATE OR REPLACE FUNCTION sys.babelfish_conv_datetime_to_string(IN p_datatype TEXT,
```
ahmed-shameem, line 79: Why is this change included? Is it to address BABEL-4299 upgrade issue? Do we need all these changes?

contrib/babelfishpg_tsql/src/pltsql_coerce.c
```diff
@@ -323,6 +323,10 @@ tsql_cast_raw_info_t tsql_cast_raw_infos[] =
 	{TSQL_CAST_WITHOUT_FUNC_ENTRY, "pg_catalog", "bytea", "pg_catalog", "text", NULL, 'i', 'i'},
 	{TSQL_CAST_WITHOUT_FUNC_ENTRY, "sys", "datetimeoffset", "pg_catalog", "text", NULL, 'i', 'i'},
 	{TSQL_CAST_WITHOUT_FUNC_ENTRY, "pg_catalog", "time", "pg_catalog", "text", NULL, 'i', 'i'},
+/*  date/time -> string via I/O */
+	{TSQL_CAST_WITHOUT_FUNC_ENTRY, "pg_catalog", "time", "sys", "varchar", NULL, 'i', 'i'},
+	{TSQL_CAST_WITHOUT_FUNC_ENTRY, "pg_catalog", "date", "sys", "varchar", NULL, 'i', 'i'},
+	{TSQL_CAST_WITHOUT_FUNC_ENTRY, "sys", "datetimeoffset", "sys", "varchar", NULL, 'i', 'i'},
```
ahmed-shameem, line 329: Ok, so we have added these casts so the functions will work the source datatype. Please check for the other types as well.

contrib/babelfishpg_tsql/src/pltsql_coerce.c
```diff
@@ -1124,6 +1131,93 @@ validate_special_function(char *func_nsname, char *func_name, List* fargs, int n
...
+			if (nargs < 2 || nargs > 254)
+				ereport(ERROR,
+						(errcode(ERRCODE_UNDEFINED_FUNCTION),
+							errmsg("The concat function requires 2 to 254 arguments.")));
+			
+			for (int i = 0; i < nargs; i++)
+			{
+				if (input_typeids[i] == UNKNOWNOID)
+				{
+					Node *arg = (Node *) lfirst(list_nth_cell(fargs, i));
```
ahmed-shameem: nit: indentation seems a little off

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -2301,24 +2301,24 @@ pltsql_report_proc_not_found_error(List *names, List *fargs, List *given_argname
 	{
 		const char *arg_str = (max_nargs < 2) ? "argument" : "arguments";
 
+		if (!proc_call)
```
ahmed-shameem, line 2394: Why did we redefine the code block here?

ahmed-shameem (changes requested):

contrib/babelfishpg_tsql/sql/sys_functions.sql
Anikait143: Yes, we have already added tests with all datatypes, and all datatypes other than sql_variant, xml, geometry, geography will work for concat and concat_ws functions.

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--3.7.0--3.8.0.sql
Anikait143: yes, this is to address BABEL-4299 upgrade issue. Yes, we need these changes as Concat was previously used in babelfish_conv_datetime_to_string and without this older definition will still exist and concat will throw an error.

contrib/babelfishpg_tsql/src/hooks.c
Anikait143: This is because in case of CONCAT() and CONCAT_WS() since we use variadic function argument, the no. of arguments can mismatch with the definition, resulting in `found` to be false, but we need to run our special function handling in that case as well to throw appropriate error message, so moved it out of that block.

contrib/babelfishpg_tsql/src/pltsql_coerce.c
Anikait143: Only these 3 IMPLICIT Casts to VARCHAR were missing rest all are already there according to TSQL. Also tests are there to verify the same.

contrib/babelfishpg_tsql/sql/sys_functions.sql
Anikait143: The error is thrown by PG itself even before it reaches here, so it will not work, but amended the error message to allow a maximum of 100 arguments to be consistent with PG.

contrib/babelfishpg_tsql/sql/sys_functions.sql
Anikait143: Replied above.

ahmed-shameem (approved):

contrib/babelfishpg_tsql/sql/sys_functions.sql
Deepesh125, line 4005: why is default required?

contrib/babelfishpg_tsql/sql/sys_functions.sql
```diff
@@ -4001,6 +4001,84 @@ BEGIN
...
+    arr_len := array_length(args, 1);
+
+    -- PG has limitation for max number of args = 100
+    IF arr_len IS NULL OR arr_len < 2 OR arr_len > 100 THEN
+        RAISE EXCEPTION 'The concat function requires 2 to 100 arguments.';
+    END IF;
+
+    RETURN (PG_CATALOG.ARRAY_TO_STRING(args, ''));
+END;
+$$ LANGUAGE plpgsql STABLE PARALLEL SAFE;
```
Deepesh125, line 4020: why not pltsql?

contrib/babelfishpg_tsql/src/pltsql_coerce.c
```diff
@@ -1124,6 +1131,93 @@ validate_special_function(char *func_nsname, char *func_name, List* fargs, int n
...
+		else if (strlen(func_name) == 6 && strncmp(func_name, "concat", 6) == 0)
+		{
+			Oid		target_type_id = get_sys_varcharoid();
+			/* PG has limitation for max number of args = 100. */
+			if (nargs < 2 || nargs > 100)
+				ereport(ERROR,
+						(errcode(ERRCODE_UNDEFINED_FUNCTION),
+							errmsg("The concat function requires 2 to 100 arguments.")));
+			
+			for (int i = 0; i < nargs; i++)
```
Deepesh125: can we have arguments with different data types? If so then please point out test cases and if not so then we can break this loop

contrib/babelfishpg_tsql/sql/sys_functions.sql
Anikait143: Since without `default` this function would consider atleast one argument to hit this definition, so case like `CONCAT()` will give error `CONCAT function requires 1 argument` without default which would be misleading. So, marking the argument as `DEFAULT` will make it hit the definition and throw a more relevant error: `The concat function requires 2 to 100 arguments.`

contrib/babelfishpg_tsql/src/pltsql_coerce.c
Anikait143: Yes, we can have arguments with different data types. We have various test case added for the same. Please refer to prepare and verify files. One such example: https://github.com/amazon-aurora/babelfish_extensions/blob/4ad48c68062d72aa58bb98e1f12e6961f382e67d/test/JDBC/expected/concat-vu-prepare.out#L472

contrib/babelfishpg_tsql/sql/sys_functions.sql
Anikait143: PL/TSQL does not support variadic functions or parameters.
Variadic functions are a feature of PostgreSQL's PL/pgSQL which allow us to define a function that can accept a variable number of arguments which is a necessity in defining the CONCAT and CONCAT_WS functions which allow variable number of arguments. So, prefered to use PL/pgSQL. Also previously as well we have used PL/pgSQL to define all such functions.

contrib/babelfishpg_tsql/src/pltsql_coerce.c
rohit01010: For cast from `datetime/datetime2/smalldatetime` to `sys.varchar` we have defined explicit functions such as `datetime_varchar`/`datetime2_varchar`/ `smalldatetime_varchar`. Here for `date, time and datetimeoffset` we are adding a cast entry without function, so this cast will use `sys.varchar` input function to convert the `date/time/datetimeoffset` to `sys.varchar`. How are we making sure that for these cast entries we added, we do not need any explicit conversion function similar to what we used for `datetime/datetime2/smalldatetime`.

test/JDBC/input/BABEL-1994-CHAR.sql
```diff
@@ -12,13 +12,13 @@ create index pad_c5nn_idx on pad(c5nn);
...
     '[' + c5nn +']' as c5nn_to_text_concat,
-    concat('[', cast(c5nn as varchar), ']') as c5nn_to_varchar_concat,
-    concat('[', c5n, ']') as c5n_concat,
+    pg_catalog.concat('[', cast(c5nn as varchar), ']') as c5nn_to_varchar_concat,
+    pg_catalog.concat('[', c5n, ']') as c5n_concat,
     '[' + c5n + ']' as c5n_to_text_concat,
-    concat('[', cast(c5n as varchar), ']') as c5n_to_varchar_concat,
-    concat('[', cast(c5n as pg_catalog.name), ']') as c5n_to_name_concat
+    pg_catalog.concat('[', cast(c5n as varchar), ']') as c5n_to_varchar_concat,
+    pg_catalog.concat('[', cast(c5n as pg_catalog.name), ']') as c5n_to_name_concat
```
rohit01010: why are we changing `concat` to `pg_catalog.concat` in test file ?

Deepesh125 (dismissed):

shah-nirmit (approved):

contrib/babelfishpg_tsql/src/pltsql_coerce.c
```diff
@@ -1124,6 +1131,93 @@ validate_special_function(char *func_nsname, char *func_name, List* fargs, int n
...
 		}
+		else if (strlen(func_name) == 6 && strncmp(func_name, "concat", 6) == 0)
+		{
+			Oid		target_type_id = get_sys_varcharoid();
+			/* PG has limitation for max number of args = 100. */
+			if (nargs < 2 || nargs > 100)
+				ereport(ERROR,
+						(errcode(ERRCODE_UNDEFINED_FUNCTION),
+							errmsg("The concat function requires 2 to 100 arguments.")));
+
```
rohit01010: Suggestion: 
This implementation for supporting variadic functions(concat & concat_ws) looks messy, very specific, repetitive code and hard to maintain. We can extend our existing logic for special functions to support variadic special functions in a generic way, as follows
- Add a variable is_variadic in tsql_special_function struct to identify the variadic special function as follows
```
typedef struct tsql_special_function
{
	const char             *nsp;                              /* namespace of special function */
	const char             *funcname;                         /* name of special function */
	const char             *formatted_funcname;				  /* formatted name of special function */
	bool					is_variadic;
	int                     nargs;                            /* number of arguments of special function (for variadic function number of fixed arguments will be stored) */
	tsql_valid_arg_type_t   valid_arg_types[SFUNC_MAX_ARGS];  /* list for storing details of all the valid types supported for each arguments (for variadic function valid types of only fixed arguments will be stored) */
} tsql_special_function_t;
```
- For variadic function we can specify the fixed number of arguments and their details as follows
```
{"sys", "concat", "concat", true, 0, {}},
{"sys", "concat_ws", "concat_ws", true, 1, {{6, {"char","varchar","nchar","nvarchar","text","ntext"}, {InvalidOid, InvalidOid, InvalidOid, InvalidOid, InvalidOid, InvalidOid}}}}
```
- Instead of having different if blocks for almost same logic, a function can be created to validate variadic argument types, as follows. This function will be called only when special_func is variadic
```
static bool
validate_variadic_special_function_args(char *func_name, List* fargs, int nargs, int nfixed_args, Oid *input_typeids)
{
	Oid		input_type_id, base_type_id;
	Oid		sys_varcharoid = get_sys_varcharoid();

	if ((strlen(func_name) == 6 && strncmp(func_name, "concat", 6) == 0) ||
		 (strlen(func_name) == 9 && strncmp(func_name, "concat_ws", 9) == 0))
	{
		for (int i = nfixed_args; i < nargs; i++)
		{
			if (input_typeids[i] == UNKNOWNOID)
			{
				Node *arg = (Node *) lfirst(list_nth_cell(fargs, i));
				if (IsA(arg, Const) && ((Const *)arg)->constisnull)
					continue;
				else
					input_type_id = sys_varcharoid;
			}
			else
				input_type_id = input_typeids[i];

			base_type_id = get_immediate_base_type_of_UDT_internal(input_type_id);
			if (OidIsValid(base_type_id))
				input_type_id = base_type_id;

			/* Check if the input type is implicitly castable to varchar. */
			if (!can_coerce_type(1, &input_type_id, &sys_varcharoid, COERCION_IMPLICIT))
				ereport(ERROR,
					(errcode(ERRCODE_UNDEFINED_FUNCTION),
						errmsg("Implicit conversion from data type %s to varchar is not allowed.", format_type_be(input_typeids[i]))));
		}
		return true;
	}
	return false;
}
```

contrib/babelfishpg_tsql/src/pltsql_coerce.c
Anikait143: Earlier, we didn't allow implicit casts from `date, time and datetimeoffset` to `varchar` whereas TSQL does, Although we already have casts from these functions to `TEXT` without function, so I don't think there would any issue here, but in future if there is need for adding functions for these casts then we can do it as a part of incremental development.

contrib/babelfishpg_tsql/src/pltsql_coerce.c
Anikait143: Done

test/JDBC/input/BABEL-1994-CHAR.sql
Anikait143: `name` datatype is not castable to varchar so it will throw an error with new definitions, so appended PG_CATALOG in that to preserve the test, although this is not required at other places, so removed that.

rohit01010 (approved):

Deepesh125 (dismissed):

Deepesh125 (approved):

PR #2950, opened 2024-09-19
Fix crash when trying to alter temp table column type.

Sairakan:
### Description

In cases where a temp table column is being altered so that only the size changes (e.g. char(20) -> char(5)), the code path needs to drop the pg_depend entry from the normal catalog table (to be later recreated using the new target type). However, for ENR temp tables all catalog tuples are stored in currentQueryEnv and not actually in the catalogs themselves. Thus, the code would not be able to find the tuple for deletion, leading to an assertion error. Fix the issue by using ENRDropTuple() instead of CatalogTupleDelete() so that the code drops the tuple in the correct place.

### Issues Resolved

BABEL-5273

* **Use case based -**
Tests altering temp table column types
```
ALTER TABLE #t1 ALTER COLUMN b char(5)
```

* **Boundary conditions -**

* **Arbitrary inputs -**

* **Negative test cases -**
Try to alter a temp table column type to a smaller type when there is already data in the table larger than the new type would allow (raises an error)
```
ALTER TABLE #t1 ALTER COLUMN b char(4)
GO
~~ERROR (Code: 8152)~~

~~ERROR (Message: value too long for type character(4))~~
```

* **Performance tests -**

* **Tooling impact -**

* **Client tests -**

### Check List


test/JDBC/expected/temp_table.out
```diff
@@ -63,6 +63,32 @@ int#!#varchar#!#int
 ~~END~~
 
 
+-- BABEL-5273 ALTER COLUMN to another char type
```
timchang514: can we add some tests alter column for other types? with an empty table - like some of the other char types like nchar/nvarchar, money, datetime, int?

test/JDBC/expected/temp_table.out
```diff
@@ -63,6 +64,176 @@ int#!#varchar#!#int
...
+ALTER TABLE #t1 ALTER COLUMN b SMALLDATETIME
+GO
+
+ALTER TABLE #t1 ALTER COLUMN b CHAR
+GO
+
+ALTER TABLE #t1 ALTER COLUMN b VARCHAR
+GO
+
+-- TODO: fix this, it should work and not raise a syntax error
```
timchang514, line 176: should this be filed in a separate JIRA? is this temp table specific or just general issue?

test/JDBC/expected/temp_table.out
```diff
@@ -63,6 +64,176 @@ int#!#varchar#!#int
...
+
+-- should raise error due to incompatible types
+ALTER TABLE #t1 ALTER COLUMN b GEOMETRY
+GO
+~~ERROR (Code: 33557097)~~
+
+~~ERROR (Message: column "b" cannot be cast automatically to type geometry)~~
+
+
+-- TODO: fix this, it should work and not raise a syntax error
```
timchang514, line 229: same q as above

timchang514 (approved):

test/JDBC/expected/temp_table.out
Sairakan: I believe this is a general issue with altering table datatypes. I'll file a JIRA for this.

lejaokri (approved):

PR #2970, opened 2024-09-25
Support fixed database role db_accessadmin

tanscorpio7:
### Description

Support fixed database role db_accessadmin in babelfish.
Member of db_accessadmin can do the following
1. CREATE SCHEMA
2. CREATE/DROP USER
3. ALTER USER
4. They always have connect permission

### Issues Resolved

[BABEL-5136]

### Check List


Imran-imtiaz48 (commented):
### Review Summary

This update appears to refine the logic related to parsing and validating statements for different user roles (guest and owner) and the creation of database roles. It includes adjustments to the number of expected statements based on login status and the role of the user, along with improvements to the role-granting process.

### Code Review

1. **Statement Parsing Logic**:
   - The adjustment to `expected_stmt_num` based on whether the user is a guest or the owner is well thought out. It accounts for different conditions like the presence of logins and the role of the owner, adding clarity and control over how many statements are expected to be parsed.
   - One suggestion here is to add comments or documentation around the conditions that differentiate between when an owner is `sa` or not, to improve readability.

   ```c
   if (!owner_is_sa)
       expected_stmt_num = list_length(logins) > 0 ? 12 : 11;	
   else
       expected_stmt_num = list_length(logins) > 0 ? 11 : 10;
   ```

2. **Role Assignment**:
   - Granting the `dbo` role to the owner only if they are not `sa` adds an extra layer of security and role distinction, which is important for preventing unnecessary privilege escalation.
   - The use of `update_GrantRoleStmt` and `update_CreateRoleStmt` functions is logical and structured, ensuring that roles are correctly assigned and updated.

   ```c
   if (!owner_is_sa)
   {
       /* Grant dbo role to owner */
       stmt = parsetree_nth_stmt(res, i++);
       update_GrantRoleStmt(stmt, list_make1(make_accesspriv_node(dbo)),
                           list_make1(make_rolespec_node(owner)));
   }
   ```

   This logic is solid, but for added clarity, consider extracting the `if (!owner_is_sa)` block into a function to further modularize the code.

3. **Error Handling**:
   - The error reporting for mismatched statement counts is correctly implemented. However, a more informative error message would be helpful. Instead of a generic `ERROR`, provide details on the expected versus actual number of statements, which would make debugging easier.

   ```c
   if (list_length(res) != expected_stmt_num)
       ereport(ERROR, (errmsg("Expected %d statements, but found %d", expected_stmt_num, list_length(res))));
   ```

4. **General Improvements**:
   - It's great that the changes focus on proper role handling and enforcing statement expectations.
   - If possible, avoid magic numbers such as `12`, `11`, `10`, and `8`. It would enhance code maintainability if these values were replaced by well-named constants or enums to represent the different scenarios.

### Recommendations

- Add inline comments or documentation for more complex conditions like role differentiation (`owner_is_sa`) to make the code easier to follow.
- Consider modularizing the block that grants roles into its own function.
- Provide more detailed error messages in case of statement mismatch.
- Use named constants for the statement numbers to avoid potential confusion.

---

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -875,45 +868,76 @@ is_user(Oid role_oid)
 		 * Only sysadmin can not be dropped. For the rest of the cases i.e., type
 		 * is "S" or "U" etc, we should drop the user
 		 */
```
HarshLunagariya: Lets remove this comment as well.

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -849,24 +849,17 @@ get_authid_login_ext_idx_oid(void)
  *****************************************/
 
 bool
-is_user(Oid role_oid)
+is_user(Oid role_oid, bool current_db_only)
```
HarshLunagariya: Let's add comment for this function usage.

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -875,45 +868,76 @@ is_user(Oid role_oid)
...
 
-	ReleaseSysCache(authtuple);
+	pfree(rolname);
 
 	return is_user;
 }
 
 bool
-is_role(Oid role_oid)
+is_role(Oid role_oid, bool current_db_only)
```
HarshLunagariya: nit: Does it make sense to refactor is_user and is_role to avoid duplication?

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -875,45 +868,76 @@ is_user(Oid role_oid)
...
+			if (current_db_only)
+			{
+				Datum db_name = SysCacheGetAttr(AUTHIDUSEREXTROLENAME, tuple,
+												 Anum_bbf_authid_user_ext_database_name, &isnull);
+				char *db_name_cstring = TextDatumGetCString(db_name);
+				char *current_db_name = get_cur_db_name();
+
+				Assert(!isnull);
+
+				is_user = (strcmp(db_name_cstring, current_db_name) == 0);
```
HarshLunagariya: Please make sure any of these db_names doesn't mismatch due to truncation of name(in of the values) and let's add testcase for it.

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -964,20 +988,32 @@ get_authid_user_ext_physical_name(const char *db_name, const char *login)
...
 
-		userform = (Form_authid_user_ext) GETSTRUCT(tuple_user_ext);
-		user_name = pstrdup(NameStr(userform->rolname));
+		datum = heap_getattr(tuple_user_ext, Anum_bbf_authid_user_ext_user_can_connect,
+							 RelationGetDescr(bbf_authid_user_ext_rel), &isnull);
+		Assert(!isnull);
+		user_can_connect = DatumGetInt32(datum);
+
+		if (user_can_connect == 1 ||
+			(has_privs_of_role(get_role_oid(login, false), get_role_oid(db_accessadmin, false))))
```
HarshLunagariya: Can you please explain this change? Let's add comment for clarification.
Does it imply that even if user is disabled then also it can be utilized if it is member of db_accessadmin?

contrib/babelfishpg_tsql/src/dbcmds.c
```diff
@@ -254,13 +271,17 @@ gen_dropdb_subcmds(const char *dbname, List *db_users)
...
+			strcmp(user_name, db_accessadmin) != 0)
 		{
 			appendStringInfo(&query, "DROP OWNED BY dummy CASCADE; ");
 			appendStringInfo(&query, "DROP ROLE dummy; ");
 			expected_stmts += 2;
 		}
 	}
+	/* Drop db_accessadmin*/
+	appendStringInfo(&query, "REVOKE CREATE ON DATABASE dummy FROM dummy; ");
+	appendStringInfo(&query, "DROP ROLE dummy; ");
```
HarshLunagariya, line 286: Let's execute DROP OWNED BY db_accessadmin statement as well -- For seamless cleanup even if object ownership got assigned by mistake in interoperability or any miscellaneous cases.

contrib/babelfishpg_tsql/sql/babelfishpg_tsql.sql
```diff
@@ -2348,13 +2348,16 @@ CREATE OR REPLACE PROCEDURE sys.sp_helpdbfixedrole("@rolename" sys.SYSNAME = NUL
 $$
 BEGIN
 	-- Returns a list of the fixed database roles. 
-	-- Only fixed role present in babelfish is db_owner.
 	IF LOWER(RTRIM(@rolename)) IS NULL OR LOWER(RTRIM(@rolename)) = 'db_owner'
 	BEGIN
 		SELECT CAST('db_owner' AS sys.SYSNAME) AS DbFixedRole, CAST('DB Owners' AS sys.nvarchar(70)) AS Description;
 	END
+	ELSE IF LOWER(RTRIM(@rolename)) IS NULL OR LOWER(RTRIM(@rolename)) = 'db_accessadmin'
```
rishabhtanwar29: No need to check nullability again in ELSE.

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -849,24 +849,17 @@ get_authid_login_ext_idx_oid(void)
...
 
-	authtuple = SearchSysCache1(AUTHOID, ObjectIdGetDatum(role_oid));
-	if (!HeapTupleIsValid(authtuple))
-		ereport(ERROR,
-				(errcode(ERRCODE_UNDEFINED_OBJECT),
-				 errmsg("role with OID %u does not exist", role_oid)));
-	rolname = ((Form_pg_authid) GETSTRUCT(authtuple))->rolname;
-	tuple = SearchSysCache1(AUTHIDUSEREXTROLENAME, NameGetDatum(&rolname));
+	rolname = GetUserNameFromId(role_oid, false);
+	tuple = SearchSysCache1(AUTHIDUSEREXTROLENAME, CStringGetDatum(rolname));
```
rishabhtanwar29: A lot of this logic is already there, I guess your target branch is not up to date.

contrib/babelfishpg_tsql/src/dbcmds.c
```diff
@@ -1225,3 +1258,125 @@ create_guest_schema_for_all_dbs(PG_FUNCTION_ARGS)
...
+				 errmsg("%s can only be called from an SQL script executed by CREATE/ALTER EXTENSION",
+						"create_database_roles_for_all_dbs()")));
+
+	saved_nest_level = pltsql_new_guc_nest_level();
+	set_config_option("babelfishpg_tsql.sql_dialect", "tsql",
+						GUC_CONTEXT_CONFIG,
+						PGC_S_SESSION, GUC_ACTION_SAVE, true, 0, false);
+	set_config_option("babelfishpg_tsql.migration_mode", physical_schema_name_exists("dbo") ? "single-db" : "multi-db",
+						GUC_CONTEXT_CONFIG,
+						PGC_S_SESSION, GUC_ACTION_SAVE, true, 0, false);
```
HarshLunagariya: Let's consult @rishabhtanwar29 specifically for this change.

contrib/babelfishpg_tsql/src/dbcmds.c
```diff
@@ -1225,3 +1258,125 @@ create_guest_schema_for_all_dbs(PG_FUNCTION_ARGS)
...
+		stmt = parsetree_nth_stmt(parsetree_list, i++);
+		update_GrantStmt(stmt, get_database_name(MyDatabaseId), NULL, db_accessadmin, NULL);
+
+		GetUserIdAndSecContext(&save_userid, &save_sec_context);
+
+		PG_TRY();
+		{
+			ListCell *parsetree_item;
+
+			SetConfigOption("createrole_self_grant", "inherit", PGC_USERSET, PGC_S_OVERRIDE);
```
HarshLunagariya: do we require inherit here?

contrib/babelfishpg_tsql/src/dbcmds.c
```diff
@@ -1225,3 +1258,125 @@ create_guest_schema_for_all_dbs(PG_FUNCTION_ARGS)
...
+				}
+
+				wrapper = makeNode(PlannedStmt);
+				wrapper->commandType = CMD_UTILITY;
+				wrapper->canSetTag = false;
+				wrapper->utilityStmt = ((RawStmt *) lfirst(parsetree_item))->stmt;
+				wrapper->stmt_location = 0;
+
+				ProcessUtility(wrapper,
+							"(CREATE LOGICAL DATABASE )",
```
HarshLunagariya: Let's change "CREATE LOGICAL DATABASE" to something meaning ful in this context.

contrib/babelfishpg_tsql/src/multidb.c
```diff
@@ -249,6 +249,17 @@ rewrite_object_refs(Node *stmt)
...
+				else if (strcmp(role_name, DB_ACCESSADMIN) == 0)
+				{
+					if (has_privs_of_role(GetUserId(), get_role_oid(get_db_owner_name(db_name), false)))
+					{
+						/* only members of db_owner can alter drop members to fixed db roles */
+					}
+					else
+						ereport(ERROR,
+								(errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
+								 errmsg("Cannot alter the role '%s', because it does not exist or you do not have permission.", role_name)));
```
HarshLunagariya: This doesn't look like a suitable place to do such checks -- currently we are doing it in rewriter. I would vote to do it in executor(just following the existing model).

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -3202,16 +3215,25 @@ bbf_ProcessUtility(PlannedStmt *pstmt,
...
-						char	   *cur_user;
-						Oid     	prev_current_user;
+						char		*db_name = get_cur_db_name();
+						bool 		is_member_of_db_owner = false;
+						bool 		is_member_of_db_accessadmin = false;
+						int 		save_sec_context;
+						Oid 		save_userid;
+						Oid 		db_owner = get_role_oid(get_db_owner_name(db_name), false);
+						Oid 		db_accessadmin = get_role_oid(get_db_accessadmin_role_name(db_name), false);
+						Oid 		user_name = get_role_oid(stmt->role->rolename, false);
```
HarshLunagariya: replace it with user_id

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -3222,19 +3244,37 @@ bbf_ProcessUtility(PlannedStmt *pstmt,
...
 											(errcode(ERRCODE_INSUFFICIENT_PRIVILEGE),
 											 errmsg("Current user does not have privileges to change schema")));
+								}
 							}
 							else if (strcmp(defel->defname, "rename") == 0)
 							{
-								if (strcmp(cur_user, dbo_name) != 0 &&
-									strcmp(cur_user, user_name) != 0)
+								if (is_member_of_db_owner || (isuser && is_member_of_db_accessadmin &&
+									!has_privs_of_role(user_name, db_owner)))
```
HarshLunagariya: Please explain the second condition specially `&& !has_privs_of_role(user_name, db_owner))`

HarshLunagariya: Isn't `|| user_name == GetUserId()` condition missing here?

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -3244,23 +3284,30 @@ bbf_ProcessUtility(PlannedStmt *pstmt,
 
 							if (strcmp(defel->defname, "rolemembers") == 0)
 							{
-								if (strcmp(cur_user, dbo_name) != 0 &&
-									strcmp(cur_user, user_name) != 0)
+								if (is_member_of_db_owner)
```
HarshLunagariya, line 3359: Isn't || user_name == GetUserId() condition missing here?

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -3276,12 +3323,11 @@ bbf_ProcessUtility(PlannedStmt *pstmt,
 						}
 						PG_FINALLY();
 						{
-							SetCurrentRoleId(prev_current_user, true);
+							SetUserIdAndSecContext(save_userid, save_sec_context);
```
HarshLunagariya, line 3398: A bit off-topic from this change - Acc. to our offline discussion, This won't be required if we decrement the GUC level(of PG side), right? I think the same happens in `AtEOXact_GUC`. In PG code as well, they resetting the UserId explicitly. I assume UserId is not handled by GUC stack[Might not be correct].

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -3560,6 +3615,22 @@ bbf_ProcessUtility(PlannedStmt *pstmt,
 						else
 							orig_schema = "dbo";
 					}
+					else
+					{
+						owner_oid = rolspec ? get_rolespec_oid(rolspec, true) : InvalidOid;
+						if (OidIsValid(owner_oid) && OidIsValid(db_accessadmin) && !member_can_set_role(GetUserId(), owner_oid) &&
+							has_privs_of_role(GetUserId(), db_accessadmin) &&
+							(is_user(owner_oid, true) || is_role(owner_oid, true)))
```
HarshLunagariya: What happens if owner_oid is invalid? It looks like it will allow the creation with current user as owner. Please correct me if it's not the case.

contrib/babelfishpg_tsql/src/catalog.c
tanscorpio7: Yes

contrib/babelfishpg_tsql/src/catalog.c
tanscorpio7: Yes, db_accessadmin has connect privilege, so its member always have access, unless DENY is used which babelfish does not support.

contrib/babelfishpg_tsql/src/dbcmds.c
tanscorpio7: Nope, but existing roles were created this way, so keeping it same for consistency.
I think we should have one single PR were we fix this for new and existing databases

contrib/babelfishpg_tsql/src/pl_handler.c
tanscorpio7: > Please explain the second condition specially && !has_privs_of_role(user_name, db_owner))

db_accessadmin cannot rename a user who is a member of db_owner.

> Isn't || user_name == GetUserId() condition missing here?

A user can only change its default_schema. renaming itself or changing its own login mapping permission is not implicitly available.

contrib/babelfishpg_tsql/src/pl_handler.c
tanscorpio7: same as [this reply](https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/2970#discussion_r1776727575)

contrib/babelfishpg_tsql/src/pl_handler.c
tanscorpio7: Do you mean the first check `OidIsValid(owner_oid)` is not enough ?

contrib/babelfishpg_tsql/src/multidb.c
tanscorpio7: Agreed

contrib/babelfishpg_tsql/src/dbcmds.c
```diff
@@ -1225,3 +1258,125 @@ create_guest_schema_for_all_dbs(PG_FUNCTION_ARGS)
...
+		dbname = text_to_cstring(&(bbf_db->name));
+		db_owner = get_db_owner_name(dbname);
+		db_accessadmin = get_db_accessadmin_role_name(dbname);
+
+		initStringInfo(&query);
+
+		appendStringInfo(&query, "CREATE ROLE dummy ROLE dummy; ");
+		appendStringInfo(&query, "GRANT CREATE ON DATABASE dummy TO dummy; ");
+
+		parsetree_list = raw_parser(query.data, RAW_PARSE_DEFAULT);
```
rishabhtanwar29, line 1381: This can be done only once outside of the while loop. Here we just need to update the rolename/dbname.

contrib/babelfishpg_tsql/src/dbcmds.c
rishabhtanwar29: +1, it is important to provide right query string so that logs contain meaning information.

contrib/babelfishpg_tsql/src/dbcmds.c
```diff
@@ -1225,3 +1258,125 @@ create_guest_schema_for_all_dbs(PG_FUNCTION_ARGS)
...
 }
+
+
+/*
+* This function is only being used during upgrade to v4.4.0
+* to create database roles db_accessadmin for each database
+*/
+PG_FUNCTION_INFO_V1(create_database_roles_for_all_dbs);
+Datum
+create_database_roles_for_all_dbs(PG_FUNCTION_ARGS)
```
rishabhtanwar29, line 1348: Why not make the name specific to what role it is creating or do we intend to add more roles with this function in this release?

contrib/babelfishpg_tsql/src/multidb.c
```diff
@@ -249,6 +249,17 @@ rewrite_object_refs(Node *stmt)
 								 errmsg("Dropping members to db_owner is not currently supported "
 										"in Babelfish")));
 				}
+				else if (strcmp(role_name, DB_ACCESSADMIN) == 0)
+				{
+					if (has_privs_of_role(GetUserId(), get_role_oid(get_db_owner_name(db_name), false)))
+					{
+						/* only members of db_owner can alter drop members to fixed db roles */
+					}
+					else
```
rishabhtanwar29: Why not write a standalone inverse `if` instead of this empty `if` and then else? Like:
```
if (!has_privs_of_role(GetUserId(), get_role_oid(get_db_owner_name(db_name), false)))
```
Anyways, why don't we do it it in processUtility instead of object_rewrite?

contrib/babelfishpg_tsql/src/multidb.c
rishabhtanwar29: +1

contrib/babelfishpg_tsql/src/multidb.c
```diff
@@ -1414,6 +1428,21 @@ get_guest_role_name(const char *dbname)
 	}
 }
 
+char *
+get_db_accessadmin_role_name(const char *dbname)
+{
+	char	   *name = palloc0(MAX_BBF_NAMEDATALEND);
+
+	if (get_migration_mode() == SINGLE_DB && strcmp(dbname, "master") != 0
+	    && strcmp(dbname, "tempdb") != 0 && strcmp(dbname, "msdb") != 0)
```
rishabhtanwar29: Looks like we need to a utility function called `is_builtin_db(const char *dbname)` to check if it is one of master/tempdb/msdb.

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -3560,6 +3615,22 @@ bbf_ProcessUtility(PlannedStmt *pstmt,
 						else
 							orig_schema = "dbo";
 					}
+					else
+					{
+						owner_oid = rolspec ? get_rolespec_oid(rolspec, true) : InvalidOid;
```
rishabhtanwar29: I think owner_oid should be `GetUserId()` when rolspec is invalid so that we create schema with current_user as owner.

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -3560,6 +3615,22 @@ bbf_ProcessUtility(PlannedStmt *pstmt,
...
+						if (OidIsValid(owner_oid) && OidIsValid(db_accessadmin) && !member_can_set_role(GetUserId(), owner_oid) &&
```
rishabhtanwar29: Both of these owner and db_Accessadmin OIDs must be valid or otherwise we should throw error.

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -3551,6 +3603,9 @@ bbf_ProcessUtility(PlannedStmt *pstmt,
 					GrantStmt  *stmt;
 					PlannedStmt *wrapper;
 					RoleSpec *rolspec = create_schema->authrole;
+					Oid       owner_oid = InvalidOid;
+					Oid       db_accessadmin = get_role_oid(get_db_accessadmin_role_name(get_cur_db_name()), true);
```
rishabhtanwar29: Missing_ok should be false.

contrib/babelfishpg_tsql/src/pl_handler.c
rishabhtanwar29: Let's add a comment regarding what this condition is checking.

test/JDBC/input/db_accessadmin-vu-cleanup.mix
rishabhtanwar29, line 1: Let's add this test to upgrade schedule files.

contrib/babelfishpg_tsql/src/pl_handler.c
tanscorpio7: Problem here is during upgrade from 3.0 -> 3.1 we create guest schemas for all existing babelfish databases. 
During this CREATE SCHEMA operation db_accessadmin will not exists.
`create_guest_schema_for_all_dbs` <-- Problematic function name

Should I instead reuse this query string `(CREATE LOGICAL DATABASE )` in `create_guest_schema_for_all_dbs` ?
Then we can change missing okay to false.

contrib/babelfishpg_tsql/src/pl_handler.c
tanscorpio7: Owner oid will only be valid if `rolspec` exists. `rolspec` == NULL means current user should be owner.
for `db_Accessadmin` we can remove it based on how we moved forward with [this comment](https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/2970#discussion_r1776827788)

tanscorpio7: Actually I can refactor this a bit.
`else ---> else if (rolspec)` Then owner_oid will always be valid inside the block

contrib/babelfishpg_tsql/src/pl_handler.c
tanscorpio7: We are trying to check if current user is trying to create schema for some one else.
If the current user itself is the owner, rolspec will be null and `CreateSchemaCommand` will handle it.

contrib/babelfishpg_tsql/src/multidb.c
tanscorpio7: Will change this to a macro

contrib/babelfishpg_tsql/src/multidb.c
tanscorpio7: > Why not write a standalone inverse if instead of this empty if and then else? Like:

Readability, these are important permission checks and reading inverse condition felt a bit confusing.
Although I think this one might be better of written inverse.

> Anyways, why don't we do it it in processUtility instead of object_rewrite?

Doing it in object_rewrite meant we just need to compare the original name. Doing it in processUtility means we will need to fetch the original name from the physical name using babelfish catalog.
But since we have syscache now, I'll do that.

contrib/babelfishpg_tsql/src/pl_handler.c
tanscorpio7: Right, transaction abort/cleanup handles the clean up of this local CurrentUserId change. But I think this might be more safe, since we really do not want to leak this state.

contrib/babelfishpg_tsql/src/dbcmds.c
tanscorpio7: Yes, three more fixed db roles will use this.

contrib/babelfishpg_tsql/src/catalog.c
tanscorpio7: I changed 
```
authtuple = SearchSysCache1(AUTHOID, ObjectIdGetDatum(role_oid));
if (!HeapTupleIsValid(authtuple))
ereport(ERROR,
		(errcode(ERRCODE_UNDEFINED_OBJECT),
			errmsg("role with OID %u does not exist", role_oid)));
rolname = ((Form_pg_authid) GETSTRUCT(authtuple))->rolname;
```
to
```
rolname = GetUserNameFromId(role_oid, false);
```
I don't think any other commit did this change.

contrib/babelfishpg_tsql/src/catalog.c
tanscorpio7: Just checked. db_names are truncated when stored in `sys.sys.babelfish_authid_user_ext`. So is `current_db_name`

contrib/babelfishpg_tsql/src/dbcmds.c
tanscorpio7: @rishabhtanwar29

contrib/babelfishpg_tsql/src/catalog.c
tanscorpio7: Further changed this to make is_role & is_user into one function.

contrib/babelfishpg_tsql/src/dbcmds.c
tanscorpio7: Done

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--4.3.0--4.4.0.sql
```diff
@@ -1906,6 +1915,211 @@ CALL sys.babelfish_drop_deprecated_object('function', 'sys', 'sp_statistics_inte
...
+		WHERE 1=0;	
+	END
+	ELSE
+		RAISERROR('''%s'' is not a known fixed role.', 16, 1, @rolename);
+END
+$$
+LANGUAGE 'pltsql';
+GRANT EXECUTE ON PROCEDURE sys.sp_helpdbfixedrole TO PUBLIC;
+
+CREATE OR REPLACE PROCEDURE sys.sp_helpuser("@name_in_db" sys.SYSNAME = NULL) AS
```
HarshLunagariya, line 4048: There is already sys.sp_helpuser() in this upgrade script. I just figured it out while rebasing my PR over your branch.

tanscorpio7: Fixed

contrib/babelfishpg_tsql/src/pl_handler.c
tanscorpio7: Made some change and setting missing ok to false now

contrib/babelfishpg_tsql/src/pl_handler.c
tanscorpio7: Done

contrib/babelfishpg_tsql/src/multidb.c
tanscorpio7: Done

contrib/babelfishpg_tsql/src/dbcmds.c
tanscorpio7: Done

contrib/babelfishpg_tsql/sql/babelfishpg_tsql.sql
tanscorpio7: Changed the definition as a whole. For null input we should be returning all fixed db roles. Earlier implementation did not take this into account

contrib/babelfishpg_tsql/src/pl_handler.c
tanscorpio7: Done

contrib/babelfishpg_tsql/src/pl_handler.c
tanscorpio7: Done

contrib/babelfishpg_tsql/src/multidb.c
tanscorpio7: Done, moved logic to processUtility and added a helper function `get_authid_user_ext_original_name`
grantee --> get original name from this (only in current db) --> if db_accessadmin then only allow db_owner to add/drop members.

contrib/babelfishpg_tsql/src/dbcmds.c
tanscorpio7: Done

contrib/babelfishpg_tsql/src/catalog.c
tanscorpio7: Even sysdatabases stores mangled truncated name

contrib/babelfishpg_tsql/src/catalog.c
tanscorpio7: Done

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -848,74 +848,54 @@ get_authid_login_ext_idx_oid(void)
...
-bool
-is_user(Oid role_oid)
+/*
+ * Check if role is a bbf db principal. Returns BBF_ROLE if it is
+ * a db role, returns BBF_USER if it is db user else returns 0
+ * Looks only in current bbf db when current_db_only is set to true
+ */
+
+const int
+is_database_principal(Oid role_oid, bool current_db_only)
```
rishabhtanwar29: Suggestion: The function no longer returns a boolean now so we can name it like `get_bbf_rolekind(Oid role_oid, const char *db_name)`. Note that passing `db_name` makes it more generic and usable.

tanscorpio7: Right, changed name to `get_db_principal_kind`. Did not want to use "role" in function name since it can be used for both pltsql database users and roles.
Also changes the bool flag to const char *db_name.

contrib/babelfishpg_tsql/src/dbcmds.c
```diff
@@ -254,15 +271,19 @@ gen_dropdb_subcmds(const char *dbname, List *db_users)
 	{
 		char	   *user_name = (char *) lfirst(elem);
 
-		if (strcmp(user_name, db_owner) != 0 && strcmp(user_name, dbo) != 0)
+		if (strcmp(user_name, db_owner) != 0 && strcmp(user_name, dbo) != 0 &&
+		    strcmp(user_name, db_accessadmin) != 0)
```
HarshLunagariya, line 275: Can we use `IS_FIXED_DB_PRINCIPAL(role)` macro here?

contrib/babelfishpg_tsql/src/session.c
```diff
@@ -48,12 +48,23 @@ get_cur_db_id(void)
...
 {
 	return pstrdup(current_db_name);
 }
 
+/*
+ * returns current active babelfish db name
+ * Callers should NOT pfree the return value
+ */
+const char *
+get_current_db_name(void)
```
HarshLunagariya: Better naming required IMO

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -3224,19 +3245,37 @@ bbf_ProcessUtility(PlannedStmt *pstmt,
...
 											(errcode(ERRCODE_INSUFFICIENT_PRIVILEGE),
 											 errmsg("Current user does not have privileges to change schema")));
+								}
 							}
 							else if (strcmp(defel->defname, "rename") == 0)
 							{
-								if (strcmp(cur_user, dbo_name) != 0 &&
-									strcmp(cur_user, user_name) != 0)
+								if (is_member_of_db_owner || (isuser && is_member_of_db_accessadmin &&
+									!has_privs_of_role(user_oid, db_owner)))
```
HarshLunagariya, line 3337: As discussed offline, This should be fixed in older versions as well, right? If yes then we should open a separate PR for it.

contrib/babelfishpg_tsql/src/pl_handler.c
HarshLunagariya: As discussed offline, This should be fixed in older versions as well, right? If yes then we should open a separate PR for it.

test/JDBC/expected/single_db/db_accessadmin-vu-verify.out
```diff
@@ -0,0 +1,337 @@
...
+
+~~ERROR (Message: User 'guest' cannot be dropped, it can only be disabled. The user is already disabled in the current database.)~~
+
+
+-- db_accessadmin can not add/drop members to any other role
+ALTER ROLE db_accessadmin ADD MEMBER babel_5136_u1
+GO
+~~ERROR (Code: 33557097)~~
+
+~~ERROR (Message: Cannot alter the role 'db_accessadmin', because it does not exist or you do not have permission.)~~
```
HarshLunagariya, line 332: Please add testcases for
1. ALTER ROLE db_accessadmin ADD MEMBER [any other fixed role -- for now sysadmin & db_owner]
2. ALTER ROLE db_accessadmin DROP MEMBER
3. sp_addrolemember
4. sp_droprolemember

test/JDBC/expected/single_db/db_accessadmin-vu-verify.out
```diff
@@ -0,0 +1,337 @@
...
+GO
+-- terminate-tsql-conn
+
+-- tsql user=babel_5136_l1 password=12345678 database=babel_5136
+-- Only db_owner members should be able to do this
+ALTER ROLE db_accessadmin ADD MEMBER babel_5136_db_accessadmin_user
+GO
+~~ERROR (Code: 33557097)~~
+
+~~ERROR (Message: Cannot alter the role 'db_accessadmin', because it does not exist or you do not have permission.)~~
```
HarshLunagariya, line 173: Let's add positive test case for this explicitly.

test/JDBC/expected/single_db/db_accessadmin-vu-verify.out
```diff
@@ -0,0 +1,337 @@
...
+GO
+~~ERROR (Code: 33557097)~~
+
+~~ERROR (Message: permission denied for database jdbc_testdb)~~
+
+CREATE DATABASE new_database
+GO
+~~ERROR (Code: 33557097)~~
+
+~~ERROR (Message: Only one user database allowed under single-db mode. User database "babel_5136" already exists)~~
```
HarshLunagariya: I think this is not intended. Please update.

test/JDBC/expected/single_db/db_accessadmin-vu-verify.out
```diff
@@ -0,0 +1,337 @@
...
+GO
+~~ERROR (Code: 2714)~~
+
+~~ERROR (Message: Function 'f1' already exists with the same name)~~
+
+CREATE DATABASE new_database
+GO
+~~ERROR (Code: 33557097)~~
+
+~~ERROR (Message: Only one user database allowed under single-db mode. User database "babel_5136" already exists)~~
```
HarshLunagariya: This is not intended output. Please update.

test/JDBC/expected/single_db/db_accessadmin-vu-verify.out
```diff
@@ -0,0 +1,337 @@
...
+SELECT current_user
+GO
+~~START~~
+varchar
+babel_5136_db_accessadmin_user
+~~END~~
+
+
+-- should be a member of db_accessadmin in babel_5136 database
+SELECT IS_ROLEMEMBER('db_accessadmin')
```
HarshLunagariya, line 192: Let's test If dbo & db_owner is member of 'db_accessadmin' or not by default.

test/JDBC/expected/single_db/db_accessadmin-vu-verify.out
```diff
@@ -0,0 +1,337 @@
...
+ALTER USER babel_5136_u2 WITH NAME = babel_5136_u1
+GO
+
+-- CREATE SCHEMA should be allowed db_accessadmin
+CREATE SCHEMA s1
+GO
+CREATE SCHEMA s2 AUTHORIZATION dbo
+GO
+CREATE SCHEMA s3 AUTHORIZATION babel_5136_u1
+GO
```
HarshLunagariya, line 254: Let's add testcase for GRANT ON SCHEMA.

test/JDBC/expected/single_db/db_accessadmin-vu-verify.out
HarshLunagariya: \+ Please add testcases for is_member as well.

test/JDBC/expected/single_db/db_accessadmin-vu-verify.out
```diff
@@ -0,0 +1,337 @@
...
+~~ERROR (Message: permission denied for database jdbc_testdb)~~
+
+CREATE DATABASE new_database
+GO
+~~ERROR (Code: 33557097)~~
+
+~~ERROR (Message: Only one user database allowed under single-db mode. User database "babel_5136" already exists)~~
+
+
+
```
HarshLunagariya, line 577: Please make sure that
1. It is being tested in both the migration mode. 
2. we are tracking the testcase about the restrictions on the role that it should not be altered/dropped from either of endpoint as separate effort.
3. we test SSMS utility as well for visibility and usage of this new role via UI.
4. We are verifying that it should not be the owner of anything else apart from schema and schema-contained objects.

contrib/babelfishpg_tsql/src/dbcmds.c
tanscorpio7: No, the fixed role names are physical rolename here.

contrib/babelfishpg_tsql/src/pl_handler.c
tanscorpio7: I will file a separate JIRA for this but keeping it same as before in my PR feels wrong 😅

test/JDBC/expected/single_db/db_accessadmin-vu-verify.out
tanscorpio7: Added

test/JDBC/expected/single_db/db_accessadmin-vu-verify.out
tanscorpio7: Already added, also added a comment above that line for easier test case review

test/JDBC/expected/single_db/db_accessadmin-vu-verify.out
tanscorpio7: Updated

test/JDBC/expected/single_db/db_accessadmin-vu-verify.out
tanscorpio7: Updated

test/JDBC/expected/single_db/db_accessadmin-vu-verify.out
tanscorpio7: Fixed `IS_ROLEMEMBER()` 
"dbo" should be a member of all fixed roles ONLY
No fixed db role can be a member of anyone else, so always zero.

contrib/babelfishpg_tsql/src/session.c
tanscorpio7: Any suggestions ? I could not think of any better name. and it is returing const char so should be safe from pfree.

test/JDBC/expected/single_db/db_accessadmin-vu-verify.out
tanscorpio7: Done

contrib/babelfishpg_tsql/src/dbcmds.c
shalinilohia50: Can we move the creation of db roles to another function like `create_db_roles_if_not_exists` mentioned here? https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/3018/files 

Since our PRs have common changes, that would be better to add other db roles in the same function.

shalinilohia50 (changes requested):

tanscorpio7:
Closing this PR since the target branch is moved to amazon aurora repo
New PR https://github.com/amazon-aurora/babelfish_extensions/pull/66/

PR #2974, opened 2024-09-25
grant on schema does not take effect on future objects created in a schema by schema's authorized users

shalinilohia50:
### Description

grant on schema does not take effect on future objects created in a schema by the schema's authorized users

When a user  `u1` is granted ownership on a schema `s1` at the time of creation. And there is `grant select on schema permission on s1 to u2`. The future objects created by the users `dbo` and `u1` should be accessible by the user `u2`

Earlier, if `u1` created a table in `s1` schema since it's authorized to do so, the user `u2` was not allowed to access the object created by `u1`.

With this commit, future objects created by `u1` and `dbo` will be accessible to the user `u2`.

### Issues Resolved

Task: [BABEL-5172](https://jira.rds.a2z.com/browse/BABEL-5172)

* **Use case based -** Added

* **Boundary conditions -**

* **Arbitrary inputs -**

* **Negative test cases -**

* **Performance tests -**

* **Tooling impact -**

* **Client tests -**

### Check List


contrib/babelfishpg_tsql/src/pltsql_utils.c
```diff
@@ -2206,12 +2258,21 @@ static List
...
-exec_grantschema_subcmds(const char *schema, const char *rolname, bool is_grant, bool with_grant_option, AclMode privilege)
+exec_grantschema_subcmds(const char *schema, const char *rolname, bool is_grant, bool with_grant_option, AclMode privilege, bool is_create_schema)
 {
 	List		*parsetree_list;
 	ListCell	*parsetree_item;
+	const char *prev_current_user;
+	const char *dbo_role;
+	const char *dbname = get_cur_db_name();
+	MigrationMode baseline_mode = is_user_database_singledb(dbname) ? SINGLE_DB : MULTI_DB;
+	dbo_role = get_dbo_role_name_by_mode(dbname, baseline_mode);
```
rishabhtanwar29, line 2264: Why do we need to infer the migration mode? Is `exec_grantschema_subcmds` is intended to be executed during extension upgrade?

contrib/babelfishpg_tsql/src/pltsql_utils.c
```diff
@@ -2206,12 +2258,21 @@ static List
...
+
+	/* Need dbo user to execute the statements. */
+	prev_current_user = GetUserNameFromId(GetUserId(), false);
+	bbf_set_current_user(dbo_role);
```
rishabhtanwar29: Let's use better API `SetUserIdAndSecContext` to set the role. Also, there might be chances of user getting leaked if an error comes just before resetting the user. Better use TRY/CATCH.

contrib/babelfishpg_tsql/src/pltsql_utils.c
```diff
@@ -2306,3 +2368,30 @@ make_rolespec_node(const char *rolename)
 
 	return n;
 }
+
+/* 
+ * Returns the oid of schema owner.
+ */
+int
```
rishabhtanwar29: Oid?

HarshLunagariya: Let's use Oid instead of int.

contrib/babelfishpg_tsql/src/pltsql_utils.c
```diff
@@ -2306,3 +2368,30 @@ make_rolespec_node(const char *rolename)
...
+get_owner_of_schema(const char *schema)
+{
+	HeapTuple	tup;
+	Form_pg_namespace nspform;
+	int result;
+	NameData   *schema_name;
+	schema_name = (NameData *) palloc0(NAMEDATALEN);
+	snprintf(schema_name->data, NAMEDATALEN, "%s", schema);
+
+	tup = SearchSysCache1(NAMESPACENAME, NameGetDatum(schema_name));
```
rishabhtanwar29: I think you can directly use CstringGetDatum:
```
tup = SearchSysCache1(NAMESPACENAME, CStringGetDatum(schema));
```

contrib/babelfishpg_tsql/src/pltsql_utils.c
```diff
@@ -2306,3 +2368,30 @@ make_rolespec_node(const char *rolename)
...
+
+	if (!HeapTupleIsValid(tup))
+		ereport(ERROR,
+				(errcode(ERRCODE_UNDEFINED_SCHEMA),
+				 errmsg("schema \"%s\" does not exist", schema)));
+
+	nspform = (Form_pg_namespace) GETSTRUCT(tup);
+	result = ((int) nspform->nspowner);
```
rishabhtanwar29: *(oid)

test/JDBC/expected/BABEL-5172.out
rishabhtanwar29, line 1: Let's use some verbose names instead of jira numbers. Even better if we can add these new test cases to some existing test files, like GRANT_SCHEMA-vu-*.

contrib/babelfishpg_tsql/src/pltsql_utils.c
```diff
@@ -2135,13 +2142,30 @@ exec_alter_role_cmd(char *query_str, RoleSpec *role)
...
 	StringInfoData query;
 	List	   *stmt_list;
 	Node	   *stmt;
 	int			expected_stmts = 2;
 	int			i = 0;
+	bool		owner_other_than_dbo = false;
+	char 		*dbname = get_cur_db_name();
+	const char	*dbo_role = get_dbo_role_name(dbname);
+	const char	*db_owner_role = get_db_owner_name(dbname);
+	const char	*schema_owner = GetUserNameFromId(get_owner_of_schema(schema), false);
```
rishabhtanwar29, line 2156: Now that I think about this, none of these objects should be `const`. The functions like `get_dbo_role_name`, `get_db_owner_name` etc return both const and palloc'd strings so it can't be const depending upon whether the database is builtin or user defined. This is not correct since in case of user defined database, the function returns palloc'd string but we can't free it due to `const` qualifier. Rather such functions should return palloc'd string even for builtin databases and then caller should always free the returned value.
*Suggestion*: Can take this opportunity to fix few of such functions.

contrib/babelfishpg_tsql/src/pltsql_utils.c
```diff
@@ -2135,13 +2142,30 @@ exec_alter_role_cmd(char *query_str, RoleSpec *role)
...
+
+	/*
+	 * Don't need multiple ALTER DEFAULT PRIVILEGE statements, if:
+	 * 1. It's a part of CREATE SCHEMA statement
+	 * 2. If the schema owner is not dbo
+	 * 3. If the schema owner is not db_owner
```
rishabhtanwar29: Question: why is this needed only if schema owner is dbo or db_owner? Ideally if a user is granted SELECT on a particular schema then no matter who the owner is, that user should be able SELECT from any existing and new objects in that schema isn't it?

contrib/babelfishpg_tsql/src/pltsql_utils.c
```diff
@@ -2164,7 +2188,16 @@ static List
...
+			
+			if (owner_other_than_dbo)
+			{
+				/* Grant ALTER DEFAULT PRIVILEGES on schema owner and dbo user. */
+				appendStringInfo(&query, "ALTER DEFAULT PRIVILEGES FOR ROLE dummy IN SCHEMA dummy GRANT dummy ON TABLES TO dummy; ");
+				appendStringInfo(&query, "ALTER DEFAULT PRIVILEGES FOR ROLE dummy IN SCHEMA dummy GRANT dummy ON TABLES TO dummy; ");
+				expected_stmts = 3;
+			}
+			else
+				appendStringInfo(&query, "ALTER DEFAULT PRIVILEGES IN SCHEMA dummy GRANT dummy ON TABLES TO dummy; ");
```
HarshLunagariya: Is there a possibility of db_owner role as an owner of any schema -- dbo schema of master, tempdb and msdb databases have db_owner roles as an owner?
If that's the case then for db_owner also we will need to execute ALTER DEF. PRIVILEGES for db_owner.

contrib/babelfishpg_tsql/src/pltsql_utils.c
```diff
@@ -2177,7 +2210,15 @@ static List
 		else
 		{
 			appendStringInfo(&query, "REVOKE dummy ON ALL TABLES IN SCHEMA dummy FROM dummy; ");
-			appendStringInfo(&query, "ALTER DEFAULT PRIVILEGES IN SCHEMA dummy REVOKE dummy ON TABLES FROM dummy; ");
+			if (owner_other_than_dbo)
+			{
+				/* Grant ALTER DEFAULT PRIVILEGES on schema owner and dbo user. */
+				appendStringInfo(&query, "ALTER DEFAULT PRIVILEGES FOR ROLE dummy IN SCHEMA dummy REVOKE dummy ON TABLES FROM dummy; ");
+				appendStringInfo(&query, "ALTER DEFAULT PRIVILEGES FOR ROLE dummy IN SCHEMA dummy REVOKE dummy ON TABLES FROM dummy; ");
+				expected_stmts = 3;
```
HarshLunagariya: nit: better would be to do `expected_stmts++;`

contrib/babelfishpg_tsql/src/pltsql_utils.c
HarshLunagariya, line 2264: Was this failing the upgrade?

contrib/babelfishpg_tsql/src/pltsql_utils.c
```diff
@@ -2177,7 +2210,15 @@ static List
...
+			}
+			else
+				appendStringInfo(&query, "ALTER DEFAULT PRIVILEGES IN SCHEMA dummy REVOKE dummy ON TABLES FROM dummy; ");
```
HarshLunagariya, line 2217: This is equivalent to `ALTER DEFAULT PRIVILEGES FOR ROLE *dbo* IN SCHEMA dummy REVOKE dummy ON TABLES FROM dummy;`, right ? Can we refactor/simplify the logic.

contrib/babelfishpg_tsql/src/pltsql_utils.c
```diff
@@ -2135,13 +2142,30 @@ exec_alter_role_cmd(char *query_str, RoleSpec *role)
...
 	int			i = 0;
+	bool		owner_other_than_dbo = false;
+	char 		*dbname = get_cur_db_name();
+	const char	*dbo_role = get_dbo_role_name(dbname);
+	const char	*db_owner_role = get_db_owner_name(dbname);
+	const char	*schema_owner = GetUserNameFromId(get_owner_of_schema(schema), false);
+
+	/*
+	 * Don't need multiple ALTER DEFAULT PRIVILEGE statements, if:
+	 * 1. It's a part of CREATE SCHEMA statement
```
HarshLunagariya, line 2160: Can you please explain the 1st point? Which grant statement gets executed at the time of grant schema? Is it `grant USAGE on schema to public` ?

contrib/babelfishpg_tsql/src/pltsql_utils.c
```diff
@@ -2164,7 +2188,16 @@ static List
 				appendStringInfo(&query, "GRANT dummy ON ALL TABLES IN SCHEMA dummy TO dummy WITH GRANT OPTION; ");
 			else
 				appendStringInfo(&query, "GRANT dummy ON ALL TABLES IN SCHEMA dummy TO dummy; ");
-			appendStringInfo(&query, "ALTER DEFAULT PRIVILEGES IN SCHEMA dummy GRANT dummy ON TABLES TO dummy; ");
+			
+			if (owner_other_than_dbo)
+			{
+				/* Grant ALTER DEFAULT PRIVILEGES on schema owner and dbo user. */
+				appendStringInfo(&query, "ALTER DEFAULT PRIVILEGES FOR ROLE dummy IN SCHEMA dummy GRANT dummy ON TABLES TO dummy; ");
+				appendStringInfo(&query, "ALTER DEFAULT PRIVILEGES FOR ROLE dummy IN SCHEMA dummy GRANT dummy ON TABLES TO dummy; ");
```
HarshLunagariya: We can improve the performance by executing single alter def. privileges stmt for multiple roles in single go. Reference : https://www.postgresql.org/docs/current/sql-alterdefaultprivileges.html
```
[ FOR { ROLE | USER } target_role [, ...] ]
```

HarshLunagariya (changes requested):

contrib/babelfishpg_tsql/src/pltsql_utils.c
shalinilohia50: As a part of `CREATE SCHEMA .. AUTHORIZATION ..`, we call `exec_grantschema_subcmds`. We don't need to execute these `ALTER DEFAULT PRIVILEGES` on multiple roles as a part of `CREATE SCHEMA` statement.

contrib/babelfishpg_tsql/src/pltsql_utils.c
shalinilohia50: Yes, this function was called as a part of upgrade. Since it's in single-db mode, it was checking of `dbo` always and failing the upgrade.

shalinilohia50: Yes, this function gets executed at the time of upgrade.

contrib/babelfishpg_tsql/src/pltsql_utils.c
shalinilohia50: Let's take that separately since this JIRA is marked for Vole and we don't have enough time left.

contrib/babelfishpg_tsql/src/pltsql_utils.c
shalinilohia50: We don't need more than 1 role for ALTER DEFAULT DEFAULT PRIVILEGES in case of schema owner being `dbo` (schema created by dbo user) or db_owner(schemas like master_dbo, tempdb_dbo etc which are owned by db_owner).

contrib/babelfishpg_tsql/src/pltsql_utils.c
shalinilohia50: Do we need this now? Or any test case which you think should be added for this? 
There are existing tests for schema which are either owned by `dbo` or `db_owner`. 
```
ALTER DEFAULT PRIVILEGES IN SCHEMA dummy GRANT dummy ON TABLES TO dummy;
```
This statement seems to handle those scenarios because only the owners of schema could execute the GRANT statement and this statement will take the role as current user. That's why I didn't specify "dbo" here.

contrib/babelfishpg_tsql/src/pltsql_utils.c
shalinilohia50: https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/2974#discussion_r1777158869

contrib/babelfishpg_tsql/src/pltsql_utils.c
```diff
@@ -2206,36 +2252,59 @@ static List
...
+						"(GRANT SCHEMA )",
+						false,
+						PROCESS_UTILITY_SUBCOMMAND,
+						NULL,
+						NULL,
+						None_Receiver,
+						NULL);
+		}
+	}
+	PG_CATCH();
```
rishabhtanwar29, line 2298: We should either change this to PG_FINALLY or add the missing pg_rethrow in catch.

contrib/babelfishpg_tsql/src/pltsql_utils.c
```diff
@@ -2206,36 +2252,59 @@ static List
...
-		/* do this step */
-		ProcessUtility(wrapper,
-					"(GRANT SCHEMA )",
-					false,
-					PROCESS_UTILITY_SUBCOMMAND,
-					NULL,
-					NULL,
-					None_Receiver,
-					NULL);
+		if(is_set_userid)
```
rishabhtanwar29: I don't we need to use this boolean, can do this unconditionally.

test/JDBC/input/1_GRANT_SCHEMA-vu-prepare.mix
```diff
@@ -161,4 +161,4 @@ go
 
 -- check for inconsistent metadata before upgrade
```
HarshLunagariya, line 162: Please add upgrade tests as well.

shalinilohia50: Added. 
Future objects like Tables and views  created by the schema owner other than `dbo`  would throw permission error. But we have a simple workaround of granting the schema permission again, we can document this. Fixing this in upgrade involves iterating over all the schema permission entries in the catalog and then altering the default privilege of schema if the schema owner is not dbo/db_owner. Is it worth the effort?

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -3937,3 +3937,155 @@ update_db_owner(const char *new_owner_name, const char *db_name)
...
+				if ((current_permission & permissions[i]) &&  permissions[i] != ACL_EXECUTE)
+				{
+					PG_TRY();
+					{
+						const char	*query = NULL;
+						List		*res;
+						AlterDefaultPrivilegesStmt	*alterdefpriv;
+						PlannedStmt	*wrapper;
+						query = psprintf("ALTER DEFAULT PRIVILEGES FOR ROLE %s, %s IN SCHEMA %s GRANT %s ON TABLES TO %s", dbo_user, schema_owner, physical_schema, privilege_to_string(permissions[i]), grantee);
+						res = raw_parser(query, RAW_PARSE_DEFAULT);
```
HarshLunagariya, line 4023: Shouldn't we have to execute `GRANT ... ON ALL TABLES` as well just to handle the privileges of old tables created by u1 in older version? (Referring u1 from commit description example).

HarshLunagariya: We need to execute `GRANT ... ON ALL TABLES` well to handle the older objects created by u1 user(referring the example given in commit desc.) after granting the schema privs. in older version.
Pseudo Code of example:
```
-- Old version
create schema sch1 owner u1
grant privs on schema to u2
create table t1()
t1 will not be accessible to u2 even if it got schema perms.
-- Latest Version
t1 should be accessible to u2
```

HarshLunagariya: Please add testcase for the same.

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -3937,3 +3937,136 @@ update_db_owner(const char *new_owner_name, const char *db_name)
...
+				F_TEXTEQ,
+				CStringGetTextDatum(OBJ_SCHEMA));
+
+	scan = systable_beginscan(bbf_schema_rel, get_bbf_schema_perms_idx_oid(),
+							true, NULL, 2, scanKey);
+	tuple_bbf_schema = systable_getnext(scan);
+
+	/* Set current user to session user for create permissions */
+	prev_current_user = GetUserNameFromId(GetUserId(), false);
+	bbf_set_current_user("sysadmin");
```
HarshLunagariya: why is it required? We don't support the granting CREATE permission.

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -3937,3 +3937,136 @@ update_db_owner(const char *new_owner_name, const char *db_name)
...
+
+	old_dbid = get_cur_db_id();
+	old_dbname = get_cur_db_name();
+	set_cur_db(dbid, dbname);	/* temporarily set current dbid as the new id */
```
HarshLunagariya: - It will get executed in the PG dialect.
- So after executing the set_cur_db() for last db, we should setting the search_path and current user to the existing value what was before execution of the function.
- Just wondering, Do we actually need setting and resetting?

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -3937,3 +3937,136 @@ update_db_owner(const char *new_owner_name, const char *db_name)
...
+		int			i;
+
+		schema_name = TextDatumGetCString(heap_getattr(tuple_bbf_schema, Anum_bbf_schema_perms_schema_name, dsc, &isnull));
+		grantee = TextDatumGetCString(heap_getattr(tuple_bbf_schema, Anum_bbf_schema_perms_grantee, dsc, &isnull));
+		current_permission = DatumGetInt32(heap_getattr(tuple_bbf_schema, Anum_bbf_schema_perms_permission, dsc, &isnull));
+
+		physical_schema = get_physical_schema_name_by_mode((char *) dbname, schema_name, baseline_mode);
+		dbo_user = get_dbo_role_name_by_mode((char *)dbname, baseline_mode);
+		db_owner = get_db_owner_name_by_mode((char *) dbname, baseline_mode);
+		schema_owner = GetUserNameFromId(get_owner_of_schema(physical_schema), false);
```
HarshLunagariya: Let's add NULL check on args here if these functions are handling it themselves.

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -3937,3 +3937,136 @@ update_db_owner(const char *new_owner_name, const char *db_name)
...
+	scan = table_beginscan_catalog(db_rel, 0, NULL);
+	tuple = heap_getnext(scan, ForwardScanDirection);
+
+	while (HeapTupleIsValid(tuple))
+	{
+		Datum		db_name_datum = heap_getattr(tuple, Anum_sysdatabases_name,
+												 db_rel->rd_att, &is_null);
+		const char *db_name = TextDatumGetCString(db_name_datum);
+
+		alter_default_privilege_for_db(db_name);
```
HarshLunagariya, line 4045: Can db_name be NULL?

shalinilohia50: No, it can't be.

contrib/babelfishpg_tsql/src/catalog.c
shalinilohia50: The arguments can't be NULL here. dbname, schema_name can't be NULL.

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -3937,3 +3937,111 @@ update_db_owner(const char *new_owner_name, const char *db_name)
...
+		/* If schema owner is other that dbo or db_owner user, only then execute ALTER DEFAULT PRIVILEGES. */
+		if ((strcmp(schema_owner, dbo_user) != 0) && (strcmp(schema_owner, db_owner) != 0))
+		{
+			/* For each permission, grant alter default privileges explicitly. */
+			for (i = 0; i < NUMBER_OF_PERMISSIONS; i++)
+			{
+				if ((current_permission & permissions[i]) &&  permissions[i] != ACL_EXECUTE)
+				{
+					const char	*alter_query = NULL;
+					const char	*grant_query = NULL;
```
rishabhtanwar29: Why CONST? Infact we should free them at the end otherwise a lot of memory will be accumulated for large number of databases and schemas.

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -3937,3 +3937,111 @@ update_db_owner(const char *new_owner_name, const char *db_name)
...
+		int			i;
+
+		schema_name = TextDatumGetCString(heap_getattr(tuple_bbf_schema, Anum_bbf_schema_perms_schema_name, dsc, &isnull));
+		grantee = TextDatumGetCString(heap_getattr(tuple_bbf_schema, Anum_bbf_schema_perms_grantee, dsc, &isnull));
+		current_permission = DatumGetInt32(heap_getattr(tuple_bbf_schema, Anum_bbf_schema_perms_permission, dsc, &isnull));
+
+		physical_schema = get_physical_schema_name_by_mode((char *) dbname, schema_name, baseline_mode);
+		dbo_user = get_dbo_role_name_by_mode((char *)dbname, baseline_mode);
+		db_owner = get_db_owner_name_by_mode((char *) dbname, baseline_mode);
+		schema_owner = GetUserNameFromId(get_owner_of_schema(physical_schema), false);
```
rishabhtanwar29: Should free whatever we can.

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -3937,3 +3937,111 @@ update_db_owner(const char *new_owner_name, const char *db_name)
...
+
+	db_rel = table_open(sysdatabases_oid, AccessShareLock);
+	scan = table_beginscan_catalog(db_rel, 0, NULL);
+	tuple = heap_getnext(scan, ForwardScanDirection);
+
+	while (HeapTupleIsValid(tuple))
+	{
+		Datum		db_name_datum = heap_getattr(tuple, Anum_sysdatabases_name,
+												 db_rel->rd_att, &is_null);
+		const char *db_name = TextDatumGetCString(db_name_datum);
```
rishabhtanwar29: Same, don't need to be CONST and should free.

test/JDBC/expected/GRANT_SCHEMA-before-15_9-16_5-vu-prepare.out
```diff
@@ -0,0 +1,126 @@
...
+1
+~~END~~
+
+
+select babel_5172_s1.f5();
+go
+~~START~~
+int
+1
+~~END~~
```
rishabhtanwar29, line 125: So we do expect execute function/procedure permission?

shalinilohia50: Yes, the problem was with tables and views.

rishabhtanwar29 (approved):

HarshLunagariya (approved):

PR #2980, opened 2024-09-26
Update babel3 bbf_pivot function upgrade script

RIC06X:
We wrapped rename and drop bbf_pivot function with a exception capture
code to raise warning instead of error.

### DIFF
--- a/contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--3.7.0--3.8.0.sql
+++ b/contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--3.7.0--3.8.0.sql
@@ -37,14 +37,38 @@ LANGUAGE plpgsql;
  * So make sure that any SQL statement (DDL/DML) being added here can be executed multiple times without affecting
  * final behaviour.
  */
+DO $$
+DECLARE
+    exception_message text;
+BEGIN
+    -- Rename bbf_pivot for dependencies
+    ALTER FUNCTION sys.bbf_pivot() RENAME TO bbf_pivot_deprecated_in_3_8_0;
 
-ALTER FUNCTION sys.bbf_pivot() RENAME TO bbf_pivot_deprecated_in_4_4_0;
-CALL sys.babelfish_drop_deprecated_object('function', 'sys', 'bbf_pivot_deprecated_in_4_4_0');
+EXCEPTION WHEN OTHERS THEN
+    GET STACKED DIAGNOSTICS
+    exception_message = MESSAGE_TEXT;
+    RAISE WARNING '%', exception_message;
+END;
+$$;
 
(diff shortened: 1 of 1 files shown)


Imran-imtiaz48 (commented):
The code effectively handles the renaming of the `bbf_pivot` function to mark it as deprecated, replacing it with an updated version while incorporating exception handling. This approach ensures backward compatibility and controlled deprecation, which is crucial for maintaining stability in systems with dependencies. Below are my observations and suggestions for improvement.

#### Positive Aspects:
1. **Exception Handling:** 
   - Including `EXCEPTION WHEN OTHERS` with `GET STACKED DIAGNOSTICS` ensures that any issues during the renaming or dropping of objects are caught and logged. This is a good practice for production environments.
   
2. **Controlled Deprecation:**
   - Deprecating the `bbf_pivot` function with a clear naming convention (`bbf_pivot_deprecated_in_3_8_0`) is a solid approach for managing versioning and future code maintenance.

3. **New Function Creation:**
   - Replacing the old function with `CREATE OR REPLACE FUNCTION sys.bbf_pivot` allows for seamless transition and minimizes the impact on existing systems using this function.

4. **Use of `babelfishpg_tsql`:**
   - Leveraging Babelfish for PostgreSQL and specifying the language as `'C'` while ensuring the function is `STABLE` shows a strong understanding of optimizing performance for system-level functions.

#### Suggestions for Improvement:
1. **Explicit Comments:**
   - Add more detailed comments explaining the purpose of each step. For instance, explain why the function is being renamed or dropped, and what implications this has for users.
   - For example:
     ```sql
     -- Rename the existing bbf_pivot function to mark it deprecated in version 3.8.0
     ```

2. **Improving Exception Handling:**
   - Although the exception handling captures the error and raises a warning, consider logging the errors into a dedicated error logging table for better auditing.
   - Example:
     ```sql
     EXCEPTION WHEN OTHERS THEN
         GET STACKED DIAGNOSTICS exception_message = MESSAGE_TEXT;
         INSERT INTO error_log_table (error_message, log_time) VALUES (exception_message, now());
         RAISE WARNING '%', exception_message;
     ```

3. **Edge Case Consideration:**
   - Ensure that dependent objects for `bbf_pivot` are carefully analyzed before dropping or renaming, as some systems may not handle this gracefully without careful checks.
   - Consider adding a check before renaming the function to see if any object depends on it:
     ```sql
     IF EXISTS (SELECT 1 FROM pg_depend WHERE objid = 'sys.bbf_pivot'::regprocedure) THEN
         RAISE WARNING 'bbf_pivot has dependent objects.';
     END IF;
     ```

4. **Role Assignment (Pending Completion):**
   - It seems that the code for assigning the `dbo` role to the `db_owner` login is incomplete. Make sure to properly close the `DO` block and complete the logic for role assignment.

#### Conclusion:
The script is well-structured and adheres to best practices for deprecating and replacing functions. Enhancing the code with more detailed comments, improving exception handling, and considering edge cases will further strengthen its robustness.

forestkeeper (approved):

PR #2960, opened 2024-09-23
Modify APIs to return palloc'ed values and catalog lookup to validate the user in get_physical_user_name

aswiniip:
### Description

Earlier in babelfish,
- The `get_*_role_name()` and `get_*_schema_name()` APIs returned constant values for a few cases. 
- The `get_physical_user_name()` used to return the user_name after appending the db_name to the user_name without any checks to verify its existence/validity. 

With this change,
- We return Palloc'ed values for the APIs instead of constant values.
- In `get_physical_user_name()` if the user or role is not found in the `sys.babelfish_authid_user_ext` catalog, then an error is thrown.

### Changes Made
- In `multidb.c`, made changes to the APIs like get_*_role_name() and get_*_schema_name() to return palloc values instead of constant values.
- Implemented a catalog lookup in `get_physical_user_name()` to verify if the user or role is found in the `sys.babelfish_authid_user_ext` catalog. The 'suppress_role_error' flag indicates if it is ok for the user or role to be absent from the catalog (required for cases when a new user/role is created and it can't be found in the catalog).
- Checked all the functions where the APIs are called and `pfree` the variables containing values from the above APIs after they are no longer used.

### Issues Resolved
Task: BABEL-4933

### Test Scenarios Covered ###
* **Use case based -**

* **Boundary conditions -**

* **Arbitrary inputs -**

* **Negative test cases -**

* **Minor version upgrade tests -**

* **Major version upgrade tests -**

* **Performance tests -**

* **Tooling impact -**

* **Client tests -**

### Check List
- [x] Commits are signed per the DCO using --signoff 

Signed-off-by: P Aswini Kumar [aswinii@amazon.com](mailto:aswinii@amazon.com)

By submitting this pull request, I confirm that my contribution is under the terms of the Apache 2.0 and PostgreSQL licenses, and grant any person obtaining a copy of the contribution permission to relicense all or a portion of my contribution to the PostgreSQL License solely to contribute all or a portion of my contribution to the PostgreSQL open source project.

For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/babelfish-for-postgresql/babelfish_extensions/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #2997, opened 2024-10-01
Modify APIs to return palloc'ed values and catalog lookup to validate the user in get_physical_user_name #2960

tanscorpio7:
### Description
Earlier in babelfish,

The get_*_role_name() and get_*_schema_name() APIs returned constant values for a few cases.
The get_physical_user_name() used to return the user_name after appending the db_name to the user_name without any checks to verify its existence/validity.
With this change,

We return Palloc'ed values for the APIs instead of constant values.
In get_physical_user_name() if the user or role is not found in the sys.babelfish_authid_user_ext catalog, then an error is thrown.
Changes Made
In multidb.c, made changes to the APIs like get_role_name() and get_schema_name() to return palloc values instead of constant values.
Implemented a catalog lookup in get_physical_user_name() to verify if the user or role is found in the sys.babelfish_authid_user_ext catalog. The 'suppress_role_error' flag indicates if it is ok for the user or role to be absent from the catalog (required for cases when a new user/role is created and it can't be found in the catalog).
Checked all the functions where the APIs are called and pfree the variables containing values from the above APIs after they are no longer used.

### Issues Resolved

[BABEL-4933]

### Sign Off

Signed-off-by: P Aswini Kumar [aswinii@amazon.com](mailto:aswinii@amazon.com)

### Check List
 Commits are signed per the DCO using --signoff

### DIFF
--- a/contrib/babelfishpg_tsql/src/procedures.c
+++ b/contrib/babelfishpg_tsql/src/procedures.c
@@ -1704,12 +1704,11 @@ create_xp_qv_in_master_dbo_internal(PG_FUNCTION_ARGS)
 	char	   *tempq = "CREATE OR REPLACE PROCEDURE %s.xp_qv(IN SYS.NVARCHAR(256), IN SYS.NVARCHAR(256))"
 	"AS \'babelfishpg_tsql\', \'xp_qv_internal\' LANGUAGE C";
 
-	const char *dbo_scm = get_dbo_schema_name("master");
-
-	if (dbo_scm == NULL)
-		elog(ERROR, "Failed to retrieve dbo schema name");
+	char	   *dbo_scm = get_dbo_schema_name("master");
 
 	query = psprintf(tempq, dbo_scm);
+
+	pfree(dbo_scm);
 
 	PG_TRY();
 	{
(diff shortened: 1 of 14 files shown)


contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -2442,7 +2442,7 @@ sp_addrolemember(PG_FUNCTION_ARGS)
 					 errmsg("User or role '%s' does not exist in this database.", membername)));
 
 		/* Map the logical role name to its physical name in the database. */
-		physical_role_name = get_physical_user_name(get_cur_db_name(), lowercase_rolname);
+		physical_role_name = get_physical_user_name(get_cur_db_name(), lowercase_rolname, true);
```
Deepesh125, line 2445: pfree before re-assignment

Deepesh125 (changes requested):

contrib/babelfishpg_tsql/src/procedures.c
tanscorpio7: We did not do this in 4X. Should we still do it here ?

tanscorpio7: Actually this is first use of this variable.

Deepesh125 (approved):

PR #3007, opened 2024-10-04
Fix CONVERT function behavior for BINARY and VARBINARY types

R4hul04:
### Description

This change addresses inconsistencies in the CONVERT function's behavior when used with BINARY and VARBINARY types in Babelfish. Previously, Babelfish was not correctly handling style arguments and hexadecimal input for these conversions. With this change, the behavior now more closely aligns with SQL Server's implementation.

1. For CONVERT(BINARY(n), '0x...', 1), the function now correctly interprets the input string as hexadecimal values instead of ASCII characters, adhering to the specified style.

2. When NULL is provided as the style argument, the function now correctly returns NULL for both BINARY and VARBINARY types.

3. For both BINARY and VARBINARY conversions, the function now properly considers the    style argument, ensuring consistent behavior across different styles.

4. Fixed an issue where style 2 was not being handled correctly, leading to unexpected results.

### Issues Resolved

Task: BABEL-2736

* **Use case based -** Yes

* **Arbitrary inputs -** Yes

* **Performance tests -**

* **Tooling impact -**

* **Client tests -**

### Check List


Jakeowen1 (approved):

contrib/babelfishpg_tsql/src/backend_parser/gram-tsql-epilogue.y.c
```diff
@@ -216,6 +216,22 @@ TsqlFunctionConvert(TypeName *typename, Node *arg, Node *style, bool try, int lo
 		 */
 		result = makeTypeCast(helperFuncCall, typename, location);
 	}
+
+	else if (strcmp(typename_string, "binary") == 0 || strcmp(typename_string, "varbinary") == 0)
```
fizzlr, line 220: Understand the existing code is using `strcmp`, but will it be safer to use `strncmp` here?

fizzlr:
Is this fix only required for [babelfish-for-postgresql:BABEL_4_X_DEV](https://github.com/babelfish-for-postgresql/babelfish_extensions/tree/BABEL_4_X_DEV)? Is it also needed in BABEL_3_X_DEV?

fizzlr:
Not sure if there's a convenient way to check the test check results of the tests you added, can you add the test output in the PR as well or a link to where they are?

contrib/babelfishpg_tsql/src/backend_parser/gram-tsql-epilogue.y.c
R4hul04: For strncmp we need to specify the length for comparison. 

for eg:
```else if (strncmp(typename_string, "binary", 6) == 0 || strncmp(typename_string, "varbinary", 9) == 0)```

 If "typename_string" is shorter than the comparison length, strncmp will still return 0 (equality) if the available characters match. For example, "bin" would match "binary". This may not be desirable for this usecase.

File: contrib/babelfishpg_tsql/src/backend_parser/gram-tsql-epilogue.y.c [hunk 2]
```diff
@@ -216,6 +216,22 @@ TsqlFunctionConvert(TypeName *typename, Node *arg, Node *style, bool try, int lo
...
+	{
+		if (style == NULL)
+			result = makeTypeCast(arg, typename, location);
```
rohit01010: when style is NULL, we should use its value as default which is 0 in this case. We can use `babelfish_conv_helper_to_varbinary` function with style parameter's default value as 0.

contrib/babelfishpg_tsql/sql/sys_function_helpers.sql
```diff
@@ -9655,6 +9655,43 @@ $BODY$
...
+    IN input_value TEXT,
+    IN style NUMERIC
+) RETURNS sys.varbinary 
+AS 
+$BODY$
+DECLARE
+    result bytea; 
+BEGIN
+    IF style = 0 THEN
+        result := CAST(input_value AS bytea);
```
rohit01010: we can directly cast the input_value as `sys.varbinary` and return it.

contrib/babelfishpg_tsql/sql/sys_function_helpers.sql
```diff
@@ -9655,6 +9655,43 @@ $BODY$
...
+    
+    ELSIF style = 1 THEN
+        -- Handle hexadecimal conversion
+        IF (left(input_value, 2) = '0x' COLLATE sys.database_default AND length(input_value) % 2 = 0) OR input_value IS NULL THEN
```
rohit01010: For PostgreSQL functions such as `left`, `length`, `decode`, `substring` etc. Kindly use namespace `PG_CATALOG` explicitly such as `PG_CATALOG.left()`

contrib/babelfishpg_tsql/sql/sys_function_helpers.sql
```diff
@@ -9655,6 +9655,43 @@ $BODY$
...
+            result := decode(substring(input_value from 3), 'hex');
+        ELSE
+            RAISE EXCEPTION 'Error converting data type varchar to varbinary.';
+        END IF;
+    ELSIF style = 2 THEN
+        IF left(input_value, 2) = '0x' COLLATE sys.database_default THEN
+            RAISE EXCEPTION 'Error converting data type varchar to varbinary.';
+        ELSE
+            result := decode((input_value), 'hex');
```
rohit01010: why we need this extra parenthesis around `input_value` ?

contrib/babelfishpg_tsql/src/backend_parser/gram-tsql-epilogue.y.c
rohit01010: > If "typename_string" is shorter than the comparison length, strncmp will still return 0 (equality) if the available characters match. For example, "bin" would match "binary". This may not be desirable for this usecase.

we can use strlen to check the string length of typename_string and then we can use strncmp to only compare with fixed length.

contrib/babelfishpg_tsql/src/backend_parser/gram-tsql-epilogue.y.c
```diff
@@ -216,6 +216,22 @@ TsqlFunctionConvert(TypeName *typename, Node *arg, Node *style, bool try, int lo
...
+		else
+		{
+			Node	   *helperFuncCall;
+
+			args = list_make2(arg, style);
+			helperFuncCall = (Node *) makeFuncCall(TsqlSystemFuncName("babelfish_conv_helper_to_varbinary"), args, COERCE_EXPLICIT_CALL, location);
```
rohit01010: `TsqlFunctionConvert` also handles behaviour of `TRY_CONVERT`, here irrespective of value of `try`, we are using `babelfish_conv_helper_to_varbinary` function which will throw error for `CONVERT` function but should return `NULL` for `TRY_CONVERT`. Kindly add appropriate handling for `TRY_CONVERT` as well.

rohit01010 (changes requested):

rohit01010:
Kindly make sure tests covering following scenario's are added for both `CONVERT` and `TRY_CONVERT`
- expression is empty string
- when style is not specified, style is 0, style is 1 and style is 2. And its combination with all valid and invalid expression.
- input parameters (expression and/or style) is/are `NULL`
- expression contains multibyte character and/or surrogate pair
- when expression is of datatype: All system types (such as varchar, char, nchar, nvarchar, binary, varbinary, int, float, etc), sysname, User Defined Types. 
- Following dependent objects on convert function which is converting expression to binary/varbinary
   - Function
   - Procedure
   - View
   - Computed Columns
   - Column level constraints
   - Index

Also, Kindly test your changes against all the upgrades we have.

contrib/babelfishpg_tsql/sql/sys_function_helpers.sql
```diff
@@ -9655,6 +9655,43 @@ $BODY$
 LANGUAGE plpgsql
 STABLE;
 
+-- convertion to binary
```
rohit01010: NIT: Kindly update the comment to
```
-- Helper function to convert to binary or varbinary
```

rohit01010 (changes requested):

forestkeeper (approved):

forestkeeper (changes requested):
I have reviewed the pr, the code change looks good to me, but my concern is the test cases seems to be too simple to cover type conversion cases customer may potentially have.
So pls add more cases like :
```
SELECT CONVERT(BINARY(5), col1, null) from table 
```
And maybe :
```
SELECT CONVERT(BINARY(5), cast (col1 as xxx ), null) from table
```
And maybe also variable usage :
```
declare xx as varchar(30);
xx = 'xxxx'
SELECT CONVERT(BINARY(5), @xx, null)
```
And also :
```
SELECT CONVERT(BINARY(5), '', null) + something others 
```
Test the binary result will work fine with other cast/conversions
Or
```
SELECT CONVERT(BINARY(5), '', null)  
Union
.. with other types maybe
```

rohit01010 (approved):
LGTM

rohit01010 (approved):

forestkeeper (dismissed):

forestkeeper (approved):

PR #3011, opened 2024-10-08
Make local variable to evaluate dynamically (only if it appears on TargetList)

Deepesh125:
### Description

Currently, declared local variables are assumed to be static by analyser and hence it is replaced with const node by optimiser. Currently, the parameter value will be supplied through pltsql_param_fetch(..) which is setup during pltsql_estate_setup as part of ParamListInfo setup. This hook is purely used during planning phase and is not used during query execution.

If parameter value is required during execution then values will be read from estate->ndatums through pltsql_param_eval_var callback for EEOP_PARAM_CALLBACK step. This estate->ndatums setup during pltsql_estate_setup and values are copied through copy_pltsql_datums(…). And currently, any assignments to declare variables are implemented using "select target" which will be executed at the end of executor. Hence, any use of variables will read original (value copied during pltsql_estate_setup(...)) and it would not reflect any dynamic change in the value of any variable. 

In order to make this behaviour dynamic, this commit introduces internal function sys.pltsql_assign_var(dno, expr). Argument dno is item number of declared variable and expr could be anything whose value will be assigned to declared variable pointed by dno during execution. We will use this function to rewrite any variables assignment operation during ANTLR parsing as stated in following example,

```
select @a = expr
```
Will be written to
```
select @a = sys.pltsql_assign_var(dno, expr)
```

Internally, sys.pltsql_assign_var invokes exec_assign_value(…) to assign appropriate value directly to variable by updating estate->ndatums array and hence any subsequent read will read latest value.

Engine PR: https://github.com/babelfish-for-postgresql/postgresql_modified_for_babelfish/pull/469
Task: BABEL-5325

### Check List


test/JDBC/expected/BABEL-2843.out
```diff
@@ -225,7 +225,7 @@ Insert on babel_2843_t1  (cost=0.00..0.01 rows=0 width=0) (actual rows=0 loops=1
 text
 Query Text: select * from babel_2843_t1 where a1 = "@param"
 Seq Scan on babel_2843_t1  (cost=0.00..38.25 rows=11 width=8) (actual rows=1 loops=1)
-  Filter: (a1 = 1)
+  Filter: (a1 = $1)
```
KushaalShroff: Can you add explanation on expected behaviour and our change

test/JDBC/expected/BABEL-1181.out
```diff
@@ -64,6 +64,8 @@ set @a = NULL;
 -- top (NULL) should throw error
 select top (@a) * from t1;
 go
+~~START~~
+int
```
KushaalShroff: Interesting change, I get the code behind this change but is it TSQL behaviour?

contrib/babelfishpg_tsql/src/pl_exec.c
```diff
@@ -60,9 +60,9 @@
 #include "err_handler.h"
 #include "iterative_exec.h"
 #include "guc.h"
+#include "hooks.h"
```
KushaalShroff: Why?

contrib/babelfishpg_tsql/src/tsqlIface.cpp
```diff
@@ -3393,6 +3471,76 @@ static void process_query_specification(
...
+				throw PGErrorWrapperException(ERROR, ERRCODE_SYNTAX_ERROR, format_errmsg("\"%s\" is not a known variable", var_str.c_str()), getLineAndPos(elem));
+
+			/* Rewrite @var = expr to @var=sys.pltsql_assign_var(dno, cast((expr) as type)) */
+			repl_text = psprintf("sys.pltsql_assign_var(%d, %s)",
+									nse->itemno,
+									rewrite_assignment_expression((PLtsql_var *) pltsql_Datums[nse->itemno],
+																	elem->expression()));
+
+			handle_local_ids_for_expression(elem->expression());
+			mutator->add(elem->expression()->start->getStartIndex(), ::getFullText(elem->expression()), std::string(repl_text));
```
KushaalShroff, line 3669: After clearing up in the previous function we are using the same elem variable to get the full text.
How is local_id_positions getting init?

test/JDBC/expected/Test-spatial-functions-vu-verify.out
```diff
@@ -404,7 +404,7 @@ GO
 ~~START~~
 numeric#!#int
 <NULL>#!#1
-0#!#2
+0E-8#!#2
```
KushaalShroff, line 407: Why?

test/JDBC/expected/parallel_query/with_recompile-vu-verify.out
```diff
@@ -946,7 +946,7 @@ Query Text: EXEC    p_recomp_22 0 /* EXECUTE WITH RECOMPILE */
 
 ~~START~~
 text
-Babelfish T-SQL Batch Parsing Time: 0.080 ms
+Babelfish T-SQL Batch Parsing Time: 0.111 ms
```
KushaalShroff: Is the parsing time being affected by your changes?

test/JDBC/expected/test_dynamic_local_vars.out
```diff
@@ -0,0 +1,1172 @@
...
+GO
+~~START~~
+int#!#int
+10#!#20
+~~END~~
+
+
+declare @i int
+declare @j int
+select @i = 10, @j = @i + 10
```
KushaalShroff, line 38: Are we testing recursive cases like @j = @j + @i

test/JDBC/expected/test_dynamic_local_vars.out
```diff
@@ -0,0 +1,1172 @@
...
+        SET @IsEven = 0;
+    
+    IF @IsEven = 1
+        PRINT CAST(@Counter AS VARCHAR(2)) + ' is even';
+    ELSE
+        PRINT CAST(@Counter AS VARCHAR(2)) + ' is odd';
+    
+    SET @Counter = @Counter + 1;
+END
+GO
```
KushaalShroff, line 108: No output?

test/JDBC/expected/test_dynamic_local_vars.out
```diff
@@ -0,0 +1,1172 @@
...
+
+~~START~~
+numeric
+300.8200
+~~END~~
+
+
+declare @a varbinary
+set @a = cast('test_bin' as varbinary)
+select @a
```
KushaalShroff, line 198: Mostly we need select @a = cast('test_bin' as varbinary) as well

test/JDBC/expected/test_dynamic_local_vars.out
```diff
@@ -0,0 +1,1172 @@
...
+go
+~~START~~
+int
+-1
+~~END~~
+
+
+-- quoted identifiers
+declare @v varchar(20) = "ABC", @v2 varchar(20)="XYZ";
+select @v += "a""b''c'd", @v2 += "x""y''z";
```
KushaalShroff, line 456: Turn quoted ident ON?

test/JDBC/expected/test_dynamic_local_vars.out
```diff
@@ -0,0 +1,1172 @@
...
+~~START~~
+int#!#int
+3#!#-3
+~~END~~
+
+
+-- xml methods
+DECLARE @a bit = 1
+DECLARE @xml XML = '<artists> <artist name="John Doe"/> <artist name="Edward Poe"/> <artist name="Mark The Great"/> </artists>'
+SELECT @a |= @xml.exist('/artists/artist/@name')
```
KushaalShroff, line 541: Lets update/set the xml in the same select

test/JDBC/expected/test_dynamic_local_vars.out
```diff
@@ -0,0 +1,1172 @@
...
+~~ROW COUNT: 1~~
+
+~~ROW COUNT: 1~~
+
+
+
+
+
+
+-- declare @i int
```
KushaalShroff: Why is this commented?

test/JDBC/expected/test_dynamic_local_vars.out
```diff
@@ -0,0 +1,1172 @@
...
+select @j
+go
+~~START~~
+int
+3
+~~END~~
+
+
+DECLARE @ans INT
+SELECT @ans = AVG(id) FROM local_var_tst
```
KushaalShroff, line 779: Can we create a UDF and pass local vars and/or columns?

test/JDBC/expected/test_dynamic_local_vars.out
```diff
@@ -0,0 +1,1172 @@
...
+GO  
+-- NEXT VALUE FOR gets re-written by ANTLR
+DECLARE @myvar1 BIGINT = NEXT VALUE FOR CountBy1  
+DECLARE @myvar2 BIGINT ;  
+DECLARE @myvar3 BIGINT ;  
+select @myvar2 = NEXT VALUE FOR CountBy1 ;  
+SELECT @myvar3 = NEXT VALUE FOR CountBy1 ;  
+SELECT @myvar1 AS myvar1, @myvar2 AS myvar2, @myvar3 AS myvar3 ;  
+GO
+~~ERROR (Code: 33557097)~~
```
KushaalShroff: Something wrong here?

KushaalShroff (changes requested):

test/JDBC/expected/BABEL-2843.out
Deepesh125: When evaluating constant expressions during the planning phase, we were incorrectly marking local variables as constants by setting param->flags = const through the pltsql_param_fetch function. Due to this, the optimizer was replacing these variables with constant nodes, which was an incorrect. This is fixed now, We no longer make any assumptions about the values of parameters or variables. And hence, the optimizer no longer folds parameters into constants during the optimization process.

contrib/babelfishpg_tsql/src/pl_exec.c
Deepesh125: Removed

contrib/babelfishpg_tsql/src/tsqlIface.cpp
Deepesh125: local_id_positions is getting init inside enter_Local_id which registers all positions of local_id within tsql batch which will be re-written using quotes later. We don't want to do it for assignment expression and so we are just removing positions of local_ids that are part of given expression to avoid re-writing.

test/JDBC/expected/Test-spatial-functions-vu-verify.out
Deepesh125: Previously, we were folding param to const which is fixed in this PR. We also fixed our logic of resolve_numeric_typmod to handle param. Due to that changes, we are seeing this difference. It is expected and correct.

test/JDBC/expected/parallel_query/with_recompile-vu-verify.out
Deepesh125: No

test/JDBC/expected/test_dynamic_local_vars.out
Deepesh125: Added new tests

test/JDBC/expected/test_dynamic_local_vars.out
Deepesh125: Somehow print does not output anything for JDBC tests. It is fixed by replacing print with select

test/JDBC/expected/test_dynamic_local_vars.out
Deepesh125: done

test/JDBC/expected/test_dynamic_local_vars.out
Deepesh125: It will be fixed with separate work. https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/3028 https://github.com/babelfish-for-postgresql/postgresql_modified_for_babelfish/pull/461/files

test/JDBC/expected/test_dynamic_local_vars.out
Deepesh125: done

test/JDBC/expected/test_dynamic_local_vars.out
Deepesh125: fixed it

test/JDBC/expected/test_dynamic_local_vars.out
Deepesh125: fixed

test/JDBC/expected/test_dynamic_local_vars.out
Deepesh125: Done

test/JDBC/expected/BABEL-1181.out
Deepesh125: Yes, I re-confirmed that its expected behaviour

contrib/babelfishpg_tsql/src/tsqlIface.cpp
```diff
@@ -3299,11 +3300,88 @@ class MyParserErrorListener: public antlr4::BaseErrorListener
...
+/*
+ * Necessary checks and mutations for query_specification.
+ * @process_local_id_assignement indicates whether local_id assignement should be re-written or not. Passed false when we are handling
+ * statement like create or alter function.
  */
 static void process_query_specification(
 	TSqlParser::Query_specificationContext *qctx,
-	PLtsql_expr_query_mutator *mutator)
+	PLtsql_expr_query_mutator *mutator,
+	bool process_local_id_assignement)
```
robverschoor: nit: 'assignment', not 'assignement'

test/JDBC/expected/BABEL-2844.out
```diff
@@ -21,7 +21,7 @@ Query Text: create table babel_2844_t1 (a1 int, b1 int);
 
 ~~START~~
 text
-Babelfish T-SQL Batch Parsing Time: 64.545 ms
+Babelfish T-SQL Batch Parsing Time: 64.592 ms
```
robverschoor: these time values cannot be in expected results -- they will always be different

test/JDBC/expected/alter_proc_recompile-vu-verify.out
```diff
@@ -186,12 +186,12 @@ text
 Query Text: EXEC    p_alter_recomp_02 0
   Query Text: select 'p_alter_recomp_02', * from tb_alter_recomp_1 where c1 = "@p"
   ->  Index Scan using tb_alter_recomp_1_pkey on tb_alter_recomp_1
-        Index Cond: (c1 = 0)
+        Index Cond: (c1 = $1)
```
robverschoor: this looks like query plans for regular queries can be affected by this change, since it affects the planner's inputs. My concern is that this fix can lead to changes in query plans for any query involving a variable. In that case there should be a mechanism to stop variables from being handled in this new way.

test/JDBC/expected/test_dynamic_local_vars.out
```diff
@@ -0,0 +1,1303 @@
...
+update local_var_tst set id = 5, @i = id * 5
+select @i
+ROLLBACK tran
+select @i
+GO
+~~ROW COUNT: 3~~
+
+~~START~~
+int
+25
```
robverschoor, line 656: when I try these steps in SQL Server, I get 30 as result

test/JDBC/expected/test_dynamic_local_vars.out
```diff
@@ -0,0 +1,1303 @@
...
+~~END~~
+
+~~START~~
+int
+25
```
robverschoor, line 661: 30 as well here

robverschoor (changes requested):
I tried a random query in MSSQL and I got different results (see review comment).

Also requesting to include tests where SELECT and UPDATE with dependent variables against an actual table, as in a stored procedure.

test/JDBC/expected/test_dynamic_local_vars.out
Deepesh125: That will be fixed with https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/3028

test/JDBC/expected/test_dynamic_local_vars.out
Deepesh125: will be fixed with https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/3028

test/JDBC/expected/alter_proc_recompile-vu-verify.out
Deepesh125: Previously, we were generating wrong result (affecting principal of correctness). And new way is correct way of handling dynamic parameter. Changes in query plan is expected -- previously all the param were being folded as const which was wrong. And with this change, it can definitely spoils plan in some cases but thats expected as we cant assume anything about values at planning time.

test/JDBC/expected/BABEL-2844.out
Deepesh125: Yes, our JDBC framework is capable to ignore these diff.

Deepesh125:
> I tried a random query in MSSQL and I got different results (see review comment).
> 
> Also requesting to include tests where SELECT and UPDATE with dependent variables against an actual table, as in a stored procedure.

Yes, There is diff in UPDATE tests cases which is expected and being worked upon with https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/3028. We have added test cases for test SELECT/INSERT as well as stored procedure

contrib/babelfishpg_tsql/src/tsqlIface.cpp
Deepesh125: Fixed

contrib/babelfishpg_tsql/src/pl_exec.c
```diff
@@ -7858,12 +7858,22 @@ pltsql_param_fetch(ParamListInfo params,
 		}
 	}
 
+	if (saved_expr_kind == EXPRKIND_TARGET)
```
KushaalShroff, line 7862: Safe to reset saved_expr_kind appropriately.

contrib/babelfishpg_tsql/src/pl_exec.c
```diff
@@ -7858,12 +7858,22 @@ pltsql_param_fetch(ParamListInfo params,
...
+	{
+		/* Let extension to set value of param dynamically during execution when variables appears in TargetList */
+		prm->pflags = 0;
+	}
+	else
+	{
+		/* We can always mark params as "const" for executor's purposes */
```
KushaalShroff: Update comment "In other cases we can mark params as const"

contrib/babelfishpg_tsql/src/pl_exec.c
```diff
@@ -10423,3 +10431,32 @@ pltsql_exec_function_cleanup(PLtsql_execstate *estate, PLtsql_function *func, Er
 	}
 	PG_END_TRY();
 }
+
+PG_FUNCTION_INFO_V1(pltsql_assign_var);
+
+Datum
+pltsql_assign_var(PG_FUNCTION_ARGS)
```
KushaalShroff, line 10450: Lets add a descriptive comment on top for the use and also about params

test/JDBC/input/test_dynamic_local_vars.sql
```diff
@@ -0,0 +1,714 @@
+-- simple vars
```
KushaalShroff, line 2: In the latest I see we are not getting $1 for explain output for all const nodes based on your changes. Since we will still get $ for targetlist vars, it would be good to have an explain test case side by side for both

KushaalShroff (changes requested):

contrib/babelfishpg_tsql/src/pl_exec.c
Deepesh125: No, we can't reset saved_expr_kind at this point because this might get called recursively for each targetlist item

test/JDBC/input/test_dynamic_local_vars.sql
Deepesh125: Added

contrib/babelfishpg_tsql/src/pl_exec.c
Deepesh125: Done

contrib/babelfishpg_tsql/src/pl_exec.c
Deepesh125: Added comment

KushaalShroff (approved):

PR #3016, opened 2024-10-11
Extend the function of eliminating unnecessary CAST function for integer

hopebo:
### Description

Some unnecessary CAST functions for integer type exist in the SQL preventing the optimizer making use of the indexes on those columns. By removing those functions, we can give the optimizer an option to choose index.

This patch extends the following supported scenarios:
1. Functions nested in AND/OR clauses.
2. Functions appear in both sides of the operator.
3. Nested CAST functions.
4. CASTs among SMAILLINT, INT, BIGINT.
5. Apart from '=', more operators are supported: </>/<=/>=

### Issues Resolved

Unnecessary CAST function in above scenarios cannot be removed. So the optimizer doesn't have a chance to consider index scan.

BABEL-4550

### Check List

### DIFF
--- a/contrib/babelfishpg_tsql/src/hooks.c
+++ b/contrib/babelfishpg_tsql/src/hooks.c
@@ -5276,56 +5276,149 @@ transform_pivot_clause(ParseState *pstate, SelectStmt *stmt)
 }
 
 
-static Node* optimize_explicit_cast(ParseState *pstate, Node *node)
+static inline bool
+is_integer_type(Oid type)
+{
+	return type == INT2OID || type == INT4OID || type == INT8OID;
+}
+
+static bool
+lower_precision_than(Oid int_type1, Oid int_type2)
+{
+	Assert(is_integer_type(int_type1));
+	Assert(is_integer_type(int_type2));
+
+	switch (int_type1)
+	{
+		case INT2OID:
+			return int_type2 == INT4OID || int_type2 == INT8OID;
+		case INT4OID:
+			return int_type2 == INT8OID;
+		default:
+			Assert(int_type1 == INT8OID);
+			return false;
+	}
+}
+
+static Node *
+optimize_explicit_cast(ParseState *pstate, Node *node)
 {
 	if (sql_dialect != SQL_DIALECT_TSQL)
 		return node;
 	if (node == NULL)
 		return NULL;
-	if (IsA(node, FuncExpr))
+
+	if (IsA(node, FuncExpr) &&
+		((FuncExpr *) node)->funcformat == COERCE_EXPLICIT_CAST)
 	{
-		FuncExpr   *f = (FuncExpr *) node;
-		if (f->funcformat == COERCE_EXPLICIT_CAST){
-			Node* arg = (Node*)linitial(f->args);
-			if (nodeTag(arg) == T_Var && 
-			   ((Var*) arg)->vartype == INT4OID && 
-			   f->funcresulttype == INT8OID)
-				return optimize_explicit_cast(pstate, linitial(f->args));
-		}
+		Node *original_node = node;
+		/*
+		 * Record the integer result type with lowest precision during CAST
+		 * process. If the original SQL casts a variable of integer type to a
+		 * type with less precision, removing the CASE funcition will lead to
+		 * different results. Initial value as highest precision integer type.
+		 */
+		Oid lowest_precision_type = INT8OID;
+		/*
+		 * Deal with nested CAST functions.
+		 * For example: CAST((CAST col_int AS BIGINT) AS INT) => col_int
+		 */
+		do {
+			FuncExpr *f = (FuncExpr *) node;
+
+			if (!is_integer_type(f->funcresulttype))
+				break;
+
+			if (lower_precision_than(f->funcresulttype, lowest_precision_type))
+				lowest_precision_type = f->funcresulttype;
+
+			node = (Node *) linitial(f->args);
+
+		} while (IsA(node, FuncExpr) &&
+                 ((FuncExpr *) node)->funcformat == COERCE_EXPLICIT_CAST);
+
+		/*
+		 * Casting a integer variable to a type with higher precision is
+		 * unnecessary. Optimizer will figure out the appropriate operator to
+		 * calculate without loss of precision.
+		 */
+		if (IsA(node, Var) && is_integer_type(((Var *) node)->vartype) &&
+			!lower_precision_than(lowest_precision_type,
+								  ((Var *) node)->vartype))
+			return node;
+		else
+			return original_node;
 	}
 	else if (IsA(node, BoolExpr))
 	{
 		BoolExpr *r = (BoolExpr *) node;
 		ListCell *l;
-		foreach (l , r->args)
+		foreach (l, r->args)
 		{
-			optimize_explicit_cast(pstate, (Node*)lfirst(l));
+			Node *new_node = optimize_explicit_cast(pstate, (Node*)lfirst(l));
+			if (lfirst(l) != new_node) lfirst(l) = new_node;
 		}
 	}
 	else if (IsA(node, OpExpr))
 	{
 		OpExpr *opExpr = (OpExpr*) node;
-		Form_pg_operator form;
-		HeapTuple	tuple;
-		Node* node1 = optimize_explicit_cast(pstate, linitial(opExpr->args));
-		Node* result = NULL;
-		if (node1 != linitial(opExpr->args))
+		Node *first_old_arg, *fist_new_arg, *second_old_arg, *second_new_arg;
+		Oid oprnamespace;
+		char *opname;
+
 		{
-			char	   *opname;
+			Form_pg_operator optup;
+			HeapTuple tp = SearchSysCache1(OPEROID,
+										   ObjectIdGetDatum(opExpr->opno));
 
-			tuple = SearchSysCache1(OPEROID, ObjectIdGetDatum(opExpr->opno));
-			if (!HeapTupleIsValid(tuple))
-				return node1; // no need to error out in here, stop transform and quite, keep the original node
-			form = (Form_pg_operator) GETSTRUCT(tuple);
+			if (!HeapTupleIsValid(tp))
+				return node;
 
-			opname = NameStr(form->oprname);
-			if (strcmp(opname, "=") == 0)
-			{
-				result =(Node*)make_op(pstate, list_make1(makeString(opname)), node1, lsecond(opExpr->args), pstate->p_last_srf, -1);
-			}
-			ReleaseSysCache(tuple);
-			if (result)
-				return result;
+			optup = (Form_pg_operator) GETSTRUCT(tp);
+			oprnamespace = optup->oprnamespace;
+			opname = pstrdup(NameStr(optup->oprname));
+			ReleaseSysCache(tp);
+		}
+
+		/*
+		 * Only when the operators are built-in operators, we assume there must
+		 * be appropriate operators for the arguments of new type after removing
+		 * CAST functions. Don't do this optimization for user-defined operator.
+		 */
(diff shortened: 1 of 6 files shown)


contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -5276,56 +5276,120 @@ transform_pivot_clause(ParseState *pstate, SelectStmt *stmt)
 }
 
 
+static inline bool is_integer_type(Oid type) {
```
2jungkook: Let's follow PG's coding style (https://www.postgresql.org/docs/current/source-conventions.html), e.g.,
```
static inline bool
is_integer_type(Oid type)
{
    ....
}
```

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -5276,56 +5276,120 @@ transform_pivot_clause(ParseState *pstate, SelectStmt *stmt)
...
+
+	switch (int_type1)
+	{
+		case INT2OID:
+			return int_type2 == INT4OID || int_type2 == INT8OID;
+		case INT4OID:
+			return int_type2 == INT8OID;
+		default:
+		  Assert(int_type1 == INT8OID);
+		  return false;
```
2jungkook: indentation

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -5276,56 +5276,120 @@ transform_pivot_clause(ParseState *pstate, SelectStmt *stmt)
...
+
+			if (!is_integer_type(f->funcresulttype)) break;
+
+			if (lower_precision_than(f->funcresulttype, lowest_precision_type))
+				lowest_precision_type = f->funcresulttype;
+
+			node = (Node *) linitial(f->args);
+
+			if (!IsA(node, FuncExpr)) break;
 		}
```
2jungkook: minor things:
1. `do ... while` is better.
2. we can do an early return if `funcresulttype` is not an integer type. 
3. `IsA(node, FuncExpr)` can be a part of `while` condition.
4. PG puts `break;` in a new line.

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -5276,56 +5276,120 @@ transform_pivot_clause(ParseState *pstate, SelectStmt *stmt)
...
+		 * calculate without loss of precision.
+		 */
+		if (IsA(node, Var) && is_integer_type(((Var *) node)->vartype) &&
+			!lower_precision_than(lowest_precision_type,
+								  ((Var *) node)->vartype))
+			return node;
+		else
+			return original_node;
 	}
 	else if (IsA(node, BoolExpr))
```
2jungkook, line 5352: Let's add a test for `NOT` expression.

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -5276,56 +5276,120 @@ transform_pivot_clause(ParseState *pstate, SelectStmt *stmt)
...
-		Node* node1 = optimize_explicit_cast(pstate, linitial(opExpr->args));
-		Node* result = NULL;
-		if (node1 != linitial(opExpr->args))
+		Node *first_old_arg, *fist_new_arg, *second_old_arg, *second_new_arg;
+
+		/*
+		 * Only the following comparsion operators could make use of index after
+		 * removing unnecessary CAST function, which are our targets.
+		 */
+		char *opname = get_opname(opExpr->opno);
```
2jungkook: This can return NULL when something goes wrong. Let's add a safety check.

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -5276,56 +5276,120 @@ transform_pivot_clause(ParseState *pstate, SelectStmt *stmt)
...
-		HeapTuple	tuple;
-		Node* node1 = optimize_explicit_cast(pstate, linitial(opExpr->args));
-		Node* result = NULL;
-		if (node1 != linitial(opExpr->args))
+		Node *first_old_arg, *fist_new_arg, *second_old_arg, *second_new_arg;
+
+		/*
+		 * Only the following comparsion operators could make use of index after
+		 * removing unnecessary CAST function, which are our targets.
+		 */
```
2jungkook, line 5394: Removing unnecessary cast is (almost) always beneficial. Do we really need to limit our scope by introducing additional syscache lookup and string comparisons? Also, just checking `opname` is not sufficient if we really want to limit the scope because users can create operators with the same names (https://www.postgresql.org/docs/current/sql-createoperator.html).

test/JDBC/expected/cast_eliminate-vu-verify.out
```diff
@@ -120,3 +120,343 @@ text
 Babelfish T-SQL Batch Parsing Time: 0.317 ms
 ~~END~~
 
+
+SELECT 1 AS [C1] FROM cast_eliminate WHERE (CAST(ROID AS BIGINT) = 1) OR (CAST(ROID AS BIGINT) = 2)
+GO
+~~START~~
+text
+Query Text: SELECT 1 AS [C1] FROM cast_eliminate WHERE (CAST(ROID AS BIGINT) = 1) OR (CAST(ROID AS BIGINT) = 2)
+Bitmap Heap Scan on cast_eliminate  (cost=8.32..13.67 rows=2 width=4)
```
2jungkook: Let's remove cost in the expected output because it can cause unnecessary failures in different major versions.

test/JDBC/expected/cast_eliminate-vu-verify.out
```diff
@@ -120,3 +120,343 @@ text
...
+  ->  BitmapOr  (cost=8.32..8.32 rows=2 width=0)
+        ->  Bitmap Index Scan on cast_eliminate_pkey  (cost=0.00..4.16 rows=1 width=0)
+              Index Cond: (roid = 1)
+        ->  Bitmap Index Scan on cast_eliminate_pkey  (cost=0.00..4.16 rows=1 width=0)
+              Index Cond: (roid = 2)
+~~END~~
+
+~~START~~
+text
+Babelfish T-SQL Batch Parsing Time: 0.518 ms
```
2jungkook: I know this is ignored by the JDBC test framework, but it is annoying anyway. Can we remove this from the expected outputs?

test/JDBC/expected/cast_eliminate-vu-verify.out
```diff
@@ -120,3 +120,343 @@ text
...
+~~START~~
+text
+Query Text: SELECT * FROM cast_eliminate WHERE -32768 < CAST(CAST(s_int AS INT) AS BIGINT) OR CAST(ROID AS BIGINT) = -2147483648
+Seq Scan on cast_eliminate (actual rows=5 loops=1)
+  Filter: (('-32768'::integer < s_int) OR (roid = '-2147483648'::integer))
+~~END~~
+
+
+-- Throws an error due to overflow. If we wrongly eliminate the CAST function, the error won't be thrown
+SELECT * FROM cast_eliminate WHERE -32768 < CAST(CAST(s_int AS INT) AS BIGINT) OR -9223372036854775808 = CAST(CAST(b_int AS BIGINT) AS INT)
```
2jungkook, line 385: Could you add the below test case to make sure removing a cast does not make an error?
```
SELECT * FROM cast_eliminate WHERE -9223372036854775808 = CAST(s_int AS BIGINT)
```

test/JDBC/expected/cast_eliminate-vu-verify.out
```diff
@@ -120,3 +120,343 @@ text
...
+  Filter: (s_int = 1)
+~~END~~
+
+~~START~~
+text
+Babelfish T-SQL Batch Parsing Time: 25.653 ms
+~~END~~
+
+
+-- Bad case: cannot remove CAST if a column is typecasted into a type with less precision
```
2jungkook, line 181: Could you add one more negative test case like
```
SELECT 1 AS [C1] FROM cast_eliminate WHERE CAST(CAST(ROID AS numeric) as int) = 1
```
?

contrib/babelfishpg_tsql/src/hooks.c
hopebo: Fixed.

contrib/babelfishpg_tsql/src/hooks.c
hopebo: Added. It works as expected.

contrib/babelfishpg_tsql/src/hooks.c
hopebo: The reasons why I choose to implement in current way have:
1. This feature is a minor optimization while seldom hand-written SQLs could involve such unnecessary CAST functions. I think most of eligible SQLs are generated by framework. I wanna reduce the risk of applying it for all operators. Though I think it'll be okay, I couldn't have the enough tests to ensure that.
2. Only when the existence of CAST function influences the execution plan causing huge performance difference, the customer will notice that. That's the reason why we have this work. If no execution plan changes, I don't think a CAST function will bring observable performance difference for most cases. So I only solve the scenarios where could influence the index choosing.
3. I guess the syscache lookup cannot be avoided. Because we need to make_op according to the new types of arguments, which relies on op_name.

As we talked offline, support only built-in =/</<=/>/>= operators.

test/JDBC/expected/cast_eliminate-vu-verify.out
hopebo: Sure, added.

test/JDBC/expected/cast_eliminate-vu-verify.out
hopebo: Yes. Removed with the following settings:
```
EXEC sp_babelfish_configure 'babelfishpg_tsql.explain_costs', 'off'
EXEC sp_babelfish_configure 'babelfishpg_tsql.explain_timing', 'off'
EXEC sp_babelfish_configure 'babelfishpg_tsql.explain_summary', 'off'
```

test/JDBC/expected/cast_eliminate-vu-verify.out
hopebo: Done.

contrib/babelfishpg_tsql/src/hooks.c
hopebo: Use another way to get information. Checked with `HeapTupleIsValid`.

contrib/babelfishpg_tsql/src/hooks.c
hopebo: 1/3/4 Fixed.
For item 2, I'd like to make them return in one place with same logic. Just one extra `if` judgement here in execution, but in unified logic.

contrib/babelfishpg_tsql/src/hooks.c
hopebo: Thank you for pointing out. I'm getting familiar with these.

test/JDBC/expected/cast_eliminate-vu-verify.out
hopebo: Added. Keep the CAST as expected.

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -5276,56 +5276,147 @@ transform_pivot_clause(ParseState *pstate, SelectStmt *stmt)
 }
 
 
-static Node* optimize_explicit_cast(ParseState *pstate, Node *node)
+static inline bool
+is_integer_type(Oid type) {
```
2jungkook: style - starting bracket `{` in a new line, e.g.,
```
is_integer_type(...)
{
    ....
}
```

hopebo: Fixed.

2jungkook (approved):

forestkeeper (approved):

PR #3036, opened 2024-10-22
cleanup pltsql states properly in exec_stmt_exec_batch in case of error

tanscorpio7:
### Description

Some system states like pltsql guc nest level, pltsql scope identity should always be cleaned up at the end of execution even in case of errors since these were defined to be outside of postgres's transaction cleanup control.
So a `pltsql_new_guc_nest_level()` should always have a `pltsql_revert_guc` in all cases and vice versa.

This was not happening inside `exec_stmt_exec_batch` where we could error out even before starting a new pltsql guc nest level and then in catch block we would try to revert a non existent nest level. 

As a fix we move out the call to pltsql_new_guc_nest_level() before try block ensuring we have a nest level to revert in catch.

### Issues Resolved

[BABEL-5344]

### Sign Off

### Check List


contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -1548,12 +1548,12 @@ exec_stmt_exec_batch(PLtsql_execstate *estate, PLtsql_stmt_exec_batch *stmt)
 	PG_FINALLY();
 	{
 		/* Restore past settings */
+		pltsql_revert_guc(save_nestlevel);
+		pltsql_revert_last_scope_identity(scope_level);
+
```
tanscorpio7, line 1553: Unsure if `set_session_properties` can fail. Seems easier to just revert the nest levels first.

2jungkook (approved):

PR #3017, opened 2024-10-12
fix for storing the correct object definition in catalog

pranavJ23:
### Description
Problem: If we create a procedure(let's say) and in that procedure we declare a variable of of some datatype with brackets around like 'DECLARE @SQL [varchar] (8000)' , the babelfish catalog stores it as 'DECLARE @SQL varchar (8000)'  i.e the brackets are not stored properly in catalog.
Or if we add a single line comment before a query(like alter, create query) then while logging the definition of that query from catalog we do not get the single line comment, which is not correct.

RCA: while parsing the query we mutate the parse tree in such a way that it modifies various things from the original string  like single line comment, brackets, multi line comments.

Fix: store the original query string before mutation and convert alter to create in case or alter procedure/function( batch level statement) and pass that query string. 
### Issues Resolved
[BABEL-5237][BABEL-5236][BABEL-5140]
### Sign Off
Signed-off-by: Pranav Jain [pranavjc@amazon.com](mailto:pranavjc@amazon.com)
### Check List
- [ ] Commits are signed per the DCO using --signoff 

By submitting this pull request, I confirm that my contribution is under the terms of the Apache 2.0 and PostgreSQL licenses, and grant any person obtaining a copy of the contribution permission to relicense all or a portion of my contribution to the PostgreSQL License solely to contribute all or a portion of my contribution to the PostgreSQL open source project.

For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/babelfish-for-postgresql/babelfish_extensions/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #3037, opened 2024-10-22
fix for storing the correct object definition in catalog

pranavJ23:
### Description
Cherry pick from : https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/3017
Problem: If we create a procedure(let's say) and in that procedure we declare a variable of of some datatype with brackets around like 'DECLARE @Sql [varchar] (8000)' , the babelfish catalog stores it as 'DECLARE @Sql varchar (8000)' i.e the brackets are not stored properly in catalog.
Or if we add a single line comment before a query(like alter, create query) then while logging the definition of that query from catalog we do not get the single line comment, which is not correct.

RCA: while parsing the query we mutate the parse tree in such a way that it modifies various things from the original string like single line comment, brackets, multi line comments.

Fix: store the original query string before mutation and convert alter to create in case or alter procedure/function( batch level statement) and pass that query string.

nit: renamed the file name with the suffix from before_15_9 to before_16_5_or_15_9

### Issues Resolved
[BABEL-5237][BABEL-5236][BABEL-5140]

### Check List


test/JDBC/expected/BABEL-5059-vu-verify.out
```diff
@@ -0,0 +1,1248 @@
+-- Procedure
```
tanscorpio7: Is this new file part of your changes ?

tanscorpio7 (changes requested):

test/JDBC/expected/BABEL-5059-vu-verify.out
pranavJ23: this file was there in 4x but not in 3x, so it get cherry picked here, removing it here as it is not necessary

PR #3067, opened 2024-10-31
[OSS-ONLY] Update antlr to 4.13.2

Sairakan:
### Description

Update the antlr version to 4.13.2. This dramatically improves ANTLR parsing performance for certain queries, reducing their execution times by up to 40%.

Note that this will require developers to have Java 11+ installed in order to develop locally.

Also update the github actions to make it so that previous versions of Babelfish will be compiled using the original antlr-4.9.3 version. This is purely a fix for testing.

### Issues Resolved

BABEL-5371

* **Use case based -**

* **Boundary conditions -**

* **Arbitrary inputs -**

* **Negative test cases -**

* **Performance tests -**

* **Tooling impact -**

* **Client tests -**

### Check List


contrib/babelfishpg_tsql/src/tsqlIface.cpp
```diff
@@ -574,7 +576,7 @@ class MyInputStream : public ANTLRInputStream
     
 		void setText(size_t pos, const char *newText)
 		{
-			UTF32String	newText32 = utf8_to_utf32(newText);
```
forestkeeper, line 577: Why this change happens ?

contrib/babelfishpg_tsql/src/tsqlIface.cpp
```diff
@@ -560,7 +562,7 @@ format_errmsg(const char *fmt, const char *arg1, const char *arg2);
 
 inline std::u32string utf8_to_utf32(const char* s)
 {
-	return antlrcpp::utf8_to_utf32(s, s + strlen(s));
```
forestkeeper, line 563: What's the difference between those two functions?

contrib/babelfishpg_tsql/src/tsqlIface.cpp
Sairakan: I think that `UTF32String` may have been an acceptable macro in an older C++ version? This was raising a compilation issue after using C++17 for compilation. Based on everything I could find, `std::u32string` seems to be the widely accepted standard type to use for 32-bit character strings.

contrib/babelfishpg_tsql/src/tsqlIface.cpp
Sairakan: utf8_to_utf32 was removed in v4.10 and replaced with lenient/strict decode https://github.com/antlr/antlr4/commit/2349fb20199303eef2a742f9449d73c4902e83c9

I opted to go with lenientDecode in order to prevent errors from being raised, but we can also change this to strictDecode if that behavior is more appropriate

.github/composite-actions/minor-version-upgrade-util/action.yml
```diff
@@ -21,13 +21,24 @@ runs:
...
         engine_branch: ${{ inputs.engine_branch }}
         install_dir: ${{ inputs.install_dir }}
 
+    - name: Compile new ANTLR
+      id: compile-new-antlr
+      if: always() && steps.build-modified-postgres-new.outcome == 'success'
+      uses: ./.github/composite-actions/compile-antlr
+      with:
+        version: ${{ inputs.engine_branch == 'latest' && '4.13.2' || '4.9.3' }}
+        install_dir: ${{inputs.install_dir}}
```
rishabhtanwar29: Are we building new antlr only for latest version? I think we should instead update this condition to build latest antlr for all the versions from now on.

Sairakan: The problem with trying to use the latest antlr for all versions including older versions is that the github workflows will break when trying to compile any older versions. The only way around this would be to find all older versions which are being built from the github actions, and first update them to use antlr-4.13.2. I think that is too much work, which is why I went with this solution of allowing older versions to compile with 4.9.3 and only use 4.13.2 for builds against the latest version.

Sairakan: You can see the extremely long chain of commits I had getting this to work, where all of the upgrade tests were failing due to mismatches in antlr version. The main reason why this happens is that all of the github workflows assume that antlr will be compiled separately from the extension, so they were written to compile antlr once before building the older version, and just use that same antlr build for the updated version as well.

rishabhtanwar29: Building old engine versions with antlr-4.9.3 is fine, what I meant was to build current and newer engine versions with new antlr-4.13.2. For example in future, in version 16.7, the above condition will build only the latest version with antlr-4.13.2 but it will still build 16.6 with old antlr-4.9.3 which doesn't seem right. So, I was thinking if we can make this condition something like:
```
if (version >= 16.6 then use antlr-4.13.2 else use antlr-4.9.3)
```

Sairakan: Good point. Let me try to figure out how to set up this condition to work properly.

rishabhtanwar29 (approved):

PR #3072, opened 2024-11-04
Fix the nvarchar-varbinary casting

pranavJ23:
### Description
PROBLEM: while casting nvarchar to varbinary we were considering the encoding as input encoding in babelfish where as in tsql we use UTF16 encoding fir nvarchar irrespective of input encoding.

RCA: we were considering varchar and nvarchar as same, whereas we should use input encoding for varchar and UTF16 encoding for nvarchar.

FIX: So we need to identify that if the input is nvarchar then we will do the UTF16 encoding.

For a casting like nvarchar->varbinary->nvarchar, now since for the casting we are encoding the input string into UTF16 encoding via function nvarcharvarbinary, so while converting varbinary-> nvarchar we will use the function varbinarynvarchar where we will convert UTF16 encoding to UTF8 with null padding.
So we created a function nvarcharvarbinary and varbinarynvarchar to handle nvarchar<-> varbinary to and fro casting.
And for this casting we have specifically applied a condition that we will not convert the datatype to basetype before choosing the casting function

### Issues Resolved
[BABEL-4891]

### Check List


contrib/babelfishpg_common/src/varchar.c
```diff
@@ -176,6 +176,178 @@ GetUTF8CodePoint(const unsigned char *in, int len, int *consumed_p)
 	return code;
 }
 
+void
+AddUTF16ToStringInfo(int32_t code, StringInfo buf)
+{
+	union
```
shah-nirmit, line 182: Can we use bitshift intead of union?
```
uint16_t temp16;

temp16 = 0xD800 + (((code - 0x010000) >> 10) & 0x03FF);
appendStringInfoChar(buf, (uint8_t)(temp16 >> 8));    // high byte
appendStringInfoChar(buf, (uint8_t)(temp16 & 0xFF));  // low byte

```

contrib/babelfishpg_common/src/varchar.h
```diff
@@ -4,8 +4,14 @@
...
 
 extern int32_t GetUTF8CodePoint(const unsigned char *in, int len, int *consumed_p);
+extern int32_t GetUTF16CodePoint(const unsigned char *in, int len, int *consumed);
+extern void AddUTF16ToStringInfo(int32_t code, StringInfo buf);
+extern void AddUTF8ToStringInfo(int32_t code, StringInfo buf);
 extern void *tsql_varchar_input(const char *s, size_t len, int32 atttypmod);
 extern void *tsql_bpchar_input(const char *s, size_t len, int32 atttypmod);
 extern int  TsqlUTF8LengthInUTF16(const void *vin, int len);
-
+extern void TsqlUTF8toUTF16StringInfo(StringInfo out, const void *vin, size_t len);
```
shah-nirmit, line 10: Do we need `void *` for `vin`? I think it will always have a fixed datatype right?

contrib/babelfishpg_tsql/src/string.c
```diff
@@ -56,11 +56,20 @@ static Datum return_varchar_pointer(char *buf, int size);
...
 	const char *algorithm = text_to_cstring(PG_GETARG_TEXT_P(0));
 	bytea	   *in = PG_GETARG_BYTEA_PP(1);
 	size_t		len = VARSIZE_ANY_EXHDR(in);
 	const uint8 *data = (unsigned char *) VARDATA_ANY(in);
 	bytea	   *result;
+	StringInfoData 		utf16_data;
+
+	if(((*common_utility_plugin_ptr->is_tsql_nvarchar_datatype)(input_type)))
+	{
+		(common_utility_plugin_ptr->tsql_utf8_to_utf16)(&utf16_data, data, len);
```
shah-nirmit: it should not make difference if you sent uint8 or uchar right? also should we free the stringinfo post its use?

contrib/babelfishpg_tsql/src/pltsql_coerce.c
```diff
@@ -1288,6 +1309,65 @@ validate_special_function(char *func_nsname, char *func_name, List* fargs, int n
...
+	Oid 						expr_second_arg;
+	char					   *proc_nsname;
+	char					   *proc_name;
+	int							ncandidates;
+	Oid							sys_oid = get_namespace_oid("sys", false);
+	Oid 						*argtypes;
+	int							nargs_func = 0;
+	Oid							second_arg;
+
+	DeconstructQualifiedName(names, &proc_nsname, &proc_name);
```
shah-nirmit: Do we need to call the same here? We already doing this in the caller of the func.

contrib/babelfishpg_tsql/src/pltsql_coerce.c
```diff
@@ -1288,6 +1309,65 @@ validate_special_function(char *func_nsname, char *func_name, List* fargs, int n
...
+	
+	/* if common_utility_plugin_ptr is not initialised */
+	if (common_utility_plugin_ptr == NULL)
+		ereport(ERROR,
+				(errcode(ERRCODE_INTERNAL_ERROR),
+					errmsg("Failed to find common utility plugin.")));
+
+	expr_second_arg = InvalidOid;
+	if (strncmp(proc_name,"hashbytes", 9) == 0)
```
shah-nirmit: Why we need to check this again here?

contrib/babelfishpg_tsql/src/pltsql_coerce.c
```diff
@@ -1308,6 +1388,12 @@ tsql_func_select_candidate_for_special_func(List *names, List *fargs, int nargs,
 
 	DeconstructQualifiedName(names, &proc_nsname, &proc_name);
 
+
```
shah-nirmit: Can you please explain why we need to do this before special function validation?

contrib/babelfishpg_tsql/src/pltsql_coerce.c
```diff
@@ -1449,6 +1536,7 @@ tsql_func_select_candidate_for_special_func(List *names, List *fargs, int nargs,
 	/* Get the candidate with matching return type */
 	ncandidates = 0;
 	best_candidate = NULL;
+
```
shah-nirmit: Nit: Please avoid whitespace changes

contrib/babelfishpg_tsql/src/pltsql_coerce.c
```diff
@@ -507,22 +513,37 @@ tsql_find_coercion_pathway(Oid sourceTypeId, Oid targetTypeId, CoercionContext c
...
 
 	/* Perhaps the types are domains; if so, look at their base types */
 	if (!isSqlVariantCast)
 	{
-		if (OidIsValid(sourceTypeId))
+		if (OidIsValid(sourceTypeId) && !isNvarchartoVarbinary)
 			sourceTypeId = getBaseType(sourceTypeId);
 
 		/*
 		 * if we are casting from INT8 to MONEY, don't look for base type of
```
shah-nirmit: Please update the comments

contrib/babelfishpg_tsql/src/pltsql_coerce.c
```diff
@@ -507,22 +513,37 @@ tsql_find_coercion_pathway(Oid sourceTypeId, Oid targetTypeId, CoercionContext c
...
+
+			/* Check if type is INT8 */
+			if (strcmp(type_nsname, "sys") == 0 && strcmp(type_name, "nvarchar") == 0)
+				isNvarchar = true;
+
 			ReleaseSysCache(tuple);
 		}
 	}
 
 	/* Perhaps the types are domains; if so, look at their base types */
```
shah-nirmit, line 569: Update comments

contrib/babelfishpg_common/src/varbinary.c
```diff
@@ -693,6 +699,82 @@ varcharvarbinary(PG_FUNCTION_ARGS)
...
+	}
+	PG_CATCH();
+	{
+		MemoryContext ectx;
+		ErrorData    *errorData;
+
+		ectx = MemoryContextSwitchTo(ccxt);
+		errorData = CopyErrorData();
+		FlushErrorState();
+		MemoryContextSwitchTo(ectx);
```
shah-nirmit, line 778: Nit: Need to update Error message here.

contrib/babelfishpg_common/src/varchar.c
```diff
@@ -443,7 +615,12 @@ tsql_varchar_input(const char *s, size_t len, int32 atttypmod)
 {
 	return varchar_input(s, len, atttypmod);
 }
-
+void
+tsql_utf8_to_utf16(StringInfoData *utf16_data, const uint8 *data, size_t len)
+{
+    initStringInfo(utf16_data);
```
shah-nirmit: lets move initStringInfo to caller

contrib/babelfishpg_common/src/varbinary.c
```diff
@@ -693,6 +699,82 @@ varcharvarbinary(PG_FUNCTION_ARGS)
...
+	int32		typmod = PG_GETARG_INT32(1);
+	bool		isExplicit = PG_GETARG_BOOL(2);
+	int32		maxlen;
+	bytea	        *result;
+	int		encodedByteLen;
+	StringInfoData  s; 
+	MemoryContext   ccxt = CurrentMemoryContext;
+
+	if (!isExplicit)
+		ereport(ERROR,
```
shah-nirmit, line 752: Do we have test cases for this scenario?

contrib/babelfishpg_common/src/varbinary.c
```diff
@@ -693,6 +699,82 @@ varcharvarbinary(PG_FUNCTION_ARGS)
...
+						"varbinary is not allowed. Use the CONVERT function "
+						"to run this query.")));
+
+	PG_TRY();
+	{
+		/*
+		 * For nvarchar convert the string to UTF16 from UTF8 irrespective of input encoding via TsqlUTF8toUTF16StringInfo()
+		 * For this we need to prepare a StringInfoData() and assign the encoded_data, encodedByteLen from the string info data we prepared
+		 */
+		initStringInfo(&s);
```
shah-nirmit: can we free stringinfo after its use?

shah-nirmit (changes requested):

contrib/babelfishpg_tsql/src/pltsql_coerce.c
pranavJ23: we are setting the proc_name there(DeconstructQualifiedName(names, &proc_nsname, &proc_name)) after which we compare proc_name with 'hashbytes'. So we need that before calling tsql_func_select_candidate_for_hashbytes

contrib/babelfishpg_tsql/src/pltsql_coerce.c
pranavJ23: Yes we can skip this check

contrib/babelfishpg_common/src/varbinary.c
pranavJ23: adding one for this

contrib/babelfishpg_tsql/src/string.c
pranavJ23: Yeah it won't make difference, it's just uint8, was used originally(before my changes) for data so I took it uint8 only.
Can change this.

contrib/babelfishpg_common/src/varchar.c
pranavJ23: I followed how we are doing this conversion in ''TdsUTF16toUTF8StringInfo''

contrib/babelfishpg_common/src/varchar.h
pranavJ23: We can remove that but
I followed how we are doing this conversion in ''TdsUTF16toUTF8StringInfo''

contrib/babelfishpg_tsql/src/pltsql_coerce.c
shah-nirmit: I did not mean that , I mean why are we not following the logic flow similar to other special functions like trim, replace etc. Why we need to diverge?

contrib/babelfishpg_tsql/src/pltsql_coerce.c
shah-nirmit: ??

contrib/babelfishpg_common/src/varchar.h
shah-nirmit: We can still have single datatype right? Helps in readability , otherwise one needs to check all the callers of the function to figure out what exactly is vin.

contrib/babelfishpg_tsql/src/pltsql_coerce.c
```diff
@@ -507,22 +513,44 @@ tsql_find_coercion_pathway(Oid sourceTypeId, Oid targetTypeId, CoercionContext c
...
+				isVarbinary = true;
+
+			/* We've found NVARCHAR TO VARBINARY casting */
+			if (isNvarchar && strcmp(type_nsname, "sys") == 0 && (strcmp(type_name, "varbinary") == 0))
+				isNvarchartoVarbinary = true;
+
+			/* Check if type is NVARCHAR */
+			if (strcmp(type_nsname, "sys") == 0 && strcmp(type_name, "nvarchar") == 0)
+				isNvarchar = true;
+
```
rohit01010: Kindly tests casting of User Defined Types on `varchar/nvarchar` to `varbinary/binary`, and vice versa.

contrib/babelfishpg_common/src/varbinary.c
```diff
@@ -835,6 +917,89 @@ varbinaryvarchar(PG_FUNCTION_ARGS)
...
+	char	        *data = VARDATA_ANY(source);		/* Source data is UTF16 encoded */
+	VarChar         *result;
+	char 	        *encoded_result;
+	size_t		len = VARSIZE_ANY_EXHDR(source);
+	int32		typmod = -1;
+	int32		maxlen = -1;
+	int		encodedByteLen;
+	int 		paddedLen;
+	char 		*paddedData;
+	StringInfoData 	s;
```
rohit01010: NIT: Lets rename this StringInfoData variable `s` to some meaningful name, like `utf8_data` or `tempBuf`

contrib/babelfishpg_common/src/varbinary.c
```diff
@@ -694,12 +700,87 @@ varcharvarbinary(PG_FUNCTION_ARGS)
...
+	char	        *data = VARDATA_ANY(source);		/* Source string is UTF-8 */
+	char	        *encoded_data;
+	char	        *rp;
+	size_t		len = VARSIZE_ANY_EXHDR(source);
+	int32		typmod = PG_GETARG_INT32(1);
+	bool		isExplicit = PG_GETARG_BOOL(2);
+	int32		maxlen;
+	bytea	        *result;
+	int		encodedByteLen;
+	StringInfoData  s;
```
rohit01010: NIT: Lets rename this StringInfoData variable `s` to some meaningful name, like `utf16_data` or `tempBuf`

contrib/babelfishpg_common/src/varbinary.c
rohit01010: Also free the StringInfoData variable's data after its use.

contrib/babelfishpg_common/src/varchar.c
```diff
@@ -55,7 +55,7 @@ static inline int varcharTruelen(VarChar *arg);
  * is_basetype_nchar_nvarchar - given datatype is nvarchar or nchar
  *     or created over nvarchar or nchar.
  */
-static bool
+bool
 is_basetype_nchar_nvarchar(Oid typid)
```
rohit01010: if we are not using `is_basetype_nchar_nvarchar` in any other files, we should keep it as static only.

contrib/babelfishpg_common/src/varchar.h
```diff
@@ -4,8 +4,14 @@
 #include "postgres.h"
 
 extern int32_t GetUTF8CodePoint(const unsigned char *in, int len, int *consumed_p);
+extern int32_t GetUTF16CodePoint(const unsigned char *in, int len, int *consumed);
+extern void AddUTF16ToStringInfo(int32_t code, StringInfo buf);
+extern void AddUTF8ToStringInfo(int32_t code, StringInfo buf);
```
rohit01010: If these functions are not being used in any other files, we should declare them as static.

contrib/babelfishpg_common/src/varchar.c
```diff
@@ -443,7 +615,11 @@ tsql_varchar_input(const char *s, size_t len, int32 atttypmod)
 {
 	return varchar_input(s, len, atttypmod);
 }
-
+void
+tsql_utf8_to_utf16(StringInfoData *utf16_data, const unsigned char *data, size_t len)
+{
+    TsqlUTF8toUTF16StringInfo(utf16_data, data, len);
+}
```
rohit01010: why we need this wrapper function `tsql_utf8_to_utf16` ?

contrib/babelfishpg_tsql/src/string.c
```diff
@@ -52,22 +52,41 @@ static Datum return_varchar_pointer(char *buf, int size);
  *
  * We use the Postgres implementations where available (for sha256 and md5),
  * and OpenSSL for the rest.
+ * 
+ * For varchar/nvarchar the input string is in UTF-8 encoding.
+ * So for varchar we keep it as it is but for nvarchar we change the encodign to UTF-16
```
rohit01010: NIT: ~encodign~ encoding

contrib/babelfishpg_tsql/src/pltsql_coerce.c
```diff
@@ -230,6 +232,7 @@ tsql_cast_raw_info_t tsql_cast_raw_infos[] =
 	{TSQL_CAST_ENTRY, "sys", "bpchar", "pg_catalog", "name", "bpchar_to_name", 'i', 'f'},
 	{TSQL_CAST_ENTRY, "pg_catalog", "varchar", "pg_catalog", "name", "varchar_to_name", 'i', 'f'},
 	{TSQL_CAST_ENTRY, "sys", "varchar", "pg_catalog", "name", "varchar_to_name", 'i', 'f'},
+	{TSQL_CAST_ENTRY, "sys", "nvarchar", "sys", "bbf_varbinary", "nvarcharvarbinary", 'i', 'f'},
```
rohit01010: `sys.nvarchar` to `sys.bbf_varbinary` cast is defined with `ASSIGNMENT` cast context, and is marked as `IMPLICIT` here. The `castcontext` of `nvarcharvarbinary` cast entry should be marked as `a` (i.e. `ASSIGNMENT`)

contrib/babelfishpg_tsql/src/pltsql_coerce.c
```diff
@@ -475,8 +478,21 @@ tsql_find_coercion_pathway(Oid sourceTypeId, Oid targetTypeId, CoercionContext c
...
 	Oid			typeIds[2] = {sourceTypeId, targetTypeId};
+	Oid			UDT_sourceBaseType;
+
+	/* Check if the UDT's base type is nvarchar or varbinary. If so, use the immediate base type for further processing. */
+	UDT_sourceBaseType = get_immediate_base_type_of_UDT_internal(sourceTypeId);
+	if(UDT_sourceBaseType != InvalidOid && ((*common_utility_plugin_ptr->is_tsql_nvarchar_datatype)(UDT_sourceBaseType) || (*common_utility_plugin_ptr->is_tsql_sys_varbinary_datatype)(UDT_sourceBaseType)))
+	{
+		typeIds[0] = UDT_sourceBaseType;
+		sourceTypeId = UDT_sourceBaseType;
+	}
```
rohit01010: kindly check whether we want to do the same with `targetTypeId` as well or not ?

contrib/babelfishpg_tsql/src/pltsql_coerce.c
```diff
@@ -1288,6 +1326,66 @@ validate_special_function(char *func_nsname, char *func_name, List* fargs, int n
...
+		/* we should only consider candidates for hashbytes function from sys schema */
+		if (get_func_namespace(current_candidate->oid) != sys_oid)
+			return NULL;
+
+		get_func_signature(current_candidate->oid,&argtypes, &nargs_func);
+		second_arg = argtypes[1];
+		if(second_arg == expr_second_arg)
+		{
+			best_candidate = current_candidate;
+			ncandidates++;
```
rohit01010: why are we maintaining `ncandidates` ?

contrib/babelfishpg_tsql/src/pltsql_coerce.c
```diff
@@ -1288,6 +1326,66 @@ validate_special_function(char *func_nsname, char *func_name, List* fargs, int n
...
+	ncandidates = 0;
+	best_candidate = NULL;
+
+	for (current_candidate = candidates;
+		current_candidate != NULL;
+		current_candidate = current_candidate->next)
+	{
+		/* we should only consider candidates for hashbytes function from sys schema */
+		if (get_func_namespace(current_candidate->oid) != sys_oid)
+			return NULL;
```
rohit01010: why are we returning `NULL` if any of the candidate does not belong to `sys` schema ? we should instead skip that particular candidate and look for required candidate from remaining candidates.

contrib/babelfishpg_tsql/src/pltsql_coerce.c
rohit01010: yeah, we can directly pass value of `proc_nsname` and `proc_name` from caller function.

rohit01010 (changes requested):

rohit01010:
kindly make sure we have tests covering following scenario's
- Add tests for casting from nvarchar to varbinary and vice versa in verify scripts
- Add Function/Procedure/View containing cast from nvarchar to varbinary and vice versa in prepare scripts and call these function/procedure/view in verify scripts
- Test cast from nvarchar <-> varbinary in when
    - Cast context is explicit, implicit, assignment 
- Test cast from 
    - (UDT on nvarchar) <-> varbinary
    - nvarchar <-> (UDT on varbinary)
    - (UDT on nvarchar) <-> (UDT on varbinary)
- Add Function/Procedure/View on hashbytes function in prepare scripts and call these function/procedure/view in verify scripts. (Hashbytes should be called on sys.VARCHAR input type, sys.NVARCHAR input type and sys.varbinary input type)
- Hashbytes function called on string literal (with and without prefixed ’N’)
- Hashbytes function called on variables of type varchar, nvarchar and varbinary
- Hashbytes function called on table columns of type varchar, nvarchar and varbinary

rohit01010:
Kindly add the test files introduced, into the schedule file of all the upgrades.

contrib/babelfishpg_tsql/src/pltsql_coerce.c
pranavJ23: we can remove this variable

contrib/babelfishpg_tsql/src/pltsql_coerce.c
pranavJ23: yes we can skip this.

contrib/babelfishpg_tsql/src/pltsql_coerce.c
pranavJ23: Done

contrib/babelfishpg_common/sql/upgrades/babelfish_common_helper--4.4.0--4.5.0.sql
```diff
@@ -0,0 +1,27 @@
...
+
+CREATE OR REPLACE FUNCTION sys.nvarcharvarbinary(sys.NVARCHAR, integer, boolean)
+RETURNS sys.BBF_VARBINARY
+AS 'babelfishpg_common', 'nvarcharvarbinary'
+LANGUAGE C IMMUTABLE STRICT PARALLEL SAFE;
+
+CREATE CAST (sys.NVARCHAR AS sys.BBF_VARBINARY)
+WITH FUNCTION sys.nvarcharvarbinary (sys.NVARCHAR, integer, boolean) AS ASSIGNMENT;
+
+CREATE OR REPLACE FUNCTION sys.varbinarysysnvarchar(sys.BBF_VARBINARY, integer, boolean)
```
tanscorpio7, line 15: let use the existing naming convention for this function. Also please commit one line at end of each file.

contrib/babelfishpg_common/sql/upgrades/babelfishpg_common--3.8.0--4.0.0.sql
```diff
@@ -0,0 +1,2 @@
+-- complain if script is sourced in psql, rather than via ALTER EXTENSION
+\echo Use "ALTER EXTENSION ""babelfishpg_common"" UPDATE TO '4.0.0'" to load this file. \quit
```
tanscorpio7: why is this file needed ?

contrib/babelfishpg_tsql/src/string.c
```diff
@@ -52,22 +52,41 @@ static Datum return_varchar_pointer(char *buf, int size);
...
+	StringInfoData 	utf16_data;
+
+	initStringInfo(&utf16_data);
+	/*
+	 * If the input_type is nvarchar then we convert it to UTF-16 encoding
+	 */
+	if(((*common_utility_plugin_ptr->is_tsql_nvarchar_datatype)(input_type)))
+	{
+		(common_utility_plugin_ptr->tsql_utf8_to_utf16)(&utf16_data, data, len);
+		data = (const unsigned char *) utf16_data.data;
```
tanscorpio7: declare `initStringInfo(&utf16_data);` inside this if block, then palloc it again into `data`.
This way you only need to pfree `utf16_data` here.

contrib/babelfishpg_tsql/src/pltsql_coerce.c
tanscorpio7: We cannot break the tie when user has also defined a hashbytes functions of their own. In this case we will simply fallback to current behvaiour.

contrib/babelfishpg_tsql/src/pltsql_coerce.c
tanscorpio7: Right, we can just return when we matched the 2nd arg.

contrib/babelfishpg_tsql/src/pltsql_coerce.c
```diff
@@ -1288,6 +1326,66 @@ validate_special_function(char *func_nsname, char *func_name, List* fargs, int n
...
+	if (common_utility_plugin_ptr == NULL)
+		ereport(ERROR,
+				(errcode(ERRCODE_INTERNAL_ERROR),
+					errmsg("Failed to find common utility plugin.")));
+
+	expr_second_arg = InvalidOid;
+	if ((*common_utility_plugin_ptr->is_tsql_nvarchar_datatype)(input_typeids[1]))
+	{
+		expr_second_arg = input_typeids[1];		
+	}
```
tanscorpio7: Are we sure there will be at least 2 arguments ? What if user defined a hasbytes function with one arg in dbo schema. and tries to call it ?

contrib/babelfishpg_tsql/src/pltsql_coerce.c
tanscorpio7: we should change the argument of the function and directly use proc_nsname & proc_name

contrib/babelfishpg_tsql/src/pltsql_coerce.c
```diff
@@ -1288,6 +1326,66 @@ validate_special_function(char *func_nsname, char *func_name, List* fargs, int n
...
+		current_candidate != NULL;
+		current_candidate = current_candidate->next)
+	{
+		/* we should only consider candidates for hashbytes function from sys schema */
+		if (get_func_namespace(current_candidate->oid) != sys_oid)
+			return NULL;
+
+		get_func_signature(current_candidate->oid,&argtypes, &nargs_func);
+		second_arg = argtypes[1];
+		if(second_arg == expr_second_arg)
```
tanscorpio7: can we just compare `argtypes[1] ==  input_typeids[1];`
and which ever function matches, we select it. For no match fall back to current behaviour.

contrib/babelfishpg_tsql/src/pltsql_coerce.c
tanscorpio7: Do we really need to check for the 2nd arg to be nvarchar ?

contrib/babelfishpg_tsql/src/pltsql_coerce.c
```diff
@@ -475,8 +478,21 @@ tsql_find_coercion_pathway(Oid sourceTypeId, Oid targetTypeId, CoercionContext c
...
 
 	Oid			typeIds[2] = {sourceTypeId, targetTypeId};
+	Oid			UDT_sourceBaseType;
+
+	/* Check if the UDT's base type is nvarchar or varbinary. If so, use the immediate base type for further processing. */
+	UDT_sourceBaseType = get_immediate_base_type_of_UDT_internal(sourceTypeId);
+	if(UDT_sourceBaseType != InvalidOid && ((*common_utility_plugin_ptr->is_tsql_nvarchar_datatype)(UDT_sourceBaseType) || (*common_utility_plugin_ptr->is_tsql_sys_varbinary_datatype)(UDT_sourceBaseType)))
+	{
+		typeIds[0] = UDT_sourceBaseType;
+		sourceTypeId = UDT_sourceBaseType;
```
tanscorpio7: why are we setting `sourceTypeId` here ?

contrib/babelfishpg_tsql/src/pltsql_coerce.c
```diff
@@ -475,8 +478,21 @@ tsql_find_coercion_pathway(Oid sourceTypeId, Oid targetTypeId, CoercionContext c
...
+	bool		isVarbinary = false;
+	bool		isNvarchar = false;
+	bool		isNvarchartoVarbinary = false;
 
 	Oid			typeIds[2] = {sourceTypeId, targetTypeId};
+	Oid			UDT_sourceBaseType;
+
+	/* Check if the UDT's base type is nvarchar or varbinary. If so, use the immediate base type for further processing. */
+	UDT_sourceBaseType = get_immediate_base_type_of_UDT_internal(sourceTypeId);
+	if(UDT_sourceBaseType != InvalidOid && ((*common_utility_plugin_ptr->is_tsql_nvarchar_datatype)(UDT_sourceBaseType) || (*common_utility_plugin_ptr->is_tsql_sys_varbinary_datatype)(UDT_sourceBaseType)))
```
tanscorpio7: please make this multi line, too many chars on one line.

contrib/babelfishpg_tsql/src/pltsql_coerce.c
```diff
@@ -507,22 +523,44 @@ tsql_find_coercion_pathway(Oid sourceTypeId, Oid targetTypeId, CoercionContext c
...
+			if (strcmp(type_nsname, "sys") == 0 && strcmp(type_name, "varbinary") == 0)
+				isVarbinary = true;
+
+			/* We've found NVARCHAR TO VARBINARY casting */
+			if (isNvarchar && strcmp(type_nsname, "sys") == 0 && (strcmp(type_name, "varbinary") == 0))
+				isNvarchartoVarbinary = true;
+
+			/* Check if type is NVARCHAR */
+			if (strcmp(type_nsname, "sys") == 0 && strcmp(type_name, "nvarchar") == 0)
+				isNvarchar = true;
```
tanscorpio7: We do not need to do it this way. We can just check if typeIds[0] = varbinary and typeIds[1] = nvarchar in single if.

tanscorpio7 (changes requested):

contrib/babelfishpg_tsql/src/pltsql_coerce.c
pranavJ23: because if we have a UDT then we need the UDT immediate base type as the sourceType, since we set key.castsource as sourcetype down there

shah-nirmit:
instead of naming test `cast_nvarchar_test` can we name it something like `cast_to_varbinary` so we can add other tests as well like nchar etc.?

contrib/babelfishpg_common/sql/upgrades/babelfishpg_common--3.8.0--4.0.0.sql
pranavJ23: Basically we need an upgrade script from 3.8 to 4, it was not there
So if someone wants to upgrade from 3.8 to 4 then without this upgrade script it won’t be possible, right ?

contrib/babelfishpg_common/sql/upgrades/babelfish_common_helper--4.4.0--4.5.0.sql
pranavJ23: for varbinaryvarchar function too we had a function varbinarysysvarchar, I created naming likewise

contrib/babelfishpg_common/sql/upgrades/babelfish_common_helper--4.4.0--4.5.0.sql
```diff
@@ -7,6 +7,38 @@
 
 SELECT set_config('search_path', 'sys, '||current_setting('search_path'), false);
 
+CREATE OR REPLACE FUNCTION sys.nvarcharvarbinary(sys.NVARCHAR, integer, boolean)
+RETURNS sys.BBF_VARBINARY
+AS 'babelfishpg_common', 'nvarcharvarbinary'
+LANGUAGE C IMMUTABLE STRICT PARALLEL SAFE;
+
+CREATE CAST (sys.NVARCHAR AS sys.BBF_VARBINARY)
+WITH FUNCTION sys.nvarcharvarbinary (sys.NVARCHAR, integer, boolean) AS ASSIGNMENT;
```
Deepesh125: Lets have a IF EXIST clause to check if cast already exists or not. And create only if it does not exist. Please check existing upgrade script to find example with similar usage.

contrib/babelfishpg_common/src/varbinary.c
```diff
@@ -874,6 +1036,52 @@ varcharbinary(PG_FUNCTION_ARGS)
...
+	if (!isExplicit)
+		ereport(ERROR,
+				(errcode(ERRCODE_DATATYPE_MISMATCH),
+				 errmsg("Implicit conversion from data type nvarchar to "
+						"binary is not allowed. Use the CONVERT function "
+						"to run this query.")));
+
+	initStringInfo(&buf);
+	TsqlUTF8toUTF16StringInfo(&buf, data, len);
+	data = buf.data;
```
Deepesh125, line 1057: do we need to free data first?

contrib/babelfishpg_common/src/varbinary.c
```diff
@@ -874,6 +1036,52 @@ varcharbinary(PG_FUNCTION_ARGS)
...
+	len= buf.len;
+
+	/* If typmod is -1 (or invalid), use the actual length */
+	if (typmod < (int32) VARHDRSZ)
```
Deepesh125, line 1061: what is default typmod for varbinary and binary? or what is tsql expected default typmod for varbinary?

test/JDBC/expected/cast_nvarchar_test-vu-verify.out
```diff
@@ -0,0 +1,384 @@
...
+SELECT * FROM dbo.HashDemoView;
+GO
+~~START~~
+varbinary#!#varbinary#!#varbinary
+185F8DB32271FE25F561A6FC938B2E264306EC304EDA518007D1764826381969#!#A07E4F7343246C82B26F32E56F85418D518D8B2F2DAE77F1D56FE7AF50DB97AF#!#185F8DB32271FE25F561A6FC938B2E264306EC304EDA518007D1764826381969
+~~END~~
+
+
+--TEST CASE 23: BINARY TO NVARCHAR
+select cast(cast(N'test string' as binary) as nvarchar)
```
Deepesh125, line 289: we should also have a test for default typmod where input string would be truncated

contrib/babelfishpg_common/src/varbinary.c
```diff
@@ -874,6 +1036,52 @@ varcharbinary(PG_FUNCTION_ARGS)
...
+		len = maxlen;
+
+	result = (bytea *) palloc(maxlen + VARHDRSZ);
+	SET_VARSIZE(result, maxlen + VARHDRSZ);
+
+	rp = VARDATA(result);
+	memcpy(rp, data, len);
+
+	/* NULL pad the rest of the space */
+	memset(rp + len, '\0', maxlen - len);
```
Deepesh125: rather directly use palloc0 at line 1073

contrib/babelfishpg_common/src/varbinary.c
```diff
@@ -874,6 +1036,52 @@ varcharbinary(PG_FUNCTION_ARGS)
 	PG_RETURN_BYTEA_P(result);
 }
 
+Datum
+nvarcharbinary(PG_FUNCTION_ARGS)
+{
+	VarChar    *source = PG_GETARG_VARCHAR_PP(0);
+	char	   *data = VARDATA_ANY(source);
+	char	   *rp;
+	size_t		len = VARSIZE_ANY_EXHDR(source);
```
Deepesh125, line 1041: max value of size_t is 65535 but input length can be more than that if we have nvarchar(max). Have we considered that case?

contrib/babelfishpg_common/src/varchar.c
```diff
@@ -176,6 +176,178 @@ GetUTF8CodePoint(const unsigned char *in, int len, int *consumed_p)
...
+	temp16.value = 0xD800 + (((code - 0x010000) >> 10) & 0x03FF);
+	appendStringInfoChar(buf, temp16.half[0]);
+	appendStringInfoChar(buf, temp16.half[1]);
+	temp16.value = 0xDC00 + ((code - 0x010000) & 0x03FF);
+	appendStringInfoChar(buf, temp16.half[0]);
+	appendStringInfoChar(buf, temp16.half[1]);
+}
+
+void
+TsqlUTF8toUTF16StringInfo(StringInfo out, const void *vin, size_t len)
```
Deepesh125: we already have functionality, TdsUTF8toUTF16StringInfo? is this new function different from it?  if not then we should refactor the code and use the common code everywhere for better maintainability

contrib/babelfishpg_common/src/varchar.c
```diff
@@ -176,6 +176,178 @@ GetUTF8CodePoint(const unsigned char *in, int len, int *consumed_p)
...
+	for (i = 0; i < len;)
+	{
+		code = GetUTF8CodePoint(&in[i], len - i, &consumed);
+		AddUTF16ToStringInfo(code, out);
+		i += consumed;
+	}
+}
+
+void
+TsqlUTF16toUTF8StringInfo(StringInfo out, void *vin, int len)
```
Deepesh125: same as above

contrib/babelfishpg_common/sql/binary.sql
```diff
@@ -99,6 +106,14 @@ LANGUAGE C IMMUTABLE STRICT PARALLEL SAFE;
 CREATE CAST (sys.BBF_BINARY AS sys.VARCHAR)
 WITH FUNCTION sys.binarysysvarchar (sys.BBF_BINARY, integer, boolean) AS IMPLICIT;
 
+CREATE OR REPLACE FUNCTION sys.binarysysnvarchar(sys.BBF_BINARY, integer, boolean)
+RETURNS sys.NVARCHAR
+AS 'babelfishpg_common', 'varbinarynvarchar'
```
Deepesh125, line 109: we should follow consistent naming convention everywhere. Look for already existed function and follow the same naming rules

contrib/babelfishpg_common/src/varbinary.c
```diff
@@ -835,6 +919,84 @@ varbinaryvarchar(PG_FUNCTION_ARGS)
 	PG_RETURN_VARCHAR_P(result);
 }
 
+Datum
+varbinarynvarchar(PG_FUNCTION_ARGS)
+{
+	bytea	        *source = PG_GETARG_BYTEA_PP(0);
+	char	        *data = VARDATA_ANY(source);		/* Source data is UTF16 encoded */
```
Deepesh125: is this valid assumption? source data will always be UTF16?

contrib/babelfishpg_common/src/varbinary.c
```diff
@@ -835,6 +919,84 @@ varbinaryvarchar(PG_FUNCTION_ARGS)
...
+	maxlen = typmod - VARHDRSZ;
+
+    /*
+    * Converts UTF-16 to UTF-8, handling odd-length inputs by padding.
+    * Respects maxlen if specified, otherwise processes full input.
+    * Uses TsqlUTF16toUTF8StringInfo for conversion, with error handling via PG_TRY.
+    */
+
+    /* truncating NULL bytes from end */
+	while(len>0  && data[len-1] == '\0')
```
Deepesh125, line 948: if input data is always UTF16 then last byte wil always be 0x00, isnt it? in that case we will always end up executing `len%2 ==1` part unnecessary

contrib/babelfishpg_common/src/varbinary.c
```diff
@@ -835,6 +919,84 @@ varbinaryvarchar(PG_FUNCTION_ARGS)
...
+	/* Initialising the paddedData as data only */
+	paddedData = (char*)palloc(len + 1);
+	memcpy(paddedData, data, len);
+
+	/* Do the Padding if lenngth is odd */
+	if(len % 2 != 0)
+	{
+		memset(paddedData + len, '\0', 1);
+		len = len + 1;
+	}
```
Deepesh125: Please refactor this code a bit. its really hard to read this code at current state.

contrib/babelfishpg_common/src/varbinary.c
```diff
@@ -835,6 +919,84 @@ varbinaryvarchar(PG_FUNCTION_ARGS)
...
+
+	if(!(maxlen < 0 || (len >> 1) <= maxlen))
+	{
+		len = maxlen << 1;
+	}
```
Deepesh125, line 959: plz add a comment explaining this logic.

contrib/babelfishpg_common/src/varbinary.c
```diff
@@ -835,6 +919,84 @@ varbinaryvarchar(PG_FUNCTION_ARGS)
...
+
+		ectx = MemoryContextSwitchTo(ccxt);
+		errorData = CopyErrorData();
+		FlushErrorState();
+		MemoryContextSwitchTo(ectx);
+
+		ereport(ERROR,
+				(errcode(ERRCODE_INTERNAL_ERROR),
+				errmsg("Failed to convert from data type varbinary to nvarchar, %s",
+				errorData->message)));
```
Deepesh125, line 984: plz add test case for this

contrib/babelfishpg_common/src/varchar.c
pranavJ23: We can simply use TsqlUTF8toUTF16StringInfo also

contrib/babelfishpg_common/src/varbinary.c
```diff
@@ -700,6 +707,82 @@ varcharvarbinary(PG_FUNCTION_ARGS)
...
+
+	if (!isExplicit)
+		ereport(ERROR,
+				(errcode(ERRCODE_DATATYPE_MISMATCH),
+				 errmsg("Implicit conversion from data type nvarchar to "
+						"varbinary is not allowed. Use the CONVERT function "
+						"to run this query.")));
+
+	initStringInfo(&buf);
+	PG_TRY();
```
Deepesh125, line 759: seems like this is common code for nvarchar -> varbinary and nvarchar -> binary. can we consider having helper function for this?

Deepesh125 (changes requested):

contrib/babelfishpg_tsql/src/pltsql_coerce.c
```diff
@@ -475,8 +492,31 @@ tsql_find_coercion_pathway(Oid sourceTypeId, Oid targetTypeId, CoercionContext c
...
 	Oid			typeIds[2] = {sourceTypeId, targetTypeId};
+	Oid			UDT_sourceBaseType;
+	Oid			UDT_targetBaseType;
+
+
+	/* Check if the UDT's base type is nvarchar or varbinary.
+	 * If so, use the immediate base type for further processing.
+	 */
+	UDT_sourceBaseType = get_immediate_base_type_of_UDT_internal(sourceTypeId);
+	UDT_targetBaseType = get_immediate_base_type_of_UDT_internal(targetTypeId);
```
Deepesh125: Have we considered performance impact of this change? on regular casts?

Deepesh125 (changes requested):
We should raise separate PR just for changes related to hashbytes

contrib/babelfishpg_common/src/varbinary.c
pranavJ23: Can have a seperate function, but it's same implementation. So do we make another function ?

contrib/babelfishpg_common/src/varbinary.c
pranavJ23: That comment is making confusion, will remove that comment
Not necessarily we will have UTF16 encoded string, So hence we need to include the padding to handle that

contrib/babelfishpg_common/sql/binary.sql
pranavJ23: Followed the naming convention as per 'binarysysvarchar'. So to maintain consistency I made similar function as binarysysnvarchar.

contrib/babelfishpg_common/src/collation.h
```diff
@@ -119,6 +119,9 @@ typedef struct collation_callbacks
 
 	void		(*set_db_collation) (Oid db_coll);
 
+	void		(*TsqlUTF8toUTF16StringInfo) (StringInfo utf16_data, const void *data, size_t len);	
+	void		(*TsqlUTF16toUTF8StringInfo) (StringInfo out, void *vin, int len);
```
Deepesh125: In collation.c:30, we need to initialise these two fields

contrib/babelfishpg_tsql/sql/datatype_string_operators.sql
```diff
@@ -1,10 +1,10 @@
 CREATE OR REPLACE FUNCTION sys.hashbytes(IN alg VARCHAR, IN data VARCHAR) RETURNS sys.bbf_varbinary
 AS 'babelfishpg_tsql', 'hashbytes' LANGUAGE C IMMUTABLE STRICT PARALLEL SAFE;
-GRANT EXECUTE ON FUNCTION sys.hashbytes(IN VARCHAR, IN VARCHAR) TO PUBLIC;
+GRANT EXECUTE ON FUNCTION sys.hashbytes(IN alg VARCHAR, IN VARCHAR) TO PUBLIC;
 
 CREATE OR REPLACE FUNCTION sys.hashbytes(IN alg VARCHAR, IN data sys.bbf_varbinary) RETURNS sys.bbf_varbinary
 AS 'babelfishpg_tsql', 'hashbytes' LANGUAGE C IMMUTABLE STRICT PARALLEL SAFE;
-GRANT EXECUTE ON FUNCTION sys.hashbytes(IN VARCHAR, IN sys.bbf_varbinary) TO PUBLIC;
+GRANT EXECUTE ON FUNCTION sys.hashbytes(IN alg VARCHAR, IN sys.bbf_varbinary) TO PUBLIC;
```
Deepesh125: is this change really needed?

contrib/babelfishpg_tds/src/backend/tds/tdsxact.c
```diff
@@ -97,7 +97,7 @@ GetTxnName(const StringInfo message, TDSRequestTxnMgmt request, int offset)
 							len, TSQL_TXN_NAME_LIMIT)));
 
 		initStringInfo(&(request->txnName));
-		TdsUTF16toUTF8StringInfo(&(request->txnName),
+		(collation_callbacks_ptr->TsqlUTF16toUTF8StringInfo)(&(request->txnName),
```
Deepesh125: or we can directly place collation_callbacks_ptr->TsqlUTF16toUTF8StringInfo inside TdsUTF16toUTF8StringInfo. does it make sense?

contrib/babelfishpg_common/src/babelfishpg_common.c
```diff
@@ -187,6 +187,7 @@ get_common_utility_plugin(void)
 		common_utility_plugin_var.is_tsql_datetime2_datatype = &is_tsql_datetime2_datatype;
 		common_utility_plugin_var.is_tsql_smalldatetime_datatype = &is_tsql_smalldatetime_datatype;
 		common_utility_plugin_var.is_tsql_datetimeoffset_datatype = &is_tsql_datetimeoffset_datatype;
+		common_utility_plugin_var.TsqlUTF8toUTF16StringInfo = &TsqlUTF8toUTF16StringInfo;
```
Deepesh125: where is this needed?

contrib/babelfishpg_common/src/varbinary.c
```diff
@@ -700,6 +707,82 @@ varcharvarbinary(PG_FUNCTION_ARGS)
 	}
 	PG_END_TRY();
 
+	/* 
+	 * If typmod is -1 (or invalid), use the actual length
+	 * Length should be checked after encoding into server encoding
+	 */
+	if (typmod < (int32) VARHDRSZ)
```
Deepesh125: in which case typmod < VARHDRSZ? there could be only undefined typmod where typmod will be -1 so can we directly have the check where typmod == -1?

contrib/babelfishpg_common/src/varbinary.c
```diff
@@ -700,6 +707,82 @@ varcharvarbinary(PG_FUNCTION_ARGS)
...
+		maxlen = encodedByteLen;
+	else
+		maxlen = typmod - VARHDRSZ;
+
+	if (encodedByteLen > maxlen)
+		encodedByteLen = maxlen;
+	result = (bytea *) palloc(encodedByteLen + VARHDRSZ);
```
Deepesh125: palloc0

contrib/babelfishpg_common/src/varbinary.c
```diff
@@ -835,6 +919,84 @@ varbinaryvarchar(PG_FUNCTION_ARGS)
...
+		len -= 1;
+
+	/* Initialising the paddedData as data only */
+	paddedData = (char*)palloc(len + 1);
+	memcpy(paddedData, data, len);
+
+	/* Do the Padding if lenngth is odd */
+	if(len % 2 != 0)
+	{
+		memset(paddedData + len, '\0', 1);
```
Deepesh125: we only need palloc len + 1 if there is odd len. So we can do something like,
```
if (len %2 == 1) then paddedData = (char*) palloc0(len + 1);
memcpy(paddedData, data, len);
```

contrib/babelfishpg_common/src/varbinary.c
Deepesh125: what is len >> 1 and maxlen << 1? is it due to *2 for UTF16?

Deepesh125 (changes requested):

contrib/babelfishpg_common/src/varbinary.c
Deepesh125: yes, another static inline function

Deepesh125 (changes requested):

contrib/babelfishpg_common/src/babelfishpg_common.c
pranavJ23: Now not needed,since we removed the hashbytes implementation. Removing it

contrib/babelfishpg_common/src/varbinary.c
pranavJ23: I think we already have 2 different functions nvarcharbinary and nvarcharbinary

contrib/babelfishpg_common/src/varbinary.c
pranavJ23: yes

contrib/babelfishpg_common/src/varbinary.c
pranavJ23: So size_t is of type long unsigned int
and nvarchar max is 4000 characters only
So nvarchar max can be accommodate with size_t
So I generated a 4000 character string and tried to convert it to varbinary
it worked fine

contrib/babelfishpg_common/sql/upgrades/babelfish_common_helper--4.4.0--4.5.0.sql
pranavJ23: Removed the cast since it was redundant

contrib/babelfishpg_common/src/varbinary.c
pranavJ23: Added a test case for default typmod too

contrib/babelfishpg_common/src/varbinary.c
Deepesh125: This is unresolved

contrib/babelfishpg_common/src/varbinary.c
```diff
@@ -835,6 +919,84 @@ varbinaryvarchar(PG_FUNCTION_ARGS)
...
+	size_t		len = VARSIZE_ANY_EXHDR(source);
+	int32		typmod = -1;
+	int		maxlen = -1;
+	int		encodedByteLen;
+	StringInfoData 	buf;
+	char 		*paddedData = (char*)palloc(len+1);
+	MemoryContext   ccxt = CurrentMemoryContext;
+
+	typmod = PG_GETARG_INT32(1);
+	maxlen = typmod - VARHDRSZ;
```
Deepesh125, line 939: what will be typmod in case of `select cast(<> as nvarchar(max))`?

contrib/babelfishpg_common/src/varbinary.c
Deepesh125: please point the code line

Deepesh125 (changes requested):

contrib/babelfishpg_common/src/varbinary.c
pranavJ23: https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/3072/files#diff-fff75a029d373c427177cfcc5166d287721a02b78cbf1aeae54b868ad6d7bc9eR216

contrib/babelfishpg_common/src/varbinary.c
pranavJ23: I added the palloc0 in nvarcharvarbinary, varcharvarbinary is a pre-existing function. Not my function

contrib/babelfishpg_common/src/varbinary.c
```diff
@@ -721,6 +727,84 @@ varcharvarbinary(PG_FUNCTION_ARGS)
...
+				errmsg("Failed to convert from data type nvarchar to varbinary, %s",
+				errorData->message)));
+	}
+	PG_END_TRY();
+
+	/* 
+	 * If typmod is -1 (or invalid), use the actual length
+	 * Length should be checked after encoding into server encoding
+	 */
+	if (typmod < 0)
```
tanscorpio7: This should be `if (typmod < (int32) VARHDRSZ)`

contrib/babelfishpg_common/src/varbinary.c
```diff
@@ -835,6 +919,84 @@ varbinaryvarchar(PG_FUNCTION_ARGS)
...
+	typmod = PG_GETARG_INT32(1);
+	maxlen = typmod - VARHDRSZ;
+
+    /*
+    * Converts UTF-16 to UTF-8, handling odd-length inputs by padding.
+    * Respects maxlen if specified, otherwise processes full input.
+    * Uses TsqlUTF16toUTF8StringInfo for conversion, with error handling via PG_TRY.
+    */
+
+    /* truncating NULL bytes from end */
```
tanscorpio7: indentation seems off

contrib/babelfishpg_common/src/varbinary.c
```diff
@@ -835,6 +919,84 @@ varbinaryvarchar(PG_FUNCTION_ARGS)
...
+
+	/* Do the Padding if lenngth is odd */
+	if(len % 2 != 0)
+	{
+		paddedData = (char*)palloc0(len+1);
+		memcpy(paddedData, data, len);
+		len = len + 1;
+	}
+	else
+		memcpy(paddedData, data, len);
```
tanscorpio7: We can just do
```
memcpy(paddedData, data, len);
len = len % 2 ? len : len + 1;
```
You are already doing palloc at line during declaration 
```
char 		*paddedData = (char*)palloc(len+1);
```
make it palloc0

contrib/babelfishpg_common/src/varbinary.c
```diff
@@ -835,6 +919,84 @@ varbinaryvarchar(PG_FUNCTION_ARGS)
...
+	if(!(maxlen < 0 || (len >> 1) <= maxlen))
+	{
+		len = maxlen << 1;
+	}
+
+	PG_TRY();
+	{
+		/* Converts UTF-16 to UTF-8 using TsqlUTF16toUTF8StringInfo */
+		initStringInfo(&buf);
+		TsqlUTF16toUTF8StringInfo(&buf,paddedData,len);
```
tanscorpio7: indent: spaces between args

contrib/babelfishpg_common/src/varbinary.c
```diff
@@ -874,6 +1036,49 @@ varcharbinary(PG_FUNCTION_ARGS)
...
+						"binary is not allowed. Use the CONVERT function "
+						"to run this query.")));
+
+	initStringInfo(&buf);
+	TsqlUTF8toUTF16StringInfo(&buf, data, len);
+	data = buf.data;
+	len= buf.len;
+
+	/* If typmod is -1 (or invalid), use the actual length */
+	if (typmod < 0)
```
tanscorpio7: should be < `VARHDRSZ`

tanscorpio7 (changes requested):

tanscorpio7 (approved):

contrib/babelfishpg_common/src/varchar.c
shah-nirmit: By anychance can we reuse TDS funcs using rendezvous variables

test/python/expected/upgrade_validation/expected_dependency.out
```diff
@@ -236,6 +236,7 @@ Function sys.binaryint2(sys.bbf_binary)
 Function sys.binaryint4(sys.bbf_binary)
 Function sys.binaryint8(sys.bbf_binary)
 Function sys.binaryrowversion(sys.bbf_binary,integer,boolean)
+Function sys.binarysysnvarchar(sys.bbf_binary,integer,boolean)
```
shah-nirmit: can we add dep tests?

shah-nirmit (approved):

rohit01010 (approved):

contrib/babelfishpg_common/src/varbinary.c
Deepesh125: why? as i made comment previously, we simply need typmod == -1

contrib/babelfishpg_common/src/varbinary.c
Deepesh125: how is this correct?
```
len = len % 2 ? len : len + 1;
```

it should be 

```
len = len % 2 ? len + 1 : len;
```

contrib/babelfishpg_common/src/varbinary.c
tanscorpio7: typmod < `VARHDRSZ` means we cannot even accommodate the header so we have to treat it as -1. We follow this in all the existing cast functions.

Deepesh125 (approved):

contrib/babelfishpg_common/src/varbinary.c
tanscorpio7: Even postgres cast functions check typmod against `VARHDRSZ`

PR #3073, opened 2024-11-05
Support for T-SQL variable/parameter names like @@var and @var#

robverschoor:
### Description

Valid names of user-defined T-SQL variables and/or parameters are `@@myvar`, `@myvar#`, `@#myvar`, etc. and indeed we see customer T-SQL applications using such names. 
Until now, such names have raised errors in the backend because we can only handle variable/parameter names with one leading `@` and without `#` characters.
This fix adds support by enclosing such variable names in square bracket delimiters; when using double quotes as delimiters (as is currently done in most places for regular-named variables), a query like `select @@V` will produce the string `'@@V'` instead of the value of variable `@@V`.
The fix is partly done through ANTLR rewriting and partly at execution time elsewhere in the `babelfish_tsql` codeline. Things are complicated by the fact that the ANTLR rewriting logic is very fragmented, as well as that for some variables reference contexts inside the body of a procedure or trigger, rewriting must be postponed until the body is executed. This means there are multiple code paths where these variable names must be intercepted. 
This fix also addresses variable names in some contexts (like a cursor variable) that previously could not be longer than 63 characters.
This fix also solves the previous issue that a variable could be declared with a name matching an internal `sys` function, but was resolved as the intern function instead of the variable, i.e. `DECLARE @@RAND INT=123 SELECT @@RAND` would return a random number instead of 123. 

### Issues Resolved

BABEL-2481 Parameter declarations containing `#` characters not handled correctly
BABEL-476  Support local variables/parameters with multiple `'@'` characters
BABEL-5384 User-defined `@@variable `should not be mapped to `sys` function 

- Use case based - Yes

- Boundary conditions - Yes

- Arbitrary inputs - N/A

- Negative test cases - Yes

- Performance tests - N/A

- Tooling impact - N/A

- Client tests - N/A

### Check List


contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -2493,6 +2513,149 @@ is_char_identpart(char c)
...
 
+/*
+ * Put delimiters around a T-SQL variable/parameter that is
+ * named '@@var' or contains a hash, e.g. '@var#'.
+ * Without delimiters, the backend will raise an error.
+ * This is used for the parameter argument of sp_executesql, so the input
+ * string may contain multiple names, e.g.: @par1 int, @par2 varchar(20), ...
+ */
+static void
+delimit_tsql_atatuservar(const char *src, char *result)
```
Sairakan: Why not have this function allocate and return the result itself, instead of requiring a pre-allocated result buffer, which imo has potential for errors if the caller doesn't properly allocate it?

robverschoor: What's the preferred way of doing it in Babelfish? There does not seem to be a consistently followed approach as far as I can tell, so I just took inspiration from some existing function.

Sairakan: Hm, I'm not sure if we have a standard for this in Babelfish. The decision of whether to accept a pre-allocated buffer or to allocate the memory within the function itself depends on the use case of the function. For example, sometimes the output just needs to be put into a temporary buffer within the caller, and isn't allocated on the heap at all. In those cases it can make sense to allow the caller to be responsible for defining/allocating the output buffer. 

In this case however, your use case so far is using palloc() anyways, so unless you have a strong idea of a future use case of this function that would _not_ be using a palloc() call, I think it's cleaner and safer to have this function allocate the memory itself.

In either case, the behavior with the output needs to be clearly documented in the function header comment. For the current case, that would be specifying that the `result` must have sufficient space to hold the expanded string, and if you change the function to return the result, you should specify that it is returned in a newly palloc'ed string.

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -2515,13 +2679,22 @@ read_param_def(InlineCodeBlockArgs *args, const char *paramdefstr)
 		return;
 	}
 
+	/* 
+	 * Put delimiters around parameter names like @@par or @par#.
+	 * Reserving twice the amount of space: since the shortest possible parameter
+	 * definition is 5 characters ('@@p x' , where x would be the type), this will 
+	 * always be enough for adding delimiters.
+	 */	
+	paramdefstr_delimited = (char *) palloc(sizeof(char *)*strlen(paramdefstr)*2);
```
Sairakan: `sizeof(char *)` is probably not correct here - I think you probably meant just `sizeof(char)`. Also, why is it not possible to just allocate `strlen(paramdefstr)+1+(2*<size_of_delimiter_string>)`?

Sairakan: Scratch that, I didn't realize that multiple parameters can be contained within this string. My previous point about using `sizeof(char)` still stands though

robverschoor: On 1: you're right, will fix.
On 2: There can be N parameters in the string that need to be delimited so that's N*2 additional chars. We should not make assumptions about the value of N, so this seems the formula that is safest + simplest.

contrib/babelfishpg_tsql/src/pl_exec-2.c
robverschoor: Changed to return the modified string.

contrib/babelfishpg_tsql/src/pl_exec-2.c
robverschoor: fixed

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -2493,6 +2513,157 @@ is_char_identpart(char c)
 			(c >= '0' && c <= '9'));
 }
 
+/*
+ * Put delimiters around a T-SQL variable/parameter that is
+ * named '@@var' or contains a hash, e.g. '@var#'.
+ * Without delimiters, the backend will raise an error.
+ * This is used for the parameter argument of sp_executesql, so the input
+ * string may contain multiple names, e.g.: @par1 int, @par2 varchar(20), ...
+ */
```
Sairakan, line 2544: I think that this header should be updated to clarify that this function returns a newly palloc'ed string.

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -2493,6 +2513,157 @@ is_char_identpart(char c)
...
+	char *varname_start = NULL;
+	bool add_delimiter = false;
+	
+	/* 
+	 * Reserving twice the amount of space of the input string: since the shortest possible
+	 * parameter definition is 5 characters ('@@p x' , where x would be the type), this will 
+	 * always be enough for adding delimiters.
+	 * Note that there can be multiple parameter names in the input string.
+	 */
+	char *result = (char *) palloc(sizeof(char)*strlen(src)*2);
```
Sairakan, line 2558: One option rather than pre-allocating the max possible string is to just use a StringInfo struct, which automatically handles string resizing. See https://github.com/babelfish-for-postgresql/babelfish_extensions/blob/BABEL_4_X_DEV/contrib/babelfishpg_tsql/src/pl_exec-2.c#L281 for examples on how to use such a struct (in this case you don't need to free the data afterwards, and can just return it)

robverschoor: We're talking about pretty small strings, a long string would be 100 bytes but often we see something like 20-30 chars. For such small sizes, it does not seem to be worth it.

Sairakan: If you are confident that the string size overhead is small, then I am okay with going with this current implementation.

robverschoor: I am, thanks

contrib/babelfishpg_tsql/src/pl_exec-2.c
robverschoor: added comment

contrib/babelfishpg_tsql/src/tsqlIface.cpp
```diff
@@ -1951,8 +1960,13 @@ class tsqlBuilder : public tsqlCommonMutator
 					throw PGErrorWrapperException(ERROR, ERRCODE_INVALID_FUNCTION_DEFINITION, "'INSERT EXEC' cannot be used within a function", getLineAndPos(ddl_object));
 				}
 				else if (ddl_object && !ddl_object->local_id()) /* insert into non-local object */
-				{
-					throw PGErrorWrapperException(ERROR, ERRCODE_INVALID_FUNCTION_DEFINITION, "'INSERT' cannot be used within a function", getLineAndPos(ddl_object));
+				{					
+					if (ddl_object && ddl_object->full_object_name() && isDelimitedAtAtUserVarName(getFullText(ddl_object->full_object_name()) ))
```
Sairakan, line 1964: I think that it's a little confusing to have an if-check which results in a no-op. Maybe you can flip this around? So something like 
```
if ( !(ddl_object && ddl_object->full_object_name() && isDelimitedAtAtUserVarName(getFullText(ddl_object->full_object_name()) )
    throw <...>
```

robverschoor: The reason I put it in here, with the comment, is to clarify the code structure. Making it a negative condition makes it harder to understand IMHO.

Sairakan: On second thought, this is still relatively readable code as-is. So probably no change is needed.

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -2493,6 +2513,159 @@ is_char_identpart(char c)
...
+			{
+				add_delimiter = true;
+			}
+		}
+
+		/* Find end of variable name */
+		while (*s)
+		{
+			/* Allowed chars in @variable name: */
+			if (isalpha(*s) || isdigit(*s) || (*s == '_') || (*s == '@') || (*s == '#'))
```
timchang514: what might happen here for names in different languages/collations, ie non latin chars?

contrib/babelfishpg_tsql/src/tsqlIface.cpp
```diff
@@ -489,10 +496,11 @@ attachPLtsql_fragment(ParseTree *node, PLtsql_stmt *fragment)
 {
 	if (fragment)
 	{
-		const char *tsqlDesc = pltsql_stmt_typename(fragment);
-
 		if (pltsql_enable_antlr_detailed_log)
+		{
+			const char *tsqlDesc = pltsql_stmt_typename(fragment);
```
timchang514, line 501: this change is just for scope cleanup, right?

robverschoor: this line was previously above the test, so it was always executed, also when the detailed logging was not enabled. that made no sense.

contrib/babelfishpg_tsql/src/pl_exec-2.c
Deepesh125: we should better use [is_identifier_char()](https://github.com/babelfish-for-postgresql/babelfish_extensions/blob/f29e512f0f8ec1e30cc310cb95980ed6025be151/contrib/babelfishpg_tsql/src/hooks.c#L1949) if that helps here

robverschoor: `is_identifier_char()` does not check for `@` characters (though it probably should since these are valid in T-SQL). The test above has been relocated into a new function `is_variable_name_char()` . Both these functions do actually not test for letters in non-ASCII ranges (e.g. Chinese etc.). It would seem to make sense to unify these two functions and to add support for beyond-ASCII characters as well.  However this exceeds the scope of the current fix and should be dealt with separately.

For example, the following is valid in T-SQL but currently fails in Babefish because the Chinese characters are not recognized by `is_variable_name_char()` .
```
1> -- variable name with two '@@' 
2> declare @@地点 int = 124 
3> select @@地点 
4> go 
Msg 33557097, Level 16, State 1, Server BABELFISH, Line 3 
function sys.地点() does not exist 
```

Note that regular identifiers (e.g. table names) and variables like the above example but with only one `@`, are currently handled correctly by Babelfish because these names are just handed off to the PG backend where the backend lexer handles these, and the lexer is Unicode-aware. For cases where Babelfish needs to explicitly check the characters in the name through `is_identifier_char()`  or `is_variable_name_char()` , this will currently result in an error for such names.

timchang514: thanks for looking into this Rob. Agree that this exceeds the scope of this change.

test/python/expected/sql_validation_framework/expected_create.out
```diff
@@ -114,9 +114,6 @@ Could not find upgrade tests for function sys.dateadd_internal
 Could not find upgrade tests for function sys.datediff_internal
 Could not find upgrade tests for function sys.datepart_internal
 Could not find upgrade tests for function sys.default_domain
-Could not find upgrade tests for function sys.error_procedure
```
timchang514, line 117: where is this diff coming from? doesn't seem related to this change

test/JDBC/expected/BABEL-3092.out
```diff
@@ -356,7 +356,7 @@ EXEC("SELECT * FROM t")
 GO
 ~~ERROR (Code: 33557097)~~
```
timchang514, line 358: where is this diff coming from? doesn't seem related to this change

test/python/expected/sql_validation_framework/expected_create.out
timchang514, line 117: where is this diff coming from? doesn't seem related to this change

timchang514 (approved):

test/python/expected/sql_validation_framework/expected_create.out
robverschoor: There's a test case with a variable named @@error_procedure

shardgupta (approved):

PR #3095, opened 2024-11-12
Display unmapped schema names in error messages

chxwang:
### Description

Previously, Babelfish displays the mapped schema names in error messages. This commit implements a hook function remove_db_name_in_schema and adds test cases. Original unmapped name will be displayed with this commit.

### Issues Resolved

[BABEL-2961](https://jira.rds.a2z.com/browse/BABEL-2961)

* **Use case based -** :white_check_mark:

* **Arbitrary inputs -** NA

* **Performance tests -** NA

* **Tooling impact -** NA

* **Client tests -** NA

### Check List


contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -5630,3 +5636,23 @@ pltsql_get_object_identity_event_trigger(ObjectAddress* address)
...
+
+/*
+ * remove_db_name_in_schema - remove the db name and underscore at the beginning
+ * 	of the given string. It is used to unmap schema name in error messages.
+ *
+ * 	@param schema_name - char *
+ * 	@return - unmapped schema name char *
+ */
+static const char *
+remove_db_name_in_schema(const char *schema_name)
```
xhfanhe: Should compare BBF & SQL Server error messages of cross-db queries. Do schema names in cross-db queries need to be mapped as well?

chxwang: No, this fucntion works only if the input `schema_name` has current db name as prefix. So if I query another db, nothing will change in the error message

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -5864,3 +5870,44 @@ is_bbf_db_ddladmin_operation(Oid namespaceId)
...
+	char		**splited_object_name;
+	char		*schema_name = NULL;
+	char		*mutable_name;
+    size_t 		db_name_len;
+    size_t 		prefix_len;
+	size_t 		schema_name_len;
+
+	mutable_name = strdup(object_name);
+	splited_object_name = split_object_name(mutable_name);
+	for (int i = 0; i < 4; i++)
```
RIC06X: is it guaranteed to have 4 element in that list? can you check the length of that list first?

RIC06X:
can you also add you test BABEL-2961 to `test/JDBC/upgrade/latest/schedule`? This is for the upgrade dependency test

contrib/babelfishpg_tsql/src/hooks.c
chxwang: Yes, the function returns 4 element list

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -5864,3 +5870,44 @@ is_bbf_db_ddladmin_operation(Oid namespaceId)
...
+	cur_db_name = get_cur_db_name();
+	db_name_len = strlen(cur_db_name);
+	schema_name_len = schema_name ? strlen(schema_name) : 0;
+	prefix_len = db_name_len + 1;
+
+	if (schema_name != NULL && schema_name_len > db_name_len && strncmp(schema_name, cur_db_name, db_name_len) == 0 && schema_name[db_name_len] == '_') {
+		object_name += prefix_len;
+	}
+	
+	free(mutable_name);
```
xhfanhe: Please use pfree here

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -5864,3 +5870,44 @@ is_bbf_db_ddladmin_operation(Oid namespaceId)
...
+{
+	const char	*cur_db_name;
+	char		**splited_object_name;
+	char		*schema_name = NULL;
+	char		*mutable_name;
+    size_t 		db_name_len;
+    size_t 		prefix_len;
+	size_t 		schema_name_len;
+
+	mutable_name = strdup(object_name);
```
timchang514: let's use `pstrdup` instead of `strdup`

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -5864,3 +5870,44 @@ is_bbf_db_ddladmin_operation(Oid namespaceId)
...
+	splited_object_name = split_object_name(mutable_name);
+	for (int i = 0; i < 4; i++)
+	{
+		if (strcmp(splited_object_name[i], "")) {
```
timchang514: I think our code checker will complain about `strcmp` usage. I would suggest either `strncmp` or in this case, maybe you can just use a `strlen` call.

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -5864,3 +5870,44 @@ is_bbf_db_ddladmin_operation(Oid namespaceId)
...
+			schema_name = splited_object_name[i];
```
xhfanhe: This is simply taking the first not-null value in splited_object_name[] as schema name - is that the right assumption? What if the input name does not have schema, e.g. "func1" ?

contrib/babelfishpg_tsql/src/hooks.c
timchang514: same as above, replace with `pfree` once you switch to `pstrdup`.

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -5864,3 +5870,44 @@ is_bbf_db_ddladmin_operation(Oid namespaceId)
...
+			break;
+		}
+	}
+
+	cur_db_name = get_cur_db_name();
```
timchang514: i think this is a `pstrdup`'d string. needs to be `pfree`d.

contrib/babelfishpg_tsql/src/hooks.c
timchang514: can you double check whether the elements in `splited_object_name` need to be freed

contrib/babelfishpg_tsql/src/hooks.c
chxwang: Yes I assumed the first non-null value in splited_object_name[] is the schema name, and check it later
If no schema in input name, it will not contain the "dbname_" prefix and will not be changed
The prefix can only exist in the first value in splited_object_name[], so I want to check the first value here

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -5864,3 +5870,48 @@ is_bbf_db_ddladmin_operation(Oid namespaceId)
...
+	char		*mutable_name;
+    size_t		db_name_len;
+    size_t		prefix_len;
+	size_t		schema_name_len;
+
+	mutable_name = pstrdup(object_name);
+	splited_object_name = split_object_name(mutable_name);
+
+	// If there are more than two parts, then it’s a cross-db object name (db.schema.object) so we don’t need to deal with it
+	if (strlen(splited_object_name[1]) == 0)
```
RIC06X: it is better to check the list length > 1 and and make sure splited_object_name[1] is not NULL;
Updates: Checked function splited_object_name, this function will guarantee to give a list of 4 not NULL elements

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -5864,3 +5870,48 @@ is_bbf_db_ddladmin_operation(Oid namespaceId)
...
+	char		*schema_name = NULL;
+	char		*mutable_name;
+    size_t		db_name_len;
+    size_t		prefix_len;
+	size_t		schema_name_len;
+
+	mutable_name = pstrdup(object_name);
+	splited_object_name = split_object_name(mutable_name);
+
+	// If there are more than two parts, then it’s a cross-db object name (db.schema.object) so we don’t need to deal with it
```
RIC06X: Please update all comment style to `/* comments */`

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -5864,3 +5870,48 @@ is_bbf_db_ddladmin_operation(Oid namespaceId)
...
+{
+	char		*cur_db_name;
+	char		**splited_object_name;
+	char		*schema_name = NULL;
+	char		*mutable_name;
+    size_t		db_name_len;
+    size_t		prefix_len;
+	size_t		schema_name_len;
+
+	mutable_name = pstrdup(object_name);
```
RIC06X, line 5893: is it necessary to copy this string? If not, we can remove it. I think  `split_object_name` will not modify the original string.

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -5864,3 +5870,48 @@ is_bbf_db_ddladmin_operation(Oid namespaceId)
...
+	{
+		schema_name = strlen(splited_object_name[2]) == 0 ? splited_object_name[3] : splited_object_name[2];
+		cur_db_name = get_cur_db_name();
+		db_name_len = strlen(cur_db_name);
+		schema_name_len = strlen(schema_name);
+		prefix_len = db_name_len + 1;
+
+		if (schema_name != NULL && schema_name_len > db_name_len && strncmp(schema_name, cur_db_name, db_name_len) == 0 && schema_name[db_name_len] == '_') {
+			// Return the part after the prefix
+			object_name += prefix_len;
```
RIC06X: Instead of modifying the original string, I think it is better to return a new string. Here we can return the copy of source string starting from prefix_len

RIC06X:
LGTM

xhfanhe (approved):
LGTM

PR #3106, opened 2024-11-12
Fixed system stored procedures, sp_columns and sp_columns_100 to handle input @fusepattern = 0 correctly

Deepesh125:
### Description

Previously, the parameter value @fUsePattern=0 is broken for sp_columns_100, returning only a partial data set. Also,
overall handling of identifier was incorrect when it comes to truncation of identifiers. This commit aims to fix these issue
by appropriately using truncate function as well as fixes issue of @fUsePattern=0 which now should return proper data.

Task: BABEL-5314

### Check List


rohit01010 (approved):

contrib/babelfishpg_tsql/sql/babelfishpg_tsql.sql
```diff
@@ -459,6 +464,9 @@ CREATE OR REPLACE PROCEDURE sys.sp_columns_100 (
     "@fusepattern" smallint = 1)
 AS $$
 BEGIN
+	-- TODO: we should be able to get rid of babelfish_truncate_identifier when we fix BABEL-5416
+	declare @truncated_ident sys.nvarchar(384);
+	select @truncated_ident = sys.babelfish_truncate_identifier(pg_catalog.lower(@table_name));
```
sumitj824, line 508: Why are we downcasing the table name explicitly and isn't the behaviour should be based on server collation? I think the function `sys.babelfish_truncate_identifier` itself the downcase and truncate the table name if required. Please Correct me if I'm wrong.

contrib/babelfishpg_tsql/sql/babelfishpg_tsql.sql
```diff
@@ -299,13 +299,18 @@ BEGIN
...
 CREATE OR REPLACE VIEW sys.sp_columns_100_view AS
   SELECT 
   CAST(t4."TABLE_CATALOG" AS sys.sysname) AS TABLE_QUALIFIER,
   CAST(t4."TABLE_SCHEMA" AS sys.sysname) AS TABLE_OWNER,
-  CAST(t4."TABLE_NAME" AS sys.sysname) AS TABLE_NAME,
-  CAST(t4."COLUMN_NAME" AS sys.sysname) AS COLUMN_NAME,
+  CAST(
+    CASE WHEN t1.reloptions[1] LIKE 'bbf_original_rel_name%' THEN substring(t1.reloptions[1], 23)
+      ELSE t4."TABLE_NAME" END
+            AS sys.sysname) AS TABLE_NAME,
```
sumitj824: 1. This is with the assumption that `bbf_original_rel_name` will be first entry in reloptions which is not always true. Please refer the definition of `information_schema_tsql.columns_internal`.
2. Why do we need the original name here?

contrib/babelfishpg_tsql/sql/babelfishpg_tsql.sql
```diff
@@ -299,13 +299,18 @@ BEGIN
...
+  CAST(
+    CASE WHEN a.attoptions[1] LIKE 'bbf_original_name%' THEN substring(a.attoptions[1], 19)
+      ELSE t4."COLUMN_NAME" END
+            AS sys.sysname) AS COLUMN_NAME,
```
sumitj824: Same as TABLE_NAME.

contrib/babelfishpg_tsql/sql/babelfishpg_tsql.sql
```diff
@@ -459,6 +464,9 @@ CREATE OR REPLACE PROCEDURE sys.sp_columns_100 (
     "@fusepattern" smallint = 1)
 AS $$
 BEGIN
+	-- TODO: we should be able to get rid of babelfish_truncate_identifier when we fix BABEL-5416
```
sumitj824, line 506: I see the definition of procedure `sys.sp_columns_100` is copy of `sys.sp_columns`, should we update the definition of `sys.sp_columns` as well?

contrib/babelfishpg_tsql/sql/babelfishpg_tsql.sql
Deepesh125: sys.babelfish_truncate_identifier only truncates the input. So it is expected that we pass lower case input so that sys.babelfish_truncate_identifier does expected work.

contrib/babelfishpg_tsql/sql/babelfishpg_tsql.sql
Deepesh125: Done

contrib/babelfishpg_tsql/sql/babelfishpg_tsql.sql
Deepesh125: Updated

contrib/babelfishpg_tsql/sql/babelfishpg_tsql.sql
Deepesh125: Updated it to use bbf_original_rel_name more precisely. Original name might be needed for multiple use cases

shah-nirmit (approved):

shardgupta (approved):

PR #3122, opened 2024-11-14
Disallow the use of user-defined functions in column defaults for temp tables and table variables.

Sairakan:
### Description

In SQL Server, it is prohibited for users to use user-defined functions as column defaults in temp tables and table variables - the only exception is in the tempdb database, but this is not currently even supported in Babelfish, so for now we will impose the same restrictions as exist elsewhere in SQL Server.

### Issues Resolved

BABEL-4868

* **Use case based -**

* **Boundary conditions -**

* **Arbitrary inputs -**

* **Negative test cases -**

* **Performance tests -**

* **Tooling impact -**

* **Client tests -**

### Check List

### DIFF
--- a/contrib/babelfishpg_tsql/src/pl_handler.c
+++ b/contrib/babelfishpg_tsql/src/pl_handler.c
@@ -149,7 +149,8 @@ extern bool pltsql_check_guc_plan(CachedPlanSource *plansource);
 bool		pltsql_function_as_checker(const char *lang, List *as, char **prosrc_str_p, char **probin_str_p);
 extern void pltsql_function_probin_writer(CreateFunctionStmt *stmt, Oid languageOid, char **probin_str_p);
 extern void pltsql_function_probin_reader(ParseState *pstate, List *fargs, Oid *actual_arg_types, Oid *declared_arg_types, Oid funcid);
-static void check_invalid_constraints(RangeVar *relation, ColumnDef *column);
+static void check_invalid_column_constraints(RangeVar *relation, ColumnDef *column);
+static void check_invalid_constraints(RangeVar *relation, Constraint *constraint);
 static bool checkAndSetTsqlSystemFunc(FuncCall *fc);
 static bool is_identity_constraint(ColumnDef *column);
 extern PLtsql_function *find_cached_batch(int handle);
@@ -1090,7 +1091,7 @@ pltsql_post_parse_analyze(ParseState *pstate, Query *query, JumbleState *jstate)
 						switch (nodeTag(element))
 						{
 							case T_ColumnDef:
-								check_invalid_constraints(stmt->relation,
+								check_invalid_column_constraints(stmt->relation,
 																   (ColumnDef *) element);
 								if (is_identity_constraint((ColumnDef *) element))
 								{
@@ -1177,7 +1178,7 @@ pltsql_post_parse_analyze(ParseState *pstate, Query *query, JumbleState *jstate)
 						switch (cmd->subtype)
 						{
 							case AT_AddColumn:
-								check_invalid_constraints(atstmt->relation,
+								check_invalid_column_constraints(atstmt->relation,
 																   castNode(ColumnDef, cmd->def));
 								if (is_identity_constraint(castNode(ColumnDef, cmd->def)))
 								{
@@ -1204,6 +1205,8 @@ pltsql_post_parse_analyze(ParseState *pstate, Query *query, JumbleState *jstate)
 							case AT_AddConstraint:
 								{
 									Constraint *c = castNode(Constraint, cmd->def);
+
+									check_invalid_constraints(atstmt->relation, c);
 
 									if (rowversion_column_name)
 										validate_rowversion_table_constraint(c, rowversion_column_name);
@@ -1254,6 +1257,17 @@ pltsql_post_parse_analyze(ParseState *pstate, Query *query, JumbleState *jstate)
 							case AT_ColumnDefault:
 								{
 									int			colnamelen = strlen(cmd->name);
+
+									if (nodeTag(cmd->def) == T_FuncCall)
+									{
+										if (atstmt->relation->relpersistence == RELPERSISTENCE_TEMP && 
+											!checkAndSetTsqlSystemFunc(castNode(FuncCall, cmd->def)))
+										{
+											ereport(ERROR,
+													(errcode(ERRCODE_INVALID_COLUMN_DEFINITION),
+													errmsg("User-defined functions, partition functions, and column references are not allowed in expressions in this context.")));
+										}
+									}
 
 									/*
 									 * Disallow defaults on a rowversion
@@ -1914,7 +1928,7 @@ revoke_type_permission_from_public(PlannedStmt *pstmt, const char *queryString, 
  * 2. User-defined functions (UDFs) are not allowed in TSQL temp table or table variable column defaults.
  */
 static void
-check_invalid_constraints(RangeVar *relation, ColumnDef *column)
+check_invalid_column_constraints(RangeVar *relation, ColumnDef *column)
 {
 	ListCell   *clist;
 	bool		is_null = false;
@@ -1923,6 +1937,7 @@ check_invalid_constraints(RangeVar *relation, ColumnDef *column)
 	foreach(clist, column->constraints)
 	{
 		Constraint *constraint = lfirst_node(Constraint, clist);
+		check_invalid_constraints(relation, constraint);
 
 		switch (constraint->contype)
 		{
@@ -1933,21 +1948,6 @@ check_invalid_constraints(RangeVar *relation, ColumnDef *column)
 			case CONSTR_IDENTITY:
 				is_identity = true;
 				break;
-			
-			case CONSTR_DEFAULT:
-			{
-				if (IsA(constraint->raw_expr, FuncCall))
-				{
-					FuncCall *fc = castNode(FuncCall, constraint->raw_expr);
-					if (relation->relpersistence == RELPERSISTENCE_TEMP && !checkAndSetTsqlSystemFunc(fc))
-					{
-						ereport(ERROR,
-								(errcode(ERRCODE_INVALID_COLUMN_DEFINITION),
-								errmsg("User-defined functions, partition functions, and column references are not allowed in expressions in this context.")));
-					}
-				}
-				break;
-			}
 
 			default:
 				break;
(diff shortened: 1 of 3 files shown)


contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -1939,6 +1959,35 @@ check_nullable_identity_constraint(RangeVar *relation, ColumnDef *column)
...
+	}
+	else
+	{
+		return false;
+	}
+
+	owa->objname = name_to_search;
+	owa->args_unspecified = true;
+
+	if (LookupFuncWithArgs(OBJECT_FUNCTION, owa, true))
```
timchang514, line 2025: how expensive is this call? I *think* it should be fine assuming it's just syscache lookups, since this would only be called in very rare cases when we need to actually check a column default function on a temp table.

timchang514 (approved):

contrib/babelfishpg_tsql/src/pl_handler.c
Sairakan: I don't know how expensive this call is, but as you said, this is buried below sufficient levels of checking that it shouldn't come up very often in the code path anyways. Basically, this would only get evaluated if we are dealing with a temp table using a function for a column default, and the function call is not schema-qualified. I suppose it could be made more efficient if we generated a list of all possible system functions and did a direct comparison against that list, but that seems like a lot of work.

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -1173,7 +1174,7 @@ pltsql_post_parse_analyze(ParseState *pstate, Query *query, JumbleState *jstate)
 						switch (cmd->subtype)
 						{
 							case AT_AddColumn:
-								check_nullable_identity_constraint(atstmt->relation,
+								check_invalid_constraints(atstmt->relation,
```
tanscorpio7: probably also need this check in ADDCONSTRAINT stmt.

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -1939,6 +1959,35 @@ check_nullable_identity_constraint(RangeVar *relation, ColumnDef *column)
...
+ */
+static bool
+isTsqlSystemFunc(FuncCall *fc)
+{
+	List *name_to_search;
+	ObjectWithArgs *owa = makeNode(ObjectWithArgs);
+	if (list_length(fc->funcname) == 1)
+	{
+		/* explicitly search in the "sys" schema */
+		name_to_search = list_make2(makeString("sys"), linitial(fc->funcname));
```
tanscorpio7: What if user created a function with the same name as a system function in dbo schema ? Do we want to allow them in defaults ?

tanscorpio7: Can we add tests for it as well.

tanscorpio7 (changes requested):

contrib/babelfishpg_tsql/src/pl_handler.c
Sairakan: In SQL Server, functions with the same name as system functions cannot be used as column defaults in temp tables:
```
1> create table #t1 (a int default dbo.isjson('hi'))
2> go
Msg 342, Level 16, State 1, Server sql1, Line 1
Column "dbo" is not allowed in this context, and the user-defined function or aggregate "dbo.isjson" could not be found.
```
And if the function call is not schema-qualified, then it will use the system function and not the user-defined function. These changes also reflect this behavior. I can add tests to validate this functionality as well.

tanscorpio7: Actually I was thinking without specifying the schema in function and relying on search path. for example
```
1> create database tdb
2> go
1> use tdb
2> go
Changed database context to 'tdb'.
1> CREATE FUNCTION isjson(@i text) returns int as begin return -99 end
2> go
1> create table #t1 (a int default isjson('1'), b int)
2> go
1> insert into #t1 (b) values (1)
2> go

(1 rows affected)
1> select * from #t1
2> go
a           b          
----------- -----------
        -99           1
```

Sairakan: In SQL Server, if a function is called without qualifying the schema, then it will default to the system function first, which is different behavior than in Babelfish (and should probably be fixed). Your example above works in Babelfish because we default to the user-defined function for non-fully-qualified function calls, but it would not work this way in SQL Server
```
1> use mydb
2> go
Changed database context to 'mydb'.
1> create function dbo.isjson(@json text) returns int as begin return 10 end
2> go
1> select isjson('hi')
2> go

-----------
          0

(1 row affected)
1> select dbo.isjson('hi')
2> go

-----------
         10

(1 row affected)
```

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -1939,6 +1959,76 @@ check_nullable_identity_constraint(RangeVar *relation, ColumnDef *column)
...
+	char *sys = palloc0(4);
+	strncpy(sys, "sys", 3);
+	if (list_length(fc->funcname) == 1)
+	{
+		/* explicitly search in the "sys" schema */
+		name_to_search = list_make2(makeString(sys), linitial(fc->funcname));
+	}
+	else
+	{
+		ereport(ERROR,
```
surendravishnoi, line 2015: does SQL server also give the error when we provide schema qualified function name?

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -1939,6 +1959,76 @@ check_nullable_identity_constraint(RangeVar *relation, ColumnDef *column)
...
+ * system function, to prevent inadvertently using any user-defined overrides for the function name.
+ */
+static bool
+checkAndSetTsqlSystemFunc(FuncCall *fc)
+{
+	List *name_to_search;
+	ObjectWithArgs *owa = makeNode(ObjectWithArgs);
+	char *sys = palloc0(4);
+	strncpy(sys, "sys", 3);
+	if (list_length(fc->funcname) == 1)
```
surendravishnoi, line 2008: this assumes that a function which matches to any sys function name should always resolve to sys function, I am not sure if that is always the case? Does SQL server always give sys functions priority over UDF with same name?

contrib/babelfishpg_tsql/src/pl_handler.c
timchang514: no, sql server allows schema qualified function name. sql server will throw an error if there is an unqualified function name in this context.

from discussion in slack: 

```
Even stronger: in SQL Server it is mandatory to specify the schema prefix for any call of a user-defined function. Without that, SQL Server gives an error, see below -- so Babelfish is OK to raise an error too in that case:
-- SQL Server
1> create table #t1 (a int default foo())
2> go
Msg 195, Level 15, State 10, Server EC2AMAZ-EULFEOJ, Line 1
```

```
1> create table #t1(a int default foo(1)
2> go
Msg 195, Level 15, State 10, Server 7ecd02cb107a, Line 1
'foo' is not a recognized built-in function name.
1> create table #t1(a int default dbo.foo(1))
2> go
Msg 342, Level 16, State 1, Server 7ecd02cb107a, Line 1
Column "dbo" is not allowed in this context, and the user-defined function or aggregate "dbo.foo" could not be found.
```

The only way in SQL Server to get a UDF as a column default function is to create that UDF in tempdb explicitly, and then it can only be used from tempdb. 

```
1> use tempdb;
2> go
1> create function bar(@a INT) RETURNS INT AS BEGIN RETURN 1 END
2> go

-- Doesn't work outside of tempdb
1> use master;
2> go
1> create table #t1(a int default tempdb.dbo.bar(1))
2> go
Msg 128, Level 15, State 1, Server 7ecd02cb107a, Line 1
The name "tempdb" is not permitted in this context. Valid expressions are constants, constant expressions, and (in some contexts) variables. Column names are not permitted.

-- Only works in tempdb
1> use tempdb
2> go
1> create table #t1(a int default dbo.bar(1))
2> go
```

contrib/babelfishpg_tsql/src/pl_handler.c
timchang514: yes, because we must specify the schema when invoking a UDF in sql server. anything unqualified means it will search in sys: 

```
1> select foo(1)
2> go
Msg 195, Level 15, State 10, Server 7ecd02cb107a, Line 1
'foo' is not a recognized built-in function name.
1> select dbo.foo(1)
2> go
           
-----------
          1

(1 rows affected)

--- built in func
1> select abs(1)
2> go
           
-----------
          1

(1 rows affected)

```

test/JDBC/expected/temp_table.out
```diff
@@ -258,3 +258,180 @@ int
...
+~~ROW COUNT: 1~~
+
+~~START~~
+int
+0
+~~END~~
+
+
+-- disallow "sys"-qualified function calls
+CREATE TABLE #t1 (a INT DEFAULT SYS.ISJSON('a'))
```
surendravishnoi, line 372: is this the same beahvior for SQL server?

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -1942,6 +1962,76 @@ check_nullable_identity_constraint(RangeVar *relation, ColumnDef *column)
...
+	{
+		/* explicitly search in the "sys" schema */
+		name_to_search = list_make2(makeString(sys), linitial(fc->funcname));
+	}
+	else
+	{
+		ereport(ERROR,
+				(errcode(ERRCODE_UNDEFINED_COLUMN),
+				 errmsg("Column \"%s\" is not allowed in this context, and the user-defined function or aggregate \"%s\" could not be found.",
```
surendravishnoi, line 2017: why are we saying Column "funcname" rather than actual column name

HarshLunagariya: Can you please add cases for following cases?
1. aggregate function
```
1> create table tbl(b int , a int default avg(b));
2> go
```
2. 3-part name - Ideal T-SQL behaviour is to throw a bit different error on that case. But We should be good IMO as long as we throw error.

```
1> create table tbl(b int , a int default db1.dbo.foo());
2> go
```

test/JDBC/expected/temp_table.out
timchang514: Yes

```
-- sql server
1> CREATE TABLE #t1 (a INT DEFAULT SYS.ISJSON('a'))
2> go
Msg 342, Level 16, State 1, Server 7ecd02cb107a, Line 1
Column "SYS" is not allowed in this context, and the user-defined function or aggregate "SYS.ISJSON" could not be found.
```

contrib/babelfishpg_tsql/src/pl_handler.c
timchang514: @surendravishnoi that wording matches the sql server error message. see below comment.

timchang514: Aggregate functions are not allowed in default expressions, and 3 part name is not allowed by this code either. I've added test cases corresponding to those.

surendravishnoi (approved):

PR #3126, opened 2024-11-18
Do not allow sp_ procedure calls in a SQL function

robverschoor:
### Description

T-SQL does not allow calling a stored procedure in a user-defined SQL function and will raise an execution-time error when such an execution is attempted. Babelfish currently enforces this for regular stored procedures, but not for system stored procedures (those whose names start with `sp_`). Some Babelfish releases ago, executing a system stored procedure in a function would crash the PG server, though in the latest codeline this may just raise an error like `InstrStartNode called twice in a row`. 
The reason is simply that the check in [exec_stmt_exec()](https://github.com/babelfish-for-postgresql/babelfish_extensions/blob/BABEL_4_X_DEV/contrib/babelfishpg_tsql/src/pl_exec-2.c#L949) does not cover the case of a call to a system stored procedure, therefore this fix adds this check to `exec_stmt_exec_sp()` as well.

NB. This is an execution-time check, presumably because SQL Server allows so-called extended stored procedures (named `xp_something`) to be called from functions, and these can also be called through a variable -- though neither variable procedure names nor extended stored procedures are currently supported in Babelfish. 

### Issues Resolved

BABEL-1730 SQL Function calling procedure crashes PG server

- Use case based - Yes

- Boundary conditions - Yes

- Arbitrary inputs - N/A

- Negative test cases - Yes

- Performance tests - N/A

- Tooling impact - N/A

- Client tests - N/A

### Check List

For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/babelfish-for-postgresql/babelfish_extensions/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).pl_exec-2.c#L949)

### DIFF
--- a/contrib/babelfishpg_tsql/src/pl_exec-2.c
+++ b/contrib/babelfishpg_tsql/src/pl_exec-2.c
@@ -1825,7 +1825,6 @@ exec_stmt_exec_sp(PLtsql_execstate *estate, PLtsql_stmt_exec_sp *stmt)
 
 	/* T-SQL doesn't allow procedure calls in a function */
 	if (estate->func && estate->func->fn_oid != InvalidOid && estate->func->fn_prokind == PROKIND_FUNCTION && estate->func->fn_is_trigger == PLTSQL_NOT_TRIGGER)
-	/* check EXEC is running in the body of function */
 	{
 		ereport(ERROR,
 				(errcode(ERRCODE_INVALID_FUNCTION_DEFINITION),
--- a/contrib/babelfishpg_tsql/src/pl_exec-2.c
+++ b/contrib/babelfishpg_tsql/src/pl_exec-2.c
@@ -945,7 +945,7 @@ exec_stmt_exec(PLtsql_execstate *estate, PLtsql_stmt_exec *stmt)
 
 		stmt->is_scalar_func = is_scalar_func;
 
-		/* T-SQL doesn't allow call procedure in function */
+		/* T-SQL doesn't allow procedure calls in a function */
 		if (estate->func && estate->func->fn_oid != InvalidOid && estate->func->fn_prokind == PROKIND_FUNCTION && estate->func->fn_is_trigger == PLTSQL_NOT_TRIGGER /* check EXEC is running
 																																									 * in the body of
 																																									 * function */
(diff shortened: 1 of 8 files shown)


test/JDBC/expected/exec_sp_in_udf-vu-prepare.out
```diff
@@ -0,0 +1,32 @@
+use master
+go
+create table t1_exec_sp_in_udf (a int)
+go
+create procedure sp_myproc as return 0
+go
+create function f_scalar_exec_sp_in_udf(@p int)
```
KushaalShroff, line 7: Ideally we would also want MVU tests where functions have already been created using sp_ procedures in the body in previous versions (prepare script) and then verified on latest

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -1823,6 +1823,15 @@ exec_stmt_exec_sp(PLtsql_execstate *estate, PLtsql_stmt_exec_sp *stmt)
 	char	   *querystr;
 	int			ret = 0;
 
+	/* T-SQL doesn't allow procedure calls in a function */
+	if (estate->func && estate->func->fn_oid != InvalidOid && estate->func->fn_prokind == PROKIND_FUNCTION && estate->func->fn_is_trigger == PLTSQL_NOT_TRIGGER)
+	/* check EXEC is running in the body of function */
```
KushaalShroff: This comment is redundant in this case since we arent using is_scalar_func

test/JDBC/expected/exec_sp_in_udf-vu-verify.out
```diff
@@ -0,0 +1,206 @@
...
+exec p1_exec_sp_in_udf
+go
+~~START~~
+int
+123
+~~END~~
+
+
+-- scalar UDF can still be called with EXEC
+declare @v int
```
KushaalShroff, line 260: Can we add the following test case as well:
```
1> declare @a int
2> set @a = dbo.f1_exec_sp_in_udf()
3> go
Msg 557, Level 16, State 2, Server EC2AMAZ-EULFEOJ, Line 2
Only functions and some extended stored procedures can be executed from within a function.
```

test/JDBC/expected/exec_sp_in_udf-vu-verify.out
```diff
@@ -0,0 +1,206 @@
...
+int
+123
+~~END~~
+
+
+create function f9_exec_sp_in_udf(@p int) 
+returns int as
+begin
+declare @v int
+exec @v = f_scalar_exec_sp_in_udf @p
```
KushaalShroff, line 274: Lets also use f1_exec_sp_in_udf

KushaalShroff (changes requested):

contrib/babelfishpg_tsql/src/pl_exec-2.c
robverschoor: removed

test/JDBC/expected/exec_sp_in_udf-vu-verify.out
robverschoor: added this case plus various other variations

test/JDBC/expected/exec_sp_in_udf-vu-verify.out
robverschoor: done

test/JDBC/expected/exec_sp_in_udf-vu-prepare.out
robverschoor: In some previous releases, this could crash the PG server, so avoiding including that in the tests.

KushaalShroff (approved):

PR #3138, opened 2024-11-20
Fix namespace for string functions in system views/functions

Anikait143:
### 1. Issues
- As an effort to fix string functions, we introduced various function definitions is sys schema, but there are a few places where we overlooked to prepend PG_CATALOG in the function calls inside system view/functions which may cause mVU/MVU failures due to error while executing upgrade scripts due to invalid argument type.
- Affected functions were - RTRIM, STRING_AGG, SUBSTRING, LOWER.

- Prepended appropriate namespace at suitable function calls as a proactive measure to prevent m/MVU failures, Since, invalid datatypes for sys function definitions which can cause issue will now be redirected to pg_catalog function definitions and will work correctly as before.

Task: BABEL- 5362

* **Use case based -**

* **Boundary conditions -**

* **Arbitrary inputs -**

* **Negative test cases -**

* **Performance tests -**

* **Tooling impact -**

* **Memory tests -**

* **Client tests -**

### Check List


rohit01010 (approved):

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--4.4.0--5.0.0.sql
```diff
@@ -288,13 +288,13 @@ $$
 DECLARE has_role BOOLEAN;
 DECLARE login_valid BOOLEAN;
 BEGIN
-	role  := TRIM(trailing from LOWER(role));
```
Deepesh125, line 291: I think we also need to fix older upgrade scripts, say, 4.x--4.y if it contains reference to any of these functions. Because upgrade scripts generally runs with latest binary where we have special handling for these functions. This may in turn fail. Please verify this and fix if needed.

Anikait143: We have verified that in previous upgrade scripts there was only one such problematic reference which we have already fixed:
```
CREATE OR REPLACE VIEW information_schema_tsql.schemata AS
	SELECT CAST(sys.db_name() AS sys.sysname) AS "CATALOG_NAME",
	CAST(CASE WHEN np.nspname LIKE PG_CATALOG.CONCAT(sys.db_name(),'%') THEN RIGHT(np.nspname, LENGTH(np.nspname) - LENGTH(sys.db_name()) - 1)
	     ELSE np.nspname END AS sys.nvarchar(128)) AS "SCHEMA_NAME",
	-- For system-defined schemas, schema-owner name will be same as schema_name
	-- For user-defined schemas having default owner, schema-owner will be dbo
	-- For user-defined schemas with explicit owners, rolname contains dbname followed
	-- by owner name, so need to extract the owner name from rolname always.
	CAST(CASE WHEN sys.bbf_is_shared_schema(np.nspname) = TRUE THEN np.nspname
		  WHEN r.rolname LIKE PG_CATALOG.CONCAT(sys.db_name(),'%') THEN
			CASE WHEN RIGHT(r.rolname, LENGTH(r.rolname) - LENGTH(sys.db_name()) - 1) = 'db_owner' THEN 'dbo'
			     ELSE RIGHT(r.rolname, LENGTH(r.rolname) - LENGTH(sys.db_name()) - 1) END ELSE 'dbo' END
			AS sys.nvarchar(128)) AS "SCHEMA_OWNER",
	CAST(null AS sys.varchar(6)) AS "DEFAULT_CHARACTER_SET_CATALOG",
	CAST(null AS sys.varchar(3)) AS "DEFAULT_CHARACTER_SET_SCHEMA",
	-- TODO: We need to first create mapping of collation name to char-set name;
	-- Until then return null for DEFAULT_CHARACTER_SET_NAME
	CAST(null AS sys.sysname) AS "DEFAULT_CHARACTER_SET_NAME"
	FROM ((pg_catalog.pg_namespace np LEFT JOIN sys.pg_namespace_ext nc on np.nspname = nc.nspname)
		LEFT JOIN pg_catalog.pg_roles r on r.oid = nc.nspowner) LEFT JOIN sys.babelfish_namespace_ext ext on nc.nspname = ext.nspname
	WHERE (ext.dbid = sys.db_id() OR np.nspname in ('sys', 'information_schema_tsql')) AND
	      (pg_has_role(np.nspowner, 'USAGE') OR has_schema_privilege(np.oid, 'CREATE, USAGE'))
```
Other than that we have also tested the previous upgrade scripts with different combinations of search path and they ran successfully, so I think it would be sufficient to include these changes in latest upgrade scriopt only.
```
set search_path = pg_catalog;
set search_path = sys;
set search_path = sys, pg_catalog;
set search_path = sys, public;
set search_path = sys, pg_catalog, public;
set search_path = sys, public, pg_catalog;
set search_path = pg_catalog, sys;
set search_path = public, sys, pg_catalog;
```

Deepesh125 (approved):

PR #3141, opened 2024-11-21
Fix the issue about Like collation not properly transformed

forestkeeper:
There is a missing condition for transforming an OpExpr when the like node is in the arg list of OpExpr

Task: BABEL-5397, BABEL-5424

### Description

Previously we didn't transform args of OpExpr, it'll miss a condition when OpExpr is '=' OpExpr, with left arg is a Case When wrapped a like Expression and right arg is a Const Value. In such case the like node inside Case When will not be transformed and will lead to a non deterministic collation runtime error.

This fix will transform all the args of OpExpr before transform the OpExpr itself.

For APG 17, we noticed this error because the in subplan will be transformed into a Join.

### Issues Resolved

BABEL-5397, BABEL-5424

### Check List

### DIFF
--- a/contrib/babelfishpg_tsql/src/collation.c
+++ b/contrib/babelfishpg_tsql/src/collation.c
@@ -1010,6 +1010,7 @@ pltsql_predicate_transformer(Node *expr)
 
 	if (IsA(expr, OpExpr))
 	{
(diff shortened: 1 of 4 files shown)


Deepesh125:
Some simple tests related to LIKE are failing. Please fix those.

Deepesh125 (approved):

PR #3154, opened 2024-11-25
Modify upgrade scripts to obtain correct server level collation during upgrade

ahmed-shameem:
### Description
In all MVU sql scripts, we update the server collation name and reset the restored_server_collation_name GUC. During the upgrade from 16.1 --> latest (17), the upgrade started failing for non-default server collation like chinese_prc_ci_as as we did not have correct collation information and the collation of columns of views were trying to get updated. This is because, we never updated the collation related information in any of the upgrade script post 16.1, hence it was always picking the default value which is bbf_unicode_cp1_ci_as. 
The issue may come up in case when user is performing upgrade from some version of 15 (before 15.6) to latest if we have defines some view which contains collatable column in upgrade script and the server level collation is different than default collation.

We address the issue by updating the collation related information in the necessary upgrade scripts. 

### Issues Resolved ###
BABEL-5415

* **Use case based -**

* **Boundary conditions -**

* **Arbitrary inputs -**

* **Negative test cases -**

* **Performance tests -**

* **Tooling impact -**

* **Client tests -**

### Check List


contrib/babelfishpg_common/sql/upgrades/babelfishpg_common--3.4.0--4.0.0.sql
```diff
@@ -1,2 +1,2 @@
 -- complain if script is sourced in psql, rather than via ALTER EXTENSION
-\echo Use "ALTER EXTENSION ""babelfishpg_common"" UPDATE TO '4.0.0'" to load this file. \quit
+\echo Use "ALTER EXTENSION ""babelfishpg_common"" UPDATE TO '4.0.0'" to load this file. \quit
```
rishabhtanwar29: Unnecessary change.

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--3.0.0--3.1.0.sql
```diff
@@ -31,6 +31,26 @@ end
...
+do
+language plpgsql
+$$
+    declare
+        query text;
+    begin
+        query := pg_catalog.format('alter database %s reset babelfishpg_tsql.restored_server_collation_name', CURRENT_DATABASE());
+        execute query;
+    end;
+$$;
```
rishabhtanwar29, line 62: I am thinking if we re-set this guc once in the first upgrade script then what happens when we again execute the above procedure in the next upgrade script with the default value (NULL?). Are we handling this in the `C` implementation of the procedure?

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--3.0.0--3.1.0.sql
```diff
@@ -31,6 +31,26 @@ end
 $$
 LANGUAGE plpgsql;
 
+CREATE OR REPLACE FUNCTION sys.babelfish_update_server_collation_name() RETURNS VOID
```
Deepesh125, line 34: if this issue only comes up when we upgrade from 15.x (x<=6) to 17 latest then how is this upgrade path coming into picture?

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--3.1.0--3.2.0.sql
```diff
@@ -31,6 +31,26 @@ end
 $$
 LANGUAGE plpgsql;
 
+CREATE OR REPLACE FUNCTION sys.babelfish_update_server_collation_name() RETURNS VOID
```
Deepesh125, line 34: I am not sure why is this required?

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--4.1.0--4.2.0.sql
```diff
@@ -31,6 +31,26 @@ end
 $$
 LANGUAGE plpgsql;
 
+CREATE OR REPLACE FUNCTION sys.babelfish_update_server_collation_name() RETURNS VOID
```
Deepesh125, line 34: this also should not be required. only 4.0--4.1 is needed to handle upgrades

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--4.4.0--5.0.0.sql
```diff
@@ -37,6 +37,27 @@ LANGUAGE plpgsql;
  * So make sure that any SQL statement (DDL/DML) being added here can be executed multiple times without affecting
  * final behaviour.
  */
+
+CREATE OR REPLACE FUNCTION sys.babelfish_update_server_collation_name() RETURNS VOID
```
Deepesh125, line 41: this should also be not required

Deepesh125 (changes requested):

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--3.0.0--3.1.0.sql
ahmed-shameem: It is possible that when source is 15.x (x <6) and targrt is 17.y, then we will come till 15.6 and then upgrade to 16 and then 17. For the first step, it is possible that the collation of some view may have been changed. As our upgrade test framework does not cover all such possible upgrade paths for non-default collations, added this to be on the safe side.

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--3.1.0--3.2.0.sql
ahmed-shameem: Same reason as above. We need to put update the collation info for all such versions considering the user is on 15.1, 15.2 and so on. Only one upgrade script is not sufficient to address the issue.

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--4.1.0--4.2.0.sql
ahmed-shameem: Actually, if a user is on extension version 4.3/4.4 (and so on) and trying to go to 5.x then they may face the same error.

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--4.4.0--5.0.0.sql
ahmed-shameem: This is required because if any user is on extension version 4.4 and trying to upgrade to 5.0, the upgrade will fail.

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--3.0.0--3.1.0.sql
Deepesh125: Got it

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--3.0.0--3.1.0.sql
Deepesh125: yes, this will cause an issue as we are raising error if restored_collation_name is NULL

rishabhtanwar29: But that doesn't sound right isn't it? We are going to repeatedly call this procedure and reset this GUC in each upgrade script so does that mean we are going to fail the second time we call this procedure since the GUC would have been reset now?

ahmed-shameem: Updated in latest rev.

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--3.0.0--3.1.0.sql
```diff
@@ -31,6 +31,37 @@ end
...
+LANGUAGE C
+AS 'babelfishpg_common', 'babelfish_update_server_collation_name';
+
+DO
+LANGUAGE plpgsql
+$$
+BEGIN
+    -- Check if the GUC is empty
+    IF current_setting('babelfishpg_tsql.restored_server_collation_name', true) <> '' 
+        OR current_setting('babelfishpg_tsql.restored_server_collation_name', true) IS NOT NULL THEN
```
rishabhtanwar29: This should be `AND` right? Also, we should first check if it is NOT NULL and then check if it not empty.

contrib/babelfishpg_common/sql/upgrades/babelfishpg_common--3.4.0--4.0.0.sql
ahmed-shameem: Removed in latest rev

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--3.0.0--3.1.0.sql
ahmed-shameem: I have updated the condition in latest rev. As discussed offline, NULL check is not required, check for empty string is sufficient.

rishabhtanwar29 (approved):
LGTM

shalinilohia50 (approved):

Deepesh125 (approved):

Deepesh125 (approved):

PR #3161, opened 2024-11-25
Fix for cache reference leak warning

pranavJ23:
### Description
Problem: Cache Reference Leaks
1. During the Alter function/procedure we are doing a sys-cache search without releasing it. 
2. During the routines-definition, in one case we were not releasing the syscache search.

Fix: 
 a. To do sys-cache search via tuple and releasing it after the sys-cache search.
 b. Also we are generating the logfile but the path that we were giving it github action workflow was not correct, updated 
  that too.
 c. Added a step in jdbc-test workflow which will compare the warnings in logfile with expected warnings

### Issues Resolved
[BABEL-5551]

### Check List


contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -2600,13 +2601,26 @@ bbf_ProcessUtility(PlannedStmt *pstmt,
 						originalFunc.objectId = oldoid;
 						originalFunc.classId = ProcedureRelationId;
 						originalFunc.objectSubId = 0;
-						if(get_bbf_function_tuple_from_proctuple(SearchSysCache1(PROCOID, ObjectIdGetDatum(oldoid))) == NULL)
+						proctup = SearchSysCache1(PROCOID, ObjectIdGetDatum(oldoid));
+
+						if (proctup == NULL)
```
tanscorpio7: we should use `HeapTupleIsValid` & curly braces can be removed in next line.

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -2600,13 +2601,26 @@ bbf_ProcessUtility(PlannedStmt *pstmt,
...
-								(errcode(ERRCODE_INVALID_FUNCTION_DEFINITION),
-									errmsg("No existing TSQL procedure found with the name for ALTER PROCEDURE")));
+						    ereport(ERROR,
+						            (errcode(ERRCODE_UNDEFINED_OBJECT),
+						             errmsg("cache lookup failed for procedure %u", oldoid)));
 						}
+						if (get_bbf_function_tuple_from_proctuple(proctup) == NULL)
+						{
+						    /* Release the tuple before reporting the error */
+						    ReleaseSysCache(proctup);
```
tanscorpio7: In case of error we do not need to release syscache

tanscorpio7 (changes requested):

.github/workflows/jdbc-tests.yml
```diff
@@ -100,6 +100,26 @@ jobs:
...
 
+      - name: Check for unexpected leak warnings
+        if: always()
+        id: check-leak-warnings
+        run: |
+          # Set the path to the expected-warnings.out file
+          EXPECTED_WARNINGS_FILE=~/work/babelfish_extensions/babelfish_extensions/test/JDBC/expected/expected-warnings.out
+
+          # Collect all leak warnings from log files, extracting only the warning type
+          grep -i "WARNING:.*leak" ~/psql/data/logfile | sed -E 's/^.*WARNING:\s*([^:]+(\s+reference\s+leak|connection leak)[^:]*):.*/WARNING:  \1/' | sort  > actual-leak-warnings.out
```
tanscorpio7: Some warnings do not contain the word leak but are still important for us & need to be resolved.
For example `elog(WARNING, "snapshot %p still active", active);`
Will this check be able to catch this WARNING if it comes up in future ?

tanscorpio7 (changes requested):

.github/workflows/jdbc-tests.yml
pranavJ23: To grep the warnings which do not contains leak, we need to either grep only "WARNING" or we need to grep the whole line containing "WARNING" keyword. 
If we only grep WARNING in the expected-file then while debugging it might be difficult to tell that which type of new warning we are getting.
If we grep whole line containing "WARNING" then there are some reference tuple which might be dynamically allocated during test run, so we cannot compare the expected and actual files.
Open to suggestions!

tanscorpio7 (approved):

contrib/babelfishpg_tsql/src/pltsql_ruleutils.c
```diff
@@ -537,7 +537,10 @@ tsql_get_functiondef(PG_FUNCTION_ARGS)
 	(void) tsql_print_function_arguments(&buf, proctup, false, true, &typmod_arr, &has_tvp);
 	/* TODO: In case of Table Valued Functions, return NULL. */
 	if (has_tvp)
+	{
+		ReleaseSysCache(proctup);
```
sumitj824, line 544: There is one more PG_RETURN_NULL() at line 528, we need to release sys cache there as well.

.github/workflows/jdbc-tests.yml
```diff
@@ -100,6 +100,73 @@ jobs:
             ~/psql/data/logfile
             ~/psql/data_5433/logfile
 
+      - name: Check for unexpected warnings
+        if: always()
+        id: check-warnings
+        run: |
+          LOG_FILE=~/psql/data/logfile
```
sumitj824: Lets have bash script for this check and you can invoke that script directly.

.github/workflows/jdbc-tests.yml
```diff
@@ -100,6 +100,73 @@ jobs:
...
+
+          if [[ "$SNAPSHOT_ACTIVE_COUNT" -ne 44 ]]; then
+              echo "Error: Expected 44 'snapshot ... still active' warnings, but found $SNAPSHOT_ACTIVE_COUNT"
+              ERROR_FOUND=true
+          fi
+
+          if [[ "$LEAK_COUNT" -ne 380 ]]; then
+              echo "Error: Expected 380 leak warnings, but found $LEAK_COUNT"
+              ERROR_FOUND=true
+          fi
```
sumitj824: Does this mean that for any new testfile which will generate warnings that are expected will increase this count and everytime developer needs to update this?

pranavJ23: Yes if that generates the LEAK or SNAPSHOT warning then only this count will change, for that case developer should check why the warning is generated
If there is some other warning, which is expected we are ignoring them. For a new kind of warning we are saying other unexpected warning

sumitj824 (approved):

shardgupta (approved):

PR #3162, opened 2024-11-25
Disallow declaring reserved @@variable names

robverschoor:
### Description

Now that variables named `@@myvar` are supported (see [PR #3073](https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/3073)), we should also address the issue of being able to declare a user-defined variable with the same name as a predefined global `@@` variable, for example `@@SPID`.
Babelfish has always allowed such declarations even though referencing the declared variable would actually reference the predefined global `@@variable`, thus producing a wrong result. For example:

```
1> declare @@spid int = 123
2> select @@spid
3> go
spid
-----------
      20902     <===== WRONG!
```
      
This fix does not allow predefined `@@` global variables to be used as user-defined variables or parameters, and will raise a syntax error as expected in T-SQL.
A special case is the `@@PGERROR` variable which was introduced by Babelfish: when declaring a variable with this name, the error message will additionally point out that `@@PGERROR` is a reserved name in Babelfish. 

### Issues Resolved

BABEL-5388: Disallow declaring reserved @@variable names

- Use case based - Yes

- Boundary conditions - Yes

- Arbitrary inputs - N/A

- Negative test cases - Yes

- Minor version upgrade tests - Yes

- Performance tests - N/A

- Tooling impact - N/A

- Client tests - N/A

### Check List


test/JDBC/expected/declare_atatglobalvars_upgrade-before-16_5-vu-verify.out
```diff
@@ -0,0 +1,16 @@
+exec p1_declare_atatglobalvars_upgrade
```
KushaalShroff: Why does this not exist post upgrade? Ideally the prepare script would run on 16.4 and when we upgrade to 16.5/16.6 then the object should still exist

robverschoor: The upgrade would fail as the object cannot be recreated anymore. With the fix in this PR, the failure happens when the object is created.

KushaalShroff: Object wont be recreated from 16.4 to 16.5/16.6, even if they did, the restore happens from psql endpoint and ANTLR does not come into picture. Dump/restore happens only for Major version upgrade.

test/JDBC/expected/declare_atatglobalvars_upgrade-before-16_5-vu-verify.out
```diff
@@ -0,0 +1,20 @@
...
+go
+~~ERROR (Code: 33557097)~~
+
+~~ERROR (Message: Incorrect syntax near '@@max_precision'.)~~
+
+exec p1_declare_atatglobalvars_upgrade
+go
+~~ERROR (Code: 33557097)~~
+
+~~ERROR (Message: Incorrect syntax near '@@max_precision'.)~~
```
KushaalShroff, line 11: So we have a case where previous functions will stop working are we expecting the customer experience to be the same?
Can we take the risk of assuming that there wouldnt be such dependent objects in the first place since these are not valid TSQL?

test/JDBC/input/declare_atatglobalvars-vu-verify.sql
```diff
@@ -0,0 +1,212 @@
+declare @@cursor_rows int
```
KushaalShroff, line 1: declare_atatglobalvars For this file , I think we need to add it to the latest/schedule file as well

KushaalShroff (changes requested):

test/JDBC/expected/declare_atatglobalvars_upgrade-before-16_5-vu-verify.out
robverschoor: If such objects had existed, which would be highly unlikely since they could not have been created in SQL Server in the first place, then they would have produced incorrect results (see PR description). That was wrong, and this fix stops it from happening. This is an extreme edge scenario and would only occur if customers had created an object using syntax that was invalid in SQL Server.

test/JDBC/input/declare_atatglobalvars-vu-verify.sql
robverschoor: done

KushaalShroff (approved):

PR #3179, opened 2024-11-28
Fix the issue about Like collation not properly transformed (#3141)

forestkeeper:
There is a missing condition for transforming an OpExpr when the like node is in the ars list of OpExpr

Task: BABEL-5397, BABEL-5424

Cherry-pick from commit : https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/3141

### Check List

### DIFF
--- a/test/JDBC/input/BABEL-4046-vu-verify.sql
+++ b/test/JDBC/input/BABEL-4046-vu-verify.sql
@@ -35,6 +35,6 @@ GO
 	select DM.empno, DM.ename, DM.baseloc from EMP_T DM where DM.baseloc in (select baseloc from t4)  order by DM.empno
 GO
 
-
-
-
+;with EMP_T AS ( select empno, ename, CASE WHEN baseloc LIKE 'AUS%' THEN REPLACE(baseloc,'AUSTIN','A') ELSE baseloc END AS baseloc, deptno from t3)
+	select DM.empno, DM.ename, DM.baseloc from EMP_T DM where DM.baseloc = 'A'  order by DM.empno
+GO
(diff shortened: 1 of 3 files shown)


test/JDBC/input/BABEL-4046-vu-verify.sql
```diff
@@ -35,6 +35,6 @@ GO
 	select DM.empno, DM.ename, DM.baseloc from EMP_T DM where DM.baseloc in (select baseloc from t4)  order by DM.empno
 GO
 
-
-
-
+;with EMP_T AS ( select empno, ename, CASE WHEN baseloc LIKE 'AUS%' THEN REPLACE(baseloc,'AUSTIN','A') ELSE baseloc END AS baseloc, deptno from t3)
```
ahmed-shameem, line 38: We should add tests for cases when the left and right arg is different node as well (wrapped inside case when). For example:
1. LA - Const, RA - ColumnRef
2. LA and RA - Const
3. LA and RA - ColumnRef
4. What is the behaviour if  CASE WHEN is wrapped inside another CASE WHEN with LIKE
5. Behaviour in this case -- CASE WHEN (like is wrapped in it) LIKE CASE WHEN (like is wrapped in it) 

Such variations needs to be tested.

forestkeeper: The issue is not happening inside case when, the operator not get transformed is actually '=' , so I added the args transform method for those operators.
For case when transform, the implement is inside : pgtsql_expression_tree_mutator function. And also, those test cases are actually already in other test cases files.

Deepesh125: @forestkeeper I agree but first three cases seems relevant here. @ahmed-shameem can you please provide example also?

ahmed-shameem: When we say '=' is not getting transformed, do we mean the case when ColRef LIKE Const (with exact match)? As in this case we simply call the '=' operator. If that is the case, then with the current test we are not testing it as Const prefix is transformed to '>=' operator. So, we need to update the test case and the cases which I mentioned applies as well because we can make some modifications to have the desired operator.

forestkeeper: It's not necessarily need to be a case when.
I don't understand what exactly is Const prefix is transformed to '>=' operator, so pls give an example.

ahmed-shameem: As discussed offline, we should add tests for other OpExpr like <,>,<=,>=. We should also test for this case DM.baseloc1 = DM.baseloc2

ahmed-shameem (approved):
LGTM.

xhfanhe (approved):

PR #3184, opened 2024-11-29
Fixed server roles except sysadmin should not have membership in database guest roles

anju15bharti:
### Description

With this commit we ensure that any fixed server role other than sysadmin should not have membership in databases' guest role.

### Issues Resolved

BABEL-5408

* **Use case based -**

* **Boundary conditions -**

* **Arbitrary inputs -**

* **Negative test cases -**

* **Performance tests -**

* **Tooling impact -**

* **Client tests -**

### Check List


contrib/babelfishpg_tsql/src/rolecmds.c
```diff
@@ -181,8 +181,9 @@ create_bbf_authid_login_ext(CreateRoleStmt *stmt)
 	/* Advance cmd counter to make the insert visible */
 	CommandCounterIncrement();
 
-	/* Grant membership to guests */
-	if (!role_is_sa(roleid))
+	/* Grant membership of guests */
+	/* sa and fixed server roles except sysadmin should not have membership in database guest roles */
+	if (!(role_is_sa(roleid) || ((get_sysadmin_oid() != roleid) && IS_BBF_FIXED_SERVER_ROLE(stmt->role))))
```
HarshLunagariya, line 186: 1. Why explicit condition for sysadmin? Fixed server roles check should cover it, isn't it?
2. Will need to revoke guest membership from sysadmin from older versions? Because existing code was not blocking guest membership grant to sysadmin it seems.

anju15bharti: The condition is to exclude sysadmin case because `IS_BBF_FIXED_SERVER_ROLE` will include that and we don't want to exclude membership of guests for sysadmin. Also Sysadmin is admin of the server so no need to remove any priv from it.

test/JDBC/expected/dbcreator_role-vu-verify.out
```diff
@@ -1,4 +1,16 @@
 -- tsql
+-- dbcreator should not have membership in database guest role
+SELECT r.rolname AS parent_role, mr.rolname as member 
+FROM pg_auth_members m 
+JOIN pg_roles r ON (m.roleid = r.oid) 
+JOIN pg_roles mr ON (m.member = mr.oid) 
+where mr.rolname = 'dbcreator';
```
tanscorpio7: suggestion: alternative smaller query we can execute in psql endpoint `SELECT roleid::regrole, member::regrole FROM pg_auth_members WHERE member = 'dbcreator'::regrole;`

test/JDBC/expected/dbcreator_role-vu-verify.out
```diff
@@ -103,6 +115,18 @@ dbcreator_login1
...
+go
+~~START~~
+varchar#!#varchar
+~~END~~
+
+
```
tanscorpio7: Please also include test cases like 
login is member of db_creator ....  login has a mapped user in a database and login is not able to select on tables that are granted to guest but not to its mapped user.

test/JDBC/expected/dbcreator_role-vu-verify.out
tanscorpio7: lets also runs this query for sysadmin

contrib/babelfishpg_tsql/src/rolecmds.c
tanscorpio7: Lets keep the check either on oid or name for server level role. I think we already have a macro for `IS_ROLENAME_SYSADMIN` so may be convert it to name.

tanscorpio7 (changes requested):

test/JDBC/expected/dbcreator_role-vu-verify.out
anju15bharti: If we don't have membership in guest that will result in not granting but will add the testcase.

contrib/babelfishpg_tsql/src/rolecmds.c
anju15bharti: We are not additionally getting the oid or the name. Both are already present and I think oid check will be comparatively faster than name, Also we have already used check with oid in the same condition, not getting reason why to make it to `IS_ROLENAME_SYSADMIN `?

test/JDBC/expected/dbcreator_role-vu-verify.out
anju15bharti: We haven't handled any change for sysadmin in this which means it will continue with the membership as earlier.

tanscorpio7: Please ensure we have automated tests, just to make sure some thing else does not break it in future.

tanscorpio7 (approved):

HarshLunagariya (approved):

shardgupta (approved):

PR #3207, opened 2024-12-05
Update  RESTORE behaviour for LIKE operator for CS_AI collation

ahmed-shameem:
### Description
We introduced the support for LIKE operator for AI collations for Latin characters from versions 15.7 and 16.3 onwards. We use function remove_accents_internal to remove accents from the characters to find the base character. To improve the performance further, we introduced a cached version of remove_accents_internal function named remove_accents_internal_using_cache from 15.8 and 16.4 onwards. 

In cases where there is CHECK constraint in the table definition for CS_AI, we observed that after upgrade from any version before 15.8 to 16.latest, the INSERT was failing on the table as another call to remove_accents_internal_using_cache was getting added on top of current node during restore. This issue is only observed for CS_AI as we do not modify the LIKE operator whereas for CI_AI we modify LIKE to ILIKE and BETWEEN operators. 

We fix this issue by checking whether the incoming LIKE operator is for CS_AI. If yes, then we update the collation information for the OpExpr for that node and return from there directly. In this way, during restore we won't go in the code path where we add extra call to remove_accents_internal* on already modified node. 

### Issues Resolved ###
BABEL-5471

* **Use case based -**

* **Boundary conditions -**

* **Arbitrary inputs -**

* **Negative test cases -**

* **Performance tests -**

* **Tooling impact -**

* **Client tests -**

### Check List


contrib/babelfishpg_tsql/src/collation.c
```diff
@@ -942,12 +942,27 @@ transform_likenode(Node *node)
...
 		 * during dump and restore.
+		 * For CS_AI collation, we keep the LIKE operator but convert it to
+		 * CS_AS by adding a call to remove_accents_internal*. During restore,
+		 * it will take the same code path and will try to add a new call to
+		 * remove_accents_internal* on top of current node. So, we take care
+		 * of that case here.
 		 */
 		init_and_check_collation_callbacks();
-		if ((*collation_callbacks_ptr->has_ilike_node) (node) && babelfish_dump_restore)
+		if (((*collation_callbacks_ptr->has_ilike_node) (node) || (*collation_callbacks_ptr->has_like_node) (node)) && babelfish_dump_restore)
```
Deepesh125: lets move babelfish_dump_restore to the front of the condition

contrib/babelfishpg_tsql/src/collation.c
```diff
@@ -942,12 +942,27 @@ transform_likenode(Node *node)
...
+			/* 
+			 * We only need to update the collateflags for CS_AI to CS_AS
+			 * No need to check whether ilike or like is present as
+			 * ilike comes only for CI collations. 
+			 * So, we know for sure that if this is CS_AI, only LIKE node 
+			 * will be present, hence we can update.
+			 */
+			if (coll_info_of_inputcollid.collateflags == 0x000e)
+				coll_info_of_inputcollid.collateflags = 0x000c;
+
```
Deepesh125: Changing flags just to avoid following if condition is really not needed. Lets refactor this code a bit. 

Leave original code for ILIKE as it is and introduce new piece of code above that to handle our requirement and return early from this function.

Deepesh125 (changes requested):

contrib/babelfishpg_tsql/src/collation.c
ahmed-shameem: There are 2 things to consider here:
1. Provide the correct collation information to OpExpr node
2. Need to update the collateflags
The first point is needed to have the correct collation information. For our case, it will be CS_AS as for all the other collation we update the underlying node in that way. Till now we were taking care of this for ILIKE, which means CI collations were getting updated with correct information. 
But for LIKE, which means CS_AI (we do not modify the operator) it does not pick correct information. So we need to update that anyway.

Secondly, the collateflags needs to updated as we are setting the collation information to CS_AS and we do not want to go through the handling of CS_AI collation for LIKE. We handle different collation based on collateflags and whether the operator is LIKE or not. So in the first step we have the correct meta-data, in the second step we update the collateflags to correct collateflag, else it will be CS_AI which is not expected.

contrib/babelfishpg_tsql/src/collation.c

contrib/babelfishpg_tsql/src/collation.c
ahmed-shameem: As discussed offline, updated in latest rev

Deepesh125 (approved):

PR #3217, opened 2024-12-06
Fix transaction leak when DMLs contains a non pltsql language function which calls another non tsql function inside try catch

tanscorpio7:
### Description

When DMLs are executed outside a user transaction block, babelfish starts an implicit transaction to handle errors inside triggers. This is done irrespective of whether the statement actually fires a trigger.
Now if the DML calls a non pltsql function which in calls another non tsql func or procedure inside a try catch, we do not properly decrement the `pltsql_sys_func_entry_count` variable properly. So the return value of `pltsql_support_tsql_transactions()` is different at beginning of the DML compared to after the end of the statement. This difference leads to babelfish not commit the implict transaction block.
As a fix, store the value of pltsql_support_tsql_transactions() at start of execution of DML and just reuse it for after execution.

### Issues Resolved

[BABEL-5446]

### Sign Off

### Check List

### DIFF
--- a/contrib/babelfishpg_tsql/src/pl_exec.c
+++ b/contrib/babelfishpg_tsql/src/pl_exec.c
@@ -4622,6 +4622,7 @@ exec_stmt_execsql(PLtsql_execstate *estate,
 	bool		fmtonly_enabled = true;
 	CmdType		cmd = CMD_UNKNOWN;
 	bool		enable_txn_in_triggers = !pltsql_disable_txn_in_triggers;
+	bool		support_tsql_trans = pltsql_support_tsql_transactions();
 	StringInfoData query;
 	bool		need_path_reset = false;
 	char	   *cur_dbname = get_cur_db_name();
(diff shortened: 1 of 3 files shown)


contrib/babelfishpg_tsql/src/pl_exec.c
```diff
@@ -4622,6 +4622,7 @@ exec_stmt_execsql(PLtsql_execstate *estate,
 	bool		fmtonly_enabled = true;
 	CmdType		cmd = CMD_UNKNOWN;
 	bool		enable_txn_in_triggers = !pltsql_disable_txn_in_triggers;
+	bool		support_tsql_trans = pltsql_support_tsql_transactions();
```
surendravishnoi, line 4626: Please see if this fix is needed for other callers of pltsql_support_tsql_transactions

tanscorpio7: Also made this change in `pltsql_exec_trigger`. Change was not needed there, but did it anyway for consistency in code.

surendravishnoi (approved):

PR #3218, opened 2024-12-06
Update upgrade script for fixed roles in APG17

anju15bharti:
### Description

Earlier we failed upgrade if fixed roles exist.

With this commit, we updated the script for fixed roles as:
* throw warning instead of exception for fixed server roles.
* check if expected fixed database roles exist, throw the warning else create fixed db roles.

### DIFF
--- a/contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--4.4.0--5.0.0.sql
+++ b/contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--4.4.0--5.0.0.sql
@@ -187,12 +187,35 @@ UPDATE sys.babelfish_authid_login_ext SET is_fixed_role = 1 WHERE rolname = 'sys
 
 -- At this point, there will be only one database fixed role which is db_owner.
 UPDATE sys.babelfish_authid_user_ext SET is_fixed_role = 1 WHERE orig_username = 'db_owner';
-UPDATE sys.babelfish_authid_user_ext SET is_fixed_role = 0 WHERE orig_username != 'db_owner';
 
 CREATE OR REPLACE PROCEDURE sys.create_db_roles_during_upgrade()
 LANGUAGE C
 AS 'babelfishpg_tsql', 'create_db_roles_during_upgrade';
-CALL sys.create_db_roles_during_upgrade();
+
+DO $$
+DECLARE
+    role_info record;
+BEGIN
+    WITH fixed_db_roles AS (
+        SELECT DISTINCT orig_username
+        FROM sys.babelfish_authid_user_ext
+        WHERE is_fixed_role = 1
+          AND orig_username IN ('db_owner', 'db_securityadmin', 'db_accessadmin', 
+                                'db_datareader', 'db_datawriter', 'db_ddladmin')
+    )
+    SELECT 
+        string_agg(orig_username, ', ') AS existing_db_roles,
+        COUNT(*) AS role_count
+    INTO role_info
+    FROM fixed_db_roles;
+
+    IF role_info.role_count = 1 THEN
+        CALL sys.create_db_roles_during_upgrade();
+    ELSE
+        RAISE WARNING 'The following fixed database role(s) already exist(s): %', role_info.existing_db_roles;
+    END IF;
+END $$;
+
 DROP PROCEDURE sys.create_db_roles_during_upgrade();
 
 DO
@@ -206,7 +229,7 @@ BEGIN
     WHERE rolname IN ('securityadmin', 'dbcreator');
         
     IF existing_server_roles IS NOT NULL THEN
-            RAISE EXCEPTION 'The following role(s) already exist(s): %', existing_server_roles;   
+            RAISE WARNING 'The following fixed server role(s) already exist(s): %', existing_server_roles;   
     ELSE
         EXECUTE format('CREATE ROLE securityadmin CREATEROLE INHERIT PASSWORD NULL');
         EXECUTE format('GRANT securityadmin TO bbf_role_admin WITH ADMIN TRUE');


contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--4.4.0--5.0.0.sql
```diff
@@ -206,7 +229,7 @@ BEGIN
     WHERE rolname IN ('securityadmin', 'dbcreator');
         
     IF existing_server_roles IS NOT NULL THEN
-            RAISE EXCEPTION 'The following role(s) already exist(s): %', existing_server_roles;   
+            RAISE WARNING 'The following fixed server role(s) already exist(s): %', existing_server_roles;
```
tanscorpio7, line 232: we should check for is_fixed_role for server roles as well.

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--4.4.0--5.0.0.sql
```diff
@@ -187,12 +187,35 @@ UPDATE sys.babelfish_authid_login_ext SET is_fixed_role = 1 WHERE rolname = 'sys
...
+
+DO $$
+DECLARE
+    role_info record;
+BEGIN
+    WITH fixed_db_roles AS (
+        SELECT DISTINCT orig_username
+        FROM sys.babelfish_authid_user_ext
+        WHERE is_fixed_role = 1
+          AND orig_username IN ('db_owner', 'db_securityadmin', 'db_accessadmin',
```
tanscorpio7, line 203: let us remove 'db_owner' from this list otherwise the error message would be misleading.

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--4.4.0--5.0.0.sql
```diff
@@ -187,12 +187,35 @@ UPDATE sys.babelfish_authid_login_ext SET is_fixed_role = 1 WHERE rolname = 'sys
...
+        string_agg(orig_username, ', ') AS existing_db_roles,
+        COUNT(*) AS role_count
+    INTO role_info
+    FROM fixed_db_roles;
+
+    IF role_info.role_count = 1 THEN
+        CALL sys.create_db_roles_during_upgrade();
+    ELSE
+        RAISE WARNING 'The following fixed database role(s) already exist(s): %', role_info.existing_db_roles;
+    END IF;
```
tanscorpio7, line 216: I think `role_count = 5` would be a more strict condition (excluding db_owner), plus lets RAISE NOTICE instead of warning.

tanscorpio7 (changes requested):

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--4.4.0--5.0.0.sql
anju15bharti: We only support fixed server role currently not user defined server role. So no need for that.

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--4.4.0--5.0.0.sql
anju15bharti: warning message will only come when other roles will exist where the it will show db_owner also exists along with other roles which is valid.

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--4.4.0--5.0.0.sql
anju15bharti: we want to call the proc if only is_fixed role count is 1 the ele condition will fall down to role_count = 6 since all the other roles are created at once.

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--4.4.0--5.0.0.sql
tanscorpio7: What if user created a login named 'securityadmin' or uses this name to create role from postgres endpoint ?

tanscorpio7 (changes requested):

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--4.4.0--5.0.0.sql
anju15bharti: I think we need to use bbf catalog in that case.

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--4.4.0--5.0.0.sql
```diff
@@ -187,12 +187,35 @@ UPDATE sys.babelfish_authid_login_ext SET is_fixed_role = 1 WHERE rolname = 'sys
 
 -- At this point, there will be only one database fixed role which is db_owner.
 UPDATE sys.babelfish_authid_user_ext SET is_fixed_role = 1 WHERE orig_username = 'db_owner';
-UPDATE sys.babelfish_authid_user_ext SET is_fixed_role = 0 WHERE orig_username != 'db_owner';
```
tanscorpio7, line 190: Lets set `is_fixed_role = 0` for every entry except fixed roles we support. We need this since in older versions of babelfish customer could execute DMLs on sys catalogs.

tanscorpio7 (approved):

anju15bharti:
Not required.

PR #3220, opened 2024-12-09
Modify upgrade scripts to obtain correct server level collation during upgrade

ahmed-shameem:
### Description
In all MVU sql scripts, we update the server collation name and reset the restored_server_collation_name GUC. During the upgrade from some version of 14 --> latest (15), the upgrade started failing for non-default server collation like chinese_prc_ci_as as we did not have correct collation information and the collation of columns of views were trying to get updated. This is because, we never updated the collation related information in any of the upgrade script post some version of 14, hence it was always picking the default value which is bbf_unicode_cp1_ci_as. 
The issue may come up in case when user is performing upgrade from some version of 14 to latest if we have defines some view which contains collatable column in upgrade script and the server level collation is different than default collation.

We address the issue by updating the collation related information in the necessary upgrade scripts. 

### Issues Resolved ###
BABEL-5415

* **Use case based -**

* **Boundary conditions -**

* **Arbitrary inputs -**

* **Negative test cases -**

* **Performance tests -**

* **Tooling impact -**

* **Client tests -**

### Check List

### DIFF
--- a/contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--3.0.0--3.1.0.sql
+++ b/contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--3.0.0--3.1.0.sql
@@ -31,6 +31,36 @@ end
 $$
 LANGUAGE plpgsql;
 
+CREATE OR REPLACE FUNCTION sys.babelfish_update_server_collation_name() RETURNS VOID
+LANGUAGE C
+AS 'babelfishpg_common', 'babelfish_update_server_collation_name';
+
+DO
+LANGUAGE plpgsql
+$$
+BEGIN
+    -- Check if the GUC is empty
+    IF current_setting('babelfishpg_tsql.restored_server_collation_name', true) <> '' THEN
+        -- Call the function to update the collation
+        EXECUTE 'SELECT sys.babelfish_update_server_collation_name()';
+    END IF;
+END;
+$$;
+
+DROP FUNCTION sys.babelfish_update_server_collation_name();
+
+-- reset babelfishpg_tsql.restored_server_collation_name GUC
+do
+language plpgsql
+$$
+    declare
+        query text;
+    begin
+        query := pg_catalog.format('alter database %s reset babelfishpg_tsql.restored_server_collation_name', CURRENT_DATABASE());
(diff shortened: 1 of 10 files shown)


contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--3.8.0--3.9.0.sql
```diff
@@ -10,6 +10,35 @@ SELECT set_config('search_path', 'sys, '||current_setting('search_path'), false)
  * So make sure that any SQL statement (DDL/DML) being added here can be executed multiple times without affecting
  * final behaviour.
  */
+CREATE OR REPLACE FUNCTION sys.babelfish_update_server_collation_name() RETURNS VOID
```
rishabhtanwar29, line 13: I think we require this only upon 3.3.0--3.4.0 isn't it?

Deepesh125 (approved):

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--3.0.0--3.1.0.sql
```diff
@@ -31,6 +31,36 @@ end
 $$
 LANGUAGE plpgsql;
 
+CREATE OR REPLACE FUNCTION sys.babelfish_update_server_collation_name() RETURNS VOID
```
Deepesh125, line 34: which upgrade will this be useful?

Deepesh125 (changes requested):

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--3.0.0--3.1.0.sql
ahmed-shameem, line 34: As discussed offline, we do not need it in 3_X at all. Because we only allow upgrade from 14.6 to 15 and so on. And the upgrade script from 2.3.0 (14.6) to 3.0.0 contains that info. So we do not need to add in any such upgrade script. Also, as Cx is in 3_X they can only to 4_X and beyond for which we have already updated the upgrade script.

Hence closing the PR.

PR #3225, opened 2024-12-09
Fix alter table drop column drops trigger on the table

tanscorpio7:
### Description

In postgres trigger and functions are two different objects. You first define a function and then use it in your trigger definition. In babelfish we handle this internally so that for users it feels like a single object. As part of this internal handing we also drop the function created for this trigger as part of post object access hook. Currently the condition for DROP TABLE was incomplete and even for DROP COLUMN we would drop the triggers associated with the table.
As a fix add another condition to only execute this internal drop of triggers for DROP TABLE.

### Issues Resolved

[BABEL-5417]

### Check List

### DIFF
--- a/contrib/babelfishpg_tsql/src/tablecmds.c
+++ b/contrib/babelfishpg_tsql/src/tablecmds.c
@@ -130,7 +130,7 @@ lookup_and_drop_triggers(ObjectAccessType access, Oid classId,
 		return;
 
 	/* We only want to execute this function for the DROP TABLE case */
-	if (classId != RelationRelationId || access != OAT_DROP)
+	if (classId != RelationRelationId || access != OAT_DROP || subId != 0)
 		return;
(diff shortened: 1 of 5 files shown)


contrib/babelfishpg_tsql/src/tablecmds.c
```diff
@@ -130,7 +130,7 @@ lookup_and_drop_triggers(ObjectAccessType access, Oid classId,
 		return;
 
 	/* We only want to execute this function for the DROP TABLE case */
-	if (classId != RelationRelationId || access != OAT_DROP)
+	if (classId != RelationRelationId || access != OAT_DROP || subId != 0)
```
forestkeeper, line 133: pls add a comment for the subId != 0

forestkeeper (approved):
LGTM

PR #3236, opened 2024-12-11
Fixed CASE statement to work correctly when branch expression is of Numeric and Decimal type

Yvinayak07:
### 1. Issue:

The CASE statement was unable to calculate the correct common typmod when the common type of the CASE statement was NUMERIC or DECIMAL. For CASE expressions, we relied on a PostgreSQL function to calculate the typmod when the common type was NUMERIC or DECIMAL. However, this approach showed different behavior from T-SQL. Therefore, we need to calculate the common typmod according to T-SQL documentation to ensure consistency with T-SQL behavior.

```
select case 1
    when 1 then cast(12 as DECIMAL(10,2))
    when 2 then cast(1.23 as numeric(10,5))
END
go

-- BBF
~~START~~
numeric
12.00
~~END~~

-- TSQL
~~START~~
decimal
12.00000
~~END~~
```

```
CREATE TABLE #pivotoutput
(
    [integer_one] INT,
    [integer_two] INT
);
GO

INSERT INTO #pivotoutput
(
    [integer_one],
    [integer_two]
)
VALUES
(1, 3);
GO

DECLARE @test_int int = 1

SELECT a.integer_one,
       a.integer_two,
       ratio = a.integer_one / CONVERT(DECIMAL, a.integer_two),
       ratio_case_one = CASE WHEN a.integer_two = 0 THEN NULL
                                   ELSE a.integer_one / CONVERT(DECIMAL, a.integer_two)
                              END,
       ratio_case_two = CASE WHEN 1 = 0 THEN NULL
                                   ELSE a.integer_one / CONVERT(DECIMAL, a.integer_two)
                              END,
       ratio_case_three = CASE WHEN @test_int = 0 THEN NULL
                                   ELSE a.integer_one / CONVERT(DECIMAL, a.integer_two)
                              END
FROM #pivotoutput a;
GO

-- BBF
~~START~~
integer_one integer_two ratio                                   ratio_case_one                           ratio_case_two                          ratio_case_three
----------- ----------- --------------------------------------- ---------------------------------------- --------------------------------------- ---------------------------------------
          1           3                   0.3333333333333333333                               0.33333333                   0.3333333333333333333                   0.3333333333333333333
~~END~~

-- TSQL
~~START~~
integer_one integer_two ratio                                   ratio_case_one                           ratio_case_two                          ratio_case_three
----------- ----------- --------------------------------------- ---------------------------------------- --------------------------------------- ---------------------------------------
          1           3                   0.3333333333333333333                               0.3333333333333333333                   0.3333333333333333333                   0.3333333333333333333
~~END~~
```

- The tsql_select_common_typmod_hook() function has been modified to return a common typmod for all branches when the branch expression is of NUMERIC or DECIMAL data types. This calculation now follows the T-SQL documentation. Additionally, we've added a 'case T_CoerceToDomain' in the 'resolve_numeric_typmod_from_exp' function to address issues related to user-defined types (UDTs).
- We have enhanced the 'resolve_numeric_typmod_from_exp' function by adding support for 'T_SubLink' and 'T_CoerceToDomain' nodes. This improvement enables accurate typmod calculation for expressions involving subqueries (T_SubLink) and user-defined types (T_CoerceToDomain). 

Authored-by: yashneet vinayak [yashneet@amazon.com](mailto:yashneet@amazon.com)

### Issues Resolved
Task: BABEL-5341

* **Use case based -**Yes

* **Boundary conditions -**

* **Arbitrary inputs -**Yes

* **Negative test cases -**

* **Minor version upgrade tests -**Yes

* **Major version upgrade tests -**Yes

* **Performance tests -**

* **Tooling impact -**

* **Client tests -**Yes

### Check List


contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
```diff
@@ -69,6 +69,17 @@
...
 #define NUMERIC_UMINUS_OID 1771
+#define INT4_NUMERIC 1740
+#define INT8_NUMERIC 1781
+#define INT2_NUMERIC 1782
+#define FLOAT4_NUMERIC 1742
+#define FLOAT8_NUMERIC 1743
+#define INT48 481
+#define INT84 480
+#define INT28 754
+#define INT82 714
```
KushaalShroff: Do we have test cases for them? Can you explain their need?

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
```diff
@@ -517,6 +527,40 @@ resolve_numeric_typmod_outer_var(Plan *plan, AttrNumber attno)
...
+								  Anum_pg_proc_prosrc, &isnull);
+	if (isnull)
+		elog(ERROR, "null prosrc");
+	proc_source = DatumGetCString(prosrcdatum);
+
+	if (func_oid == INT4_NUMERIC ||
+		func_oid == INT8_NUMERIC ||
+		func_oid == INT2_NUMERIC ||
+		func_oid == FLOAT4_NUMERIC ||
+		func_oid == FLOAT8_NUMERIC ||
```
KushaalShroff: Should we also consider these cast functions:
```
1> select typname from pg_type where oid in (select castsource from pg_cast where casttarget = 1700)
2> go
typname                                                                                                                                                                                                                                                         
int8                                                                                                                                                                                                                                                            
int2                                                                                                                                                                                                                                                            
int4                                                                                                                                                                                                                                                            
float4                                                                                                                                                                                                                                                          
float8                                                                                                                                                                                                                                                          
money                                                                                                                                                                                                                                                           
numeric                                                                                                                                                                                                                                                         
jsonb                                                                                                                                                                                                                                                           
fixeddecimal                                                                                                                                                                                                                                                    
varchar                                                                                                                                                                                                                                                         
bit                                                                                                                                                                                                                                                             
sql_variant 
```

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
```diff
@@ -517,6 +527,40 @@ resolve_numeric_typmod_outer_var(Plan *plan, AttrNumber attno)
 	return resolve_numeric_typmod_from_exp(outerplan, (Node *)tle->expr);
 }
 
+static bool
+is_convertable_to_numeric(Oid func_oid)
```
KushaalShroff: Lets add comments on top to describe the need of this util function

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
```diff
@@ -517,6 +527,40 @@ resolve_numeric_typmod_outer_var(Plan *plan, AttrNumber attno)
...
+{
+	HeapTuple	procTup;
+	Datum		prosrcdatum;
+	char	   *proc_source;
+	bool		isnull;
+
+	procTup = SearchSysCache1(PROCOID, ObjectIdGetDatum(func_oid));
+	if (!HeapTupleIsValid(procTup))
+		elog(ERROR, "cache lookup failed for function %u", func_oid);
```
KushaalShroff: 1. Lets do this conditionally since for all of the cast functions except money ones do not require lookup.
2. Even for looking up for money cast function, we can cache the oid in a static variable so a so to reuse it. For example:
```
static Oid fixeddecimal_numeric_oid = InvalidOid;

fixeddecimal_numeric_oid_func()
if (fixeddecimal_numeric_oid == InvalidOid):
     find the oid of fixeddecimal_numeric function
else return fixeddecimal_numeric_oid
```
and then we can use this helper function directly inside the if condition like:
```
func_oid == fixeddecimal_numeric_oid_func()
```
This will have 2 look ups, but thinking of larger picture where this function gets called n times then this approach gives us O(2) complexity rather than O(n)

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
```diff
@@ -560,14 +604,22 @@ resolve_numeric_typmod_from_exp(Plan *plan, Node *expr)
 			{
 				Const	   *con = (Const *) expr;
 				Numeric		num;
-
-				if (!is_numeric_datatype(con->consttype) || con->constisnull)
+				int64		val;
+				
+				if ((!(con->consttype == INT4OID) && !is_numeric_datatype(con->consttype)) ||
```
KushaalShroff, line 668: What about bigint, smallint, tinyint. Also bit?

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
```diff
@@ -560,14 +604,22 @@ resolve_numeric_typmod_from_exp(Plan *plan, Node *expr)
...
+					con->constisnull)
 				{
 					/* typmod is undefined */
 					return -1;
 				}
 				else
 				{
+					if (!plan && con->consttype == INT4OID)
```
KushaalShroff: Lets add comments for the conditions we have added

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
```diff
@@ -957,6 +1022,36 @@ resolve_numeric_typmod_from_exp(Plan *plan, Node *expr)
...
+				CoerceToDomain *rlt = (CoerceToDomain *) expr;
+
+				if (rlt->resulttypmod != -1)
+					return rlt->resulttypmod;
+				else
+					return resolve_numeric_typmod_from_exp(plan, (Node *) rlt->arg);
+			}
+		case T_SubLink:
+			{
+				const SubLink *sublink = (const SubLink *) expr;
```
KushaalShroff, line 1108: comment with /* Copied from exprTypmod. */

contrib/babelfishpg_tsql/src/pltsql_coerce.c
```diff
@@ -2019,36 +2026,107 @@ tsql_select_common_typmod_hook(ParseState *pstate, List *exprs, Oid common_type)
 			 !utilptr->is_tsql_binary_datatype(common_type) &&
 			 !utilptr->is_tsql_sys_binary_datatype(common_type) &&
 			 !utilptr->is_tsql_varbinary_datatype(common_type) &&
-			 !utilptr->is_tsql_sys_varbinary_datatype(common_type))
+			 !utilptr->is_tsql_sys_varbinary_datatype(common_type) &&
+			 !utilptr->is_tsql_decimal_datatype(common_type) &&
```
KushaalShroff, line 2029: Lets check the need of this

contrib/babelfishpg_tsql/src/pltsql_coerce.c
```diff
@@ -1999,16 +2001,21 @@ select_common_type_for_coalesce_function(ParseState *pstate, List *exprs)
...
 tsql_select_common_typmod_hook(ParseState *pstate, List *exprs, Oid common_type)
 {
-	int32		max_typmods=0;
+	int32		max_typmods=0,
+			max_precision = 0,
+			max_scale = 0,
+			precision = 0,
+			scale = 0,
+			integralDigitCount = 0,
+			result_typmod = -1;
```
KushaalShroff: Lets name this as numeric_result_typmod

contrib/babelfishpg_tsql/src/pltsql_coerce.c
```diff
@@ -2019,36 +2026,107 @@ tsql_select_common_typmod_hook(ParseState *pstate, List *exprs, Oid common_type)
...
-			Oid   base_type = getBaseTypeAndTypmod(type, &base_typmod);
+			/* If Udt then calculate typmod.*/
+			if (OidIsValid(immediate_base_type))
+			{
+				type = getBaseTypeAndTypmod(type, &typmod);
+			}
+			
+			if (typmod == -1 && (*pltsql_protocol_plugin_ptr))
+			{
+				Plan* temp = NULL;
```
KushaalShroff: Why do we need temp?

KushaalShroff (changes requested):

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
Yvinayak07: FOR:

1. bigint we have INT8_NUMERIC
2. tinyint  we have INT2_NUMERIC
3.smallint we have I4TOI2
4. bit we have is_tsql_bit_numeric

So these will be converted to int or numeric. and get handled in these conditions.

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
Yvinayak07: Added most of these that could be castable to numeric.

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
Yvinayak07: Added comment, resolved.

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
Yvinayak07: As suggested made helper functions for:

1. bit2numeric
2. fixeddecimal_numeric
3. numeric_fixeddecimal
4. int4bit

stored their Oid in a static variable.

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
Yvinayak07: added comment for explanation, resolved.

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
Yvinayak07: added comment /* Copied from exprTypmod. */ for both T_SubLink and T_CoerceToDomain.

contrib/babelfishpg_tsql/src/pltsql_coerce.c
Yvinayak07: We need decimal type, if all the branches are of decimal type then its common type is decimal.

contrib/babelfishpg_tsql/src/pltsql_coerce.c
Yvinayak07: renamed result_typmod to numeric_result_typmod

contrib/babelfishpg_tsql/src/pltsql_coerce.c
Yvinayak07: I used for testing purpose, removed the variable temp.

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
Yvinayak07: We have existing test cases for INT48 conversions. As a precautionary measure, we have included OIDs for all integer type casts to ensure comprehensive coverage.

Yvinayak07: Added comment on above of these OID's used.

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
```diff
@@ -786,6 +927,20 @@ resolve_numeric_typmod_from_exp(Plan *plan, Node *expr)
 					rettypmod = pltsql_plugin_handler_ptr->pltsql_read_numeric_typmod(func_oid,
 																					  func->args == NIL ? 0 : func->args->length,
 																					  func->funcresulttype);
+
+				/*
+				 * !plan means we are not invoking resolve_numeric_typmod_from_exp
+				 * from tds side. Here we make resursive call so as to calculate 
+				 * typmod from other base nodes in parse tree.
+				 */
+				if (!plan &&
```
rohit01010: NIT: rewrite `!plan` to `plan == NULL`

contrib/babelfishpg_tds/src/include/tds_response.h
```diff
@@ -98,5 +98,7 @@ extern void TDSStatementExceptionCallback(PLtsql_execstate *estate, PLtsql_stmt
 										  bool terminate_batch);
 extern void SendColumnMetadata(TupleDesc typeinfo, List *targetlist, int16 *formats);
 extern bool GetTdsEstateErrorData(int *number, int *severity, int *state);
+extern int32 resolve_numeric_typmod_from_exp(Plan *plan, Node *expr);
+extern bool is_numeric_cast(Oid func_oid);
```
rohit01010: if is `is_numeric_cast` function is not being used in anywhere else, kindly mark it as static.

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
```diff
@@ -786,6 +927,20 @@ resolve_numeric_typmod_from_exp(Plan *plan, Node *expr)
 					rettypmod = pltsql_plugin_handler_ptr->pltsql_read_numeric_typmod(func_oid,
 																					  func->args == NIL ? 0 : func->args->length,
 																					  func->funcresulttype);
+
+				/*
+				 * !plan means we are not invoking resolve_numeric_typmod_from_exp
+				 * from tds side. Here we make resursive call so as to calculate 
+				 * typmod from other base nodes in parse tree.
+				 */
```
rohit01010: kindly update the comment to explain the checks in the if condition.

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
```diff
@@ -577,7 +719,7 @@ resolve_numeric_typmod_from_exp(Plan *plan, Node *expr)
 				Var		   *var = (Var *) expr;
 
 				/* If this var referes to tuple returned by its outer plan then find the original tle from it */
-				if (var->varno == OUTER_VAR)
+				if (plan && var->varno == OUTER_VAR)
```
rohit01010: How are we ensuring that `if plan is NULL then it is called from TSQL and not from TDS` ?

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
```diff
@@ -560,14 +686,30 @@ resolve_numeric_typmod_from_exp(Plan *plan, Node *expr)
...
+					 * converts the INT4 value to NUMERIC and then determines 
+					 * the appropriate typmod. This process ensures correct 
+					 * numeric precision handling in Babelfish TSQL operations.
+					 */
+					if (!plan && con->consttype == INT4OID)
+					{
+						val = con->constvalue;
+						num = int64_to_numeric(val);
+						return numeric_get_typmod(num);
+					}
```
rohit01010: When plan is `NOT NULL` i.e. function is not called from TSQL extension then shouldn't we return `-1`, as we use to do previously ?

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
```diff
@@ -517,6 +541,108 @@ resolve_numeric_typmod_outer_var(Plan *plan, AttrNumber attno)
...
 
+static bool is_tsql_bit_numeric(Oid oid)
+{
+	if (tsql_bit_numeric_oid == InvalidOid)
+	{
+		Oid			nspoid;
+		Oid			typoid;
+		Oid 		funcargtypes[1];
+
+		nspoid = get_namespace_oid("sys", true);
```
rohit01010: we should keep the second parameter as `false` which will ensure that if `get_namespace_oid` function is unable to find oid of `sys` then we should throw error.

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
```diff
@@ -517,6 +541,108 @@ resolve_numeric_typmod_outer_var(Plan *plan, AttrNumber attno)
...
+		if (nspoid == InvalidOid)
+			return InvalidOid;
+
+		typoid = GetSysCacheOid2(TYPENAMENSP, Anum_pg_type_oid, CStringGetDatum("bit"), ObjectIdGetDatum(nspoid));
+
+		funcargtypes[0] = typoid;
+		tsql_bit_numeric_oid = LookupFuncName(list_make2(makeString("sys"), makeString("bit2numeric")), 1, funcargtypes, true);
```
rohit01010: should we also consider setting the `missing_ok` parameter value as `false` ?

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
```diff
@@ -70,6 +70,19 @@
...
+#define INT8_NUMERIC 1781
+#define INT2_NUMERIC 1782
+#define FLOAT4_NUMERIC 1742
+#define FLOAT8_NUMERIC 1743
+#define INT48 481
+#define INT84 480
+#define INT28 754
+#define INT82 714
+#define I2TOI4 313
+#define I4TOI2 314
```
rohit01010: These Oids are already defined in `fmgroids.h` header file, we can directly use those constants.

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
Deepesh125: what if OID gets changed tomorrow due to addition of cast or other object?

test/JDBC/expected/BABEL-CASE_EXPR-vu-verify.out
```diff
@@ -13014,3 +13014,1522 @@ GO
...
+GO
+
+CREATE TYPE SmallDecimalType_BABEL_5341 FROM DECIMAL(5,2);
+GO
+
+-- Test Case 45: Basic CASE with all UDTs
+SELECT CASE 4
+   WHEN 1 THEN CAST(12345.6789 AS NumericType_BABEL_5341)
+   WHEN 2 THEN CAST(98765.4321 AS DecimalType_BABEL_5341)
+   WHEN 3 THEN CAST(1234567.8901234567890 AS BigNumericType_BABEL_5341)
```
rohit01010, line 13786: kindly use 18 digits before decimal for value with type `BigNumericType_BABEL_5341`, so result precision can also be verified.

rohit01010 (changes requested):

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
Deepesh125: you can cache this by making it static/extern and use it everywhere

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
Deepesh125: missing_ok is passed as true which means it may return invalid oid. We should detect that and throw an error. IMO, we should pass missing-ok = false

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
Deepesh125: what is meaning of these INT48, INT84 etc?

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
Deepesh125: IIRC, const node can have int4, numeric, bool or unknown/text oid. so we probably need to think about bool/bit for tsql

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
```diff
@@ -786,6 +927,20 @@ resolve_numeric_typmod_from_exp(Plan *plan, Node *expr)
 					rettypmod = pltsql_plugin_handler_ptr->pltsql_read_numeric_typmod(func_oid,
 																					  func->args == NIL ? 0 : func->args->length,
 																					  func->funcresulttype);
+
+				/*
+				 * !plan means we are not invoking resolve_numeric_typmod_from_exp
+				 * from tds side. Here we make resursive call so as to calculate
```
Deepesh125: can you please explain meaning of "tds side"?

test/JDBC/input/NUMERIC_TYPMOD.sql
```diff
@@ -0,0 +1,1231 @@
...
+-- TODO FIX: SQRT function returns float in TSQL. while
+-- selecting commontype between flaot ansd numeric,
+-- is selected. where as in BBF SQRT function returns numeric.
+-- Test Case: T_SubPlan with different numeric operations
+SELECT CASE 
+    WHEN (SELECT POWER(CAST(2 AS NUMERIC(5,2)), 3)) > 7
+    THEN (SELECT SQRT(CAST(100 AS NUMERIC(10,4))))
+    ELSE (SELECT CAST(PI() AS DECIMAL(10,8)))
+END AS result;
+GO
```
Deepesh125, line 1231: these test cases are simple in terms of typmod. can you add cases where we can test boundaries of typmod? (38,0), (38,38), etc?

test/JDBC/input/BABEL-CASE_EXPR-vu-prepare.sql
```diff
@@ -45,4 +45,112 @@ BEGIN
...
+
+INSERT INTO BABEL_5341_T1 (COL1, COL2, COL3, COL4, COL5, COL6)
+VALUES (33333.33, 44444.44, 3333333333333333333, 4444444444444444444, 33, 44);
+GO
+
+INSERT INTO BABEL_5341_T1 (COL1, COL2, COL3, COL4, COL5, COL6)
+VALUES (55555.55, 66666.66, 5555555555555555555, 6666666666666666666, 55, 66);
+GO
+
+CREATE VIEW BABEL_5341_V1 AS
```
Deepesh125, line 80: in verify part, we should check the precision, scale of the column using some system metadata views

Deepesh125 (changes requested):

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
Yvinayak07: changed to plan == NULL. Resolved.

contrib/babelfishpg_tds/src/include/tds_response.h
Yvinayak07: Marked as static. resolved.

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
Yvinayak07: Updated the comments to explain this condition point wise.

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
Yvinayak07: Yes, we need to add that condition we missed that here. added this condition here resolved.

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
Yvinayak07: changed the parameter to false so as we can throw error if oid is not found.

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
Yvinayak07: changed the parameter to false so as we can throw error if oid is not found.

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
Yvinayak07: Picked Oid's from fmgroids.h so if Oid's change in future we need to change at one place.

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
Yvinayak07: We have picked the Oid's from fmgroids.h so in future if Oid's change we need to make change at one place. If we do lookup from catalog it will increase time.

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
Yvinayak07: Made a static variable sys_nspoid for getting oid for sys.

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
Yvinayak07: changed the parameter 'missing_ok' to false so as we can throw error if oid is not found.

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
Yvinayak07: When 'plan' is NULL, it means we're calling this function from the Babelfish TSQL extension not tds( Resolved with correct comment). In this case, we don't have a query plan to work with.

test/JDBC/input/NUMERIC_TYPMOD.sql
Yvinayak07: Added in BABEL-CASE_EXPR-verify file:
- https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/3236/files#diff-d8342fa5f1c401e85f0a2c76f761031caed3e894ea483b9c12b53c8b2a20b098R13619
- https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/3236/files#diff-d8342fa5f1c401e85f0a2c76f761031caed3e894ea483b9c12b53c8b2a20b098R14506

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
Yvinayak07: - We have picked the Oid's from fmgroids.h so in future if Oid's change we need to make change at one place. If we do lookup from catalog it will increase time.

- INT48 and INT84 are casting functions. INT48 represents a cast from INT4 (4-byte integer) to INT8 (8-byte integer), while INT84 represents the reverse cast from INT8 to INT4.

test/JDBC/expected/BABEL-CASE_EXPR-vu-verify.out
Yvinayak07:
```
-- BBF After Our build.
SELECT CASE 4
   WHEN 1 THEN CAST(12345.6789 AS NUMERIC(20,2))
   WHEN 2 THEN CAST(98765.4321 AS DECIMAL(18,4))
   WHEN 3 THEN CAST(1234567.8901234567890 AS NUMERIC(38,10))
   WHEN 4 THEN CAST(99.99 AS DECIMAL(5,2))
   ELSE CAST(0 AS NUMERIC(20,2))
END AS RESULT

RESULT                                  
----------------------------------------
                           99.9900000000
```

```
-- SQLserver 
SELECT CASE 4
   WHEN 1 THEN CAST(12345.6789 AS NUMERIC(20,2))
   WHEN 2 THEN CAST(98765.4321 AS DECIMAL(18,4))
   WHEN 3 THEN CAST(1234567.8901234567890 AS NUMERIC(38,10))
   WHEN 4 THEN CAST(99.99 AS DECIMAL(5,2))
   ELSE CAST(0 AS NUMERIC(20,2))
END AS RESULT

RESULT                                  
----------------------------------------
                           99.9900000000

```

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
Yvinayak07: - When a expression is of bit it enter T_Funcexpr and call it recursively with bit2numeric and solve the typmod.
```
test case: 

SELECT CASE 
    WHEN 1 = 1 THEN 
        CASE 
            WHEN 1=1
            THEN CAST(1234.56 AS NUMERIC(10,2)) * cast(1 as bit)
            ELSE CAST(9876.54 AS numeric(12,2)) / cast(1 as bit)
        END
    ELSE CAST(0 AS DECIMAL(12,2))
END AS result;
```

-  The bit data type is used for storing Boolean values.

test/JDBC/input/BABEL-CASE_EXPR-vu-prepare.sql
Yvinayak07: Added test cases related to this.

rohit01010 (approved):

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
Deepesh125: does it mean we will have to update IF condition whenever we introduce new cast (xyz -> numeric)? it will have maintenance issue

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
```diff
@@ -786,6 +910,21 @@ resolve_numeric_typmod_from_exp(Plan *plan, Node *expr)
 					rettypmod = pltsql_plugin_handler_ptr->pltsql_read_numeric_typmod(func_oid,
 																					  func->args == NIL ? 0 : func->args->length,
 																					  func->funcresulttype);
+
+				/*
+				 * If the following conditions are met then we will recursively find typmod from arg.
+				 * 1) plan == NULL means we are invoking this function from babelfishpg_tsql extension.
```
Deepesh125: or we can say, during parsing phase

Deepesh125 (dismissed):

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
Yvinayak07: updated the comment. Resolved

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
```diff
@@ -545,18 +534,13 @@ static bool is_tsql_bit_numeric(Oid oid)
...
-		Oid			nspoid;
 		Oid			typoid;
 		Oid 		funcargtypes[1];
 
-		nspoid = get_namespace_oid("sys", true);
-		if (nspoid == InvalidOid)
-			return InvalidOid;
-
-		typoid = GetSysCacheOid2(TYPENAMENSP, Anum_pg_type_oid, CStringGetDatum("bit"), ObjectIdGetDatum(nspoid));
+		typoid = GetSysCacheOid2(TYPENAMENSP, Anum_pg_type_oid, CStringGetDatum("bit"), ObjectIdGetDatum(sys_nspoid));
```
KushaalShroff: For sanity we should always check if (sys_nspoid == InvalidOid) and assign it before using it.

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
```diff
@@ -527,24 +541,87 @@ resolve_numeric_typmod_outer_var(Plan *plan, AttrNumber attno)
...
+		Oid 		funcargtypes[1];
+
+		nspoid = get_namespace_oid("sys", true);
+		if (nspoid == InvalidOid)
+			return InvalidOid;
+
+		typoid = GetSysCacheOid2(TYPENAMENSP, Anum_pg_type_oid, CStringGetDatum("fixeddecimal"), ObjectIdGetDatum(nspoid));
+
+		funcargtypes[0] = typoid;
+		tsql_fixeddecimal_numeric_oid = LookupFuncName(list_make2(makeString("sys"), makeString("fixeddecimal_numeric")), 1, funcargtypes, true);
```
KushaalShroff: Can we lookup in pg_cast instead?
```
	tuple = SearchSysCache2(CASTSOURCETARGET,
							ObjectIdGetDatum(typeId),
							ObjectIdGetDatum(typeId));

	if (HeapTupleIsValid(tuple))
	{
		Form_pg_cast castForm = (Form_pg_cast) GETSTRUCT(tuple);

		*funcid = castForm->castfunc;
		ReleaseSysCache(tuple);
	}
```
Seems cleaner and faster

KushaalShroff (changes requested):

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
Yvinayak07: Made this sanity check.

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
Yvinayak07: Lookup done in pg_cast for faster access.

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
```diff
@@ -530,68 +530,56 @@ resolve_numeric_typmod_outer_var(Plan *plan, AttrNumber attno)
 	return resolve_numeric_typmod_from_exp(outerplan, (Node *)tle->expr);
 }
 
-static bool is_tsql_bit_numeric(Oid oid)
+static Oid
+func_oid_of_cast_from_source_target(Oid castsource, Oid casttarget)
```
KushaalShroff: Why cant we have the same naming convention: LookupCastFuncName

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
```diff
@@ -517,6 +530,97 @@ resolve_numeric_typmod_outer_var(Plan *plan, AttrNumber attno)
...
+ * This information is used when processing T_FuncExpr nodes to determine
+ * if resolve_numeric_typmod_from_exp should be called recursively.
+ * This ensures proper typmod resolution for nested numeric conversions.
+ */
+static bool
+is_numeric_cast(Oid func_oid)
+{
+	sys_nspoid = get_namespace_oid("sys", false);
+
+	if (OidIsValid(sys_nspoid))
```
KushaalShroff: By my comment earlier I didnt mean to add the valid check like this. Here it would be obvious that sys_nspoid is valid.

What I meant was coding it this way:
```
if (!OidIsValid(sys_nspoid))
	sys_nspoid = get_namespace_oid("sys", false);
```

KushaalShroff (changes requested):

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
Yvinayak07: Replaced func_oid_of_cast_from_source_target -> LookupCastFuncName.

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
Yvinayak07: Corrected, I thought it other way.
Resolved

KushaalShroff (approved):

PR #3259, opened 2024-12-15
Fix hashbytes behaviour for nvarchar input

pranavJ23:
### Description
PROBLEM: while having nvarchar input in hashbytes we were considering the encoding as input encoding in babelfish where as in tsql we use UTF16 encoding for nvarchar irrespective of input encoding.

RCA: we were considering varchar and nvarchar as same, whereas we should use input encoding for varchar and UTF16 encoding for nvarchar.

FIX: So we need to identify that if the input is nvarchar then we will do the UTF16 encoding.
In the function hashbytes, we were converting varchar/nvarchar to varbinary, so if we get input as nvarchar then we will first convert the input string to UTF16 encoding then convert to varbinary.
There we created the overloaded function and also select the best candidate among depending on the second argument(i.e varchar/nvarchar/varbinary).

### Issues Resolved
[BABEL-4891]

### Check List


contrib/babelfishpg_tsql/src/string.c
```diff
@@ -56,11 +56,29 @@ static Datum return_varchar_pointer(char *buf, int size);
...
+    /*
+     * If the input_type is nvarchar then we convert it to UTF-16 encoding
+     */
+    if(((*common_utility_plugin_ptr->is_tsql_nvarchar_datatype)(input_type)))
+    {
+        unsigned char     *temp_data;
+        initStringInfo(&utf16_data);
+        (common_utility_plugin_ptr->TsqlUTF8toUTF16StringInfo)(&utf16_data, data, len);
+        len = utf16_data.len;
+        temp_data = (unsigned char *)palloc(len);
```
shah-nirmit: why we need temp_data separate?

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--4.4.0--4.5.0.sql
```diff
@@ -10,6 +10,45 @@ SELECT set_config('search_path', 'sys, '||current_setting('search_path'), false)
  * So make sure that any SQL statement (DDL/DML) being added here can be executed multiple times without affecting
  * final behaviour.
  */
+
+DO $$
+DECLARE
+    exception_message text;
+BEGIN
+    ALTER FUNCTION sys.hashbytes(IN alg VARCHAR, IN data VARCHAR) RENAME TO hashbytes_varchar_deprecated_4_5_0;
```
Deepesh125: do we really need to deprecate this function? what if some cx objects are dependent on this and they are upgrading their database to new version (major version upgrade)?

contrib/babelfishpg_tsql/src/pltsql_coerce.c
```diff
@@ -1308,6 +1403,12 @@ tsql_func_select_candidate_for_special_func(List *names, List *fargs, int nargs,
 
 	DeconstructQualifiedName(names, &proc_nsname, &proc_name);
 
+	/* Specific handling for hashbytes function based on second input argument via tsql_func_select_candidate_for_hashbytes() */
+	if ((proc_nsname == NULL || strcmp(proc_nsname, "sys") == 0) && strcmp(proc_name, "hashbytes") == 0)
```
Deepesh125: what if input is not schema qualified?

contrib/babelfishpg_tsql/src/pltsql_coerce.c
```diff
@@ -1288,6 +1348,41 @@ validate_special_function(char *func_nsname, char *func_name, List* fargs, int n
...
+
+    /* Get the candidate with matching second argument type */
+    best_candidate = NULL;
+    for (current_candidate = candidates;
+        current_candidate != NULL;
+        current_candidate = current_candidate->next)
+    {
+        /* we should only consider candidates for hashbytes function from sys schema */
+        if (get_func_namespace(current_candidate->oid) != sys_oid)
+            return NULL;
```
Deepesh125: there could be sys.hashbytes available in next iteration

contrib/babelfishpg_tsql/src/pltsql_coerce.c
```diff
@@ -1288,6 +1348,41 @@ validate_special_function(char *func_nsname, char *func_name, List* fargs, int n
...
+
+        get_func_signature(current_candidate->oid,&argtypes, &nargs_func);
```
Deepesh125: space before agruments

contrib/babelfishpg_tsql/src/pltsql_coerce.c
```diff
@@ -1288,6 +1348,41 @@ validate_special_function(char *func_nsname, char *func_name, List* fargs, int n
...
+        if(input_typeids[1] == argtypes[1])
+        {
+            best_candidate = current_candidate;
```
Deepesh125: we found it, should we get out of the loop?

contrib/babelfishpg_tsql/src/string.c
```diff
@@ -56,11 +56,29 @@ static Datum return_varchar_pointer(char *buf, int size);
...
+        memcpy(temp_data,utf16_data.data,len);
+        data = temp_data;
+        pfree(utf16_data.data);
```
Deepesh125: why cant we directly do data = utf16_data.data?

Deepesh125 (changes requested):

contrib/babelfishpg_tsql/src/pltsql_coerce.c
pranavJ23: But since user has defined their own function hashbytes, so we don't need to look/wait for any sys.hashbytes function.

contrib/babelfishpg_tsql/src/pltsql_coerce.c
pranavJ23: then proc_nsname will be NULL, right?
If it is null and function is hashbytes then we will go inside the condition 
like select hashbytes('sha1','test string') does not have any schema, hence proc_nsname will be null

contrib/babelfishpg_tsql/src/string.c
pranavJ23: It will throw error while pfree, since data is a pointer and if we free utf16_data then data will also be allocated with a garbage value

contrib/babelfishpg_tsql/src/pltsql_coerce.c
pranavJ23: but we need to do best_candidate->next = NULL too, so directly returning might not address that issue

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--4.4.0--4.5.0.sql
pranavJ23: we are renaming the object, right? 
 Not deprecating the function

test/JDBC/expected/cast_nvarchar_test-vu-prepare.out
```diff
@@ -1,14 +1,13 @@
+CREATE TABLE TestHash(
```
Deepesh125, line 1: can we have some tests where we create userdefined hashbytes and create view/func/proc which is dependent on that?

Deepesh125 (changes requested):

contrib/babelfishpg_tsql/src/pltsql_coerce.c
shah-nirmit: +1

contrib/babelfishpg_tsql/src/string.c
pranavJ23: It will throw error while pfree, since data is a pointer and if we free utf16_data then data will also be allocated with a garbage value

contrib/babelfishpg_tsql/src/string.c
Deepesh125: I am not getting this part. Why would it throw error using pfree? You can directly use utf16_data and then pfree once you dont need it. Adding additional palloc and memcpy will cause performance penalty

contrib/babelfishpg_tsql/src/pltsql_coerce.c
```diff
@@ -1349,13 +1373,22 @@ tsql_func_select_candidate_for_special_func(List *names, int nargs, Oid *input_t
...
+			second_arg_type = argtypes[1];
+		}
 		
 		/* Ignore following definitions as these are used when no other potential definition can be used. */
 		if ((current_candidate->args[0] == TEXTOID && rettype == get_sys_varcharoid())
 			|| (current_candidate->args[0] == BYTEAOID && rettype == BYTEAOID))
 			continue;
-
-		if (expr_result_type == rettype)
+		/* find the best candidate based on second_arg_type(this will be valid only for the case of hasbytes)
```
Deepesh125: use postgres style of comment

contrib/babelfishpg_tsql/src/pltsql_coerce.c
```diff
@@ -1370,9 +1403,15 @@ tsql_func_select_candidate_for_special_func(List *names, int nargs, Oid *input_t
...
 	{
-		ereport(ERROR,
-			(errcode(ERRCODE_INTERNAL_ERROR),
-				errmsg("multiple definitions of function %s.%s with return type %s found.", proc_nsname, proc_name, format_type_be(expr_result_type))));
+		if(OidIsValid(expr_result_type))
+			ereport(ERROR,
+				(errcode(ERRCODE_INTERNAL_ERROR),
+					errmsg("multiple definitions of function %s.%s with return type %s found.", proc_nsname, proc_name, format_type_be(expr_result_type))));
+
+		else if(OidIsValid(expr_arg_type))
```
Deepesh125: space after if

contrib/babelfishpg_tsql/src/pltsql_coerce.c
```diff
@@ -1370,9 +1403,15 @@ tsql_func_select_candidate_for_special_func(List *names, int nargs, Oid *input_t
...
+			ereport(ERROR,
+				(errcode(ERRCODE_INTERNAL_ERROR),
+					errmsg("multiple definitions of function %s.%s with second arg type %s found.", proc_nsname, proc_name, format_type_be(expr_arg_type))));
```
Deepesh125: how can it run into this error? can you give an example?

contrib/babelfishpg_tsql/src/string.c
```diff
@@ -56,11 +56,29 @@ static Datum return_varchar_pointer(char *buf, int size);
 Datum
 hashbytes(PG_FUNCTION_ARGS)
 {
-	const char *algorithm = text_to_cstring(PG_GETARG_TEXT_P(0));
-	bytea	   *in = PG_GETARG_BYTEA_PP(1);
-	size_t		len = VARSIZE_ANY_EXHDR(in);
-	const uint8 *data = (unsigned char *) VARDATA_ANY(in);
-	bytea	   *result;
+    Oid            input_type = get_fn_expr_argtype(fcinfo->flinfo,1);
+    const char *algorithm = text_to_cstring(PG_GETARG_TEXT_P(0));
```
Deepesh125: where is it being freed? and why const char *?

contrib/babelfishpg_tsql/src/string.c
```diff
@@ -56,11 +56,29 @@ static Datum return_varchar_pointer(char *buf, int size);
...
+    bytea	   *in = PG_GETARG_BYTEA_PP(1);
+    size_t		len = VARSIZE_ANY_EXHDR(in);
+    bytea	   *result;
+    unsigned char     *data = (unsigned char *) VARDATA_ANY(in);
+    StringInfoData     utf16_data;
```
Deepesh125: plz fix indentation

Deepesh125 (changes requested):

contrib/babelfishpg_tsql/src/pltsql_coerce.c
pranavJ23: yes, my fault. It will not run into this error.

contrib/babelfishpg_tsql/src/pltsql_coerce.c
```diff
@@ -1329,12 +1334,31 @@ tsql_func_select_candidate_for_special_func(List *names, int nargs, Oid *input_t
...
+				|| (*common_utility_plugin_ptr->is_tsql_text_datatype)(new_input_typeids[1])
+				|| new_input_typeids[1] == UNKNOWNOID)
+		{
+			expr_arg_type = get_sys_varcharoid();
+		}
+		else if((*common_utility_plugin_ptr->is_tsql_nvarchar_datatype)(new_input_typeids[1])
+				|| (*common_utility_plugin_ptr->is_tsql_nchar_datatype)(new_input_typeids[1]))
+		{
+			expr_arg_type = (*common_utility_plugin_ptr->lookup_tsql_datatype_oid) ("nvarchar");
+		}
```
rohit01010: what will be the behaviour when argument type is `ntext` ?

contrib/babelfishpg_tsql/src/pltsql_coerce.c
```diff
@@ -1329,12 +1334,31 @@ tsql_func_select_candidate_for_special_func(List *names, int nargs, Oid *input_t
...
 
 	/* free new_input_typeids, as they are no longer needed */
 	if (new_input_typeids)
 		pfree(new_input_typeids);
 
-	if (!OidIsValid(expr_result_type))
+	if (!OidIsValid(expr_result_type) && !OidIsValid(expr_arg_type))
 		return NULL;
 
 	/* Get the candidate with matching return type */
```
rohit01010: Now since we are also selecting candidate with argument type, we should update this comment.

contrib/babelfishpg_tsql/src/string.c
pranavJ23: I have not freed the data,since we are not doing any palloc for data

contrib/babelfishpg_tsql/src/string.c
```diff
@@ -56,36 +56,34 @@ static Datum return_varchar_pointer(char *buf, int size);
...
     bytea	   *result;
-    unsigned char     *data = (unsigned char *) VARDATA_ANY(in);
-    StringInfoData     utf16_data;
+    StringInfoData	utf16_data;
 
 
     /*
      * If the input_type is nvarchar then we convert it to UTF-16 encoding
      */
+	initStringInfo(&utf16_data);
```
Deepesh125, line 68: indentation

contrib/babelfishpg_tsql/src/string.c
```diff
@@ -56,36 +56,34 @@ static Datum return_varchar_pointer(char *buf, int size);
...
     {
-        unsigned char     *temp_data;
-        initStringInfo(&utf16_data);
         (common_utility_plugin_ptr->TsqlUTF8toUTF16StringInfo)(&utf16_data, data, len);
         len = utf16_data.len;
-        temp_data = (unsigned char *)palloc(len);
-        memcpy(temp_data,utf16_data.data,len);
-        data = temp_data;
-        pfree(utf16_data.data);
+		data = (const uint8 *)utf16_data.data;
```
Deepesh125: indentation

contrib/babelfishpg_tsql/src/pltsql_coerce.c
Deepesh125: this is not resolved

Deepesh125 (dismissed):

contrib/babelfishpg_tsql/src/pltsql_coerce.c
pranavJ23: Tsql give invalid argument with ntext and text datatype, so I think we don't need to be worried about ntext. But just to be on safer side added ntext and text cases too

rohit01010 (approved):

Deepesh125 (approved):

PR #3291, opened 2024-12-20
Performance Improvement of create login in BBF

anju15bharti:
### Description

Engine PR : https://github.com/babelfish-for-postgresql/postgresql_modified_for_babelfish/pull/507

In Major version 16 Babelfish introduced a new role bbf_admin_role which is a member of all the roles which created implicitly from TSQL. This increases the Load on any Grant/Revoke operations that happen since in BBF to "select the role through which it gets permission to current user to grant", is inefficient and loops over members of members to find the all members of the role even though the required role is find.
This commit optimizes the selection of role with is_admin_option true for babelfish by escaping the search for any TSQL roles conditionally.
Without the changes the time taken to create a login increases drastically and linearly:

Flame graph : 
kernel

The average time consumed by create login stmt(with 125 dbs and 125 logins present) : 12836 ms
with these changes we see the time taken to create login increasing linearly but not as drastically as it was before the changes:

Flame graph : 
perf-kernel-new-hook

The average time consumed by create login stmt(with 125 dbs and 125 logins present) : 49 ms

### Check List


test/JDBC/expected/db_securityadmin-vu-verify.out
```diff
@@ -1312,6 +1312,15 @@ t
 ~~END~~
 
 
+-- Wait to sync with another session
```
tanscorpio7, line 1315: better to use the terminate conn flag

tanscorpio7 (changes requested):

test/JDBC/input/test_windows_login-vu-prepare.mix
```diff
@@ -1,5 +1,4 @@
--- sla 180000
--- TODO: BABEL-5349
+-- sla 55000
```
tanscorpio7, line 1: let us also merge this fix in 17.2 & 16.7. This will help us skip creating before and after files for this test case.

tanscorpio7 (changes requested):

test/JDBC/expected/db_securityadmin-vu-verify.out
anju15bharti: Since we have terminated the process from psql endpoint already in this testfile just waiting time was missing. So I think for terminating from psql endpoint, we can go with this fix for now and if we want to fix this thoroughly, respective author can do it later.

test/JDBC/input/test_windows_login-vu-prepare.mix
anju15bharti: We will still need to create before file for version before 16.7 as it was failing for 16.1 and some other versions as well.
This commit can be cherry-picked into 16.7 and 17.2 branch as well.

tanscorpio7 (approved):

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -6287,3 +6288,49 @@ get_tvp_typename_typeschemaname(char *proc_name, char *target_arg_name, char **t
...
+		HeapTuple	tup = &memlist->members[i]->tuple;
+		Form_pg_auth_members form = (Form_pg_auth_members) GETSTRUCT(tup);
+	
+		/* Return true if admin_option is true */
+		if (form->admin_option)
+		{
+			ReleaseSysCacheList(memlist);
+			return true;
+		}
+	}
```
HarshLunagariya, line 6332: Here, we are relying on the fact that bbf_role_admin is always be the direct member of every TSQL login/user/role. Is there a possibility to have an exception?
One suggestion to add testcase which validates all the role entries of babelfish_authid_user_ext and babelfish_authid_login_ext to validate whether they are member of bbf_role_admin with admin option or not.

HarshLunagariya (changes requested):

contrib/babelfishpg_tsql/src/catalog.c
anju15bharti: It can be possible if do some grants from psql endpoint but that is then interoperability case.
And even in this case, it will fall back to earlier path of iterating through all of it's member.

We rely on the fact that in bbf we grant all roles to bbf_role_admin. Added testcase explicitly to validate the same.

HarshLunagariya (approved):

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -6287,3 +6288,49 @@ get_tvp_typename_typeschemaname(char *proc_name, char *target_arg_name, char **t
...
+ * Checks if the given member is 'bbf_role_admin' role and has the privilege 
+ * to grant a specified role through direct membership. This privilege is 
+ * determined by the presence of the 'admin_option' flag, which allows the 
+ * member to grant the role to others. 
+ * 
+ * @return true if the member is 'bbf_role_admin' and also has the admin option 
+ *         to grant the specified role; otherwise, returns false.
+ */
+bool
+bbf_check_member_has_direct_priv_to_grant_role(Oid member, Oid role)
```
surendravishnoi, line 6302: this hook is critical as it overrides admin role membership. Please make sure we have test case for each use case which should return true and false from this function.

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -6287,3 +6288,49 @@ get_tvp_typename_typeschemaname(char *proc_name, char *target_arg_name, char **t
...
+	 */
+	if (sql_dialect != SQL_DIALECT_TSQL || !IS_TDS_CONN()
+		|| member !=  get_bbf_role_admin_oid())
+		return false;
+
+	/* 
+	 * Find all existing tuples for given role
+	 * whose member is bbf_role_admin
+	 */
+	memlist = SearchSysCacheList2(AUTHMEMROLEMEM,
```
surendravishnoi, line 6318: we are checking two things
1) bbf admin is member of "role"
2) membership is with admin option ON

do we have a use case where this not true for bbf_admin and bbf_roles? if not, it might be sufficient to check for role being bbf_role (assuming that check is very fast to do)

anju15bharti: No, for every bbf_roles, these conditions will be true. I checked `is_babelfish_role` API is costlier than the current approach since it uses `roles_is_member_of` API internally to find all memberships.

contrib/babelfishpg_tsql/src/catalog.c
anju15bharti: This hook will always return true for valid scenario if member is bbf_role_admin and dialect is tsql and In BBF for all scenarios through which we hit this hook(i.e create/alter role and create database) we set bbf_role_admin as current user before execution so it is ought to return true in every bbf session valid case.
In order to add negative testcases for this hook we need to add scenarios of grant roles from psql endpoint which already exists in bbf https://github.com/babelfish-for-postgresql/babelfish_extensions/blob/BABEL_5_X_DEV/test/JDBC/expected/db_securityadmin-vu-verify.out#L1291

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -6287,3 +6288,49 @@ get_tvp_typename_typeschemaname(char *proc_name, char *target_arg_name, char **t
...
+									ObjectIdGetDatum(role),
+									ObjectIdGetDatum(member));
+	for (int i = 0; i < memlist->n_members; i++)
```
surendravishnoi, line 6321: let us not assume that memlist will be non-null, we should check for it and fallback to PG in case of null

anju15bharti: `SearchCatCacheList` inside `SearchSysCacheList2` ensures that memlist value should not be null as it allocates memory to memlist even if no entry matches in the search and assign memlist->n_members as 0.
And when memlist->n_members will be 0, this API will return false falling back to PG API.

surendravishnoi (approved):

PR #3295, opened 2024-12-20
Fix is_srvrolemember output for login same as specified server role

anju15bharti:
Earlier is_srvrolemember was returning 1 for login same as specified server role.

With this commit, now it will return expected output.


shalinilohia50 (approved):

jsudrik (approved):
Cherrypick of https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/3293 . Approved.

PR #3348, opened 2025-01-03
Improve performance of sys.sp_column_privileges by removing an unutilized check

shalinilohia50:
### Description

Improve performance of sys.sp_column_privileges by removing an unutilized check. 

### Issues Resolved

Task: BABEL-5353

Performance Testing
20 login, 50 databases, each with 10 tables in dbo and 10 schema with 10 tables each making it 5.5k tables 
Each database has 20 users in each database making it 500+ users
Each database has 10 new roles.

```
1> select count(*) from sp_column_privileges_view
2> go

 4748 rows (1469 ms)


1> select * from sp_table_privileges_view
2> go

585 rows (1630 ms)
```

| query | Execution time before [with func](ms) | Execution time after [with func, current change](ms)| Execution time before [with join](ms)|Execution time before [without join, it returns incorrect value]|
|-------| -------| ------| ------| ------|
|EXEC sp_column_privileges @table_name = 'db1_table3' | 2903 | 1453| 1927| 1338|
|EXEC sp_column_privileges @table_name = 'db1_table3', @table_qualifier = 'db1', @COLUMN_NAME='column1' |  2900 | 1442| 1812| 1290|
|CREATE PROCEDURE usp_GetColumnPrivileges @TableName NVARCHAR(128) AS begin EXEC sys.sp_column_privileges @table_name = @TableName; end ; usp_GetColumnPrivileges @TableName = 'db1_table3'; | 2893 | 1454| 1860| 1339|
|EXEC sp_table_privileges @table_name = 'db1_table3' | 1590 | 1590| 1473 | 1435|

* **Use case based -**
* **Boundary conditions -**
* **Arbitrary inputs -**
* **Negative test cases -**
* **Performance tests -**
* **Tooling impact -**
* **Client tests -**

### DIFF
--- a/contrib/babelfishpg_tsql/sql/babelfishpg_tsql.sql
+++ b/contrib/babelfishpg_tsql/sql/babelfishpg_tsql.sql
@@ -1370,6 +1370,7 @@ FROM sys.sp_column_privileges_internal()
 WHERE GRANTEE NOT IN ('db_datareader', 'db_datawriter');
 GRANT SELECT ON sys.sp_column_privileges_view TO PUBLIC;
 
+-- TODO: BABEL-5523
 CREATE OR REPLACE PROCEDURE sys.sp_column_privileges(
     "@table_name" sys.sysname,
     "@table_owner" sys.sysname = '',
--- a/contrib/babelfishpg_tsql/sql/babelfishpg_tsql.sql
+++ b/contrib/babelfishpg_tsql/sql/babelfishpg_tsql.sql
@@ -1354,7 +1354,7 @@ BEGIN
         JOIN information_schema.column_privileges t5 ON t1.relname = t5.table_name AND t2.nspname = t5.table_schema
         JOIN pg_attribute t6 ON t6.attrelid = t1.oid AND t6.attname = t5.column_name;
 END;
-$$ LANGUAGE plpgsql;
+$$ LANGUAGE plpgsql STABLE PARALLEL SAFE;
 
 CREATE OR REPLACE VIEW sys.sp_column_privileges_view AS
 SELECT
@@ -1450,7 +1450,7 @@ BEGIN
         JOIN information_schema.table_privileges t4 ON t1.relname = t4.table_name
     WHERE t4.privilege_type = 'DELETE';
 END;
-$$ LANGUAGE plpgsql;
+$$ LANGUAGE plpgsql STABLE PARALLEL SAFE;
 
 CREATE OR REPLACE VIEW sys.sp_table_privileges_view AS
 -- Will use sp_column_priivleges_view to get information from SELECT, INSERT and REFERENCES (only need permission from 1 column in table)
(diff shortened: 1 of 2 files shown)


contrib/babelfishpg_tsql/sql/babelfishpg_tsql.sql
```diff
@@ -1385,46 +1385,21 @@ BEGIN
...
+		TABLE_OWNER,
+		TABLE_NAME,
+		COLUMN_NAME,
+		GRANTOR,
+		GRANTEE,
+		PRIVILEGE,
+		IS_GRANTABLE
+		FROM sys.sp_column_privileges_view
+		WHERE pg_catalog.lower(@table_name) = pg_catalog.lower(table_name)
+			AND (pg_catalog.lower('dbo')= pg_catalog.lower(table_owner))
```
sumitj824, line 1400: If table_owner parameter is not specified, it will look for database owner i.e. 'dbo', right?

contrib/babelfishpg_tsql/sql/babelfishpg_tsql.sql
```diff
@@ -1354,7 +1354,7 @@ BEGIN
         JOIN information_schema.column_privileges t5 ON t1.relname = t5.table_name AND t2.nspname = t5.table_schema
         JOIN pg_attribute t6 ON t6.attrelid = t1.oid AND t6.attname = t5.column_name;
 END;
-$$ LANGUAGE plpgsql;
+$$ LANGUAGE plpgsql STABLE PARALLEL SAFE;
```
sumitj824, line 1357: Please check if there are sufficient test cases present for it in parallel mode.

sumitj824 (commented):
1. Please add the number and kind of objects used for performance testing.
2. Please add the execution time with/without JOIN in the view without the internal function we have introduced.
3. Please create a task to look for possibility of bringing down the performance numbers.
4. Please also add perf numbers for sp_table_privileges procedure.

contrib/babelfishpg_tsql/sql/babelfishpg_tsql.sql
shalinilohia50: Yes.

contrib/babelfishpg_tsql/sql/babelfishpg_tsql.sql
shalinilohia50: We are running the same test suite in parallel mode. Do we add any specific tests?

shalinilohia50:
> 1. Please add the number and kind of objects used for performance testing.
> 2. Please add the execution time with/without JOIN in the view without the internal function we have introduced.
> 3. Please create a task to look for possibility of bringing down the performance numbers.
> 4. Please also add perf numbers for sp_table_privileges procedure.

2. Not adding the numbers for without JOIN since it would give incorrect results.

HarshLunagariya (commented):
Correctness-wise looks good!

contrib/babelfishpg_tsql/sql/babelfishpg_tsql.sql
sumitj824: Just make sure that we have sufficient coverage for it.

sumitj824 (approved):

jsudrik (approved):
Approved.

PR #3367, opened 2025-01-07
Fix formatting issue in CAST MONEY to CHAR/VARCHAR

rishabhtanwar29:
### Description
While CASTing money/smallmoney to (N)VARCHAR/(N)CHAR, the value get rounded
to 4 decimal places instead of 2. Fixed this by implementing CASTs from money to
CHAR/VARCHAR datatype which rounds off input money value to 2 decimal places and
converts it to string.

Additionally, while CASTing money values to (N)CHAR, the resultant string gets left padded
instead of usual right padding. This commit also fixes this issue.

Task: BABEL-5478

* **Use case based -**
test_conv_money_to_varchar-vu-*

* **Boundary conditions -**

* **Arbitrary inputs -**

* **Negative test cases -**

* **Performance tests -**

* **Tooling impact -**

* **Client tests -**

### Check List


contrib/babelfishpg_common/src/varchar.c
```diff
@@ -1102,6 +1250,109 @@ varchar2numeric(PG_FUNCTION_ARGS)
...
 }
 
+/*
+ * fixeddecimal2str_helper
+ * Converts a fixed decimal value to a string, rounding it off to 2 decimal points.
+ */
+#define FIXEDDECIMAL_MULTIPLIER_NEW 100LL
+#define FIXEDDECIMAL_SCALE_NEW 2
+static char *
+fixeddecimal2str_helper(int64 val, char *buffer)
```
KushaalShroff: I understand that this is copied from fixeddecimal code, can we somehow, conditionally, do it over there? Would the create extension dependency chain be a deadlock in that case since varchar wont exist?

contrib/babelfishpg_common/src/varchar.c
```diff
@@ -48,9 +48,155 @@ void		TsqlCheckUTF16Length_bpchar(const char *s, int32 len, int32 maxlen, int ch
...
+/*
+  * pg_int64tostr Converts 'value' into a decimal string representation of the
+  * number.
+  *
+  * Caller must ensure that 'str' points to enough memory to hold the result
+  * (at least 21 bytes, counting a leading sign and trailing NUL). Return
+  * value is a pointer to the new NUL terminated end of string.
+  */
+static char *
+pg_int64tostr(char *str, int64 value)
```
KushaalShroff: Copied? Have we considered other options like rendezvous variables or moving the code appropriately?
One option could be to call fixeddecimal2str_helper(the OG one from fixeddecimal.c) inside fixeddecimal2varchar either through rendezvous or SQL API

test/JDBC/expected/non_default_server_collation/chinese_prc_ci_as/concat_ws-before-15_9-or-16_5-vu-verify.out
```diff
@@ -1244,67 +1244,67 @@ SELECT CONCAT_WS(col_varchar, col_varchar, col_smallmoney) from babel_3409_conca
...
+????14.00
+比尔·拉比尔·拉15.00
+莫斯莫斯16.00
 
 
-AnikaitAnikait17.0000
+AnikaitAnikait17.00
 
-AgrawalAgrawal18.0000
+AgrawalAgrawal18.00
```
KushaalShroff, line 1290: @Anikait143 Can you please review and validate the expected behaviour here?

contrib/babelfishpg_common/src/varchar.c
rishabhtanwar29: Yep, we create fixeddecimal before sys.varchar so can't move this logic there.

contrib/babelfishpg_common/src/varchar.c
rishabhtanwar29: Yeah I initially thought that copying these helper functions would be simpler compared to rendezvous or unnecessary SQL function but looks like we can use function pointers for cross module calls. Removed the copied code in the newer revision.

contrib/babelfishpg_common/src/varchar.c
```diff
@@ -1102,6 +1110,82 @@ varchar2numeric(PG_FUNCTION_ARGS)
...
+	end = fixeddecimal2str_p(val, buf, FIXEDDECIMAL_MULTIPLIER_NEW, FIXEDDECIMAL_SCALE_NEW);
+	len = (end - buf);
+
+	if (maxByteLen < 0)
+		maxByteLen = len + VARHDRSZ;
+	maxByteLen -= VARHDRSZ;
+	if (len > maxByteLen)
+		ereport(ERROR,
+				(errcode(ERRCODE_STRING_DATA_RIGHT_TRUNCATION),
+				 errmsg("There is insufficient result space to convert a money/smallmoney value to varchar.")));
```
KushaalShroff: to char/nchar

contrib/babelfishpg_common/src/varchar.c
```diff
@@ -1102,6 +1110,82 @@ varchar2numeric(PG_FUNCTION_ARGS)
...
+			load_external_function("$libdir/babelfishpg_money", "fixeddecimal2str", true, NULL);
+
+	end = fixeddecimal2str_p(val, buf, FIXEDDECIMAL_MULTIPLIER_NEW, FIXEDDECIMAL_SCALE_NEW);
+	len = (end - buf);
+
+	maxByteLen -= VARHDRSZ;
+	if (maxByteLen >= 0 && len > maxByteLen)
+		ereport(ERROR,
+				(errcode(ERRCODE_STRING_DATA_RIGHT_TRUNCATION),
+				 errmsg("There is insufficient result space to convert a money/smallmoney value to varchar.")));
```
KushaalShroff: to varchar/nvarchar

contrib/babelfishpg_common/src/varchar.c
```diff
@@ -1102,6 +1110,82 @@ varchar2numeric(PG_FUNCTION_ARGS)
...
+		fixeddecimal2str_p = (fixeddecimal2str_t)
+			load_external_function("$libdir/babelfishpg_money", "fixeddecimal2str", true, NULL);
+
+	end = fixeddecimal2str_p(val, buf, FIXEDDECIMAL_MULTIPLIER_NEW, FIXEDDECIMAL_SCALE_NEW);
+	len = (end - buf);
+
+	if (maxByteLen < 0)
+		maxByteLen = len + VARHDRSZ;
+	maxByteLen -= VARHDRSZ;
+	if (len > maxByteLen)
```
KushaalShroff, line 1171: Why no:
```
if (maxByteLen >= 0 && len > maxByteLen)
```

contrib/babelfishpg_common/src/varchar.c
KushaalShroff: Anyway to test this condition?

KushaalShroff (changes requested):

contrib/babelfishpg_money/fixeddecimal.c
```diff
@@ -393,12 +396,30 @@ pg_int64tostr_zeropad(char *str, int64 value, int64 padding)
  *		Prints the fixeddecimal 'val' to buffer as a string.
  *		Returns a pointer to the end of the written string.
  */
-static char *
-fixeddecimal2str(int64 val, char *buffer)
+char *
+fixeddecimal2str(int64 val, char *buffer,
+				int64 fixeddecimal_multiplier,
+				int64 fixeddecimal_scale)
```
hopebo, line 402: If you're adding parameter to a general function, you shouldn't only deal with the scenario when  `fixeddecimal_scale` equals to 2. How about 3?

contrib/babelfishpg_common/src/varchar.c
```diff
@@ -1102,6 +1110,82 @@ varchar2numeric(PG_FUNCTION_ARGS)
 	PG_RETURN_NUMERIC(result);
 }
 
+#define FIXEDDECIMAL_MULTIPLIER_NEW 100LL
+#define FIXEDDECIMAL_SCALE_NEW 2
```
hopebo: Can we use better meaningful macro naming like `FIXEDDECIMAL_2VARCHAR_SCALE`?

contrib/babelfishpg_common/src/varchar.c
rishabhtanwar29: 1. This is always `varchar` in T-SQL for all string types.
2. There is already one test case in test_conv_money_to_varchar-vu-verify file.

contrib/babelfishpg_common/src/varchar.c
rishabhtanwar29: Note that we are checking if (maxByteLen < 0) then we are adjusting it to (len + VARHDRSZ) so it will never be negative at this point. I kept this different from the function for varchar (`fixeddecimal2bpchar`) as there I am not using this maxByteLen but here I am using it to add padding.
But for simplicity, updated the logic to be same at both the places.

contrib/babelfishpg_common/src/varchar.c
rishabhtanwar29: his is always varchar in T-SQL for all string types.

contrib/babelfishpg_money/fixeddecimal.c
rishabhtanwar29: So far I don't see any other scale values being used although I have now made the logic generic.

contrib/babelfishpg_money/fixeddecimal.c
```diff
@@ -409,9 +409,10 @@ fixeddecimal2str(int64 val, char *buffer,
 	 * By default, fixeddecimal contains 4 digits after decimal but if we only need
 	 * 2 decimal points, remove last two decimal digits, round off remaining value.
 	 */
-	if (fixeddecimal_scale == 2)
+	if (fixeddecimal_scale < 4)
```
KushaalShroff, line 412: For < 4 you have commented we will remove 2 decimal points, should it be the scale_factor decimal points?

KushaalShroff (approved):

contrib/babelfishpg_money/fixeddecimal.c
rishabhtanwar29: Will address this separately.

test/JDBC/expected/concat-vu-prepare.out
```diff
@@ -466,7 +466,7 @@ RETURN (SELECT TOP 1 CONCAT(col_nvarchar, col_nchar, col_int, col_bigint, col_sm
 GO
 
 CREATE VIEW babel_3409_concat_dep_view2 AS
-    SELECT CONCAT(col_nvarchar, col_nchar, col_varchar, col_decimal, col_smallmoney, col_smalldatetime) AS concatenated_column FROM babel_3409_concat_t1
+    SELECT CONCAT(col_nvarchar, col_nchar, col_varchar, col_decimal, col_smalldatetime) AS concatenated_column FROM babel_3409_concat_t1
```
Anikait143, line 469: Why we had to remove `smallmoney` column in this test case in all the prepare files?

rishabhtanwar29: Cause otherwise we had to unnecessarily create a bunch of -before/after test files with almost same content just for a single line diff.

contrib/babelfishpg_money/fixeddecimal.c
hopebo: Have you really fixed the issue? I don't see the change.

contrib/babelfishpg_common/src/varchar.c
hopebo: Have you fixed it?

contrib/babelfishpg_money/fixeddecimal.c
hopebo, line 412: You should use existing macro to replace `4`. Otherwise you can raise an error, if the input argument is greater than 4

test/JDBC/expected/concat-vu-prepare.out
Anikait143: Ok, just wanted to confirm if we are not getting any unexpected results here. Otherwise LGTM.

Anikait143 (approved):

contrib/babelfishpg_money/fixeddecimal.c
rishabhtanwar29: Fixeddecimal scale has maximum value 4 so it can't be higher. Moreover, this is an internal function where we have made sure to pass valid scale values, there is no way a user can directly interact with this function and supply invalid input values.

PR #3386, opened 2025-01-09
[OSS-ONLY] Restrict grant to/from any babelfish created role via PG port

shalinilohia50:
### Description

Restrict grant to/from any babelfish created role via PG port
All server and database roles including fixed server roles and database roles have 'bbf_role_admin' as its member. Added a check to identify if 'bbf_role_admin' is a member of any role in addition to the existing checks to avoid granting any Babelfish db/server role  to another Babelfish db/server roles including fixed db and server roles.

### Issues Resolved

Task: BABEL-5505

* **Use case based -**
* **Boundary conditions -**
* **Arbitrary inputs -**
* **Negative test cases -**
* **Performance tests -**
* **Tooling impact -**
* **Client tests -**

### DIFF
--- a/contrib/babelfishpg_tds/src/backend/tds/tdsutils.c
+++ b/contrib/babelfishpg_tds/src/backend/tds/tdsutils.c
@@ -1246,10 +1246,7 @@ handle_grant_role(GrantRoleStmt *grant_stmt)
 
 		roleid = get_role_oid(rolename, false);
 		if (OidIsValid(roleid) && is_babelfish_role(rolename))
-		{
-			pfree(rolename);
 			check_babelfish_alterrole_restictions(false);
-		}
 	}
 
 	/* Restrict grant to/from bbf created role */
--- a/contrib/babelfishpg_tds/src/backend/tds/tdsutils.c
+++ b/contrib/babelfishpg_tds/src/backend/tds/tdsutils.c
@@ -949,11 +949,13 @@ is_babelfish_role(const char *role)
 	Oid			bbf_msdb_guest_oid;
 	Oid			securityadmin;
 	Oid			dbcreator;
+	Oid			bbf_admin_oid;
 
 	sysadmin_oid = get_role_oid(BABELFISH_SYSADMIN, true);	/* missing OK */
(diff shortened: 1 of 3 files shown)


contrib/babelfishpg_tds/src/backend/tds/tdsutils.c
```diff
@@ -1242,19 +1245,22 @@ handle_grant_role(GrantRoleStmt *grant_stmt)
 			continue;
 
 		roleid = get_role_oid(rolename, false);
-		if (OidIsValid(roleid) && IS_DEFAULT_BBF_SERVER_ROLE(rolename))
+		if (OidIsValid(roleid) && is_babelfish_role(rolename))
+		{
+			pfree(rolename);
```
rishabhtanwar29: We didn't palloc'd this object so shouldn't be free'ing it as well.

contrib/babelfishpg_tds/src/backend/tds/tdsutils.c
```diff
@@ -973,7 +975,8 @@ is_babelfish_role(const char *role)
 		&& OidIsValid(bbf_msdb_guest_oid)
 		&& is_member_of_role(role_oid, bbf_master_guest_oid)
 		&& is_member_of_role(role_oid, bbf_tempdb_guest_oid)
-		&& is_member_of_role(role_oid, bbf_msdb_guest_oid))
+		&& is_member_of_role(role_oid, bbf_msdb_guest_oid)
+		&& is_member_of_role(bbf_admin_oid, role_oid))
```
rishabhtanwar29, line 979: What about only checking membership against bbf_role_admin? Is it sufficient or do we still need the above checks?

shalinilohia50: Tried, that wasn't sufficient for a case where a TSQL user having sysadmin privilege was able to create/alter a user via PG port. Since TSQL user created the PG user, it was being considered as a BBF user and didn't get dropped.

HarshLunagariya (approved):

PR #3394, opened 2025-01-13
fix object resolution inside pltsql routines

tanscorpio7:
### Description

When an non fully qualified object is used inside a pltsql routine, it must get resolved to the logical database that contains the routine being executed. If the schema is also not specified then the object must be first looked inside the procedure's schema then the dbo schema of the database.

For DDL and EXEC statements the order of schemas is different
<default_schema_of_user>, <dbo_schema_of_database>, sys, pg_catalog

We achieve this behaviour by maintain two global static variables.
1. current_db_search_path - search path for the current active logical database for the current login
2. pltsql_search_path - search path of the inner most executing routine

Now based on the statement being executed, we will decided what the search path should be inside dispatch_stmt().
This is needed since search path is a transactional GUC and setting it once inside a procedure is not enough since any rollback or error could reset the search path to its previous value. Also attempts at making search path a scope based GUC failed since engine internally sets the search path for some cases and expects it work transactionally.

Some other changes included in this PR

1. Merged `set_session_properties` function into `set_cur_user_db_and_path`. Just needed pass one more argument.
2. "sp_" prefixed procedures called like `db.dbo.sp_` & `db..sp_` must also look into dbo schema of master database and sys schema for the procedure even though the db name and schema has been specified.
We fix this by removing the db name and schema name at ANTLR parsing.
3. Non pltsql "sp_" system procedure in master.dbo schema will not have tsql error handling. After the fix we did in 2, we can just remove this sp_ procedures in master.dbo schema. Since master.dbo.sp_* will anyway look into sys schema as well now.
4. db_id() & db_name() were defined as immutable which is not correct, since db_id can change in the same session.
5. Use syscache for `get_authid_user_ext_schema_name`

### Issues Resolved

[BABEL-5448]

### Sign Off

### Check List


shalinilohia50:
Please update the PR description explaining the changes.

contrib/babelfishpg_tsql/src/pl_comp.c
```diff
@@ -967,6 +969,18 @@ do_compile(FunctionCallInfo fcinfo,
 	for (i = 0; i < function->fn_nargs; i++)
 		function->fn_argvarnos[i] = in_arg_varnos[i];
 
+	/* store the logical db which contains this function */
+	if (IS_TDS_CONN() && procStruct->provolatile != PROVOLATILE_IMMUTABLE && procStruct->proparallel == PROPARALLEL_UNSAFE)
```
HarshLunagariya, line 973: Why `procStruct->provolatile != PROVOLATILE_IMMUTABLE && procStruct->proparallel == PROPARALLEL_UNSAFE` ?

contrib/babelfishpg_tsql/src/session.c
```diff
@@ -128,8 +127,6 @@ set_session_properties(const char *db_name)
 				(errcode(ERRCODE_UNDEFINED_DATABASE),
 				 errmsg("database \"%s\" does not exist", db_name)));
 
-	check_session_db_access(db_name);
-
 	set_cur_user_db_and_path(db_name);
```
HarshLunagariya: Let's analyze all other instances of `set_cur_user_db_and_path()` and make sure that it is expecting to throw error everywhere. Ideally it should but just double check.

contrib/babelfishpg_tsql/src/iterative_exec.c
```diff
@@ -2184,3 +2190,77 @@ send_env_change_token_on_txn_abort(void)
...
+ * before executing any pltsql statement
+ */
+static void
+set_search_path_for_pltsql_stmt(PLtsql_stmt *stmt)
+{
+	char		*cur_dbname;
+	char 		*new_search_path = NULL;
+
+	if (!IS_TDS_CONN() || IsInParallelMode())
+		return;
```
HarshLunagariya: What is the effect on interoperability behaviour?

contrib/babelfishpg_tsql/src/tsqlIface.cpp
```diff
@@ -2182,8 +2151,14 @@ class tsqlBuilder : public tsqlCommonMutator
...
-		// record that the stmt is ddl
-	 	stmt->is_ddl = true;
+
+		/*
+		 * This flag is only used while setting the search path for ddl statements in
+		 * execsql. we do not change search path for alter user because it does not depend
+		 * on search path and doing so has side effects like change to user default schema
+		 * does not take effect in current session
+		 */
+	 	stmt->is_ddl = !ctx->alter_user();
```
HarshLunagariya: It makes sense. Can you please elaborate more regarding alter user's problematic scenario?

contrib/babelfishpg_tsql/src/session.h
```diff
@@ -14,12 +14,18 @@ extern void check_session_db_access(const char *dn_name);
...
 
 /* Hooks for parallel workers for babelfish fixed state */
 extern void babelfixedparallelstate_insert(ParallelContext *pcxt, bool estimate);
 extern void babelfixedparallelstate_restore(shm_toc *toc);
 
+/* search path when executing pltsql routines */
+extern char *pltsql_search_path;
+/* search path for current active logical database */
+extern char *current_db_search_path;
+
```
HarshLunagariya: 1. What is the transactional/Error handling behaviour of these?
2. How are we resetting it in case of rollback?

For current_db_search_path, It straight forward because any error doesn't switch the database context. But what happens with `pltsql_search_path` when there is a cross-db procedure execution and there is an error. Do we reset it to something in those cases? or we just really on setting it to new value when we execution a new procedure.

contrib/babelfishpg_tsql/src/pl_comp.c
tanscorpio7: Immutable functions cannot rely on search path or object resolution so we do not do this.
Same for parallel safe functions, GUCs cannot change inside parallel sage functions.

contrib/babelfishpg_tsql/src/session.c
tanscorpio7: `check_session_db_access` was a redundant call here. `set_cur_user_db_and_path` will anyways make a call to `get_user_for_database` we can throw error there if no user is found.

contrib/babelfishpg_tsql/src/iterative_exec.c
tanscorpio7: This should not break anything from interoperability case. Whatever worked before should continue to work.

contrib/babelfishpg_tsql/src/tsqlIface.cpp
tanscorpio7: Actually I can remove this change now. This was needed in an earlier approach. The current approach does not need this change.

contrib/babelfishpg_tsql/src/session.h
tanscorpio7: `pltsql_search_path` will be a scope based static variable.
Errors / transaction rollback / commit will not affect it.
This value will only change when entering or exiting pltsql_call_handler (for both error case and no error case)

contrib/babelfishpg_tsql/src/iterative_exec.c
tanscorpio7: We will stop seeing error like `database "" does not exists `. But that also means that object resolution will not work from postgres endpoint for tsql routines because we do not have the curr_db_id set & the search path is not in our control.

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -1002,47 +1002,33 @@ get_authid_user_ext_physical_name(const char *db_name, const char *login)
...
+											   Anum_bbf_authid_user_ext_default_schema_name);
+			char  	*default_schema_name = TextDatumGetCString(schema_datum);
+
+			if (strlen(default_schema_name))
+				schema_name = default_schema_name;
+		}
+
+		pfree(db_name_cstring);
+		ReleaseSysCache(tuple_user_ext);
+	}
```
tanscorpio7, line 1049: Using syscache instead of index scan.

contrib/babelfishpg_tsql/src/iterative_exec.c
```diff
@@ -2184,3 +2190,89 @@ send_env_change_token_on_txn_abort(void)
...
+		else if (execsqlstmt->is_ddl && current_db_search_path)
+		{
+			new_search_path = pstrdup(current_db_search_path);
+		}
+	}
+	else if (stmt->cmd_type == PLTSQL_STMT_EXEC)
+	{
+		PLtsql_stmt_exec *execstmt = (PLtsql_stmt_exec *)stmt;
+
+		if (strncmp(execstmt->proc_name, "sp_", 3) == 0 &&
```
thephantomthief, line 2237: can a user not create `sp_` prefixed procedures?

contrib/babelfishpg_tsql/src/tsqlIface.cpp
```diff
@@ -6898,29 +6856,16 @@ makeExecuteProcedure(ParserRuleContext *ctx, std::string call_type)
 		
 		// For sp_* procs, truncate proc name to sp_* if the schema is "dbo" or "sys" or has leading dots
 		// ToDo: handle 'EXEC mydb..sp_proc' where sp_proc gets executed in the context of 'mydb', even when the current DB is not 'mydb'	
-		if ((pg_strncasecmp(name.c_str(), "..dbo.sp_", 9) == 0) || (pg_strncasecmp(name.c_str(), "..sys.sp_", 9) == 0))
+		if ((!proc_name.empty() && pg_strncasecmp(proc_name.c_str(), "sp_", 3) == 0) &&
```
thephantomthief, line 6879: Please update the comment to reflect new handling

contrib/babelfishpg_tsql/src/iterative_exec.c
tanscorpio7: They can, but `sp_` procedure still looks into master.dbo schema as last resort when user executes
```
sp_
<db>.dbo.sp_
dbo.sp_
<db>..sp_
```

contrib/babelfishpg_tsql/sql/ownership.sql
```diff
@@ -178,54 +178,6 @@ BEGIN
...
-
-  ALTER PROCEDURE master_dbo.sp_addlinkedserver OWNER TO sysadmin;
-
-  CREATE OR REPLACE PROCEDURE master_dbo.sp_addlinkedsrvlogin( IN "@rmtsrvname" sys.sysname,
-                                                      IN "@useself" sys.varchar(8) DEFAULT 'TRUE',
-                                                      IN "@locallogin" sys.sysname DEFAULT NULL,
-                                                      IN "@rmtuser" sys.sysname DEFAULT NULL,
-                                                      IN "@rmtpassword" sys.sysname DEFAULT NULL)
-  AS 'babelfishpg_tsql', 'sp_addlinkedsrvlogin_internal'
-  LANGUAGE C;
```
HarshLunagariya, line 199: Can it affect metadata in any way? For example, entries in sys.objects /sys.procedures.

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -6038,4 +6010,11 @@ remove_db_name_in_schema(const char *object_name, const char *object_type)
...
-}
\ No newline at end of file
+}
+
+/* Check if current connection is a tds connection */
+static bool
+is_bbf_tds_connection(void)
+{
+	return IS_TDS_CONN();
+}
```
HarshLunagariya, line 6030: Is it really required? Can't we directly access the is_tds_conn from MyProcPort structure.

contrib/babelfishpg_tsql/src/tsqlIface.cpp
```diff
@@ -1908,11 +1906,10 @@ class tsqlBuilder : public tsqlCommonMutator
 				if (ctx_name->database)
 				{
 					db_name = stripQuoteFromId(ctx_name->database);
-					if (!string_matches(db_name.c_str(), get_cur_db_name()))
-					{
-						is_cross_db = true;
-					}
+					is_cross_db = true;
```
HarshLunagariya, line 1909: Seems like we are marking is_cross_db when there is db specified in the statement. If that's the case then I suggest to name it better.

contrib/babelfishpg_tsql/src/session.c
```diff
@@ -177,44 +175,49 @@ set_cur_user_db_and_path(const char *db_name)
...
+	SetConfigOption("search_path", path,
+					PGC_SUSET, PGC_S_SESSION);
+	/*
+	 * Incase of error inside parallel worker we could reach here before transaction
+	 * abort. Do not try to set_config since it will fail inside a parallel worker
+	 * This function is used in code path where error is never expected
+	 */
+	if (!(IsParallelWorker() && IsAbortedTransactionBlockState()))
+		SetConfigOption("search_path", path,
+						PGC_SUSET, PGC_S_DATABASE_USER);
```
HarshLunagariya: 1. What is the reason behind keeping GUC source level PGC_S_DATABASE_USER?
2. Can you elaborate a bit more on parallel worker scenario? Will we be able to cover it with testcases?

contrib/babelfishpg_tsql/src/session.c
```diff
@@ -177,44 +175,49 @@ set_cur_user_db_and_path(const char *db_name)
...
-	{
-		const char *guest_schema = get_authid_user_ext_schema_name(db_name, "guest");
+	Assert (db_name != NULL && strlen(db_name) > 0);
 
-		if (!guest_schema)
-			guest_schema = "guest";
-		physical_schema = get_physical_schema_name(pstrdup(db_name), guest_schema);
-	}
-	else
+	schema = get_authid_user_ext_schema_name(db_name, user);
```
HarshLunagariya, line 203: get_authid_user_ext_schema_name() handles the guest and dbo case? Do we have default_schema set for them in the our catalogs?

contrib/babelfishpg_tsql/src/pl_comp.c
```diff
@@ -3252,3 +3268,34 @@ reset_cached_batch()
 		delete_cached_batch(cur_handle_id--);
 	cur_handle_id = 1;
 }
+
+static char *
+get_local_schema_for_bbf_functions(Oid proc_nsp_oid, int16 dbid)
```
HarshLunagariya, line 3273: Can proc_nsp_oid be in the separate database from dbid?

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -5569,6 +5571,12 @@ pltsql_call_handler(PG_FUNCTION_ARGS)
 		{
 			set_procid(func->fn_oid);
 
+			/* for cross db func/proc calls switch to that db */
+			if (DbidIsValid(func->fn_dbid) && get_cur_db_id() != func->fn_dbid)
+				set_cur_user_db_and_path(get_db_name(func->fn_dbid), false);
+
+			pltsql_search_path = func->fn_search_path;
```
HarshLunagariya: Is it redundant to reset it in catch/finally in case of error?

test/JDBC/expected/objectproperty-vu-verify.out
```diff
@@ -233,42 +233,42 @@ SELECT OBJECTPROPERTY(OBJECT_ID('dbo.sp_addlinkedserver'), 'IsMSShipped')
 GO
 ~~START~~
 int
-1
+<NULL>
```
HarshLunagariya: We should fix this.

test/JDBC/expected/getdate-vu-verify.out
```diff
@@ -429,6 +429,7 @@ int#!#int#!#int#!#int
 
 
 -- psql
+SET search_path = master_dbo, sys, pg_catalog;
```
HarshLunagariya, line 432: Seems like consequence of blocking the "pltsql" search path specific logic from PG endpoint. Let's verify that there are no such cx scenarios/workarounds suggested in the past. Or any features relying on executing pltsql procedures from PG endpoint.

test/JDBC/expected/BABEL_4442.out
```diff
@@ -80,15 +80,15 @@ SELECT babel_4442_f(), current_setting('search_path'), babel_4442_s1.babel_4442_
 GO
 ~~START~~
 int#!#text#!#int#!#text#!#int#!#int#!#text#!#int
-1#!#master_dbo, "$user", sys, pg_catalog#!#2#!#master_dbo, "$user", sys, pg_catalog#!#3#!#1#!#master_dbo, "$user", sys, pg_catalog#!#1
+1#!#master_dbo, master_dbo, sys, pg_catalog#!#2#!#master_babel_4442_s1, master_dbo, sys, pg_catalog#!#3#!#1#!#master_babel_4442_s2, master_dbo, sys, pg_catalog#!#1
```
HarshLunagariya: Shouldn't we reset the search_path are executing the function? Here, What happens if instead of `babel_4442_s1 .babel_4442_f(),current_setting('search_path')` ,`babel_4442_s1 .babel_4442_f(),non-pltsql routine/non-pltsql stmt` gets executed and it behaves acc. to `master_babel_4442_s1, master_dbo, sys, pg_catalog` search path which is incorrect.

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -5612,10 +5620,16 @@ pltsql_call_handler(PG_FUNCTION_ARGS)
...
+		pltsql_search_path = saved_pltsql_search_path;
+		pltsql_check_search_path = true;
 
 		pltsql_remove_current_query_env();
 		pltsql_revert_guc(save_nestlevel);
 		pltsql_revert_last_scope_identity(scope_level);
+
+		/* reset db context must always be the last line in this block */
+		if (get_cur_db_id() != saved_dbid)
+			set_cur_user_db_and_path(get_db_name((saved_dbid)), false);
```
HarshLunagariya, line 5631: How the error handling takes place here?

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -4654,3 +4612,45 @@ exec_stmt_partition_scheme(PLtsql_execstate *estate, PLtsql_stmt_partition_schem
...
+	initStringInfo(&res);
+
+	foreach(lc, namelist)
+	{
+		if (first)
+			first = false;
+		else
+			appendStringInfoChar(&res, ',');
+
+		appendStringInfoString(&res, (char *)lfirst(lc));
```
HarshLunagariya: Can't we just build a new search_path instead of doing all this? to make it less error-prone.

test/JDBC/expected/BABEL_OBJECT_RESOLUTION_IN_ROUTINES-vu-prepare.out
```diff
@@ -0,0 +1,511 @@
...
+
+
+
+
+
+
+
+
+-- tsql
+-- In this test file we will test babelfish object resolution inside procedure functions and triggers
```
HarshLunagariya, line 11: Let's add some test for 
1. sp_executesql proc execution variations
2.

thephantomthief (approved):

contrib/babelfishpg_tsql/sql/ownership.sql
tanscorpio7: Yes, sys.objects will not show these procedures for master database which is also expected from tsql point of view.

tanscorpio7: We only created these procedures in master.dbo because we wanted some queries like master.dbo.sp_add ... to work.

contrib/babelfishpg_tsql/src/hooks.c
tanscorpio7: Was not able to access this structure in namespace.c. Importing the required headers caused other issues.

contrib/babelfishpg_tsql/src/tsqlIface.cpp
tanscorpio7: Yes, parsing is not the right place to determine if this is a cross db reference, it should be done during execution.

contrib/babelfishpg_tsql/src/session.c
tanscorpio7: > What is the reason behind keeping GUC source level PGC_S_DATABASE_USER?

USE DB statement is non transactional, it does not rollback on transaction rollback. 
`PGC_S_DATABASE_USER` sets the default value of a GUC for the session, so even if we rolled back an entire transaction which executed a USE DB, we will still end up with the right search path.

This is covered by existing test cases.

tanscorpio7: > Can you elaborate a bit more on parallel worker scenario? Will we be able to cover it with testcases?
Yes, I actually found this issue when I was running some local tests.

CREATE A pltsql function which executed something like SELECT COUNT(*) FROM t WHERE CAST(col AS TINY) < 1000;
Now make a call to this function from another db.

`SELECT * FROM db.dbo.pltsqfunction();`

The entire function will not run inside parallel worker but the SELECT query inside the function will.
To enter the error code path we can add a row in table t which should throw an error during `CAST(col AS TINY)`

contrib/babelfishpg_tsql/src/pl_exec-2.c
tanscorpio7: Do you mean directly using the `current_db_name` ?
Yes, that might be more readable.

test/JDBC/expected/BABEL_OBJECT_RESOLUTION_IN_ROUTINES-vu-prepare.out
tanscorpio7: Sure, 2nd point is empty, any other tests to add ?

contrib/babelfishpg_tsql/src/session.c
tanscorpio7: Yes, I will add upgrade test cases for it.

contrib/babelfishpg_tsql/src/pl_comp.c
tanscorpio7: No, caller needs to ensure that does not happen.

contrib/babelfishpg_tsql/src/pl_handler.c
tanscorpio7: We need to reset the db_context in case of an error inside proc/function. Otherwise we will leak the db_context

contrib/babelfishpg_tsql/src/pl_handler.c
tanscorpio7: Could not understand this. We are resetting the db_context here in both cases (error and no error)

test/JDBC/expected/BABEL_4442.out
tanscorpio7: For non tsql functions/procedure we are not handling anything from search path point of view. They will run with the last active tsql search path.

test/JDBC/expected/BABEL_OBJECT_RESOLUTION_IN_ROUTINES-vu-prepare.out
tanscorpio7: added tests for `sp_executesql`

contrib/babelfishpg_tsql/src/pl_exec-2.c
tanscorpio7: Done

contrib/babelfishpg_tsql/src/session.c
tanscorpio7: Added tests

contrib/babelfishpg_tsql/src/session.c
tanscorpio7: Added tests

.github/scripts/scan-warnings.sh
```diff
@@ -49,7 +49,7 @@ if [[ "$SNAPSHOT_ACTIVE_COUNT" -ne 44 ]]; then
     ERROR_FOUND=true
 fi
 
-if [[ "$LEAK_COUNT" -ne 324 ]]; then
+if [[ "$LEAK_COUNT" -ne 350 ]]; then
     echo "Error: Expected 324 leak warnings, but found $LEAK_COUNT"
```
thephantomthief, line 53: nit: should be 350 now

shardgupta (approved):

PR #3400, opened 2025-01-13
[OSS-ONLY] Restrict grant to/from any babelfish created role via PG port

shalinilohia50:
### Description

Restrict grant to/from any babelfish created role via PG port
All server and database roles including fixed server roles and database roles have 'bbf_role_admin' as its admin. Added a check to identify if 'bbf_role_admin' is the direct admin of the role to avoid granting any Babelfish role  to another Babelfish roles including fixed db and server roles. 

### Issues Resolved

Task: BABEL-5505

* **Use case based -**
* **Boundary conditions -**
* **Arbitrary inputs -**
* **Negative test cases -**
* **Performance tests -**
* **Tooling impact -**
* **Client tests -**

### DIFF
--- a/contrib/babelfishpg_tds/src/backend/tds/tdsutils.c
+++ b/contrib/babelfishpg_tds/src/backend/tds/tdsutils.c
@@ -35,7 +35,6 @@
 #include "utils/fmgroids.h"
 #include "utils/rel.h"
 #include "utils/syscache.h"
-#include "utils/lsyscache.h"
 #include "miscadmin.h"
 #include "utils/builtins.h"
 
@@ -934,11 +933,11 @@ check_babelfish_droprole_restrictions(char *role)
 static bool
 is_bbf_admin_of_role(Oid roleid, Oid bbf_role_admin)
 {
-	CatCList   *memlist;
-	int			i;
+	CatCList	*memlist;
+	int		i;
 	bool		is_bbf_admin_of_role = true;
 
-	/* Find roles that roleid is directly a member of */
+	/* Find roles that are direct membes of roleid */
 	memlist = SearchSysCacheList1(AUTHMEMROLEMEM,
                                   ObjectIdGetDatum(roleid));
 	for (i = 0; i < memlist->n_members; i++)
@@ -968,7 +967,7 @@ is_bbf_admin_of_role(Oid roleid, Oid bbf_role_admin)
  * 	Since role related DDLs could be executed in any PG databases,
  * 	This function check the underlying assumption:
  * 	1. Given role is the bbf_role_admin.
- * 	2. If bbf_role_admin is admin of the given role.
+ * 	2. If bbf_role_admin is only admin of the given role.
  */
 static bool
 is_babelfish_role(const char *role)
--- a/contrib/babelfishpg_tds/src/backend/tds/tdsutils.c
+++ b/contrib/babelfishpg_tds/src/backend/tds/tdsutils.c
@@ -928,14 +928,14 @@ check_babelfish_droprole_restrictions(char *role)
 }
 
 /*
- * Checks if bbf_role_admin is the direct admin of a role.
+ * Checks if admin_of is member of the given roleid and its immediate admin.
  */
 static bool
-is_bbf_admin_of_role(Oid roleid, Oid bbf_role_admin)
+member_and_immediate_admin_of_role(Oid admin_of, Oid roleid)
 {
 	CatCList	*memlist;
 	int		i;
-	bool		is_bbf_admin_of_role = true;
+	bool		member_and_immediate_admin_of_role = false;
 
 	/* Find roles that are direct membes of roleid */
 	memlist = SearchSysCacheList1(AUTHMEMROLEMEM,
@@ -945,15 +945,15 @@ is_bbf_admin_of_role(Oid roleid, Oid bbf_role_admin)
 		HeapTuple	tup = &memlist->members[i]->tuple;
 		Form_pg_auth_members form = (Form_pg_auth_members) GETSTRUCT(tup);
 
-		/* If role is member of a login other than bbf_role_admin WITH ADMIN OPTION. */
-		if (form->member != bbf_role_admin && form->admin_option)
+		/* If admin_of is the direct member of roleid WITH ADMIN OPTION. */
+		if (form->member == admin_of && form->admin_option)
 		{
-			is_bbf_admin_of_role = false;
+			member_and_immediate_admin_of_role = true;
 			break;
 		}
 	}
 	ReleaseSysCacheList(memlist);
-	return is_bbf_admin_of_role;
+	return member_and_immediate_admin_of_role;
 }
 
 /*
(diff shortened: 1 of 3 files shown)


contrib/babelfishpg_tds/src/backend/tds/tdsutils.c
```diff
@@ -925,6 +927,35 @@ check_babelfish_droprole_restrictions(char *role)
 	}
 }
 
+/*
+ * Checks if bbf_role_admin is the direct admin of a role.
+ */
+static bool
+is_bbf_admin_of_role(Oid roleid, Oid bbf_role_admin)
```
anju15bharti: We already have an API in catalog.c for this `bbf_check_member_has_direct_priv_to_grant_role`, we should check that.

anju15bharti (changes requested):

contrib/babelfishpg_tds/src/backend/tds/tdsutils.c
shalinilohia50: `if (sql_dialect != SQL_DIALECT_TSQL || !IS_TDS_CONN()
		|| member !=  get_bbf_role_admin_oid())
		return false;`
		We can't directly use this function, it will return false for PG dialect.  Also, in this function `bbf_check_member_has_direct_priv_to_grant_role` we are checking if bbf_admin_role is the member of role -> It's not what's needed here. 
		
If a TSQL login L1 creates a user U1 via TSQL -> bbf_role_admin is it's direct member with admin option.
If a TSQL login L2 created a U2 via PG -> L2 is it's direct member with admin option. 
`is_bbf_admin_of_role` it differentiates between the above 2 cases.

contrib/babelfishpg_tds/src/backend/tds/tdsutils.c
```diff
@@ -925,6 +927,35 @@ check_babelfish_droprole_restrictions(char *role)
...
+	{
+		HeapTuple	tup = &memlist->members[i]->tuple;
+		Form_pg_auth_members form = (Form_pg_auth_members) GETSTRUCT(tup);
+
+		/* If role is member of a login other than bbf_role_admin WITH ADMIN OPTION. */
+		if (form->member != bbf_role_admin && form->admin_option)
+		{
+			is_bbf_admin_of_role = false;
+			break;
+		}
```
rishabhtanwar29: We should rather check if given admin is bbf_role_admin, like this
```
if (form->admin_option && form->member == bbf_role_admin)
{
    is_bbf_admin_of_role = true;
    break;
}
```

contrib/babelfishpg_tds/src/backend/tds/tdsutils.c
rishabhtanwar29: I think we should make this function generic by naming it like `is_member_immediate_admin_of_role` and then checking whether given member is immediate admin of role or not.

tanscorpio7: What happens if customer granted babelfish role to a postgres endpoint created role, now it is possible that babelfish role has some other admin other then bbf_role_admin.

tanscorpio7: I think for simplicity we could block anything that bbf_role_admin has admin privileges over. Using babelfish roles to create other roles from postgres endpoint seems less likely to occur compared to granting bbf roles to pg roles

contrib/babelfishpg_tds/src/backend/tds/tdsutils.c
```diff
@@ -925,6 +927,35 @@ check_babelfish_droprole_restrictions(char *role)
...
+	/* Find roles that are direct membes of roleid */
+	memlist = SearchSysCacheList1(AUTHMEMROLEMEM,
+                                  ObjectIdGetDatum(roleid));
+	for (i = 0; i < memlist->n_members; i++)
+	{
+		HeapTuple	tup = &memlist->members[i]->tuple;
+		Form_pg_auth_members form = (Form_pg_auth_members) GETSTRUCT(tup);
+
+		/* If role is member of a login other than bbf_role_admin WITH ADMIN OPTION. */
+		if (form->member != bbf_role_admin && form->admin_option)
```
thephantomthief: Can 2 roles be admin of a Babelfish role? If yes, do we want to disallow dropping of role from PG endpoint, if one of the admin is bbf_role_admin. From current condition it seems like a user can drop such a role from PG endpoint.

contrib/babelfishpg_tds/src/backend/tds/tdsutils.c
tanscorpio7: I think this function does something entirely different to what its name suggests.
It will return true only if bbf_role_admin has admin privs over it.

tanscorpio7 (changes requested):

contrib/babelfishpg_tds/src/backend/tds/tdsutils.c
tanscorpio7: this will also break if super users create a role from pg_endpoint.

contrib/babelfishpg_tds/src/backend/tds/tdsutils.c
```diff
@@ -934,46 +965,24 @@ check_babelfish_droprole_restrictions(char *role)
...
-	bbf_tempdb_guest_oid = get_role_oid("tempdb_guest", true);
-	bbf_msdb_guest_oid = get_role_oid("msdb_guest", true);
-	if (OidIsValid(bbf_master_guest_oid)
-		&& OidIsValid(bbf_tempdb_guest_oid)
-		&& OidIsValid(bbf_msdb_guest_oid)
-		&& is_member_of_role(role_oid, bbf_master_guest_oid)
-		&& is_member_of_role(role_oid, bbf_tempdb_guest_oid)
-		&& is_member_of_role(role_oid, bbf_msdb_guest_oid))
+	if ((pg_strcasecmp(role, BABELFISH_ROLE_ADMIN) == 0) ||
+			(is_bbf_admin_of_role(role_oid, bbf_admin_oid) && is_admin_of_role(bbf_admin_oid, role_oid)))
```
thephantomthief: Aren't these two conditions checking the same thing? The latter should suffice since it will check both direct and indirect membership `WITH ADMIN OPTION`

contrib/babelfishpg_tds/src/backend/tds/tdsutils.c
thephantomthief: Do we also need to pass bbf_role_admin to function? We can find the oid inside function itself by doing `get_role_oid(BABELFISH_ROLE_ADMIN, true);`

shalinilohia50: > What happens if customer granted babelfish role to a postgres endpoint created role, now it is possible that babelfish role has some other admin other then bbf_role_admin.

We have blocked GRANT on/to any BBF roles via PG with this change. 

`member_and_direct_admin_of_role` It will check if admin_of(i.e., bbf_role_oid) is member of the given roleid and its direct admin. This will be true for BBF users via PG port.

contrib/babelfishpg_tds/src/backend/tds/tdsutils.c
shalinilohia50: We are checking direct members of a role. BBF role will only have bbf_role_admin as a member and direct admin of role. PG role will have login as a member and direct admin of role.

PR #3402, opened 2025-01-14
Fix IDENTITY_INSERT to work cross-db

thephantomthief:
### Description

Previously, passing three part object names to SET IDENTITY_INSERT was
throwing an error. In this commit, if the relation name is a three part
name, we will parse the catalog name and set the IDENTITY_INSERT
property for the relation in that catalog.

Also with this commit, we will check if current user has permission to SET
IDENTITY_INSERT for the relation.

Task: BABEL-3212

* **Use case based -** Yes

* **Arbitrary inputs -** Yes

* **Performance tests -** N/A

* **Tooling impact -** N/A

* **Client tests -** N/A

### Check List


test/JDBC/expected/BABEL-IDENTITY.out
```diff
@@ -1208,7 +1208,7 @@ Index Only Scan using babel_3384_test_pkey on babel_3384_test
 
 ~~START~~
 text
-Babelfish T-SQL Batch Parsing Time: 0.159 ms
+Babelfish T-SQL Batch Parsing Time: 0.154 ms
```
shalinilohia50: Can't these values change?

thephantomthief: Yes they can but the test framework ignores the Batch Parsing time when taking the diff. See: https://github.com/babelfish-for-postgresql/babelfish_extensions/blob/BABEL_5_X_DEV/test/JDBC/src/test/java/com/sqlsamples/TestQueryFile.java#L380

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -317,6 +317,19 @@ check_identity_insert(char** newval, void **extra, GucSource source)
...
+static void
+throw_error_for_identity_insert(char *database, char *schema, char* object)
+{
+	ereport(ERROR,
+			(errcode(ERRCODE_INVALID_PARAMETER_VALUE),
+			 errmsg("Cannot find the object \"%s%s%s%s%s\" because it does not exist or you do not have permissions.",
+				database ? database : "",
+				database ? "." : "",
+				schema ? schema : "",
+				schema ? "." : "",
```
HarshLunagariya, line 329: It will produce db1.obj1 instead of db1..obj1 when there is no schema name supplied.

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -362,40 +379,61 @@ assign_identity_insert(const char *newval, void *extra)
...
-			schema_name = (char *) lthird(elemlist);
+			logical_schema_name = (char *) lthird(elemlist);
 
 			if (cur_db_name)
 				schema_name = get_physical_schema_name(cur_db_name,
-													   schema_name);
+													   logical_schema_name);
+
+			/* If no schema name is provided, we should use default schema */
+			if (!schema_name)
```
HarshLunagariya, line 416: By this logic, We will be using default schema when user supplies wrong schema.

HarshLunagariya: Do we have any testcase covering it?

test/JDBC/expected/IDENTITY-INSERT-CROSS-DB.out
```diff
@@ -0,0 +1,278 @@
+-- tsql
+CREATE DATABASE identity_insert_db
+GO
+
```
HarshLunagariya: Can we avoid making this new file IDENTITY-INSERT-CROSS-DB? , instead I suggest moving this to BABEL-IDENTITY.

HarshLunagariya (changes requested):

contrib/babelfishpg_tsql/src/pl_handler.c
thephantomthief: Actually `SET IDENTITY_INSERT db1..obj1 ON/OFF` will throw `syntax error at or near ".."`

contrib/babelfishpg_tsql/src/pl_handler.c
thephantomthief: `get_physical_schema_name()` doesn't perform the catalog look up. The actual error will get thrown at https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/3402/files#diff-ce1101bab57d0e7f8b36f0df103ed3ae49f4b488e473ac87bf2320c32f2697e1R430. For test case see https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/3402/files#diff-c16577e6f7e7da2222539138bf10997966a06ded7b507a823fe31328fbfe27d4R146.

shalinilohia50 (approved):

contrib/babelfishpg_tsql/src/pltsql_utils.c
```diff
@@ -2461,6 +2461,23 @@ privilege_to_string(AclMode privilege)
...
+	if (!HeapTupleIsValid(tuple))
+		ereport(ERROR,
+				(errcode(ERRCODE_UNDEFINED_TABLE),
+				 errmsg("relation with OID %u does not exist", relation_oid)));
+	ownerId = ((Form_pg_class) GETSTRUCT(tuple))->relowner;
+	ReleaseSysCache(tuple);
+
+	return ownerId;
+}
+
```
tanscorpio7: We can use existing engine function `object_ownercheck` for checking if user is owner of table.

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -335,6 +348,10 @@ assign_identity_insert(const char *newval, void *extra)
 		char	   *id_insert_rel_name = NULL;
 		char	   *id_insert_schema_name = NULL;
 		char	   *cur_db_name;
+		char	   *catalog_name = NULL;
```
tanscorpio7, line 351: Why are we naming this as `catalog_name` ?

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -407,15 +445,30 @@ assign_identity_insert(const char *newval, void *extra)
...
+		 * For SET IDENTITY_INSERT, ALTER permission on relation is needed. In
+		 * Babelfish, current user has ALTER permission on relation if either:
+		 *
+		 * 1. User is owner of object
+		 * 2. User is member of db_ddladmin or db_owner database role
+		 */
+		if (!(has_privs_of_role(curr_user_id, get_pg_class_obj_owner_id(rel_oid)) ||
+			has_privs_of_role(curr_user_id, get_db_ddladmin_oid(cur_db_name, false)) ||
+			has_privs_of_role(curr_user_id, get_db_owner_oid(cur_db_name, false))))
+			throw_error_for_identity_insert(catalog_name, logical_schema_name, rel_name);
```
tanscorpio7: I think the ideal way to check permission is to figure out the logical db_id from the relation schema OID.
Relation schema OID --> belongs to which logical database --> now check permissions using this db_id.

tanscorpio7: The problem with the current approach is that it assumes relation (without three part name) will always resolve to a schema in the current logical database but that is not true since search path can contain schemas from other logical databases.
Also there could be case where users can refer another database object from a different db.
ex: 
CREATE DATABASE db1 & db1_db1
`SELECT * FROM db1.db1_s1.t`

tanscorpio7: It is possible to force scenarios like
1. Current db is `some_db_1` but search path contains schema from `some_other_db_2`
2. If we create database db1 & db1_db1, I can refer db1_db1 database object using db1.db1_schema_name.object.

tanscorpio7 (changes requested):

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -317,6 +317,19 @@ check_identity_insert(char** newval, void **extra, GucSource source)
...
+				database ? database : "",
+				database ? "." : "",
+				schema ? schema : "",
+				schema ? "." : "",
+				object)));
+}
+
 static void
 assign_identity_insert(const char *newval, void *extra)
 {
```
tanscorpio7, line 335: We need to move this to check hook. A GUC assign hook must never fail.
see src/backend/utils/misc/README
```
Note that there is no provision for a failure result code.  assign_hooks
should never fail except under the most dire circumstances, since a failure
may for example result in GUC settings not being rolled back properly during
transaction abort.  In general, try to do anything that could conceivably
fail in a check_hook instead, and pass along the results in an "extra"
struct, so that the assign hook has little to do beyond copying the data to
someplace.  This applies particularly to catalog lookups: any required
lookups must be done in the check_hook, since the assign_hook may be
executed during transaction rollback when lookups will be unsafe.
```

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -362,40 +379,61 @@ assign_identity_insert(const char *newval, void *extra)
 		option_flag = (char *) linitial(elemlist);
 		rel_name = (char *) lsecond(elemlist);
 
+		/* Use catalog name if provided */
+		if (list_length(elemlist) == 4)
```
tanscorpio7, line 383: We need to add test cases where schema name/table name has a full stop. Using `!SplitGUCList(input_string, '.', &elemlist)` does not make me feel good.

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -362,40 +379,61 @@ assign_identity_insert(const char *newval, void *extra)
...
+			 */
+			if (!curr_user_if_cross_db)
+				throw_error_for_identity_insert(catalog_name, logical_schema_name, rel_name);
+		}
+
 		/* Check the user provided schema value */
 		if (list_length(elemlist) >= 3)
 		{
-			schema_name = (char *) lthird(elemlist);
+			logical_schema_name = (char *) lthird(elemlist);
```
tanscorpio7, line 409: Let use add a test case with schema name missing `set identity_insert db1..t5 on`

tanscorpio7 (changes requested):

contrib/babelfishpg_tsql/src/pl_handler.c
tanscorpio7: Actually this one might not be a problem in this case since we are not actually using the GUC value anywhere and instead maintaining are own static variables.

contrib/babelfishpg_tsql/src/pl_handler.c
thephantomthief: database name is also commonly referred as catalog name

contrib/babelfishpg_tsql/src/pl_handler.c
thephantomthief: I had added them before but everything threw syntax error. I'll add them back.

contrib/babelfishpg_tsql/src/pl_handler.c
thephantomthief: Not sure how I can force (1). For (2) if there exists database `db1` and `db1_db1`, you cannot create schema `db1_dbo` in `db1` database. It will throw error `schema "db1_db1_dbo" already exists`

contrib/babelfishpg_tsql/src/pl_handler.c
thephantomthief: Yeah this is wrong. We need to remove the grammar from bison and set the GUC from ANTLR.

contrib/babelfishpg_tsql/src/pl_handler.c
tanscorpio7:
```
CREATE DATABASE db1

CREATE DATABASE db1_db1

USE db1_db1

Changed database context to 'db1_db1'.
CREATE SCHEMA s1

CREATE VIEW s1.v1 AS SELECT 'I AM IN DB1_DB1'

USE db1

Changed database context to 'db1'.
SELECT * FROM db1.db1_s1.v1

                                                                                                                                                                                                                                                                
I AM IN DB1_DB1                                                                                                                                                                                                                                                 

(1 rows affected)
```

tanscorpio7: For 1, easiest way is to set the search path yourself using set_config() and we internally do some set_config("Search_path", ..) that will have the same effect.

tanscorpio7: For 2
```
1> use master
2> go
Changed database context to 'master'.
1> create procedure sp_proc_in_master_dbo as select * from v1
2> go
1> create view v1 as select 'I am in master_dbo'
2> go
1> use tempdb
2> go
Changed database context to 'tempdb'.
1> sp_proc_in_master_dbo
2> go
                                                                                                                                                                                                                                                                
I am in master_dbo                                                                                                                                                                                                                                              

(1 rows affected)
```
This is just one case. I can force many other such cases where we pick an object which does not exists in current logical database when object is not schema qualified.

contrib/babelfishpg_tsql/src/pl_handler.c
tanscorpio7: This is problematic because customer can do [schema.name].table for set identity insert and we will incorrectly change identity insert value of some other logical databases table.

contrib/babelfishpg_tsql/src/pl_handler.c
tanscorpio7: Let keep the failing tests even if we are not fixing it right now. This will ensure any future changed does not crash here because of NULL `logical_schema_name`

contrib/babelfishpg_tsql/src/pl_handler.c
thephantomthief: I was actually thinking on lines of incorrectly resolving from object A to object B where both A and B exist. But you are talking about the case where A does not exist and we resolve to B. Will fix this.

HarshLunagariya: I was talking about following scenarios in our discussion, but it seems like INSERT statement is resolving it properly. We are good here:
```
1> create database [db1.sch1];
2> go
1> create database [db1_sch1];
2> go
1> use [db1_sch1];
2>
3> go
Changed database context to 'db1_sch1'.
1> create schema [sch2];
2> go
1> create table [sch2].t1(a int identity not null);
2> go
1> set identity_insert [db1.sch1].[sch2].t1 ON;  --> It should have thrown error since there is no such table in [db1.sch1_sch2]
2>
3> go
1> insert into [db1.sch1].[sch2].t1 values (1);
2> go
relation "db1.sch1_sch2.t1" does not exist
1>

```

tanscorpio7 (approved):

HarshLunagariya (approved):

shardgupta (approved):

PR #3405, opened 2025-01-15
Fix a corner case '=' not properly rewritten to AS

forestkeeper:
Previously we rewrite "SELECT COL=expr" to "SELECT expr as "COL"", but there's a corner case that for if condition, it didn't do the rewrite, because different parser path from if condition to dml statement.

This commit fix this issue by adding the mutator logic to if statement rewrite.

Task: BABEL-5320

### Check List

### DIFF
--- a/contrib/babelfishpg_tsql/src/tsqlIface.cpp
+++ b/contrib/babelfishpg_tsql/src/tsqlIface.cpp
@@ -2936,17 +2936,38 @@ public:
 		clear_rewritten_query_fragment();
 	}
 
-	PLtsql_expr *rewrite_if_condition(TSqlParser::Search_conditionContext *ctx)
+	PLtsql_expr *rewrite_if_condition(TSqlParser::Search_conditionContext *ctx, PLtsql_expr *expr)
 	{
-		PLtsql_expr *expr = makeTsqlExpr(ctx, false);
-		PLtsql_expr_query_mutator mutator(expr, ctx);
-		add_rewritten_query_fragment_to_mutator(&mutator);
-		mutator.run();
+		if (!expr)
+			expr = makeTsqlExpr(ctx, false);
+		if (statementMutator)
+		{
+			add_rewritten_query_fragment_to_mutator(statementMutator.get());
+			statementMutator->run();
+			statementMutator = nullptr;
+		}
+		else
+		{
+			PLtsql_expr_query_mutator mutator(expr, ctx);
+			add_rewritten_query_fragment_to_mutator(&mutator);
+			mutator.run();
+		}
 		clear_rewritten_query_fragment();
 
 		/* Now we can prepend SELECT to rewritten search_condition */
 		expr->query = strdup((std::string("SELECT ") + std::string(expr->query)).c_str());
 		return expr;
+	}
+
+	void enterSearch_condition(TSqlParser::Search_conditionContext *ctx) override
+	{
+		if (((TSqlParser::Cfl_statementContext *) ctx->parent->parent)->if_statement())
+		{
+			PLtsql_stmt_if *fragment = (PLtsql_stmt_if *) getPLtsql_fragment(ctx->parent->parent);
+			fragment->cond = makeTsqlExpr(ctx, false);
+			clear_rewritten_query_fragment();
+			statementMutator = std::make_unique<PLtsql_expr_query_mutator>(fragment->cond, ctx);
+		}
 	}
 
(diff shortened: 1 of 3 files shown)


test/JDBC/expected/BABEL-2514.out
```diff
@@ -197,3 +197,9 @@ int#!#int
 
 drop view if exists babel_2514_complex_view;
 go
+
+IF EXISTS 
+(
+SELECT tblnam = a.name FROM sysobjects a where a.name = 'tablename'
+) PRINT 'Exists'; ELSE PRINT 'Not Exists';
+GO
```
Deepesh125: can you add test cases where we have similar statement inside procedure?

Deepesh125: Also, can you add tests where if condition itself gets re-written? For example, 
```
IF EXISTS(
select trim(col) = '' from tab1 where ...
) PRINT 'Exists'; ELSE PRINT 'Not Exists';

File: contrib/babelfishpg_tsql/src/tsqlIface.cpp [hunk 2]
```diff
@@ -2936,34 +2936,54 @@ class tsqlBuilder : public tsqlCommonMutator
...
 		clear_rewritten_query_fragment();
 
 		/* Now we can prepend SELECT to rewritten search_condition */
 		expr->query = strdup((std::string("SELECT ") + std::string(expr->query)).c_str());
 		return expr;
 	}
 
+	void enterSearch_condition(TSqlParser::Search_conditionContext *ctx) override
+	{
+		if (((TSqlParser::Cfl_statementContext *) ctx->parent->parent)->if_statement())
```
KushaalShroff, line 2964: Need Sanity check?
```
		if (!ctx->parent || !ctx->parent->parent)
			return;
```

contrib/babelfishpg_tsql/src/tsqlIface.cpp
```diff
@@ -2936,34 +2936,54 @@ class tsqlBuilder : public tsqlCommonMutator
...
+			mutator.run();
+		}
 		clear_rewritten_query_fragment();
 
 		/* Now we can prepend SELECT to rewritten search_condition */
 		expr->query = strdup((std::string("SELECT ") + std::string(expr->query)).c_str());
 		return expr;
 	}
 
+	void enterSearch_condition(TSqlParser::Search_conditionContext *ctx) override
```
KushaalShroff, line 2962: Didnt get why we need this change?

contrib/babelfishpg_tsql/src/tsqlIface.cpp
KushaalShroff: I think we should use dynamic_cast function instead of directly casting it

contrib/babelfishpg_tsql/src/tsqlIface.cpp
```diff
@@ -2936,34 +2936,54 @@ class tsqlBuilder : public tsqlCommonMutator
...
+	{
+		if (((TSqlParser::Cfl_statementContext *) ctx->parent->parent)->if_statement())
+		{
+			PLtsql_stmt_if *fragment = (PLtsql_stmt_if *) getPLtsql_fragment(ctx->parent->parent);
```
KushaalShroff, line 2966: Might be a naive question, I see there are multiple possible parent->parent of search_condition like
```
table_sources->table_source_item->search_condition
```
So are we saying that only in few conditions there will be parse tree path to if_condition which will always be parent->parent?

test/JDBC/expected/BABEL-2514.out
xhfanhe, line 205: Why is it not printing anything?

forestkeeper: print cmd is not working in jdbc framework

xhfanhe (approved):

contrib/babelfishpg_tsql/src/tsqlIface.cpp
```diff
@@ -2936,34 +2936,54 @@ class tsqlBuilder : public tsqlCommonMutator
...
+			fragment->cond = makeTsqlExpr(ctx, false);
+			clear_rewritten_query_fragment();
```
Deepesh125, line 2968: why are we calling  clear_rewritten_query_fragment  without calling add_rewritten_query_fragment_to_mutator?

PR #3419, opened 2025-01-21
Fix to update suser_sname behaviour with invalid input

pranavJ23:
### Description
In suser_sname() function if we run the query 
```
select suser_sname(0x0)  
```
then it was giving the session user which is not correct as expected output is NULL
Fix the above issue by making changes in the function suser_name where we were setting the user_id to session_user_id if input is invalid, changed it to return null if input is invalid. This fixes the above problem.

### Issues Resolved
[BABEL-5469]
### Check List


contrib/babelfishpg_tsql/src/rolecmds.c
```diff
@@ -968,6 +968,39 @@ get_original_login_name(char *login)
...
 
+PG_FUNCTION_INFO_V1(suser_name_current_session);
+Datum
+suser_name_current_session(PG_FUNCTION_ARGS)
+{
+	Oid			server_user_id;
+	char	   *ret;
+	char	   *orig_loginname;
+
+	server_user_id = GetSessionUserId();
```
anju15bharti: why can't we utilize the existing API to handle this case as well?

anju15bharti (changes requested):

contrib/babelfishpg_tsql/src/rolecmds.c
pranavJ23: In the existing API we were handelling both NULL and empty input as invalid input so unable distinguish there.

anju15bharti (approved):

anju15bharti (approved):
As discussed Offline, changes look good after removing the extra API that was not needed.

HarshLunagariya (approved):

shardgupta (dismissed):

shardgupta (approved):

PR #3420, opened 2025-01-21
Add graceful handling in some hooks when temp name space is deleted by concurrent postgres sessions

tanscorpio7:
### Description

A couple of hooks introduced as part of supporting fixed database level roles assumed that the namespace for an object will always exists, but an exception to this is temp namespace which can be deleted by a concurrent postgres sessions.

As a fix completely skip the hook logic if we are processing a temp namespace and add safety null checks.

### Issues Resolved

[BABEL-5569]

### Sign Off

### Check List

### DIFF
--- a/contrib/babelfishpg_tsql/src/hooks.c
+++ b/contrib/babelfishpg_tsql/src/hooks.c
@@ -5801,13 +5801,19 @@ pltsql_get_object_owner(Oid namespaceId, Oid ownerId)
 
 	Assert(OidIsValid(namespaceId));
 
-	if (sql_dialect != SQL_DIALECT_TSQL || !IS_TDS_CONN())
+	if (sql_dialect != SQL_DIALECT_TSQL || !IS_TDS_CONN() || isTempNamespace(namespaceId))
 		return ownerId;
 
 	if (!OidIsValid(ownerId))
 		ownerId = GetUserId();
 
 	tuple = SearchSysCache1(NAMESPACEOID, ObjectIdGetDatum(namespaceId));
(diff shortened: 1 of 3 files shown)


HarshLunagariya (commented):
1. Do we have existing testcases about verifying the ownership/access of temp tables?
2. Can you please add test scenario in description? If we cannot add the automated tests.

tanscorpio7:
> 1. Do we have existing testcases about verifying the ownership/access of temp tables?

No, but it should not be a problem since we skip every schema that does not have an entry in babelfish namespace catalog.

> 2. Can you please add test scenario in description? If we cannot add the automated tests.

I will add one test cases for this. Not too sure about dropping temp namespaces from a different sessions.

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -5801,13 +5801,19 @@ pltsql_get_object_owner(Oid namespaceId, Oid ownerId)
 
 	Assert(OidIsValid(namespaceId));
 
-	if (sql_dialect != SQL_DIALECT_TSQL || !IS_TDS_CONN())
+	if (sql_dialect != SQL_DIALECT_TSQL || !IS_TDS_CONN() || isTempNamespace(namespaceId))
```
Deepesh125: we need to verify that normal user (w/o superuser privilege) should not able drop temp schema. Please add such test cases

HarshLunagariya (approved):

shardgupta (approved):

PR #3429, opened 2025-01-24
Store original name for Babelfish indexes

rishabhtanwar29:
### Description
In Babelfish, index names are generated by first calculating hash of given original index name
and table name concatenated together and then concatenating index name, table name and
this hash together once again to form a 64 character long identifier.
But this name is not intuitive for users as it looks different from what user specified while
creating the index.
To address this issue, this commit adds ability to store original index name as a reloption for a
given index and show that stored original name in the catalog views, like sys.indexes.
This CR extends this solution to also store original index names for partitioned indexes.
Additionally, support renaming of indexes using `sp_rename` so that existing indexes can be
renamed so that their original names get stored in the catalog.
 
Task: BABEL-1517

* **Use case based -**

* **Boundary conditions -**

* **Arbitrary inputs -**

* **Negative test cases -**

* **Performance tests -**

* **Tooling impact -**

* **Client tests -**

### Check List


contrib/babelfishpg_tsql/sql/babelfishpg_tsql.sql
```diff
@@ -3104,6 +3107,24 @@ BEGIN
 						END
 					SET @currtype = 'CO';
 				END
+			ELSE IF @objtype = 'INDEX'
+				BEGIN
+					DECLARE @relid INT = 0;
+					DECLARE @index_count INT;
+					SELECT @relid = object_id FROM sys.objects o1 INNER JOIN sys.schemas s1 ON o1.schema_id = s1.schema_id 
+						WHERE s1.name = @schemaname AND o1.name = @curr_relname;
```
KushaalShroff, line 3115: babelfish_sp_rename_word_parse is making sure these are not NULL for sure, right?

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -4341,30 +4334,28 @@ bbf_ProcessUtility(PlannedStmt *pstmt,
 			}
 		case T_IndexStmt:
 			{
-				if (sql_dialect == SQL_DIALECT_TSQL)
+				IndexStmt	*stmt = (IndexStmt *) parsetree;
+
+				if (sql_dialect == SQL_DIALECT_TSQL &&
+					strcmp(queryString, "(CREATE FULLTEXT INDEX STATEMENT )") != 0) /* Skip fulltext indexes since they don't even have an original name */
```
KushaalShroff: Create macro like:
```
/* BBF SUBCOMMANDS QUERY STRING */
#define CREATE_LOGICAL_DATABASE "(CREATE LOGICAL DATABASE )"
#define CREATE_GUEST_SCHEMAS_DURING_UPGRADE "(CREATE GUEST SCHEMAS DURING UPGRADE )"
#define CREATE_FIXED_DB_ROLES "(CREATE FIXED DATABASE ROLES )"
#define ALTER_DEFAULT_PRIVILEGES "(ALTER DEFAULT PRIVILEGES )"
#define INTERNAL_GRANT_STATEMENT "(GRANT STATEMENT )"
#define INTERNAL_REVOKE_ALL_ON_ROUTINE "(REVOKE ALL ON ROUTINE )"
#define INTERNAL_ALTER_ROLE "(ALTER ROLE ADD )"
```

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--5.0.0--5.1.0.sql
```diff
@@ -5559,6 +5559,342 @@ LANGUAGE plpgsql
...
+END;
+$$
+LANGUAGE 'pltsql';
+
+CREATE OR REPLACE PROCEDURE sys.sp_renamedb(
+	IN "@objname" sys.SYSNAME,
+	IN "@newname" sys.SYSNAME
+)
+AS 'babelfishpg_tsql', 'sp_renamedb_internal'
+LANGUAGE C;
```
KushaalShroff: Why is this needed?

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -3786,6 +3787,11 @@ sp_rename_internal(PG_FUNCTION_ARGS)
 			objtype_code = OBJECT_TABLE;
 			process_util_querystr = "(ALTER TABLE )";
 		}
+		else if (strcmp(objtype, "IX") == 0)
+		{
+			objtype_code = OBJECT_INDEX;
+			process_util_querystr = "(ALTER INDEX )";
```
KushaalShroff, line 3793: Macro might be better

contrib/babelfishpg_tsql/sql/sys_views.sql
```diff
@@ -834,7 +834,16 @@ with index_id_map as MATERIALIZED(
...
+  , cast(
+		coalesce(
+			(select pg_catalog.string_agg(
+				case
+					when option like 'bbf_original_rel_name=%' then substring(option, 23 /* prefix length */)
+					else null
+				end, ',')
+			from unnest(I.reloptions) as option),
+			I.relname)
+		AS sys.sysname) AS name
```
KushaalShroff, line 846: Impact on performance? IK this is not avoidable, just curious

test/JDBC/expected/BABEL_4817-vu-verify.out
```diff
@@ -4,6 +4,15 @@ GO
 ALTER TABLE babel_4817_t3 ADD col4 INT
 GO
 
+sp_rename 'babel_4817_t1.babel_4817_t1_idx_1', 'babel_4817_t1_idx_2', 'INDEX'
```
KushaalShroff, line 7: Can we test indexname > 64 len

test/JDBC/expected/PARTITION-vu-verify.out
```diff
@@ -1009,7 +1069,7 @@ varchar#!#int#!#int
 363863941f079adaa9aa733200e57c9f_partition_3_value_idx#!#4#!#3
 363863941f079adaa9aa733200e57c9f_partition_3#!#4#!#0
 363863941f079adaa9aa733200e57c9f_partition_3_id_key#!#4#!#2
-partition_vu_prepare_normal_inded6315af14d88f45711ba24e46851b8f#!#1#!#2
+partition_vu_prepare_normal_ind5f5fcd81a18b69e8a21aef69ce61f7b6#!#1#!#2
```
KushaalShroff, line 1072: 1. How is partition_id getting changed with renaming index? Looks like an existing bug with sys.partitions. Please check with @sumitj824 
2. Also looking at this output, Eventually we should fix OBJECT_NAME to give the original name instead of hashed name.

KushaalShroff (changes requested):

contrib/babelfishpg_tsql/sql/babelfishpg_tsql.sql
rishabhtanwar29: Yep.

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--5.0.0--5.1.0.sql
rishabhtanwar29: Added by mistake.

contrib/babelfishpg_tsql/src/procedures.c
rishabhtanwar29: We are already having such hardcoded command names for all other kind of objects in this function. Moreover, macro is mostly useful when it is being used at multiple places, which is not true for this case so I think we can keep it hardcoded for now.

test/JDBC/expected/PARTITION-vu-verify.out
rishabhtanwar29: For 2nd, object_name/object_id functions only work for schema scoped functions but index is a table scoped object in T-SQL so these functions  are not actually meant to handle indexes.

test/JDBC/expected/BABEL_4817-vu-verify.out
rishabhtanwar29: Added test in Test-sp_rename-vu-verify file

test/JDBC/expected/PARTITION-vu-verify.out
rishabhtanwar29: For 1st, the output is expected since sys.partitions returns entry for both tables and indexes and index `partition_vu_prepare_normal_index` has been removed to `partition_vu_prepare_normal_index_new` in the beginning using sp_rename so partition_id has also been changed.

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -4345,30 +4338,28 @@ bbf_ProcessUtility(PlannedStmt *pstmt,
...
 		case T_IndexStmt:
 			{
-				if (sql_dialect == SQL_DIALECT_TSQL)
+				IndexStmt	*stmt = (IndexStmt *) parsetree;
+
+				if (sql_dialect == SQL_DIALECT_TSQL &&
+					strcmp(queryString, CREATE_FULLTEXT_INDEX) != 0) /* Skip fulltext indexes since they don't even have an original name */
 				{
-					IndexStmt *stmt = (IndexStmt *) parsetree;
+					char    	*original_name = stmt->idxname != NULL ? pstrdup(stmt->idxname) : NULL;
```
tanscorpio7, line 4346: Wouldn't parser already chop down the name to less than 64 ?

tanscorpio7 (changes requested):

contrib/babelfishpg_tsql/src/pl_handler.c
rishabhtanwar29: Yep but that truncation is less frequent compared to the current case where we are always generating complex names. In theory we can actually fetch the original name from the query string but we would require name location for that and currently IndexStmt doesn't have a location field and right now I am quite hesitant in changing a Node just for this unless there is some actual value in future.

tanscorpio7: makes sense, I think we are not even doing this for tables currently.

KushaalShroff (approved):

PR #3439, opened 2025-01-30
Update logic for LIKE operator for Case Insensitive collation

ahmed-shameem:
### Description

-- CASE 1
```
```

-- CASE 2
```
```

```
expression LIKE pattern
```

```
expression = pattern
```

```
```

Currently we have observed that we do not comply with the transformation above. Rather than COLLATE CS_AS, we observe COLLATE “default” in query plan. Consider the following query:

```
select a from varchar_tst where a collate chinese_prc_ci_as LIKE 'ab%'
```

-- query plan
```
Filter: (((a)::text ~~* 'ab%'::text COLLATE "default") 
AND ((a)::text >= 'ab'::text COLLATE "default") AND ((a)::text < 'ab?'::text))
```

Even though we observe COLLATE "default" in the query plan, it picks up the correct collation oid during execution. 
However, if we update the collation oid in typeTypeCollation function to pickup database collation oid for string literals in parse tree, we may observe upgrade/restore failure.

We use this task as first step to modify and incorporate improvements for such cases.

### Issues Resolved

Task: BABEL-5608

* **Use case based -**

* **Boundary conditions -**

* **Arbitrary inputs -**

* **Negative test cases -**

* **Performance tests -**

* **Tooling impact -**

* **Client tests -**

### Check List


contrib/babelfishpg_tsql/src/collation.c
```diff
@@ -409,14 +413,14 @@ transform_from_ci_as_for_likenode(Node *node, OpExpr *op, like_ilike_info_t like
...
-		optup = compatible_oper(NULL, list_make1(makeString(op_str)), ltypeId, ltypeId,
+		optup = compatible_oper(NULL, list_make1(makeString(op_str)), ltypeId, rtypeId,
 								true, -1);
 		if (optup == (Operator) NULL)
 			return node;
 
 		ret = (Node *) (make_op_with_func(oprid(optup), BOOLOID, false,
 											(Expr *) leftop, (Expr *) prefix,
-											InvalidOid, coll_info_of_inputcollid.oid, oprfuncid(optup)));
+											coll_info_of_inputcollid.oid, coll_info_of_inputcollid.oid, oprfuncid(optup)));
```
Deepesh125: upon further investigation, I found that operator that returns bool will have InvalidOid as opcollid. It does not matter collation of operator result.

contrib/babelfishpg_tsql/src/collation.c
```diff
@@ -392,11 +392,15 @@ transform_from_ci_as_for_likenode(Node *node, OpExpr *op, like_ilike_info_t like
...
 	patt = (Const *) rightop;
+	patt->constcollid = op->inputcollid;
 
 	/* extract pattern */
 	pstatus = pattern_fixed_prefix_wrapper(patt, 1, coll_info_of_inputcollid.oid,
 											&prefix, NULL);
 
+	/* update the collation for the prefix as well */
+	prefix->constcollid = coll_info_of_inputcollid.oid;
+
```
Deepesh125: Lets not touch it. collation of const would anyway be DEFAULT_COLLATION_OID assigned within pattern_fixed_prefix_wrapper. And we know for sure that either op or left hand operand will have correct collation. So lets not touch it.

ahmed-shameem: This is needed or else COLLATE "default" will appear for >= operator

contrib/babelfishpg_tsql/src/collation.c
ahmed-shameem: Right, updated

test/JDBC/expected/BABEL-5608-vu-prepare.out
```diff
@@ -0,0 +1,238 @@
...
+text
+CHECK ((((babel_5608_vu_prepare_t6_col1)::text COLLATE bbf_unicode_cp1_cs_as) ~~* ((babel_5608_vu_prepare_t6_col2)::text COLLATE bbf_unicode_cp1_cs_as)))
+~~END~~
+
+
+-- CASE 7: T_CollateExpr(T_ReLabel(T_Var)) LIKE T_Const
+CREATE TABLE BABEL_5608_vu_prepare_t7_1(BABEL_5608_vu_prepare_t7_1_col1 varchar(max), CONSTRAINT BABEL_5608_vu_prepare_t7_1_c1 CHECK (BABEL_5608_vu_prepare_t7_1_col1 COLLATE Latin1_General_CI_AI LIKE 'abc'));
+GO
+
+SELECT pg_get_constraintdef(oid) FROM pg_constraint WHERE conrelid=object_id('BABEL_5608_vu_prepare_t7_1');
```
Deepesh125: I think this should be part of verify script as well. it will ensure that we have expected constraint before and after upgrade

contrib/babelfishpg_tsql/src/collation.c
```diff
@@ -846,20 +875,34 @@ convert_node_to_funcexpr_for_like(Node *node)
...
 
 static Node *
-transform_likenode_for_AI(Node *node, OpExpr *op)
+transform_likenode_for_AI(Node *node, OpExpr *op, coll_info_t coll_info_of_inputcollid)
 {
 	Node		*leftop = (Node *) linitial(op->args);
 	Node		*rightop = (Node *) lsecond(op->args);
+	Oid			colloid_of_cs_as;
+
+	/* Find the CS_AS collation corresponding to the CS_AI collation */
```
Deepesh125: CS_AS collation corresponding to CS_AI or CI_AI

contrib/babelfishpg_tsql/src/collation.c
```diff
@@ -784,6 +812,7 @@ convert_node_to_funcexpr_for_like(Node *node)
 					if (con->constisnull)
 						return new_node;
 					con->constvalue = DirectFunctionCall1(remove_accents_internal, con->constvalue);
+					con->constcollid = colloid_of_cs_as;
```
Deepesh125: why do we only care for Const? is it because of type would be text?

test/JDBC/expected/BABEL-5608-vu-prepare.out
```diff
@@ -0,0 +1,238 @@
...
+SELECT pg_get_constraintdef(oid) FROM pg_constraint WHERE conrelid=object_id('BABEL_5608_vu_prepare_t10');
+GO
+~~START~~
+text
+CHECK (((((babel_5608_vu_prepare_t10_col1)::text COLLATE bbf_unicode_cp1_ci_as) COLLATE bbf_unicode_cp1_cs_as) ~~* (like_escape('15/% %'::text COLLATE "default", '/'::text COLLATE "default") COLLATE bbf_unicode_cp1_cs_as)))
+~~END~~
+
+
+-- CASE 11: T_CoerceViaIO
+CREATE TABLE BABEL_5608_vu_prepare_t11(BABEL_5608_vu_prepare_t11_col1 varchar(max), CONSTRAINT BABEL_5608_vu_prepare_t11_c1 CHECK (N'123' collate Latin1_General_CI_AI LIKE CAST(123 as varchar(3))));
```
Deepesh125: lets also have some cases which involves VIEW

Deepesh125 (changes requested):

test/JDBC/expected/BABEL-5608-vu-prepare.out
Deepesh125: Also have some test cases where LIKE apears in CFL condition ie, case when, if else, etc

contrib/babelfishpg_tsql/src/collation.c
```diff
@@ -305,6 +305,15 @@ transform_funcexpr(Node *node)
 	return node;
 }
 
+static CollateExpr*
+create_collate_expr(Node *arg, Oid collid)
+{
+	CollateExpr *expr = makeNode(CollateExpr);
+	expr->arg = (Expr *) arg;
+	expr->collOid = collid;
```
forestkeeper, line 313: pls set expr location = -1, if won't use.

contrib/babelfishpg_tsql/src/collation.c
```diff
@@ -428,11 +448,18 @@ transform_from_ci_as_for_likenode(Node *node, OpExpr *op, like_ilike_info_t like
 		Node	   *constant_suffix;
 		Const	   *highest_sort_key;
 
+		/* Always create a CollateExpr on top to match with op->inputcollid */
+		new_leftop->arg = linitial(op->args);
+		new_leftop->collOid = op->inputcollid;
```
forestkeeper: Why don't you reuse create_collate_expr in here ?

contrib/babelfishpg_tsql/src/collation.c
```diff
@@ -428,11 +448,18 @@ transform_from_ci_as_for_likenode(Node *node, OpExpr *op, like_ilike_info_t like
...
+		linitial(op->args) = (Node*) new_leftop;
+
 		/* construct leftop >= pattern */
-		optup = compatible_oper(NULL, list_make1(makeString(">=")), ltypeId, ltypeId,
+		optup = compatible_oper(NULL, list_make1(makeString(">=")), ltypeId, rtypeId,
```
forestkeeper, line 471: What's the purpose of change ltypeId to rtypeId in here ?

contrib/babelfishpg_tsql/src/collation.c
```diff
@@ -388,18 +398,28 @@ transform_from_ci_as_for_likenode(Node *node, OpExpr *op, like_ilike_info_t like
 	if (IsA(leftop, Const) || !IsA(rightop, Const) ||
 		((Const *) rightop)->constisnull)
```
kuntalghosh, line 399: Why do we need to explicitly collate even if the right operand is NULL? How does NULL value work w.r.t collations

contrib/babelfishpg_tsql/src/collation.c
ahmed-shameem: Updated in lates rev

contrib/babelfishpg_tsql/src/collation.c

contrib/babelfishpg_tsql/src/collation.c
ahmed-shameem: the leftop of >= is left node typeid and right op of >= should be right node typeid. We have observed both of these are same, however for clarity made this change

contrib/babelfishpg_tsql/src/collation.c
ahmed-shameem: There is no need to add explicit collation for NULL values.  There is no functional difference between NULL and NULL COLLATE latin1_general_ci_ai. The sort order is also determined by NULLS FIRST/LAST and NOT collation. I have removed the creation of new CollateExpr on NULL in latest rev anyway.

contrib/babelfishpg_tsql/src/collation.c
ahmed-shameem: This is redundant, initially added this for adding CS_AS for pattern for AI collations. However, we take care of that in transform_from_ci_as_for_likenode. Removed in latest rev.

contrib/babelfishpg_tsql/src/collation.c
ahmed-shameem: CI_AI, however the latest changes does not need it hence removed it

test/JDBC/expected/BABEL-5608-vu-prepare.out
ahmed-shameem: Updated in latest rev

test/JDBC/expected/BABEL-5608-vu-prepare.out
ahmed-shameem: Added in latest rev

contrib/babelfishpg_tsql/src/collation.c
forestkeeper: Can you add an assert that those two are the same in here ?

ahmed-shameem: Added Assert. Also, want to correct my previous statement, the operands can be of different datatypes like text LIKE name. However, for the optimisation part, both of them would be TEXT

test/JDBC/expected/BABEL-5608-vu-prepare.out
rishabhtanwar29, line 1: We should create test files with some meaningful name instead of using bare task ids.

test/JDBC/upgrade/latest/schedule
```diff
@@ -607,3 +607,4 @@ sys_sql_logins
 sys-login-property
 sys-fn-varbintohexsubstring
 BABEL_OBJECT_RESOLUTION_IN_ROUTINES
+BABEL-5608
```
forestkeeper: BABEL-5608 is only schedule in the latest version upgrade test ? Why ?

ahmed-shameem: Added in other upgrade paths as well

contrib/babelfishpg_tsql/src/collation.c
```diff
@@ -400,23 +414,45 @@ transform_from_ci_as_for_likenode(Node *node, OpExpr *op, like_ilike_info_t like
...
+	 * which is adding extra parenthesis on rightop when we invoke pg_get_constraintdef() from PG
+	 * We update the righop to equivalent CollateExpr to pick correct collation
+	 * We are clearing collation or else we observe multiple redundant COLLATE
+	 * clause in pg_get_constraintdef(), which will result in error during upgrade/restore
+	 * We also set InvalidOid for highest_sort_key during creation for the same reason,
+	 * later we enclose it withing CollateExpr
+	 */
+	prefix->constcollid = ((Const *) rightop)->constcollid = InvalidOid;
+	
+	prefix_collate = create_collate_expr((Node* ) prefix, coll_info_of_inputcollid.oid);
```
Deepesh125, line 434: can you give an example on this?

contrib/babelfishpg_tsql/src/collation.c
```diff
@@ -428,33 +464,46 @@ transform_from_ci_as_for_likenode(Node *node, OpExpr *op, like_ilike_info_t like
...
 		optup = compatible_oper(NULL, list_make1(makeString("||")), rtypeId, rtypeId,
 								true, -1);
 		if (optup == (Operator) NULL)
 			return node;
+
 		concat_expr = make_op_with_func(oprid(optup), rtypeId, false,
-										(Expr *) prefix, (Expr *) highest_sort_key,
-										InvalidOid, coll_info_of_inputcollid.oid, oprfuncid(optup));
+										(Expr *) prefix_collate,
+										(Expr *) create_collate_expr((Node* ) highest_sort_key, coll_info_of_inputcollid.oid),
```
Deepesh125, line 495: Not getting why do we need collateExpr for highest_sort_key?

contrib/babelfishpg_tsql/src/collation.c
```diff
@@ -846,7 +895,7 @@ convert_node_to_funcexpr_for_like(Node *node)
 
 
 static Node *
-transform_likenode_for_AI(Node *node, OpExpr *op)
+transform_likenode_for_AI(Node *node, OpExpr *op, coll_info_t coll_info_of_inputcollid)
```
Deepesh125: where do we need coll_info_of_inputcollid?

contrib/babelfishpg_tsql/src/collation.c
ahmed-shameem: Consider the following diff
```
-- exact match
-CHECK (((babel_5608_vu_prepare_t4_1_col1)::text = 'ABC'::text COLLATE sys.bbf_unicode_cp1_ci_as))
+CHECK (((babel_5608_vu_prepare_t4_1_col1)::text = ('ABC'::text COLLATE sys.bbf_unicode_cp1_ci_as)))
```
Extra parenthesis is being added on rightop after restore
```
-- prefix pattern match
-CHECK (((((babel_5608_vu_prepare_t4_2_col1)::text COLLATE sys.bbf_unicode_cp1_cs_as) ~~* 'a%'::text COLLATE sys.bbf_unicode_cp1_cs_as) AND (((babel_5608_vu_prepare_t4_2_col1)::text >= 'a'::text COLLATE sys.bbf_unicode_cp1_ci_as) AND ((babel_5608_vu_prepare_t4_2_col1)::text < ('a'::text COLLATE sys.bbf_unicode_cp1_ci_as || ''::text COLLATE sys.bbf_unicode_cp1_ci_as)))))
+CHECK (((((babel_5608_vu_prepare_t4_2_col1)::text COLLATE sys.bbf_unicode_cp1_cs_as) ~~* ('a%'::text COLLATE sys.bbf_unicode_cp1_cs_as)) AND (((babel_5608_vu_prepare_t4_2_col1)::text >= ('a'::text COLLATE sys.bbf_unicode_cp1_ci_as)) AND ((babel_5608_vu_prepare_t4_2_col1)::text < (('a'::text COLLATE sys.bbf_unicode_cp1_ci_as) || (''::text COLLATE sys.bbf_unicode_cp1_ci_as))))))
```

This is because while querying the pg_get_constraintdef() from PG endpoing, the dump is as expected but as the Const node has explicit COLLATE, during restore while rebuilding the parse tree that node is being considered as a CollateExpr which results adds extra parenthesis

contrib/babelfishpg_tsql/src/collation.c
```diff
@@ -400,23 +414,45 @@ transform_from_ci_as_for_likenode(Node *node, OpExpr *op, like_ilike_info_t like
 	/* If there is no constant prefix then there's nothing more to do */
 	if (pstatus == Pattern_Prefix_None)
 	{
+		/* update the collation of left and right node*/
+		linitial(op->args) = (Node *) create_collate_expr(linitial(op->args), op->inputcollid);
+		lsecond(op->args) = (Node *) create_collate_expr(lsecond(op->args), op->inputcollid);
 		return node;
```
Deepesh125, line 419: can you plz point me the example/test cases for this?

test/JDBC/expected/db_collation/test_constraint_like-vu-prepare.out
```diff
@@ -0,0 +1,388 @@
...
+CREATE VIEW BABEL_5608_13_5_VIEW AS SELECT * FROM BABEL_5608_vu_prepare_t12_2 WHERE UPPER(status) COLLATE DATABASE_DEFAULT LIKE 'ac%';
+GO
+
+-- psql
+-- CASE 1: T_Const LIKE T_CollateExpr(T_Const)
+SELECT pg_get_constraintdef(oid) FROM pg_constraint WHERE conrelid = 'master_dbo.BABEL_5608_vu_prepare_t1'::regclass;
+GO
+~~START~~
+text
+CHECK ((('abc'::text COLLATE sys.bbf_unicode_cp1_cs_as) ~~* (('A%'::text COLLATE sys.bbf_unicode_cp1_ci_as) COLLATE sys.bbf_unicode_cp1_cs_as)))
```
Deepesh125, line 239: IMO, we shouldn't be getting `COLLATE sys.bbf_unicode_cp1_ci_as` on rightop

test/JDBC/expected/db_collation/test_constraint_like-vu-prepare.out
```diff
@@ -0,0 +1,388 @@
...
+~~END~~
+
+
+-- CASE 2: T_CollateExpr(T_Const) LIKE T_Const
+SELECT pg_get_constraintdef(oid) FROM pg_constraint WHERE conrelid = 'master_dbo.BABEL_5608_vu_prepare_t2'::regclass;
+GO
+~~START~~
+text
+CHECK ((((('abc'::text COLLATE sys.bbf_unicode_cp1_ci_as) COLLATE sys.bbf_unicode_cp1_cs_as) ~~* ('A%'::text COLLATE sys.bbf_unicode_cp1_cs_as)) AND ((('abc'::text COLLATE sys.bbf_unicode_cp1_ci_as) >= ('A'::text COLLATE sys.bbf_unicode_cp1_ci_as)) AND (('abc'::text COLLATE sys.bbf_unicode_cp1_ci_as) < (('A'::text COLLATE sys.bbf_unicode_cp1_ci_as) || ('￿'::text COLLATE sys.bbf_unicode_cp1_ci_as))))))
```
Deepesh125, line 248: why are we apply transformation rule when both sides are Const nodes? it doesnt make sense

test/JDBC/expected/db_collation/test_constraint_like-vu-prepare.out
```diff
@@ -0,0 +1,388 @@
...
+-- CASE 5: T_ReLabelType(T_Var) LIKE T_CollateExpr(T_Const)
+SELECT pg_get_constraintdef(oid) FROM pg_constraint WHERE conrelid = 'master_dbo.BABEL_5608_vu_prepare_t5_1'::regclass;
+GO
+~~START~~
+text
+CHECK ((((sys.remove_accents_internal((babel_5608_vu_prepare_t5_1_col1)::text))::text COLLATE sys.bbf_unicode_cp1_cs_as) ~~* ((sys.remove_accents_internal(('ABC'::text COLLATE sys.bbf_unicode_cp1_ci_ai)))::text COLLATE sys.bbf_unicode_cp1_cs_as)))
+~~END~~
+
+
+SELECT pg_get_constraintdef(oid) FROM pg_constraint WHERE conrelid = 'master_dbo.BABEL_5608_vu_prepare_t5_2'::regclass;
```
Deepesh125, line 287: Dont we need >= and < operators for this case?

contrib/babelfishpg_tsql/src/collation.c
ahmed-shameem: Please look: https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/3439/files?diff=unified&w=0#r1986562565

test/JDBC/expected/db_collation/test_constraint_like-vu-verify.out
```diff
@@ -0,0 +1,583 @@
...
+~~START~~
+varchar
+active
+inactive
+~~END~~
+
+
+-- CASE WHEN
+INSERT INTO BABEL_5608_vu_prepare_t12_2 VALUES ('active');
+INSERT INTO BABEL_5608_vu_prepare_t12_2 VALUES ('inactive');
```
Deepesh125, line 320: we should have one negative tests

test/JDBC/expected/db_collation/test_constraint_like-vu-prepare.out
```diff
@@ -0,0 +1,388 @@
...
+GO
+
+CREATE VIEW BABEL_5608_13_3_VIEW AS SELECT * FROM BABEL_5608_vu_prepare_t12_2 WHERE status LIKE 'ac%' COLLATE estonian_ci_ai;
+GO
+
+CREATE VIEW BABEL_5608_13_4_VIEW AS SELECT * FROM BABEL_5608_vu_prepare_t12_2 WHERE UPPER(status) LIKE 'ac%';
+GO
+
+CREATE VIEW BABEL_5608_13_5_VIEW AS SELECT * FROM BABEL_5608_vu_prepare_t12_2 WHERE UPPER(status) COLLATE DATABASE_DEFAULT LIKE 'ac%';
+GO
```
Deepesh125, line 216: we should retrieve view def before/after upgrade to verify the correctness

test/JDBC/expected/db_collation/test_constraint_like-vu-prepare.out
```diff
@@ -0,0 +1,388 @@
...
+
+
+-- CASE WHEN
+SELECT pg_get_constraintdef(oid) FROM pg_constraint WHERE conrelid = 'master_dbo.BABEL_5608_vu_prepare_t12_2'::regclass;
+GO
+~~START~~
+text
+CHECK ((<newline>CASE<newline>    WHEN (((sys.remove_accents_internal((status)::text))::text COLLATE sys.bbf_unicode_cp1_cs_as) ~~* ('act%'::text COLLATE sys.bbf_unicode_cp1_cs_as)) THEN 1<newline>    WHEN ((sys.remove_accents_internal((status)::text))::text ~~* 'inactive'::text) THEN 1<newline>    ELSE 0<newline>END = 1))
+~~END~~
+
```
Deepesh125, line 438: we should also have a test cases where LIKE is part of computed column. And just for completeness, lets add test cases where LIKE is part of procedure/function and it exihibits the expected behaviour

Deepesh125 (changes requested):

test/JDBC/expected/db_collation/test_constraint_like-vu-prepare.out
ahmed-shameem: We will get that we have defined the constraint that way: 
```
CREATE TABLE BABEL_5608_vu_prepare_t1(BABEL_5608_vu_prepare_t1_col1 varchar(max), CONSTRAINT BABEL_5608_vu_prepare_t1_c1 CHECK ('abc' LIKE 'A%' COLLATE Latin1_general_ci_as));
```

test/JDBC/expected/db_collation/test_constraint_like-vu-prepare.out
ahmed-shameem: The constraint is this
```
'abc' COLLATE Latin1_general_ci_as LIKE 'A%'
```

So its CollateExpr LIKE Const, hence the transformation is coming into picture

test/JDBC/expected/db_collation/test_constraint_like-vu-prepare.out
ahmed-shameem: No, the constraint definition is
```
BABEL_5608_vu_prepare_t5_2_col1 LIKE 'a%' COLLATE Latin1_General_CI_AI
```

Relabel(Var) LIKE CollateExpr...hence NO optimisation

contrib/babelfishpg_tsql/src/collation.c
ahmed-shameem: Not needed, removed in latest rev

contrib/babelfishpg_tsql/src/collation.c
```diff
@@ -388,6 +399,9 @@ transform_from_ci_as_for_likenode(Node *node, OpExpr *op, like_ilike_info_t like
 	if (IsA(leftop, Const) || !IsA(rightop, Const) ||
 		((Const *) rightop)->constisnull)
 	{
+		/* update the collation of left and right node*/
+		linitial(op->args) = (Node *) create_collate_expr(linitial(op->args), op->inputcollid);
```
rishabhtanwar29, line 402: [Question] Will we need this only when leftop is a Const node? The above `if` condition will let us in when leftop is not Const and rightop is not Const. `IsA(leftop, Const) || !IsA(rightop, Const)`

contrib/babelfishpg_tsql/src/collation.c
```diff
@@ -428,33 +464,46 @@ transform_from_ci_as_for_likenode(Node *node, OpExpr *op, like_ilike_info_t like
 		Node	   *constant_suffix;
 		Const	   *highest_sort_key;
 
+		/* Always create a CollateExpr on top to match with op->inputcollid */
+		new_leftop = create_collate_expr(linitial(op->args), op->inputcollid);
+		linitial(op->args) = (Node*) new_leftop;
```
rishabhtanwar29: can't we do this one line as we are doing just below this?

test/JDBC/expected/BABEL-5608-vu-prepare.out
ahmed-shameem: Yes, already fixed

contrib/babelfishpg_tsql/src/collation.c
ahmed-shameem: Added one such case, please look at case 14

test/JDBC/expected/db_collation/test_constraint_like-vu-verify.out
ahmed-shameem: Added in latest rev

test/JDBC/expected/db_collation/test_constraint_like-vu-prepare.out
ahmed-shameem: Updated in latest rev

test/JDBC/expected/db_collation/test_constraint_like-vu-prepare.out
ahmed-shameem: Updated in latest rev

contrib/babelfishpg_tsql/src/collation.c
ahmed-shameem: Yes, updated in latest rev

contrib/babelfishpg_tsql/src/collation.c
ahmed-shameem: Idea is to add CollateExpr in all cases. We will get inside this block in the following conditions:
1. leftop IS const
2. rightop is NOT const
3. rightop is const but NULL

I have added test cases for the same. Please look at case 1 and 15.

forestkeeper (approved):

rishabhtanwar29 (approved):

Deepesh125 (approved):

PR #3455, opened 2025-02-05
Fix issue with truncation of result of arithmetic computations to correct scale when result is of numeric datatype

rohit01010:
### Description
Currently, whenever an arithmetic operation is performed which results in numeric/decimal, we are storing the result till maximum possible scale instead of the correct scale. This is causing incorrect results in some arithmetic operations such as numeric division, numeric multiplication etc. This PR will fix this issue by adding a hooks in engine which truncates/rounds the result value to the correct scale and also sets the correct typmod of multiple nodes whose base type is `NUMERIC`.

This PR also fixes precision overflow adjustment logic in case NUMERIC ADDITION, SUBTRACTION and MULTIPLICATION.

Engine PR: https://github.com/babelfish-for-postgresql/postgresql_modified_for_babelfish/pull/529

### Issues Resolved

BABEL-5467, BABEL-5685

* **Use case based -**

* **Boundary conditions -**

* **Arbitrary inputs -**

* **Negative test cases -**

* **Performance tests -**

* **Tooling impact -**

* **Client tests -**

### Check List


test/JDBC/expected/BABEL-5467-vu-verify.out
```diff
@@ -0,0 +1,217 @@
+-- Client Tests
```
Yvinayak07, line 1: Do we have UDT testing for this change

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
```diff
@@ -642,7 +642,7 @@ is_numeric_datatype(Oid typid)
 int32
 resolve_numeric_typmod_from_exp(Plan *plan, Node *expr)
 {
-	if (expr == NULL)
+	if (expr == NULL || !is_numeric_datatype(exprType(expr)))
```
Yvinayak07: We can't use this condition here as in `T_Const` we are checking for INT4 also.

rohit01010: Updated the condition to check for exact numeric datatypes.

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -6038,4 +6070,29 @@ remove_db_name_in_schema(const char *object_name, const char *object_type)
...
\ No newline at end of file
+}
+
+static int32
+tsql_get_numeric_typmod_from_exp(Plan *plan, Node *expr)
+{
+	int32       result_typmod = -1;
+	uint8_t     scale,
+                precision;
+	if (pltsql_protocol_plugin_ptr && *pltsql_protocol_plugin_ptr && (*pltsql_protocol_plugin_ptr)->get_numeric_typmod_from_exp)
```
KushaalShroff: if (*pltsql_protocol_plugin_ptr && (*pltsql_protocol_plugin_ptr)->get_numeric_typmod_from_exp)
should be enough, right?

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -6038,4 +6070,29 @@ remove_db_name_in_schema(const char *object_name, const char *object_type)
 	pfree(splited_object_name);
 
 	return (const char *)pstrdup(object_name + prefix_len);
-}
\ No newline at end of file
+}
+
+static int32
+tsql_get_numeric_typmod_from_exp(Plan *plan, Node *expr)
```
KushaalShroff: Lets add meaningful comments over the functions you create.

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -6038,4 +6070,29 @@ remove_db_name_in_schema(const char *object_name, const char *object_type)
...
+	uint8_t     scale,
+                precision;
+	if (pltsql_protocol_plugin_ptr && *pltsql_protocol_plugin_ptr && (*pltsql_protocol_plugin_ptr)->get_numeric_typmod_from_exp)
+		result_typmod = (*pltsql_protocol_plugin_ptr)->get_numeric_typmod_from_exp(plan, expr);
+	if (result_typmod != -1)
+	{
+		/* 
+		* If we are unable to get correct precision and scale for overflow cases
+		* then return -1
+		*/
```
KushaalShroff: Indentation

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -6038,4 +6070,29 @@ remove_db_name_in_schema(const char *object_name, const char *object_type)
...
+		scale = (result_typmod - VARHDRSZ) & 0xffff;
+		precision = ((result_typmod - VARHDRSZ) >> 16) & 0xffff;
+		if (precision > TDS_NUMERIC_MAX_PRECISION)
+		{
+			if (!(precision - scale > 32 && scale > 6) && !(precision - scale <= TDS_NUMERIC_MAX_PRECISION))
+				return -1;
```
KushaalShroff: Can you explain this?

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
```diff
@@ -638,11 +641,57 @@ is_numeric_datatype(Oid typid)
...
+	{
+		TypeName *typename = makeTypeNameFromNameList(list_make2(makeString("sys"), makeString("bit")));
+		tsql_bit_oid = LookupTypeNameOid(NULL, typename, false);
+	}
+	
+	return (decimal_oid == typid
+			|| tsql_tinyint_oid == typid
+			|| tsql_money_oid == typid
+			|| tsql_smallmoney_oid == typid
+			|| tsql_bit_oid == typid);
```
KushaalShroff: Seems inefficient to load all oids first and then return. Better if we decimal_oid == typid check after !OidIsValid(decimal_oid)

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -1190,6 +1202,26 @@ pltsql_bbfViewHasInsteadofTrigger(Relation view, CmdType event)
...
+
+	if (result && OidIsValid(result_type) && getBaseType(result_type) == NUMERICOID)
+	{
+		if (fn_expr != NULL && result_typmod == -1)
+			result_typmod = tsql_get_numeric_typmod_from_exp(NULL, fn_expr);
+
+		if (result_typmod != -1)
+		{
+			scale = (result_typmod - VARHDRSZ) & 0xffff; 
+			return DirectFunctionCall2(numeric_trunc, result, Int32GetDatum(scale));
```
KushaalShroff: Should we be calling tsql_numeric_trunc?

KushaalShroff (changes requested):

contrib/babelfishpg_tsql/src/hooks.c
rohit01010: Yes, updated.

contrib/babelfishpg_tsql/src/hooks.c
rohit01010: Added

contrib/babelfishpg_tsql/src/hooks.c
rohit01010: Fixed

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
rohit01010: Updated code

contrib/babelfishpg_tsql/src/hooks.c
rohit01010: `tsql_numeric_trunc` is a wrapper function which internally calls `numeric_trunc` when the third parameter passed to function `tsql_numeric_trunc` is non-zero, else it will call `tsql_numeric_round`. But in our case we only need to truncate the result hence there is no benefit of using `tsql_numeric_trunc`.

contrib/babelfishpg_tsql/src/hooks.c
rohit01010: In function `resolve_numeric_typmod_from_exp()`, arithmetic operations such as division and multiplication might results in precision overflow cases, we had computed precision and scale for some cases but for other cases we keep the precision as it is. For such unhandled overflow cases incorrect typmod leads to incorrect truncation of result, hence returned -1, to get the existing behaviour.

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -1190,6 +1208,31 @@ pltsql_bbfViewHasInsteadofTrigger(Relation view, CmdType event)
...
+ * pltsql_trunc_numeric_result
+ *  truncates the result value to the correct scale based on result_typmod.
+ *  for result_typmod = -1, computes the result_typmod using pltsql_exprTypmod function.
+ */
+static Datum
+pltsql_trunc_numeric_result(Plan *plan, Node *fn_expr, Datum result, Oid result_type, int32 result_typmod)
+{
+	int32       scale;
+
+	if (result && OidIsValid(result_type) && getBaseType(result_type) == NUMERICOID)
```
Deepesh125, line 1255: we are calling pltsql_trunc_numeric_result before calling this hook so lets get rid of one of this call

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -6048,4 +6091,95 @@ remove_db_name_in_schema(const char *object_name, const char *object_type)
...
+			}
+		}
+	}
+	return result_typmod;
+}
+
+/*
+ * pltsql_ExecInitResultTypeTL -
+ *  Initialize result type, using the plan node's targetlist.
+ *  Following function is same as PostgreSQL ExecInitResultTypeTL() function,
```
Deepesh125: If the most of the logic is same with ExecInitResultTypeTL then IMHO we should be calling this hook only to update typmod in typdesc instead of copying whole logic here. In this way, it will be better option in terms of maintainability in the event of community making changes to ExecInitResultTypeTL.

Deepesh125: If there is specific reason for doing it in this way then plz let me know or all better comments

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -6048,4 +6091,95 @@ remove_db_name_in_schema(const char *object_name, const char *object_type)
...
+                precision;
+	Oid         expr_type = exprType(expr);
+	
+	if (!OidIsValid(expr_type))
+		return -1;
+
+	if (getBaseType(expr_type) == NUMERICOID)
+	{
+		if (plan == NULL && IsA(expr, Aggref) && ((Aggref *)expr)->aggsplit == AGGSPLIT_FINAL_DESERIAL)
+			return -1;
```
Deepesh125: plz explain this part

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -6048,4 +6091,95 @@ remove_db_name_in_schema(const char *object_name, const char *object_type)
...
+			 * If we are unable to get correct precision and scale for overflow cases
+			 * then return -1
+			 */
+			scale = (result_typmod - VARHDRSZ) & 0xffff;
+			precision = ((result_typmod - VARHDRSZ) >> 16) & 0xffff;
+			if (precision > TDS_NUMERIC_MAX_PRECISION)
+			{
+				if (!(precision - scale > 32 && scale > 6) && !(precision - scale <= TDS_NUMERIC_MAX_PRECISION))
+					return -1;
+			}
```
Deepesh125: which test cases exercise this part of code?

Deepesh125 (changes requested):

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
```diff
@@ -665,6 +710,9 @@ resolve_numeric_typmod_from_exp(Plan *plan, Node *expr)
 				Numeric		num;
 				int64		val;
 				
+				if (con->consttypmod != -1)
```
Deepesh125, line 576: irrespective of consttype?

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
```diff
@@ -1031,15 +1146,27 @@ resolve_numeric_typmod_from_exp(Plan *plan, Node *expr)
...
+					Assert(aggref->args != NIL);
 
-				te = (TargetEntry *) linitial(aggref->args);
-				typmod = resolve_numeric_typmod_from_exp(plan, (Node *) te->expr);
+					te = (TargetEntry *) linitial(aggref->args);
+					typmod = resolve_numeric_typmod_from_exp(plan, (Node *) te->expr, &found_typmod);
+					if (!found_typmod)
+					{
+						if (found != NULL) *found = false;
+						return -1;
```
Deepesh125: condition on line 1177 says that we have never returned -1 for funexpr before and now we are doing so. were there any test cases which was exercising that part of code? and how are we ensuring that nothing will be regressed with this code changes?

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
```diff
@@ -909,17 +1000,21 @@ resolve_numeric_typmod_from_exp(Plan *plan, Node *expr)
 
 				/*
 				 * If the following conditions are met then we will recursively find typmod from arg.
-				 * 1) plan == NULL means we are invoking this function during parsing phase.
-				 * 2) rettypmod == -1 means unable to find typmod till now.
-				 * 3) check if only one args and then is that castable to numeric.
+				 * 1) rettypmod == -1 means unable to find typmod till now.
```
Deepesh125, line 920: why did we remove plan == NULL check?

Deepesh125 (changes requested):

contrib/babelfishpg_tsql/src/hooks.c
rohit01010: For constant expressions, before we use to compute the result and would have stored the typmod as -1 for the result const node, whose typmod is later get computed based on its value. But now since we want to store the correct typmod information while creating such const node, we added this hook and are using this to get the typmod. But in overflow cases where we are unable to compute correct result, we should not use the incorrect typmod and use -1 instead so that correct typmod can be computed later. Test cases for these would as follows
```
SELECT CAST(CAST(12465781.46792 as real) as numeric(38,0)) * CAST(CAST(12465781.4679213254 as real) as numeric(38,0));
GO
~~START~~
numeric
155395695939961
~~END~~
```
Here if we set the incorrect typmod in const expression for overflow case, it will not calculate the typmod information at TDS layer based on result value and will result in following type of error
```
Arithmetic overflow error for data type numeric.
```
These tests are there in `BABEL-3066-vu-verify.sql` test file.

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
rohit01010: This `-1` will only be returned when we are not able to get the correct typmod, which only happens when there is any T_Var in recursive call and while computing the typmod of the T_Var its varno is set to OUTER_VAR and plan == NULL. This will not happen when plan is not NULL. Also, condition on line 1177 is true only when typmod found by recursive calls are correct, which we use to assume so far, as we always use to send the plan to get the correct typmod. But with the changes from this PR we also calls this function during parsing and analysing stage where we do not have the plan information, and hence to avoid getting the incorrect typmod and keep the existing behaviour returned -1.

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
rohit01010: This handling and check was added in this commit: [link](https://github.com/babelfish-for-postgresql/babelfish_extensions/commit/4e61d3a62bcd76686884ee0758b2fc055c35d6c9): , the reason behind the check was back at that time we were using this method `resolve_numeric_typmod_from_exp` only at two places one in `PrepareRowDescription()` where we have the plan information and other in `tsql_select_common_typmod_hook` where we do not have the plan. Now to avoid any regression in `PrepareRowDescription()`, this NULL check was added. But now when I investigate it appears we need this handling in both cases irrespective of plan. For example refer following query
```
-- Here column type is NULL
-- For following query plan == NULL check is there
1> select cast(1.12 as numeric(10, 2)) * CASE WHEN type IS NULL THEN 1 ELSE -1 END AS FTD_AMOUNT_FC from testdecimal_tab;
2> go
FTD_AMOUNT_FC          
-----------------------
                 1.1200

(1 rows affected)

-- For following query plan == NULL check is not there
1> select cast(1.12 as numeric(10, 2)) * CASE WHEN type IS NULL THEN 1 ELSE -1 END AS FTD_AMOUNT_FC from testdecimal_tab;
2> go
FTD_AMOUNT_FC 
--------------
          1.12

(1 rows affected)

-- expected results
1> select cast(1.12 as numeric(10, 2)) * CASE WHEN type IS NULL THEN 1 ELSE -1 END AS FTD_AMOUNT_FC from testdecimal_tab;
2> go
FTD_AMOUNT_FC          
-----------------------
                   1.12

(1 rows affected)
```
We have such tests in TestDecimal, TestNumeric test files.

contrib/babelfishpg_tsql/src/hooks.c
rohit01010: This check is redundant, during partial aggregation the correct typmod information cannot be calculated without using plan so added this check to return typmod as -1. This scenario is being handled with following logic
```
/* If this var referes to tuple returned by its outer plan then find the original tle from it */
if (var->varno == OUTER_VAR)
{
	if (plan != NULL)
	{
		Assert(plan);
		rettypmod = resolve_numeric_typmod_outer_var(plan, var->varattno, &found_typmod);
		if (!found_typmod)
		{
			if (found != NULL) *found = false;
			return -1;
		}
		return rettypmod;
	}
	else
	{
		if (found != NULL)
			*found = false;
		return -1;
	}
}
```
in `T_Var` node in `resolve_numeric_typmod_from_exp` function. Hence removed. Tests for these are added in BABEL-5467-vu-verify.sql at line no 170 to 259, when run in parallel mode.

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
rohit01010: > condition on line 1177 says that we have never returned -1 for funexpr before and now we are doing so.

This is true when found_typmod is true. That means, for cases where found_typmod is false we are returning -1 indicating that we cannot compute typmod information without plan information and the correct typmod will be computed later at TDS layer when we have the plan information. This will keep the existing behaviour for cases where we are not able to get the typmod information without plan. Although, if we are getting typmod as -1 even if we have the plan information then we should send the default precision and scale. 

> how are we ensuring that nothing will be regressed with this code changes?

Previously we only had used this function in `PrepareRowDescription()` where we pass the plan information hence we will not get into this code flow, so no regression will happen there. These checks are kind of sanity checks to avoid any such regression as returning incorrect result when plan is not there might cause regression, this handling of adding another parameter which will be set or unset based on whether we are able to calculate the typmod will help us to decide whether to use the typmod returned or keep it as -1, which is the existing behaviour.

> were there any test cases which was exercising that part of code ?

Yes there are tests across different files such as TestNumeric, TestDecimal, NUMERIC_TYPMOD, BABEL-CASE_EXPR-vu-verify, money_aggregate-vu-verify, numericOverflow etc which were giving incorrect scale and precision value in parallel mode with aggregates, as the T_VAR arguments of aggregate expression has type and typmod gets set to `BYTEA` and `-1` during partial aggregation and we were ending up using incorrect typmod. (These output difference in existing test files were detected with this PR changes as we are now passing the correct typmod information and truncating the result earlier during optimising and execution stage.)

shah-nirmit (approved):

contrib/babelfishpg_tsql/src/hooks.c
rohit01010: Updated logic to only update the typmod of TupleDescriptor

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -6022,6 +6065,81 @@ remove_db_name_in_schema(const char *object_name, const char *object_type)
...
+		TargetEntry *tle = lfirst(l);
+		Form_pg_attribute attr = TupleDescAttr(desc, cur_resno - 1);
+
+		if (getBaseType(attr->atttypid) == NUMERICOID)
+		{
+			attr->atttypmod = pltsql_exprTypmod(planstate->plan, (Node *) tle->expr);
+		}
+
+		cur_resno++;
+	}
```
Deepesh125, line 6176: lets also check cur_resno == size of targetList, raise an error if its not

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -6022,6 +6065,81 @@ remove_db_name_in_schema(const char *object_name, const char *object_type)
...
+		return;
+
+	foreach(l, targetList)
+	{
+		TargetEntry *tle = lfirst(l);
+		Form_pg_attribute attr = TupleDescAttr(desc, cur_resno - 1);
+
+		if (getBaseType(attr->atttypid) == NUMERICOID)
+		{
+			attr->atttypmod = pltsql_exprTypmod(planstate->plan, (Node *) tle->expr);
```
Deepesh125, line 6172: is there any better way of updating tupledesc?

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -6022,6 +6065,81 @@ remove_db_name_in_schema(const char *object_name, const char *object_type)
...
+ * pltsql_exprTypmod -
+ *  returns the type-specific modifier of the expression's result type,
+ *  if it can be determined, else we return -1.
+ */
+static int32
+pltsql_exprTypmod(Plan *plan, Node *expr)
+{
+	int32       result_typmod = -1;
+	uint8_t     scale,
+                precision;
```

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
```diff
@@ -675,41 +723,57 @@ resolve_numeric_typmod_from_exp(Plan *plan, Node *expr)
...
+							if (found != NULL) *found = false;
+							return -1;
+						}
+						return rettypmod;
+					}
+					else
+					{
+						if (found != NULL)
+							*found = false;
+						return -1;
```
Deepesh125: why dont we return var->vartypmod?

test/JDBC/expected/non_default_server_collation/chinese_prc_ci_as/babel_datatype.out
```diff
@@ -607,7 +607,7 @@ select CAST(2 AS money) / CAST(0.5 AS numeric(4,2));
 GO
 ~~START~~
 numeric
-4.000000
+4.0000000
```
Deepesh125, line 610: is this correct tsql behaviour?

Deepesh125 (changes requested):

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
rohit01010: If `var->varno == OUTER_VAR` then `var->vartypmod` does not hold correct value as it signifies to use the plan to get the correct value from the plan, hence returning `-1` would be a better option as we will keep the existing behaviour instead of returning an incorrect typmod information

contrib/babelfishpg_tsql/src/hooks.c
rohit01010: I am not aware of any better way than this as PostgreSQL does same thing for updating collation of existing tupleDesc in function `TupleDescInitEntryCollation`

contrib/babelfishpg_tsql/src/hooks.c
rohit01010: removed these checks from engine

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
rohit01010: Yes, if the `consttypmod` is not `-1` then it means it already have correct value and we should that instead of computing on our own, this is irrespective of consttype.

contrib/babelfishpg_tsql/src/hooks.c
rohit01010: Fixed

test/JDBC/expected/non_default_server_collation/chinese_prc_ci_as/babel_datatype.out
rohit01010: This difference is because, currently money datatype typmod is considered as -1, but should have been considered as corresponding to precision = 19, scale = 4. TSQL returns result till 6 scale. This difference should get fixed once the above issue with money & smallmoney is fixed.

test/JDBC/expected/BABEL-5467-vu-verify.out
```diff
@@ -0,0 +1,524 @@
...
+nvarchar#!#nvarchar#!#nvarchar#!#nvarchar#!#tinyint#!#smallint#!#int
+master#!#dbo#!#babel_5467_avgdata_8#!#p#!#10#!#10#!#0
+master#!#dbo#!#babel_5467_avgdata_8#!#q#!#10#!#10#!#0
+master#!#dbo#!#babel_5467_avgdata_8#!#r#!#10#!#10#!#0
+master#!#dbo#!#babel_5467_avgdata_8#!#s#!#10#!#10#!#0
+~~END~~
+
+
+-- Tests for aggregates
+CREATE TABLE babel_5467_t5 (id integer PRIMARY KEY, amount decimal(38,6));
```
shah-nirmit, line 542: Suggestion: We should also test variables along with aggregates

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
```diff
@@ -1080,32 +1207,62 @@ resolve_numeric_typmod_from_exp(Plan *plan, Node *expr)
...
 				else
-					return resolve_numeric_typmod_from_exp(plan, (Node *) rlt->arg);
+				{
+					rettypmod = resolve_numeric_typmod_from_exp(plan, (Node *) rlt->arg, &found_typmod);
+					if (!found_typmod)
+					{
+						if (found != NULL) *found = false;
+						return -1;
+					}
+					return rettypmod;
```
KushaalShroff: Reusing the same piece of code everywhere, chance of encapsulating it?

KushaalShroff (dismissed):

test/JDBC/expected/BABEL-5467-vu-verify.out
rohit01010: Added tests for the same.

test/JDBC/expected/BABEL-NUMERIC_EXPR-vu-verify.out
```diff
@@ -970,8 +970,8 @@ GO
 int#!#float#!#int#!#numeric#!#numeric#!#numeric
 100#!#100.5#!#10#!#100.12345#!#100.1234#!#150.12345000
 200#!#200.5#!#20#!#200.12345#!#200.1234#!#466.79011666
-1100#!#1100.5#!#110#!#1100.12345#!#1100.1234#!#833.45678333
-1200#!#1200.5#!#120#!#1200.12345#!#1200.1234#!#1150.12345000
+1100#!#1100.5#!#110#!#1100.12345#!#1100.0000#!#833.45678333
```
KushaalShroff: Why? Is this expected?

rohit01010: Currently, our typmod computation from expression logic (resolve_numeric_typmod_from_exp) is unable to get the right typmod information for User Defined Datatypes, hence we are getting this behaviour.

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
```diff
@@ -517,27 +516,32 @@ is_numeric_cast(Oid func_oid)
...
-resolve_numeric_typmod_from_exp(Plan *plan, Node *expr)
+resolve_numeric_typmod_from_exp(Plan *plan, Node *expr, bool *found)
 {
+	/*
+	 * set found value as true by default, if we are unable to 
+	 * find the expression typmod found will be set to false.
+	 */
+	if (found != NULL)
+		*found = true;
+
```
tanscorpio7, line 544: Where is this being used ? How is this different from return typmod = -1 ?

rohit01010: only setting return typmod as -1 is not sufficient to identify whether the typmod value we have computed is correct or not, hence added this explicit flag which helps us to identify such scenario's and we can set appropriate details during analysing and executing stage.

tanscorpio7: Can you share an example where `resolve_numeric_typmod_from_exp != -1` and found = false ?

shah-nirmit: +1, Not able to understand the need for maintaining a separate bool var

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -1205,6 +1223,66 @@ pltsql_bbfViewHasInsteadofTrigger(Relation view, CmdType event)
...
+	{
+		if (expr != NULL && result_typmod == -1)
+			result_typmod = pltsql_exprTypmod(plan, expr);
+
+		if (result_typmod != -1)
+		{
+			scale = (result_typmod - VARHDRSZ) & 0xffff; 
+			func_id = get_func_id_from_exp(expr);
+
+			if (func_id == F_NUMERIC_DIV || func_id == F_AVG_NUMERIC)
```
shah-nirmit: Nit: Can we add a comment for this?

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
rohit01010:
```
select POWER(99999999999999999999.12, 1) * cast(1.23 as numeric(3,2)) 
go
```
Here the entire expression typmod when calculated from pltsql_exprTypmod is `458760` which corresponds to precision 7 and scale 4, But the found is false, as the value computed by POWER function is not correct and we want to identify and keep behaviour of all such unhandled cases same as existing behaviour.

shah-nirmit: Thanks , got it 👍

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -1205,6 +1223,66 @@ pltsql_bbfViewHasInsteadofTrigger(Relation view, CmdType event)
 	return false;
 }
 
+/*
+ * get_func_id_from_exp
+ *  For expressions T_FuncExpr, T_OpExpr, T_Aggref returns its function Oid
+ */
+static Oid
+get_func_id_from_exp(Node *expr)
```
shah-nirmit: Suggestion: 
```
bool
check_functions_in_node(Node *node, check_function_callback checker,
						void *context)
```

Can we use this ?

contrib/babelfishpg_tsql/src/hooks.c
rohit01010: Done

contrib/babelfishpg_tsql/src/hooks.c
rohit01010: Updated the logic to use above suggested function.

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
```diff
@@ -643,21 +668,40 @@ resolve_numeric_typmod_from_exp(Plan *plan, Node *expr)
 							tle = get_tle_by_resno(innerplan->targetlist, var->varattno);
 							if (!tle)
 								elog(ERROR, "bogus varattno for INNER_VAR var: %d", var->varattno);
-							return resolve_numeric_typmod_from_exp(innerplan, (Node *)tle->expr);
+							return resolve_numeric_typmod_from_exp(innerplan, (Node *)tle->expr, &found_typmod);
+							if (!found_typmod)
+							{
+								if (found != NULL) *found = false;
+							}
+							return rettypmod;
```
KushaalShroff: Returning before the if?

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -6034,6 +6104,81 @@ remove_db_name_in_schema(const char *object_name, const char *object_type)
...
+ * pltsql_exprTypmod -
+ *  returns the type-specific modifier of the expression's result type,
+ *  if it can be determined, else we return -1.
+ */
+static int32
+pltsql_exprTypmod(Plan *plan, Node *expr)
+{
+	int32       result_typmod = -1;
+	Oid         expr_type;
+	bool		found_typmod;
```
KushaalShroff: Move declaration in the block its required

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -6034,6 +6104,81 @@ remove_db_name_in_schema(const char *object_name, const char *object_type)
...
+		return;
+
+	targetList = planstate->plan->targetlist;
+
+	foreach(l, targetList)
+	{
+		TargetEntry *tle = lfirst(l);
+		Form_pg_attribute attr = TupleDescAttr(desc, cur_resno - 1);
+
+		if (getBaseType(attr->atttypid) == NUMERICOID)
```
KushaalShroff: Why even lookup in case of ``attr->atttypid == NUMERICOID``?
Better to write it as
``attr->atttypid == NUMERICOID || getBaseType(attr->atttypid) == NUMERICOID``

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
```diff
@@ -1028,32 +1148,59 @@ resolve_numeric_typmod_from_exp(Plan *plan, Node *expr)
...
 			}
 		case T_SubLink:
 			{
 				/* Copied from exprTypmod. */
 				const SubLink *sublink = (const SubLink *) expr;
+				int            rettypmod;
+				bool           found_typmod;
 
 				if (sublink->subLinkType == EXPR_SUBLINK ||
 					sublink->subLinkType == ARRAY_SUBLINK)
```
KushaalShroff, line 1185: Why have we added check for ARRAY_SUBLINK?

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
```diff
@@ -1065,15 +1212,23 @@ resolve_numeric_typmod_from_exp(Plan *plan, Node *expr)
...
+					{
+						if (found != NULL) *found = false;
+					}
+					return rettypmod;
 					/* note we don't need to care if it's an array */
 				}
+
+				if (found != NULL) *found = false;
 				/* otherwise, result is RECORD or BOOLEAN, typmod is -1 */
 				return -1;
```
KushaalShroff, line 1206: How do we get record? Using system plpgsql functions? And for boolean do we have tests where this gets triggered?

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
```diff
@@ -1028,32 +1148,59 @@ resolve_numeric_typmod_from_exp(Plan *plan, Node *expr)
...
+				int             rettypmod;
+				bool            found_typmod;
 
-				return resolve_numeric_typmod_from_exp(plan, (Node *) phv->phexpr);
+				rettypmod = resolve_numeric_typmod_from_exp(plan, (Node *) phv->phexpr, &found_typmod);
+				if (!found_typmod)
+				{
+					if (found != NULL) *found = false;
+				}
+				return rettypmod;
```
KushaalShroff, line 1138: We have no code coverage for this. How are we sure this works?

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
```diff
@@ -979,15 +1085,29 @@ resolve_numeric_typmod_from_exp(Plan *plan, Node *expr)
...
 				uint8_t		precision,
 							scale;
+				bool        found_typmod;
 
-				Assert(aggref->args != NIL);
+				if (aggref->aggstar)
+				{
+					if (found != NULL) *found = false;
+					typmod = -1;
+				}
```
KushaalShroff: From assert to allowing aggstar, not enough reasoning in comments to understand this. Do we have tests for agg_func(*)?

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
```diff
@@ -979,15 +1085,29 @@ resolve_numeric_typmod_from_exp(Plan *plan, Node *expr)
...
+				else
+				{
+					Assert(aggref->args != NIL);
```
KushaalShroff, line 1076: Why assert if we know its not null here?

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
```diff
@@ -517,29 +516,37 @@ is_numeric_cast(Oid func_oid)
...
+resolve_numeric_typmod_from_exp(Plan *plan, Node *expr, bool *found)
 {
+	/*
+	 * set found value as true by default, if we are unable to 
+	 * find the expression typmod found will be set to false.
+	 */
+	if (found != NULL)
+		*found = true;
+
 	if (expr == NULL)
```
KushaalShroff, line 545: When will expr be NULL?

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
```diff
@@ -606,7 +621,12 @@ resolve_numeric_typmod_from_exp(Plan *plan, Node *expr)
...
 							elog(ERROR, "bogus varattno for SubqueryScan's subplan: %d", var->varattno);
-						return resolve_numeric_typmod_from_exp(subplan, (Node *)tle->expr);
+						rettypmod = resolve_numeric_typmod_from_exp(subplan, (Node *)tle->expr, &found_typmod);
+						if (!found_typmod)
+						{
+							if (found != NULL) *found = false;
+						}
+						return rettypmod;
 					}
 					else
```
KushaalShroff, line 636: Do we have such errors in the engine as well? can be an assert right? How is subplan null in SubqueryScan

KushaalShroff: Same for MergeAppend

KushaalShroff (changes requested):

KushaalShroff:
It would be interesting to notify any performance penalties with this change by adequate testing

test/JDBC/input/babel_test_int4_numeric_oper_before_16_3-vu-verify.mix
```diff
@@ -1,3 +1,4 @@
+-- sla 50000
```
KushaalShroff, line 1: whats the earlier execution-time?

KushaalShroff (changes requested):

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
rohit01010: This is just a sanity check, that was already there. Ideally we should not call this method on NULL expr.

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
rohit01010: Existing code. Original Author @tanyagupta17 can help answer this. PR: https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/3457 which added this code.

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
rohit01010: Moved the following existing logic which was covering aggregates without start, 
```
Assert(aggref->args != NIL);

te = (TargetEntry *) linitial(aggref->args);
typmod = resolve_numeric_typmod_from_exp(plan, (Node *) te->expr);
```
to else block and added the if block before it which returns typmod as -1 and found false as we currently we are unable to calculate the typmod for this case. Addded tests with `count(*)` in BABEL-5467-vu-verify test file.

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
rohit01010: we do not now whether it is null or not as the previous check was checking the aggstar value here we are checking args value.

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
rohit01010: Here we haven't changed any logic, its just we are storing the typmod in a variable and returning the same instead of directly returning. Also if found_typmod is false the we are falling on to the existing behaviour and if it is true then we will get improved behaviour, in either of the case the value of rettypmod still remains the same.

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
rohit01010: Not sure, Existing logic.

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
rohit01010: Existing logic, not aware about the tests for the same.

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
rohit01010: Fixed

contrib/babelfishpg_tsql/src/hooks.c
rohit01010: moved

contrib/babelfishpg_tsql/src/hooks.c
rohit01010: Updated

test/JDBC/input/babel_test_int4_numeric_oper_before_16_3-vu-verify.mix
rohit01010: previously it was around 22145ms, currently it is around 37142ms.

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
tanyagupta17: Please refer to this thread https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/3457#discussion_r1973714236
This was a safety check, as assertion does not work in production, it was recommended by @tanscorpio7 to add these sanity checks, to avoid any crashes. (Fix as a part of BABEL-5588)

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
KushaalShroff: @Yvinayak07 Please address this in the next changes we raise in this area

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
KushaalShroff: @Yvinayak07 Please address this in the next changes we raise in this area

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
KushaalShroff: @Yvinayak07 Please address this in the next changes we raise in this area

KushaalShroff (approved):

shah-nirmit (approved):

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
shah-nirmit: Should we really throw an error here? , Not sure if this helps User in any way?

Deepesh125 (approved):
Approving based on other approval.

PR #3457, opened 2025-02-05
Numeric/decimal datatype typmod resolution for outer/inner query

tanyagupta17:
### Description

Previously in Babelfish when we have sub-query with UNIONS and in the outer query we have columns like a+b , a*b etc, then we were parsing the whole query plan again in TDS layer and were getting calculating wrong typmod in some cases which would hang the system. 

Now with this change we will directly get the value from the top most node or the next node in case of parallel mode or order by. It also takes care of SubqueryScan and searching for the VAR node by appropriate number

### Issues Resolved

BABEL-5454, BABEL-5588

* **Use case based -** Yes

* **Arbitrary inputs -** Yes

* **Performance tests -** NA

* **Tooling impact -** NA

* **Client tests -** Yes

### Check List


test/JDBC/expected/BABEL-NUMERIC_EXPR-vu-verify.out
```diff
@@ -0,0 +1,918 @@
+-- basic operator testing
+-- test 1
+SELECT COL2_T2 + COL3_T2 FROM BABEL_5454_T2
```
Deepesh125, line 3: we should have some cases where (38,6) + (38, 0) and (38, 0) + (38, 0)

Deepesh125: similarly for each operator which will breach max limit of precision/scale. We want to evaluate the behaviour of sacrificing scale at the cost of integral part.

test/JDBC/expected/BABEL-NUMERIC_EXPR-vu-verify.out
```diff
@@ -0,0 +1,918 @@
...
+25.99345678
+113.62345678
+5.37345678
+6.12345678
+~~END~~
+
+
+
+-- Selecting varchar and numeric, JIRA QUERY
+SELECT
```
Deepesh125: please have some cases where we utilises CTE and as well as cases which breach max precision/scale as mentioned above.

test/JDBC/expected/BABEL-NUMERIC_EXPR-vu-verify.out
```diff
@@ -0,0 +1,918 @@
...
+
+
+-- cte, limit, 5588
+WITH cte AS (SELECT TOP 10 (id+1) AS id FROM BABEL_5454_T9 WHERE id >50 ORDER BY id) SELECT TOP 1 1, id FROM cte
+GO
+~~START~~
+int#!#numeric
+1#!#52
+~~END~~
+
```
Deepesh125, line 1376: can you add tests which exercise `IsA(plan, Limit)` option

Deepesh125 (changes requested):

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
```diff
@@ -695,13 +591,34 @@ resolve_numeric_typmod_from_exp(Plan *plan, Node *expr)
...
+				/* If we are in parallel mode or have a Sort or limit node,
+     			 * find the original target list entry from the outer plan
+				 */
+				if (plan && outerplan && (
+					(IsA(plan, Agg) && IsA(outerplan, Gather)) ||
+					IsA(plan, Gather) ||
+					IsA(plan, Sort) ||
+					IsA(plan, GatherMerge) ||
+					IsA(plan, Limit)) &&
+					var->varno == OUTER_VAR)
```
tanscorpio7: How do we know we are covering everything ? For example what if the outerplan's top node is WIndowAgg ?

tanscorpio7 (changes requested):

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
```diff
@@ -695,13 +591,34 @@ resolve_numeric_typmod_from_exp(Plan *plan, Node *expr)
...
 				{
 					Assert(plan);
-					return (resolve_numeric_typmod_outer_var(plan, var->varattno));
+					Assert(outerplan);
+					if (IsA(outerplan, SubqueryScan))
+					{
+						outerplan = ((SubqueryScan *)outerplan)->subplan;
+						tle = get_tle_by_resno(outerplan->targetlist, var->varattnosyn);
+					}
```
tanscorpio7: I think this should another recursion, in resolve_numeric_typmod_from_exp() under CASE T_VAR:

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
```diff
@@ -695,13 +591,34 @@ resolve_numeric_typmod_from_exp(Plan *plan, Node *expr)
 			}
 		case T_Var:
 			{
-				Var		   *var = (Var *) expr;
+				Var		    *var = (Var *) expr;
+				Plan		*outerplan = NULL;
+				TargetEntry	*tle;
+				if (plan) outerplan = outerPlan(plan);
```
tanscorpio7: These variables should be definied inside the if condition, when we see that VAR is an OUTER VAR

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
tanscorpio7: For example try this
```
CREATE TABLE t (id NUMERIC(4, 2));
GO
INSERT INTO t VALUES (1.0)
GO
SELECT id + 10 FROM (SELECT TOP 5 ROW_NUMBER() OVER(ORDER BY id) AS srNo, id + 1.0 AS id FROM t) AS t
GO
```

tanscorpio7: In fact why are we not recursing for `INNER_VAR` ? 
For example please try 
```
WITH cte AS (SELECT id + 1.0 AS id FROM t) SELECT id + 10, id1 + 10 FROM cte JOIN (SELECT TOP 5 ROW_NUMBER() OVER(ORDER BY id) AS srNo, id + 1.0 AS id1 FROM t) AS t ON (1=1)
```

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
tanyagupta17: As discussed offline, made it general as there are many nodes and we might miss out some specific combination

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
tanyagupta17: Added handling for `SubqueryScan`, `OUTTER_VAR` and `INNER_VAr`

test/JDBC/expected/BABEL-NUMERIC_EXPR-vu-verify.out
tanyagupta17: Added

test/JDBC/expected/BABEL-NUMERIC_EXPR-vu-verify.out
tanyagupta17: Added

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
```diff
@@ -695,13 +591,42 @@ resolve_numeric_typmod_from_exp(Plan *plan, Node *expr)
...
+				 */
+				if (plan && (!IsA(plan, Append) && !IsA(plan, MergeAppend)))
 				{
 					Assert(plan);
-					return (resolve_numeric_typmod_outer_var(plan, var->varattno));
+					if (var->varno == OUTER_VAR)
+					{
+						outerplan = outerPlan(plan);
+						Assert(outerplan);
+						tle = get_tle_by_resno(outerplan->targetlist, var->varattno);
```
tanscorpio7: Let us add null checks for get_tle_by_resno
```
tle = get_tle_by_resno(dpns->outer_tlist, var->varattno);
if (!tle)
    elog(ERROR, "bogus varattno for OUTER_VAR var: %d", var->varattno);
```

tanscorpio7: I would also add null check for outerplan & inner plan in both condition respectively.

tanscorpio7:
Please think hard on `varattnosyn` and ensure we have good coverage for it.

tanscorpio7 (approved):

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
tanyagupta17: As discussed offline, Assertion does not work in production and although OUTER_VAR represents that there will outer query as per postgres, still adding this handling to be on safer side.

Deepesh125 (approved):

PR #3461, opened 2025-02-06
Added support for sp_xml_preparedocument and sp_xml_removedocument procedures

harshdubey166:
Description

Currently babelfish does not support procedures like `sp_xml_preparedocument` and `sp_xml_removedocument`. This PR adds support for the `sp_xml_preparedocument` and `sp_xml_removedocument` procedures.
The procedure  `sp_xml_preparedocument` parses an XML text given as input , stores it at session level and returns a handle to this document, which can be used to fetch the document. 
Similarly , the purpose of `sp_xml_removedocument` is to remove the internal representation of the XML document and invalidate that handle which was created using the `sp_xml_preparedocument` procedure.

The syntax for the procedures is as folllows:

- `sp_xml_preparedocument hdoc OUTPUT [ , xmltext ] [ , xpath_namespaces ] [ ; ]`

    * hdoc is the handle to the newly created document. hdoc is an integer.
    * xmltext is the given XML document in text format. The default value is NULL.
    * xpath_namespaces is a text parameter which specifies the namespace declarations that are used in row and column XPath expressions in OPENXML. The default value is NULL.
  

- `sp_xml_removedocument hdoc [ ; ]`

    * The handle to the already existing created document. A handle that isn't valid returns an error. hdoc is an integer.

BABEL-1168
Signed-off-by: Harsh Dubey harshdu@amazon.com

### Check List


contrib/babelfishpg_tsql/sql/sys_procedures.sql
```diff
@@ -333,3 +333,68 @@ GRANT EXECUTE on PROCEDURE sys.sp_enum_oledb_providers() TO PUBLIC;
...
+AS $$
+DECLARE
+    xml_data XML; 
+    ns_data XML;  
+BEGIN
+    CREATE TEMPORARY TABLE IF NOT EXISTS SessionXMLDocuments (
+        Handle INTEGER GENERATED BY DEFAULT AS IDENTITY PRIMARY KEY,  
+        Xml_content XML,              
+        NamespaceDefinitions XML      
+    ) ON COMMIT PRESERVE ROWS;
```
Deepesh125: We need to prevent this table from dropping accidentally by end user (may be).

contrib/babelfishpg_tsql/sql/sys_procedures.sql
```diff
@@ -333,3 +333,68 @@ GRANT EXECUTE on PROCEDURE sys.sp_enum_oledb_providers() TO PUBLIC;
 CREATE OR REPLACE PROCEDURE sys.sp_reset_connection()
 AS 'babelfishpg_tsql', 'sp_reset_connection_internal' LANGUAGE C;
 GRANT EXECUTE ON PROCEDURE sys.sp_reset_connection() TO PUBLIC;
+
+CREATE OR REPLACE PROCEDURE sys.sp_xml_preparedocument(
```
Deepesh125: with this changes, can you please try out following scenario?
```
create table #SessionXMLDocuments(handle text, xml)

insert some dummy rows in the this table

DECLARE @hdoc INT;
EXEC sp_xml_preparedocument @hdoc OUTPUT, '<root><child></root', 'xmlns:ns1="http://example.com/ns1"/>';
SELECT @hdoc as handle;
EXEC sp_xml_removedocument @hdoc;
GO
```

contrib/babelfishpg_tsql/sql/sys_procedures.sql
shah-nirmit: +1

contrib/babelfishpg_tsql/sql/sys_procedures.sql
```diff
@@ -333,3 +333,68 @@ GRANT EXECUTE on PROCEDURE sys.sp_enum_oledb_providers() TO PUBLIC;
...
+    INOUT hdoc INTEGER,                 
+    IN xmltext sys.VARCHAR DEFAULT NULL,    
+    IN xpath_namespaces sys.VARCHAR DEFAULT NULL 
+) 
+AS $$
+DECLARE
+    xml_data XML; 
+    ns_data XML;  
+BEGIN
+    CREATE TEMPORARY TABLE IF NOT EXISTS SessionXMLDocuments (
```
shah-nirmit: if this is session specific can we name this using something like xml_data_sha256(pid of backend) to prevent name collisions? also would likely decrease chances of dropping this table.

contrib/babelfishpg_tsql/sql/sys_procedures.sql
```diff
@@ -333,3 +333,68 @@ GRANT EXECUTE on PROCEDURE sys.sp_enum_oledb_providers() TO PUBLIC;
 CREATE OR REPLACE PROCEDURE sys.sp_reset_connection()
 AS 'babelfishpg_tsql', 'sp_reset_connection_internal' LANGUAGE C;
 GRANT EXECUTE ON PROCEDURE sys.sp_reset_connection() TO PUBLIC;
+
+CREATE OR REPLACE PROCEDURE sys.sp_xml_preparedocument(
+    INOUT hdoc INTEGER,                 
+    IN xmltext sys.VARCHAR DEFAULT NULL,
```
shah-nirmit: Add test cases to check max length limits of arguments

contrib/babelfishpg_tsql/sql/sys_procedures.sql
```diff
@@ -333,3 +333,68 @@ GRANT EXECUTE on PROCEDURE sys.sp_enum_oledb_providers() TO PUBLIC;
...
+        NamespaceDefinitions XML      
+    ) ON COMMIT PRESERVE ROWS;
+
+    IF xmltext IS NULL OR length(xmltext)=0 OR xml_is_well_formed_document(xmltext) THEN
+        xml_data := xmltext::XML; 
+    ELSE
+        RAISE EXCEPTION 'The XML input is not well-formed.';
+    END IF;
+
+    IF xpath_namespaces IS NULL OR length(xpath_namespaces)=0 OR xml_is_well_formed_document(xpath_namespaces) THEN
```
shah-nirmit: Lets use fully qualified name whereever possible to avoid future issues

shah-nirmit:
Also as Rob mentioned there are procedures in T-SQL which returns a list of handles based on backend/session Id, I dont think temp table approach will be able to support that usecase

contrib/babelfishpg_tsql/sql/sys_procedures.sql
harshdubey166: Created catalog table instead of temporary table as we need to see the handles across the sessions. Now the drop issue is fixed.

contrib/babelfishpg_tsql/sql/sys_procedures.sql
harshdubey166: Tried to create the same table name as the new catalog table(sys.babelfish_xml_handles), it showed error that permission denied for sys schema.

contrib/babelfishpg_tsql/sql/sys_procedures.sql
harshdubey166: We are now creating a catalog table instead of session specific table

contrib/babelfishpg_tsql/sql/sys_procedures.sql
harshdubey166: Yes, added that tests in the test file.

contrib/babelfishpg_tsql/sql/sys_procedures.sql
harshdubey166: Okay, done that.

contrib/babelfishpg_tsql/src/catalog.h
```diff
@@ -505,9 +505,7 @@ extern void	add_entry_to_bbf_partition_depend(int16 dbid, char* partition_scheme
 extern void	remove_entry_from_bbf_partition_depend(int16 dbid, char *schema_name, char *table_name);
 extern bool	is_bbf_partitioned_table(int16 dbid, char *schema_name, char *table_name);
 extern char	*get_partition_scheme_for_partitioned_table(int16 dbid, char *schema_name, char *table_name);
-extern void	rename_table_update_bbf_partition_depend_catalog(RenameStmt *stmt, char *logical_schema_name, int16 dbid);
-
-
+extern void	rename_table_update_bbf_partition_depend_catalog(RenameStmt *stmt, char *logical_schema_name, int16 dbid);
```
shah-nirmit: nit: fix whitespaces and other unwanted changes

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -5110,6 +5110,7 @@ _PG_init(void)
 	init_tsql_coerce_hash_tab(fcinfo);
 	init_tsql_datatype_precedence_hash_tab(fcinfo);
 	init_tsql_cursor_hash_tab(fcinfo);
+	init_tsql_xml_handle_hash_tab(fcinfo);
```
shah-nirmit: Question: why should we init hashtab with connection init? Can we init it on first xml_prep_doc/xml_drop_doc call?

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -108,6 +112,29 @@ bool		sp_describe_first_result_set_inprogress = false;
 char	   *orig_proc_funcname = NULL;
 static bool is_supported_case_sp_describe_undeclared_parameters = true;
 
+/* xml handle hash table */
+static HTAB         *XMLHandleHashTable = NULL;
+static MemoryContext XMLHashtabContext = NULL;
+const uint32         XML_HANDLE_START = 0;
+const uint32         XML_HANDLE_INVALID = 0x80000000; 
+const uint32         MAX_XML_HANDLES_PER_SESSION = 4000;  /* Maximum xml handles per session (size of hash table)*/
```
shah-nirmit: Question: Is this the initial size of htab? or the max num of handles we allow in a concurrent session? if its the max value i dont see us throwing error if num handles grows beyond this value.

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -4290,3 +4317,243 @@ sp_reset_connection_internal(PG_FUNCTION_ARGS)
...
+uint32 pltsql_insert_xml_handle_entry(xmltype *xml_data, xmltype *ns_data, 
+                                     int32 xml_data_length, int32 ns_data_length)
+{
+    XMLHandleHashEnt *hentry;
+    bool found;
+    uint32 handle;
+    uint32 document_id;
+    
+    handle = get_next_xml_handle();      /* get the next handle */
+    document_id = 2*handle - 1;          /* document_id is always odd and unique */
```
shah-nirmit: Can this go OOB?

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -4290,3 +4317,243 @@ sp_reset_connection_internal(PG_FUNCTION_ARGS)
...
+        
+        if (current_xml_handle == XML_HANDLE_INVALID)
+            current_xml_handle = XML_HANDLE_START + 1;
+            
+        if (unlikely(current_xml_handle == old_handle))
+            elog(ERROR, "out of XML handles");
+            
+        if (hash_search(XMLHandleHashTable, &current_xml_handle, 
+                       HASH_FIND, NULL) == NULL)
+            break;  /* found */
```
shah-nirmit: Nit: Update Comment

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -4290,3 +4317,243 @@ sp_reset_connection_internal(PG_FUNCTION_ARGS)
...
+{
+    XMLHandleHashEnt *hentry;
+
+    /* First find the entry */
+    hentry = (XMLHandleHashEnt *) hash_search(XMLHandleHashTable, &document_id, 
+                                             HASH_FIND, NULL);
+    if (hentry == NULL)
+    {
+        ereport(ERROR, errmsg("Could not find prepared statement with handle %d", document_id));
+        return;
```
shah-nirmit: Nit: redundant return

shah-nirmit (changes requested):

contrib/babelfishpg_tsql/src/catalog.h
harshdubey166: Yes, fixed that

contrib/babelfishpg_tsql/src/procedures.c
harshdubey166: Yes, this is the max number of handles we allow in a concurrent session. Added the condition which throws error, if current_xml_handle > MAX_XML_HANDLES_PER_SESSION.

contrib/babelfishpg_tsql/src/procedures.c
harshdubey166: We have set the invalid condition to half the size of uint32 (2147483648), so it does not go OOB.

contrib/babelfishpg_tsql/src/procedures.c
harshdubey166: Yes, resolved that

contrib/babelfishpg_tsql/src/procedures.c
harshdubey166: Yes fixed that

contrib/babelfishpg_tsql/src/pl_handler.c
harshdubey166: Having pre-allocation ensures consistent behavior within transactions without risking memory allocation failures mid-transaction. It is easier to handle errors at connection time than during query execution. Also we have taken reference from cursor hash table where we initialize at the connection time.

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -108,6 +112,29 @@ bool		sp_describe_first_result_set_inprogress = false;
...
 static bool is_supported_case_sp_describe_undeclared_parameters = true;
 
+/* xml handle hash table */
+static HTAB         *XMLHandleHashTable = NULL;
+static MemoryContext XMLHashtabContext = NULL;
+const  int           XML_HANDLE_START = 0;
+const  int           XML_HANDLE_INVALID = INT32_MAX/2; 
+const  int           MAX_XML_HANDLES_PER_SESSION = 4000;  /* Maximum xml handles per session (size of hash table)*/ 
+static int           current_xml_handle;
+int     	         get_next_xml_handle(void);
```
Yvinayak07: NIT: please could you ommit the extra space before `get_next_xml_handle`

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -4290,3 +4317,284 @@ sp_reset_connection_internal(PG_FUNCTION_ARGS)
 
 	PG_RETURN_VOID();
 }
+
+/* Function to generate the xml handles */
+int 
+get_next_xml_handle()
+{
+    int  old_handle = current_xml_handle;
+	int  current_doc_id;
```
Yvinayak07: NIT: please correct the indentation here.

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -4290,3 +4317,284 @@ sp_reset_connection_internal(PG_FUNCTION_ARGS)
...
+    while (true)
+    {
+        ++current_xml_handle;
+        current_doc_id = 2*current_xml_handle -1;
+        
+        if (current_xml_handle == XML_HANDLE_INVALID || current_xml_handle > MAX_XML_HANDLES_PER_SESSION )
+		{
+            current_xml_handle = XML_HANDLE_START + 1;
+            current_doc_id = 2*current_xml_handle -1;
+		}
```
Yvinayak07: NIT: please correct the indentation and space between operators and operands eg. `2*current_xml_handle -1` to be written as `2 * current_xml_handle - 1`

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -4290,3 +4317,284 @@ sp_reset_connection_internal(PG_FUNCTION_ARGS)
...
+            
+        if (unlikely(current_xml_handle == old_handle))
+            elog(ERROR, "out of XML handles");
+            
+        if (hash_search(XMLHandleHashTable, &current_doc_id, 
+                       HASH_FIND, NULL) == NULL)
```
Yvinayak07: could you use:
-  multiline arguments for `hash_search ` function.
```
hash_search(XMLHandleHashTable,
                      &current_doc_id,
                      HASH_FIND,
                      NULL)

or 
```
in a single line:
```
hash_search(XMLHandleHashTable, &current_doc_id, HASH_FIND, NULL)
```

Yvinayak07: Better to put arguments single line as we are comapring it with NULL.

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -4290,3 +4317,284 @@ sp_reset_connection_internal(PG_FUNCTION_ARGS)
...
+            break;  /* handle not found */
```
Yvinayak07: Add comment at top of if condition.

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -4290,3 +4317,284 @@ sp_reset_connection_internal(PG_FUNCTION_ARGS)
...
+void 
+pltsql_create_xml_handle_htab()    /* Creating the hash table */
+{
+    HASHCTL ctl;
+
+    if (XMLHashtabContext == NULL)    /* intialize memory context */
+    {
+        XMLHashtabContext = AllocSetContextCreateInternal(NULL,
+                           "PLtsql XML handle hashtab Memory Context",
+                            ALLOCSET_DEFAULT_SIZES);
```
Yvinayak07: NIT: Could you correct the indentation here :
```
AllocSetContextCreateInternal(NULL,
                                                              "PLtsql XML handle hashtab Memory Context",
                                                              ALLOCSET_DEFAULT_SIZES);
```

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -4290,3 +4317,284 @@ sp_reset_connection_internal(PG_FUNCTION_ARGS)
...
+    current_xml_handle = XML_HANDLE_INVALID;
+}
+
+static xmltype*
+copyXMLType(xmltype *xml_data)
+{
+	xmltype*      new_xml_data;
+	MemoryContext oldcontext;
+
+	if(xml_data == NULL)
```
Yvinayak07: NIT : please add space between `if` and `(`.
```
if (xml_data == NULL)
```

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -4290,3 +4317,284 @@ sp_reset_connection_internal(PG_FUNCTION_ARGS)
...
+
+	/*
+	 * VARSIZE_ANY_EXHDR is the size of the struct in bytes, minus the
+	 * VARHDRSZ or VARHDRSZ_SHORT of its header.  Construct the copy with a
+	 * full-length header.
+	 */
+	oldcontext = MemoryContextSwitchTo(XMLHashtabContext);
+	new_xml_data = (xmltype *) palloc(VARSIZE_ANY_EXHDR(xml_data) + VARHDRSZ);
+	MemoryContextSwitchTo(oldcontext);
+	// new_xml_data = (xmltype *) palloc(VARSIZE_ANY_EXHDR(xml_data) + VARHDRSZ);
```
Yvinayak07: This comment looks redundant.

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -4290,3 +4317,284 @@ sp_reset_connection_internal(PG_FUNCTION_ARGS)
...
+    int  handle;
+    int  document_id;
+    
+    handle = get_next_xml_handle();      /* get the next handle */
+    document_id = 2*handle - 1;          /* document_id is always odd and unique */
+   
+    hentry = (XMLHandleHashEnt *) hash_search(XMLHandleHashTable, &document_id, 
+                                             HASH_ENTER, &found);
+    if (found)
+        ereport(ERROR, errmsg("Duplicate XML handles"));
```
Yvinayak07: Please look if could provide `errcode`.
for ref: https://www.postgresql.org/docs/current/error-message-reporting.html

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -4290,3 +4317,284 @@ sp_reset_connection_internal(PG_FUNCTION_ARGS)
...
+    /* First find the entry */
+    hentry = (XMLHandleHashEnt *) hash_search(XMLHandleHashTable, &document_id, 
+                                             HASH_FIND, NULL);
+    if (hentry == NULL)
+    {
+        ereport(ERROR, errmsg("Could not find prepared statement with handle %d", document_id));
+    }
+
+    /* Remove the entry from hash table */
+    hash_search(XMLHandleHashTable, &document_id, HASH_REMOVE, NULL);
```
HarshLunagariya: Do we need to free the assigned memory of xml_data & ns_data here from hash table's memory context?

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -108,6 +112,29 @@ bool		sp_describe_first_result_set_inprogress = false;
...
+
+typedef struct XMLHandleHashEnt           /* Entries of hash table */
+{
+	int         document_id;
+	int         namespace_id;
+	int         original_document_size_bytes;
+	int         original_namespace_document_size_bytes;
+	bool        is_namespace_null;
+	xmltype    *xml_data; 
+	xmltype    *ns_data;
```
HarshLunagariya: What are the restrictions on xml_data & ns_data field acc. to ideal TSQL behaviour? Currently, there are no limitations for these fields, which can cause the bloat. Let's verify and add length restrictions if applicable.

contrib/babelfishpg_tsql/sql/sys_procedures.sql
```diff
@@ -333,3 +333,23 @@ GRANT EXECUTE on PROCEDURE sys.sp_enum_oledb_providers() TO PUBLIC;
 CREATE OR REPLACE PROCEDURE sys.sp_reset_connection()
 AS 'babelfishpg_tsql', 'sp_reset_connection_internal' LANGUAGE C;
 GRANT EXECUTE ON PROCEDURE sys.sp_reset_connection() TO PUBLIC;
+
+CREATE OR REPLACE PROCEDURE sys.sp_xml_preparedocument(
+    INOUT "@hdoc"  INTEGER,                 
+    IN "@xmltext" sys.VARCHAR DEFAULT NULL,    
+    IN "@xpath_namespaces" sys.VARCHAR DEFAULT NULL
```
Deepesh125: is this expected default value?

contrib/babelfishpg_tsql/src/pl_handler.c
Deepesh125: I think that is not correct way. Error at connection time is not good, its basically means that we may potentially cause "denial of service" attack if there is any unwanted bug somewhere down in the code. Secondly, Babelfish connections are already slower and we are putting more burden on that by loading it at connection time. So IMO, we should load the hash table on demand. Cursor related hash table is there probably because it has different use case.

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -7497,4 +7498,4 @@ gen_command_grant_revoke_priv_to_role(StringInfo query, const char *rolename,
 				(revoke_createrole ? "nocreaterole" : ""), grant_createdb ? "createdb" : 
 					(revoke_createdb ? "nocreatedb" : ""));
 
-}
\ No newline at end of file
+}
```
Deepesh125: unnecessary change

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -108,6 +112,29 @@ bool		sp_describe_first_result_set_inprogress = false;
...
+static int           current_xml_handle;
+int     	         get_next_xml_handle(void);
+void                 pltsql_create_xml_handle_htab(void);
+void                 pltsql_delete_xml_handle_entry(int  handle);
+int                  pltsql_insert_xml_handle_entry(xmltype *xml_data,xmltype *ns_data, int xml_data_length, int ns_data_length);
+
+typedef struct XMLHandleHashEnt           /* Entries of hash table */
+{
+	int         document_id;
+	int         namespace_id;
```
Deepesh125: should be unsigned int

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -108,6 +112,29 @@ bool		sp_describe_first_result_set_inprogress = false;
...
+	int         original_document_size_bytes;
+	int         original_namespace_document_size_bytes;
```
Deepesh125: should be of type "size"

contrib/babelfishpg_tsql/src/procedures.c
Deepesh125: what could be max length of xml_data?

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -4290,3 +4317,284 @@ sp_reset_connection_internal(PG_FUNCTION_ARGS)
...
+
+    /* Validating the given xml text string */
+    if (xml_data_length == 0)
+    {
+        xml_data = NULL;  // Handle NULL case
+    }
+    else
+    {
+        is_xml_text_well_formed = DatumGetBool(
+            DirectFunctionCall1(xml_is_well_formed_document, PG_GETARG_DATUM(1))
```
Deepesh125: this function call through DirectFunctionCall1 will go through security_defined which will take its own time. I would suggest you to call `wellformed_xml` directly

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -4290,3 +4317,284 @@ sp_reset_connection_internal(PG_FUNCTION_ARGS)
...
+        );
+    
+        if (is_xml_text_well_formed)
+        {
+            xml_data = DatumGetXmlP(DirectFunctionCall1(xml_in, CStringGetDatum(xml_text)));
```
Deepesh125: this xml_in and xml_is_well_formed_document (or wellformed_xml) both calls xml_parse which actually returns xmlDocPtr which could be NULL if xml is not well formed. So we can be little bit more creative here by calling xml_in directly on input xml_text. If it returns NULL then you can say that its error case or continue with execution if its not null

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -4290,3 +4317,284 @@ sp_reset_connection_internal(PG_FUNCTION_ARGS)
...
+    else
+    {
+        is_xml_text_well_formed = DatumGetBool(
+            DirectFunctionCall1(xml_is_well_formed_document, PG_GETARG_DATUM(1))
+        );
+    
+        if (is_xml_text_well_formed)
+        {
+            xml_data = DatumGetXmlP(DirectFunctionCall1(xml_in, CStringGetDatum(xml_text))); 
+			xml_data_length = strlen(xml_text);
```
Deepesh125: is this correct? shouldn't we extract length from  xml_data?

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -4290,3 +4317,284 @@ sp_reset_connection_internal(PG_FUNCTION_ARGS)
 
 	PG_RETURN_VOID();
 }
+
+/* Function to generate the xml handles */
+int 
+get_next_xml_handle()
```
Deepesh125: Please add comment above the function on its functionality. It's really hard to get what function is trying to achieve by just looking at code

contrib/babelfishpg_tsql/src/procedures.c
Deepesh125: if we throw at this point then we are leaking document_id because current_xml_handle is already updated inside get_next_xml_handle

Deepesh125 (changes requested):

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -4290,3 +4317,284 @@ sp_reset_connection_internal(PG_FUNCTION_ARGS)
...
+    values[0] = Int32GetDatum(document_id);
+    tuple = heap_form_tuple(tupdesc, values, &isnull);
+
+    result = (HeapTupleHeader) palloc(tuple->t_len);
+    memcpy(result, tuple->t_data, tuple->t_len);
+
+    heap_freetuple(tuple);
+    ReleaseTupleDesc(tupdesc);
+
+    PG_RETURN_HEAPTUPLEHEADER(result);
```
Deepesh125: Can you please explain this part why is it needed?

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -4290,3 +4317,284 @@ sp_reset_connection_internal(PG_FUNCTION_ARGS)
...
+    handle = get_next_xml_handle();      /* get the next handle */
+    document_id = 2*handle - 1;          /* document_id is always odd and unique */
+   
+    hentry = (XMLHandleHashEnt *) hash_search(XMLHandleHashTable, &document_id, 
+                                             HASH_ENTER, &found);
+    if (found)
+        ereport(ERROR, errmsg("Duplicate XML handles"));
+
+    hentry->xml_data = copyXMLType(xml_data);
+    hentry->ns_data = copyXMLType(ns_data);
```
Deepesh125: how are we ensuring that its being copied (or memory being allocated) is inside hash memory context we created?

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -4290,3 +4317,284 @@ sp_reset_connection_internal(PG_FUNCTION_ARGS)
...
+    XMLHandleHashTable = NULL;
+
+    /* Re-create the xml handles-related data structures. */
+    pltsql_create_xml_handle_htab();
+    
+    current_xml_handle = XML_HANDLE_INVALID;     /* Reset xml handles. */
+}
+
+Datum
+sp_xml_preparedocument(PG_FUNCTION_ARGS)
```
Deepesh125, line 4827: We also need to update some shared metadata so that procedures like `dm_exec_xml_handles` can return XML handles prepared in this session? how are we doing it?

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -4290,3 +4317,284 @@ sp_reset_connection_internal(PG_FUNCTION_ARGS)
...
+    ReleaseTupleDesc(tupdesc);
+
+    PG_RETURN_HEAPTUPLEHEADER(result);
+}
+
+Datum
+sp_xml_removedocument(PG_FUNCTION_ARGS)
+{
+    int doc_handle;
+    TSQLInstrumentation(INSTR_TSQL_SP_XML_REMOVEDOCUMENT);
```
Deepesh125: isnt this required for prepare doc also?

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -4290,3 +4317,284 @@ sp_reset_connection_internal(PG_FUNCTION_ARGS)
...
+pltsql_delete_xml_handle_entry(int document_id)
+{
+    XMLHandleHashEnt *hentry;
+
+    /* First find the entry */
+    hentry = (XMLHandleHashEnt *) hash_search(XMLHandleHashTable, &document_id, 
+                                             HASH_FIND, NULL);
+    if (hentry == NULL)
+    {
+        ereport(ERROR, errmsg("Could not find prepared statement with handle %d", document_id));
```
Deepesh125: use ereport with proper error code

contrib/babelfishpg_tsql/src/procedures.c
Deepesh125: Interesting! how are we ensuring it?

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -4290,3 +4317,284 @@ sp_reset_connection_internal(PG_FUNCTION_ARGS)
...
+
+    pltsql_create_xml_handle_htab();
+    
+    PG_RETURN_INT32(0);
+}
+
+/*
+ * reset_cached_xml_handle:
+ *		Cleans up all the stale states and resets the xml handles.
+ *		This function should be called when a connection is cancelled or terminated.
```
Deepesh125: Just say connection is reset or connection is terminated. There is no thing like "connection is cancelled"

Deepesh125 (changes requested):

contrib/babelfishpg_tsql/src/pl_handler.c
harshdubey166: Initialised & created the table when the procedure is being called instead of initialisation at connection time.

contrib/babelfishpg_tsql/src/procedures.c
harshdubey166: Yes, added that.

contrib/babelfishpg_tsql/src/procedures.c
harshdubey166: To return the handle after the execution of the procedure, we are doing that.

contrib/babelfishpg_tsql/src/procedures.c
harshdubey166: The implementation of dm_exec_xml_handles is different from this. We will update that metadata during the implementation of dm_exec_xml_handles

contrib/babelfishpg_tsql/src/procedures.c
harshdubey166: We are now using temp tables which store data on disk. So there is no limit for xml data.

contrib/babelfishpg_tsql/sql/sys_procedures.sql
harshdubey166: Yes.

contrib/babelfishpg_tsql/src/procedures.c
harshdubey166: We are extracting the length of string xml_text which is given as input.

contrib/babelfishpg_tsql/sql/sys_procedures.sql
harshdubey166: Approach has been changed. Only bbf_role_admin has the drop permissions. User can't drop it.

contrib/babelfishpg_tsql/src/procedures.c
harshdubey166: Approach has been changed. Now we are using ENR temp tables instead of hash table.

contrib/babelfishpg_tsql/src/procedures.c
harshdubey166: SQL server has doc_id as INT and size of xml data as BIGINT so we are using the same.

contrib/babelfishpg_tsql/src/procedures.c
harshdubey166: Approach has been changed.

contrib/babelfishpg_tsql/src/procedures.c
harshdubey166: Approach has been changed. Now we are not doing that.

contrib/babelfishpg_tsql/src/procedures.c
harshdubey166: Removed this redundant part.

contrib/babelfishpg_tsql/src/procedures.c
harshdubey166: In the current implementation, the XML document is validated using PostgreSQL's xml_is_well_formed_document function. If validation passes, the C string (xml_text) is converted to PostgreSQL's internal XML type (xmltype) via (cstring_to_text) for storage in temp table.

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -4290,3 +4304,511 @@ sp_reset_connection_internal(PG_FUNCTION_ARGS)
...
+ * Returns: Next available XML handle (integer)
+ * Errors: Throws error when all possible handles are in use
+ */
+int 
+get_next_xml_handle()
+{
+    int           old_handle = current_xml_handle;
+    MemoryContext oldContext = NULL;
+    
+    while (true)
```
ayushdsh: Do you think this can be improved? What if the user keeps removing the last document inserted. This is O(n); but what if we keep a set of prepared handles i.e. available handles and keep giving the handle from the start, whenever removed put it back in the set.

Just something that was on the top of my head at the time. Let me know your thoughts.

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -108,6 +112,16 @@ bool		sp_describe_first_result_set_inprogress = false;
 char	   *orig_proc_funcname = NULL;
 static bool is_supported_case_sp_describe_undeclared_parameters = true;
 
+const  int           XML_HANDLE_START = 0;
+const  int           XML_HANDLE_INVALID = INT_MAX / 2;
```
ayushdsh: Why is the limit of handles INT_MAX/2 ? Just a small explanation would be good.

harshdubey166: Since document_id is calculated as (2 * handle - 1) and is of type int, its maximum value must not exceed INT_MAX. Therefore, the maximum number of handles is capped at INT_MAX / 2 — no more to avoid overflow, and no less to ensure we fully utilize the available handle space.

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -4290,3 +4304,528 @@ sp_reset_connection_internal(PG_FUNCTION_ARGS)
...
+     {
+		/* Set current user to bbf_role_admin for create permissions.*/
+		sql_dialect = SQL_DIALECT_TSQL;
+		SetUserIdAndSecContext(get_bbf_role_admin_oid(), save_sec_context | SECURITY_LOCAL_USERID_CHANGE);
+		address = DefineRelation(stmt, RELKIND_RELATION, InvalidOid, NULL, NULL);
+
+		/* Store the OID in TopMemoryContext to persist across transactions */
+		oldContext = MemoryContextSwitchTo(TopMemoryContext);
+		xml_handle_temp_table_relid = address.objectId;
+		MemoryContextSwitchTo(oldContext);
```
rohit01010: why do we need to change context for storing the relid ?

contrib/babelfishpg_tsql/sql/sys_procedures.sql
```diff
@@ -333,3 +333,20 @@ GRANT EXECUTE on PROCEDURE sys.sp_enum_oledb_providers() TO PUBLIC;
 CREATE OR REPLACE PROCEDURE sys.sp_reset_connection()
 AS 'babelfishpg_tsql', 'sp_reset_connection_internal' LANGUAGE C;
 GRANT EXECUTE ON PROCEDURE sys.sp_reset_connection() TO PUBLIC;
+
+CREATE OR REPLACE PROCEDURE sys.sp_xml_preparedocument( INOUT "@hdoc"  INTEGER,                 
+                                                              IN "@xmltext" sys.VARCHAR DEFAULT NULL,    
+                                                              IN "@xpath_namespaces" sys.VARCHAR DEFAULT NULL )
```
Deepesh125, line 340: Please have test cases which very large xml data

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -20,6 +20,7 @@
 #include "catalog/indexing.h"
 #include "commands/defrem.h"
 #include "commands/prepare.h"
+#include "commands/tablecmds.h"
```
Deepesh125, line 23: why is this needed?

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -4290,3 +4304,528 @@ sp_reset_connection_internal(PG_FUNCTION_ARGS)
...
+    {
+        ns_data = NULL;
+    }
+    else
+    {
+        is_xpath_namespaces_well_formed = DatumGetBool(DirectFunctionCall1(xml_is_well_formed_document, PG_GETARG_DATUM(2)));
+    
+        if (is_xpath_namespaces_well_formed)
+        {
+            ns_data = (xmltype *) cstring_to_text(xpath_namespaces);
```
Deepesh125: this would be same as `PG_GETARG_DATUM(2)` then why do we need cstring_to_text?

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -4290,3 +4304,528 @@ sp_reset_connection_internal(PG_FUNCTION_ARGS)
...
+    Datum                 values[7];
+    Oid                   save_userid;
+    int                   save_sec_context;
+    int                   saved_dialect = sql_dialect;
+    HeapTuple             tuple;
+    bool                  nulls[7] = {false, true, false, true, true, false, true};
+    bool                  table_exists = false;
+
+   /* Check if the table exists using ENR lookup */
+   if(get_ENR_withoid(currentQueryEnv, xml_handle_temp_table_relid, ENR_TSQL_TEMP) != NULL)
```
Deepesh125: Make sure to get this reviewed by @timchang514  or @lejaokri

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -4290,3 +4304,528 @@ sp_reset_connection_internal(PG_FUNCTION_ARGS)
...
+ * The function temporarily elevates privileges to "bbf_role_admin" to ensure it has
+ * the necessary permissions to create the table. It also sets the SQL dialect
+ * to TSQL to ensure proper table creation semantics.
+ */
+
+ void 
+ create_xml_handle_temp_table()  
+ {
+     CreateStmt   *stmt = makeNode(CreateStmt);
+     RangeVar     *relation = makeRangeVar(NULL,"#xml_handle_temp_table", -1);
```
Deepesh125: use macro for constant

contrib/babelfishpg_tsql/src/procedures.c
Deepesh125: +1

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -4290,3 +4304,528 @@ sp_reset_connection_internal(PG_FUNCTION_ARGS)
...
+     stmt->tableElts = create_xml_handle_columns();
+     stmt->constraints = NIL;
+     stmt->inhRelations = NIL;  
+     stmt->partspec = NULL;
+     stmt->ofTypename = NULL; 
+     stmt->oncommit = ONCOMMIT_PRESERVE_ROWS; 
+     stmt->tablespacename = NULL;
+     stmt->if_not_exists = false;  
+     stmt->options = NIL;
+     stmt->accessMethod = NULL;
```
Deepesh125: please make sure that its same as if we run `create table #temp()`

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -4290,3 +4304,528 @@ sp_reset_connection_internal(PG_FUNCTION_ARGS)
...
+}
+
+/* Define column definitions for #xml_handle_temp_table
+ The table has the following structure:
+ *
+ * - document_id: A unique identifier for the XML document (INT, NOT NULL)
+ * - namespace_id: An identifier for the namespace, if present (INT)
+ * - doc_size: The size of the XML document in bytes (BIGINT)
+ * - ns_size: The size of the namespace in bytes (BIGINT)
+ * - is_namespace_null: Flag indicating if namespace is NULL (BOOLEAN)
```
Deepesh125, line 4408: why do we need this explicitly?

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -4290,3 +4304,528 @@ sp_reset_connection_internal(PG_FUNCTION_ARGS)
...
+    ColumnDef *docIdCol;
+    ColumnDef *nsIdCol;
+    ColumnDef *docSizeCol;
+    ColumnDef *nsSizeCol;
+    ColumnDef *isNsNullCol;
+    ColumnDef *xmlDataCol;
+    ColumnDef *nsDataCol;
+    
+    /* Create column definitions */
+    docIdCol = makeNode(ColumnDef);
```
Deepesh125: instead of repeating code for each column, we should either use macro or function call.

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -4290,3 +4304,528 @@ sp_reset_connection_internal(PG_FUNCTION_ARGS)
...
+     stmt->accessMethod = NULL;
+     
+     GetUserIdAndSecContext(&save_userid, &save_sec_context);
+ 
+     PG_TRY();
+     {
+		/* Set current user to bbf_role_admin for create permissions.*/
+		sql_dialect = SQL_DIALECT_TSQL;
+		SetUserIdAndSecContext(get_bbf_role_admin_oid(), save_sec_context | SECURITY_LOCAL_USERID_CHANGE);
+		address = DefineRelation(stmt, RELKIND_RELATION, InvalidOid, NULL, NULL);
```
Deepesh125: instead of elevating the privilege to get_bbf_role_admin_oid and running this DefineRelation, can we pass ownerId=get_bbf_role_admin_oid (3rd arg)?

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -4290,3 +4304,528 @@ sp_reset_connection_internal(PG_FUNCTION_ARGS)
...
+    /*
+     * Switch to TopMemoryContext to ensure the bitmap set persists across transactions.
+     * This prevents the bitmap set from being freed when the current memory context
+     * is reset, which would cause the handles to be lost and potentially reused.
+     */
+    oldContext = MemoryContextSwitchTo(TopMemoryContext);
+    active_xml_handles_counter = bms_add_member(active_xml_handles_counter, current_xml_handle_counter);
+    MemoryContextSwitchTo(oldContext);
+    
+    return current_xml_handle_counter;
```
Deepesh125: how are we ensuring that there is no handle leak in case of txn rollback?

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -4290,3 +4304,528 @@ sp_reset_connection_internal(PG_FUNCTION_ARGS)
...
+    
+    /* Set up the values for insertion */
+    values[0] = Int32GetDatum(document_id);
+    values[1] = Int32GetDatum(namespace_id);
+    values[2] = Int32GetDatum(xml_data_length);
+    values[3] = Int32GetDatum(ns_data_length);
+    values[4] = BoolGetDatum(is_namespace_null);
+    
+    if (xml_data == NULL)
+        nulls[5] = true;
```
Deepesh125: values[5] will be garbage

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -4290,3 +4304,528 @@ sp_reset_connection_internal(PG_FUNCTION_ARGS)
...
+    else
+        values[5] = PointerGetDatum(xml_data);
+    
+    /* For namespace data */
+    if (!nulls[6])
+        values[6] = PointerGetDatum(ns_data);
```
Deepesh125: same here, values[6] could be garbage

contrib/babelfishpg_tsql/src/procedures.c
harshdubey166: Since we are using the DefineRelation() function which is defined there, so we are using that header file.

contrib/babelfishpg_tsql/src/procedures.c
timchang514: is it safe to have the name of the table be constant? for ex, can there be nested calls to this function?

what if a user already has a table named `#xml_handle_temp_table` for some reason?

contrib/babelfishpg_tsql/src/procedures.c
timchang514: TopMemoryContext is the one that is guaranteed to never be cleared for a backend, although you've already defined `xml_handle_temp_table_relid ` as a `static int`. but why do we need to store relid across multiple xacts?

contrib/babelfishpg_tsql/src/procedures.c
timchang514: similar to above comment - what if there are errors? what if the table is created but an error is thrown and the metadata hangs around. then all future XML calls in the session might fail because the relation name is not unique.

contrib/babelfishpg_tsql/src/procedures.c
timchang514: Use GetENRTempTableWithOid instead. Also `!= NULL` is unnecessary

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -4290,3 +4304,528 @@ sp_reset_connection_internal(PG_FUNCTION_ARGS)
...
+      relation = relation_open(xml_handle_temp_table_relid, RowExclusiveLock);
+      table_exists = true;
+   }
+
+    if (!table_exists)
+    {
+        /* Table doesn't exist or was dropped, create it */
+        create_xml_handle_temp_table();
+        
+        if (!OidIsValid(xml_handle_temp_table_relid))
```
timchang514: if this condition is hit, then we will rollback. is there any other state that needs to be cleaned up if we exit here?

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -4290,3 +4304,528 @@ sp_reset_connection_internal(PG_FUNCTION_ARGS)
...
+    tuple = heap_form_tuple(RelationGetDescr(relation), values, nulls);
+
+    GetUserIdAndSecContext(&save_userid, &save_sec_context);
+    
+    PG_TRY();
+    {
+        /* Set current user to bbf_role_admin for insert permissions */
+        sql_dialect = SQL_DIALECT_TSQL;
+        SetUserIdAndSecContext(get_bbf_role_admin_oid(), save_sec_context | SECURITY_LOCAL_USERID_CHANGE);
+        CatalogTupleInsert(relation, tuple);
```
timchang514: why are we calling `CatalogTupleInsert` on a temp table? this function is reserved for system catalogs. maybe you want `heap_insert`? 

comments here explaining what is happening would be helpful as well

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -4290,3 +4304,528 @@ sp_reset_connection_internal(PG_FUNCTION_ARGS)
...
+    if (HeapTupleIsValid(tuple))
+    {
+        GetUserIdAndSecContext(&save_userid, &save_sec_context);
+         
+        PG_TRY();
+        {
+            /* Set current user to bbf_role_admin for delete permissions */
+            sql_dialect = SQL_DIALECT_TSQL;
+            SetUserIdAndSecContext(get_bbf_role_admin_oid(), save_sec_context | SECURITY_LOCAL_USERID_CHANGE);
+            CatalogTupleDelete(relation, &tuple->t_self);
```
timchang514: same, why are we using table_beginscan_catalog and CatalogTupleDelete on a temp table?

contrib/babelfishpg_tsql/src/procedures.c
harshdubey166: The namespace_id, ns_size, is_namespace_null all the metadata will be used in the sys.dm_exec_xml_handles view(to be implemented later) .So we are storing them too in the table.

contrib/babelfishpg_tsql/src/procedures.c
harshdubey166: Yes, this part looks redundant. We have not to switch memory context. We have added the check that if the table gets dropped in case of xact rollback , create a new table on demand.So changing memory context is redundant.

contrib/babelfishpg_tsql/src/procedures.c
harshdubey166: Okay, got that.

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -4290,3 +4304,528 @@ sp_reset_connection_internal(PG_FUNCTION_ARGS)
...
+    	relation = relation_open(xml_handle_temp_table_relid, RowExclusiveLock);
+    	table_exists = true;
+    }
+
+    if (!table_exists)
+    {
+        /* Table doesn't exist, so the handle definitely doesn't exist */
+        ereport(ERROR,
+                (errcode(ERRCODE_UNDEFINED_OBJECT),
+                 errmsg("Could not find prepared statement with handle %d", document_id)));
```
Deepesh125: please update this error message

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -4290,3 +4304,528 @@ sp_reset_connection_internal(PG_FUNCTION_ARGS)
...
+        }
+        PG_FINALLY();
+        {
+            SetUserIdAndSecContext(save_userid, save_sec_context);
+            sql_dialect = saved_dialect;
+        }
+        PG_END_TRY();
+        
+        oldContext = MemoryContextSwitchTo(TopMemoryContext);
+        active_xml_handles_counter = bms_del_member(active_xml_handles_counter, curr_handle_counter);
```
Deepesh125: this should be within above TRY block

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -4290,3 +4304,528 @@ sp_reset_connection_internal(PG_FUNCTION_ARGS)
...
+        ereport(ERROR,
+                (errcode(ERRCODE_UNDEFINED_OBJECT),
+                 errmsg("Could not find prepared statement with handle %d", document_id)));
+    }
+
+    /* Calculate the handle counter */
+    curr_handle_counter = (document_id + 1) / 2;
+
+    /* Check if the handle is still active */
+    if (!bms_is_member(curr_handle_counter, active_xml_handles_counter))
```
Deepesh125: there is risk of leaking handles when we try to create XML entry in the table. If we dont fix it then this may produce false negative. We should ensure correctness of end to end behaviour

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -4290,3 +4304,528 @@ sp_reset_connection_internal(PG_FUNCTION_ARGS)
...
+{
+    /* Reset active handles bitmap */
+    bms_free(active_xml_handles_counter);
+    active_xml_handles_counter = NULL;
+
+    /* Reset xml handles */
+    current_xml_handle_counter = XML_HANDLE_COUNTER_INVALID; 
+
+    /* Invalidate the Oid */
+    xml_handle_temp_table_relid = InvalidOid;
```
Deepesh125: how do we ensure that table is truncated?

Deepesh125 (changes requested):

contrib/babelfishpg_tsql/src/procedures.c
harshdubey166: Okay, made changes.

contrib/babelfishpg_tsql/src/procedures.c
harshdubey166: Yes, since the input is already of type text, and we're not modifying the content after extracting it as a C string, we can directly use PG_GETARG_DATUM(2) instead of converting back using cstring_to_text().

contrib/babelfishpg_tsql/src/procedures.c
harshdubey166: Added explicitly, values[5]= (Datum)0, when xml_data == NULL.

contrib/babelfishpg_tsql/src/procedures.c
harshdubey166: Added explicitly, values[6]= (Datum)0, if ns_data == NULL.

contrib/babelfishpg_tsql/sql/sys_procedures.sql
harshdubey166: Yes, added

contrib/babelfishpg_tsql/src/procedures.c
harshdubey166: Yes, we can pass ownerId in the 3rd arg of DefineRelation. Made changes

contrib/babelfishpg_tsql/src/procedures.c
harshdubey166: Since we're working with temporary tables (ENRs) and not system catalogs, instead of using CatalogTupleInsert and CatalogTupleDelete  we can use simple_heap_insert and simple_heap_delete, which are more appropriate for ENRs.

contrib/babelfishpg_tsql/src/procedures.c
harshdubey166: Also, since table_beginscan_catalog is meant for catalog tables, I have used systable_beginscan instead.

contrib/babelfishpg_tsql/src/procedures.c
harshdubey166: Got that.

contrib/babelfishpg_tsql/src/procedures.c
harshdubey166: Since we are using temp tables (ENRs), so they are automatically dropped when a connection is reset. We don't have to explicitly add checks for that.

contrib/babelfishpg_tsql/src/procedures.c
harshdubey166: Since we are performing all operations like insertion,deletion etc. after this step, there is not any other state that needs to be cleaned up if we exit here.

contrib/babelfishpg_tsql/src/procedures.c
harshdubey166: If the table does not exist and user tries to remove a handle which does not exist , we should throw this error message as sql server gives the same error message when handle does not exist.

contrib/babelfishpg_tsql/src/procedures.c
harshdubey166: yes, this is same as if we run `create table #temp()`

contrib/babelfishpg_tsql/src/procedures.c
harshdubey166: Generated unique table name using md5 hash to avoid collisions.

contrib/babelfishpg_tsql/src/procedures.c
harshdubey166: Fixed that issue. The implementation prevents handle leaks during transaction rollbacks by verifying if entries actually exist in the table. When a handle is in the active set bitmap, we check if a corresponding entry exists in the temporary table. If no entry exists (which happens after a rollback), we consider the handle valid and reuse it. This approach ensures handles don't remain marked as "in use" after their corresponding table entries have been removed due to transaction rollbacks, effectively preventing handle leaks.

contrib/babelfishpg_tsql/src/procedures.c
harshdubey166: Fixed that issue.

contrib/babelfishpg_tsql/src/procedures.c
harshdubey166: This would increase the space complexity too much.

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -4290,3 +4306,633 @@ sp_reset_connection_internal(PG_FUNCTION_ARGS)
...
+	
+	/* Find and delete the tuple */
+	if (HeapTupleIsValid(tuple))
+	{
+		GetUserIdAndSecContext(&save_userid, &save_sec_context);
+		 
+		PG_TRY();
+		{
+			/* Set current user to bbf_role_admin for delete permissions */
+			sql_dialect = SQL_DIALECT_TSQL;
```
Deepesh125: should be outside of TRY block

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -4290,3 +4306,633 @@ sp_reset_connection_internal(PG_FUNCTION_ARGS)
...
+			SetUserIdAndSecContext(get_bbf_role_admin_oid(), save_sec_context | SECURITY_LOCAL_USERID_CHANGE);
+			simple_heap_delete(relation, &tuple->t_self);
```
Deepesh125, line 4777: do `CommandCounterIncrement();`

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -4290,3 +4306,633 @@ sp_reset_connection_internal(PG_FUNCTION_ARGS)
...
+		values[6] = (Datum)0;
+	}
+	tuple = heap_form_tuple(RelationGetDescr(relation), values, nulls);
+
+	GetUserIdAndSecContext(&save_userid, &save_sec_context);
+	
+	PG_TRY();
+	{
+		/* Set current user to bbf_role_admin for insert permissions */
+		sql_dialect = SQL_DIALECT_TSQL;
```
Deepesh125: should be outside of TRY

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -4290,3 +4306,633 @@ sp_reset_connection_internal(PG_FUNCTION_ARGS)
...
+		simple_heap_insert(relation, tuple);
+
+		/*
+		* Switch to TopMemoryContext to ensure the bitmap set persists across transactions.
+		* This prevents the bitmap set from being freed when the current memory context
+		* is reset, which would cause the handles to be lost and potentially reused.
+		*/
+		oldContext = MemoryContextSwitchTo(TopMemoryContext);
+		active_xml_handles_counter = bms_add_member(active_xml_handles_counter, current_xml_handle_counter);
+		MemoryContextSwitchTo(oldContext);
```
Deepesh125, line 4678: do `CommandCounterIncrement();`

Deepesh125 (changes requested):

contrib/babelfishpg_tsql/src/procedures.c
harshdubey166: Ok, did that

contrib/babelfishpg_tsql/src/procedures.c
harshdubey166: Added that

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -4290,3 +4306,637 @@ sp_reset_connection_internal(PG_FUNCTION_ARGS)
...
+	/* Generate a unique input string for the MD5 hash */
+	char *random_input = psprintf("%s%d", base_name, rand());
+	
+	/* Generate MD5 hash */
+	success = pg_md5_hash(random_input, strlen(random_input), md5, &errstr);
+	
+	if (unlikely(!success))
+		ereport(ERROR,
+			(errcode(ERRCODE_INTERNAL_ERROR),
+			 errmsg("could not compute MD5 hash: %s", errstr)));
```
Deepesh125: Little bit more context is required

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -4290,3 +4306,637 @@ sp_reset_connection_internal(PG_FUNCTION_ARGS)
...
+		*/
+		address = DefineRelation(stmt, RELKIND_RELATION, get_bbf_role_admin_oid(), NULL, NULL);
+
+		/* 
+		* Create a TOAST table for the relation if needed.
+		* TOAST tables store large values that don't fit in the main table pages.
+		* This is important for XML data which can be large.
+		* The second parameter (0) means no special options for the TOAST table.
+		*/
+		NewRelationCreateToastTable(address.objectId, (Datum)0);
```
Deepesh125, line 4537: Are we doing similar for other places? In the engine?

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -4290,3 +4306,637 @@ sp_reset_connection_internal(PG_FUNCTION_ARGS)
...
+	}
+	tuple = heap_form_tuple(RelationGetDescr(relation), values, nulls);
+
+	GetUserIdAndSecContext(&save_userid, &save_sec_context);
+	sql_dialect = SQL_DIALECT_TSQL;
+	
+	PG_TRY();
+	{
+		/* Set current user to bbf_role_admin for insert permissions */
+		SetUserIdAndSecContext(get_bbf_role_admin_oid(), save_sec_context | SECURITY_LOCAL_USERID_CHANGE);
```
Deepesh125: Is this really required? I don't think simple_heap_insert does any perms checks

contrib/babelfishpg_tsql/src/procedures.c
harshdubey166: Added more error details.

contrib/babelfishpg_tsql/src/procedures.c
harshdubey166: Removed perms checks from simple_heap_insert and simple_heap_delete. We only need at the time of table creation.

contrib/babelfishpg_tsql/src/procedures.c
harshdubey166: Yes toast table for a temp table is created though this function in the engine also.

Deepesh125: Is this applicable for ENR also?

harshdubey166: Yes, in case of ENRs also we create toast table by that function.

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -4290,3 +4306,619 @@ sp_reset_connection_internal(PG_FUNCTION_ARGS)
...
+
+					systable_endscan(scan);
+					relation_close(relation, AccessShareLock);
+				}
+			}
+			/* If no entry exists in the table, we can use this handle */
+			if (!entry_exists)
+			{
+				handle_valid = true;
+			}
```
HarshLunagariya, line 4388: Just curious, When can we hit this?

test/JDBC/expected/sp_xml_preparedocument-vu-verify.out
```diff
@@ -0,0 +1,495 @@
+
```
HarshLunagariya, line 1: Do we see possibility of adding testcase for following scenario?
Current user tries dropping internally created temp table.

Changes looks good overall!

harshdubey166: I have tested this locally. When a user less than `bbf_role_admin` permissions tries to drop this temp table, it shows permission denied for that. Since we are using random name for the temp created everytime, so it is not feasible to add automated tests for them .

contrib/babelfishpg_tsql/src/procedures.c
harshdubey166: When we start a transaction and insert some entries into the temp table by preparing some handles and after the transaction is rolled back, those entries from the temp table are lost but the `active_xml_handles_counter` in the bitmap set are not rolled back. In this case, we have set this condition that since the entry does not exist in the table, we can still use this handle after all handles are exhausted.

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -4290,3 +4306,619 @@ sp_reset_connection_internal(PG_FUNCTION_ARGS)
...
+			int                    document_id = 2 * current_xml_handle_counter - 1;
+			Relation               relation = NULL;
+			ScanKeyData            skey[1];
+			SysScanDesc            scan;
+			HeapTuple              tuple;
+			bool                   entry_exists = false;
+			EphemeralNamedRelation enr = NULL;
+			
+			/* Only check the table if it exists using ENR lookup */
+			if (xml_handle_temp_table_name != NULL)
```
Yvinayak07: I think we can combine this condition with the above if condition.

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -4290,3 +4306,619 @@ sp_reset_connection_internal(PG_FUNCTION_ARGS)
...
+ * gets a unique temp table name, avoiding conflicts when multiple sessions are
+ * creating these tables simultaneously.
+ *
+ * Parameters:
+ *   base_name - The base name to use for the table
+ *
+ * Returns:
+ *   A unique table name with the base name and an MD5 hash suffix
+ */
+#define MD5_HASH_LEN 32
```
Yvinayak07: we should declare this above in this file.

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -4290,3 +4306,619 @@ sp_reset_connection_internal(PG_FUNCTION_ARGS)
...
+	MemoryContext           oldContext = NULL;
+
+	/* Check if the table exists using ENR lookup */
+	if (xml_handle_temp_table_name != NULL)
+	{
+		enr = get_ENR(currentQueryEnv, xml_handle_temp_table_name, true);
+		if (enr)
+		{
+			relation = relation_open(enr->md.reliddesc, RowExclusiveLock);
+			table_exists = true;
```
Yvinayak07, line 4591: seems like this code part is used at many places. could we make a function for this?

contrib/babelfishpg_tsql/src/procedures.c
harshdubey166: Yes included that in the above if condition

contrib/babelfishpg_tsql/src/procedures.c
harshdubey166: Declared that above in the file.

contrib/babelfishpg_tsql/src/procedures.c
harshdubey166: If we create a function for this we have to add additional checks for null cases also, i.e, if relation becomes null. So to avoid those we are using this.

rohit01010 (approved):

Yvinayak07 (approved):

contrib/babelfishpg_tsql/src/procedures.c
Deepesh125: Is that expected TSQL?

harshdubey166: Yes

Deepesh125 (approved):

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -4290,3 +4307,620 @@ sp_reset_connection_internal(PG_FUNCTION_ARGS)
...
+	active_xml_handles_counter = NULL;
+
+	/* Reset xml handles */
+	current_xml_handle_counter = XML_HANDLE_COUNTER_INVALID;
+
+	/* Reset the table name */
+	xml_handle_temp_table_name = NULL;
+
+	/* Reset the query environment */
+	pltsql_remove_current_query_env();
```
ayushdsh, line 4823: Nit: explicitly calling this is not a problem, but we need to determine the reason why wasn't it dropped during reset conn.

ayushdsh (approved):

PR #3465, opened 2025-02-07
Fix upgrade script to resolve MVU issue from 15.11

Anikait143:
### Description
- Fix MVU issue from 15.11 by fixing upgrade script `babelfishpg_tsql--4.2.0--4.3.0.sql` to prepend `PG_CATALOG` schema on string_agg function call

### Issues Resolved
BABEL-5595

* **Use case based -**

* **Boundary conditions -**

* **Arbitrary inputs -**

* **Negative test cases -**

* **Performance tests -**

* **Tooling impact -**

* **Client tests -**

### Check List

### DIFF
--- a/contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--4.2.0--4.3.0.sql
+++ b/contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--4.2.0--4.3.0.sql
@@ -6719,7 +6719,7 @@ CREATE OR REPLACE VIEW information_schema_tsql.columns_internal AS
 			CAST(ext.orig_name AS sys.nvarchar(128)) AS "TABLE_SCHEMA",
 			CAST(
 				COALESCE(
-					(SELECT string_agg(
+					(SELECT pg_catalog.string_agg(
 						CASE
 						WHEN option LIKE 'bbf_original_rel_name=%' THEN substring(option, 23 /* prefix length */)
 						ELSE NULL
@@ -6730,7 +6730,7 @@ CREATE OR REPLACE VIEW information_schema_tsql.columns_internal AS
 
 			CAST(
 				COALESCE(
-					(SELECT string_agg(
+					(SELECT pg_catalog.string_agg(
 						CASE
 						WHEN option LIKE 'bbf_original_name=%' THEN substring(option, 19 /* prefix length */)
 						ELSE NULL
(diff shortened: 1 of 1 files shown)


rohit01010 (approved):

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--4.2.0--4.3.0.sql
```diff
@@ -6719,7 +6719,7 @@ CREATE OR REPLACE VIEW information_schema_tsql.columns_internal AS
 			CAST(ext.orig_name AS sys.nvarchar(128)) AS "TABLE_SCHEMA",
 			CAST(
 				COALESCE(
-					(SELECT string_agg(
+					(SELECT pg_catalog.string_agg(
```
ahmed-shameem, line 6722: This should fix the issue. Although I think it would be better to add PG_CATALOG for all the string_agg even if it is not invoked from inside COALESCE function as that would be a good coding practice. There are such occurrences in other upgrade scripts where we call string_agg instead of PG_CATALOG.string_agg which  are not invoked from inside COALESCE function. I think we should update that.

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--4.2.0--4.3.0.sql
```diff
@@ -6719,7 +6719,7 @@ CREATE OR REPLACE VIEW information_schema_tsql.columns_internal AS
 			CAST(ext.orig_name AS sys.nvarchar(128)) AS "TABLE_SCHEMA",
```
ahmed-shameem, line 6719: Let's update the search path to ```sys, pg_catalog, public``` to be consistent.

Anikait143: This needs more investigation as to why we have different search paths, will take this up later

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--4.2.0--4.3.0.sql
Anikait143: For now, we have verified locally as well as in pre-prod upgrades and this change fixes the issue, as far as other occurences are concerned we have closely verified that either of the 2 definitions would work correctly with no differences and for prepending all such occurences there is a long list of previous such function calls which will not affect us but we can update them as a part of future changes

ahmed-shameem (changes requested):
Let's update the description of the PR as this is not only true for 15.11 but from previous versions as well.

ahmed-shameem (approved):

rishabhtanwar29 (approved):

PR #3495, opened 2025-02-14
Fix OBJECT_ID() to exclude indexes

roshan0708:
### Description

This commit modifies `OBJECT_ID` function to exclude indexes when searching for objects including temp table as well, ensuring that it correctly returns OIDs for constraints and other objects without mistakenly returning index OIDs.

### Issues Resolved

JIRA-5303

* **Use case based -** y

* **Boundary conditions -**

* **Arbitrary inputs -**

* **Negative test cases -**

* **Performance tests -**

* **Tooling impact -**

* **Client tests -**

### Check List


contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -2361,13 +2361,39 @@ object_id(PG_FUNCTION_ARGS)
 			} 
 			else if (enr == NULL)
 			{
-				result = get_relname_relid((const char *) object_name, LookupNamespaceNoError("pg_temp"));
+				Oid temp_nsp_oid = LookupNamespaceNoError("pg_temp");
+				if (OidIsValid(temp_nsp_oid))
+				{
+					result = get_relname_relid((const char *) object_name, temp_nsp_oid);
+				}
```
thephantomthief: Is this really necessary? I'm assuming even before if pg_temp lookup fails we would still set result as `InvalidOid`

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -2361,13 +2361,39 @@ object_id(PG_FUNCTION_ARGS)
...
+					{
+						result = relid;
+					}
+				}
+
+				ReleaseSysCache(class_tuple);
+			}
 
 			if (OidIsValid(relid) && pg_class_aclcheck(relid, user_id, ACL_SELECT) == ACLCHECK_OK)
 			{
```
thephantomthief, line 2392: Should we remove this check then?

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -2361,13 +2361,39 @@ object_id(PG_FUNCTION_ARGS)
...
+				/* Check if the relation is not an index */
+				if (classform->relkind != RELKIND_INDEX)
+				{
+					relid = classform->oid;
+
+					/* Perform ACL check */
+					if (pg_class_aclcheck(relid, user_id, ACL_SELECT) == ACLCHECK_OK)
+					{
+						result = relid;
+					}
```
thephantomthief: we should also check for `OidIsValid(relid)`

contrib/babelfishpg_tsql/runtime/functions.c
roshan0708: You're right it's not strictly needed, but I think it adds clarity and safety. Unless it slows things down, I'd like to keep it.

contrib/babelfishpg_tsql/runtime/functions.c
roshan0708: Good catch! I missed removing this. Thanks for spotting that.

thephantomthief (approved):

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -2361,17 +2361,38 @@ object_id(PG_FUNCTION_ARGS)
...
 		{
-			/* search in pg_class by name and schema oid */
-			Oid			relid = get_relname_relid((const char *) object_name, schema_oid);
+			HeapTuple	class_tuple;
+			Oid			relid = InvalidOid;
+
+			/* Perform a single lookup in pg_class by name and schema oid */
+			class_tuple = SearchSysCache2(RELNAMENSP, CStringGetDatum(object_name), ObjectIdGetDatum(schema_oid));
 
-			if (OidIsValid(relid) && pg_class_aclcheck(relid, user_id, ACL_SELECT) == ACLCHECK_OK)
```
tanscorpio7, line 2379: We can just add another condition `get_rel_relkind(relid) != RELKIND_INDEX`

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -2361,17 +2361,38 @@ object_id(PG_FUNCTION_ARGS)
 			} 
 			else if (enr == NULL)
 			{
-				result = get_relname_relid((const char *) object_name, LookupNamespaceNoError("pg_temp"));
+				Oid temp_nsp_oid = LookupNamespaceNoError("pg_temp");
+				if (OidIsValid(temp_nsp_oid))
+				{
+					result = get_relname_relid((const char *) object_name, temp_nsp_oid);
```
tanscorpio7: `get_relname_relid` is used in multiple other places in OBJECT ID. Do we need to add relkind check at all the places ?

tanscorpio7 (changes requested):

contrib/babelfishpg_tsql/runtime/functions.c
roshan0708: We don't need to check this at all the places since only constraints can be confused as indexes.

tanscorpio7: What about cases like `SELECT OBJECT_ID('some constraint name', 's')`

tanscorpio7 (approved):

thephantomthief (approved):

shardgupta (approved):

PR #3499, opened 2025-02-18
SELECT INTO does not keep nullability or identity

thephantomthief:
### Description

Previously, if the source table contains an identity column or a not
nullable column, then the destination table where these columns are
SELECTed INTO did not keep their identity or nullability property.

In this commit, if the columns in the INTO clause, have the identity or
not nullable property attributed with them, we will implicitly create an
ALTER TABLE statement that would do the following:

 1. NOT NULL: ALTER TABLE dest_rel ALTER COLUMN col SET NOT NULL
 2. IDENTITY: ALTER TABLE dest_rel ALTER COLUMN col ADD GENERATED ...
              AS IDENTITY

Task: BABEL-5485
Signed-off-by: Sharu Goel <goelshar@amazon.com>

### Issues Resolved

BABEL-5485

### Check List
- [x] Commits are signed per the DCO using --signoff 

By submitting this pull request, I confirm that my contribution is under the terms of the Apache 2.0 and PostgreSQL licenses, and grant any person obtaining a copy of the contribution permission to relicense all or a portion of my contribution to the PostgreSQL License solely to contribute all or a portion of my contribution to the PostgreSQL open source project.

For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/babelfish-for-postgresql/babelfish_extensions/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #3534, opened 2025-02-26
SELECT INTO does not keep nullability or identity (#3499)

thephantomthief:
### Description

Previously, if the source table contains an identity column or a not nullable column, then the destination table where these columns are SELECTed INTO did not keep their identity or nullability property.

In this commit, if the columns in the INTO clause, have the identity or not nullable property attributed with them, we will implicitly create an ALTER TABLE statement that would do the following:

 1. IDENTITY: ALTER TABLE dest_rel ADD COLUMN col. Where col is an IDENTITY column with the same properties as the source column.
 2. NOT NULL: ALTER TABLE dest_rel ALTER COLUMN col SET NOT NULL

Task: BABEL-5485

### Check List

### DIFF
--- a/contrib/babelfishpg_tsql/src/pl_handler.c
+++ b/contrib/babelfishpg_tsql/src/pl_handler.c
@@ -48,6 +48,7 @@
 #include "nodes/pg_list.h"
 #include "parser/analyze.h"
 #include "parser/parser.h"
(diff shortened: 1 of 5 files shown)


jsudrik (approved):
Cherry pick from the other branch.

PR #3560, opened 2025-03-09
Objects should be owned by schema owner in Babelfish

shalinilohia50:
### Description

When a new object is created in Babelfish, its owner should be same as its schema owner.
Upgrade approach:
1. Older objects will continue to have the owner (which can be different than the schema owner)
2. There are functions provided to check the objects where there is owner mismatch between object and its schema. User can run the function and update the ownership of mismatched objects themselves.

### Issues Resolved

Task: BABEL-5361

* **Use case based -** Added
* **Boundary conditions -**
* **Arbitrary inputs -**
* **Negative test cases -**
* **Performance tests -**
* **Tooling impact -**
* **Client tests -**


contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -5806,16 +5805,22 @@ pltsql_get_object_owner(Oid namespaceId, Oid ownerId)
...
-				has_privs_of_role(GetUserId(), db_ddladmin)))
+			/*
+			 * Ignore cases when:
+			 * 1. Object owner is dbo AND current user is a member of db_ddladmin
+			 * 2. Schema owner is dbo, It happens when CREATE permission is granted via PG port
+			 */
+			if ((schema_db_id == get_cur_db_id()) &&
+				(((ownerId != dbo_oid) && has_privs_of_role(GetUserId(), db_ddladmin))
+				|| (nsp_owner != dbo_oid)))
 				ownerId = nsp_owner;
```
tanscorpio7: Could you explain why we need these special handling ?

tanscorpio7 (changes requested):

contrib/babelfishpg_tsql/src/hooks.c
shalinilohia50: 1. If we don't add this check, it will set current user 'l2' (a member of ddladmin) as object owner which is wrong. We should have dbo as the object_owner. The user 'l2' got the DML privileges, that's where it was detected.
2. CREATE privilege is granted on `dbo` schema since in `dbo` schema no one other than sysadmin can create objects, we want to skip those cases. Examples in existing tests.

shalinilohia50: We need this as there were a lot of existing tests where we have granted CREATE on dbo schema via PG Port. 
If the schema is dbo, only change the object owner if the current user is a member of db_owner or db_ddladmin. For other case, that would mean that permission was granted via PG Port.

tanscorpio7: I think we should "always" keep schema owners as object owner for objects created from TDS endpoint.

shalinilohia50: We are keeping that same for schemas other than dbo, there are many tests relying on this fact. As discussed with Alex, the object owner won't necessarily be same as schema owner if user interferes via PG port.

tanscorpio7: I am still not clear why fixing this for create granted though PG port is an issue. If it is just a test issue, can we mend the tests instead ?

tanscorpio7: Plus figuring out if a CREATE privilege was granted from PG endpoint or TDS endpoint without help from catalog metadata seems dodgy.

shalinilohia50: > I am still not clear why fixing this for create granted though PG port is an issue. If it is just a test issue, can we mend the tests instead ?

We can fix this actually. CREATE table runs some ALTER statement internally. In case of GRANTs via PG port, current user doesn't have enough privilege to execute ALTER statement.
We should set the current user to schema owner for ALTER statement in TSQL dialect since we are now saying that table owner would be same as schema owner.

tanscorpio7: Makes sense to me. We can also rope in @HarshLunagariya @anju15bharti in this discussion.

shalinilohia50: Discussed with Alex:
We should let these CREATE from TSQL endpoint fail since this extra handling is done in Babelfish for CREATE statements.  We should not modify our TSQL code to handle any interoperability behaviour. Earlier users could create, after the upgrade they won't have CREATE privilege in BBF schema if permission granted via PG port. We will document this in the release notes.

tanscorpio7: Agreed, as long as we have consistent ownership for new objects created from Babelfish endpoint.

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -4079,16 +4079,6 @@ bbf_ProcessUtility(PlannedStmt *pstmt,
...
-					/* Grant ALL schema privileges to the user.*/
-					if (rolspec && strcmp(queryString, CREATE_LOGICAL_DATABASE) != 0)
-					{
-						int i;
-						for (i = 0; i < NUMBER_OF_PERMISSIONS; i++)
-						{
-							/* Execute the GRANT SCHEMA subcommands. */
-							exec_grantschema_subcmds(create_schema->schemaname, rolspec->rolename, true, false, permissions[i], true);
-						}
-					}
```
HarshLunagariya, line 4090: Why are we removing this logic?

test/JDBC/expected/PARTITION-vu-verify.out
```diff
@@ -1291,14 +1291,17 @@ partition_u1
...
+/** CREATE PERMISSION GRANTED VIA PG PORT MAY NOT TAKE EFFECT ON TSQL OBJECTS **/
+/** NOT A RECOMMENDED WAY **/
 CREATE TABLE PartitionDb_TestPartitionTable (
     Id INT,
     Value sys.varchar(20)
 ) ON PartitionDb_PartitionScheme(Id);
 GO
+~~ERROR (Code: 33557097)~~
+
+~~ERROR (Message: must be owner of table partitiondb_testpartitiontable)~~
```
HarshLunagariya, line 1303: What is the decided behaviour about above interoperability scenario?

HarshLunagariya: Looks like we are removing some test scenarios, Let's create task to track and assign it to author.

test/JDBC/expected/jira-BABEL-3504-upgrade-vu-cleanup.out
```diff
@@ -1,4 +1,4 @@
--- tsql user=jira_babel_3504_log1 password=123456789
+-- tsql
```
HarshLunagariya, line 1: Why are we not using jira_babel_3504_log1 ?

test/JDBC/expected/jira-BABEL-3504-vu-cleanup.out
```diff
@@ -104,7 +104,6 @@ WHERE sys.suser_name(usesysid) = 'jira_babel_3504_log1' AND backend_type = 'clie
 GO
 ~~START~~
 bool
-t
```
HarshLunagariya: nit: Redundant query, We can remove it.

test/JDBC/expected/object_ownership.out
```diff
@@ -0,0 +1,701 @@
...
+    '2025-01-01'
+);
+go
+
+create partition scheme ps2
+as partition pf2 
+all to ([primary]);
+go
+
+/** CREATE OBJECTS IN DBO SCHEMA USING DBO USER **/
```
HarshLunagariya: This comment seems incorrect.

test/JDBC/expected/object_ownership.out
```diff
@@ -0,0 +1,701 @@
...
+    '2023-01-01', 
+    '2024-01-01',
+    '2025-01-01'
+);
+go
+
+create partition scheme ps2
+as partition pf2 
+all to ([primary]);
+go
```
HarshLunagariya, line 58: Can we please add index , trigger, table type , type creation cases for objects for `/** SCHEMA OWNER AND CURRENT USER ARE SAME **/` as well?

test/JDBC/expected/object_ownership.out
```diff
@@ -0,0 +1,701 @@
...
+create table object_ownership_s3.pt2
+(
+    saleid int identity(1,1),
+    saledate datetime,
+    amount decimal(10,2),
+    customerid int,
+    productid int
+) on ps2(saledate);  -- specify the partitioning column
+go
+
```
HarshLunagariya, line 236: Can we please add queries to verify the pg metadata for ownership of object?

HarshLunagariya (changes requested):

contrib/babelfishpg_tsql/src/pl_handler.c
shalinilohia50: We don't need this logic now, object owner is the schema owner so it will have the required privileges which we explicitly gave earlier.

test/JDBC/expected/PARTITION-vu-verify.out
shalinilohia50: BABEL-5681 Will be tracking it. We should have appropriate CREATE permission via TSQL port. Any CREATE permission via PG may not take effect as earlier.

test/JDBC/expected/jira-BABEL-3504-upgrade-vu-cleanup.out
shalinilohia50: In this test, we were granting the CREATE privilege via PG port. Sequence is getting created, but the owner would be dbo which is the schema owner. Now, `jira_babel_3504_log1` doesn't have the privilege to drop the sequence, which is expected.

test/JDBC/upgrade/16_8/schedule
```diff
@@ -598,4 +598,5 @@ test_conv_money_to_varchar
 fixeddecimal_modulo
 BABEL-NUMERIC_EXPR
 test_conv_float_to_varchar_char
+object_ownership-before-16_8-or-17_4
```
HarshLunagariya, line 601: Why are we only adding this to 16_8 schedule file? Shouldn't we add for few other previous versions as well?

shalinilohia50: In the older versions, we had more restrictive privileges due to which there were more "permission denied" errors after upgrade. 
We are not losing any test coverage. Adding this test in the older version means adding redundant files for `before-16_7` which I think we can skip.

HarshLunagariya (approved):

tanscorpio7 (approved):

shardgupta (approved):

PR #3565, opened 2025-03-10
Handle storing original index name through cached plan

rishabhtanwar29:
### Description
Previously, index created through cached plan was being created with a wrong/
invalid name as we were modifying the index name in index statement with composed
index name. This composed index name was being allocated in the `SPI` context which
gets destroyed at the end of the query execution so index name in the cached plan was
pointing to a freed memory address.

Fix this by keeping the original index in the IndexStmt node so that it still remains in the
cached plan.

Task: BABEL-1517

* **Use case based -**

* **Boundary conditions -**

* **Arbitrary inputs -**

* **Negative test cases -**

* **Performance tests -**

* **Tooling impact -**

* **Client tests -**

### Check List

### DIFF
--- a/contrib/babelfishpg_tsql/src/pl_handler.c
+++ b/contrib/babelfishpg_tsql/src/pl_handler.c
@@ -4354,7 +4354,7 @@ bbf_ProcessUtility(PlannedStmt *pstmt,
 				if (sql_dialect == SQL_DIALECT_TSQL &&
 					strcmp(queryString, CREATE_FULLTEXT_INDEX) != 0) /* Skip fulltext indexes since they don't even have an original name */
 				{
-					char    	*original_name = stmt->idxname != NULL ? pstrdup(stmt->idxname) : NULL;
+					char    	*original_name = stmt->idxname != NULL ? stmt->idxname : NULL;
 					List    	*partition_schemes = stmt->excludeOpNames;
 
 					stmt->excludeOpNames = NIL;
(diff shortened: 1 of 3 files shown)


contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -4382,7 +4382,12 @@ bbf_ProcessUtility(PlannedStmt *pstmt,
 						}
 					}
 					if (original_name && !stmt->isconstraint)
+					{
+						/* Store the original index name in reloptions */
 						exec_add_original_index_name(stmt->idxname, stmt->relation->schemaname, original_name);
+						/* Restore the original index name so that cached plan remains valid */
+						stmt->idxname = original_name;
+					}
```
tanscorpio7, line 4402: Reset would not happen if we error out some where before. Can we copy the tree before working on it instead ?
In standard_ProcessUtility
```
	/*
	 * If the given node tree is read-only, make a copy to ensure that parse
	 * transformations don't damage the original tree.  This could be
	 * refactored to avoid making unnecessary copies in more cases, but it's
	 * not clear that it's worth a great deal of trouble over.  Statements
	 * that are complex enough to be expensive to copy are exactly the ones
	 * we'd need to copy, so that only marginal savings seem possible.
	 */
```

tanscorpio7 (changes requested):

contrib/babelfishpg_tsql/src/pl_handler.c
rishabhtanwar29: Makes sense, copied the plannedstmt.

tanscorpio7 (approved):

KushaalShroff (approved):

PR #3567, opened 2025-03-11
Improve round() function return type handling based on input arguments data types

fizzlr:
### Description

Make the round() function clearer on return types based on argument types and number of arguments.

Changes:

1. Add lookup functions for data types that are represented differently
2. Introduce PLpgSQL functions for each return type to reuse the existing underlying C functions
3. Implement logic to determine return types based on argument types and number of arguments
4. Add tests to ensure comprehensive coverage

### Check List


contrib/babelfishpg_tsql/src/pltsql_coerce.c
```diff
@@ -426,7 +426,11 @@ tsql_special_function_t tsql_special_function_list[] =
 	{"sys", "lower", "lower", false, 1},
 	{"sys", "upper", "upper", false, 1},
 	{"sys", "concat", "concat", true, 0},
-	{"sys", "concat_ws", "concat_ws", true, 1}
+	{"sys", "concat_ws", "concat_ws", true, 1},
+
+	/* Since sql server round() function can take either 2 or 3 parameters, it doesn't fit into case of variable length of paraemters */
```
hopebo, line 431: I'm not so familiar with this piece of codes. Can you explain to me what will happen if we don't add the codes here?
I see the call with number of arguments not 2 or 3 has been banned in following codes.

contrib/babelfishpg_tsql/src/pltsql_coerce.c
```diff
@@ -1353,6 +1357,58 @@ tsql_func_select_candidate_for_special_func(List *names, int nargs, Oid *input_t
...
 	}
+	/* The return type of sys.round() will be based on input argument type */
+	else if (strlen(proc_name) == 5 && strncmp(proc_name, "round", 5) == 0)
+	{
+		/*
+		 * 2 and 3 are the only valid options for nargs of sys.round() in sql server 2022 (version 16), but since the version supported here
+		 * is version 12, will follow the existing behavior.
+		 * The existing behavior throws error message if nargs is not from 1 to 3.
+		 */
+		if (nargs == 2 || nargs == 3)
```
hopebo, line 1368: So you also fix the issue of this function could be called with only one argument in this patch. Do we have test case for that?

contrib/babelfishpg_tsql/src/pltsql_coerce.c
```diff
@@ -1353,6 +1357,58 @@ tsql_func_select_candidate_for_special_func(List *names, int nargs, Oid *input_t
...
+		{
+			/* tinyint, smallint and int will have return type int */
+			if ((*common_utility_plugin_ptr->is_tsql_tinyint_datatype)(new_input_typeids[0])
+			|| (new_input_typeids[0] == INT2OID)
+			|| (new_input_typeids[0] == INT4OID))
+			{
+				expr_result_type = (*common_utility_plugin_ptr->lookup_tsql_datatype_oid) ("int");
```
hopebo, line 1375: What's the difference here if we directly set it to INT4OID instead of `lookup_tsql_datatype_oid`, since you already use it to compare with the type of argument.

contrib/babelfishpg_tsql/src/pltsql_coerce.c
```diff
@@ -1353,6 +1357,58 @@ tsql_func_select_candidate_for_special_func(List *names, int nargs, Oid *input_t
...
+			}
+			/* bigint will have return type bigint */
+			else if (new_input_typeids[0] == INT8OID)
+			{
+				expr_result_type = (*common_utility_plugin_ptr->lookup_tsql_datatype_oid) ("bigint");
```
hopebo, line 1380: ditto

contrib/babelfishpg_tsql/src/pltsql_coerce.c
fizzlr: otherwise data type will be incorrect, since the oid for that in sql server and babelfish are different.

contrib/babelfishpg_tsql/src/pltsql_coerce.c
fizzlr: otherwise data type will be incorrect, since the oid for that in sql server and babelfish are different.

contrib/babelfishpg_tsql/src/pltsql_coerce.c
```diff
@@ -1353,6 +1357,58 @@ tsql_func_select_candidate_for_special_func(List *names, int nargs, Oid *input_t
...
+			}
+			/* float and real will have return type float */
+			else if (new_input_typeids[0] == FLOAT8OID
+			|| new_input_typeids[0] == FLOAT4OID)
+			{
+				expr_result_type = (*common_utility_plugin_ptr->lookup_tsql_datatype_oid) ("float");
+			}
+			else
+			{
+				expr_result_type = (*common_utility_plugin_ptr->lookup_tsql_datatype_oid) ("decimal");
```
hopebo: Do we have situation hitting the last `else` branch? From what I see in SQL Server doc, all situations have been covered above.

contrib/babelfishpg_tsql/sql/sys_functions.sql
```diff
@@ -1570,11 +1570,84 @@ AS 'babelfishpg_tsql', 'datalength' LANGUAGE C IMMUTABLE STRICT PARALLEL SAFE;
...
+CREATE OR REPLACE FUNCTION sys.round(number sys.BIGINT, length INTEGER)
+RETURNS sys.BIGINT
+AS $$
+BEGIN
+    RETURN sys.round(number::PG_CATALOG.NUMERIC, length);
+END;
+$$
+LANGUAGE plpgsql IMMUTABLE PARALLEL SAFE;
+
+CREATE OR REPLACE FUNCTION sys.round(number sys.BIGINT, length INTEGER, function INTEGER)
```
hopebo, line 1607: I think it could work if you only define functions for `sys.BIGINT`. For other integer types, it will automatically pick up the function for `BIGINT`.

test/JDBC/input/BABEL-2057.sql
```diff
@@ -0,0 +1,193 @@
...
+select round(cast(42 as smallmoney), 0)
+go
+
+select round(cast(42 as smallmoney), 1)
+go
+
+select round(cast(42 as smallmoney), 2)
+go
+
+select round(cast(42 as smallmoney), 999)
```
hopebo: Could you use some complex float number instead of integer `42` as much as possible to test?

hopebo (commented):
Overall, looks good to me. Some minor comments and questions left.

hopebo (commented):
Also, please squash your commits into one.

contrib/babelfishpg_tsql/src/pltsql_coerce.c
hopebo: But for input arguments, they always have the same oid as postgres?

contrib/babelfishpg_tsql/src/pltsql_coerce.c
fizzlr: here i think is more of including it to tsql_special_function_list that we need to treat them specially.

contrib/babelfishpg_tsql/src/pltsql_coerce.c
hopebo: What if ROUND(a, 2) while `a` is a column of babelfish `smallint`.

test/JDBC/expected/BABEL-2057.out
```diff
@@ -0,0 +1,492 @@
...
+~~END~~
+
+
+select round(cast(42 as tinyint), -2)
+go
+~~START~~
+int
+~~ERROR (Code: 33557097)~~
+
+~~ERROR (Message: value overflows for numeric format)~~
```
forestkeeper: Is this expected ?

test/JDBC/expected/BABEL-2057.out
```diff
@@ -0,0 +1,492 @@
...
+int
+40
+~~END~~
+
+
+select round(cast(42 as smallint), -2)
+go
+~~ERROR (Code: 33557097)~~
+
+~~ERROR (Message: value overflows for numeric format)~~
```
forestkeeper: Is this expected ?

test/JDBC/expected/BABEL-2057.out
```diff
@@ -0,0 +1,492 @@
...
+numeric
+42.00
+~~END~~
+
+
+select round(cast(42 as decimal), 999)
+go
+~~ERROR (Code: 33557097)~~
+
+~~ERROR (Message: Arithmetic overflow error for data type numeric.)~~
```
forestkeeper: What is this error msg from SQL Server ?

contrib/babelfishpg_tsql/src/pltsql_coerce.c
```diff
@@ -1353,6 +1357,58 @@ tsql_func_select_candidate_for_special_func(List *names, int nargs, Oid *input_t
...
+			}
+		}
+		else
+		{
+			ereport(ERROR,
+				(errcode(ERRCODE_INVALID_PARAMETER_VALUE),
+				 errmsg("ROUND function expects 2 or 3 arguments")));
```
forestkeeper: Where is the test case for this ?
What's the output of sql server for this ?

contrib/babelfishpg_tsql/src/pltsql_coerce.c
```diff
@@ -1353,6 +1357,58 @@ tsql_func_select_candidate_for_special_func(List *names, int nargs, Oid *input_t
 			expr_arg_type = (*common_utility_plugin_ptr->lookup_tsql_datatype_oid) ("bbf_varbinary");
 		}
 	}
+	/* The return type of sys.round() will be based on input argument type */
+	else if (strlen(proc_name) == 5 && strncmp(proc_name, "round", 5) == 0)
```
forestkeeper, line 1361: Case sensitive ?

contrib/babelfishpg_tsql/src/pltsql_coerce.c
fizzlr: will add a jira to document the behavior and future improvement if it need to

contrib/babelfishpg_tsql/src/pltsql_coerce.c
fizzlr: will put an assert here instead

contrib/babelfishpg_tsql/src/pltsql_coerce.c
fizzlr: Will test that

contrib/babelfishpg_tsql/src/pltsql_coerce.c
fizzlr: got it, will check sql server behavior

test/JDBC/input/BABEL-2057.sql
```diff
@@ -0,0 +1,193 @@
...
+select round(cast(748.58 as float), 0)
+go
+
+select round(cast(748.58 as float), 1)
+go
+
+select round(cast(748.58 as float), 2)
+go
+JBA
+select round(cast(748.58 as float), 3)
```
hopebo: Also supplement a test case with only one argument and comment on it.

hopebo (changes requested):

contrib/babelfishpg_tsql/src/pltsql_coerce.c
fizzlr: within this function, following the pattern to use lower case as other special function.

contrib/babelfishpg_tsql/src/pltsql_coerce.c
rohit01010: We do not need this else block, as the return type in other cases are not specified. For other cases lets have the existing behaviour.

contrib/babelfishpg_tsql/src/pltsql_coerce.c
rohit01010: Currently babelfish supports round with 1 arguments due to underlying PostgreSQL, if we do not have any good enough reason to block round function call with number of arguments other than 2 or 3, we should not block it. Lets keep the existing behaviour in this case, if we do not have any good enough reason to throw error.

rohit01010 (changes requested):

contrib/babelfishpg_tsql/src/pltsql_coerce.c
fizzlr: removed else block

contrib/babelfishpg_tsql/src/pltsql_coerce.c
fizzlr: also during my testing, this ereport is unable to be reached. The error about the number of arguments is caught else where already. Remove this ereport

contrib/babelfishpg_tsql/src/pltsql_coerce.c
fizzlr: after removing this else block, it will fail tests from BABEL-1193.out and babel_function.out. Annotated in the code: add else here for backward compatibility and passing old tests

test/JDBC/input/BABEL-2057.sql
fizzlr: added in test

test/JDBC/expected/BABEL-2057.out
fizzlr: yes

test/JDBC/expected/BABEL-2057.out
fizzlr: yes

test/JDBC/expected/BABEL-2057.out
fizzlr: for the overflow message shown in the .out file, sql server files just return 0 or .00

fizzlr: but since those are the existing babelfish behavior, I will leave them as that. If required in future for behavior change, we can revisit this.

contrib/babelfishpg_tsql/src/pltsql_coerce.c
hopebo: How do the old tests fail? We don't need to keep the backward compatibility. The behaviour of old version is wrong. We only need to keep consistent with SQL Server.

contrib/babelfishpg_tsql/src/pltsql_coerce.c
hopebo: If SQL Server doesn't allow the function call with only one argument, I hope you can ban it in this patch. It won't be so hard I think.

contrib/babelfishpg_tsql/src/pltsql_coerce.c
fizzlr: +~~ERROR (Code: 33557097)~~
+
+~~ERROR (Message: function round(text, integer) is not unique)~~

and another one for NULL case is:
+~~ERROR (Code: 33557097)~~
+
+~~ERROR (Message: function round(unknown, integer) is not unique)~~

fizzlr: Chatted about behavior for this, will remove else block and modify old tests accordingly and move them to new test file I added.

contrib/babelfishpg_tsql/src/pltsql_coerce.c
fizzlr: test added

contrib/babelfishpg_tsql/src/pltsql_coerce.c
fizzlr: resolved

test/JDBC/input/BABEL-2057.sql
fizzlr: added

hopebo (approved):

test/JDBC/input/round_return_type_test-vu-cleanup.sql
```diff
@@ -0,0 +1,23 @@
...
+GO
+DROP VIEW IF EXISTS dbo.RoundDemoView;
+GO
+
+-- Drop Functions
+DROP FUNCTION IF EXISTS dbo.RoundMultipleTypes;
+GO
+DROP FUNCTION IF EXISTS dbo.RoundFloat;
+GO
+DROP FUNCTION IF EXISTS dbo.RoundDecimal;
```
forestkeeper, line 14: This test need to be added to schedule files.

test/JDBC/expected/round-return-type.out
```diff
@@ -0,0 +1,638 @@
...
+~~ERROR (Code: 33557097)~~
+
+~~ERROR (Message: value overflows for numeric format)~~
+
+
+
+select round(cast(42 as decimal), 0)
+go
+~~START~~
+numeric
```
forestkeeper, line 198: I think for decimal, it should return decimal according to this doc : 
https://github.com/MicrosoftDocs/sql-docs/blob/live/docs/t-sql/functions/round-transact-sql.md

test/JDBC/expected/round-return-type.out
```diff
@@ -0,0 +1,638 @@
...
+
+select round(cast(42 as decimal), -2)
+go
+~~ERROR (Code: 33557097)~~
+
+~~ERROR (Message: value overflows for numeric format)~~
+
+
+
+select round(cast(42 as numeric), 0)
```
forestkeeper, line 242: For numeric, it should return decimal according to this doc : https://github.com/MicrosoftDocs/sql-docs/blob/live/docs/t-sql/functions/round-transact-sql.md

hopebo: Decimal is just an alias of numeric in SQL Server. Not sure if we need to pay more efforts to this.

test/JDBC/expected/round-return-type.out
fizzlr: You right. But after checking the logic in sys_functions.sql and pltsql_coerce.c, I've set them to be decimal already.

In https://learn.microsoft.com/en-us/sql/t-sql/data-types/decimal-and-numeric-transact-sql?view=sql-server-ver16, it says "decimal and numeric are numeric data types that have a fixed precision and scale. decimal and numeric are synonyms and can be used interchangeably."

test/JDBC/expected/round-return-type.out
fizzlr: Correct, but after checking the logic in sys_functions.sql and pltsql_coerce.c, I've set them to be decimal already.

In https://learn.microsoft.com/en-us/sql/t-sql/data-types/decimal-and-numeric-transact-sql?view=sql-server-ver16, it says "decimal and numeric are numeric data types that have a fixed precision and scale. decimal and numeric are synonyms and can be used interchangeably."

contrib/babelfishpg_tsql/src/pltsql_coerce.c
```diff
@@ -1353,6 +1357,48 @@ tsql_func_select_candidate_for_special_func(List *names, int nargs, Oid *input_t
...
+			else if ((*common_utility_plugin_ptr->is_tsql_money_datatype)(new_input_typeids[0])
+			|| (*common_utility_plugin_ptr->is_tsql_smallmoney_datatype)(new_input_typeids[0]))
+			{
+				expr_result_type = (*common_utility_plugin_ptr->lookup_tsql_datatype_oid) ("money");
+			}
+			/* float and real will have return type float */
+			/* in all other cases it will default to float data type */
+			else
+			{
+				expr_result_type = (*common_utility_plugin_ptr->lookup_tsql_datatype_oid) ("float");
```
fizzlr, line 1398: tested in sql server that other data types not mentioned in round sql server documentation will be defaulted to float data type. Matching that behavior here

test/JDBC/input/round_return_type_test-vu-cleanup.sql
fizzlr: added accordingly

contrib/babelfishpg_tsql/src/pltsql_coerce.c
hopebo: Good finding! So SQL Server allows other types other than listed in the doc as input, right?
Have you supplemented such test cases?

fizzlr: yes, added to round_return_type_test-vu-verify.sql

forestkeeper (approved):

PR #3570, opened 2025-03-12
Add support for ALTER VIEW and CREATE OR ALTER VIEW syntax on schema bound views in Babelfish

R4hul04:
### Description

Currently, Babelfish does not support ALTER VIEW and CREATE OR ALTER VIEW syntax statement in the TSQL dialect. With this change, Babelfish will support ALTER VIEW operations, allowing users to modify existing views. 

- We assume views to not have any dependencies for alter to succeed. 
- For dependent procedures, alter view will continue to work and will not be affected.
- Permissions on views won't be affected

The change was made to enhance Babelfish's compatibility with SQL Server by supporting the ALTER VIEW statement. This allows users to modify views in a manner consistent with SQL Server. 

[NOTE]: ALTER VIEW for views having dependencies will be supported once Babelfish supports weak schema bound views.

Engine PR - https://github.com/babelfish-for-postgresql/postgresql_modified_for_babelfish/pull/545

Signed off by: Rahul Parande rparande@amazon.com
### Issues Resolved

JIRA - BABEL-2017
* **Use case based -**
```
CREATE OR ALTER VIEW v1 AS SELECT a, b FROM t;
GO
SELECT * FROM v1;
GO
~~START~~
int#!#varchar
1#!#a
2#!#b
3#!#c
```
* **Boundary conditions -**
```
CREATE OR ALTER VIEW v1 AS SELECT a, b, c, d, e, f FROM t;
GO
~~ERROR (Code: 33557097)~~

~~ERROR (Message: column "e" does not exist)~~
```
```
-- Create view with existing object name [Error]
CREATE VIEW v1 AS SELECT a, b FROM t;
GO
~~ERROR (Code: 2714)~~

~~ERROR (Message: relation "v1" already exists)~~
```
* **Arbitrary inputs -**
```
ALTER VIEW v7 AS SELECT a, b, CAST(c AS date) AS c, d, e FROM t4;
GO
~~ERROR (Code: 33557097)~~

~~ERROR (Message: cannot cast type "decimal" to date)~~
```

* **Negative test cases -**
```
-- Error as dependent view exists
Drop table t;
GO
~~ERROR (Code: 3726)~~

~~ERROR (Message: cannot drop table t because other objects depend on it)~~
```

* **Performance tests -**

* **Tooling impact -**

* **Client tests -**

### Check List


test/JDBC/expected/alter-view-vu-cleanup.out
```diff
@@ -0,0 +1,43 @@
+-- Drop the views
```
forestkeeper, line 1: alter-view should be added into schedule of latest

contrib/babelfishpg_tsql/src/tsqlUnsupportedFeatureHandler.cpp
```diff
@@ -485,8 +485,8 @@ antlrcpp::Any TsqlUnsupportedFeatureHandlerImpl::visitCreate_or_alter_trigger(TS
 
 antlrcpp::Any TsqlUnsupportedFeatureHandlerImpl::visitCreate_or_alter_view(TSqlParser::Create_or_alter_viewContext *ctx)
 {
-	if (ctx->ALTER())
-		handle(INSTR_UNSUPPORTED_TSQL_ALTER_VIEW, "ALTER VIEW", getLineAndPos(ctx->ALTER()));
+	// if (ctx->ALTER())
+	// 	handle(INSTR_UNSUPPORTED_TSQL_ALTER_VIEW, "ALTER VIEW", getLineAndPos(ctx->ALTER()));
```
forestkeeper: Delete those line instead of commenting out.

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -2766,6 +2770,119 @@ bbf_ProcessUtility(PlannedStmt *pstmt,
...
 			}
+
+			/* 
+			* This case handles the `ALTER VIEW` and `CREATE OR ALTER VIEW` statements in Babelfish.
+			*/
+			case T_ViewStmt:
+			{
+				ViewStmt *stmt = (ViewStmt *) parsetree;
+				
+				if (sql_dialect == SQL_DIALECT_TSQL && (stmt->createoralter))
```
forestkeeper: In what scenario will createoralter be false ? ( if it's not create nor alter, then what's the sql ? )

contrib/babelfishpg_tsql/src/dbcmds.c
```diff
@@ -155,8 +156,15 @@ gen_createdb_subcmds(const char *dbname, const char *owner)
 	if (guest)
 		appendStringInfo(&query, "CREATE SCHEMA dummy AUTHORIZATION dummy; ");
 
+	old_dialect = sql_dialect;
+	sql_dialect = SQL_DIALECT_PG;
+	/* this query must be run in PG dialect or we will get a syntax error due to different
+	 * ALTER VIEW behavior between PG and TSQL
```
forestkeeper: Is this working in dump database only ?

contrib/babelfishpg_tsql/src/pl_handler.c
R4hul04: Since we are using `ViewStmt` node for alter view, createoralter flag will help us identify if it is CREATE OR ALTER/ ALTER VIEW statement. We have kept `createoralter = false` in `gram.y` for `CREATE VIEW`, `CREATE OR REPLACE VIEW`, `CREATE RECURSIVE VIEW`. Also we want to avoid `CREATE VIEW` query in our function

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -2766,6 +2770,119 @@ bbf_ProcessUtility(PlannedStmt *pstmt,
...
+						if (oldViewAcl)
+							pfree(oldViewAcl);
+						if (needCleanup)
+							EventTriggerEndCompleteQuery();
+					}
+					PG_END_TRY();
+					return; 
+				}
+				break;
+			}
```
forestkeeper, line 2888: So what if it's not match this if condition ? Is it possible? If not , let's add an assert in here.

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -2766,6 +2770,119 @@ bbf_ProcessUtility(PlannedStmt *pstmt,
...
+					*/
+					ObjectAddress address, originalView;
+					Oid oldViewOid;
+					Acl *oldViewAcl = NULL;
+					bool isCompleteQuery = (context != PROCESS_UTILITY_SUBCOMMAND);
+					bool needCleanup;
+			
+					if (!IS_TDS_CLIENT())
+					{
+						if(stmt->createoralter)
```
forestkeeper: IS_TDS_CLIENT() is enough to determine whether it's from psql endpoint or not.

contrib/babelfishpg_tsql/src/dbcmds.c
timchang514: what is the syntax error? why doesn't this issue occur with other shared syntaxes between PG and BBF?

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -2766,6 +2770,119 @@ bbf_ProcessUtility(PlannedStmt *pstmt,
...
+			
+							/* Create new view */
+							stmt->replace = true; 
+							address = DefineView(stmt, queryString, pstmt->stmt_location, pstmt->stmt_len);
+							CommandCounterIncrement();
+							
+							/* Store the view definition in babelfish_view_def */
+							if (store_view_definition_hook)
+								store_view_definition_hook(queryString, address);
+							CommandCounterIncrement();
```
timchang514: this should be inside the if stmt. 

or rather, should we throw an error if `babelfish_view_def` is not defined for some reason? what are the consequences if the code completes without storing the view in bbf_view_def?

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -2766,6 +2770,119 @@ bbf_ProcessUtility(PlannedStmt *pstmt,
...
+			
+							/* Update ACL info */
+							if (oldViewAcl)
```
timchang514: should we throw an error instead if oldViewAcl is not found? under what conditions could it be missing? what are consequences if it is missing?

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -2766,6 +2770,119 @@ bbf_ProcessUtility(PlannedStmt *pstmt,
...
+							{
+								pg_class_update_acl(address.objectId, oldViewAcl);
+								pfree(oldViewAcl);
+							}
+							CommandCounterIncrement();
+
+							if (isCompleteQuery)
+								EventTriggerDDLCommandEnd(parsetree);
```
timchang514: why is EventTriggerDDLCommandEnd  needed?

test/JDBC/input/alter/alter-view-vu-verify.sql
timchang514, line 1: Please add more test cases such as: 

- DROP VIEW and then create with same name. 
- Create view with existing object name (should throw error)
- Error scenarios that are not expected (eg. ALTER TABLE for dependencies, DROP TABLE, etc). 
- Test all the above in transactions (ROLLBACK, subtransaction, etc). 
- Error scenarios that are expected and in the code. 
- Ensure that no orphaned entries left behind as well, whether due to normal operation, rolback, or errors. Check pg and bbf catalogs both.

contrib/babelfishpg_tsql/src/pl_handler.c
timchang514: also, is it necessary for so many calls to CommandCounterIncrement?

contrib/babelfishpg_tsql/src/pl_handler.c
R4hul04: We always need to store view in `babelfish_view_def`. If that's the case should I remove if stmt? Also can I use `pltsql_store_view_definition` function instead?

contrib/babelfishpg_tsql/src/pl_handler.c
R4hul04: Maybe we should always update the acl, even though it is NULL?

contrib/babelfishpg_tsql/src/dbcmds.c
R4hul04: If I initialize the database without this change, it throws error : `syntax error at or near "OWNER"`. Postgres has it's own Alter View which does not modify views defining query (for that we use CREATE OR REPLACE) and is different from t-sql Alter view: https://www.postgresql.org/docs/current/sql-alterview.html

contrib/babelfishpg_tsql/src/pl_handler.c
R4hul04: The condition won't match if we are in TSQL dialect from non TDS client and we perform alter. Added an assert

contrib/babelfishpg_tsql/src/pl_handler.c
R4hul04: removed unnecessary EventTriggerDDLCommandEnd and CommandCounterIncrement

contrib/babelfishpg_tsql/src/dbcmds.c
timchang514: i worry that this change can have unintended side effects. after all, sql_dialect is a flag used in many many places across the code. my question is still, why does this happen with your change? specifically, can we debug the syntax error and see whether that is a case that we did not properly support?

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -4814,6 +4928,73 @@ pltsql_proc_get_oid_proname_proacl(AlterFunctionStmt *stmt, ParseState *pstate,
...
+    HeapTuple newtup;
+
+    /* Get the tuple from syscache */
+    classtup = SearchSysCache1(RELOID, ObjectIdGetDatum(newViewOid));
+    if (!HeapTupleIsValid(classtup))
+        elog(ERROR, "cache lookup failed for relation %u", newViewOid);
+
+    pg_class_rel = table_open(RelationRelationId, RowExclusiveLock);
+
+	memset(values, 0, sizeof(values));
```
timchang514, line 4994: nit: whitespace

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -2766,6 +2770,116 @@ bbf_ProcessUtility(PlannedStmt *pstmt,
...
+                 *
+                 * Since both CREATE VIEW and CREATE OR ALTER VIEW set replace = false initially,
+                 * we use the 'createOrAlter' flag to distinguish between them and implement the
+                 * correct behavior when a view already exists
+				 */
+
+				if (sql_dialect == SQL_DIALECT_TSQL && (stmt->createOrAlter))    
+				{
+					/*
+					*	1. Retrieve the OID of the old view using `RangeVarGetRelid()`.
```
timchang514: nit: here and above - block comment formatting. 


```
/*
 * lorem ipsum
 */
```

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -2766,6 +2770,116 @@ bbf_ProcessUtility(PlannedStmt *pstmt,
...
+							performDeletion(&originalView, DROP_RESTRICT, 0);
+							CommandCounterIncrement();
+			
+							/* Create new view */
+							stmt->replace = true; 
+							address = DefineView(stmt, queryString, pstmt->stmt_location, pstmt->stmt_len);
+							CommandCounterIncrement();
+							
+							/* Store the view definition in babelfish_view_def */
+							store_view_definition_hook(queryString, address);
```
timchang514: Typically when calling a hook we will wrap it in a check to ensure it exists. 

```
if (hook)
    hook(args)
```

I also don't think this is the correct syntax to invoke the hook, so i'm surprised it is not throwing a syntax error?

R4hul04: No it doesn't throw any syntax errors. I will add the above syntax. Also I don't think of any case where  `babelfish_view_def` will not be defined for tsql views.

contrib/babelfishpg_tsql/src/dbcmds.c
R4hul04: changed the BBF database initialization query from `ALTER VIEW dummy.sysdatabases OWNER TO dummy; ` to 
`ALTER TABLE dummy.sysdatabases OWNER TO dummy; `. Since PG can use `ALTER TABLE` cmd as well to change ownership of view

contrib/babelfishpg_tsql/src/dbcmds.c
```diff
@@ -149,7 +149,7 @@ gen_createdb_subcmds(const char *dbname, const char *owner)
 
 	/* create sysdatabases under current DB's DBO schema */
 	appendStringInfo(&query, "CREATE VIEW dummy.sysdatabases AS SELECT * FROM sys.sysdatabases; ");
-	appendStringInfo(&query, "ALTER VIEW dummy.sysdatabases OWNER TO dummy; ");
```
forestkeeper, line 152: What's the purpose for this line ?

R4hul04: If we use ALTER VIEW, it will raise syntax error near OWNER while doing initbbf because of different ALTER VIEW behavior between PG and tsql. So we use ALTER TABLE instead to avoid that, as PG can use ALTER TABLE as well to change the ownership of view. ([Postgresql Alter Table](https://www.postgresql.org/docs/8.1/sql-altertable.html#:~:text=This%20form%20changes%20the%20owner%20of%20the%20table%2C%20sequence%2C%20or%20view%20to%20the%20specified%20user.))

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -2766,6 +2770,117 @@ bbf_ProcessUtility(PlannedStmt *pstmt,
...
+					Oid oldViewOid;
+					Acl *oldViewAcl = NULL;
+					bool isCompleteQuery = (context != PROCESS_UTILITY_SUBCOMMAND);
+					bool needCleanup;
+			
+					if (!IS_TDS_CLIENT())
+					{
+							ereport(ERROR,
+									(errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
+										errmsg("TSQL ALTER VIEW is not supported from PostgreSQL endpoint.")));
```
timchang514, line 2813: can we add a test case for this error message? 

use `.mix` files to test psql endpoint.

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -2766,6 +2770,117 @@ bbf_ProcessUtility(PlannedStmt *pstmt,
...
+
+						/* Get the old view's OID and verify it exists */
+						oldViewOid = RangeVarGetRelid(stmt->view, AccessExclusiveLock, true);
+						
+						/* If the view does not exist, check if the stmt is CREATE OR ALTER VIEW / ALTER VIEW */
+						if (!OidIsValid(oldViewOid))
+						{
+							if(stmt->replace)	/* we have set replace to false for CREATE OR ALTER VIEW to avoid view does not exist error */
+							{
+								ereport(ERROR,
```
timchang514, line 2837: i don't see a test case for this error message

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -2766,6 +2770,117 @@ bbf_ProcessUtility(PlannedStmt *pstmt,
...
+							CommandCounterIncrement();
+			
+							/* Create new view */
+							stmt->replace = true; 
+							address = DefineView(stmt, queryString, pstmt->stmt_location, pstmt->stmt_len);
+							CommandCounterIncrement();
+							
+							/* Store the view definition in babelfish_view_def */
+							if(store_view_definition_hook)
+								store_view_definition_hook(queryString, address);
```
timchang514, line 2867: store_view_definition_hook is already called in `DefineView`, is this needed?

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -4814,6 +4929,73 @@ pltsql_proc_get_oid_proname_proacl(AlterFunctionStmt *stmt, ParseState *pstate,
 	*isSameFunc = OidIsValid(funcOid);
 }
 
+/*
+ * Get the acl of view from pg_class given its oid
+ */
+static Acl *
+get_old_view_acl(Oid oldViewOid)
+	{
```
timchang514: nit: whitespace

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -4814,6 +4929,73 @@ pltsql_proc_get_oid_proname_proacl(AlterFunctionStmt *stmt, ParseState *pstate,
...
+
+	return oldViewAcl;
+	}
+
+/*
+ * Update the acl info of view in pg_class given its oid and acl info
+ */
+static void
+pg_class_update_acl(Oid newViewOid, Acl *oldViewAcl)
+	{
```
timchang514: nit: whitespace

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -4814,6 +4929,73 @@ pltsql_proc_get_oid_proname_proacl(AlterFunctionStmt *stmt, ParseState *pstate,
...
+	ReleaseSysCache(tuple);
+
+	return oldViewAcl;
+	}
+
+/*
+ * Update the acl info of view in pg_class given its oid and acl info
+ */
+static void
+pg_class_update_acl(Oid newViewOid, Acl *oldViewAcl)
```
timchang514, line 4977: let's also add a test that verifies that the acl is properly updated when we do ALTER VIEW.

contrib/babelfishpg_tsql/src/tsqlIface.cpp
```diff
@@ -4376,6 +4377,25 @@ storeOriginalQueryForBatchLevelStatement(TSqlParser::Batch_level_statementContex
 		originalQueryCopy.replace(startIndex, endIndex - startIndex, "CREATE");
 		return pstrdup(originalQueryCopy.c_str());
 	}
+	else if (ctx->create_or_alter_view() && ctx->create_or_alter_view()->ALTER())
```
timchang514, line 4381: what is the purpose of these changes?

contrib/babelfishpg_tsql/src/pl_handler.c
R4hul04: Yes, because in DefineView condition is `if (store_view_definition_hook && !(stmt->replace))`, !(stmt->replace) will always be false for Alter View, so we need to call `store_view_definition_hook` in our function to record the definition

contrib/babelfishpg_tsql/src/tsqlIface.cpp
R4hul04: After the alter operation is complete, the definition that gets stored in catalogs is something like `Alter view v1 as select 1;` instead of `Create view v1 as select 1;`. This change allows storing correct definitions after `ALTER/CREATE OR ALTER VIEW` operations. Appropriate tests are added in the test files. SQL server does the same, it stores the definitions as `CREATE VIEW` instead of `ALTER VIEW`

contrib/babelfishpg_tsql/src/dbcmds.c
forestkeeper: Pls add a comment into the code for this behavior and the purpose for this change.

contrib/babelfishpg_tsql/src/pl_handler.c
forestkeeper: Agree, we should add test case for this.

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -2766,6 +2770,117 @@ bbf_ProcessUtility(PlannedStmt *pstmt,
...
+					}
+
+					needCleanup = isCompleteQuery && EventTriggerBeginCompleteQuery();
+			
+					PG_TRY();
+					{
+						StartTransactionCommand();
+						
+						if (isCompleteQuery)
+							EventTriggerDDLCommandStart(parsetree);
```
forestkeeper, line 2827: Which is the test case for this ?

contrib/babelfishpg_tsql/src/tsqlIface.cpp
forestkeeper: Pls add the comments into the code as well.

contrib/babelfishpg_tsql/src/pl_handler.c
R4hul04: Without this code, DDL event triggers won't fire for ALTER VIEW operations. Currently we do not support DDL triggers for T-SQL, but tested it out from from PSQL endpoint. This code can be used later, once we support DDL triggers in T-SQL. Added comment in the code.

contrib/babelfishpg_tsql/src/tsqlIface.cpp
R4hul04: Added comments in the code

contrib/babelfishpg_tsql/src/dbcmds.c
R4hul04: Added comments in the code

contrib/babelfishpg_tsql/src/pl_handler.c
R4hul04: Added this test case as well as tests for ACL changes.

contrib/babelfishpg_tsql/src/pl_handler.c
R4hul04: Added the test case, thanks for noticing 👍

contrib/babelfishpg_tsql/src/pl_handler.c
R4hul04: Added `.mix` test file to test ACL changes.

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -2766,6 +2770,121 @@ bbf_ProcessUtility(PlannedStmt *pstmt,
...
+							}
+							/* View doesn't exist - create it */
+							address = DefineView(stmt, queryString, pstmt->stmt_location, pstmt->stmt_len);
+							CommandCounterIncrement();
+						}
+						/* View exists */
+						else
+						{
+							/* Save ACL before dropping the view */
+							oldViewAcl = get_old_view_acl(oldViewOid);
```
timchang514, line 2850: To me it seems like get_old_view_acl can potentially return NULL. There is even a check for it in pg_class_update_acl. Then it would cause a crash once you hit `pfree(oldViewAcl)`.

So either add a check that `get_old_view_acl` cannot return NULL - throw an error (then the case in pg_class_update_acl is not needed), or ensure that pfree isn't called on NULL.

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -2766,6 +2770,121 @@ bbf_ProcessUtility(PlannedStmt *pstmt,
...
+							CacheInvalidateRelcacheByRelid(oldViewOid);
+			
+							/* Drop the old view */
+							originalView.objectId = oldViewOid;
+							originalView.classId = RelationRelationId;
+							originalView.objectSubId = 0;
+							performDeletion(&originalView, DROP_RESTRICT, 0);
```
timchang514, line 2857: DROP_RESTRICT means this will fail if there's any objects dependent on the view. Could we test this - create an object dependent on a view and attempt to alter the underlying view?

forestkeeper (dismissed):

contrib/babelfishpg_tsql/src/pl_handler.c
R4hul04: Added to test case

contrib/babelfishpg_tsql/src/dbcmds.c
tanscorpio7: Please add test cases for postgres endpoint. ALTER VIEW ... should continue to work as before for postgres endpoint.

contrib/babelfishpg_tsql/src/pl_handler.c
R4hul04: Older version views may have NULL ACL, added check to pfree(oldViewAcl) only if oldViewAcl is non null

contrib/babelfishpg_tsql/src/dbcmds.c
R4hul04: Added test cases to check psql ALTER VIEW

timchang514 (approved):

forestkeeper (approved):

PR #3571, opened 2025-03-12
Reseed identity for new table created using select into and only inherit nullability for simple queries

tanscorpio7:
### Description

This commit fixes a couple of thing.
1. Reseed identity sequence for new table created using
SELECT INTO statement. For reseeding we will run a post
Select Into internal sub command.
```
SELECT setval(seq, SELECT MAX(identity_column) FROM new_table);
```
If identity is decreasing then we will use MIN instead of MAX.

2. Only carry over nullability when the query has single
table reference in FROM clause. Table could be either
permanent or temporary.

### Issues Resolved

[BABEL-5661]

### Sign Off

### Check List

### DIFF
--- a/contrib/babelfishpg_tsql/src/hooks.c
+++ b/contrib/babelfishpg_tsql/src/hooks.c
@@ -904,7 +904,7 @@ pltsql_ExecFuncProc_AclCheck(Oid funcid)
 	Oid userid = GetUserId();
 
 	/* In TDS client, the permissions might need to be checked against session user. */
-	if (IS_TDS_CLIENT() && !InSecurityRestrictedOperatio())
+	if (IS_TDS_CLIENT() && !InSecurityRestrictedOperation())
 	{
 		Oid schema_id = get_func_namespace(funcid);
 
--- a/contrib/babelfishpg_tsql/src/hooks.c
+++ b/contrib/babelfishpg_tsql/src/hooks.c
@@ -903,7 +903,12 @@ pltsql_ExecFuncProc_AclCheck(Oid funcid)
 {
 	Oid userid = GetUserId();
 
-	/* In TDS client, the permissions might need to be checked against session user. */
+	/*
+	 * In TDS client, the permissions might need to be checked against session user.
+	 * Do not do this when in SECURITY_RESTRICTED_OPERATION because in this context
+	 * we temprorarily switch to a different user to execute some internal subcommands.
+	 * For example see reseed_identity_post_select_into
+	 */
 	if (IS_TDS_CLIENT() && !InSecurityRestrictedOperation())
 	{
 		Oid schema_id = get_func_namespace(funcid);
@@ -958,6 +963,7 @@ pltsql_ExecutorStart(QueryDesc *queryDesc, int eflags)
 	/*
 	 * In TDS client, the RTE permissions might need to be checked against login mapped to given checkAsUser,
 	 * if it is valid, otherwise permissions are checked against session user (current login).
+	 * For SECURITY_RESTRICTED_OPERATION handling see comments in pltsql_ExecFuncProc_AclCheck
 	 */
 	if (IS_TDS_CLIENT() && queryDesc->plannedstmt != NULL && !InSecurityRestrictedOperation())
 	{
--- a/contrib/babelfishpg_tsql/src/hooks.c
+++ b/contrib/babelfishpg_tsql/src/hooks.c
@@ -904,7 +904,7 @@ pltsql_ExecFuncProc_AclCheck(Oid funcid)
 	Oid userid = GetUserId();
 
 	/* In TDS client, the permissions might need to be checked against session user. */
-	if (IS_TDS_CLIENT() && !InSecurityRestrictedOperation())
+	if (IS_TDS_CLIENT() && !InSecurityRestrictedOperatio())
 	{
 		Oid schema_id = get_func_namespace(funcid);
 
--- a/contrib/babelfishpg_tsql/src/hooks.c
+++ b/contrib/babelfishpg_tsql/src/hooks.c
@@ -904,7 +904,7 @@ pltsql_ExecFuncProc_AclCheck(Oid funcid)
 	Oid userid = GetUserId();
 
 	/* In TDS client, the permissions might need to be checked against session user. */
-	if (IS_TDS_CLIENT())
+	if (IS_TDS_CLIENT() && !InSecurityRestrictedOperatio())
 	{
 		Oid schema_id = get_func_namespace(funcid);
 
@@ -959,7 +959,7 @@ pltsql_ExecutorStart(QueryDesc *queryDesc, int eflags)
 	 * In TDS client, the RTE permissions might need to be checked against login mapped to given checkAsUser,
 	 * if it is valid, otherwise permissions are checked against session user (current login).
 	 */
-	if (IS_TDS_CLIENT() && queryDesc->plannedstmt != NULL)
+	if (IS_TDS_CLIENT() && queryDesc->plannedstmt != NULL && !InSecurityRestrictedOperation())
 	{
 		ListCell	*lc;
 
--- a/contrib/babelfishpg_tsql/src/pl_handler.c
+++ b/contrib/babelfishpg_tsql/src/pl_handler.c
@@ -6765,32 +6765,24 @@ transformSelectIntoStmt(CreateTableAsStmt *stmt)
 		Index identity_ressortgroupref = 0;
 		List *modifiedTargetList = NIL;
 
-		bool persist_identity = true;
+		/*
+		 * currently we only carry over identity and nullability for
+		 * simple query with a single relation in from expression
+		 */
+		bool persist_identity_and_nullability = false;
 
 		/*
 		 * In T-SQL, identity property of column in destination
 		 * table is not persisted if the SELECT ... INTO statement
 		 * does not contain a single base relation
 		 */
-		if (q->jointree && q->jointree->fromlist != NIL)
+		if (q->jointree && q->jointree->fromlist != NIL && list_length(q->jointree->fromlist) == 1)
 		{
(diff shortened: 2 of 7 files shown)


contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -904,7 +904,7 @@ pltsql_ExecFuncProc_AclCheck(Oid funcid)
 	Oid userid = GetUserId();
 
 	/* In TDS client, the permissions might need to be checked against session user. */
-	if (IS_TDS_CLIENT())
```
Deepesh125, line 907: Please add more explanation that it could be the case only if we are reseeding the identity during SELECT...INTO operation. You can also add reference link to reseed_identity_post_select_into

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -6745,57 +6745,33 @@ get_identity_into_args(Node *node)
...
-		 * In T-SQL, identity property of column in destination
-		 * table is not persisted if the SELECT ... INTO statement
-		 * does not contain a single base relation
+		 * currently we only inherit identity and nullability for
+		 * simple query i.e. a single relation in from expression
 		 */
-		if (q->jointree && q->jointree->fromlist != NIL)
+		bool persist_identity_and_nullability = false;
+
+		if (q->jointree && q->jointree->fromlist != NIL && list_length(q->jointree->fromlist) == 1)
```
Deepesh125, line 6779: will we (in future) support more cases around this?

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -6745,57 +6745,33 @@ get_identity_into_args(Node *node)
...
-				{
-					RangeTblRef		*rtr = (RangeTblRef *) n_from;
-					RangeTblEntry	*rte = rt_fetch(rtr->rtindex, q->rtable);
-
-					/* Do not persist identity if not an ordinary relation */
-					if (rte->rtekind != RTE_RELATION)
-						persist_identity = false;
-				}
+				RangeTblEntry	*rte = rt_fetch(rtr->rtindex, q->rtable);
+				persist_identity_and_nullability = rte->rtekind == RTE_RELATION;
```
Deepesh125: what if I have single temp table in from clause? I think we can still support that case?

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -6880,8 +6854,10 @@ transformSelectIntoStmt(CreateTableAsStmt *stmt)
 				lcmd->missing_ok = false;
 				lcmd->def = (Node *)def;
 				altstmt->cmds = lappend(altstmt->cmds, lcmd);
+
+				tle->resjunk = true;
```
Deepesh125: why are we doing it forcefully?

contrib/babelfishpg_tsql/src/pltsql_identity.c
```diff
@@ -470,3 +472,100 @@ pltsql_revert_last_scope_identity(int nest_level)
...
+		seq_name = quote_identifier(get_rel_name(seq_relid));
+		identity_colname = quote_identifier(NameStr(attr->attname));
+
+		ReleaseSysCache(pgstuple);
+		break;
+	}
+
+	RelationClose(rel);
+
+	if (!identity_colname || !OidIsValid(relowner))
```
Deepesh125: in which can this could be true?

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -6929,43 +6905,9 @@ transformSelectIntoStmt(CreateTableAsStmt *stmt)
...
-					{
-						SortGroupClause *sortcl = (SortGroupClause *)sortnode;
-						if (sortcl->tleSortGroupRef != identity_ressortgroupref)
-							modifiedSortClause = lappend(modifiedSortClause, sortcl);
-					}
-				}
-				q->sortClause = modifiedSortClause;
-			}
-
-			if (q->distinctClause && list_length(q->distinctClause) > (identity_ressortgroupref - 1))
```
Deepesh125: Please explain why dont we need this change

contrib/babelfishpg_tsql/src/pltsql_identity.c
```diff
@@ -470,3 +472,100 @@ pltsql_revert_last_scope_identity(int nest_level)
...
+		return;
+
+	initStringInfo(&querybuf);
+
+	appendStringInfo(&querybuf,
+					 "SELECT pg_catalog.setval('%s.%s', (SELECT pg_catalog.%s(%s) FROM %s.%s))",
```
Deepesh125, line 533: is there any better way of doing this? May be directly calling some backend function to retrieve maxvalue and we can then call setConfig?

test/JDBC/expected/select_into-identity-notnull.out
```diff
@@ -418,3 +414,316 @@ go
...
+-- INCREASING IDENTITY
+-- id1 represent expected identity value
+CREATE TABLE babel_5661_source (id INT IDENTITY(1,1), id1 INT)
+GO
+INSERT INTO babel_5661_source VALUES (1), (2), (3), (4), (5), (6)
+GO
+~~ROW COUNT: 6~~
+
+
+SELECT * INTO #babel_5661_new1 FROM babel_5661_source
```
Deepesh125, line 440: can you add some similar tests where source table is temp table

Deepesh125 (changes requested):

contrib/babelfishpg_tsql/src/pl_handler.c
tanscorpio7: Yes, we could

contrib/babelfishpg_tsql/src/pl_handler.c
tanscorpio7: Yes, in fact this check should also be checking fro relkind, which will actually block views.

contrib/babelfishpg_tsql/src/pl_handler.c
tanscorpio7: This target entry is for identity function. Earlier we were doing
1. Add alter table commands for IDENTITY
2. Skip the identity function tle in new tle list

Instead of doing all this we can simply mark the tle as resjunk and move along.

contrib/babelfishpg_tsql/src/pltsql_identity.c
tanscorpio7: When new table does not have any identity.

tanscorpio7: and checking for `!OidIsValid(relowner)` just for safety.

test/JDBC/expected/select_into-identity-notnull.out
tanscorpio7: Done

contrib/babelfishpg_tsql/src/pltsql_identity.c
tanscorpio7: Could not find any other alternative for this.

contrib/babelfishpg_tsql/src/pl_handler.c
tanscorpio7: Reverted this, I was refactoring a few things but will keep it out of this PR for simplicity.

contrib/babelfishpg_tsql/src/pl_handler.c
tanscorpio7: Reverted this.

contrib/babelfishpg_tsql/src/pl_handler.c
tanscorpio7: Now we will allow permanent and temporary tables only.

contrib/babelfishpg_tsql/src/hooks.c
tanscorpio7: Done

contrib/babelfishpg_tsql/src/pltsql_identity.c
Deepesh125: if we are checking !OidIsValid(relowner) just for safety then it should throw an error instead of just returning silently

contrib/babelfishpg_tsql/src/pltsql_identity.c
```diff
@@ -470,3 +472,100 @@ pltsql_revert_last_scope_identity(int nest_level)
...
+		int 	rc;
+
+		sql_dialect = SQL_DIALECT_PG;
+
+		if ((rc = SPI_connect()) < 0)
+			elog(ERROR, "SPI_connect() failed in SELECT INTO "
+					"with return code %d", rc);
+
+		/* The new table could already be transferred to schema owner and current user may not have privs for it */
+		SetUserIdAndSecContext(relowner, save_sec_context | SECURITY_RESTRICTED_OPERATION);
```
HarshLunagariya, line 555: I agree that cross-db `pltsql_ExecFuncProc_AclCheck` should be executed in case of `SECURITY_RESTRICTED_OPERATION`. But is it relevant in this case? Here I think mostly all the function/proc included are from pg_catalog , isn't it or Am I missing something?

tanscorpio7: We are executing a query against the newly created table for which current user may not have permission for.

HarshLunagariya: Oh, We have added this to ExecutorStart, Not just for FuncProc. Essentially we just want to execute this command in PG dialect and avoid T-SQL specific logic. Am I interpreting it correctly?

contrib/babelfishpg_tsql/src/pltsql_identity.c
tanscorpio7: Done

Deepesh125 (approved):

PR #3573, opened 2025-03-13
Update logic for LIKE operator for Case Insensitive collation

ahmed-shameem:
### Description

-- CASE 1
```
```

-- CASE 2
```
```

```
expression LIKE pattern
```

```
expression = pattern
```

```
```

Currently we have observed that we do not comply with the transformation above. Rather than COLLATE CS_AS, we observe COLLATE “default” in query plan. Consider the following query:

```
select a from varchar_tst where a collate chinese_prc_ci_as LIKE 'ab%'
```

-- query plan
```
Filter: (((a)::text ~~* 'ab%'::text COLLATE "default") 
AND ((a)::text >= 'ab'::text COLLATE "default") AND ((a)::text < 'ab?'::text))
```

Even though we observe COLLATE "default" in the query plan, it picks up the correct collation oid during execution. 
However, if we update the collation oid in typeTypeCollation function to pickup database collation oid for string literals in parse tree, we may observe upgrade/restore failure.

We use this task as first step to modify and incorporate improvements for such cases.

### Issues Resolved

Task: BABEL-5608

* **Use case based -**

* **Boundary conditions -**

* **Arbitrary inputs -**

* **Negative test cases -**

* **Performance tests -**

* **Tooling impact -**

* **Client tests -**

### Check List

### DIFF
--- a/test/JDBC/upgrade/16_6/schedule
+++ b/test/JDBC/upgrade/16_6/schedule
@@ -590,3 +590,4 @@ test_conv_money_to_varchar
 fixeddecimal_modulo
 test_conv_float_to_varchar_char_before_17_4
 BABEL_OBJECT_RESOLUTION_IN_ROUTINES
+test_constraint_like-before-17-5-or-16-9
--- a/contrib/babelfishpg_tsql/src/collation.c
+++ b/contrib/babelfishpg_tsql/src/collation.c
@@ -305,6 +305,16 @@ transform_funcexpr(Node *node)
 	return node;
 }
 
+static CollateExpr*
+create_collate_expr(Node *arg, Oid collid)
+{
+	CollateExpr *expr = makeNode(CollateExpr);
+	expr->arg = (Expr *) arg;
+	expr->collOid = collid;
+	expr->location = -1;
+	return expr;
+}
+
 /*
  * If the node is OpExpr and the colaltion is ci_as, then
  * transform the LIKE OpExpr to ILIKE OpExpr:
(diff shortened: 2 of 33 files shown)


test/JDBC/upgrade/16_6/schedule
```diff
@@ -590,3 +590,4 @@ test_conv_money_to_varchar
 fixeddecimal_modulo
 test_conv_float_to_varchar_char_before_17_4
 BABEL_OBJECT_RESOLUTION_IN_ROUTINES
+test_constraint_like-before-17-5-or-16-9
```
Deepesh125, line 593: should we also test some 15.x to 16.latest path?

Deepesh125 (changes requested):

test/JDBC/upgrade/16_6/schedule
ahmed-shameem: For major version upgrade we are already testing from 16 to 17. We do not plan to cherry-pick the change into 15X.

ahmed-shameem: We know that the definition does not change for the objects after m/MVU. 
Moreover, down the line we have improved upon like operator and etc. We will have multiple copies of same test files and multiple output files with minor diff for a single test which includes multiple before after file. For example, absence of [BABEL-5424] in 15X creates a new diff for few cases. In that case we need create separate before after file.

Deepesh125 (approved):

PR #3582, opened 2025-03-18
Use Index Scan for EXACT and PREFIX pattern matches for LIKE 

ahmed-shameem:
### Description

-- CASE 1
```
```

-- CASE 2
```
```

```
expression LIKE pattern
```

```
expression = pattern
```

```
```

Currently we have observed that we do not comply with the transformation above for CS collations. Rather than COLLATE CS_AS, we observe COLLATE “default” in query plan. Consider the following query:

```
select a from varchar_tst where a collate chinese_prc_cs_as LIKE 'ab%'
```

-- query plan
```
Filter: (((a)::text ~~ 'ab%'::text COLLATE "default") 
```

This commit handles case sensitive collation in babelfish side as well with the above mentioned optimisation for EXACT and PREFIX pattern match.

We have the optimisation for EXACT and PREFIX pattern match so that =, > and <= operators can use Index when the left operand has an Index on it. Till now babelfish was unable to use Index Scan efficiently. The reason for this is because babelfish was unable to fulfil the following condition to use Index:
```
must contain an operator which is in the index's operator family for this column
 ```

As LIKE operator is defined for TEXT datatype in PG, the operands of LIKE in babelfish were getting type casted into TEXT,  hence whenever we were trying to find  =, > and <= operators, we were always obtaining the operators best compatible with TEXT. Now these operators are NOT a member of operator family like varchar_ops, bpchar_ops in babelfish. 
For accent insensitive collations, as we put extra remove_accents_internal function call on operands, to use Index scan for this case user needs to create index on remove_accents_internal (functional index) from PG endpoint.

This commit fixes that issue by finding the best compatible  =, > and <= operators for the original operands. This operator should be in expected operator class and hence we should be able to use Index properly. 
We also handle the CHECK CONSTRAINT here to NOT use any optimisation as that does not add any value and rather adds additional overhead. Similar approach is used by native PG. Moreover, we have observed that such optimisation in check constraint adds additonal meta-data in the constraint definition after restore and upgrade which basically changes the constraint definition pre and post upgrade/restore. 

### Issues Resolved

Task: BABEL-5660, BABEL-5708

* **Use case based -**

* **Boundary conditions -**

* **Arbitrary inputs -**

* **Negative test cases -**

* **Performance tests -**

* **Tooling impact -**

* **Client tests -**

### Check List

### CODE REVIEW

**Review by Imran-imtiaz48 — COMMENTED** [review]
Summary: This update performs a comprehensive evaluation of the LIKE and NOT LIKE clause behavior using Babelfish for PostgreSQL. The test cases cover various scenarios including pattern matching, Unicode support, and edge cases with escape characters.
________________________________________
What's Good
•	Diverse Test Coverage: Excellent coverage for patterns like 'jones', 'Jon%', 'jone_', 'ab%', 'äb%', '\%ones', and others. This ensures robust verification of pattern matching in Babelfish.
•	Unicode Handling: Special characters such as 'äḃĆ_' were tested to verify Unicode collation and LIKE clause behavior under bbf_unicode_cp1_cs_as collation.
•	Escaped Characters: Validation for edge cases like '\%ones' and '\_ones' showcases attention to escape handling.
•	Consistent Use of Indexes: Most queries make good use of the Bitmap Index Scan, optimizing performance across a range of LIKE filters.
________________________________________
Performance Insights
•	T-SQL batch parsing times remained low and consistent (0.12–0.20 ms range), confirming low overhead.
•	Seq Scan usage on testing5 indicates absence of indexes on that table – might consider indexing if performance tuning is needed.
•	Indexed scans on testing4 suggest good database design for pattern search efficiency.
________________________________________
Suggestions
•	Consider adding test cases for:
o	Case-sensitive vs case-insensitive LIKE (COLLATE variations).
o	Wildcard at different positions: %abc, abc%, %abc%, _abc, etc.
o	LIKE '[a-z]%' patterns if Babelfish supports regex-style evaluation (just for completeness).
•	It may be helpful to clarify any Locale or Collation settings impact, especially with Unicode strings, in your README or test documentation.
________________________________________
Final Thoughts
Great work validating pattern-matching behavior using Babelfish’s execution plans. This level of test coverage is invaluable when assessing compatibility and performance in migration scenarios from SQL Server to PostgreSQL via Babelfish.

File: contrib/babelfishpg_tsql/src/collation.c [hunk 1]
```diff
@@ -1026,50 +1022,34 @@ transform_likenode(Node *node)
...
 		{
-			if (supported_collation_for_db_and_like(coll_info_of_inputcollid.code_page))
-				return transform_from_cs_ai_for_likenode(node, op, like_entry, coll_info_of_inputcollid);
-			else
-				ereport(ERROR,
-				(errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
-				 errmsg("LIKE operator is not supported for \"%s\"", coll_info_of_inputcollid.collname)));
-		}
+			if (coll_info_of_inputcollid.collateflags == 0x000d || coll_info_of_inputcollid.collateflags == 0x000c)	/* AS */
+				return optimise_likenode(node, op, like_entry, coll_info_of_inputcollid, is_constraint);
```
Deepesh125, line 1032: why CS_AS also?

contrib/babelfishpg_tsql/src/collation.c
```diff
@@ -893,67 +930,26 @@ convert_node_to_funcexpr_for_like(Node *node)
...
+transform_likenode_for_AI(OpExpr *op)
 {
 	Node		*leftop = (Node *) linitial(op->args);
 	Node		*rightop = (Node *) lsecond(op->args);
 
 	linitial(op->args) = coerce_to_target_type(NULL,
-												convert_node_to_funcexpr_for_like(leftop),
-												get_sys_varcharoid(),
+												convert_node_to_funcexpr_for_like(leftop, op->inputcollid),
+												get_sys_nvarcharoid(),
```
Deepesh125, line 944: why nvarchar for leftop? I think remove_accents_internal_using_cache returns nvarchar so we should coerce result of remove_accents_internal_using_cache from nvarchar to original type of leftop unless I am missing something

contrib/babelfishpg_tsql/src/collation.c
```diff
@@ -798,12 +829,18 @@ Datum remove_accents_internal(PG_FUNCTION_ARGS)
...
 
 static Node *
-convert_node_to_funcexpr_for_like(Node *node)
+convert_node_to_funcexpr_for_like(Node *node, Oid inputcollid)
 {
 	FuncExpr *newFuncExpr = makeNode(FuncExpr);
 	Node *new_node;
 	newFuncExpr->funcid = remove_accents_internal_oid;
-	newFuncExpr->funcresulttype = get_sys_varcharoid();
+	newFuncExpr->funcresulttype = get_sys_nvarcharoid();
```
Deepesh125, line 841: please explain this change

contrib/babelfishpg_tsql/src/collation.c
```diff
@@ -893,67 +930,26 @@ convert_node_to_funcexpr_for_like(Node *node)
...
+												get_sys_nvarcharoid(),
 												exprType(leftop), -1,
 												COERCION_EXPLICIT,
 												COERCE_EXPLICIT_CAST,
 												-1);
 	lsecond(op->args) = coerce_to_target_type(NULL,
-												convert_node_to_funcexpr_for_like(rightop),
-												get_sys_varcharoid(),
+												convert_node_to_funcexpr_for_like(rightop, op->inputcollid),
+												get_sys_nvarcharoid(),
```
Deepesh125, line 951: same here

test/JDBC/expected/babel_collection.out
```diff
@@ -183,14 +183,15 @@ GO
 ~~START~~
 text
 Query Text: select c1 from testing4 where c1 LIKE 'jones'
-Bitmap Heap Scan on testing4  (cost=11.40..29.53 rows=3 width=32)
-  Filter: ((c1)::text = 'jones'::text)
-  ->  Bitmap Index Scan on c1_idxtesting49a168d73f3ba5aacdfd495b931b8d187  (cost=0.00..11.40 rows=650 width=0)
+Bitmap Heap Scan on testing4  (cost=4.17..11.28 rows=3 width=32)
+  Recheck Cond: (c1 = 'jones'::"varchar")
+  ->  Bitmap Index Scan on c1_idxtesting49a168d73f3ba5aacdfd495b931b8d187  (cost=0.00..4.17 rows=3 width=0)
+        Index Cond: (c1 = 'jones'::"varchar")
```
Deepesh125, line 189: I am not getting this, why isn't it picking simple seq scan or index scan and is picking Bitmap heap scan instead?

Deepesh125: Can we please check postgres if it shows same behaviour?

contrib/babelfishpg_tsql/src/collation.c
```diff
@@ -393,22 +397,34 @@ transform_from_ci_as_for_likenode(Node *node, OpExpr *op, like_ilike_info_t like
 			rightop = (Node *) lsecond(op->args);
 		}
 	}
+	op->inputcollid = tsql_get_oid_from_collidx(collidx_of_cs_as);
 
-	/* no constant prefix found in pattern, or pattern is not constant */
+	/* 
+	 * no constant prefix found in pattern, or pattern is not constant 
+	 * OR if it is for CHECK CONSTRAINT, we do NOT need any optimisation
+	 * for it. Rather it will add extra overhead, moreover vanilla PG
```
Deepesh125: Postgres*

contrib/babelfishpg_tsql/src/collation.c
```diff
@@ -419,6 +435,22 @@ transform_from_ci_as_for_likenode(Node *node, OpExpr *op, like_ilike_info_t like
...
+	 * Similarly, update the type of prefix to have appropriate datatypes of operands
+	 * 
+	 * Optimiser will remove Relabel Node during Index scan, see match_index_to_operand
+	 */
+	if (IsA(leftop, RelabelType))
+	{
+		RelabelType	*relabel = (RelabelType *) leftop;
+		leftop = copyObject((Node*) relabel->arg);
+		prefix->consttype = rtypeId = ltypeId = exprType(leftop);
+	}
```
Deepesh125, line 456: can we move this above where we are handing right operand for Relabel node?

contrib/babelfishpg_tsql/src/collation.c
```diff
@@ -316,7 +319,7 @@ create_collate_expr(Node *arg, Oid collid)
 }
 
 /*
- * If the node is OpExpr and the colaltion is ci_as, then
+ * If the node is OpExpr and the colaltion is ci_as/ci_ai , then
```
Deepesh125, line 322: also add comment if collation is cs_as

contrib/babelfishpg_tsql/src/collation.c
```diff
@@ -372,12 +375,13 @@ transform_from_ci_as_for_likenode(Node *node, OpExpr *op, like_ilike_info_t like
...
-
+	/* Change the opno and oprfuncid to ILIKE if CI collation */
+	if (coll_info_of_inputcollid.collateflags == 0x000f || coll_info_of_inputcollid.collateflags == 0x000d) /* CI */
+	{
+		op->opno = like_entry.ilike_oid;
+		op->opfuncid = like_entry.ilike_opfuncid;
+	}
+	
 	/* 
 	 * This is needed to process CI_AI for Const nodes
```
Deepesh125, line 390: If condition doesnt justify this statement. Either we need to update if condition or improve comment

contrib/babelfishpg_tsql/src/collation.c
```diff
@@ -482,7 +514,7 @@ transform_from_ci_as_for_likenode(Node *node, OpExpr *op, like_ilike_info_t like
 											oprfuncid(optup));
 		ReleaseSysCache(optup);
 		/* construct pattern||E'\uFFFF' */
-		highest_sort_key = makeConst(TEXTOID, -1, InvalidOid, -1,
```
Deepesh125, line 485: Line 505/506 should be converted to error condition. It is always expected that we will find all the required operators - 
```
if (optup == (Operator) NULL)
			return node;
```

Deepesh125: same for 522/523 line
```		
if (optup == (Operator) NULL)
			return node;

```

Deepesh125 (changes requested):

test/JDBC/expected/db_collation/babel_collection.out
```diff
@@ -184,13 +184,13 @@ GO
 text
 Query Text: select c1 from testing4 where c1 LIKE 'jones'
 Bitmap Heap Scan on testing4  (cost=11.40..31.15 rows=3 width=32)
-  Filter: ((remove_accents_internal((c1)::text))::text = 'jones'::text)
+  Filter: (remove_accents_internal((c1)::text) = 'jones'::nvarchar)
```
Deepesh125, line 187: shouldn't we get COLLATE on right operand? Seems like it should be just by looking at look

test/JDBC/expected/non_default_server_collation/chinese_prc_ci_as/babel_collection.out
```diff
@@ -361,14 +372,16 @@ GO
...
 text
 Query Text: select c1 from testing4 where c1 LIKE '\%ones'
-Bitmap Heap Scan on testing4  (cost=11.40..32.78 rows=1 width=32)
-  Filter: (((c1)::text ~~* '\%ones'::text COLLATE chinese_prc_cs_as) AND ((c1)::text >= '\'::text) AND ((c1)::text < '\?'::text))
-  ->  Bitmap Index Scan on c1_idxtesting49a168d73f3ba5aacdfd495b931b8d187  (cost=0.00..11.40 rows=650 width=0)
+Bitmap Heap Scan on testing4  (cost=4.18..11.30 rows=1 width=32)
+  Recheck Cond: ((c1 >= '\'::"varchar") AND (c1 < '\?'::"varchar"))
+  Filter: ((c1)::text ~~* '\%ones'::text COLLATE chinese_prc_cs_as)
+  ->  Bitmap Index Scan on c1_idxtesting49a168d73f3ba5aacdfd495b931b8d187  (cost=0.00..4.18 rows=3 width=0)
+        Index Cond: ((c1 >= '\'::"varchar") AND (c1 < '\?'::"varchar"))
```
Deepesh125, line 379: why is transformation applied here? because pattern is not const prefix

test/JDBC/expected/non_default_server_collation/chinese_prc_ci_as/test_constraint_like-vu-prepare.out
```diff
@@ -263,15 +263,15 @@ SELECT pg_get_constraintdef(oid) FROM pg_constraint WHERE conrelid = 'master_dbo
 GO
 ~~START~~
 text
-CHECK (((babel_5608_vu_prepare_t4_1_col1)::text = ('ABC'::text COLLATE sys.chinese_prc_ci_as)))
+CHECK ((((babel_5608_vu_prepare_t4_1_col1)::text COLLATE sys.chinese_prc_cs_as) ~~* ('ABC'::text COLLATE sys.chinese_prc_cs_as)))
```
Deepesh125, line 266: Does postgres converts LIKE to = for check constraint if it is exact match pattern?

test/JDBC/upgrade/17_5/schedule
```diff
@@ -613,9 +613,10 @@ sys-login-property
 sys-fn-varbintohexsubstring
 BABEL_OBJECT_RESOLUTION_IN_ROUTINES
 alter-mvu-test
-test_constraint_like
+test_constraint_like-17-5-or-16-9
```
Deepesh125, line 615: this is bit unconventional. We should rename existing test to before_17_5_or_16_9 instead of doing this way

test/JDBC/upgrade/16_9/schedule
```diff
@@ -607,7 +607,8 @@ BABEL-NUMERIC_EXPR
 test_conv_float_to_varchar_char
 BABEL-5467
 BABEL_OBJECT_RESOLUTION_IN_ROUTINES
-test_constraint_like
+test_constraint_like-17-5-or-16-9
 round_return_type_test
 alter-view
 test_conv_int_to_varbinary_binary
+test_like_index
```
Deepesh125, line 614: isnt this same?

test/JDBC/expected/parallel_query/babel_collection.out
```diff
@@ -183,16 +183,18 @@ GO
 ~~START~~
 text
 Query Text: select c1 from testing4 where c1 LIKE 'jones'
-Gather  (cost=11.40..24.02 rows=3 width=32)
-  Workers Planned: 3
-  ->  Parallel Bitmap Heap Scan on testing4  (cost=11.40..24.02 rows=1 width=32)
-        Filter: ((c1)::text = 'jones'::text)
-        ->  Bitmap Index Scan on c1_idxtesting49a168d73f3ba5aacdfd495b931b8d187  (cost=0.00..11.40 rows=650 width=0)
+Gather  (cost=4.17..11.28 rows=3 width=32)
```
Deepesh125, line 186: can you please examplain Parallel Bitmap heap scan before vs Bitmap scan currently?

Deepesh125 (changes requested):

test/JDBC/expected/db_collation/test_like_index-vu-verify.out
```diff
@@ -0,0 +1,972 @@
...
+-- VARCHAR
+-- CS_AS
+ANALYZE master_dbo.test_like_index_vu_prepare_v_csas;
+GO
+
+-- tsql
+SET BABELFISH_SHOWPLAN_ALL ON
+GO
+
+SELECT * FROM test_like_index_vu_prepare_v_csas WHERE test_like_index_vu_prepare_v_csas_col LIKE 'cs1'
```
Deepesh125: have test case which will verify that index scan wont be picked up if you supply explicit collation that doesnt match with column collation

test/JDBC/expected/db_collation/test_like_index-vu-verify.out
```diff
@@ -0,0 +1,972 @@
+
```
Deepesh125, line 32: 1. Should ensure to add test cases for toast rows. Verify that it picks up index scan or bitmap index scan for Toasted entries.
2. Have complex index on multiple columns and check that query picks up index scan if predicate with combination from those column is there 
3. Have a join between two tables with LIKE in join predicate. Verify that inner side pick up index scan for outer join. Or even for hash join

test/JDBC/expected/db_collation/test_like_index-vu-verify.out
```diff
@@ -0,0 +1,972 @@
...
+-- tsql
+SET BABELFISH_SHOWPLAN_ALL ON
+GO
+
+SELECT * FROM test_like_index_vu_prepare_v_cias WHERE test_like_index_vu_prepare_v_cias_col LIKE 'ci1'
+GO
+~~START~~
+text
+Query Text: SELECT * FROM test_like_index_vu_prepare_v_cias WHERE test_like_index_vu_prepare_v_cias_col LIKE 'ci1'
+Index Only Scan using test_like_index_vu_prepare_v_ci2d75c4ad76ed9b58a88ba0a93bab0133 on test_like_index_vu_prepare_v_cias  (cost=0.28..8.29 rows=1 width=5)
```
Deepesh125: Add test where query would pick up `index scan` and not `index only scan`

Deepesh125 (changes requested):

contrib/babelfishpg_tsql/src/collation.c
ahmed-shameem: We are handling CS_AS in babelfish side so that it can use Index. Else, if we leave it to PG engine directly, the same issue of not finding compatible operators for original datatype will come up.

test/JDBC/expected/babel_collection.out
ahmed-shameem: postgres also shows same behaviour for same input. Based on my observation if the the of rows or the row size is high, the PG usually chooses Index. Else, it may opt for bitmap index scan

test/JDBC/expected/db_collation/babel_collection.out
ahmed-shameem: No, as the the column does NOT have explicit collation (column level collation), it picks up the database default. As this is exact match, we use that collation for ```=``` operator as well. If the query would have been something like this:
```
select c1 from testing4 where c1 LIKE 'jones' COLLATE DATABASE_DEFAULT
```
then we would expect the COLLATE in plan

test/JDBC/expected/non_default_server_collation/chinese_prc_ci_as/test_constraint_like-vu-prepare.out
ahmed-shameem: No it does not.

contrib/babelfishpg_tsql/src/collation.c
ahmed-shameem: remove_accente_internal* returns NVARCHAR. This info is used during index scan for  AI collations, PG validates whether the functional index's each field matches with the operand. As the return type is NVARCHAR in functional index, we must make sure that operand type is also the same

contrib/babelfishpg_tsql/src/collation.c
ahmed-shameem: That should not be the case, if we can not find any compatible operator then we should be using LIKE directly (which can be then handled by PG engine if there is any issue). As this is an optimisation and NOT measure of correctness, putting such restriction is NOT ideal .

contrib/babelfishpg_tsql/src/collation.c
ahmed-shameem: The third arg is exprtype which is the result type of the expression. After RAI is applied the result type is NVARCHAR, we want to coearce it into original type. Same for rightop.

contrib/babelfishpg_tsql/src/collation.c
ahmed-shameem: The third arg is exprtype which is the result type of the expression. After RAI is applied the result type is NVARCHAR, we want to coearce it into original type. Same for rightop.

contrib/babelfishpg_tsql/src/collation.c

contrib/babelfishpg_tsql/src/collation.c
ahmed-shameem: Updated in latest rev

contrib/babelfishpg_tsql/src/collation.c
ahmed-shameem: We can not because:
1. prefix->consttype is set here, howeverer prefix is initialised later
2. We only need to find the correct typeid if we want to have the optimisation, so logically this makes sense

contrib/babelfishpg_tsql/src/collation.c
ahmed-shameem: Updated in latest rev

test/JDBC/upgrade/17_5/schedule
ahmed-shameem: As discussed offline, this is because only one version has specific output

test/JDBC/expected/db_collation/test_like_index-vu-verify.out
ahmed-shameem: Added in latest rev

test/JDBC/expected/db_collation/test_like_index-vu-verify.out
ahmed-shameem: Added in latest rev except 1st one (toastable entry) as it is not always possible to make sure that there are entries in toast table

test/JDBC/expected/db_collation/test_like_index-vu-verify.out
ahmed-shameem: Added in latest rev

test/JDBC/expected/parallel_query/babel_collection.out
ahmed-shameem: There was no Index condition before:
```
 Index Cond: (c1 = 'jones'::"varchar")
```
which means, we were unable to use index efficiently

test/JDBC/expected/non_default_server_collation/chinese_prc_ci_as/babel_collection.out
ahmed-shameem: This is a const node, PG also shows optimisation

test/JDBC/upgrade/16_9/schedule
ahmed-shameem: As discussed offline, this is because this input file has output for these specific versions

Deepesh125 (dismissed):

Yvinayak07 (approved):

test/JDBC/upgrade/16_10/schedule
```diff
@@ -607,8 +607,9 @@ BABEL-NUMERIC_EXPR
 test_conv_float_to_varchar_char
 BABEL-5467
 BABEL_OBJECT_RESOLUTION_IN_ROUTINES
-test_constraint_like
+#test_constraint_like
```
rishabhtanwar29: Nit: why are we disabling this test?

rishabhtanwar29 (dismissed):

test/JDBC/upgrade/16_10/schedule
ahmed-shameem: We will re-enable it again when we have these changes in 16.latest else it will fail here.

forestkeeper (dismissed):

Deepesh125 (approved):

PR #3588, opened 2025-03-18
Support prefix term grammar for CONTAINS clause

jaspal007:
Previously in Babelfish, the feature of fulltext search for prefix term as search string in CONTAINS clause was not supported. Prefix term over TDS endpoint would throw unsupported error. To fix this, helper functions are created to
support the grammar of prefix term to provide the intended result.

### 1. Issue:
- Fulltext search over `prefix_term` is not supported in Babelfish. When a `prefix_term` is passed as a search string in FTS TSQL CONTAINS query it throws `prefix term not supported error`
  - current output
     ```
     1> SELECT * FROM prefix_table WHERE CONTAINS(col, '"work*"')
     2> GO
    Msg 33557097, Level 16, State 1, Server persist-tanyagp-perf-diligent-16-6-i, Line 1
    Prefix term is not currently supported in Babelfish
     ```
### 2. Changes made to fix the issue:
- Grammar is added in FTS parser to handle whitespaces to access prefix_term token from the search string.  
- Custom function is created in the FTS parser to handle the prefix term token and handle all the cases of leading, trailing  and unwanted whitespaces within the `prefix_term` token. 

TASK: BABEL-4380
Authored-by: Jaspal Singh ijaspals@amazon.com
Signed-off-by: Jaspal Singh ijaspals@amazon.com

### Issues Resolved
BABEL-4380

* **Use case based -** YES

* **Arbitrary inputs -** YES

### Check List


test/JDBC/input/full_text_search/fts-contains-prefix-vu-verify.sql
```diff
@@ -0,0 +1,66 @@
+-- enable FULLTEXT
+-- tsql user=jdbc_user password=12345678
```
roshan0708: This is redundant. Please remove this at other places as well since it's a `.sql` file not `.mix`.

contrib/babelfishpg_tsql/src/fts_parser.y
```diff
@@ -195,6 +205,95 @@ static char
...
+      ereport(ERROR,
+          (errcode(ERRCODE_INTERNAL_ERROR),
+           errmsg("Null or empty full-text predicate.")));
+    }
+    
+    trimmedInputStr = pstrdup(inputStr);
+
+    /* removing leading and trailing spaces outside of double quotes */
+    trim(trimmedInputStr, false);
+    inputLength = strlen(trimmedInputStr);
```
roshan0708: Seems redundant, we can remove this.

contrib/babelfishpg_tsql/src/fts_parser.y
```diff
@@ -195,6 +205,95 @@ static char
...
+            }
+            if(output.len > 0){
+              if(i == inputLength-1) {
+                  appendStringInfoString(&output, ":*");
+              }
+              else {
+                  appendStringInfoString(&output, ":*<->");
+              }
+            } else {
+                  continue;
```
roshan0708: Better to mention example cases which are being handled here.

contrib/babelfishpg_tsql/src/fts_parser.y
```diff
@@ -195,6 +205,95 @@ static char
...
+                if(output.len > 0){
+                    if(i == inputLength-1) {
+                        appendStringInfoString(&output, ":*");
+                    }
+                    else {
+                        appendStringInfoString(&output, ":*<->");
+                    }
+                } else {
+                        continue;
+                }
```
roshan0708: Please mention example cases here also.

test/JDBC/jdbc_schedule
```diff
@@ -10,6 +10,11 @@
 
 all
 
+# Ignore fts-contains-before files as they don't have test cases that support prefix term support
+ignore#!#fts-contains-before-vu-prepare
+ignore#!#fts-contains-before-vu-verify
+ignore#!#fts-contains-before-vu-cleanup
+
```
roshan0708: Seems redundant, we can altogether remove this if not needed.

test/JDBC/input/full_text_search/fts-contains-prefix-vu-verify.sql
```diff
@@ -0,0 +1,66 @@
...
+-- Create a unique-text index
+CREATE UNIQUE INDEX uid ON ArticleSentences(id)
+GO
+-- Create a full-text index
+CREATE FULLTEXT INDEX ON ArticleSentences(Content)
+KEY INDEX uid 
+GO
+
+-- Sample search query using CONTAINS
+-- 1. Basic prefix search across all text fields
```
roshan0708: Add more tests cases around different patterns of prefix terms as mentioned in the test plan:
- "one two three*"
- "one* two three*"
- "one* two* three*"
- "one*two*three*"
- other combinations with spaces and `*` characters

roshan0708 (changes requested):

contrib/babelfishpg_tsql/src/fts_parser.y
jaspal007: This function is being used to remove the extra spaces between the leading delimiters `"` and single quotes `'` and the trailing delimiters `"` and single quotes `'`
It handles cases when a customer might input search string with more than required spaces between single quotes and the delimiter `"`:
```
SELECT * FROM table_name WHERE CONTAINS({column_name | column_list}, '   "word1*"    ')
GO
```
This is treated as
```
SELECT * FROM table_name WHERE CONTAINS({column_name | column_list}, '"word1*"')
GO
```

contrib/babelfishpg_tsql/src/fts_parser.y
```diff
@@ -195,6 +205,109 @@ static char
...
+*translate_prefix_term(const char* inputStr) {
+    int             inputLength;
+    char            *trimmedInputStr;
+    char            *outputStr;
+    StringInfoData  output;
+    const char      *specialChars = "~!&|@#$%^+=\\;:<>?.\\/";
+
+
+    /* Check for empty input - this should not be possible based on lexer rules, but check just in case */
+    if (!inputStr || !(inputLength = strlen(inputStr))) {
```
thephantomthief: do we need to do assignment here to `inputLength`?

contrib/babelfishpg_tsql/src/fts_parser.y
thephantomthief: do we need to do assignment here to `inputLength`? We are again overwriting it in L248

contrib/babelfishpg_tsql/src/fts_parser.y
```diff
@@ -195,6 +205,109 @@ static char
...
+     */
+    trim(trimmedInputStr, true);
+    
+    /* Rewriting search string in format word1:*<->word2:*  */
+    outputStr = pstrdup(trimmedInputStr);
+    inputLength = strlen(outputStr);
+    initStringInfo(&output);
+
+    for(int i=0; i<inputLength; i++) {
+        if (strchr(specialChars, outputStr[i]) != NULL || strchr("`'_", outputStr[i]) != NULL) {
```
thephantomthief: why don't we put all the special chars in `specialChars`

contrib/babelfishpg_tsql/src/fts_parser.y
```diff
@@ -195,6 +205,109 @@ static char
...
+            ereport(ERROR,
+                (errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
+                errmsg("Special characters in the prefix term search condition are not currently supported in Babelfish")));
+        }
+        if (outputStr[i] == ' ') {
```
thephantomthief: what about tab, new line characters?

contrib/babelfishpg_tsql/src/fts_parser.y
```diff
@@ -195,6 +205,109 @@ static char
...
+            /* 
+             * Removing multiple spaces and * from the search string 
+             * If a space is encountered, we remove all the next occurances of * and spaces 
+             * before end of the input or if next word is encountered
+             * Case 1: '"word1   * * ** *"' = 'word1:*'
+             * Case 2: '"word1   * * ** * word2*"' = 'word1:*<->word2:*'
+             */
+            while(i < inputLength-1 && (outputStr[i+1] == '*' || outputStr[i+1] == ' ')) {
```
thephantomthief: what about tab, new line characters?

contrib/babelfishpg_tsql/src/fts_parser.y
```diff
@@ -195,6 +205,108 @@ static char
...
+              if(i == inputLength-1) {
+                  appendStringInfoString(&output, ":*");
+              }
+              else {
+                  appendStringInfoString(&output, ":*<->");
+              }
+            } else {
+                  continue;
+            }
+        } else if(outputStr[i] == '*') {
```
thephantomthief: Can we combine the above else if with this block

contrib/babelfishpg_tsql/src/fts_parser.y
```diff
@@ -195,6 +205,108 @@ static char
...
+                        appendStringInfoString(&output, ":*<->");
+                    }
+                } else {
+                        continue;
+                }
+        } else {
+                appendStringInfoChar(&output, outputStr[i]);
+        }
+    }
+    appendStringInfoChar(&output, '\0');
```
thephantomthief: This is not needed

contrib/babelfishpg_tsql/src/fts_parser.y
roshan0708: for strings with special characters "`'_", these result in exact matches for simple term but @jaspal007 We should check if such prefix term search conditions match the same results, then only we should use this.

contrib/babelfishpg_tsql/src/fts_parser.y
roshan0708: We can just simply check `(!inputStr || !*inputStr)`

contrib/babelfishpg_tsql/src/fts_parser.y
roshan0708: @jaspal007 Current implementation of this function can be further simplified. IMO:
* Simplify nested conditions for spaces/asterisks handling into single pattern
* Reduce memory buffers from two to one for better memory management
* Consolidate redundant string operations for `':*'` and `'<->'` separators
* Use pointer arithmetic instead of array indexing for better performance

contrib/babelfishpg_tsql/src/fts_parser.y
jaspal007: The intended output of sql server is that it returns no effective records if we have tab or newline character present in prefix term search string.

contrib/babelfishpg_tsql/src/fts_parser.y
```diff
@@ -195,6 +205,93 @@ static char
...
+
+    /* 
+     * Rewriting search string in format word1:*<->word2:*  
+     */
+    while(*trimmedOutputStr) {
+        if (strchr(specialChars, *trimmedOutputStr) != NULL) {
+            ereport(ERROR,
+                (errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
+                errmsg("Special characters in the prefix term search condition are not currently supported in Babelfish")));
+        }
```
roshan0708: When an error is thrown here, both `outputStr` and `output.data` memory will be leaked. Please free these allocations before calling ereport().

contrib/babelfishpg_tsql/src/fts_parser.y
```diff
@@ -195,6 +205,93 @@ static char
...
+              }
+            }
+        } else if (*trimmedOutputStr == '\n' || *trimmedOutputStr == '\t') {
+                /*
+                 * To match the behavior of SQL server.
+                 * When FTS is performed over prefix_term,
+                 * it outputs no result
+                 */ 
+                pfree(outputStr);
+                output.data = "";
```
roshan0708: setting `output.data` to a string literal after initialization without proper cleanup will cause memory leaks. Use `resetStringInfo(&output)` instead to properly clean up the previously allocated memory.

roshan0708 (changes requested):

roshan0708:
We should add test cases for emojis, CTEs as well

contrib/babelfishpg_tsql/src/fts_parser.y
```diff
@@ -195,6 +205,141 @@ static char
...
+
+    initStringInfo(&outputStr);
+    output = pstrdup(inputStr);
+
+    /*
+     * removing spaces between the leading single quote (') and leading delimiter (") and
+     * trailing single quote (') and trailing delimiter (")
+     * '   "word1 word2*"  ' = '"word1 word2*"'
+     */
+    trim(output, false);
```
thephantomthief, line 242: doesn't `trim(output, true)` not also do this implicitly?

contrib/babelfishpg_tsql/src/fts_parser.y
```diff
@@ -195,6 +205,141 @@ static char
...
+    start = output;
+    end = output + (strlen(output) - 1);
+        
+    /*
+     * rewriting search string in format word1:*<->word2:* 
+     */
+    while(start<=end) {
+        if (strchr(specialChars, *start) != NULL) {
+            pfree(output);
+            resetStringInfo(&outputStr);
```
thephantomthief, line 271: isn't this already empty at this point?

contrib/babelfishpg_tsql/src/fts_parser.y
```diff
@@ -195,6 +205,141 @@ static char
...
+        /* 
+        * removing multiple spaces, tabs and * from the search string 
+        * If a space is encountered, we remove all the next occurances of * and spaces and tabs
+        * before end of the input or if next word is encountered
+        * Case 1: '"word1   * * ** *"' = 'word1:*'
+        * Case 2: '"word1   * * ** * word2*"' = 'word1:*<->word2:*'
+        * Case 3: '"word1* *' + CHAR(9) + '** *"' = 'word1:*'
+        * Case 4: '"word1* * *  ' + CHAR(9) + '* ' + CHAR(9) + ' *** * word2*"' = 'word1:*<->word2:*'
+        */
+        if(*start == ' ' || *start == '*' || *start == '\t') {
```
thephantomthief: if there is a newline character, we shouldn't remove the following spaces, tabs etc.?

contrib/babelfishpg_tsql/src/fts_parser.y
```diff
@@ -195,6 +205,141 @@ static char
...
+        }
+        /*
+         * trim the extra 
+         * space, 
+         * asterisk, 
+         * a tab character or 
+         * a newline character 
+         * at the end of the search string
+         */
+        while(start<=end && (*end == ' ' || *end == '*' || *end == '\t' || *end == '\n')) {
```
thephantomthief: we only need to run this loop once so we can take it out of the outer while loop

contrib/babelfishpg_tsql/src/fts_parser.y
```diff
@@ -195,6 +205,141 @@ static char
...
+            }
+
+            /*
+             * to handle the case when a newline character is encountered
+             * while removing extra space, asterisk and tab character
+             * '"word1' + CHAR(9) + ' ' + CHAR(10) + 'word2*"' = 'word1:*<->uniqueHash:*<->word2:*'
+             */
+            if(*(start + 1) == '\n') {
+                start++;
+                continue;
```
thephantomthief, line 297: why do we need `continue` here?

contrib/babelfishpg_tsql/src/fts_parser.y
```diff
@@ -195,6 +205,141 @@ static char
...
+                }
+                /*
+                 * the trailing newline characters are removed in the beginning
+                 * but added this safety check
+                 */
+                if(start!=end) {
+                    char *newlineHash = replace_special_chars_fts_impl("\n");
+                    trim(newlineHash, false);
+                    appendStringInfoString(&outputStr, ":*<->");
+                    appendStringInfo(&outputStr, "%s", newlineHash);
```
thephantomthief: why not use `appendStringInfoString` again here?

contrib/babelfishpg_tsql/src/pltsql_utils.c
```diff
@@ -2187,21 +2188,23 @@ char
...
-	StringInfoData	output_str;
+	char            	*unique_hashes[5];
+	const char		*special_chars[5] = {"~!&|@#$%^*+=\\;:<>?./", "`", "'", "_", "\n"};
+	StringInfoData		output_str;
 	
-	for (int i = 0; i < 4; i++) {
-		unique_hashes[i] = construct_unique_index_name("specialChars", psprintf("cat%d", i + 1));
+	for (int i = 0; i < 5; i++) {
+		unique_hashes[i] = construct_unique_index_name("ftSpecial", psprintf("cat%d", i + 1));
+		unique_hashes[i][21] = '\0';
```
thephantomthief: Are we certain every unique hash will at least be 22 chars long?

jaspal007: the unique hash value generated by `construct_unique_index_name` is quite large, we require a hash value that is smaller and yet has a good collision avoidance

contrib/babelfishpg_tsql/src/fts_parser.y
jaspal007: we require `continue` here as the newline character is handled in a separate `if` clause

contrib/babelfishpg_tsql/src/fts_parser.y
jaspal007: That is handled in `if` clause for newline character separately

contrib/babelfishpg_tsql/src/fts_parser.y
jaspal007: This is for the case if special characters are present in the middle of search stirng
`'"word@1 wor$d2*"'`

contrib/babelfishpg_tsql/src/fts_parser.y
jaspal007: When we call `trim(output, true)` it processes the leading spaces within the double quotes and search string.
To remove the trailing spaces outside double quotes, we implement `trim(output, false)`

contrib/babelfishpg_tsql/src/fts_parser.y
```diff
@@ -195,6 +205,142 @@ static char
...
+    rightPtr = output + (strlen(output) - 1);
+        
+    /*
+        * trim the extra 
+        * space, 
+        * asterisk, 
+        * a tab character or 
+        * a newline character 
+        * at the end of the search string
+        */
```
roshan0708: nit: Indentation. Also, we can condense this comment in a line or two.

contrib/babelfishpg_tsql/src/fts_parser.y
```diff
@@ -195,6 +205,142 @@ static char
...
+        /* 
+        * removing multiple spaces, tabs and * from the search string 
+        * If a space is encountered, we remove all the next occurances of * and spaces and tabs
+        * before end of the input or if next word is encountered
+        * Case 1: '"word1   * * ** *"' = 'word1:*'
+        * Case 2: '"word1   * * ** * word2*"' = 'word1:*<->word2:*'
+        * Case 3: '"word1* *' + CHAR(9) + '** *"' = 'word1:*'
+        * Case 4: '"word1* * *  ' + CHAR(9) + '* ' + CHAR(9) + ' *** * word2*"' = 'word1:*<->word2:*'
+        */
+        if(*leftPtr == ' ' || *leftPtr == '*' || *leftPtr == '\t') {
```
roshan0708: nit: Spacing after `if`. Check other places as well.

contrib/babelfishpg_tsql/src/pltsql_utils.c
```diff
@@ -50,6 +50,7 @@ extern bool restore_tsql_tabletype;
 extern bool babelfish_dump_restore;
 extern char *get_cur_db_name(void);
 extern char *construct_unique_index_name(char *index_name, char *relation_name); 
+extern char *generate_fts_char_hash(char *prefix, char *bucket);
```
roshan0708: We can remove this now since it's redundant.

contrib/babelfishpg_tsql/src/pltsql_utils.c
```diff
@@ -2200,8 +2201,9 @@ char
 	for (size_t i = 0; i < input_len; i++) {
 		replacement = NULL;
 
-		for (int j = 0; j < 4; j++) {
-			if (strchr(special_chars[j], input_str[i]) != NULL) {
+		for (int j = 0; j < 5; j++) {
+			char *chr;
+			if ((chr = strchr(special_chars[j], input_str[i])) != NULL) {
```
roshan0708: What is this change for? Are we using `chr` somewhere else?

roshan0708:
Regarding tests, I think creating three separate tables for testing is redundant and we could write tests using a single table itself.

roshan0708 (approved):

jaspal007:
nit: fix spacings and indentations

thephantomthief (approved):

jsudrik (approved):
Approved.

PR #3595, opened 2025-03-20
Objects should be owned by schema owner in Babelfish

shalinilohia50:
When a new object is created in Babelfish, its owner should be same as its schema owner.

Upgrade approach: Older objects will continue to have the owner (which can be different than the schema owner). There are functions provided to check the objects where there is owner mismatch between object and its schema. User can run the scripts and update the ownership themselves.

Cherry-picked from https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/3560

Task: BABEL-5361

Signed-off-by: Shalini Lohia lshalini@amazon.com


HarshLunagariya (approved):

jsudrik (approved):
Approved.

PR #3644, opened 2025-04-01
Avoid checking permission of Babelfish temp tables on parallel worker

Deepesh125:
### Description

Consider following facts,

Babelfish temp tables are implemented using ENR which is not shared between different backends. So if Parallel worker
tries to check permissions on Babelfish then it will fail.

Any user should be able to access Babelfish temp tables under given session.

Postgres by default does not allow parallel operations on temp tables. Attempt to do so will result in run time error.

Due to above facts, we should avoid permission check on temp tables within parallel workers while ensuring that leader
does required permission check on other tables. This commits achieves this behaviour by implementating following three
hooks,

bbf_ParallelQueryMain_hook -- implements ParallelQueryMain_hook. It constructs Bitmapset (list of temp table oids)
by looping through es_range_table checking if given relation is temp table, checks permission if it is and add it to the
set. It will then pass this set of temp table defined under current with Parallel workers.

bbf_ExecInitParallelPlan -- implements ExecInitParallelPlan_hook. Parallel workers will gather additional details (oid set of
temp table) passed by Leader node. This set will later be used to avoid checking permissions on temp tables.

bbf_ExecCheckOneRelPerms_hook -- implements ExecCheckOneRelPerms_hook. It will avoid permission check relation
if relid is part of the set of oids constructed using bbf_ExecInitParallelPlan. 

### Issues Resolved

BABEL-5703

### Check List

### DIFF
--- a/contrib/babelfishpg_tsql/src/bbf_parallel_query.c
+++ b/contrib/babelfishpg_tsql/src/bbf_parallel_query.c
@@ -152,9 +152,9 @@ bbf_ParallelQueryMain(shm_toc *toc)
 bool
 bbf_ExecCheckOneRelPerms(RTEPermissionInfo *perminfo)
 {
-	if (prev_ExecCheckOneRelPerms_hook && (*prev_ExecCheckOneRelPerms_hook)(perminfo))
+	if (prev_ExecCheckOneRelPerms_hook && !(*prev_ExecCheckOneRelPerms_hook)(perminfo))
 	{
-		return true;
+		return false;
 	}
 
 	if (!OidIsValid(perminfo->relid))
--- a/contrib/babelfishpg_tsql/src/bbf_parallel_query.c
+++ b/contrib/babelfishpg_tsql/src/bbf_parallel_query.c
@@ -17,6 +17,7 @@
 #include "utils/acl.h"
 #include "utils/elog.h"
 #include "utils/lsyscache.h"
+#include "utils/queryenvironment.h"
 
 #include "bbf_parallel_query.h"
 #include "pltsql.h"
@@ -81,7 +82,7 @@ bbf_ExecInitParallelPlan(EState *estate, ParallelContext *pcxt, bool estimate)
 			RangeTblEntry *rte = (RangeTblEntry *) lfirst(lc);
 			if (rte->rtekind == RTE_RELATION &&
 				OidIsValid(rte->relid) &&
-				get_rel_persistence(rte->relid) == RELPERSISTENCE_TEMP)
+				get_ENR_withoid(currentQueryEnv, rte->relid, ENR_TSQL_TEMP) != NULL)
 			{
 				/* probably (re)do perm check */
 				RTEPermissionInfo *perminfo = getRTEPermissionInfo(estate->es_plannedstmt->permInfos, rte);
--- a/contrib/babelfishpg_tsql/src/bbf_parallel_query.c
+++ b/contrib/babelfishpg_tsql/src/bbf_parallel_query.c
@@ -134,7 +134,7 @@ bbf_ParallelQueryMain(shm_toc *toc)
 		(*prev_ParallelQueryMain_hook)(toc);
 
 	/* Another line of defense to make sure no regular backend calls this function. */
-	if (IsBabelfishParallelWorker())
+	if (!IsBabelfishParallelWorker())
 	{
 		return;
 	}
--- a/contrib/babelfishpg_tsql/src/bbf_parallel_query.c
+++ b/contrib/babelfishpg_tsql/src/bbf_parallel_query.c
@@ -2,6 +2,7 @@
 #include "postgres.h"
 
 #include "access/parallel.h"
+#include "catalog/pg_class.h"
 #include "executor/executor.h"
 #include "fmgr.h"
 #include "nodes/bitmapset.h"
@@ -80,7 +81,7 @@ bbf_ExecInitParallelPlan(EState *estate, ParallelContext *pcxt, bool estimate)
 			RangeTblEntry *rte = (RangeTblEntry *) lfirst(lc);
 			if (rte->rtekind == RTE_RELATION &&
 				OidIsValid(rte->relid) &&
-				get_rel_persistence(rte->relid) == 't')
+				get_rel_persistence(rte->relid) == RELPERSISTENCE_TEMP)
 			{
 				/* probably (re)do perm check */
 				RTEPermissionInfo *perminfo = getRTEPermissionInfo(estate->es_plannedstmt->permInfos, rte);
--- a/contrib/babelfishpg_tsql/src/bbf_parallel_query.c
+++ b/contrib/babelfishpg_tsql/src/bbf_parallel_query.c
@@ -1,0 +1,148 @@
+
+#include "postgres.h"
+
+#include "access/parallel.h"
+#include "fmgr.h"
+#include "nodes/bitmapset.h"
+#include "nodes/execnodes.h"
+#include "nodes/nodes.h"
+#include "nodes/parsenodes.h"
+#include "nodes/pg_list.h"
+#include "parser/parser.h"
+#include "storage/shm_toc.h"
+#include "utils/elog.h"
+#include "utils/lsyscache.h"
+
+#include "bbf_parallel_query.h"
+
+/*
+ * temp_relids - maintains relid list of temp table shared by leader node. This should be
+ * strictly accessed within parallel worker context.
+ */
+static Bitmapset   *temp_relids = NULL;
+
+/*
+ * string representation of list of temp table oids computed by Leader node to be shared
+ * with parallel worker.
+ */
+static char		   *temp_relids_str = NULL;
+
+/*
+ * In Babelfish, Any user under given session should be able access the temp tables. 
+ * And Postgres doesn't allow parallel scan on temp tables. But it still does the permission
+ * check on temp tables under parallel workers. But since Babelfish temp tables implemented
+ * using ENR, it can't be accessed inside Parallel worker.
+ * 
+ * So we would like to avoid permission checking on temp tables under parallel workers while
+ * ensuring that Leader node does require permission checks on temp table.
+ * 
+ * In order to achieve this, we (probably re)do permission checks on temp tables and share
+ * the list of oids with parallel worker. And parallel worker will avoid permission check
+ * on these oids.
+ */
+
+
+/*
+ * bbf_ExecInitParallelPlan -- implements ExecInitParallelPlan_hook.
+ * It iterates through es_range_tables checking persistence of given relation. Probably,
+ * re-do permission checking (better to redo perm checking instead of never doing it) if
+ * relation is temp table and adds oid/relid to the set. This set will be shared with
+ * parallel worker so that parallel worker avoids permission check on temp tables.
+ */
+void 
+bbf_ExecInitParallelPlan(EState *estate, ParallelContext *pcxt, bool estimate)
+{
+	if (prev_ExecInitParallelPlan_hook)
+		(*prev_ExecInitParallelPlan_hook)(estate, pcxt, estimate);
+
+	if (sql_dialect != SQL_DIALECT_TSQL)
+		return;
+
+	if (estimate)
+	{
+		ListCell   *lc;
+		Bitmapset  *temp_relids_local = NULL;
+
+		foreach(lc, estate->es_range_table)
+		{
+			RangeTblEntry *rte = (RangeTblEntry *) lfirst(lc);
+			if (rte->rtekind == RTE_RELATION &&
+				OidIsValid(rte->relid) && 
+				get_rel_persistence(rte->relid) == 't')
+			{
+				temp_relids_local = bms_add_member(temp_relids_local, rte->relid);
+			}
+		}
+		temp_relids_str = bmsToString(temp_relids_local);
+
+		/*
+		 * Estimate extra context for Babelfish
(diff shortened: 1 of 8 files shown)


contrib/babelfishpg_tsql/src/bbf_parallel_query.c
```diff
@@ -0,0 +1,175 @@
...
+ * warning: should stricktly call under parallel worker.
+ */
+void
+bbf_ParallelQueryMain(shm_toc *toc)
+{
+	if (prev_ParallelQueryMain_hook)
+		(*prev_ParallelQueryMain_hook)(toc);
+
+	/* Another line of defense to make sure no regular backend calls this function. */
+	if (IsBabelfishParallelWorker())
```
surendravishnoi: this seems incorrect

contrib/babelfishpg_tsql/src/bbf_parallel_query.c
```diff
@@ -0,0 +1,175 @@
...
+
+/*
+ * bbf_ExecCheckOneRelPerms -- implements ExecCheckOneRelPerms_hook.
+ * Returns true if this is Babelfish parallel worker and provided relid is Babelfish temp table. 
+ * Note that temp_relids must have communicated by Leader node.
+ */
+bool
+bbf_ExecCheckOneRelPerms(RTEPermissionInfo *perminfo)
+{
+	if (prev_ExecCheckOneRelPerms_hook && (*prev_ExecCheckOneRelPerms_hook)(perminfo))
```
surendravishnoi: we should return false if prev hook returns false and do perm check in case it returns true

test/JDBC/expected/Test_parallel_query.out
```diff
@@ -0,0 +1,223 @@
+-- tsql
+CREATE LOGIN l1 WITH PASSWORD = '12345678'
```
surendravishnoi, line 3: make sure to add tests that non BBF is working as expected

contrib/babelfishpg_tsql/src/bbf_parallel_query.c
```diff
@@ -0,0 +1,175 @@
...
+	{
+		ListCell   *lc;
+		Bitmapset  *temp_relids_local = NULL;
+
+		foreach(lc, estate->es_range_table)
+		{
+			RangeTblEntry *rte = (RangeTblEntry *) lfirst(lc);
+			if (rte->rtekind == RTE_RELATION &&
+				OidIsValid(rte->relid) &&
+				get_rel_persistence(rte->relid) == 't')
```
surendravishnoi: is this sufficient to to know that it is BBF temp table?

test/JDBC/expected/Test_parallel_query.out
```diff
@@ -0,0 +1,223 @@
...
+GO
+
+CREATE USER l1
```
surendravishnoi, line 6: we should have a test where
accessing temp table inside parallel worker should give error
accessing temp table as outer scan should not give an error
accessing PG temp table inside parallel worker should give error
accessing PG temp table as outer scan should not give an error (but this will not skip the check, rather check will pass)
accessing PG temp table as outer scan should give an error (due to perm issue)
accessing PG regular table in both case should always give an error
mix of all three tables in same query

contrib/babelfishpg_tsql/src/bbf_parallel_query.c
```diff
@@ -0,0 +1,175 @@
...
+			{
+				/* probably (re)do perm check */
+				RTEPermissionInfo *perminfo = getRTEPermissionInfo(estate->es_plannedstmt->permInfos, rte);
```
surendravishnoi: should we do perm check while setting memory for parallel workers

contrib/babelfishpg_tsql/src/bbf_parallel_query.c
Deepesh125: Checking with @timchang514

contrib/babelfishpg_tsql/src/bbf_parallel_query.c
Deepesh125: we are doing it on next line.

contrib/babelfishpg_tsql/src/bbf_parallel_query.c
timchang514: yes, all bbf temp tables will have relpersistence 't'. it would be clearer to use `RELPERSISTENCE_TEMP` here or better yet, use the function `IsTsqlTempTable(relpersistence)`, although that also performs a dialect check which i'm not sure if it will affect the intended behavior here.

I saw in the comment above though that the intention is to skip permissions checks on temp tables in parallel workers since they are in ENR. There can also be non-ENR temp tables in rare cases from users (such as if they are created with user-defined types), though I think we wouldn't necessarily want differing behavior here between non-ENR and ENR temp tables

Deepesh125: Yes, we want to avoid dialect check. But please note that we are not avoid permission checks on Leader node. It will be done for sure but we are avoiding it on parallel worker where we anyway dont need temp table

contrib/babelfishpg_tsql/src/bbf_parallel_query.c
```diff
@@ -0,0 +1,172 @@
...
+
+		foreach(lc, estate->es_range_table)
+		{
+			RangeTblEntry *rte = (RangeTblEntry *) lfirst(lc);
+			if (rte->rtekind == RTE_RELATION &&
+				OidIsValid(rte->relid) &&
+				get_rel_persistence(rte->relid) == RELPERSISTENCE_TEMP)
+			{
+				/* probably (re)do perm check */
+				RTEPermissionInfo *perminfo = getRTEPermissionInfo(estate->es_plannedstmt->permInfos, rte);
```
surendravishnoi: let us double check that all -> references are safe to do or put additional NULL checks

contrib/babelfishpg_tsql/src/bbf_parallel_query.c
```diff
@@ -0,0 +1,172 @@
...
+		 */
+		shm_toc_estimate_chunk(&pcxt->estimator, strlen(temp_relids_str) + 1);
+		shm_toc_estimate_keys(&pcxt->estimator, 1);
+	}
+	else
+	{
+		char *temp_relids_space;
+
+		/*temp_relids_str will never be NULL even if there is no temp tables in the query. */
+		if (temp_relids_str == NULL)
```
surendravishnoi, line 124: this should be an assert or internal error. User has no context of temp table rel id list

contrib/babelfishpg_tsql/src/bbf_parallel_query.c
```diff
@@ -0,0 +1,172 @@
...
+					(errcode(ERRCODE_OBJECT_NOT_IN_PREREQUISITE_STATE),
+					 errmsg("Unexpected list of temp table relids")));
+		}
+
+		temp_relids_space = shm_toc_allocate(pcxt->toc, strlen(temp_relids_str) + 1);
+		memcpy(temp_relids_space, temp_relids_str, strlen(temp_relids_str) + 1);
+		shm_toc_insert(pcxt->toc, BABELFISH_PARALLEL_KEY_TEMP_RELIDS, temp_relids_space);
+
+		/* And reset temp_relids_str */
+		pfree(temp_relids_str);
```
surendravishnoi, line 136: we leak it in case of error, make sure to reset at some known query entry point or at-least at the beginning of this function

contrib/babelfishpg_tsql/src/bbf_parallel_query.c
```diff
@@ -0,0 +1,172 @@
...
+ * bbf_ExecCheckOneRelPerms -- implements ExecCheckOneRelPerms_hook.
+ * Returns true if this is Babelfish parallel worker and provided relid is Babelfish temp table. 
+ * Note that temp_relids must have communicated by Leader node.
+ */
+bool
+bbf_ExecCheckOneRelPerms(RTEPermissionInfo *perminfo)
+{
+	if (!OidIsValid(perminfo->relid))
+	{
+		ereport(ERROR,
```
surendravishnoi, line 172: should be internal error

test/JDBC/expected/Test_parallel_query.out
```diff
@@ -0,0 +1,367 @@
...
+GO
+
+CREATE TABLE pq_t (id INT)
```
surendravishnoi, line 9: It is not clear if all test cases are covered, can you put one line comment for each test to describe what use case it covers like
-- Skip permission check for temp table in BBF on parallel worker
-- Skip permission check for PG temp table in BBF on parallel worker
-- Permission check error for PG temp table without parallel worker and so on

contrib/babelfishpg_tsql/src/bbf_parallel_query.c
Deepesh125: Added sanity check for estate->es_plannedstmt->permInfos

contrib/babelfishpg_tsql/src/bbf_parallel_query.c
Deepesh125: Converted to internal error

contrib/babelfishpg_tsql/src/bbf_parallel_query.c
Deepesh125: Done

contrib/babelfishpg_tsql/src/bbf_parallel_query.c
```diff
@@ -0,0 +1,184 @@
...
+			if (rte->rtekind == RTE_RELATION &&
+				OidIsValid(rte->relid) &&
+				get_rel_persistence(rte->relid) == RELPERSISTENCE_TEMP)
+			{
+				if (estate->es_plannedstmt == NULL ||
+					list_length(estate->es_plannedstmt->permInfos) == 0)
+				{
+					/* If there is RTE present then corresponding permInfo must be there */
+					ereport(ERROR,
+							(errcode(ERRCODE_INTERNAL_ERROR),
```
surendravishnoi, line 94: Add more context to error like during parallel query execution, failed to check permissions due to missing permission info in planned statement

contrib/babelfishpg_tsql/src/bbf_parallel_query.c
```diff
@@ -0,0 +1,184 @@
...
+	else
+	{
+		char *temp_relids_space;
+
+		/*temp_relids_str will never be NULL even if there is no temp tables in the query. */
+		if (temp_relids_str == NULL)
+		{
+			ereport(ERROR,
+					(errcode(ERRCODE_INTERNAL_ERROR),
+					 errmsg("Unexpected list of temp table relids")));
```
surendravishnoi, line 128: more context

contrib/babelfishpg_tsql/src/bbf_parallel_query.c
```diff
@@ -0,0 +1,184 @@
...
+ * Note that temp_relids must have communicated by Leader node.
+ */
+bool
+bbf_ExecCheckOneRelPerms(RTEPermissionInfo *perminfo)
+{
+	if (!OidIsValid(perminfo->relid))
+	{
+		ereport(ERROR,
+			(errcode(ERRCODE_INTERNAL_ERROR),
+			 errmsg("Unexpected perminfo is found")));
```
surendravishnoi: more context

contrib/babelfishpg_tsql/src/bbf_parallel_query.c
```diff
@@ -0,0 +1,184 @@
...
+	 * doing plpgsql function scan on leader node.
+	 */
+	if (!IS_TDS_CONN())
+		return;
+
+	if (estimate)
+	{
+		ListCell   *lc;
+		Bitmapset  *temp_relids_local = NULL;
+		temp_relids_str = NULL;
```
surendravishnoi, line 79: what about memory? is it leaked or corresponding context is cleaned up by now?

Deepesh125: In normal case, we are explicitely cleaning up the memory. But it should get free by error handling in case of error

contrib/babelfishpg_tsql/src/bbf_parallel_query.c
Deepesh125: Added

contrib/babelfishpg_tsql/src/bbf_parallel_query.c
Deepesh125: Added

contrib/babelfishpg_tsql/src/bbf_parallel_query.c
Deepesh125: fixed

test/JDBC/expected/Test_parallel_query.out
Deepesh125: done

contrib/babelfishpg_tsql/src/bbf_parallel_query.c
Deepesh125: removed prev hook

test/JDBC/expected/Test_parallel_query.out
Deepesh125: Added

kuntalghosh (dismissed):

surendravishnoi (approved):

kuntalghosh (approved):

PR #3661, opened 2025-04-08
Prepare 5_2_STABLE branch

timchang514:
- Disable fault injection tests
- Point workflows towards new stable branches
- Update branch name where appropriate


roshan0708:
@timchang514 We also need to update `singledb-version-upgrade.yml` for fixing singledb-version-upgrade from 17.4 to 17.5. You can check this PR: #3475 

This needs to be done for other stable branches as well.

roshan0708:
@rishabhtanwar29 Should we also update `upgrade-test-configuration.yml` for new version upgrade?

timchang514:
@roshan0708 thanks, fixed that. 

In general do we keep track of the changes needed for these branch creation commits in a PR somewhere, with explanations? 

For instance, it's not immediately clear why `singledb-version-upgrade` is based on 2X in 5_0_STABLE branch, but we moved to 4_5_STABLE in 5_1. Is it because of antlr version, or a miss in 5_0? Why don't tests fail if we have the wrong branch?

roshan0708:
> @roshan0708 thanks, fixed that.
> 
> In general do we keep track of the changes needed for these branch creation commits in a PR somewhere, with explanations?
> 
> For instance, it's not immediately clear why `singledb-version-upgrade` is based on 2X in 5_0_STABLE branch, but we moved to 4_5_STABLE in 5_1. Is it because of antlr version, or a miss in 5_0? Why don't tests fail if we have the wrong branch?

@timchang514 As far as I know, we don't keep a list, perhaps @rishabhtanwar29 can give more clarity on this. 

Also, right now, the singledb version upgrade test is failing because we missed updating the antlr version in the `.yml` file.

timchang514:
Sure, I corrected that file as well. 

What I mean is, when we create these new branches, the tests are all passing if we don't make the proper changes. Or even if we only make a subset of the required changes (since all the tests were green before the two latest commits on this branch). So we don't have any proper verification that all the proper changes have been made. Even a quip with the instructions would be insufficient unless it is always kept up-to-date, but test failures would always indicate a mismatch in the github actions metadata.

For ex, I would expect that as soon as I change the Version.config or maybe the `babelfish_internal_version.h` (though it looks like that is only meant for display by internal function call) for all of the tests in the PR to fail until exactly all the right changes were made, but that isn't the case. Could we devise a set of test cases that verifies the cases in the github actions scripts based on whether we are on a dev/stable branch?

.github/workflows/minor-version-upgrade.yml
```diff
@@ -4,8 +4,8 @@ on: [push, pull_request]
 jobs:
   extension-tests:
     env:
-      ENGINE_VER_FROM: BABEL_5_1_STABLE__PG_17_4
-      EXTENSION_VER_FROM: BABEL_5_1_STABLE
+      ENGINE_VER_FROM: BABEL_5_2_STABLE__PG_17_5
+      EXTENSION_VER_FROM: BABEL_5_2_STABLE
```
rishabhtanwar29: Can keep it same

.github/configuration/upgrade-test-configuration.yml
```diff
@@ -178,6 +178,18 @@ upgrade-version: [{
...
+    {
+      version: 17.5,
+      upgrade-type: null
+    },
+    { 
+      version: target.latest,
+      upgrade-type: minor
+    }
+  ]
+},
```
rishabhtanwar29: Don't need to add this.

.github/configuration/upgrade-test-configuration.yml
```diff
@@ -292,5 +304,17 @@ upgrade-version: [{
...
+  upgrade-path: [
+    {
+      version: 16.10,
+      upgrade-type: null
+    },
+    { 
+      version: target.latest,
+      upgrade-type: major
+    }
+  ]
```
rishabhtanwar29: Don't need to add this.

.github/configuration/upgrade-test-configuration.yml
```diff
@@ -133,7 +133,7 @@ upgrade-version: [{
 {
   upgrade-path: [
     {
-      version: '15.12',
```
roshan0708, line 136: @rishabhtanwar29 Do we need this?

roshan0708 (approved):

priyansx (approved):
LGTM

PR #3718, opened 2025-04-23
Chunk server_hello into size of 4096 during SSL handshake

Deepesh125:
### Description

During SSL handshake, when server_hello message exceeds the TDS packet size limit of 4096 bytes, the connection fails as the max allowed packets size is 4096 while doing SSL handshake. 

This commit make changes to split server_hello messages into chunks of of 4096 bytes. Mandatory TDS header (8 byes) are included in the each chunk.

Task: BABEL-5787
Authored-by: Sumit Jaiswal (sumiji@amazon.com)
Co-Authored-by: Dipesh Dhameliya (dddhamel@amazon.com)

### Check List

### DIFF
--- a/contrib/babelfishpg_tds/src/backend/tds/tdssecure.c
+++ b/contrib/babelfishpg_tds/src/backend/tds/tdssecure.c
@@ -193,7 +193,6 @@ SslWrite(BIO * h, const char *buf, int size)
 static int
 SslHandShakeWrite(BIO * h, const char *buf, int size)
 {
-	#define SSL_MAX_PACKET_SIZE 4096
 	StringInfoData	str;
 	char		tmp[2];
 	uint16_t	tsize;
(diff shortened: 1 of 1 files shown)


contrib/babelfishpg_tds/src/backend/tds/tdssecure.c
```diff
@@ -193,56 +193,71 @@ SslWrite(BIO * h, const char *buf, int size)
...
+		appendStringInfoChar(&str, 0x00);
+		appendStringInfoChar(&str, 0x00);
+		appendStringInfoChar(&str, packet_id++);
+		appendStringInfoChar(&str, 0x00);
+
+		appendBinaryStringInfo(&str, buf + total_written, chunk_size);
+
+		/* Write the current chunk */
+		res = 0;
+		while (res < chunk_size)
```
Deepesh125: This should be res < chunk_size + header size

contrib/babelfishpg_tds/src/backend/tds/tdssecure.c
```diff
@@ -193,56 +193,71 @@ SslWrite(BIO * h, const char *buf, int size)
 static int
 SslHandShakeWrite(BIO * h, const char *buf, int size)
 {
-	StringInfoData str;
+	#define SSL_MAX_PACKET_SIZE 4096
```
surendravishnoi: can we get this macro from place where it is already defined?

contrib/babelfishpg_tds/src/backend/tds/tdssecure.c
```diff
@@ -193,56 +193,71 @@ SslWrite(BIO * h, const char *buf, int size)
...
-	appendStringInfoChar(&str, 0x00);
-
-	appendBinaryStringInfo(&str, buf, size);
-	buf = str.data;
-	size += TDS_PACKET_HEADER_SIZE;
-
-	/* Write the complete data */
-	while (res < size)
+	/* Process data in chunks of SSL_MAX_PACKET_SIZE */
+	while (size > 0)
```
surendravishnoi, line 212: is there use case where size is 0? how do we handle that?

contrib/babelfishpg_tds/src/backend/tds/tdssecure.c
sumitj824: yeah right, fixed it next commit.

contrib/babelfishpg_tds/src/backend/tds/tdssecure.c
sumitj824: We will return 0 when size is 0, the API from library calling this function will raise error based on the return value.

contrib/babelfishpg_tds/src/backend/tds/tdssecure.c
```diff
@@ -193,56 +193,69 @@ SslWrite(BIO * h, const char *buf, int size)
...
+		appendStringInfoChar(&str, packet_id++);
+		appendStringInfoChar(&str, 0x00);
+
+		appendBinaryStringInfo(&str, buf + total_written, chunk_size);
+
+		/* Write the current chunk */
+		res = 0;
+		while (res < (chunk_size + TDS_PACKET_HEADER_SIZE))
+		{
+			int tmp_res = SslWrite(h, str.data + res, (chunk_size + TDS_PACKET_HEADER_SIZE - res));
```
surendravishnoi, line 243: for size = 0, we will write a header without payload, is that expected behavior?

sumitj824: yes, this is expected behavior

surendravishnoi (dismissed):

contrib/babelfishpg_tds/src/backend/tds/tdssecure.c
Deepesh125: As per https://github.com/aws/aws-lc/blob/629fcf2553b4b6eeadd839ffa520f65d99f42fbb/crypto/bio/bio.c#L336. BIO_write(...) will never call the callback to write data of length <= 0

so this should be error condition

surendravishnoi (approved):

kuntalghosh (approved):

PR #3719, opened 2025-04-23
Handling CASE expression when one branch is of decimal/numeric and other branch is of smallmoney/money.

tanyagupta17:
### Description

Currently, we treat typmod for smallmoney/money datatype as default which leads to output differences in many cases.

Following are the fixes introduced in this PR : 

- Introduce typmod handling for smallmoney/money datatype for objects like table/procedure/views/functions/casts and handling their upgrade scenarios.
- We are also handling p&s for declared variables, nested case expressions and unions having smallmoney/money as one of it's node.
- It also takes case of round vs truncation behaviour around edge cases for fixeddecimal multiplication
- Money/smallmoney and Numeric Arithmetic operations - precision and sclae correction and fixing TDS protocol Errors.
- T_Param node handling for fixeddecimal dataypes (money and smallmoney) when numeric is another operand.
- T_Aggref handling for Sum/Avg for fixeddecimal and other fixed length datatypes when numeric is another operand.
- Handing typmod for smallmoney/money in pg catalogs for views metadata and during dump-restore.
- Handling money/smallmoney typmod and fixeddecimal typmodin during dump-restore and by-passing typmod while deparsing for ISC check constraint. 
- Avoiding crashes and wrong results by updating sanity checks and setting intermittent result size appropirately for bigint and fixeddecimal multiplication.

Engine PR : https://github.com/babelfish-for-postgresql/postgresql_modified_for_babelfish/pull/575

### Issues Resolved
BABEL-5512, BABEL-5718, BABEL-5889, BABEL-5744, BABEL-5750, BABEL-5755, BABEL-5758, BABEL-5892, BABEL-5893, BABEL-5894, BABEL-5895, BABEL-5896, BABEL-5897, BABEL-5898

* **Use case based -** Yes

* **Arbitrary inputs -** Yes

* **Performance tests -** NA

* **Tooling impact -** NA

* **Client tests -** NA

### Check List


contrib/babelfishpg_tsql/src/pltsql_coerce.c
```diff
@@ -2144,7 +2148,17 @@ tsql_select_common_typmod_hook(ParseState *pstate, List *exprs, Oid common_type)
...
+
+			/* Handling for money/smallmoney. */
+				if (typmod == -1 && (*common_utility_plugin_ptr->is_tsql_money_datatype)(type))
+			{
+				typmod = ((TDS_MONEY_PRECISION << 16) | TDS_FIXEDDECIMAL_SCALE) + VARHDRSZ;
+			}
+			else if (typmod == -1 && (*common_utility_plugin_ptr->is_tsql_smallmoney_datatype)(type))
+			{
+				typmod = ((TDS_SMALLMONEY_PRECISION << 16) | TDS_FIXEDDECIMAL_SCALE) + VARHDRSZ;
+			}
```
rishabhtanwar29: nit: indentation seems a bit off.

contrib/babelfishpg_tsql/src/pltsql_utils.c
```diff
@@ -41,6 +41,9 @@ bool		suppress_string_truncation_error = false;
 bool		pltsql_suppress_string_truncation_error(void);
 
 
+#define TDS_MONEY_PRECISION 19
+#define TDS_SMALLMONEY_PRECISION 10
+#define TDS_FIXEDDECIMAL_SCALE 4
```
rishabhtanwar29: Duplicate declaration, can we move this to a command header instead?

contrib/babelfishpg_tsql/src/pltsql_coerce.c
```diff
@@ -72,7 +76,7 @@ PG_FUNCTION_INFO_V1(get_immediate_base_type_of_UDT);
 static Oid select_common_type_setop(ParseState *pstate, List *exprs, Node **which_expr, const char *context);
 static Oid select_common_type_for_isnull(ParseState *pstate, List *exprs);
 static Oid select_common_type_for_coalesce_function(ParseState *pstate, List *exprs);
-static Oid get_immediate_base_type_of_UDT_internal(Oid typeid);
+Oid get_immediate_base_type_of_UDT_internal(Oid type_oid);
```
rishabhtanwar29: We can get rid of this completely.

contrib/babelfishpg_tsql/src/pl_comp.c
```diff
@@ -2368,12 +2372,30 @@ pltsql_build_variable(const char *refname, int lineno, PLtsql_type *dtype,
...
+				Oid	base_typeid = InvalidOid;
 
 				var = palloc0(sizeof(PLtsql_var));
 				var->dtype = PLTSQL_DTYPE_VAR;
 				var->refname = pstrdup(refname);
 				var->lineno = lineno;
 				var->datatype = dtype;
+
+				typ_oid = var->datatype->typoid;
+				base_typeid = get_immediate_base_type_of_UDT_internal(typ_oid);
```
rishabhtanwar29: Question: Why are we looking only for immediate basetype and not the actual basetype here?

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
```diff
@@ -527,6 +532,38 @@ is_numeric_datatype(Oid typid)
...
+/* TODO : remove later. */
+static bool
+is_tsql_money_datatype(Oid oid)
+{
+	if (tsql_money_oid == InvalidOid)
+		tsql_money_oid = lookup_tsql_datatype_oid("money");
+	return tsql_money_oid == oid;
+}
+
+/* TODO : remove later after resolve_numeric_typmod_from_exp is moved to tsql */
```
rishabhtanwar29: Who will address this TODO? Is there a tracking ticket?

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--5.2.0--5.3.0.sql
```diff
@@ -191,6 +191,261 @@ WHERE(pg_has_role(sys.suser_id(), 'sysadmin'::TEXT, 'MEMBER')
...
+			FROM pg_type t
+			WHERE t.oid = a.atttypid
+			AND t.typbasetype::regtype::text IN ('sys.money', 'sys.smallmoney')
+		)
+	);
+
+
 -- Please add your SQLs here
 /*
  * Note: These SQL statements may get executed multiple times specially when some features get backpatched.
```
rishabhtanwar29: Oh, so seems like folks have been ignoring this Note and advice for a while which clearly mentions to add the SQL statements starting from here.

contrib/babelfishpg_money/fixeddecimal.c
```diff
@@ -789,14 +789,16 @@ fixeddecimaltypmodin(PG_FUNCTION_ARGS)
 	}
 	else if (n == 1)
 	{
-		if (tl[0] < FIXEDDECIMAL_SCALE || tl[0] > FIXEDDECIMAL_MAX_PRECISION)
+		int val_scale = (tl[0] - VARHDRSZ) & 0xffff;
+		int val_precision = ((tl[0] - VARHDRSZ) >> 16) & 0xffff;
+		if (val_scale < FIXEDDECIMAL_SCALE || (val_precision - val_scale) > FIXEDDECIMAL_MAX_PRECISION)
```
rishabhtanwar29: Can you explain why this change is needed?

contrib/babelfishpg_money/fixeddecimal.c
```diff
@@ -1729,8 +1731,12 @@ fixeddecimalmul(PG_FUNCTION_ARGS)
 	 * we must divide the result by this to get the correct result.
 	 */
 	result = (int128) arg1 * arg2 / FIXEDDECIMAL_MULTIPLIER;
+	/* Round off the result to TDS_DIFXEDECIMAL_SCALE. */
+	if ((((int128) arg1 * arg2 % FIXEDDECIMAL_MULTIPLIER)) >= 5000)
+	result++;
 
-	if (result != ((int64) result))
+
+	if (result != (int128) ((int64) result))
```
rishabhtanwar29: Why is this change needed? Do we have a test case where the earlier condition was giving wrong results?

contrib/babelfishpg_tsql/sql/sys_views.sql
```diff
@@ -169,6 +169,8 @@ BEGIN
 	CASE v_type 
 	WHEN 'decimal' THEN scale = (typemod - 4) & 65535;
 	WHEN 'numeric' THEN scale = (typemod - 4) & 65535;
+	WHEN 'money' THEN scale = (typemod - 4) & 65535;
+	WHEN 'smallmoney' THEN scale = (typemod - 4) & 65535;
```
KushaalShroff: Can we have fixed typmod here?

contrib/babelfishpg_tsql/sql/sys_views.sql
```diff
@@ -256,6 +258,8 @@ BEGIN
 	CASE v_type
 	WHEN 'numeric' THEN precision = ((typemod - 4) >> 16) & 65535;
 	WHEN 'decimal' THEN precision = ((typemod - 4) >> 16) & 65535;
+	WHEN 'money' THEN precision = ((typemod - 4) >> 16) & 65535;
+	WHEN 'smallmoney' THEN precision = ((typemod - 4) >> 16) & 65535;
```
KushaalShroff: Same, lets have fixed typmod here

contrib/babelfishpg_tsql/src/pl_comp.c
```diff
@@ -2368,12 +2372,30 @@ pltsql_build_variable(const char *refname, int lineno, PLtsql_type *dtype,
...
+
+				if (OidIsValid(base_typeid))
+					typ_oid = base_typeid;
+
+				if ((*common_utility_plugin_ptr->is_tsql_money_datatype)(typ_oid))
+				{
+					var->datatype->atttypmod = ((TDS_MONEY_PRECISION << 16) | TDS_FIXEDDECIMAL_SCALE) + VARHDRSZ;
```
KushaalShroff: This can be a macro named as TSQL_MONEY_TYPMOD:
((TDS_MONEY_PRECISION << 16) | TDS_FIXEDDECIMAL_SCALE) + VARHDRSZ

Same for smallmoney

contrib/babelfishpg_tsql/src/pl_comp.c
```diff
@@ -2368,12 +2372,30 @@ pltsql_build_variable(const char *refname, int lineno, PLtsql_type *dtype,
 			{
 				/* Ordinary scalar datatype */
 				PLtsql_var *var;
+				Oid typ_oid;
```
KushaalShroff: Why do we need this variable?

contrib/babelfishpg_tsql/src/pltsql_coerce.c
```diff
@@ -52,6 +52,10 @@
 #define BPCHAR_MAX_TYPMOD 8000
 
 #define TDS_MAX_NUM_PRECISION 38
+#define TDS_MONEY_PRECISION 19
+#define TDS_SMALLMONEY_PRECISION 10
+#define TDS_FIXEDDECIMAL_SCALE 4
```
KushaalShroff: Lets remove TDS Prefix

contrib/babelfishpg_tsql/src/pltsql_coerce.c
```diff
@@ -2144,7 +2148,17 @@ tsql_select_common_typmod_hook(ParseState *pstate, List *exprs, Oid common_type)
...
-			
+
+			/* Handling for money/smallmoney. */
+				if (typmod == -1 && (*common_utility_plugin_ptr->is_tsql_money_datatype)(type))
+			{
+				typmod = ((TDS_MONEY_PRECISION << 16) | TDS_FIXEDDECIMAL_SCALE) + VARHDRSZ;
+			}
+			else if (typmod == -1 && (*common_utility_plugin_ptr->is_tsql_smallmoney_datatype)(type))
+			{
+				typmod = ((TDS_SMALLMONEY_PRECISION << 16) | TDS_FIXEDDECIMAL_SCALE) + VARHDRSZ;
```
KushaalShroff: For money types I think we are doing redundant runtime calculations further to remove the scale and precision later and then again building the typmod to return it

contrib/babelfishpg_tsql/src/pltsql_ruleutils.c
```diff
@@ -2844,8 +2844,18 @@ tsql_format_type_extended(Oid type_oid, int32 typemod, bits16 flags)
...
 		typemod = -1;
+	/*
+	 * Assign -1 as typmod which is equivalent to not printing the typmod for
+	 * smallmoney/money
+	 */
+	if ((*common_utility_plugin_ptr->is_tsql_money_datatype)(type_oid) ||
+		(*common_utility_plugin_ptr->is_tsql_smallmoney_datatype)(type_oid))
+	{
+		typemod = -1;
+	}
```
KushaalShroff: Why cant we set with_typemod = false instead

contrib/babelfishpg_tsql/src/pltsql_utils.c
KushaalShroff: Being declared more than once, lets move it in a common header file. Also lets avoid TDS prefix

contrib/babelfishpg_tsql/src/pltsql_utils.c
```diff
@@ -441,9 +444,32 @@ pltsql_check_or_set_default_typmod_helper(TypeName *typeName, int32 *typmod, boo
 		char	   *schemaname;
 		char	   *typname;
 		bool		is_sys_schema = false;
+		// const char *base_typname;
```
KushaalShroff: Please cleanup any redundant lines of code

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--5.2.0--5.3.0.sql
```diff
@@ -191,6 +191,261 @@ WHERE(pg_has_role(sys.suser_id(), 'sysadmin'::TEXT, 'MEMBER')
...
+	SELECT jsonb_set(
+		p1.probin::jsonb,
+		'{typmod_array}',
+		to_jsonb(
+			(
+				SELECT jsonb_agg(
+					CASE
+						WHEN p2.prokind = 'p' OR (p2.prokind = 'f' AND p2.proargtypes[typ_index-1] IS NOT NULL) THEN
+							CASE
+								WHEN typmod = '-1' AND (p2.proargtypes[typ_index-1]::regtype::text = 'sys.money' OR
```
rishabhtanwar29: Generic comment: should cast like 'sys.money'::regtype and compare OIDs instead of comparing TEXTs.

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--5.2.0--5.3.0.sql
```diff
@@ -191,6 +191,261 @@ WHERE(pg_has_role(sys.suser_id(), 'sysadmin'::TEXT, 'MEMBER')
...
+									-- Handling UDT.
+									EXISTS (SELECT 1 FROM pg_type WHERE oid = p2.proargtypes[typ_index-1] AND
+										typbasetype::regtype::text = 'sys.money')) THEN '1245192'
```
rishabhtanwar29: Shouldn't we be recursively checking for basetype instead of immediate basetype?

contrib/babelfishpg_money/fixeddecimal.c
```diff
@@ -1729,8 +1731,12 @@ fixeddecimalmul(PG_FUNCTION_ARGS)
 	 * we must divide the result by this to get the correct result.
 	 */
 	result = (int128) arg1 * arg2 / FIXEDDECIMAL_MULTIPLIER;
+	/* Round off the result to TDS_DIFXEDECIMAL_SCALE. */
+	if ((((int128) arg1 * arg2 % FIXEDDECIMAL_MULTIPLIER)) >= 5000)
```
KushaalShroff: Why 5000?
Also create a macro for this

contrib/babelfishpg_money/fixeddecimal.c
```diff
@@ -1906,28 +1912,21 @@ fixeddecimalint8mul(PG_FUNCTION_ARGS)
...
-	 *
-	 * Since the division is likely much more expensive than the actual
-	 * multiplication, we'd like to skip it where possible.  The best bang for
-	 * the buck seems to be to check whether both inputs are in the int32
-	 * range; if so, no overflow is possible.
+	 * Overflow check. The best bang for the buck seems to be to check
+	 * whether result in the int32 range; if so, no overflow is possible.
 	 */
-	if (arg1 != (int64) ((int32) arg1) &&
-		result / arg1 != arg2)
```
KushaalShroff, line 1932: Add comment explaining why we dont need this check anymore

KushaalShroff (changes requested):

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--5.2.0--5.3.0.sql
```diff
@@ -191,6 +191,261 @@ WHERE(pg_has_role(sys.suser_id(), 'sysadmin'::TEXT, 'MEMBER')
...
+							END
+						ELSE typmod
+					END
+				)
+				FROM jsonb_array_elements_text(p1.probin::jsonb->'typmod_array') WITH ORDINALITY AS elem(typmod,typ_index)
+			)
+		)
+	)
+)
+FROM pg_proc p2
```
rishabhtanwar29: Why do we need to join with pg_proc again? can't we just use the same result relation of UPDATE statement.

contrib/babelfishpg_tsql/src/pl_comp.c
tanyagupta17: Removed

contrib/babelfishpg_tsql/src/pltsql_ruleutils.c
tanyagupta17: Done for smalldatetime, money, smallmoney.

contrib/babelfishpg_tsql/src/pltsql_utils.c
tanyagupta17: Moved to pltsql.h

tanyagupta17: Moved to pltsql.h

contrib/babelfishpg_tsql/src/pltsql_coerce.c
tanyagupta17: Corrected it

contrib/babelfishpg_tsql/src/pltsql_coerce.c
tanyagupta17: Removed. We no longer need to look at get_immediate_base_type_of_UDT_internal to resolve UDT, as we are keeping the original behavior -1 for it.

contrib/babelfishpg_tsql/src/pl_comp.c
tanyagupta17: We were looking at immediate basetype and not the actual basetype here as money/smallmoney are domains themselves built over fixeddecimal. So if we look at actual basetype then we would land on fixeddecimal and would not have been able to distinguish between money and smallmoney. 
Though, now we are handling UDT like other datatype like decimal, varchar etc. so this handling is not required here.

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
tanyagupta17: This TODO was kept till Yashneet's changes were merged where he has moved `resolve_numeric_typmod_from_exp` from tds to tsql layer. So now we can directly use the common_utility_plugin_ptr for checking money/smoll money types. 

This has been addressed now with this PR itself.

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--5.2.0--5.3.0.sql
tanyagupta17: Moved the upgrade script after this Note now.

contrib/babelfishpg_money/fixeddecimal.c
tanyagupta17: This is a fix as a part of dump-restore issue. Now that we are using `fixeddecimaltypmodin` for money/smallmoney restore, here n==2 represents p&s where n==1 represents typmod. It's a day1 issue, for we were earlier having a sanity check to compare typmod with FIXEDDECIMAL_SCALE and FIXEDDECIMAL_MAX_PRECISION. Whereas we should be extracting p&s from typmod first and then doing the sanity check appropriately.

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--5.2.0--5.3.0.sql
tanyagupta17: Tsql only allows one level for UDT, so this was sufficient as per that. But we no longer need to handle UDT at object level. So removed it now.

contrib/babelfishpg_tsql/src/pltsql_coerce.c
tanyagupta17: Updated now we are setting direct scale and precision and keeping rest of behavior as it is.

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--5.2.0--5.3.0.sql
tanyagupta17: Done

contrib/babelfishpg_money/fixeddecimal.c
```diff
@@ -1908,28 +1914,21 @@ fixeddecimalint8mul(PG_FUNCTION_ARGS)
...
+	if (result != (int128) (int64) ((int32) result))
```
ayushdsh: Since fixeddecimal is a int64 we should only be converting the result till int64.
For example - 
`select cast(922337203685477.5807 as money) * cast(1 as bigint)`

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--5.2.0--5.3.0.sql
tanyagupta17: Updated

contrib/babelfishpg_money/fixeddecimal.c
```diff
@@ -791,14 +791,16 @@ fixeddecimaltypmodin(PG_FUNCTION_ARGS)
 	}
 	else if (n == 1)
 	{
-		if (tl[0] < FIXEDDECIMAL_SCALE || tl[0] > FIXEDDECIMAL_MAX_PRECISION)
+		int val_scale = (tl[0] - VARHDRSZ) & 0xffff;
+		int val_precision = ((tl[0] - VARHDRSZ) >> 16) & 0xffff;
+		if (val_scale < FIXEDDECIMAL_SCALE || (val_precision - val_scale) > FIXEDDECIMAL_MAX_PRECISION)
```
rohit01010: Here the condition should be 
```
if (val_precision < FIXEDDECIMAL_SCALE || val_precision > FIXEDDECIMAL_MAX_PRECISION)
```

contrib/babelfishpg_money/fixeddecimal.c
```diff
@@ -791,14 +791,16 @@ fixeddecimaltypmodin(PG_FUNCTION_ARGS)
...
 			ereport(ERROR,
 					(errcode(ERRCODE_INVALID_PARAMETER_VALUE),
 					 errmsg("FIXEDDECIMAL precision %d must be between %d and %d",
 							tl[0], FIXEDDECIMAL_SCALE, FIXEDDECIMAL_MAX_PRECISION)));
```
rohit01010: we should have similar check on val_scale, to validate if val_scale is equals to FIXEDDECIMAL_SCALE or not. And throw error if not equals.

contrib/babelfishpg_money/fixeddecimal.c
```diff
@@ -791,14 +791,16 @@ fixeddecimaltypmodin(PG_FUNCTION_ARGS)
...
 
 		/* scale defaults to FIXEDDECIMAL_SCALE */
-		typmod = ((tl[0] << 16) | FIXEDDECIMAL_SCALE) + VARHDRSZ;
+		typmod = ((val_precision << 16) | FIXEDDECIMAL_SCALE) + VARHDRSZ;
```
rohit01010: why are we using `FIXEDDECIMAL_SCALE`, we should use `val_scale` here

contrib/babelfishpg_money/fixeddecimal.c
tanyagupta17: The condition should be like 
`if (val_precision < FIXEDDECIMAL_SCALE || (val_precision - val_scale) > FIXEDDECIMAL_MAX_PRECISION)` 
As FIXEDDECIMAL_MAX_PRECISION represents 15 and FIXEDDECIMAL_SCALE as 4. So the precision should be > 4 but the integral part (val_precision - val_scale) should be < 15

contrib/babelfishpg_money/fixeddecimal.c
tanyagupta17: Added the check for val_scale

contrib/babelfishpg_money/fixeddecimal.c
tanyagupta17: Updated to val_scale

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--5.2.0--5.3.0.sql
```diff
@@ -208,6 +208,254 @@ RETURNS INTEGER AS
...
+-- To handle the typmod value of UDT created over money/smallmoney.
+-- This would be required in cases where we create a new table using this earlier UDT or directly using this UDT.
+UPDATE pg_type t
+SET typtypmod =
+	CASE
+		WHEN t.typtypmod = '-1' AND t.typtype = 'd' AND
+			sys.bbf_get_immediate_base_type_of_UDT(t.oid) = (SELECT 'sys.money'::regtype::oid)
+		THEN 1245192
+		WHEN t.typtypmod = '-1' AND t.typtype = 'd' AND
+			sys.bbf_get_immediate_base_type_of_UDT(t.oid) = (SELECT 'sys.smallmoney'::regtype::oid)
```
rohit01010: we are already filtering the rows based on `t.typtypmod = -1 AND t.typtype = 'd';`, why do we again want to add these checks here ?

contrib/babelfishpg_tsql/src/pl_comp.c
```diff
@@ -2374,6 +2374,16 @@ pltsql_build_variable(const char *refname, int lineno, PLtsql_type *dtype,
...
 				var->datatype = dtype;
+
+				if ((*common_utility_plugin_ptr->is_tsql_money_datatype)(var->datatype->typoid))
+				{
+					var->datatype->atttypmod = TSQL_MONEY_TYPMOD;
+				}
+				else if ((*common_utility_plugin_ptr->is_tsql_smallmoney_datatype)(var->datatype->typoid))
+				{
+					var->datatype->atttypmod = TSQL_SMALLMONEY_TYPMOD;
+				}
```
rohit01010, line 2385: do we also need to add handling for UDT on money or smallmoney?

rohit01010: Do we need similar handling for other type of variables?

contrib/babelfishpg_tsql/src/pltsql_coerce.c
```diff
@@ -1222,6 +1223,36 @@ resolve_numeric_typmod_from_exp(Plan *plan, Node *expr, bool *found)
...
+
+					/* handling for fixed length datatypes */
+					if (param->paramtype == INT4OID)
+						return ((INT_PRECISION_RADIX << 16) | 0) + VARHDRSZ;
+					else if (param->paramtype == INT8OID)
+						return ((BIGINT_PRECISION_RADIX << 16) | 0) + VARHDRSZ;
+					else if (param->paramtype == INT2OID)
+						return ((SMALLINT_PRECISION_RADIX << 16) | 0) + VARHDRSZ;
+
+					if (found != NULL) *found = false;
```
rohit01010: why we need this ?

contrib/babelfishpg_tsql/src/pltsql_coerce.c
```diff
@@ -1781,6 +1812,25 @@ resolve_numeric_typmod_from_exp(Plan *plan, Node *expr, bool *found)
 					precision = tds_default_numeric_precision;
 				}
 
+				if (aggFuncName && strlen(aggFuncName) == 3 &&
+					((strncmp(aggFuncName, "sum", 3) == 0) ||
+					(strncmp(aggFuncName, "avg", 3) == 0)))
+				{
+					/* Handling for fixed length datatype. */
+					if (typmod != -1 && ((*common_utility_plugin_ptr->is_tsql_money_datatype)(aggref->aggtype)))
```
rohit01010: why we are having this check of typmod != -1 ? Also do we need similar check for small money or not ?

contrib/babelfishpg_tsql/src/pltsql_coerce.c
```diff
@@ -2973,15 +3030,28 @@ tsql_select_common_typmod_hook(ParseState *pstate, List *exprs, Oid common_type)
...
+
+			/* Handling for money/smallmoney. */
+			if (typmod == -1 && (*common_utility_plugin_ptr->is_tsql_money_datatype)(type))
+			{
+				precision = MONEY_PRECISION;
+				scale = FIXEDDECIMAL_SCALE;
+			}
+			else if (typmod == -1 && (*common_utility_plugin_ptr->is_tsql_smallmoney_datatype)(type))
+			{
+            	precision = SMALLMONEY_PRECISION;
```
rohit01010: NIT: Indentation

contrib/babelfishpg_money/fixeddecimal.c
rohit01010: Here result is the multiplication of two int64 values, so the result can exceed int64 hence we should store the result into an int128 variable and add a check whether lies in int64 range or not. I think this is what is done here, except the condition should have been `result != (int128) ((int64) result)`.

contrib/babelfishpg_tsql/src/pltsql_coerce.c
```diff
@@ -1222,6 +1223,36 @@ resolve_numeric_typmod_from_exp(Plan *plan, Node *expr, bool *found)
 		case T_Param:
 			{
 				Param *param = (Param *) expr;
+				if (param->paramtypmod != -1 &&
+					((*common_utility_plugin_ptr->is_tsql_money_datatype)(param->paramtype) ||
+					(*common_utility_plugin_ptr->is_tsql_smallmoney_datatype)(param->paramtype)))
+				{
+					return param->paramtypmod;
+				}
```
rohit01010: can we move this to the bottom where we are returning `param->paramtypmod` when expr type is numeric ?

contrib/babelfishpg_money/fixeddecimal.c
tanyagupta17: We were earlier just calculating arg1*arg2 and truncating it to 4 digits after decimal using FIXEDDECIMAL_MULTIPLIER.
Now, we check if the remainder after modulus with 10000 gives a number >=5000 that means the 5th decimal place is 5 and above, so we would need to round up the 4th decimal place in our result and return that.
Example 
`select cast(2.5 as smallmoney) * cast(2.4999 as smallmoney)`
Here arg1 = 25000, arg2 = 24999 and result = 624975000 
result % 10000 = 5000. So we need to round up the result to be 6.2498

tanyagupta17: Created a Marco for this

contrib/babelfishpg_tsql/src/pltsql_coerce.c
tanyagupta17: Removed.

contrib/babelfishpg_tsql/src/pltsql_coerce.c
tanyagupta17: Moved.

contrib/babelfishpg_tsql/src/pltsql_coerce.c
tanyagupta17: We will not need a seprate handling for smallmoney as in case of both smallmoney and money we will get `aggref->aggtype` as money. Added comment for same.
Removed check typmod != -1.

contrib/babelfishpg_tsql/src/pltsql_coerce.c
tanyagupta17: Corrected.

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--5.2.0--5.3.0.sql
tanyagupta17: removed

contrib/babelfishpg_tsql/src/pl_comp.c
tanyagupta17: We don't need handling for UDT here it would be taken care in coerce_to_domain using `getBaseTypeAndTypmod` Just like other datatypes UDT like varchar, decimal etc. 

We also don't need handling for other types of variables like table variable etc, as they will be handled by `check_or_set_default_typmod_hook`.

contrib/babelfishpg_money/fixeddecimal.c
tanyagupta17: Just checking if arg1 is convertible to int32 is not enough, as the result could still overflow. Hence we should be rather checking if the whole result itself is convertible to int64. That would confirm that no overflow would happen.

Example :
`SELECT CAST(1.0001 AS SMALLMONEY) * CAST(9223372036854775807 AS BIGINT)`
`SELECT CAST(1.0001 AS money) * CAST(9223372036854775807 AS BIGINT)`

Without this fix the first query is crashing and the second one is giving wrong output. 
It's expected for both of these to throw overflow error.

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--5.2.0--5.3.0.sql
```diff
@@ -208,6 +208,262 @@ RETURNS INTEGER AS
...
+UPDATE pg_proc p
+SET probin = (
+	SELECT jsonb_set(
+		p.probin::jsonb,
+		'{typmod_array}',
+		to_jsonb(
+			(
+				SELECT jsonb_agg(
+					CASE
+						WHEN p.prokind = 'p' OR (p.prokind = 'f' AND p.proargtypes[typ_index-1] IS NOT NULL) THEN
```
rishabhtanwar29, line 398: Why do we check nullability of argument type for functions and not for the procedures?

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--5.2.0--5.3.0.sql
```diff
@@ -208,6 +208,262 @@ RETURNS INTEGER AS
...
+				FROM jsonb_array_elements_text(p.probin::jsonb->'typmod_array') WITH ORDINALITY AS elem(typmod,typ_index)
+			)
+		)
+	)
+)
+FROM sys.babelfish_namespace_ext sch, pg_language l
+WHERE sch.nspname = p.pronamespace::regnamespace::name
+	AND p.prolang = l.oid
+	AND l.lanname = 'pltsql'
+	AND ((p.prokind = 'p' AND p.proargtypes <> '') OR (p.prokind = 'f' AND p.proallargtypes IS NULL));
```
rishabhtanwar29, line 422: Why `proallargtypes` should be null? Shouldn't we be fixing the occurrences of money/smallmoney in `proallargtypes` as well?

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--5.2.0--5.3.0.sql
```diff
@@ -208,6 +208,262 @@ RETURNS INTEGER AS
...
+	END
+FROM pg_class c
+INNER JOIN sys.babelfish_namespace_ext sch
+	ON sch.nspname = c.relnamespace::regnamespace::name
+WHERE a.attrelid = c.oid
+	AND a.atttypmod = -1
+	AND NOT a.attisdropped
+	AND (a.atttypid = 'sys.money'::regtype::oid OR
+		a.atttypid = 'sys.smallmoney'::regtype::oid)
+	AND c.relkind IN ('r', 'i', 'v', 'p', 'I');
```
rishabhtanwar29, line 447: Why not composite types, foreign tables or materialized views?

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--5.2.0--5.3.0.sql
```diff
@@ -208,6 +208,262 @@ RETURNS INTEGER AS
...
+
+/*
+ * Updates typmod values for UDTs based on money/smallmoney types in babelfish_namespace_ext schemas.
+ * Required when creating new tables or using these UDTs directly to ensure proper type handling.
+ */
+UPDATE pg_type t
+SET typtypmod =
+	CASE
+		WHEN sys.bbf_get_immediate_base_type_of_UDT(t.oid) = (SELECT 'sys.money'::regtype::oid)
```
rishabhtanwar29: Why to create a subquery instead of a direct cast like this?
```
WHEN sys.bbf_get_immediate_base_type_of_UDT(t.oid) = 'sys.money'::regtype::oid
```

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--5.2.0--5.3.0.sql
```diff
@@ -208,6 +208,262 @@ RETURNS INTEGER AS
...
+	SELECT jsonb_set(
+		p.probin::jsonb,
+		'{typmod_array}',
+		to_jsonb(
+			(
+				SELECT jsonb_agg(
+					CASE
+						WHEN p.prokind = 'p' OR (p.prokind = 'f' AND p.proargtypes[typ_index-1] IS NOT NULL) THEN
+							CASE
+								WHEN typmod = '-1' AND (p.proargtypes[typ_index-1] = 'sys.money'::regtype::oid) THEN '1245192'
```
rishabhtanwar29, line 400: Why are we no longer checking for domains here?

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--5.2.0--5.3.0.sql
```diff
@@ -208,6 +208,262 @@ RETURNS INTEGER AS
...
+/*
+ * Updates typmod values in pg_attribute for smallmoney/money columns
+ * in r = ordinary table, i = index, v = view, p = partitioned table, I = partitioned index.
+ * For other relkinds, we either don't create from TDS side or they don't support money/smallmoney.
+ * Sets money typmod to 1245192 and smallmoney to 655368 where. typmod is -1
+ */
+UPDATE pg_attribute a
+SET atttypmod =
+	CASE
+		WHEN atttypmod = '-1' AND (a.atttypid = 'sys.money'::regtype::oid)
```
rishabhtanwar29, line 433: Why are we no longer checking for domains here?

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--5.2.0--5.3.0.sql
tanyagupta17: Updated

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--5.2.0--5.3.0.sql
tanyagupta17: As discussed offline, in case of function there is a possibility that there is no input argument and hence proargtypes will be NULL. We map typmods in typmod_array by proargtypes first then proretype, so to make sure that we actually have an input argument and the corresponding typmod mapping is correct, we have added this check. 

In case of procedure we will always have proargtypes, and in case we don't have proargtypes for below case 
`CREATE PROCEDURE babel_5512_upgrade_p7 AS (SELECT cast(1234 as money));`
The typmod_array will be [] which shouldn't be updated, that we handling in the where clause filter.

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--5.2.0--5.3.0.sql
tanyagupta17: As discussed offline, for ordinary user defined function we stores values in proargtypes and prorettype only. For ITVF and MSTVF we store values in `proallargtypes`. We don't want to change ITVF and MSTVF functions typmod as they are handled in a different manner(expect -1 both before and after), so we have excluded those using this check.

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--5.2.0--5.3.0.sql
tanyagupta17: Current TSQL supported syntax  for composite types are handed as ordinary tables itself which is already being taken care off here. 
Rest foreign tables and materialized views can't be created from TSQL endpoint. And sequence - does not support smallmoney/money.

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--5.2.0--5.3.0.sql
tanyagupta17: As discussed offline, for UDT being created and used as proc/func arg or table column etc. we don't store there typmod in catalogs and rather handle it at runtime using `getBaseTypeAndTypmod` like other datatypes like varchar, decimal etc.

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--5.2.0--5.3.0.sql
tanyagupta17: Discussed offline, same reasoning as above

rohit01010 (approved):

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -6478,3 +6492,31 @@ pltsql_validateCachedPlanSearchPath(SPIPlanPtr plan)
...
+ * and UDTs created on them, in case of dump_restore.
+ */
+static void
+handle_basetype_typmodin(Type typ, Oid *typmodin)
+{
+	Oid 		typbasetype;
+	Oid 		basetypeid;
+	Oid 		typeoid;
+	HeapTuple	tup;
+	const char *dump_restore = GetConfigOption("babelfishpg_tsql.dump_restore", true, false);
```
rishabhtanwar29: Can directly use `babelfish_dump_restore` variable.

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -6478,3 +6492,31 @@ pltsql_validateCachedPlanSearchPath(SPIPlanPtr plan)
 	return true;
 	
 }
+
+/*
+ * Return the typmodin for domains like smallmoney/money
+ * and UDTs created on them, in case of dump_restore.
+ */
```
rishabhtanwar29: Should say something like: Returns the oid of typmodin function for a given domain tuple which is essentially same as the oid of it's basetype's typmodin function. This is only used during restore to handle domains like smallmoney/money and UDTs created on them.

rishabhtanwar29 (dismissed):

rohit01010 (approved):

kuntalghosh (dismissed):

contrib/babelfishpg_tsql/src/pltsql_coerce.c
```diff
@@ -1801,14 +1801,36 @@ resolve_numeric_typmod_from_exp(Plan *plan, Node *expr, bool *found)
 					precision = tds_default_numeric_precision;
 				}
 
+				if (aggFuncName && strlen(aggFuncName) == 3 &&
+					((strncmp(aggFuncName, "sum", 3) == 0) ||
+					(strncmp(aggFuncName, "avg", 3) == 0)))
+				{
+					/* Handling for fixed length datatype. */
+					if (typmod != -1 && ((*common_utility_plugin_ptr->is_tsql_money_datatype)(aggref->aggtype)))
```
KushaalShroff: Are we sure we do not need to add small money? There is scope of being defensive by adding is_tsql_smallmoney_datatype check even if currently we think it is redundant

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--5.2.0--5.3.0.sql
```diff
@@ -208,6 +208,262 @@ RETURNS INTEGER AS
...
+	END CASE;
+	RETURN scale;
+END;
+$$ LANGUAGE plpgsql IMMUTABLE STRICT;
+
+/*
+ * Updates typmod values in pg_proc for smallmoney/money data types in
+ * PLTSQL procedures/functions, defined in babelfish_namespace_ext schemas.
+ * Sets money typmod to 1245192(19,4) and smallmoney to 655368(10,4) where typmod is -1.
+ */
```
KushaalShroff, line 388: Is there any risk for performance? Lets think if an existing customer is mvu'ing to 5.3.0 and have a lot of tables/procedures -- for tables we are scanning entire pg_attribute and pg_class and filtering for our use case. Will the mvu take more time for this? Does mvu have a time out -- I am guessing it should be same as statement timeout since we will run alter extension upgrade.

contrib/babelfishpg_tsql/src/pltsql_coerce.c
```diff
@@ -1781,6 +1806,31 @@ resolve_numeric_typmod_from_exp(Plan *plan, Node *expr, bool *found)
 					precision = tds_default_numeric_precision;
 				}
 
+				if (aggFuncName && strlen(aggFuncName) == 3 &&
```
KushaalShroff, line 1809: Why checking strlen and then doing strncmp again?

contrib/babelfishpg_tsql/src/pltsql_coerce.c
```diff
@@ -2973,15 +3024,28 @@ tsql_select_common_typmod_hook(ParseState *pstate, List *exprs, Oid common_type)
...
-				typmod = resolve_numeric_typmod_from_exp(NULL, expr, NULL);
-			
-			if (typmod == -1 || !is_tsql_exact_numeric_type(type))
-				continue;
-			
-			scale = (typmod - VARHDRSZ) & 0xffff;
-			precision = ((typmod - VARHDRSZ) >> 16) & 0xffff;
+
+			/* Handling for money/smallmoney. */
+			if (typmod == -1 && (*common_utility_plugin_ptr->is_tsql_money_datatype)(type))
```
KushaalShroff, line 3030: Can we add typmod!=-1 at the top so that we can avoid looking up types incase of existing typmod

test/JDBC/input/numeric_money-vu-verify.sql
```diff
@@ -0,0 +1,1154 @@
...
+
+select 21474836 * CAST(1 AS MONEY)
+GO
+
+-- fixeddecimal multiplication
+select cast(922337203685477.5807 as money) * cast(922337203685477.5807 as money)
+GO
+
+-- crashing
+-- select cast(214748.3647 as smallmoney) * cast(214748.3647 as smallmoney)
```
KushaalShroff: Regression or even before your change?

KushaalShroff (changes requested):

test/JDBC/input/numeric_money-vu-verify.sql
tanyagupta17: Not a regression, it was crashing even before my fix. This is due to `fixeddecimalmul` returns the result as int64 but we are later not differentiating between money and smallmoney, so it crashes later on. It might get fixed with @ayushdsh changes. PR : https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/3797

contrib/babelfishpg_tsql/src/pltsql_coerce.c
tanyagupta17: Though it would be redundant as we will get `aggref->aggtype` as money in case of both smallmoney and money arguments in sum/avg function. Still added it to be more defensive

contrib/babelfishpg_tsql/src/pltsql_coerce.c
tanyagupta17: We check the strlen to be exactly 3 before strncmp to match the string exactly and avoid doing any substring matchings for sum/avg.

contrib/babelfishpg_tsql/src/pltsql_coerce.c
tanyagupta17: As we have added the `typmod == -1` at the start of the condition, if that is not met, we will break out of that the `if condition` directly, without looking up types. And directly go to the `else statement` and calculate p&s when `typmod!=-1`.

KushaalShroff (approved):

PR #3720, opened 2025-04-23
Fix invalid read in tsql_round_var function

sharathbp:
### Description

To detect for overflow and underflow we need to determine total integral digits which in-turn uses leading digits in the digits array. But digits array will be empty in case of 0 numeric values which leads to invalid read. 
When the rscale is negative and numeric value is 0 we can skip checking for numeric overflow as the base-NBASE digits will be empty for 0.

### Issues Resolved
BABEL-5546

* **Use case based -**
Yes

* **Boundary conditions -**
Yes

* **Arbitrary inputs -**
Yes

* **Negative test cases -**
Yes

NA

NA

* **Performance tests -**
NA

* **Tooling impact -**
NA

* **Client tests -**
NA

### Check List

### DIFF
--- a/test/JDBC/expected/round_return_type_test-vu-verify.out
+++ b/test/JDBC/expected/round_return_type_test-vu-verify.out
@@ -449,4 +449,11 @@ go
 ~~ERROR (Message: The function round is found but cannot be used. Possibly due to datatype mismatch and implicit casting is not allowed.)~~
 
 
+select round(0.0, -1)
+go
+~~START~~
+numeric
+0
+~~END~~
+
 
--- a/contrib/babelfishpg_common/src/numeric.c
+++ b/contrib/babelfishpg_common/src/numeric.c
@@ -804,7 +804,7 @@ tsql_round_var(NumericVar *var, int rscale)
 	di = (var->weight + 1) * DEC_DIGITS + rscale;
 
 	/* checking numeric overflow for TSQL */
-	if (rscale < 0)
+	if (rscale < 0 && var->ndigits > 0)
 	{
 		int			integral_digits = di - DEC_DIGITS;
 		int			leading_digits = digits[0];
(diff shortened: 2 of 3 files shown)


test/JDBC/expected/round_return_type_test-vu-verify.out
```diff
@@ -449,4 +449,11 @@ go
 ~~ERROR (Message: The function round is found but cannot be used. Possibly due to datatype mismatch and implicit casting is not allowed.)~~
 
 
+select round(0.0, -1)
+go
+~~START~~
+numeric
+0
```
tanscorpio7, line 456: Is this also what SQL Server outputs ?

tanscorpio7 (changes requested):

test/JDBC/expected/round_return_type_test-vu-verify.out
sharathbp: In sqlserver it includes decimal

```
1> select round(0.0, -1)
2> go

---
0.0
```
But currently in babelfish it doesn't show decimal with/without this change as well

tanscorpio7: Can we figure out what the RCA is for this issue. I am just trying to confirm if this needs to be handled in `tsql_round_var`

tanscorpio7 (changes requested):

test/JDBC/expected/round_return_type_test-vu-verify.out
sharathbp: They are forcing it to 0 here 
https://github.com/babelfish-for-postgresql/babelfish_extensions/blob/BABEL_5_X_DEV/contrib/babelfishpg_common/src/numeric.c#L672
So we have to handle it here I think

tanscorpio7 (approved):

test/JDBC/expected/round_return_type_test-vu-verify.out
Yvinayak07: We have jira for correct scale from round function BABEL-5666.
Maybe this will be handled in `resolve_numeric_typmod_from_exp`.

Yvinayak07: @sharathbp could we add this query in the comments of that jira so we don't miss while fixing that change.

Yvinayak07 (approved):

shardgupta (approved):

PR #3725, opened 2025-04-28
Allow parallelism for pltsql return expression after community commit 556f7b7

Deepesh125:
### Description

With the commit 556f7b7, we unintentionally blocked parallelism to
evaluate plpgsql return expression because maxtuples = 2 is being passed
to exec_run_select(...) from exec_eval_expr(...) to evaluate the return
expression of the plpgsql function.

Idea to fix this issue to pass maxtuples = 0. It is safe to do it
because number of processed rows is anyway being checked later in
exec_eval_expr(...). But with this, there is not real caller remained
which calls exec_run_select(...) with maxtuple != 0 so updated definition
of exec_run_select(...) to remove maxtuples argument.

This commit also fixes parallel context leak in case of statement terminating 
error originating from PLTSQL function body. This is important because 
subsequent resetting search_path or table variable cleanup operations may
end up error.

Task: BABEL-5599
Signed-off-by: Dipesh Dhameliya <dipeshdhameliya125@gmail.com>

### Check List


contrib/babelfishpg_tsql/src/iterative_exec.c
```diff
@@ -1397,6 +1397,15 @@ dispatch_stmt_handle_error(PLtsql_execstate *estate,
 		/* In case of RO functions we want the caller to handle errors. */
 		if (ro_func)
 		{
+			if (IsInParallelMode())
```
tanscorpio7, line 1400: How did this cleanup happen today for cases like `SELECT pltsql_func();` & pltsql_func is executing a query in its body that goes parallel.

contrib/babelfishpg_tsql/src/iterative_exec.c
```diff
@@ -1397,6 +1397,15 @@ dispatch_stmt_handle_error(PLtsql_execstate *estate,
...
+			{
+				/*
+				 * If this subxact has started any unfinished parallel operation, clean up
+				 * its workers and exit parallel mode.  Don't warn about leaked resources.
+				 */
+				AtEOSubXact_Parallel(false, GetCurrentSubTransactionId());
+				ExitParallelMode();
```
tanscorpio7: Is it guaranteed we will be inside sub transaction and not transaction ? I do not think a simple SELECT f(); starts a transaction block.

tanscorpio7 (changes requested):

contrib/babelfishpg_tsql/src/iterative_exec.c
Deepesh125: Based on missing handling for parallel context, it seems like no stmt inside pltsql body is going parallel currently.

But I can only think of select stmt as part of RETURN expression that can go parallel. Do you have anything specific in mind?

contrib/babelfishpg_tsql/src/iterative_exec.c
Deepesh125: What all we need is SubTransactionId in order to cleanup parallel context. You can see implementation of `AtEOSubXact_Parallel`. And even if we are inside explicit transaction then SubTransactionId would be `TopSubTransactionId` aka 1 which will be handled with this code too.

And there is possibility of calling `AtEOSubXact_Parallel` through `AbortSubTransaction` again but this function will become no-op during second call because its already cleaned.

surendravishnoi: why this special handling is required? we are throwing error anyway below, why is that not sufficient? also, why to use sub transaction here and not transaction itself, is there a use case where we want to cleanup just the sub txn but not whole txn?

surendravishnoi (approved):

PR #3729, opened 2025-04-28
Adjust precision and scale in Numeric Multiplication and Division

Yvinayak07:
### 1. Issues

In Babelfish's implementation of numeric operations (multiplication and division), logic for adjustment of precision and scale (if precision is greater then NUMERIC_MAX_PRECISION (38)) is incorrect as compared to TSQL.

```
-- Babelfish 
1> select cast(1.289473 as numeric(38, 6)) * 100
2> go
                                       
----------------------------------------
                                  128.95
 
(1 rows affected)
 
-- Expected behaviour (TSQL) 
1> select cast(1.289473 as numeric(38, 6)) * 100
2> go
                                       
----------------------------------------
                              128.947300
 
(1 rows affected)
```

```
-- bbf 
1> DECLARE @o NUMERIC(38,0) = 9999999999999999999999999999999999999; DECLARE @p NUMERIC(38,0)= 3; SELECT @o / @p AS result; 
3> go 
result 
Sqlcmd: Error: Internal error at ReadAndHandleColumnData (Reason: Error reading column data). SqlState HY000, Protocol error in TDS stream
 
-- TSQL 
1> DECLARE @o NUMERIC(38,0) = 9999999999999999999999999999999999999; DECLARE @p NUMERIC(38,0)= 3; SELECT @o / @p AS result; 
3> go 
Msg 8115, Level 16, State 2, Server EC2AMAZ-T4MF7PS, Line 1 Arithmetic overflow error converting expression to data type numeric.

```

- The numeric operations (multiplication and division) logic for precision and scale adjustment needs to be corrected. To prevent TDS protocol violations, errors should be thrown in the execution stage (before setting metadata) instead of the TDS stage.
- `tsql_numeric_get_typmod` doesn’t handle edge cases wisely, fixed the API to give correct typmod for a numeric value.
- Could not calculate typmod from `T_var` node in `resolve_numeric_typmod_from_exp` unable to calculate typmod for fixed length datatypes and UDT's. added handling to calculate typmod from fixed length datatypes and UDT's from `T_Var` node.
```
-- BBF
1> create table testTbl(size int) insert into testTbl values (707072) 
2> go 
(1 rows affected) 
1> select size * 8.00 from testTbl 
2> go 
--------- 
Sqlcmd: Error: Internal error at ReadAndHandleColumnData (Reason: Error reading column data). SqlState HY000, Protocol error in TDS stream

-- TSQL
1> select size * 8.00 from testTbl
2> go
                
----------------
      5656576.00

(1 rows affected)

```
- Failed in parallel query mode as worker node couldn’t load pltsql_protocol_plugin_ptr , So moved resolve_numeric_typmod_from_exp API from TDS side to TSQL.

We have added `sys.dm_os_sys_info` in both `contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--5.1.0--5.2.0.sql` and `contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--5.2.0--5.3.0.sql`. Will remove it from `contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--5.2.0--5.3.0.sql` once it is merged in stable branch.

### Issues Resolved

Task: BABEL-5689

* **Use case based -**

* **Boundary conditions -**

* **Arbitrary inputs -**

* **Negative test cases -**

* **Performance tests -**

* **Tooling impact -**

* **Client tests -**

### Check List


contrib/babelfishpg_tsql/sql/sys_views.sql
```diff
@@ -3680,7 +3680,7 @@ GRANT SELECT ON sys.sequences TO PUBLIC;
 CREATE OR REPLACE VIEW sys.dm_os_sys_info 
 AS SELECT 
   CAST(0 AS BIGINT) AS cpu_ticks,
-  CAST(ROUND(EXTRACT(EPOCH FROM NOW()) * 1000.0, 0) AS BIGINT) AS ms_ticks, 
+  CAST(ROUND(CAST(EXTRACT(EPOCH FROM NOW()) AS NUMERIC(38,0)) * 1000.0, 0) AS BIGINT) AS ms_ticks,
```
rohit01010, line 3683: why we need this change ?

contrib/babelfishpg_common/src/numeric.c
```diff
@@ -1010,7 +1010,7 @@ tsql_numeric_get_typmod(Numeric num)
 	int32_t		weight = NUMERIC_WEIGHT(num);
 	int32_t		precision;
 
-	if (weight >= 0)
+	if (weight > 0)
```
rohit01010: This condition does not seem to be enough, weight can be zero in case of example like `123.112`, so in this case according to new logic the precision = scale = 3 which is incorrect.

contrib/babelfishpg_tsql/sql/sys_views.sql
Yvinayak07: We don't have handling for calculating correct typmod from `extract` function as it returns `numeric` in our case. we already have jira to fix other mathematical functions.

contrib/babelfishpg_common/src/numeric.c
Yvinayak07: i have added the logic for handling edge cases.

rohit01010: Updated logic looks good to me. Do we need similar changes in `numeric_get_typmod` as well ?

Also currently we have two functions `tsql_numeric_get_typmod` and `numeric_get_typmod`, both of them seems to be doing the same thing, can we have single function to improve maintainability ?

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -1282,15 +1288,30 @@ adjust_numeric_result(Plan *plan, Node *expr, Datum result, bool result_isnull,
...
+				result = DirectFunctionCall2(numeric_round, result, Int32GetDatum(scale));
+			
+			result_numeric_val = DatumGetNumeric(result);
+			val_typmod = (*common_utility_plugin_ptr->tsql_numeric_get_typmod) (result_numeric_val);
+
+			if (val_typmod != -1)
+			{
+				val_scale = (val_typmod - VARHDRSZ) & 0xffff;
+				val_precision = ((val_typmod - VARHDRSZ) >> 16) & 0xffff;
+			}
```
rohit01010: when can val_typmod be `-1` ?

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -6387,12 +6408,9 @@ pltsql_exprTypmod(Plan *plan, Node *expr)
...
 		 */ 
-		if (*pltsql_protocol_plugin_ptr && (*pltsql_protocol_plugin_ptr)->get_numeric_typmod_from_exp)
-		{
-			result_typmod = (*pltsql_protocol_plugin_ptr)->get_numeric_typmod_from_exp(plan, expr, &found_typmod);
-			if (!found_typmod)
-				return -1;
-		}
+		result_typmod = resolve_numeric_typmod_from_exp(plan, expr, &found_typmod);
+		if (!plan && !found_typmod)
+			return -1;
```
rohit01010, line 6416: Why do we need this extra check `!plan` ?

contrib/babelfishpg_common/src/numeric.c
Yvinayak07: yes, `numeric_get_typmod` looks similar to `tsql_numeric_get_typmod`. In future we can make a single function to   improve maintainability. we can take this outside of this PR as we will need plugin pointer and use it everywhere. Will create a jira for this.

contrib/babelfishpg_tsql/src/hooks.c
Yvinayak07: we introduced this `found_typmod` check for execution stage (parsing stage). So we will need this `plan` check before `found_typmod` so we don't need to depend on `found_typmod` check if we have `plan` i.e. we are not in parsing stage.

contrib/babelfishpg_tsql/src/hooks.c
Yvinayak07: `val_typmod` will be calculated from the numeirc value itself so yes it seems to be a extra check. Will remove this check.

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -1282,15 +1288,27 @@ adjust_numeric_result(Plan *plan, Node *expr, Datum result, bool result_isnull,
...
+			val_typmod = (*common_utility_plugin_ptr->tsql_numeric_get_typmod) (result_numeric_val);
+
+			val_scale = (val_typmod - VARHDRSZ) & 0xffff;
+			val_precision = ((val_typmod - VARHDRSZ) >> 16) & 0xffff;
+			
+			target_precision = val_precision + (scale - val_scale);
+			if (target_precision > TDS_NUMERIC_MAX_PRECISION ||
+				target_precision > precision)
+				ereport(ERROR, (errcode(ERRCODE_NUMERIC_VALUE_OUT_OF_RANGE),
+								errmsg("Arithmetic overflow error for data type numeric.")));
```
KushaalShroff, line 1317: Need comment explaining why we have moved this logic higher in the code flow

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -1266,7 +1266,13 @@ contains_truncation_functions_checker(Oid func_id, void *context)
 static Datum
 adjust_numeric_result(Plan *plan, Node *expr, Datum result, bool result_isnull, Oid result_type, int32 result_typmod)
 {
-	int32       scale;
+	int32		scale,
+			precision,
+			val_scale = 0,
+			val_precision = 0,
+			val_typmod = -1,
+			target_precision = 0;
```
KushaalShroff: Some of these variables are not needed to declared in function scope, lets move it in lower blocks

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -1282,15 +1288,27 @@ adjust_numeric_result(Plan *plan, Node *expr, Datum result, bool result_isnull,
...
-				return DirectFunctionCall2(numeric_round, result, Int32GetDatum(scale));
+				result = DirectFunctionCall2(numeric_round, result, Int32GetDatum(scale));
+			
+			result_numeric_val = DatumGetNumeric(result);
+			val_typmod = (*common_utility_plugin_ptr->tsql_numeric_get_typmod) (result_numeric_val);
+
+			val_scale = (val_typmod - VARHDRSZ) & 0xffff;
+			val_precision = ((val_typmod - VARHDRSZ) >> 16) & 0xffff;
+			
+			target_precision = val_precision + (scale - val_scale);
```
KushaalShroff, line 1313: Please explain the logic behind this.
Also lets add this test case:
```
1> create table tn(a numeric(5,3))
2> go
1> insert into tn values (12.345)
2> go

(1 rows affected)
1> insert into tn values (12.34)
2> go

(1 rows affected)
1> insert into tn values (123.4)
2> go
Msg 8115, Level 16, State 8, Server EC2AMAZ-T4MF7PS, Line 1
Arithmetic overflow error converting numeric to data type numeric.
The statement has been terminated.
```

test/JDBC/expected/money_aggregate-vu-verify.out
```diff
@@ -209,6 +209,6 @@ varchar#!#varchar#!#smallint#!#tinyint#!#tinyint
 avgamount#!#money#!#8#!#19#!#4
 maxamount#!#money#!#8#!#19#!#4
 minamount#!#money#!#8#!#19#!#4
-totalwithmarkup#!#numeric#!#17#!#38#!#38
+totalwithmarkup#!#numeric#!#17#!#38#!#6
```
KushaalShroff, line 212: I believe this should be 22,5. Lets check with @tanyagupta17 to check this with her PR

KushaalShroff (changes requested):
As discussed offline, we also need to fix the overflow conditions for addition and subtraction.
As per: https://github.com/MicrosoftDocs/sql-docs/blob/live/docs/t-sql/data-types/precision-scale-and-length-transact-sql.md
> In addition and subtraction operations, we need max(p1 - s1, p2 - s2) places to store the integral part of the decimal number. If there isn't enough space to store them (that is, max(p1 - s1, p2 - s2) < min(38, precision) - scale), the scale is reduced to provide enough space for the integral part. The resulting scale is min(precision, 38) - max(p1 - s1, p2 - s2), so the fractional part might be rounded to fit into the resulting scale.

contrib/babelfishpg_tsql/src/pltsql_coerce.c
```diff
@@ -1064,6 +1096,807 @@ get_immediate_base_type_of_UDT_internal(Oid typeid)
...
+						scale = 6;
+					}
+					/*
+					 * Control reaching here for only arithmetic overflow
+					 * cases
+					 */
+					else
+					{
+						if (found != NULL) *found = false;
+					}
```
rohit01010: based on the above checks it seems like the control will never go into this else block. Lets remove it.

contrib/babelfishpg_tsql/src/pltsql_coerce.c
```diff
@@ -1064,6 +1096,807 @@ get_immediate_base_type_of_UDT_internal(Oid typeid)
...
+
+				if (var->vartypmod == -1)
+				{
+					/* UDT handling in T_var */
+					Oid immediate_base_type = get_immediate_base_type_of_UDT_internal(var->vartype);
+					if (OidIsValid(immediate_base_type))
+					{
+						Oid type = var->vartype;
+						var->vartype = getBaseTypeAndTypmod(var->vartype, &var->vartypmod);
+						var->vartype = type;
```
rohit01010: 1. Can be refactored as follows
```
getBaseTypeAndTypmod(var->vartype, &var->vartypmod);
```
2. Seems like here we are changing the input `expr` by setting the `vartypmod` field, where as in the rest of the code of this function we only compute the typmod and return without changing the input expression, is this  change intentional ?

contrib/babelfishpg_tsql/src/pltsql_coerce.c
```diff
@@ -1064,6 +1096,807 @@ get_immediate_base_type_of_UDT_internal(Oid typeid)
...
+							return var->vartypmod;
+					}
+
+					/* handling for fixed length datatypes */
+					if (plan && var->vartype == INT4OID)
+						return ((INT_PRECISION_RADIX << 16) | 0) + VARHDRSZ;
+					else if (plan && var->vartype == INT8OID)
+						return ((BIGINT_PRECISION_RADIX << 16) | 0) + VARHDRSZ;
+					else if (plan && var->vartype == INT2OID)
+						return ((SMALLINT_PRECISION_RADIX << 16) | 0) + VARHDRSZ;
```
rohit01010, line 1393: why we need to check whether plan is NULL or not here ?

rohit01010: Also, can you please point to the testcases added for this change.

contrib/babelfishpg_tsql/src/pltsql_coerce.c
```diff
@@ -1064,6 +1096,807 @@ get_immediate_base_type_of_UDT_internal(Oid typeid)
...
+					 */
+					if (con->consttype == INT4OID ||
+						con->consttype == INT8OID ||
+						con->consttype == INT2OID)
+					{
+						val = con->constvalue;
+						num = int64_to_numeric(val);
+						if (*pltsql_protocol_plugin_ptr && (*pltsql_protocol_plugin_ptr)->get_numeric_get_typmod)
+							return (*pltsql_protocol_plugin_ptr)->get_numeric_get_typmod(num);
+						return -1;
```
rohit01010, line 1277: shouldn't we set the found flag as false here since we are not able to compute the typmod ?

contrib/babelfishpg_tsql/src/pltsql_coerce.c
```diff
@@ -1064,6 +1096,807 @@ get_immediate_base_type_of_UDT_internal(Oid typeid)
...
+					}
+
+					num = (Numeric) con->constvalue;
+					if (*pltsql_protocol_plugin_ptr && (*pltsql_protocol_plugin_ptr)->get_numeric_get_typmod)
+						return (*pltsql_protocol_plugin_ptr)->get_numeric_get_typmod(num);
+					return -1;
```
rohit01010, line 1285: shouldn't we set the found flag as false here since we are not able to compute the typmod ?

Yvinayak07:
> As discussed offline, we also need to fix the overflow conditions for addition and subtraction. As per: https://github.com/MicrosoftDocs/sql-docs/blob/live/docs/t-sql/data-types/precision-scale-and-length-transact-sql.md
> 
> > In addition and subtraction operations, we need max(p1 - s1, p2 - s2) places to store the integral part of the decimal number. If there isn't enough space to store them (that is, max(p1 - s1, p2 - s2) < min(38, precision) - scale), the scale is reduced to provide enough space for the integral part. The resulting scale is min(precision, 38) - max(p1 - s1, p2 - s2), so the fractional part might be rounded to fit into the resulting scale.

We already have handling for this the addition error is thrown from `numeric_sum`.

contrib/babelfishpg_tsql/src/pltsql_coerce.c
Yvinayak07: i have added test cases: https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/3729/files#diff-098cb2091c99eec2a21b720b72244823f1052a784ac004c93dcb595cd0269124R135

test/JDBC/expected/money_aggregate-vu-verify.out
Yvinayak07: it seems like we are not able to calculate typmod for money correctly, it will be fixed after @tanyagupta17 changes.

tanyagupta17: We are able to get the correct typmod for money/smallmoney after fix. The issue lies here as we are hardcoding the precision for sum/avg: https://github.com/babelfish-for-postgresql/babelfish_extensions/blob/BABEL_5_X_DEV/contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c#L1108
We should restrict this for numeric/decimal only, others should be handled differently.

contrib/babelfishpg_tsql/src/hooks.c
Yvinayak07: I have decalred the variables in the lower block.

contrib/babelfishpg_tsql/src/hooks.c
Yvinayak07: added the test case.
Explanation: if their is a value with integral part greater then calculated integral part then we can't accomodate that value in calculate typmod so we will throw error in that case.
let's take the above example you mentioned:
- actual integral part should be 5-3 = 2.
- in 123.4 the integral pat will be 3 and scale is 1 where as calculated integral part is 2 so we can't fix number with integral part 3 in calculated integral part as 2.

contrib/babelfishpg_tsql/src/hooks.c
Yvinayak07: added comments here.

contrib/babelfishpg_tsql/src/pltsql_coerce.c
Yvinayak07: removed tis condition.

contrib/babelfishpg_tsql/src/pltsql_coerce.c
Yvinayak07: yes made these changes.

contrib/babelfishpg_tsql/src/pltsql_coerce.c
Yvinayak07: added flag

contrib/babelfishpg_tsql/src/pltsql_coerce.c
Yvinayak07: added flag

rohit01010 (approved):

KushaalShroff (approved):
Please add comment why we have added changes in both upgrade scripts for clarity of reviewers

Yvinayak07:
> Please add comment why we have added changes in both upgrade scripts for clarity of reviewers

Added comments in description.

PR #3561, opened 2025-03-09
fix issue with incorrect precision and scale value is computed of constant with value 0

rohit01010:
### Description
Currently when typmod information is computed of a constant using `numeric_get_typmod` function, it returns `precision = 18 and scale = 0` when constant value is 0 irrespective of its scale. Although this should be the case only if constant value is  0 and its scale is also 0. Fixed this logic by adding a check on scale.

### Issues Resolved
BABEL-5655

### Test Scenarios Covered ###
* **Use case based -** 

* **Boundary conditions -**

* **Arbitrary inputs -**

* **Negative test cases -**

* **Minor version upgrade tests -**

* **Major version upgrade tests -**

* **Performance tests -**

* **Tooling impact -**

* **Client tests -**

### Check List
- [ ] Commits are signed per the DCO using --signoff 

By submitting this pull request, I confirm that my contribution is under the terms of the Apache 2.0 and PostgreSQL licenses, and grant any person obtaining a copy of the contribution permission to relicense all or a portion of my contribution to the PostgreSQL License solely to contribute all or a portion of my contribution to the PostgreSQL open source project.

For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/babelfish-for-postgresql/babelfish_extensions/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #3731, opened 2025-04-30
Check array not empty before accessing leading digits in Numeric

sharathbp:
### Description

Before fetching the leading digits from numeric added check if the numeric array is empty or not as it leads to invalid read when the numeric is zero.

### Issues Resolved
BABEL-4631

* **Use case based -**
Yes

* **Boundary conditions -**
Yes

* **Arbitrary inputs -**
Yes

* **Negative test cases -**
Yes

NA

NA

* **Performance tests -**
NA

* **Tooling impact -**
NA

* **Client tests -**
NA

### Check List

### DIFF
--- a/contrib/babelfishpg_common/src/numeric.c
+++ b/contrib/babelfishpg_common/src/numeric.c
@@ -1018,7 +1018,7 @@ tsql_numeric_get_typmod(Numeric num)
 			10,
 			1,
 		};
-		int			leading_digits = NUMERIC_DIGITS(num)[0];
+		int			leading_digits = NUMERIC_NDIGITS(num) != 0 ? NUMERIC_DIGITS(num)[0] : 0;
 
 		precision = weight * DEC_DIGITS + scale;
 
--- a/contrib/babelfishpg_common/src/numeric.c
+++ b/contrib/babelfishpg_common/src/numeric.c
@@ -1010,7 +1010,17 @@ tsql_numeric_get_typmod(Numeric num)
 	int32_t		weight = NUMERIC_WEIGHT(num);
 	int32_t		precision;
 
-	if (weight >= 0)
+	/*
+	 * We can identify a zero by the fact that there are no digits at all. In
+	 * case of zero both precision and scale will be evaluated to zero, so we
+	 * will set (precision,scale) to T-SQL default (18,0).
+	 */
+	if (NUMERIC_NDIGITS(num) == 0)
+	{
+		precision = 18;
+		scale = 0;
+	}
+	else if (weight >= 0)
 	{
 		static const int32 timescales[DEC_DIGITS] = {
 			1000,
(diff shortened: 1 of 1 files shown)


contrib/babelfishpg_common/src/numeric.c
```diff
@@ -1018,7 +1018,7 @@ tsql_numeric_get_typmod(Numeric num)
 			10,
 			1,
 		};
-		int			leading_digits = NUMERIC_DIGITS(num)[0];
+		int			leading_digits = NUMERIC_NDIGITS(num) != 0 ? NUMERIC_DIGITS(num)[0] : 0;
```
tanscorpio7: in tds extension there is a copy of this function. `numeric_get_typmod` that handles zero digits this way.
```
/*
* We can identify a zero by the fact that there are no digits at all. In
* case of zero both precision and scale will be evaluated to zero, so we
* will set (precision,scale) to T-SQL default (18,0).
*/
if (NUMERIC_NDIGITS(num) == 0)
{
precision = 18;
scale = 0;
}
```
Which one should we use ?

tanscorpio7 (changes requested):

contrib/babelfishpg_common/src/numeric.c
sharathbp: Yeah right, client doesn't accept 0 prevision will update it to set default value
```
1> select cast(0.0 as numeric(0,0))
2> go
Specified column precision 0 for 'numeric' datatype must be within the range 1 to maximum precision(38)
```

contrib/babelfishpg_common/src/numeric.c
```diff
@@ -1010,7 +1010,17 @@ tsql_numeric_get_typmod(Numeric num)
...
+	/*
+	 * We can identify a zero by the fact that there are no digits at all. In
+	 * case of zero both precision and scale will be evaluated to zero, so we
+	 * will set (precision,scale) to T-SQL default (18,0).
+	 */
+	if (NUMERIC_NDIGITS(num) == 0)
+	{
+		precision = 18;
+		scale = 0;
+	}
```
rishabhtanwar29, line 1022: @rohit01010 Has this PR open https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/3561, which will add some more logic at this place. Can you check with him if we still need that logic?

contrib/babelfishpg_common/src/numeric.c
rishabhtanwar29, line 1: Can we add some tests for this? Have we atleast tested it with valgrind.

sharathbp: Yes I have tested with Valgrind and there is no Invalid read as it doesnt calls NUMERIC_DIGITS(num)[0] which is in else code block. Will add more tests.

contrib/babelfishpg_common/src/numeric.c
sharathbp: It will be fixed as part of this pr https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/3729
Will close this pr as its not required.

sharathbp:
Fixed as part of this pr https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/3729

PR #3733, opened 2025-05-01
Added sp_helplogins - stored proc that provides information about logins and users associated with those logins.

ayushdsh:
### Description

sp_helplogins [ [ @LoginNamePattern = ] N'LoginNamePattern' ] [ ; ] @LoginNamePattern is sys.sysname, with a default of NULL . If specified, @loginnamepattern must exist.

It returns two result sets -
First view - information about all logins present on the server. Second view - information about each login and its mapping with a user in a database or its membership with a database-role

Task - BABEL-5742

#### Perf testing numbers

```
100 logins and users
Average execution time over 20 runs: 592 ms
Single connection Average time: 35 ms

500 logins and users
Average execution time over 20 runs: 605 ms
Single connection Average time: 57 ms

1000 logins and users
Average execution time over 20 runs: 637 ms
Single connection Average time: 100 ms
```

### Issues Resolved

* **Use case based -**
    - Logged-in user is a member of sysadmin role (superuser)
    - Logged-in user is a member of securityadmin  role
    - Logged-in user is a member of securityadmin role and also a member of db_securityadmin or db_accessadmin
    - Logged-in user is not a member of securityadmin role
    - Login does not have any users

* **Boundary conditions -**
    - Parameter passed is an invalid login

* **Arbitrary inputs -**
    - Parameter passed is an invalid login

* **Negative test cases -**
    - Parameter passed is empty
    - Parameter passed contains padded spaces L-R

    * NA

    * NA

* **Performance tests -**
    * NA

* **Tooling impact -**
    * NA

* **Client tests -**
    * NA

### Check List


test/JDBC/input/storedProcedures/Test-sp_helplogins-vu-verify.mix
```diff
@@ -0,0 +1,139 @@
+-- prepare script for sp_helplogins
```
roshan0708: Can you move the prepare sqls into a prepare test file? In that we can test this via version upgrades as well.

test/JDBC/input/storedProcedures/Test-sp_helplogins-vu-verify.mix
```diff
@@ -0,0 +1,139 @@
...
+
+-- tsql user=jdbc_user password=12345678
+EXEC sp_helplogins 'testloginwithsecurityadmin2  '
+GO
+
+-- tsql user=jdbc_user password=12345678
+EXEC sp_helplogins ' testloginwithsecurityadmin2 '
+GO
+
+-- tsql user=jdbc_user password=12345678
```
roshan0708: We don't need to use `-- tsql` for each and every test query. Once we create a connection with a login, we can re-use that connection for other test queries as well.

test/JDBC/jdbc_schedule
```diff
@@ -579,3 +579,7 @@ ignore#!#test_constraint_like-16-6-vu-cleanup
 ignore#!#sys-fn-varbintohexsubstring-before-16_10-or-17_6-vu-prepare
 ignore#!#sys-fn-varbintohexsubstring-before-16_10-or-17_6-vu-verify
 ignore#!#sys-fn-varbintohexsubstring-before-16_10-or-17_6-vu-cleanup
+
+# More info: BABEL-5742
+ignore#!#Test-sp_helplogins-vu-verify
+ignore#!#Test-sp_helplogins-vu-cleanup
```
roshan0708: We need to run these in jdbc test suite? If yes, we'll need to remove this.

Also, shouldn't we add these in `latest` upgrade file for version upgrade tests?

contrib/babelfishpg_tsql/sql/babelfishpg_tsql.sql
```diff
@@ -3754,3 +3754,192 @@ BEGIN
...
+    UExt2.database_name as database_name,
+	UExt1.orig_username as role_name,
+	UExt2.login_name as member_login
+  FROM pg_catalog.pg_auth_members AS Authmbr
+  INNER JOIN pg_catalog.pg_roles AS PGR1 ON PGR1.oid = Authmbr.roleid
+  INNER JOIN pg_catalog.pg_roles AS PGR2 ON PGR2.oid = Authmbr.member
+  INNER JOIN sys.babelfish_authid_user_ext AS UExt1 ON PGR1.rolname = UExt1.rolname
+  INNER JOIN sys.babelfish_authid_user_ext AS UExt2 ON PGR2.rolname = UExt2.rolname
+  WHERE UExt1.orig_username IN ('db_securityadmin', 'db_accessadmin')
+  AND UExt2.login_name = sys.suser_name()
```
roshan0708: We should store username in a variable to avoid multiple `sys.suser_name()` calls.

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--5.2.0--5.3.0.sql
```diff
@@ -112,6 +112,195 @@ END;
...
+  INNER JOIN sys.babelfish_authid_user_ext AS UExt1 ON PGR1.rolname = UExt1.rolname
+  INNER JOIN sys.babelfish_authid_user_ext AS UExt2 ON PGR2.rolname = UExt2.rolname
+  WHERE UExt1.orig_username IN ('db_securityadmin', 'db_accessadmin')
+  AND UExt2.login_name = sys.suser_name()
+
+  SELECT
+    CASE ISNULL(UExt.login_name, '')
+		WHEN '' THEN CAST(Db.owner AS sys.SYSNAME)
+		ELSE CAST(UExt.login_name AS sys.SYSNAME)
+	END AS LoginName,
```
roshan0708: I think using `COALESCE/NULLIF` would be more robust for `NULL` handling.

contrib/babelfishpg_tsql/sql/babelfishpg_tsql.sql
```diff
@@ -3754,3 +3754,192 @@ BEGIN
...
+    CAST(UExt1.orig_username AS SYS.SYSNAME) AS UserName,
+    'Member of' AS UserOrAlias 
+  FROM pg_catalog.pg_auth_members AS Authmbr
+  INNER JOIN pg_catalog.pg_roles AS PGR1 ON PGR1.oid = Authmbr.roleid
+  INNER JOIN pg_catalog.pg_roles AS PGR2 ON PGR2.oid = Authmbr.member
+  INNER JOIN sys.babelfish_authid_user_ext AS UExt1 ON PGR1.rolname = UExt1.rolname
+  INNER JOIN sys.babelfish_authid_user_ext AS UExt2 ON PGR2.rolname = UExt2.rolname
+  LEFT JOIN sys.babelfish_sysdatabases Db ON Db.name = UExt1.database_name
+  WHERE UExt1.type = 'R' AND 
+        UExt2.orig_username != 'db_owner' AND
```
roshan0708: Can we move these checks to `JOIN` clauses to filter rows earlier and reduce extra processing post multiple joins?

contrib/babelfishpg_tsql/sql/babelfishpg_tsql.sql
```diff
@@ -3754,3 +3754,192 @@ BEGIN
 END;
 $$ LANGUAGE pltsql;
 GRANT EXECUTE ON PROCEDURE sys.sp_procedure_params_100_managed TO PUBLIC;
+
+CREATE OR REPLACE PROCEDURE sys.sp_helplogins()
```
roshan0708: Both `sp_helplogins` procedures have nearly identical logic with minor differences in filtering. An internal wrapper function would reduce duplication, improve maintainability, and ensure consistent behavior.

roshan0708 (changes requested):

test/JDBC/jdbc_schedule
ayushdsh: We do need to run them in all tests except singledb-mode .. however currently I am stuck on this because this proc returns two result sets i.e. two tables, one of which contains SID which is an identifier (which will keep changing on each test run). The way we tackle this is using a separate table to store the results of the proc, but INSERT INTO .. EXEC does not work if the proc returns more than 1 result set.
I will work on the test separately. That's why added it as ignored for now.

> Also, shouldn't we add these in latest upgrade file for version upgrade tests?

And yes, this should also be added to the upgrade tests as well, but the same reason as above.

contrib/babelfishpg_tsql/sql/babelfishpg_tsql.sql
ayushdsh: Oh yes. Agreed. Will make the change.

contrib/babelfishpg_tsql/sql/babelfishpg_tsql.sql
ayushdsh: Right. That's true. However, do you think in that case, I need to have a explicit revoke for execution, or will that be taken care implicitly?

test/JDBC/input/storedProcedures/Test-sp_helplogins-vu-verify.mix
ayushdsh: But for the verification, we create connections using different logins like superuser and other logins to get the behaviour difference of the proc.

test/JDBC/input/storedProcedures/Test-sp_helplogins-vu-verify.mix
ayushdsh: So in this particular case, we create certain logins with server role permissions as well like `securityadmin`. Another test for a proc - `sp_helpsrvrolemember` gets all the logins which have server role. Since all prepare scripts run first, if I write this in the prepare the output for that test is also altered. 

I just wanted to confirm if it is correct to modify that test as well in this case?

test/JDBC/input/storedProcedures/Test-sp_helplogins-vu-verify.mix
roshan0708: Right, we do create connections for different logins. However, when we're executing consecutive queries with the same user like:
```
-- tsql user=user_1
QUERY 1
GO

-- tsql user=user_1
QUERY 2
GO
```

This unnecessarily creates a new connection for each block. We could optimize this by reusing the existing connection:
```
-- tsql user=user_1
QUERY 1
GO

QUERY 2
GO
```

contrib/babelfishpg_tsql/sql/babelfishpg_tsql.sql
roshan0708: No explicit `REVOKE` is needed as long as you don't grant explicit `EXECUTE` permissions on the wrapper.

test/JDBC/input/storedProcedures/Test-sp_helplogins-vu-verify.mix
roshan0708: Output of these test files largely depends upon the order of execution of different test files. If test (JDBC) file for `sp_helplogins` runs before, cleanup is all done and nothing needs to be handled. Although, this won't be the case for version upgrade/dump-restore tests, in such case, we'll need to update these outputs for test cases for proc `sp_helpsrvrolemember`. 

@HarshLunagariya / @shalinilohia50 could answer this question better.

contrib/babelfishpg_tsql/sql/babelfishpg_tsql.sql
ayushdsh: I tried to implement this using a view and sp as well .. the problem is - when user executes the `sp_helplogins` proc, unless the user has permissions on the internal views/sp they won't be able to get the result set for sp_helplogins. 

If you don't mind, can we connect for me to understand this better. Thanks

roshan0708: You can refer to implementation of `sp_describe_undeclared_parameters_internal` and `sp_column_privileges_internal` and their usage. We are using them inside the parent procedure and view.

ayushdsh: Yes. I got some idea. Added the new commit.. let me know if it makes sense!

test/python/expected/upgrade_validation/expected_dependency.out
```diff
@@ -981,6 +981,7 @@ View information_schema_tsql.columns_internal
 View information_schema_tsql.constraint_column_usage
 View information_schema_tsql.domains
 View information_schema_tsql.routines
+View sys.all_database_users
```
ayushdsh: Afaiu, upgrade_validation here tests if there are any sys views created that are a dependencies for others; but it does not consider database objects within the sys or other internal catalog itself. Hence, it is expected to be added here. 

Please help me understand if otherwise. @roshan0708 @shalinilohia50

contrib/babelfishpg_tsql/sql/babelfishpg_tsql.sql
```diff
@@ -3754,3 +3754,154 @@ BEGIN
...
+FROM pg_catalog.pg_roles AS Base INNER JOIN sys.babelfish_authid_user_ext AS Ext
+ON Base.rolname = Ext.rolname
+LEFT OUTER JOIN pg_catalog.pg_roles Base2
+ON Ext.login_name = Base2.rolname
+LEFT OUTER JOIN sys.babelfish_sysdatabases AS Db
+ON Ext.database_name COLLATE sys.database_default = Db.name
+LEFT OUTER JOIN pg_catalog.pg_roles AS Base3
+ON Db.owner = Base3.rolname
+WHERE Ext.type != 'R' AND Ext.orig_username IS NOT NULL;
+GRANT SELECT on sys.all_database_users TO PUBLIC;
```
HarshLunagariya: Why do we need this view? Can't we re-use sys.database_principals?

contrib/babelfishpg_tsql/sql/babelfishpg_tsql.sql
```diff
@@ -3754,3 +3754,154 @@ BEGIN
...
+	UExt2.database_name as database_name,
+	UExt1.orig_username as role_name,
+	UExt2.login_name as member_login
+FROM pg_catalog.pg_auth_members AS Authmbr
+INNER JOIN pg_catalog.pg_roles AS PGR1 ON PGR1.oid = Authmbr.roleid
+INNER JOIN pg_catalog.pg_roles AS PGR2 ON PGR2.oid = Authmbr.member
+INNER JOIN sys.babelfish_authid_user_ext AS UExt1 ON PGR1.rolname = UExt1.rolname
+INNER JOIN sys.babelfish_authid_user_ext AS UExt2 ON PGR2.rolname = UExt2.rolname
+WHERE UExt1.orig_username IN ('db_securityadmin', 'db_accessadmin');
+GRANT SELECT on sys.db_role_mapping TO PUBLIC;
```
HarshLunagariya: Have we defined this just to check db_securityadmin & db_accessadmin membership?
Better would be to utilise pg_has_role() function directly.

contrib/babelfishpg_tsql/sql/babelfishpg_tsql.sql
```diff
@@ -3754,3 +3754,154 @@ BEGIN
...
+AS $$
+BEGIN
+	EXEC sp_helplogins_internal_logins;
+	EXEC sp_helplogins_internal_user_mappings;
+  RETURN 0;
+END;
+$$;
+GRANT EXECUTE ON PROCEDURE sys.sp_helplogins() TO PUBLIC;
+
+CREATE OR REPLACE PROCEDURE sys.sp_helplogins("@loginname" sys.SYSNAME)
```
HarshLunagariya: Just double checking:
Do we actually need two variations - `sys.sp_helplogins()` & `sys.sp_helplogins("@loginname" sys.SYSNAME)` ?

contrib/babelfishpg_tsql/sql/babelfishpg_tsql.sql
```diff
@@ -3754,3 +3754,154 @@ BEGIN
...
+    CAST(LExt.default_database_name AS SYS.SYSNAME) AS DefDBName,
+    CAST(LExt.default_language_name AS SYS.SYSNAME) AS DefLangName,
+    CASE 
+      WHEN Dp.oid IS NOT NULL THEN 'YES'
+      ELSE 'NO'
+    END as AUser,
+    'NO' AS ARemote -- Currently we do not support linking local logins to remote logins
+  FROM pg_catalog.pg_roles AS Base 
+  INNER JOIN sys.babelfish_authid_login_ext AS LExt ON Base.rolname = LExt.rolname
+  LEFT JOIN sys.all_database_users Dp ON Dp.oid = Base.oid -- In order to find out if a login has any users associated with it
```
HarshLunagariya: Instead for defining a separate view, please expand `all_database_users` here itself.

contrib/babelfishpg_tsql/sql/babelfishpg_tsql.sql
```diff
@@ -3754,3 +3754,154 @@ BEGIN
...
+
+  SELECT
+    CAST(COALESCE(NULLIF(UExt.login_name, ''), Db.owner) AS sys.SYSNAME) AS LoginName,
+		CAST(UExt.database_name AS sys.SYSNAME) AS DBName,
+		CAST(UExt.orig_username AS SYS.SYSNAME) AS UserName,
+		'User' AS UserOrAlias 
+  FROM sys.babelfish_authid_user_ext UExt
+  LEFT JOIN sys.babelfish_sysdatabases Db ON Db.name = UExt.database_name COLLATE database_default
+  WHERE UExt.type != 'R' AND  
+		UExt.orig_username != 'guest' AND
```
HarshLunagariya: Are we checking whether guest user is enabled or not ?
I'm refering to:
> Therefore, for each database on the server, at least one of the following conditions must be true:
>
> The user that is executing sp_helplogins has permission to access the database.
>
> The guest user account is enabled in the database.

contrib/babelfishpg_tsql/sql/babelfishpg_tsql.sql
ayushdsh: sys.database_principals is specific to the current database (has additional filter - database = db_name()). But we need a server level view.

contrib/babelfishpg_tsql/sql/babelfishpg_tsql.sql
ayushdsh: We actually might not. Previously, I had a very complex logic for this, and that's why I thought of segregating for ease of maintainence. But now, I think I can merge those to have just one.

contrib/babelfishpg_tsql/sql/babelfishpg_tsql.sql
ayushdsh: The `has_dbaccess()` condition takes care of that. If the user has access to the database (is a member of db roles) or guest user is enabled, `has_dbaccess(<dbname>)` returns true. 

The condition -  `UExt.orig_username != 'guest'` is only used for filtering out guest users from the result, since it is not returned in the result set of `sp_helplogins`

contrib/babelfishpg_tsql/sql/babelfishpg_tsql.sql
ayushdsh: I think it's the same problem with pg_has_role as well that it works on the current database session. However for this, we want to check if the login is member of role (db_securityadmin & db_accessadmin ) for each database.

Please correct me if I am wrong. I might completely have the wrong idea

test/JDBC/upgrade/latest/schedule
```diff
@@ -453,6 +453,7 @@ Test-sp_rename
 Test-sp_rename-dep
 Test-sp_set_session_context
 Test-sp_set_session_context-dep
+# BABEL-5742-sp_helplogins
```
ayushdsh: Since this is a server level proc, I am unable to find a good way to isolate this tests from others. The problem here is that other tests are also creating logins and users which are visible here because all prepares run first in case of dump-restore/mVU/MVU

ayushdsh:
> [Tests / JDBC Tests / run-babelfish-jdbc-tests (pull_request)](https://github.com/babelfish-for-postgresql/babelfish_extensions/actions/runs/14890593557/job/41821421518?pr=3733)

This test fails because of increased warnings - 
> ./.github/scripts/scan-warnings.sh: line 38: [[: 0
0: syntax error in expression (error token is "0")
Error: Expected 324 leak warnings, but found 394
Unexpected warning counts detected. Please investigate.
Snapshot 'still active' warnings: 44
Leak warnings: 394
Error: Process completed with exit code 1.

After checking, this is because of how the test and actual proc is implemented i.e. using `INSERT INTO ... EXEC`

Currently - 
> tail -f logfile | grep TupleDesc
2025-05-08 14:24:49.693 UTC [18932] WARNING:  resource was not closed: TupleDesc 0x362915d8 (85682,-1)
2025-05-08 14:24:49.693 UTC [18932] WARNING:  resource was not closed: TupleDesc 0x362915d8 (85682,-1)
2025-05-08 14:24:49.708 UTC [18932] WARNING:  resource was not closed: TupleDesc 0x7f18efb57648 (85687,-1)
2025-05-08 14:24:49.708 UTC [18932] WARNING:  resource was not closed: TupleDesc 0x7f18efb57648 (85687,-1)

After removing `INSERT INTO .. EXEC`

> tail -f logfile | grep TupleDesc


This might be because of an incorrect impl of `INSERT INTO .. EXEC` 
I think this is unavoidable and hence we need to mould the scan-warnings.sh file accordingly. What are your thoughts? @HarshLunagariya @roshan0708 @shalinilohia50

.github/scripts/scan-warnings.sh
```diff
@@ -49,8 +49,8 @@ if [[ "$SNAPSHOT_ACTIVE_COUNT" -ne 44 ]]; then
     ERROR_FOUND=true
 fi
 
-if [[ "$LEAK_COUNT" -ne 350 ]]; then
-    echo "Error: Expected 324 leak warnings, but found $LEAK_COUNT"
+if [[ "$LEAK_COUNT" -ne 394 ]]; then
+    echo "Error: Expected 394 leak warnings, but found $LEAK_COUNT"
```
ayushdsh: created an internal task to track this fix

.github/composite-actions/dump-restore-util/action.yml
```diff
@@ -104,7 +104,8 @@ runs:
 
           # Get the list of all the Babelfish logins in a file
           query="SELECT orig_loginname, default_database_name, default_language_name, type, \
-                pg_has_role(rolname, 'sysadmin', 'member') AS is_sysadmin_member \
+                pg_has_role(rolname, 'sysadmin', 'member') AS is_sysadmin_member, \
+                pg_has_role(rolname, 'securityadmin', 'member') AS is_securityadmin_member \
```
HarshLunagariya, line 108: Why this change?

contrib/babelfishpg_tds/babelfishpg_tds--1.0.0.sql
```diff
@@ -8,7 +8,7 @@ RETURNS text
 AS 'MODULE_PATHNAME'
 LANGUAGE C VOLATILE STRICT;
 
-CREATE FUNCTION sys.inject_fault(
+CREATE FUNCTION sys.inject_fault(
```
HarshLunagariya: Please remove this change.

contrib/babelfishpg_tsql/sql/babelfishpg_tsql.sql
```diff
@@ -3754,3 +3754,152 @@ BEGIN
...
+    CAST(LExt.default_database_name AS SYS.SYSNAME) AS DefDBName,
+    CAST(LExt.default_language_name AS SYS.SYSNAME) AS DefLangName,
+    CASE 
+      WHEN Dp.oid IS NOT NULL THEN 'YES'
+      ELSE 'NO'
+    END as AUser,
+    'NO' AS ARemote -- Currently we do not support linking local logins to remote logins
+  FROM pg_catalog.pg_roles AS Base 
+  INNER JOIN sys.babelfish_authid_login_ext AS LExt ON Base.rolname = LExt.rolname
+  LEFT JOIN #all_database_users Dp ON Dp.oid = Base.oid -- In order to find out if a login has any users associated with it
```
HarshLunagariya: Please fix the indentation.

contrib/babelfishpg_tsql/sql/babelfishpg_tsql.sql
```diff
@@ -3754,3 +3754,152 @@ BEGIN
...
+LANGUAGE pltsql
+AS $$
+BEGIN
+	IF is_srvrolemember('securityadmin') = 0 
+  BEGIN
+    RAISERROR('User does not have permission to perform this action.', 16, 1);
+		RETURN 0;
+  END
+
+	SELECT DISTINCT
```
HarshLunagariya: Some possible optimizations:
1. Do we need DISTINCT?
2. Instead of temp table, Can we utilize CTE? Can you please post the plan of both the variation. It will help us decide with one to choose.

HarshLunagariya (changes requested):

.github/composite-actions/dump-restore-util/action.yml
ayushdsh: Currently, when we migrate logins (dump/restore at the db level) we create files like logins, users (dump part) and then use these files for restoring logins and users. When restoring logins we take care of logins which are member of sysadmin (when parsing the logins at restore time, we check if the login was a member of sysadmin; if it was then we make it a member), but we do not handle other server-level roles (like securityadmin).

contrib/babelfishpg_tsql/sql/babelfishpg_tsql.sql
ayushdsh: 1. We would need DISTINCT, because this temp table outputs OIDs per user per database, so lets say we create 5 databases which has jdbc_user as the owner, this query would output 5 + 3(master, tempdb, msdb) = 8 identical OIDs. But our view of logins is independent of databases since its a server level view and only cares if there exists users for this login 
2. That's a very good suggestion. I am attaching the query plan for both cte and temp table below. Here's what I found.

First run - 
CTE - 395ms
temp table - 405ms

Subsequent runs - 
CTE - 0.16 - 0.18
temp tables - 0.17-0.18

[plan.xlsx](https://github.com/user-attachments/files/20347612/plan.xlsx)

contrib/babelfishpg_tsql/sql/babelfishpg_tsql.sql
```diff
@@ -3754,3 +3754,160 @@ BEGIN
 END;
 $$ LANGUAGE pltsql;
 GRANT EXECUTE ON PROCEDURE sys.sp_procedure_params_100_managed TO PUBLIC;
+
+CREATE OR REPLACE PROCEDURE sys.sp_helplogins_internal_logins()
```
roshan0708: nit: Please fix indentations at all places.

.github/composite-actions/dump-restore-util/action.yml
HarshLunagariya: Nice catch! Thanks for fixing.

contrib/babelfishpg_tsql/sql/babelfishpg_tsql.sql
```diff
@@ -3754,3 +3754,140 @@ BEGIN
...
+        INNER JOIN sys.babelfish_authid_login_ext AS LExt ON Base.rolname = LExt.rolname
+        LEFT JOIN sys.babelfish_authid_user_ext AS Ext ON Ext.login_name = Base.rolname AND Ext.type != 'R'
+        LEFT JOIN sys.babelfish_sysdatabases AS Db ON Db.name COLLATE database_default = LExt.default_database_name
+        WHERE LExt.type NOT IN ('R', 'Z')
+    
+        SELECT
+            CAST(COALESCE(NULLIF(UExt.login_name, ''), Db.owner) AS sys.SYSNAME) AS LoginName,
+            CAST(UExt.database_name AS sys.SYSNAME) AS DBName,
+            CAST(UExt.orig_username AS SYS.SYSNAME) AS UserName,
+            'User' AS UserOrAlias
```
HarshLunagariya: Let's include the `CAST(... AS ..) AS UserOrAlias`.

contrib/babelfishpg_tsql/sql/babelfishpg_tsql.sql
```diff
@@ -3754,3 +3754,140 @@ BEGIN
...
+
+CREATE OR REPLACE PROCEDURE sys.sp_helplogins(IN "@loginname" sys.sysname DEFAULT NULL)
+LANGUAGE pltsql
+AS $$
+DECLARE @input_loginname sys.sysname;
+DECLARE @current_username sys.nvarchar(128)
+DECLARE @is_sysadmin BIT
+BEGIN
+
+	IF is_srvrolemember('securityadmin') = 0
```
HarshLunagariya: Fix indentation

test/JDBC/src/main/java/com/sqlsamples/FilterConditions.java
```diff
@@ -0,0 +1,15 @@
+package com.sqlsamples;
```
roshan0708, line 1: Why is this change required?

ayushdsh: I've made the changes - wherein now we use inplace join. Since this returns a SID which is non-deterministic .. we can't directly use it in test; also we cannot capture a single result set from a stored proc which returns > 1 result sets.
With this, I am trying to introduce a new option - -- ignore_columns <int>,<int>  which when used with a query .. ignores the given columns and does not produce its output. 

This was not a problem previously, because with that we were using 2 internal procs .. and we were using the same internal procs while testing as well, with which we were able to ignore the SID column.

test/JDBC/src/main/java/com/sqlsamples/FilterConditions.java
```diff
@@ -0,0 +1,15 @@
...
+import java.util.Set;
+import java.util.HashSet;
+
+public class FilterConditions {
```
sumitj824, line 5: Do we really need a class for this? We can directly maintain  `Set<Integer> colsToIgnore` and pass as argument to other functions.

test/JDBC/input/BABEL_GRANT_CONNECT-vu-verify.mix
```diff
@@ -39,6 +39,10 @@ go
 drop user guest;
 go
 
+-- re-enabling connect on guest user
```
sumitj824, line 42: I assume this to fix intermittent test failures which is not related to your changes, right?

test/JDBC/input/storedProcedures/z_sp_helplogins-vu-prepare.mix
```diff
@@ -0,0 +1,3 @@
+-- create a dummy db
+create database sp_helplogins_db1
```
sumitj824, line 2: Just curious, what does the 'z' indicate the test files name? Also, If we are just creating only a database in prepare file and not creating objects/logins then there is no use of having upgrade test (prepare/verify/cleanup) for this.

contrib/babelfishpg_tsql/sql/babelfishpg_tsql.sql
```diff
@@ -3754,3 +3754,140 @@ BEGIN
...
+            CAST(LExt.orig_loginname AS sys.SYSNAME) AS LoginName,
+            CAST(CAST(Base.oid AS INT) AS sys.varbinary(85)) AS sid,
+            CAST(LExt.default_database_name AS SYS.SYSNAME) AS DefDBName,
+            CAST(LExt.default_language_name AS SYS.SYSNAME) AS DefLangName,
+            CASE 
+                WHEN Ext.login_name IS NOT NULL AND Ext.login_name = LExt.orig_loginname COLLATE database_default THEN 'YES'
+                WHEN Db.owner COLLATE database_default = LExt.orig_loginname THEN 'YES'
+                ELSE 'NO'
+            END AS AUser,
+            CAST('NO' AS VARCHAR(8)) AS ARemote -- Currently we do not support linking local logins to remote logins
```
sumitj824: Please use sys.varchar instead of varchar.

contrib/babelfishpg_tsql/sql/babelfishpg_tsql.sql
sumitj824: Please define the exact column type by adding explicit casts.

test/JDBC/input/BABEL_GRANT_CONNECT-vu-verify.mix
ayushdsh: That's correct. In this test, we are removing the guest user but never adding them back. This is to add the guest access back

test/JDBC/input/storedProcedures/z_sp_helplogins-vu-prepare.mix
ayushdsh: the 'z_' because i want these tests to run at the very end of all upgrade/dump-restore tests. This is just to avoid other tests interfering with this (since they create logins).

Currently, our instance-level dump-restore script does not migrate logins. That's why creating logins in the verify script.
I agree that not having them in prepare script is weird and upgrade and dump/restore might not make sense.

test/JDBC/src/main/java/com/sqlsamples/FilterConditions.java
ayushdsh: I have in mind to extend this more. For ex. `-- order-results`, `-- ignore-result-set` etc.

contrib/babelfishpg_tsql/sql/babelfishpg_tsql.sql
```diff
@@ -3754,3 +3754,140 @@ BEGIN
...
+    BEGIN    
+        SELECT DISTINCT
+            CAST(LExt.orig_loginname AS sys.SYSNAME) AS LoginName,
+            CAST(CAST(Base.oid AS INT) AS sys.varbinary(85)) AS sid,
+            CAST(LExt.default_database_name AS SYS.SYSNAME) AS DefDBName,
+            CAST(LExt.default_language_name AS SYS.SYSNAME) AS DefLangName,
+            CASE 
+                WHEN Ext.login_name IS NOT NULL AND Ext.login_name = LExt.orig_loginname COLLATE database_default THEN 'YES'
+                WHEN Db.owner COLLATE database_default = LExt.orig_loginname THEN 'YES'
+                ELSE 'NO'
```
HarshLunagariya: Please add explicit datatype cast for AUser

test/JDBC/input/BABEL_GRANT_CONNECT-vu-verify.mix
HarshLunagariya: Please think about cherry-picking this stabilizing changes to other branches as well if we are observing the same.

test/JDBC/src/main/java/com/sqlsamples/batch_run.java
```diff
@@ -301,6 +305,18 @@ static void batch_run_sql(Connection con_bbl, BufferedWriter bw, String testFile
                         checkSingleDbModeExpected = true;
                         continue;
                     }
+
+                    if (strLine.toLowerCase().startsWith("-- ignore_columns")) {
```
HarshLunagariya, line 309: 1. Please make sure to update the README file to include the directions to use this option.
2. Does it handle the scenario when we get quoted column names in result metadata? Please test once.

test/JDBC/src/main/java/com/sqlsamples/FilterConditions.java
HarshLunagariya: This can be achieved via SQL with inserting the output of the procedure to temp table. Was there any specific benefit? Please make sure framework owner is aware about this change.

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--5.2.0--5.3.0.sql
```diff
@@ -119,6 +119,143 @@ END;
...
+            (
+                @is_sysadmin = 1 OR
+                UExt2.login_name = @current_username OR
+                ISNULL(UExt2.login_name, '') = ''
+            )
+    END
+    ELSE
+    BEGIN
+
+        SET @input_loginname = sys.RTRIM(@loginname);
```
HarshLunagariya, line 302: No need for LTRIM?

test/JDBC/expected/z_sp_helplogins-vu-verify.out
```diff
@@ -0,0 +1,276 @@
...
+jdbc_user#!#tempdb#!#db_owner#!#Member of
+jdbc_user#!#tempdb#!#dbo#!#User
+sp_helplogins_testlogin#!#master#!#db_securityadmin#!#Member of
+sp_helplogins_testlogin#!#master#!#userof_sp_helplogins_testlogin#!#User
+testloginindb1#!#sp_helplogins_db1#!#userof_testloginindb1#!#User
+testloginwithotherdefdb#!#sp_helplogins_db1#!#u_testloginwithotherdefdb#!#User
+testloginwithsecurityadmin#!#master#!#userof_testloginwithsecurityadmin#!#User
+testloginwithsecurityadmin#!#sp_helplogins_db1#!#userof_testloginwithsecurityadmin_indb1#!#User
+testloginwithsecurityadmin2#!#sp_helplogins_db1#!#db_securityadmin#!#Member of
+testloginwithsecurityadmin2#!#sp_helplogins_db1#!#userof_testloginwithsecurityadmin2#!#User
```
HarshLunagariya: Please create database roles as well in the test to validate that it shouldn't show user-created database roles.

test/JDBC/expected/z_sp_helplogins-vu-verify.out
```diff
@@ -0,0 +1,276 @@
...
+-- creating logins because instance level dump/restore does not migrate logins
+-- create a login with user in master with securityadmin server role
+-- create login with user in master with db_securityadmin role
+create login sp_helplogins_testlogin with password = '12345678'
+create user userof_sp_helplogins_testlogin for login sp_helplogins_testlogin
+GO
+
+alter role db_securityadmin add member userof_sp_helplogins_testlogin
+GO
+
```
HarshLunagariya: Please add a case where some login is owner of the database and the member of sysadmin server role.

test/JDBC/expected/z_sp_helplogins-vu-verify.out
```diff
@@ -0,0 +1,276 @@
...
+~~START~~
+varchar#!#varchar#!#varchar#!#varchar
+testloginwithsecurityadmin2#!#sp_helplogins_db1#!#db_securityadmin#!#Member of
+testloginwithsecurityadmin2#!#sp_helplogins_db1#!#userof_testloginwithsecurityadmin2#!#User
+~~END~~
+
+
+-- ignore_columns 2
+EXEC sp_helplogins ' testloginwithsecurityadmin2 '
```
HarshLunagariya, line 447: Include the testcase where we are passing login names longer than 64 letters. And also include some cases where mapped user is also of longer length. This will help to validate whether we correct join filter condition to incorporate truncated names. And in the output, We should always show original name - untruncated name.

HarshLunagariya: Add cases for mixed case names with special chars, quoted names, enclosed with square brackets.
Add cases for windows login/user.
We need to add dependency test as well.

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--5.2.0--5.3.0.sql
ayushdsh: left spaces are considered part of the string. for example, for this proc - 
`testlogin  ` is same as `testlogin`, but `  testlogin` is considered as a whole

test/JDBC/src/main/java/com/sqlsamples/FilterConditions.java
ayushdsh: Since the proc returns 2 result sets, `insert into .. exec` does not work. And that's why we cannot also ignore the SID column

test/JDBC/expected/z_sp_helplogins-vu-verify.out
ayushdsh: It does show user defined roles in the second result set as well..

```
##MS_PolicyEventProcessingLogin## msdb ##MS_PolicyEventProcessingLogin## User    
##MS_PolicyEventProcessingLogin## msdb PolicyAdministratorRole MemberOf
##MS_PolicyTsqlExecutionLogin## msdb ##MS_PolicyTsqlExecutionLogin## User    
##MS_PolicyTsqlExecutionLogin## msdb PolicyAdministratorRole MemberOf
...
testlogin3 master sp_helplogins_role MemberOf ---> user defined db role
...
```

test/JDBC/expected/z_sp_helplogins-vu-verify.out
```diff
@@ -0,0 +1,324 @@
...
+sp_helplogins_testlogin#!#master#!#userof_sp_helplogins_testlogin#!#User
+testloginindb1#!#sp_helplogins_db1#!#userof_testloginindb1#!#User
+testloginwithotherdefdb#!#sp_helplogins_db1#!#db_datareader#!#Member of
+testloginwithotherdefdb#!#sp_helplogins_db1#!#u_testloginwithotherdefdb#!#User
+testloginwithsecurityadmin#!#master#!#userof_testloginwithsecurityadmin#!#User
+testloginwithsecurityadmin#!#sp_helplogins_db1#!#userof_testloginwithsecurityadmin_indb1#!#User
+testloginwithsecurityadmin2#!#sp_helplogins_db1#!#db_securityadmin#!#Member of
+testloginwithsecurityadmin2#!#sp_helplogins_db1#!#sp_helplogins_role#!#Member of
+testloginwithsecurityadmin2#!#sp_helplogins_db1#!#userof_testloginwithsecurityadmin2#!#User
+win@HELPLOGINS#!#master#!#sp_helplogins_vu_windows_user#!#User
```
HarshLunagariya: Both login and username should come from orig_name column.

HarshLunagariya (changes requested):

HarshLunagariya (approved):

contrib/babelfishpg_tsql/sql/babelfishpg_tsql.sql
```diff
@@ -3754,3 +3754,161 @@ BEGIN
...
+                    AND UExt2.database_name = UExt.database_name -- filter to check if the processing db is equal to the outer query db, since we want to find if the user is a member of the roles in the outer db
+                    AND UExt2.login_name = @current_username
+                )
+            )
+        UNION
+        SELECT
+            CAST(LExt.orig_loginname AS sys.SYSNAME) AS LoginName,
+            CAST(UExt2.database_name AS sys.SYSNAME) AS DBName,
+            CAST(UExt1.orig_username AS sys.SYSNAME) AS UserName,
+            CAST('Member of' AS sys.varchar(10)) AS UserOrAlias
```
sumitj824: The column type should be char(8), please fix it at other places as well.

contrib/babelfishpg_tsql/sql/babelfishpg_tsql.sql
```diff
@@ -3754,3 +3754,161 @@ BEGIN
...
+            CAST(LExt.orig_loginname AS sys.SYSNAME) AS LoginName,
+            CAST(CAST(Base.oid AS INT) AS sys.varbinary(85)) AS sid,
+            CAST(LExt.default_database_name AS SYS.SYSNAME) AS DefDBName,
+            CAST(LExt.default_language_name AS SYS.SYSNAME) AS DefLangName,
+            CASE 
+                WHEN Ext.login_name IS NOT NULL AND Ext.login_name = LExt.rolname COLLATE database_default THEN CAST('YES' AS sys.char(5))
+                WHEN Db.owner COLLATE database_default = LExt.orig_loginname THEN CAST('YES' AS sys.char(5))
+                ELSE CAST('NO' AS sys.char(5))
+            END AS AUser,
+            CAST('NO' AS sys.char(7)) AS ARemote -- Currently we do not support linking local logins to remote logins
```
sumitj824: I think you missed to fix the column type at other places.

contrib/babelfishpg_tsql/sql/babelfishpg_tsql.sql
```diff
@@ -3754,3 +3754,161 @@ BEGIN
...
+        LEFT JOIN sys.babelfish_sysdatabases AS Db ON Db.name COLLATE database_default = LExt.default_database_name
+        WHERE LExt.type NOT IN ('R', 'Z')
+
+        -- first selector in the union is to get all the mapped users
+        -- second selector in the union is to get all the mapped database/user-defined roles
+        SELECT
+            CAST(LExt.orig_loginname AS sys.SYSNAME) AS LoginName,
+            CAST(UExt.database_name AS sys.SYSNAME) AS DBName,
+            CAST(UExt.orig_username AS SYS.SYSNAME) AS UserName,
+            CAST('User' AS sys.char(8)) AS UserOrAlias
```
sumitj824, line 3800: I think you missed to fix the column type at other places.

contrib/babelfishpg_tsql/sql/babelfishpg_tsql.sql
```diff
@@ -3754,3 +3754,160 @@ BEGIN
...
+    END
+
+    SET @current_username = LOWER(sys.suser_name());
+    SET @is_sysadmin = is_srvrolemember('sysadmin');
+    
+    IF @loginname IS NULL
+    BEGIN    
+        SELECT DISTINCT
+            CAST(LExt.orig_loginname AS sys.SYSNAME) AS LoginName,
+            CAST(CAST(Base.oid AS INT) AS sys.varbinary(85)) AS sid,
```
sumitj824: NIT: column name should be in capital letters here.

contrib/babelfishpg_tsql/sql/babelfishpg_tsql.sql
```diff
@@ -3754,3 +3754,160 @@ BEGIN
...
+            CAST(LExt.default_database_name AS SYS.SYSNAME) AS DefDBName,
+            CAST(LExt.default_language_name AS SYS.SYSNAME) AS DefLangName,
+            CASE 
+                WHEN Ext.login_name IS NOT NULL AND Ext.login_name = LExt.rolname COLLATE database_default THEN CAST('YES' AS sys.char(5)) -- if there exists a mapping between user and logins, then we can say that there are users attached to this login
+                WHEN Db.owner COLLATE database_default = LExt.orig_loginname THEN CAST('YES' AS sys.char(5)) -- this is the case for superuser
+                ELSE CAST('NO' AS sys.char(5))
+            END AS AUser,
+            CAST('NO' AS sys.char(7)) AS ARemote -- Currently we do not support linking local logins to remote logins
```
sumitj824: NIT: The values of these columns should be in small letters.

sumitj824 (approved):

contrib/babelfishpg_tsql/sql/babelfishpg_tsql.sql
```diff
@@ -3754,3 +3754,160 @@ BEGIN
...
+    END
+
+    SET @current_username = LOWER(sys.suser_name());
+    SET @is_sysadmin = is_srvrolemember('sysadmin');
+    
+    IF @loginname IS NULL
+    BEGIN    
+        SELECT DISTINCT
+            CAST(LExt.orig_loginname AS sys.SYSNAME) AS LoginName,
+            CAST(CAST(Base.oid AS INT) AS sys.varbinary(85)) AS SID,
```
sumitj824: For sanity, lets use BIGINT cast here. The OID might not fit in INT range.

HarshLunagariya (approved):

jsudrik (approved):
Approved.

PR #3769, opened 2025-05-19
Fix loading of data for TEXT column (NULL allowed) if primary key is defined using non-default collation from PG endpoint  for logical database

ahmed-shameem:
### Description
If the babelfish table has a column whose type is anything other than text/ntext (eg: varchar, nvarchar etc) and the collation of the column is NOT same as babelfish database/server level collation and if the query is executed from PSQL endpoint, it will throw error saying it can not decide which collation to use.

If the same query is executed from TDS endpoint, it runs successfully.

This is because when the query is executed from PSQL endpoint -- the collation of varchar is server level collation. During resolution of collation at the parent node, the collation of left and right operand does NOT match and none of their collation matches with CLUSTER_COLLATION_OID (this returns DEFAULT_COLLATION_OID as the dialect is PG)

The query is working from TDS endpoint -- the collation of varchar is database level collation. During resolution of collation at the parent node, the collation of left and right operand does NOT match and collation of right operand matches with CLUSTER_COLLATION_OID (this returns database level collation as the dialect is TSQL)

This commit handles the case when the connection is from PG endpoint and psql_logical_babelfish_db_name GUC is set -- the case for logical/TSQL database. If both the aforementioned conditions are met, we will return babelfish server level collation irrespective of dialect (initially it was only for TDS connection).

### Issues Resolved

BABEL-5077

* **Use case based -**

* **Boundary conditions -**

* **Arbitrary inputs -**

* **Negative test cases -**

* **Performance tests -**

* **Tooling impact -**

* **Client tests -**

### Check List

### DIFF
--- a/contrib/babelfishpg_common/src/babelfishpg_common.c
+++ b/contrib/babelfishpg_common/src/babelfishpg_common.c
@@ -34,21 +34,6 @@ char	   *pltsql_server_collation_name = NULL;
 /* Dump and Restore */
 char	   *babelfish_restored_server_collation_name = NULL;
 
-/* Global cache variable */
-LogicalDBCache g_logical_db_cache = {NULL, false};
-
-/* Previous ProcessUtility hook if any */
-static ProcessUtility_hook_type prev_ProcessUtility = NULL;
-static void
-babelfishpg_logicalDB_ProcessUtility(PlannedStmt *pstmt,
-						const char *queryString,
-						bool readOnlyTree,
-						ProcessUtilityContext context,
-						ParamListInfo params,
-						QueryEnvironment *queryEnv,
-						DestReceiver *dest,
-						QueryCompletion *qc);
-
 const char *BabelfishTranslateCollation(
 										const char *collname,
 										Oid collnamespace,
@@ -165,14 +150,6 @@ _PG_init(void)
  	prev_get_like_collation_hook = get_like_collation_hook;
 	get_like_collation_hook = bbf_get_like_collation;
 
-	/* Initialize logical db name */
-	g_logical_db_cache.logical_db_name = NULL;
-	g_logical_db_cache.is_valid = false;
-
-	/* Save previous hook and install hook for logical db name */
-	prev_ProcessUtility = ProcessUtility_hook;
-	ProcessUtility_hook = babelfishpg_logicalDB_ProcessUtility;
-
 }
 void
 _PG_fini(void)
@@ -182,8 +159,6 @@ _PG_fini(void)
 	PreCreateCollation_hook = prev_PreCreateCollation_hook;
 	set_like_collation_hook = prev_set_like_collation_hook;
 	get_like_collation_hook = prev_get_like_collation_hook;
-	/* Restore previous hook */
-	ProcessUtility_hook = prev_ProcessUtility;
 }
 
 common_utility_plugin *
@@ -244,62 +219,4 @@ get_common_utility_plugin(void)
 		common_utility_plugin_var.tsql_numeric_get_typmod = &tsql_numeric_get_typmod;
 	}
 	return &common_utility_plugin_var;
-}
-
-/* Hook function to catch utility commands */
-static void
-babelfishpg_logicalDB_ProcessUtility(PlannedStmt *pstmt,
-						const char *queryString,
-						bool readOnlyTree,
-						ProcessUtilityContext context,
-						ParamListInfo params,
-						QueryEnvironment *queryEnv,
-						DestReceiver *dest,
-						QueryCompletion *qc)
-{
-	Node *parsetree = pstmt->utilityStmt;
-
-	switch (nodeTag(parsetree))
-	{
-		/* SET [ SESSION | LOCAL ] configuration_parameter */
-		case T_VariableSetStmt:
-		{
-			VariableSetStmt *stmt = (VariableSetStmt *) parsetree;
-			if (strcasecmp(stmt->name, "psql_logical_babelfish_db_name") == 0)
-			{
-				g_logical_db_cache.is_valid = false;
-			}
-			break;
-		}
-		/* ALTER DATABASE name SET configuration_parameter */
-		case T_AlterDatabaseSetStmt:
-		{
-			AlterDatabaseSetStmt *stmt = (AlterDatabaseSetStmt *) parsetree;
-			if (strcasecmp(stmt->setstmt->name, "psql_logical_babelfish_db_name") == 0)
-			{
-				g_logical_db_cache.is_valid = false;
-			}
-			break;
-		}
-		/* ALTER SYSTEM SET configuration_parameter */
-		case T_AlterSystemStmt:
-		{
-			AlterSystemStmt *stmt = (AlterSystemStmt *) parsetree;
-			if (strcasecmp(stmt->setstmt->name, "psql_logical_babelfish_db_name") == 0)
-			{
-				g_logical_db_cache.is_valid = false;
-			}
-			break;
-		}
-		default:
-			break;
-	}
-
-	/* Chain to previous hook if any */
-	if (prev_ProcessUtility)
-		prev_ProcessUtility(pstmt, queryString, readOnlyTree,
-						context, params, queryEnv, dest, qc);
-	else
-		standard_ProcessUtility(pstmt, queryString, readOnlyTree,
-							context, params, queryEnv, dest, qc);
 }
--- a/contrib/babelfishpg_common/src/babelfishpg_common.c
+++ b/contrib/babelfishpg_common/src/babelfishpg_common.c
@@ -20,7 +20,6 @@
 #include "typecode.h"
 #include "varchar.h"
 #include "datetimeoffset.h"
-#include "tcop/utility.h"
 
 common_utility_plugin common_utility_plugin_var = {NULL};
 static common_utility_plugin *get_common_utility_plugin(void);
--- a/contrib/babelfishpg_common/src/babelfishpg_common.c
+++ b/contrib/babelfishpg_common/src/babelfishpg_common.c
@@ -20,6 +20,7 @@
 #include "typecode.h"
 #include "varchar.h"
 #include "datetimeoffset.h"
+#include "tcop/utility.h"
 
 common_utility_plugin common_utility_plugin_var = {NULL};
 static common_utility_plugin *get_common_utility_plugin(void);
@@ -32,6 +33,21 @@ char	   *pltsql_server_collation_name = NULL;
 
 /* Dump and Restore */
 char	   *babelfish_restored_server_collation_name = NULL;
+
+/* Global cache variable */
+LogicalDBCache g_logical_db_cache = {NULL, false};
+
+/* Previous ProcessUtility hook if any */
+static ProcessUtility_hook_type prev_ProcessUtility = NULL;
+static void
+babelfishpg_logicalDB_ProcessUtility(PlannedStmt *pstmt,
+						const char *queryString,
+						bool readOnlyTree,
+						ProcessUtilityContext context,
+						ParamListInfo params,
+						QueryEnvironment *queryEnv,
+						DestReceiver *dest,
+						QueryCompletion *qc);
 
 const char *BabelfishTranslateCollation(
 										const char *collname,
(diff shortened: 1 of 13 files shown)


contrib/babelfishpg_common/src/collation.c
```diff
@@ -1243,12 +1243,23 @@ get_database_or_server_collation_oid_internal(bool missingOk)
...
 	}
+	
+	/* 
+	 * If DMS, we can return the server level collation
+	 * We assume that psql_logical_babelfish_db_name is a valid TSQL database name
+	 * Caller is reponsible to supply valid psql_logical_babelfish_db_name
+	 */
+	logical_babelfish_db_name = GetConfigOption("psql_logical_babelfish_db_name", true, true);
+	if (logical_babelfish_db_name)
+		return get_database_or_server_collation_oid_internal(false);
```
Deepesh125: This might be expensive operation to get logical_babelfish_db_name everytime on each statement. Can we somehow cache it while ensuring that any update to logical_babelfish_db_name through GUC operation invalidates that cache?

test/JDBC/expected/psql_logical_babelfish_db.out
```diff
@@ -311,6 +311,441 @@ go
...
+text#!#"sys"."varchar"
+NULL#!#unicode_cs_as
+~~END~~
+
+
+-- tsql
+USE master;
+GO
+
+DROP DATABASE dms_db_babel_5077;
```
Deepesh125, line 1298: just for completeness, can you add tests where predicate has the same column which is being updated and it also has PKEY. Cover the same for [n][var]char also

Deepesh125 (changes requested):

contrib/babelfishpg_common/src/collation.c
```diff
@@ -1243,12 +1243,23 @@ get_database_or_server_collation_oid_internal(bool missingOk)
...
+	const char* logical_babelfish_db_name;
 	if (sql_dialect == SQL_DIALECT_TSQL)
 	{
 		Oid db_coll = get_database_or_server_collation_oid_internal(false);	/* set and cache
 													 * database or server_collation_oid */
 		return db_coll;
 	}
+	
+	/* 
+	 * If DMS, we can return the server level collation
```
KushaalShroff: Lets avoid restricted keywords in OSS

KushaalShroff (changes requested):

contrib/babelfishpg_common/src/collation.c
ahmed-shameem: updated in latest rev

test/JDBC/expected/psql_logical_babelfish_db.out
ahmed-shameem: updated in latest rev

contrib/babelfishpg_common/src/collation.c
ahmed-shameem: updated in latest rev

contrib/babelfishpg_common/src/collation.c
```diff
@@ -1240,6 +1241,32 @@ get_database_or_server_collation_oid_internal(bool missingOk)
...
 }
 
+static const char *
+get_cached_logical_db_name(void)
+{
+	if (!g_logical_db_cache.is_valid)
+	{
+		MemoryContext oldcontext;
+		const char* logical_db = NULL;
+		oldcontext = MemoryContextSwitchTo(CacheMemoryContext);
```
KushaalShroff: Do we need CacheMemoryContext for GUCs?
I dont think it is required.

contrib/babelfishpg_common/src/babelfishpg_common.c
```diff
@@ -219,3 +245,61 @@ get_common_utility_plugin(void)
 	}
 	return &common_utility_plugin_var;
 }
+
+/* Hook function to catch utility commands */
+static void
+babelfishpg_logicalDB_ProcessUtility(PlannedStmt *pstmt,
```
KushaalShroff: This doesnt look clean at first glance. Can we cache it in tsql extension and use it accordingly?

KushaalShroff (changes requested):

contrib/babelfishpg_common/src/babelfishpg_common.c
```diff
@@ -149,6 +165,14 @@ _PG_init(void)
  	prev_get_like_collation_hook = get_like_collation_hook;
 	get_like_collation_hook = bbf_get_like_collation;
 
+	/* Initialize logical db name */
+	g_logical_db_cache.logical_db_name = NULL;
+	g_logical_db_cache.is_valid = false;
```
KushaalShroff: I think it will be better to modify assign_guc_hook for psql_logical_babelfish_db_name so that we can distinguish NULL and not set GUC.

KushaalShroff (changes requested):

contrib/babelfishpg_tsql/src/collation.c
```diff
@@ -2174,3 +2174,25 @@ patindex_ai_collations(PG_FUNCTION_ARGS)
...
+	/* Free old value if exists */
+	if (pltsql_psql_logical_babelfish_db_name)
+	{
+		guc_free(pltsql_psql_logical_babelfish_db_name);
+		pltsql_psql_logical_babelfish_db_name = NULL;
+	}
+
+	/* Set new value if not NULL */
+	if (newval)
+		pltsql_psql_logical_babelfish_db_name = guc_strdup(ERROR, newval);
```
Deepesh125: IIUC, this is not needed

contrib/babelfishpg_common/src/collation.c
ahmed-shameem: removed in latest rev

contrib/babelfishpg_common/src/babelfishpg_common.c
ahmed-shameem: updated in latest rev

contrib/babelfishpg_common/src/babelfishpg_common.c
ahmed-shameem: updated in latest rev

contrib/babelfishpg_tsql/src/collation.c
ahmed-shameem: Ack, updated in latest rev

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -5975,6 +5975,14 @@ default_collation_for_builtin_type(Type typ, bool handle_pg_type)
...
 		oid = CLUSTER_COLLATION_OID();
 	}
+	else if (sql_dialect == SQL_DIALECT_PG && pltsql_psql_logical_babelfish_db_name)
+	{
+		/* 
+		 * Check whether the type is collatable or not
+		 * If yes, return server level collation or return InvalidOid
+		 */
+		return OidIsValid(typtup->typcollation) ? CLUSTER_COLLATION_OID() : InvalidOid;
+	}
```
Deepesh125: we can move this section out of this if condition to later before where we are handling case for TEXT. Also, add comment explaining when and why are we doing this

contrib/babelfishpg_tsql/src/collation.c
```diff
@@ -2174,3 +2174,21 @@ patindex_ai_collations(PG_FUNCTION_ARGS)
...
+
+
+void
+assign_psql_logical_babelfish_db_name(const char *newval, void *extra)
+{
+	/* 
+	 * No need to tell babelfishpg_common that logical db name has been set if dialect is TSQL
+	 * because the psql_logical_babelfish_db_name is to be used for PG dialect only
+	 */
+	if (sql_dialect == SQL_DIALECT_PG)
```
Deepesh125: I am not 100% sure if this is really needed. Can you please explain use case/example which may break if we dont do thiss

contrib/babelfishpg_tsql/src/collation.c
```diff
@@ -2174,3 +2174,21 @@ patindex_ai_collations(PG_FUNCTION_ARGS)
...
+	{
+		/* Initialise collation callbacks */
+		init_and_check_collation_callbacks();
+
+		/* let babelfishpg_common know that psql_logical_babelfish_db_name has been updated */
+		(*collation_callbacks_ptr->set_logical_db_name_cache) (newval);
```
KushaalShroff, line 2193: Given this logic, shall we check for valid values of newval?

KushaalShroff (changes requested):

contrib/babelfishpg_tsql/src/collation.c
ahmed-shameem: Are you indicating to check whether the logical database name exists or not?

contrib/babelfishpg_tsql/src/collation.c
ahmed-shameem: The ```is_logical_db_name_set``` for TSQL endpoint as well. Now, for foreign constraint the dialect gets set to PG, hence during collation resolution in merge_collation_state, ```CLUSTER_COLLATION_OID()``` returns babelfish server/db level collation. Whereas, it expects ```DEFAULT_COLLATION_OID```. This results in collation conflict and we get the following error during insertion of row:
```
1> INSERT INTO dms_db_babel_5077_t7(dms_db_babel_5077_t7_col2) VALUES ('unicode_cs_as');
2> go
could not determine which collation to use for string comparison
```

By setting ```is_logical_db_name_set``` for PG dialect only, we ensure that for foreign constraint  case as well we return ```DEFAULT_COLLATION_OID``` from ```CLUSTER_COLLATION_OID()``` (which was the behaviour initially)

contrib/babelfishpg_tsql/src/collation.c
ahmed-shameem: IMO, that's redundant as even if the logical db name is incorrect, we won't find any relations, functions, view, procedures etc for that incorrect/invalid logical db

contrib/babelfishpg_tsql/src/hooks.c
ahmed-shameem: Updated in latest rev

contrib/babelfishpg_tsql/src/collation.c
Deepesh125: FKEY check will always be done under PG_DIALECT so same question remains. Why is this needed? can you actually show whole example you are referring?

Deepesh125 (changes requested):

contrib/babelfishpg_tsql/src/collation.c
ahmed-shameem: This test will fail: https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/3769/files#diff-baa40afc81f5544d3a1b28c72d43ab458fb930c373f4cc5687ecbd45d509726eR169

ahmed-shameem:
```
CREATE TABLE dms_db_babel_5077_t6(dms_db_babel_5077_t6_col1 text NULL, dms_db_babel_5077_t6_col2 [varchar](44) collate BBF_Unicode_General_CS_AS primary key);
GO

INSERT INTO dms_db_babel_5077_t6(dms_db_babel_5077_t6_col2) VALUES ('unicode_cs_as');
GO

CREATE TABLE dms_db_babel_5077_t7(dms_db_babel_5077_t7_col1 text NULL, dms_db_babel_5077_t7_col2 [varchar](44) collate BBF_Unicode_General_CS_AS references dms_db_babel_5077_t6(dms_db_babel_5077_t6_col2));
GO

1> INSERT INTO dms_db_babel_5077_t7(dms_db_babel_5077_t7_col2) VALUES ('unicode_cs_as');
2> go
could not determine which collation to use for string comparison
```

ahmed-shameem: Logically the guc should be used from PG endpoint only. If the GUC is modified from TDS endpoint, it should not affect any operation. This change restricts that.

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -5984,8 +5984,13 @@ default_collation_for_builtin_type(Type typ, bool handle_pg_type)
 	 * Special handling for PG datatypes such as TEXT because Babelfish does not define sys.TEXT.
 	 * This is required as Babelfish currently does not handle collation of String Const node correctly.
 	 * TODO: Fix the handling of the collation for String Const node.
+	 * 
+	 * When pltsql_psql_logical_babelfish_db_name is set, return CLUSTER_COLLATION_OID() for TEXT (e.g: string literals)
```
Deepesh125: also mention that psql_logical_babelfish_db_name can only be set inside pg dialect.

contrib/babelfishpg_tsql/src/collation.c
Deepesh125: will it be better if we check tds conn?

contrib/babelfishpg_tsql/src/hooks.c
ahmed-shameem: updated comment in latest rev

contrib/babelfishpg_tsql/src/collation.c

Deepesh125 (approved):

KushaalShroff (approved):

PR #3777, opened 2025-05-22
Parse cursor queries with antlr before executing

sharathbp:
### Description
Cursor queries are not parsed with antlr parser before sending it to Postgres parser, which leads to syntax error if there is any psql related keywords in the query. Antlr parser transforms the query by delimit the psql keywords in the query. Adding antlr parser logic for CURSOR queries in Pabelfish before sending it to Postgres for execution.

### Issues Resolved

BABEL-5743

* **Use case based -**
Yes

* **Boundary conditions -**
Yes

* **Arbitrary inputs -**
Yes

* **Negative test cases -**
Yes

NA

NA

* **Performance tests -**
NA

* **Tooling impact -**
NA

* **Client tests -**
NA

### Check List


contrib/babelfishpg_tsql/src/cursor.c
```diff
@@ -1449,6 +1449,24 @@ execute_sp_cursoropen_common(int *stmt_handle, int *cursor_handle, const char *s
...
+	PLtsql_stmt_execsql *parse_result;
+	PLtsql_function *function;
+	char *stmt_copy;
+
+	/*
+	 * We need to parse the TSQL statements using antlr parser before passing it to SPI_prepare_cursor. 
+	 * Antlr parser will converts TSQL query into PSQL syntax and properly handles idenfier delimiters,
+	 * allowing PostgreSQL reserved words to be used as column aliases.
+	 */
+	if (stmt != NULL && sql_dialect == SQL_DIALECT_TSQL)
```
tanscorpio7: ANTLR can be a pretty heavy parser, so we maintain a session level cache for the ANTLR parse result for a procedure's body. Again invoking ANTLR here for sp_cursor procs would mean we always ANTLR parse during runtime.
Can we avoid that by doing this in the first parse of the statement itself ?

tanscorpio7 (changes requested):

contrib/babelfishpg_tsql/src/cursor.c
```diff
@@ -1449,6 +1449,29 @@ execute_sp_cursoropen_common(int *stmt_handle, int *cursor_handle, const char *s
...
 	MemoryContext oldcontext;
 	MemoryContext savedPortalCxt;
+	ANTLR_result result;
+	
+	/*
+	 * We need to parse the TSQL statements using antlr parser before passing it to SPI_prepare_cursor. 
+	 * Antlr parser will converts TSQL query into PSQL syntax and properly handles idenfier delimiters,
+	 * allowing PostgreSQL reserved words to be used as column aliases.
+	 */
+	if (prepare && stmt != NULL && sql_dialect == SQL_DIALECT_TSQL)
```
tanscorpio7, line 1460: Can we merge this block with the prepare block below ? Also why do we need dialect check here ?

contrib/babelfishpg_tsql/src/cursor.c
```diff
@@ -1449,6 +1449,29 @@ execute_sp_cursoropen_common(int *stmt_handle, int *cursor_handle, const char *s
...
+		result = antlr_parser_cpp(stmt);
+		
+		if (!result.success)
+			report_antlr_error(result);
+			
+		/* Skip if NULL query was passed. */
+		if (pltsql_parse_result->body)
+		{
+			PLtsql_expr *sqlstmt = ((PLtsql_stmt_execsql *) lsecond(pltsql_parse_result->body))->sqlstmt;
+			if (sqlstmt)
```
tanscorpio7, line 1473: Can you share an example where sqlstmt is Null but input stmt was not null ?

tanscorpio7 (changes requested):

contrib/babelfishpg_tsql/src/cursor.c
sharathbp: Didn't merge it because we want to parse before creating SPI connection because if parse fails SPI connection is not required. Since SPI connection is used outside prepare if statement added it separately before SPI connection.

contrib/babelfishpg_tsql/src/cursor.c
sharathbp: If the parsed stmt is not PLtsql_stmt_execsql then it will be null
For eg: 
GOTO label
DECLARE @var INT;

contrib/babelfishpg_tsql/src/cursor.c
sharathbp: Dialect I added as safety as antlr parser should be used only for TSQL context right

KushaalShroff: I dont think we will ever reach here in non TSQL dialect, can you please confirm?

contrib/babelfishpg_tsql/src/cursor.c
```diff
@@ -1449,6 +1449,29 @@ execute_sp_cursoropen_common(int *stmt_handle, int *cursor_handle, const char *s
...
+	ANTLR_result result;
+	
+	/*
+	 * We need to parse the TSQL statements using antlr parser before passing it to SPI_prepare_cursor. 
+	 * Antlr parser will converts TSQL query into PSQL syntax and properly handles idenfier delimiters,
+	 * allowing PostgreSQL reserved words to be used as column aliases.
+	 */
+	if (prepare && stmt != NULL && sql_dialect == SQL_DIALECT_TSQL)
+	{
+		pltsql_curr_compile = NULL;
```
KushaalShroff: Why do we need to set this locally?

contrib/babelfishpg_tsql/src/cursor.c
KushaalShroff: Shouldnt we be checking for stmt type then?

test/JDBC/expected/TestCursorFetchNext.out
```diff
@@ -64,6 +64,34 @@ int#!#smallint#!#bigint#!#tinyint#!#bit
 #cursor#!#fetch#!#next
 #cursor#!#fetch#!#afterlast
 #cursor#!#fetch#!#next
+cursor#!#close
+~~SUCCESS~~
+cursor#!#open#!#SELECT a year, b month, c quarter FROM test_cursors_fetch_next#!#TYPE_SCROLL_INSENSITIVE#!#CONCUR_READ_ONLY#!#CLOSE_CURSORS_AT_COMMIT
```
KushaalShroff, line 69: Can we have more testing done:
Maybe SQL query Testing and Negative testing with different command types in body for RPC and SQL

KushaalShroff (changes requested):

contrib/babelfishpg_tsql/src/cursor.c
sharathbp: Currently when we cast to PLtsql_stmt_execsql if its not execsql it returns null and ignored. So stmt check is not required.

contrib/babelfishpg_tsql/src/cursor.c
sharathbp: Verified its only getting called through ProcessTDSRequest->ProcessRPCRequest->HandleSPCursorOpenCommon

KushaalShroff: Then we dont need the TSQL Dialect check, right?

KushaalShroff (changes requested):

contrib/babelfishpg_tsql/src/pl_gram.y
```diff
@@ -7264,10 +7265,14 @@ parse_datatype(const char *string, int location)
 
 	/* Restore former ereport callback */
 	error_context_stack = syntax_errcontext.previous;
+	
+	if (pltsql_curr_compile)
+		collation = pltsql_curr_compile->fn_input_collation;
```
KushaalShroff, line 7270: Is there a possibility that pltsql_curr_compile is NULL here? Thats serious isn't it?

KushaalShroff (changes requested):

contrib/babelfishpg_tsql/src/pl_gram.y
sharathbp: If its NULL then we will not access it. It will be the assigned value above InvalidOid. Before it was causing seg fault because we were accessing it without null check below so added this.

contrib/babelfishpg_tsql/src/cursor.c
sharathbp: Yeah I can remove that.

tanscorpio7:
Please also add test cases for sp_cursor procedures. `sp_cursoropen`, `sp_cursorexecute`, etc. Thanks.

PR #3789, opened 2025-05-29
fix for DB user drop query when granted some permission

pranavJ23:
### Description

Problem : If a user having some permissions over an object then we are not able to drop the user, which is not same as TSQL behaviour.Note:  in postgres we can't drop a role before dropping it's dependencies.
Fix: we have to remove the grants(permission) given to the user via shared_dependency_acl before dropping the user.

- Added a check if user is grantor of some permissions then it can't be dropped. 
- After dropping the user remove the entries corresponding to that user from schema permission catalog.
- To implement we have created a new function bbf_shdep_drop_owned_dependent_acl (forked from pg function shdepDropOwned).

### Issues Resolved
[BABEL-3026]

### Check List


test/JDBC/expected/restrict_drop_user_role.out
```diff
@@ -1,4 +1,22 @@
...
+go
+create user u2 for login l2
+go
+create login l3 with password = '123'
+go
+create user u3 for login l3
+go
+create table grant_tbl(a int)
+go
+grant select on grant_tbl to u2 with grant option
```
shalinilohia50, line 18: Please add tests for permission on schema.

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -4109,6 +4109,49 @@ remove_entry_from_bbf_schema_perms(const char *schema_name,
 	table_close(bbf_schema_rel, RowExclusiveLock);
 }
 
+/*
+ * Removes a row from the catalog BABELFISH_SCHEMA_PERMISSIONS when a user is dropped.
+ */
```
tanscorpio7: /* ... */ single line comments

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -4109,6 +4109,49 @@ remove_entry_from_bbf_schema_perms(const char *schema_name,
...
+	{
+		/* Delete the entry */
+		CatalogTupleDelete(bbf_schema_rel, &tuple_bbf_schema->t_self);
+		tuple_bbf_schema = systable_getnext(scan);
+	}
+
+	systable_endscan(scan);
+	table_close(bbf_schema_rel, RowExclusiveLock);
+}
+
```
tanscorpio7: Can we modify and reuse `remove_entry_from_bbf_schema_perms` ?

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -2407,6 +2411,129 @@ validateUserAndRole(char *name)
 						errmsg("'%s' is not a valid name because it contains invalid characters.", name)));
 }
 
+/* Revoke the permissions from an object present in Dependent ACL list within the same database */
+void
+bbf_shdep_drop_owned_dependent_acl(Oid roleid, DropBehavior behavior)
```
tanscorpio7: Should we move this function to rolecmds.c ?

Please also add a comment explaining we forked this function from shdepDropOwned and why.

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -2407,6 +2411,129 @@ validateUserAndRole(char *name)
...
+								/* no need to remove if grantor */
+								if (ai->ai_grantor == roleid)
+									return;
+							}
+				}
+				RemoveRoleFromObjectACL(roleid,
+										sdepForm->classid,
+										sdepForm->objid);
+				/* Update the catalog after deleting the user */
+				remove_user_entry_from_bbf_schema_perms(roleid);
```
tanscorpio7: Why do we need to do this here ? There should be an existing code for handling this already ?

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -2407,6 +2411,129 @@ validateUserAndRole(char *name)
...
+					 */
+					const AclItem *acldat;
+					old_acl = DatumGetAclPCopy(aclDatum);
+					acldat = ACL_DAT(old_acl);
+						for (int i = 0; i < ACL_NUM(old_acl); i++)
+							{
+								const AclItem *ai = &acldat[i];
+								/* no need to remove if grantor */
+								if (ai->ai_grantor == roleid)
+									return;
```
tanscorpio7: How do we cleanup opened resources in this case ?

Also indenation seems to be off for this block.

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -2407,6 +2411,129 @@ validateUserAndRole(char *name)
...
+							}
+				}
+				RemoveRoleFromObjectACL(roleid,
+										sdepForm->classid,
+										sdepForm->objid);
+				/* Update the catalog after deleting the user */
+				remove_user_entry_from_bbf_schema_perms(roleid);
+				ReleaseSysCache(tup);
+				if(!old_acl)
+					pfree(old_acl);
```
tanscorpio7: Static code analyer says this is uninitialzied in some case.

tanscorpio7 (changes requested):

contrib/babelfishpg_tsql/src/pl_handler.c
pranavJ23: We are removing the rows from schema when grantee is user(which is being dropped).
We were not doing this earlier because earlier when user is grantee it can't be dropped

contrib/babelfishpg_tsql/src/pl_handler.c
pranavJ23: Thanks for pointing this out.
To cleanup, we can have a bool variable set to true instead of direct return and call RemoveRoleFromObjectACL only when that variable( let's say user_is_grantor is false).

contrib/babelfishpg_tsql/src/catalog.c
pranavJ23: We can modify the function : remove_entry_from_bbf_schema_perms but in that case we will need 2 extra inputs, one being user oid and other being a flag to determine, which deletion operation to perform

tanscorpio7: Looking at current `remove_entry_from_bbf_schema_perms ` I think the merging them both makes it less readable, you may revert back to having a different function if you want to.

contrib/babelfishpg_tsql/src/rolecmds.c
```diff
@@ -3313,3 +3315,140 @@ drop_db_owner_related_roles(Oid roleid, const char* rolname)
...
+					if(!old_acl)
+						pfree(old_acl);
+				}
+				if(!user_is_grantor)
+				{
+					RemoveRoleFromObjectACL(roleid,
+											sdepForm->classid,
+											sdepForm->objid);
+					/* Update the catalog after deleting the user */
+					remove_entry_from_bbf_schema_perms(NULL, NULL, NULL, NULL, roleid, true);
```
tanscorpio7: Should this be outside the while loop ?

tanscorpio7 (changes requested):

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -4109,6 +4109,47 @@ remove_entry_from_bbf_schema_perms(const char *schema_name,
 	table_close(bbf_schema_rel, RowExclusiveLock);
 }
 
+/* Removes a row from the catalog BABELFISH_SCHEMA_PERMISSIONS when a user is dropped. */
```
tanscorpio7: Should this be `Removes all rows for babelfish database user ... ` ?

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -4109,6 +4109,47 @@ remove_entry_from_bbf_schema_perms(const char *schema_name,
...
+void
+remove_user_entry_from_bbf_schema_perms(Oid user_oid)
+{
+	Relation		bbf_schema_rel;
+	HeapTuple		tuple_bbf_schema;
+	ScanKeyData		scanKey[1];
+	SysScanDesc		scan;
+	const char		*user_name = GetUserNameFromId(user_oid, true);
```
tanscorpio7: `GetUserNameFromId` should be after `/* Return if the Oid is invalid */`

contrib/babelfishpg_tsql/src/rolecmds.c
```diff
@@ -3313,3 +3315,142 @@ drop_db_owner_related_roles(Oid roleid, const char* rolname)
...
+	{
+		Form_pg_shdepend sdepForm = (Form_pg_shdepend) GETSTRUCT(tuple);
+		/*
+		 * We only operate on shared objects and objects in the current
+		 * database
+		 */
+		if (sdepForm->dbid != MyDatabaseId &&
+			sdepForm->dbid != InvalidOid)
+			continue;
+		if (sdepForm->deptype == SHARED_DEPENDENCY_ACL)
```
tanscorpio7: We should quick exit if we find any other deptype except `SHARED_DEPENDENCY_ACL`. The object is no longer safe to drop. Also exit if any other database has dependencies.

tanscorpio7 (changes requested):

tanscorpio7 (approved):

shalinilohia50 (approved):

shardgupta (approved):

PR #3568, opened 2025-03-11
[WIP]Enhance Test Coverage for Money/SmallMoney Datatypes

shah-nirmit:
### Description

[Describe what this change achieves - Guidelines below (please delete the guidelines after writing the PR description)]

> 1. *What* is the change? This is best described in terms of “Currently, Babelfish does X. With this change it now does Y.” Think of “What *did* it *used* to do?” and “What *does* it do *now*?”
2. *Why* was the change made? What drove our desire to put effort into the change?
3. *How* was the code changed should only appear for large commits. This can serve as a rough roadmap to what’s contained in the commit. It should be very high level; if it’s directly referencing code it’s probably too detailed. It’s also critical that this section of a commit message does not try to replace proper code documentation (ie, block comments or README files). Generally, this section should only appear if the commit itself is large enough that it’s helpful to provide a roadmap to someone looking at the commit.
4. The last descriptive piece is the “title” for the commit: the very first line of the commit message, which should typically be less than 80 characters. A good title is *critical*, because it’s the only thing that shows up in places like the Github commit listing. No one’s got time to read through full commit messages when trying to find a single commit out of dozens.

### Issues Resolved

[List any issues this PR will resolve]

### Test Scenarios Covered ###
* **Use case based -**

* **Boundary conditions -**

* **Arbitrary inputs -**

* **Negative test cases -**

* **Minor version upgrade tests -**

* **Major version upgrade tests -**

* **Performance tests -**

* **Tooling impact -**

* **Client tests -**

### Check List
- [ ] Commits are signed per the DCO using --signoff 

By submitting this pull request, I confirm that my contribution is under the terms of the Apache 2.0 and PostgreSQL licenses, and grant any person obtaining a copy of the contribution permission to relicense all or a portion of my contribution to the PostgreSQL License solely to contribute all or a portion of my contribution to the PostgreSQL open source project.

For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/babelfish-for-postgresql/babelfish_extensions/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #3797, opened 2025-05-30
Address issues with arithmetic, mathematical functions for money/smallmoney

ayushdsh:
### Description

Handling overflow and incorrect outputs in case of certain money and smallmoney arithmetic and math functions.

There were certain issues found both w.r.t overflow handling and certain functions and operations giving incorrect result for money and smallmoney datatype. This commit addresses those issues.

Issue #1: Overflow of smallmoney arithmetic leading to TDS hang.
Issue #2: Unhandled overflow cases leading to incorrect output.
Issue #3: Fixeddecimal operators being resolved for smallmoney functions and operations leading to incorrect results.

In this commit, we have added new functions for all smallmoney operators and fixed certain overflows with money (fixeddecimal).

New operators added - 
1. arithmetic operators for bit
    a. bit and float operators
    b. bit and smallmoney operators
    c. bit and bit operators.

The reason to introduce bit,bit and bit,float operators is because once we introduced bit,smallmoney; certain operators did not remain unique - like bit, bit and bit, float.
Previously, their output was datetime which was also not valid. Hence, we've corrected the output and introduced these operators.

New functions added - 
1. All arithmetic functions for smallmoney
    a. smallmoney and int2, int4 and int8
    b. smallmoney and smallmoney
2. power, ceiling and floor functions for fixeddecimal

### Update
* we will not be introducing bit, bit operators

### Issues Resolved

BABEL-5745, BABEL-5757, BABEL-5756, BABEL-5747, BABEL-5754

Note: testcases are imported from [PR-3568](https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/3568)
* **Use case based -**
    * All arithmetic and math functions for money and smallmoney
        * with all types of operands and vice-versa 

* **Boundary conditions -**
    * All boundary conditions for overflow

* **Arbitrary inputs -**
    * Certain arbritrary input tests added

* **Negative test cases -**
    * Overflow conditions checked which were previously causing protocol hang.

    * dependency tests for all the new operators and functions

    * dependency tests for all the new operators and functions

* **Performance tests -**
    * NA

* **Tooling impact -**
    * NA

* **Client tests -**
    * NA

### Check List


contrib/babelfishpg_money/fixeddecimal.c
```diff
@@ -1848,23 +1817,37 @@ Datum
...
+     * 64 bit, then the result will definitely overflow
+	 */
+    if (adder != (int64) adder)
+    {
+        ereport(ERROR,
+				(errcode(ERRCODE_NUMERIC_VALUE_OUT_OF_RANGE),
+				 errmsg("fixeddecimal out of range")));
+        /* ensure compiler realizes we mustn't reach the division (gcc bug) */
+        PG_RETURN_NULL();
+    }
```
ayushdsh: This condition and the below condition where we check overflow are required in the following scenario - 
When arg2 = INT64_MAX, adder itself overflows and the existing overflow check condition assumes that the adder was always negative. Hence, we first check if the adder overflows and terminate if it does.

contrib/babelfishpg_money/fixeddecimal.c
```diff
@@ -1887,14 +1870,28 @@ fixeddecimalint8mi(PG_FUNCTION_ARGS)
...
+     * 64 bit, then the result will definitely overflow
+	 */
+    if (subtractor != (int64) subtractor)
+    {
+        ereport(ERROR,
+				(errcode(ERRCODE_NUMERIC_VALUE_OUT_OF_RANGE),
+				 errmsg("fixeddecimal out of range")));
+        /* ensure compiler realizes we mustn't reach the division (gcc bug) */
+        PG_RETURN_NULL();
+    }
```
ayushdsh: SImilarly, if arg2 = INT64_MAX, this leads to subtractor overflow and becomes -ve and the overflow check is not correct because it assumes that the arg2 is -ve.

contrib/babelfishpg_money/fixeddecimal.c
```diff
@@ -2096,6 +2135,19 @@ int8fixeddecimaldiv(PG_FUNCTION_ARGS)
...
+     in T-SQL
+    */
+    if (arg1 > FIXEDDECIMAL_MAX)
+    {
+        ereport(ERROR,
+				(errcode(ERRCODE_NUMERIC_VALUE_OUT_OF_RANGE),
+				 errmsg("fixeddecimal out of range")));
+		/* ensure compiler realizes we mustn't reach the division (gcc bug) */
+		PG_RETURN_NULL();
+    }
```
ayushdsh: This is a TSQL behaviour which we did not respect previously. This condition takes care of the same.

test/JDBC/expected/babel_varbinary-vu-verify.out
```diff
@@ -1162,8 +1162,8 @@ ORDER BY
     bit_binary_add;
 GO
 ~~START~~
-numeric#!#datetime#!#datetime#!#bigint#!#int#!#smallint#!#smallint#!#money#!#datetime#!#varchar#!#datetime#!#bigint
-100.50#!#1900-04-11 12:00:00.0#!#1900-04-11 12:00:00.0#!#1000#!#100#!#10#!#1#!#100.5000#!#1900-01-01 00:00:00.0#!#0A1A2A3A4-B1B2-C1C2-D1D2-E1E2E3E4E5E6#!#<NULL>#!#0
+numeric#!#datetime#!#smallmoney#!#bigint#!#int#!#smallint#!#smallint#!#money#!#bit#!#varchar#!#datetime#!#bigint
+100.50#!#1900-04-11 12:00:00.0#!#100.5000#!#1000#!#100#!#10#!#1#!#100.5000#!#0#!#0A1A2A3A4-B1B2-C1C2-D1D2-E1E2E3E4E5E6#!#<NULL>#!#0
```
ayushdsh: here the changes are because we introduced new operators for bit. Now, bit + smallmoney results into smallmoney, which previously resulted into datetime and bit + bit results into bit which previously resulted into datetime.

contrib/babelfishpg_common/sql/upgrades/babelfish_common_helper--5.2.0--5.3.0.sql
```diff
@@ -133,6 +133,458 @@ BEGIN
...
+
+DROP OPERATOR IF EXISTS sys.+ (sys.SMALLMONEY, sys.SMALLMONEY);
+CREATE OPERATOR sys.+ (
+    LEFTARG    = sys.SMALLMONEY,
+    RIGHTARG   = sys.SMALLMONEY,
+    COMMUTATOR = +,
+    PROCEDURE  = smallmoneypl
+);
+
+DROP OPERATOR IF EXISTS sys.- (sys.SMALLMONEY, sys.SMALLMONEY);
```
tanscorpio7: An operator can have dependant objects which would make dropping an operator not possible.
If we are just creating new operators that did not already exists, then we should do it inside a do block
i.e. IF operator does not exist for sometype to sometype then only create.

test/JDBC/input/money_smallmoney-vu-verify.sql
```diff
@@ -0,0 +1,2845 @@
+-- Create test tables
```
tanscorpio7: We will also need to create prepare and verify test file for this. Upgrade tests should include 
1) Create views in prepare to cover all operators we are fixing.
2) Performance - create a table in prepare with money small money datatype. in verify ensure that index scan is picked for queries on these tables when the predicate has this operator. We can also keep changing the type on one end of the operator.
```
create table t (id int small money)
create index idx on t(id)

SELECT * FROM t WHERE id (some_operator) (different data types)
```

tanscorpio7 (changes requested):

contrib/babelfishpg_money/fixeddecimal.c
```diff
@@ -1892,8 +1874,8 @@ Datum
 fixeddecimalint8mi(PG_FUNCTION_ARGS)
 {
 	int64		arg1 = PG_GETARG_INT64(0);
-	int64		subtractor = PG_GETARG_INT64(1) * FIXEDDECIMAL_MULTIPLIER;
-	int64		result;
+	int128		subtractor = (int128) PG_GETARG_INT64(1) * FIXEDDECIMAL_MULTIPLIER;
+	int128		result;
```
tanyagupta17: While returning this result at end, we should not be relying on `PG_RETURN_INT64`, so please either have an intermediate result or return like this  :
`PG_RETURN_INT64(result); -> PG_RETURN_INT64((int64)result); `

ayushdsh: Interesting; I am just trying to understand this. Any example where the pg function returns an incorrect result?

test/JDBC/input/money_smallmoney-vu-verify.sql
ayushdsh: Great points. I've added both kinds of tests.
1. Added dependency tests - create views/funcs in prepare and querying in verify
2. added the perf tests; and we can see that planner picks the index

contrib/babelfishpg_common/sql/bit.sql
```diff
@@ -504,3 +504,100 @@ WITH FUNCTION sys.varchar2bit(sys.VARCHAR) AS IMPLICIT;
...
+            RETURN (SELECT sys.int4fixeddecimalmi(1, $2));
+        END IF;
+    END;
+$$ LANGUAGE plpgsql IMMUTABLE STRICT PARALLEL SAFE;
+
+CREATE OPERATOR sys.+ (
+    LEFTARG    = sys.BIT,
+    RIGHTARG   = sys.SMALLMONEY,
+    COMMUTATOR = +,
+    PROCEDURE  = bitsmallmoneypl
```
tanscorpio7: Should we scheama qualify `bitsmallmoneypl` ?

contrib/babelfishpg_common/sql/bit.sql
```diff
@@ -504,3 +504,100 @@ WITH FUNCTION sys.varchar2bit(sys.VARCHAR) AS IMPLICIT;
...
+    RIGHTARG   = sys.BIT,
+    COMMUTATOR = +,
+    PROCEDURE  = bitpl
+);
+
+CREATE OPERATOR sys.- (
+    LEFTARG    = sys.BIT,
+    RIGHTARG   = sys.BIT,
+    PROCEDURE  = bitmi
+);
```
tanscorpio7: 1. Does SQL Server support (bit +/- bit) ?
2. Do we need handle other operators for BIT & SMALLMONEY ? like multiplication etc.

contrib/babelfishpg_common/sql/bit.sql
```diff
@@ -504,3 +504,100 @@ WITH FUNCTION sys.varchar2bit(sys.VARCHAR) AS IMPLICIT;
 
 CREATE CAST (bool AS sys.BIT)
 WITHOUT FUNCTION AS IMPLICIT;
+
+CREATE FUNCTION sys.bitsmallmoneypl(sys.BIT, sys.SMALLMONEY)
```
tanscorpio7, line 510: Are we creating these wrappers to fix the return type ? We can define this as a C LANGUAGE and reuse whatever C function is used in `sys.int4fixeddecimalpl`

contrib/babelfishpg_tsql/sql/sys_functions.sql
```diff
@@ -2450,22 +2456,21 @@ SELECT
...
-LANGUAGE SQL IMMUTABLE PARALLEL SAFE;
+LANGUAGE SQL IMMUTABLE PARALLEL SAFE RETURNS NULL ON NULL INPUT;
 GRANT EXECUTE ON FUNCTION sys.sign(INT) TO PUBLIC;
 
 CREATE OR REPLACE FUNCTION sys.sign(IN arg SMALLINT) RETURNS INT AS
 $BODY$
 SELECT sys.sign(arg::INT);
 $BODY$
-LANGUAGE SQL IMMUTABLE PARALLEL SAFE;
+LANGUAGE SQL IMMUTABLE PARALLEL SAFE RETURNS NULL ON NULL INPUT;
```
tanscorpio7: I think we have used `STRICT` instead of `RETURNS NULL ON NULL INPUT`. They can be used interchangebly but let us stick to existing format.

contrib/babelfishpg_money/fixeddecimal.c
```diff
@@ -2099,6 +2143,19 @@ int8fixeddecimaldiv(PG_FUNCTION_ARGS)
...
+     in T-SQL
+    */
+    if (arg1 > FIXEDDECIMAL_MAX)
+    {
+        ereport(ERROR, 
+                (errcode(ERRCODE_NUMERIC_VALUE_OUT_OF_RANGE),
+                 errmsg("fixeddecimal out of range")));
+        /* ensure compiler realizes we mustn't reach the division (gcc bug) */
+        PG_RETURN_NULL();
+    }
```
tanscorpio7: Comment /* ... */

Also some indentation fixes here and in some other places in the PR.

contrib/babelfishpg_common/sql/money/fixeddecimal--1.1.0_base_parallel.sql
```diff
@@ -1641,22 +1641,22 @@ LANGUAGE C IMMUTABLE STRICT PARALLEL SAFE;
...
+AS 'babelfishpg_money', 'smallmoneymul'
 LANGUAGE C IMMUTABLE STRICT PARALLEL SAFE;
 
 CREATE FUNCTION sys.fixeddecimaldiv(sys.SMALLMONEY, sys.SMALLMONEY)
 RETURNS sys.SMALLMONEY
-AS 'babelfishpg_money', 'fixeddecimaldiv'
+AS 'babelfishpg_money', 'smallmoneydiv'
 LANGUAGE C IMMUTABLE STRICT PARALLEL SAFE;
 
 CREATE FUNCTION sys.fixeddecimalmod(sys.SMALLMONEY, sys.SMALLMONEY)
```
tanscorpio7, line 1666: We do not need to handle these operators ?

contrib/babelfishpg_money/smallmoney.c
```diff
@@ -0,0 +1,1073 @@
...
+     * Overflow check. If the result cannot be converted back
+     * to a 32 bit result, then we can say that there is a 
+     * overflow
+     */
+    if (result != (int32) result)
+        ereport(ERROR,
+                (errcode(ERRCODE_NUMERIC_VALUE_OUT_OF_RANGE),
+                 errmsg("smallmoney out of range")));
+#endif							
+    PG_RETURN_INT32(result);
```
tanscorpio7: Small money and money are both 8 bytes data types in babelfish. Please check if we should be returning INT64 inside C functions for small money.

tanscorpio7 (changes requested):

contrib/babelfishpg_money/fixeddecimal.c
```diff
@@ -1863,23 +1837,37 @@ Datum
 fixeddecimalint8pl(PG_FUNCTION_ARGS)
 {
 	int64		arg1 = PG_GETARG_INT64(0);
-	int64		adder = PG_GETARG_INT64(1) * FIXEDDECIMAL_MULTIPLIER;
-	int64		result;
+	int128		adder = (int128) PG_GETARG_INT64(1) * FIXEDDECIMAL_MULTIPLIER;
```
tanscorpio7: Let us also cast `FIXEDDECIMAL_MULTIPLIER` to int128 just to be safe from any compiler tricks.

contrib/babelfishpg_money/fixeddecimal.c
tanscorpio7: PG_RETURN.. just cast the pointer to a generic pointer (Datum) it does not actually cast input type to execpted type. So in this case we are mostly returning 16 bytes when the correct size should have been 8. Please confirm this by temporarily inserting the return value of this function into money column and then selecting from it.

contrib/babelfishpg_money/fixeddecimal.c
```diff
@@ -2029,24 +2058,39 @@ int8fixeddecimalpl(PG_FUNCTION_ARGS)
...
 		ereport(ERROR,
 				(errcode(ERRCODE_NUMERIC_VALUE_OUT_OF_RANGE),
 				 errmsg("fixeddecimal out of range")));
 #else
+
+    /*
+	 * Overflow check. If the subtractor cannot be fit into 
+     * 64 bit, then the result will definitely overflow
+	 */
+    if (subtractor != (int64) subtractor)
```
tanscorpio7: Can we use overflow checks similar to PG ? For example
```
if (unlikely(arg1 < SHRT_MIN) || unlikely(arg1 > SHRT_MAX))
```

contrib/babelfishpg_common/sql/bit.sql
ayushdsh: 1. Yes. Its just like or-ing the two results. 
2. The reason for having only +- operators is to handle overflows, which cannot occur for mul and div. bit is lower precedence than money smallmoney .. thats why the result cam overflow.

contrib/babelfishpg_common/sql/bit.sql
ayushdsh: The main reason for introducing +- is to handle overflows. 
You're right. I'll rewrite this is C instead!

tanscorpio7 (changes requested):

contrib/babelfishpg_common/sql/money/fixeddecimal--1.1.0_base_parallel.sql
ayushdsh: mod returns the correct results. I've added tests for that as well, and the result is correct. That's why we don't have to handle these

tanscorpio7: Oh, this is mod operator so no overflow is possible.

contrib/babelfishpg_money/fixeddecimal.c
tanscorpio7: This might be redundant comment if we change the overflow check condition to this https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/3797#discussion_r2145136369

contrib/babelfishpg_money/fixeddecimal.c
tanscorpio7: Seems like in this case `Int64GetDatum(int64 X)` ends up casting it to int64.

ayushdsh: Yes .. exactly. I'd be very surprised if the pg return function does not actually do what it says.
At the end, it will be casted to 64bits in any case.

contrib/babelfishpg_common/sql/bit.sql
ayushdsh: I've written most of them in C. The ones that are pretty obvious like mul, I've done it in SQL to avoid the overhead of func parsing and selection.
I've also added other arithmetic on bit -
1. bit, bit
2. bit, smallmoney
3. and bit, float

other operands are not required, because they result in the correct result + rettype.

contrib/babelfishpg_money/fixeddecimal.c
ayushdsh: The `unlikely()` macro only suggests the compiler that a branch might be "unlikely" (so the compiler can optimize on the branch prediction). I think we would still need this handling. 
And you are right; we should add explicit cast to `FIXEDDECIMAL_MULTIPLIER` as well.

test/JDBC/expected/Test-spatial-functions-2-vu-verify.out
```diff
@@ -375,9 +375,9 @@ GROUP BY ROUND(PointColumn.STIntersects(@referencePoint) / @intersection_points,
 ORDER BY Intersectinggroup;
 go
 ~~START~~
-numeric#!#int
+bigint#!#int
 <NULL>#!#1
-0E-8#!#2
+0#!#2
```
ayushdsh: The query written here ideally is not valid in T-SQL. Below is the result - 
```
SELECT ROUND(PointColumn.STIntersects(@referencePoint) / @intersection_points, 0) * @intersection_points AS Intersectinggroup,
COUNT(*) AS PointCount

Msg 8117, Level 16, State 1, Line 6
Operand data type bit is invalid for divide operator.
```

This is because bit division is not supported in T-SQL. But for sanity, we are currently supporting it. Here the change comes because of the correct implementation of the `bit/bit` operator. 
And hence the rettype leads to bigint because of the `ROUND` function on a `bit`

test/JDBC/expected/Test-spatial-functions-2-vu-verify.out
```diff
@@ -869,9 +869,9 @@ bit
 ~~END~~
 
 ~~START~~
-numeric#!#int
+bigint#!#int
 <NULL>#!#1
-0E-8#!#2
+0#!#2
```
ayushdsh: The query written here ideally is not valid in T-SQL. Below is the result - 
```
SELECT ROUND(PointColumn.STIntersects(@referencePoint) / @intersection_points, 0) * @intersection_points AS Intersectinggroup,
COUNT(*) AS PointCount

Msg 8117, Level 16, State 1, Line 6
Operand data type bit is invalid for divide operator.
```

This is because bit division is not supported in T-SQL. But for sanity, we are currently supporting it. Here the change comes because of the correct implementation of the `bit/bit` operator. 
And hence the rettype leads to bigint because of the `ROUND` function on a `bit`

test/JDBC/expected/Test-spatial-functions-vu-verify.out
```diff
@@ -402,9 +402,9 @@ GROUP BY ROUND(PointColumn.STEquals(@referencePoint) / @equal, 0) * @equal
 ORDER BY Equalitygroup;
 GO
 ~~START~~
-numeric#!#int
+bigint#!#int
 <NULL>#!#1
-0E-8#!#2
+0#!#2
```
ayushdsh: The query written here ideally is not valid in T-SQL. Below is the result - 
```
SELECT ROUND(PointColumn.STIntersects(@referencePoint) / @intersection_points, 0) * @intersection_points AS Intersectinggroup,
COUNT(*) AS PointCount

Msg 8117, Level 16, State 1, Line 6
Operand data type bit is invalid for divide operator.
```

This is because bit division is not supported in T-SQL. But for sanity, we are currently supporting it. Here the change comes because of the correct implementation of the `bit/bit` operator. 
And hence the rettype leads to bigint because of the `ROUND` function on a `bit`

test/JDBC/expected/Test-spatial-functions-vu-verify.out
```diff
@@ -808,9 +808,9 @@ bit
 ~~END~~
 
 ~~START~~
-numeric#!#int
+bigint#!#int
 <NULL>#!#1
-0E-8#!#2
+0#!#2
```
ayushdsh: The query written here ideally is not valid in T-SQL. Below is the result - 
```
SELECT ROUND(PointColumn.STIntersects(@referencePoint) / @intersection_points, 0) * @intersection_points AS Intersectinggroup,
COUNT(*) AS PointCount

Msg 8117, Level 16, State 1, Line 6
Operand data type bit is invalid for divide operator.
```

This is because bit division is not supported in T-SQL. But for sanity, we are currently supporting it. Here the change comes because of the correct implementation of the `bit/bit` operator. 
And hence the rettype leads to bigint because of the `ROUND` function on a `bit`

test/JDBC/expected/babel_varbinary-vu-verify.out
```diff
@@ -917,8 +917,8 @@ ORDER BY
     float_binary_add;
 GO
 ~~START~~
-float#!#float#!#float#!#float#!#float#!#float#!#float#!#datetime#!#float#!#varchar#!#datetime#!#bigint
-201.0#!#201.0#!#1100.5#!#200.5#!#110.5#!#101.5#!#201.0#!#1900-04-11 12:00:00.0#!#201.0#!#100.5A1A2A3A4-B1B2-C1C2-D1D2-E1E2E3E4E5E6#!#<NULL>#!#100
+float#!#float#!#float#!#float#!#float#!#float#!#float#!#float#!#float#!#varchar#!#datetime#!#bigint
```
ayushdsh, line 920: Here the change comes because now we have correctly implemented the `bit + float` operator, previously which resulted into datetime, now results into float which follows the T-SQL behaviour.

test/JDBC/expected/babel_varbinary-vu-verify.out
```diff
@@ -952,8 +952,8 @@ ORDER BY
     real_binary_add;
 GO
 ~~START~~
-real#!#float#!#real#!#real#!#real#!#real#!#real#!#datetime#!#real#!#varchar#!#datetime#!#bigint
-201.0#!#201.0#!#1100.5#!#200.5#!#110.5#!#101.5#!#201.0#!#1900-04-11 12:00:00.0#!#201.0#!#100.5A1A2A3A4-B1B2-C1C2-D1D2-E1E2E3E4E5E6#!#<NULL>#!#100
+real#!#float#!#real#!#real#!#real#!#real#!#real#!#float#!#real#!#varchar#!#datetime#!#bigint
```
ayushdsh, line 955: Here the change comes because now we have correctly implemented the `bit + float` operator, previously which resulted into datetime, now results into float which follows the T-SQL behaviour.

test/JDBC/expected/babel_varbinary-vu-verify.out
```diff
@@ -1162,8 +1162,8 @@ ORDER BY
     bit_binary_add;
 GO
 ~~START~~
-numeric#!#datetime#!#datetime#!#bigint#!#int#!#smallint#!#smallint#!#money#!#datetime#!#varchar#!#datetime#!#bigint
-100.50#!#1900-04-11 12:00:00.0#!#1900-04-11 12:00:00.0#!#1000#!#100#!#10#!#1#!#100.5000#!#1900-01-01 00:00:00.0#!#0A1A2A3A4-B1B2-C1C2-D1D2-E1E2E3E4E5E6#!#<NULL>#!#0
+numeric#!#float#!#float#!#bigint#!#int#!#smallint#!#smallint#!#money#!#bit#!#varchar#!#datetime#!#bigint
```
ayushdsh: Here the change comes because now we have correctly implemented the `bit + float`, `bit + real` and `bit + bit` operator, previously which resulted into datetime, now results into float which follows the T-SQL behaviour.

contrib/babelfishpg_money/fixeddecimal.c
ayushdsh: The `unlikely()` macro is used for suggesting the compiler that this code path might be taken rarely. We can definitely add it for sanity.

The overall condition would look like -

`if (unlikely(subtractor != (int64) subtractor))`

contrib/babelfishpg_money/smallmoney.c
ayushdsh: I'm not sure how do we determine whether both money and smallmoney are 8 bytes long? (any suggestions).

But I think for sanity, we should return 64 bit result.

rohit01010:
```
babelfish_db=> select oid, typname, typnamespace, typlen from pg_type where typname like 'smallmoney' OR typname like 'money';
  oid  |  typname   | typnamespace | typlen 
-------+------------+--------------+--------
   790 | money      |           11 |      8
 20533 | money      |        20518 |      8
 20610 | smallmoney |        20518 |      8
(3 rows)
```

SMALLMONEY and MONEY is a domain created on FIXEDDECIMAL with additional check constraints, and since FIXEDDECIMAL is of 8 bytes, sys.SMALLMONEY implicitly becomes 8 bytes. This can be confirmed from above psql query.

rohit01010: Also make sure whether we need similar thing for getting the arg as well or not, i.e. do we need to use PG_GETARG_INT64 instead of  PG_GETARG_INT32 ?

ayushdsh: I agree that in babelfish it will be 8B .. but is it the same with tsql. 
It's probably not a big deal to have it 8B even if tsql is 4B

ayushdsh: > Also make sure whether we need similar thing for getting the arg as well or not, i.e. do we need to use PG_GETARG_INT64 instead of  PG_GETARG_INT32 ?

I don't think it will be a problem to keep it 32 bits .. because the domain constraints ensure that we will not overflow Int32_Max

test/python/expected/upgrade_validation/expected_dependency.out
```diff
@@ -573,6 +571,7 @@ Function sys.pltsql_call_handler()
 Function sys.pltsql_inline_handler(internal)
 Function sys.pltsql_validator(oid)
 Function sys.point(sys.geometry)
+Function sys.power(sys.fixeddecimal,numeric)
```
rohit01010: Lets add some dependent tests for this function as well

contrib/babelfishpg_money/fixeddecimal.c
tanscorpio7: Actually I meant taking overflow checks from postgres code since those are tried and tested.
For example multiplication can just be
```
	int64		arg1 = PG_GETARG_INT64(0);
	int64		arg2 = PG_GETARG_INT64(1);
	int64		result;

	if (unlikely(pg_mul_s64_overflow(arg1, arg2, &result)))
		ereport(ERROR,
				(errcode(ERRCODE_NUMERIC_VALUE_OUT_OF_RANGE),
				 errmsg("bigint out of range")));
        if (result < min_small_money && result > max small money)
                error....

	PG_RETURN_INT64(result);
```

ayushdsh: Ah I see. Thanks for the explanation. That makes sense! Thanks!

contrib/babelfishpg_money/smallmoney.c
ayushdsh: As discussed offline, we will be accepting and returning 64bit type.

test/JDBC/expected/TestExactNumeric.out
```diff
@@ -486,27 +486,27 @@ FROM exactnumeric_table;
 GO
 ~~START~~
 int#!#int#!#int#!#bigint
```
ayushdsh, line 520: The reason for this change is because we've rectified the SIGN function.
I've added a diff file which shows the side-by-side diff. You can see that now there is no diff in output of tsql vs. bbf
[diff.txt](https://github.com/user-attachments/files/20760574/diff.txt)

test/JDBC/expected/numeric_money-vu-verify.out
```diff
@@ -1166,8 +1166,9 @@ SELECT CAST(9223372036854775807 AS BIGINT) * CAST(0.0001 AS MONEY)
 GO
 ~~START~~
 money
-922337203685477.5807
-~~END~~
+~~ERROR (Code: 33557097)~~
+
+~~ERROR (Message: fixeddecimal out of range)~~
```
ayushdsh: This is the correct behaviour in T-SQL. The query
`SELECT CAST(9223372036854775807 AS BIGINT) * CAST(0.0001 AS MONEY)` does overflow in T-SQL

contrib/babelfishpg_money/smallmoney.c
```diff
@@ -0,0 +1,1140 @@
...
+     * returns true
+     */
+    if (pg_add_s32_overflow(arg1, arg2, &result))
+        ereport(ERROR,
+                (errcode(ERRCODE_NUMERIC_VALUE_OUT_OF_RANGE),
+                 errmsg("smallmoney out of range")));
+        
+#endif							
+    PG_RETURN_INT64(result);
+}
```
tanscorpio7: `pg_add_s32_overflow` already checks for `HAVE__BUILTIN_OP_OVERFLOW`. I think we can remove the #ifdef from here and just call `pg_add_s32_overflow`

contrib/babelfishpg_money/smallmoney.c
```diff
@@ -0,0 +1,1140 @@
...
+Datum
+smallmoneyint8pl(PG_FUNCTION_ARGS)
+{
+    int64		arg1 = PG_GETARG_INT64(0);
+    int64       arg2 = PG_GETARG_INT64(1);
+    int64		adder;
+    int128		result;
+
+#ifdef HAVE_BUILTIN_OVERFLOW /* HAVE_BUILTIN_OVERFLOW */
+    if (__builtin_add_overflow(arg1, adder, &result))
```
tanscorpio7: `adder` is. not yet initialized. Also the else block seems to have a different logic then if. We should try avoid having `ifdef / else` in our operator functions and directly call pg wrapper functions instead.

tanscorpio7 (changes requested):

contrib/babelfishpg_common/src/bit.c
```diff
@@ -568,3 +589,267 @@ varchar2bit(PG_FUNCTION_ARGS)
...
+                 errmsg("smallmoney out of range")));
+#else
+    result = (int64) subtractor - arg2;
+
+    /*
+     * Overflow check. If the result cannot be converted back
+     * to a 32 bit result, then we can say that there is a 
+     * overflow
+     */
+    if (result != (int32) result)
```
tanscorpio7: I am not sure how compiler equates to different types. Could we instead do it like 
```
if (unlikely(arg1 < SMALLMONEY_MIN) || unlikely(arg1 > SMALLMONEY_MAX))
```

tanscorpio7 (changes requested):

contrib/babelfishpg_common/sql/bit.sql
tanscorpio7: > I've done it in SQL to avoid the overhead of func parsing and selection.

Actually SQL body functions have overhead of parsing and analyzing.

contrib/babelfishpg_common/src/bit.c
ayushdsh: Agreed. Good point. Its definitely much safer this way.

test/JDBC/expected/TestStrings.out
```diff
@@ -4261,15 +4261,9 @@ GO
...
-varchar
-Prefix_1
-~~END~~
+~~ERROR (Code: 33557097)~~
 
-~~START~~
-nvarchar
-Prefix_1
-~~END~~
+~~ERROR (Message: invalid input syntax for type bit: "Prefix_")~~
```
ayushdsh: nvarchar and bit are incompatible operators

contrib/babelfishpg_money/smallmoney.c
```diff
@@ -0,0 +1,974 @@
...
+	 * Remember that arg2 is the number multiplied by FIXEDDECIMAL_MULTIPLIER,
+	 * we must divide the result by this to get the correct result.
+	 */
+	result = (int64) arg1 * arg2 / FIXEDDECIMAL_MULTIPLIER;
+
+	/* Round off the result to FIXEDDECIMAL_SCALE.
+	 * abs() in order to deal with -ve result as well 
+	 * if the result is negative we subtract 1, else add 1
+	 */
+	if (abs(((int64) arg1 * arg2) % FIXEDDECIMAL_MULTIPLIER) >= FIXEDDECIMAL_ROUNDUP)
```
tanscorpio7: Should we use this property so that we need not worry about overflow `(a×b)modc=(amodc×bmodc)modc`

contrib/babelfishpg_money/smallmoney.c
```diff
@@ -0,0 +1,974 @@
...
+	int64		arg1 = PG_GETARG_INT64(0);
+	int64		arg2 = PG_GETARG_INT64(1);
+	int64		result;
+
+	/*
+	 * We need to promote this to 64bit as we may overflow int32 here.
+	 * Remember that arg2 is the number multiplied by FIXEDDECIMAL_MULTIPLIER,
+	 * we must divide the result by this to get the correct result.
+	 */
+	result = (int64) arg1 * arg2 / FIXEDDECIMAL_MULTIPLIER;
```
tanscorpio7: Please wrap the mathematical operation with brackets to make it explicit what happens first.

contrib/babelfishpg_money/smallmoney.c
```diff
@@ -0,0 +1,974 @@
...
+	if (pg_mul_s64_overflow(arg2, (int64) FIXEDDECIMAL_MULTIPLIER, &adder)) 
+	{
+		ereport(ERROR,
+				(errcode(ERRCODE_NUMERIC_VALUE_OUT_OF_RANGE),
+				 errmsg("smallmoney out of range")));
+		/* ensure compiler realizes we mustn't reach the division (gcc bug) */
+		PG_RETURN_NULL();
+	}
+
+	result = arg1 + adder;
```
tanscorpio7, line 201: pg_add function ? Similarly in the below functions as well.

contrib/babelfishpg_money/smallmoney.c
```diff
@@ -0,0 +1,974 @@
...
+{
+	int64		arg1 = PG_GETARG_INT64(0);
+	int64		arg2 = PG_GETARG_INT64(1);
+	int32		result;
+
+	/*
+	 * Overflow check. If the result of addition
+	 * does not fit in 32 bit, then pg_mul_s64_overflow
+	 * returns true
+	 */
```
tanscorpio7: comment mentions `pg_mul_s64_overflow`

contrib/babelfishpg_money/smallmoney.c
```diff
@@ -0,0 +1,974 @@
...
+	if (pg_mul_s32_overflow(arg2, (int32) FIXEDDECIMAL_MULTIPLIER, &adder)) 
+	{
+		ereport(ERROR,
+				(errcode(ERRCODE_NUMERIC_VALUE_OUT_OF_RANGE),
+				 errmsg("smallmoney out of range")));
+		/* ensure compiler realizes we mustn't reach the division (gcc bug) */
+		PG_RETURN_NULL();
+	}
+
+	result = arg1 + adder;
```
tanscorpio7: pg_add ?

contrib/babelfishpg_tsql/sql/sys_functions.sql
```diff
@@ -2297,6 +2297,9 @@ AS 'babelfishpg_tsql', 'int_floor' LANGUAGE C IMMUTABLE STRICT PARALLEL SAFE;
 CREATE OR REPLACE FUNCTION sys.floor(tinyint) RETURNS TINYINT
 AS 'babelfishpg_tsql', 'int_floor' LANGUAGE C IMMUTABLE STRICT PARALLEL SAFE;
 
+CREATE OR REPLACE FUNCTION sys.floor(sys.fixeddecimal) RETURNS sys.MONEY
+AS 'babelfishpg_money', 'fixeddecimal_floor' LANGUAGE C IMMUTABLE STRICT PARALLEL SAFE;
```
tanscorpio7, line 2319: For Smallmoney input what is the output result data type ?

tanscorpio7 (changes requested):

contrib/babelfishpg_tsql/sql/sys_functions.sql
ayushdsh: It's still money.
Ref: https://github.com/MicrosoftDocs/sql-docs/blob/live/docs/t-sql/functions/floor-transact-sql.md

contrib/babelfishpg_money/smallmoney.c
ayushdsh: Yes. That's a good call. We should!

contrib/babelfishpg_money/smallmoney.c
ayushdsh: The reason for not using it is because we would anyways have to do the overflow check like below. Here we'll be trying to use pg_add_s64, but will require an explicit check `result > INT32_MAX || result < INT32_MIN` anyways.

contrib/babelfishpg_money/smallmoney.c
ayushdsh: Agreed. Here we should've used pg_add_s32

contrib/babelfishpg_common/src/bit.c
```diff
@@ -62,6 +65,32 @@ PG_FUNCTION_INFO_V1(bitint4le);
...
+PG_FUNCTION_INFO_V1(bitsmallmoneypl);
+PG_FUNCTION_INFO_V1(bitsmallmoneymi);
+PG_FUNCTION_INFO_V1(bitsmallmoneymul);
+PG_FUNCTION_INFO_V1(bitsmallmoneydiv);
+
+/* Arithmetic operations bit */
+PG_FUNCTION_INFO_V1(bitpl);
+PG_FUNCTION_INFO_V1(bitmi);
+PG_FUNCTION_INFO_V1(bitmul);
+PG_FUNCTION_INFO_V1(bitdiv);
```
rohit01010: if we are not using these functions anymore we should remove it

test/JDBC/expected/Test-spatial-functions-2-vu-verify.out
```diff
@@ -367,17 +367,17 @@ bit
...
 FROM  TestGeospatialMethods_YourTableTemp
 GROUP BY ROUND(PointColumn.STIntersects(@referencePoint) / @intersection_points, 0) * @intersection_points
 ORDER BY Intersectinggroup;
 go
 ~~START~~
-numeric#!#int
+int#!#int
 <NULL>#!#1
-0E-8#!#2
+0#!#2
```
ayushdsh, line 380: This is because bit div bit is not a valid operation and results in the following error. Here i am trying to preserve what the test does and hence changed the type from BIT to INT
> 1> select cast(1 as bit) / cast(1 as bit);
2> go
Msg 8117, Level 16, State 1, Server EC2AMAZ-CLDACUV, Line 1
Operand data type bit is invalid for divide operator.

rohit01010 (approved):

contrib/babelfishpg_common/sql/bit.sql
```diff
@@ -504,3 +504,183 @@ WITH FUNCTION sys.varchar2bit(sys.VARCHAR) AS IMPLICIT;
...
+    LEFTARG    = sys.BIT,
+    RIGHTARG   = float8,
+    PROCEDURE  = sys.bitfloatmul
+);
+
+CREATE OPERATOR sys./ (
+    LEFTARG    = sys.BIT,
+    RIGHTARG   = float8,
+    PROCEDURE  = sys.bitfloatdiv
+);
```
tanscorpio7, line 694: Let us add a new line wherever needed.

contrib/babelfishpg_common/sql/upgrades/babelfish_common_helper--5.2.0--5.3.0.sql
```diff
@@ -133,6 +133,456 @@ BEGIN
...
+
+
+CREATE OR REPLACE FUNCTION sys.bitsmallmoneypl(sys.BIT, sys.SMALLMONEY)
+RETURNS sys.SMALLMONEY
+AS 'babelfishpg_common', 'bitsmallmoneypl'
+LANGUAGE C IMMUTABLE STRICT PARALLEL SAFE;
+
+DO $$
+BEGIN
+IF NOT EXISTS(SELECT 1 FROM pg_catalog.pg_operator WHERE oprleft = 'sys.bit'::pg_catalog.regtype and oprright = 'sys.smallmoney'::pg_catalog.regtype and oprnamespace = 'sys'::regnamespace and oprname = '+' and oprresult != 0) THEN
```
tanscorpio7: What does `oprresult != 0` do ?

contrib/babelfishpg_money/fixeddecimal.c
```diff
@@ -2070,6 +2094,19 @@ int8fixeddecimalmul(PG_FUNCTION_ARGS)
 #else
 	result = (int128) arg1 * arg2;
 
+	/*
+	 multiplication with values > FIXEDDECIMAL_MAX or < FIXEDDECIMAL_MIN lead to overflow
+	 in T-SQL
+	*/
+	if (arg1 > FIXEDDECIMAL_MAX || arg1 < FIXEDDECIMAL_MIN)
```
tanscorpio7: Can we replace the entire cast function with pg_mul_64 ?

contrib/babelfishpg_money/smallmoney.c
```diff
@@ -0,0 +1,974 @@
...
+ *	Arithmetic operators on smallmoney.
+ *---------------------------------------------------------*/
+
+/*
+	Even though the range of smallmoney is within INT32_MIN and INT32_MAX,
+	we have created it as a domain on top of FIXEDDECIMAL which is represented 
+	as a 64 bit INT. Hence, we will also use 64 bit to smallmoney manipulation.
+    But we can be guaranteed that the smallmoney arg will always fit in 32bits
+    In other words, INT32_MIN <= arg <= INT32_MAX
+*/
```
tanscorpio7: indent

contrib/babelfishpg_money/smallmoney.c
```diff
@@ -0,0 +1,974 @@
...
+	 * INT64_MIN / -1 is problematic, since the result can't be represented on
+	 * a two's-complement machine.  Some machines produce INT64_MIN, some
+	 * produce zero, some throw an exception.  We can dodge the problem by
+	 * recognizing that division by -1 is the same as negation.
+	 */
+	if (arg2 == -1)
+	{
+		result = -arg1;
+		/* overflow check (needed for INT64_MIN) */
+		if (arg1 != 0 && result != (int32) result)
```
tanscorpio7: Can just check for `if (unlikely(arg1 == PG_INT64_MIN))`

contrib/babelfishpg_money/smallmoney.c
```diff
@@ -0,0 +1,974 @@
...
+	*/
+	if (arg2 > SMALLMONEY_MAX || arg2 < SMALLMONEY_MIN)
+	{
+		ereport(ERROR,
+				(errcode(ERRCODE_NUMERIC_VALUE_OUT_OF_RANGE),
+				 errmsg("smallmoney out of range")));
+		/* ensure compiler realizes we mustn't reach the division (gcc bug) */
+		PG_RETURN_NULL();
+	}
+
```
tanscorpio7: Redundant ?

contrib/babelfishpg_money/smallmoney.c
```diff
@@ -0,0 +1,974 @@
...
+	*/
+	if (arg1 > SMALLMONEY_MAX || arg1 < SMALLMONEY_MIN)
+	{
+		ereport(ERROR,
+				(errcode(ERRCODE_NUMERIC_VALUE_OUT_OF_RANGE),
+				 errmsg("smallmoney out of range")));
+		/* ensure compiler realizes we mustn't reach the division (gcc bug) */
+		PG_RETURN_NULL();
+	}
+
```
tanscorpio7: Probably can remove this.

contrib/babelfishpg_money/fixeddecimal.c
```diff
@@ -2099,6 +2136,19 @@ int8fixeddecimaldiv(PG_FUNCTION_ARGS)
...
+	{
+		ereport(ERROR, 
+				(errcode(ERRCODE_NUMERIC_VALUE_OUT_OF_RANGE),
+				 errmsg("fixeddecimal out of range")));
+		/* ensure compiler realizes we mustn't reach the division (gcc bug) */
+		PG_RETURN_NULL();
+	}
+
```
tanscorpio7: What is SQL Server return type for this operator ?

tanscorpio7 (changes requested):

contrib/babelfishpg_money/fixeddecimal.c
ayushdsh: FIXEDDECIMAL_MAX != INT64_MAX and same for MIN

contrib/babelfishpg_money/smallmoney.c
ayushdsh: SMALLMONEY_MAX is not equal INT32_MAX
INT32_MAX = 2147483647
SMALLMONEY_MAX = (INT32_MAX/FIXEDDECIMAL_MULTIPLIER) = 214748

contrib/babelfishpg_money/fixeddecimal.c
ayushdsh: Both T-SQL and BBF with these changes return money.
T-SQL
```
1> select cast(1000 as bigint) / cast(143.4243 as money);
2> go
 
-
6.9723
```

BBF
```
1> select cast(1000 as bigint) / cast(143.4243 as money);
2> go
                     
---------------------
               6.9723
```

It's just that we do the calculations in float, and the sender function takes care of converting it to money (simply by truncating decimal places)

contrib/babelfishpg_common/sql/upgrades/babelfish_common_helper--5.2.0--5.3.0.sql
ayushdsh: `oprresult` stores the rettype oid. It just verifies that the operator in pg_operator is a valid one.

contrib/babelfishpg_money/fixeddecimal.c
tanscorpio7: But the return type of this operator looks to be float8 in babelfish

ayushdsh: This function is used as a wrapper by `int8fixeddecimaldiv_money`

contrib/babelfishpg_money/fixeddecimal.c
```diff
@@ -1741,10 +1689,18 @@ fixeddecimalmul(PG_FUNCTION_ARGS)
...
 	 */
-	result = (int128) arg1 * arg2 / FIXEDDECIMAL_MULTIPLIER;
-	/* Round off the result to FIXEDDECIMAL_SCALE. */
-	if ((((int128) arg1 * arg2 % FIXEDDECIMAL_MULTIPLIER)) >= FIXEDDECIMAL_ROUNDUP)
-	result++;
+	result = ((int128) arg1 * arg2) / FIXEDDECIMAL_MULTIPLIER;
+	/* Round off the result to FIXEDDECIMAL_SCALE.
+	 * abs() in order to deal with -ve result as well 
+	 * if the result is negative we subtract 1, else add 1
+	 */
```
tanscorpio7: comment formatting

contrib/babelfishpg_common/sql/money/fixeddecimal--1.1.0_base_parallel.sql
```diff
@@ -1839,32 +1839,30 @@ CREATE OPERATOR sys./ (
...
 RETURNS sys.SMALLMONEY
-AS 'babelfishpg_money', 'int8fixeddecimalmul'
+AS 'babelfishpg_money', 'int8smallmoneymul'
 LANGUAGE C IMMUTABLE STRICT PARALLEL SAFE;
 
 CREATE FUNCTION sys.int8fixeddecimaldiv(INT8, sys.SMALLMONEY)
 RETURNS DOUBLE PRECISION
-AS 'babelfishpg_money', 'int8fixeddecimaldiv'
+AS 'babelfishpg_money', 'int8smallmoneydiv'
 LANGUAGE C IMMUTABLE STRICT PARALLEL SAFE;
```
tanscorpio7: `int8fixeddecimaldiv` still expects FLOAT8 return type from C function. Let us put the old C function logic into another function and make `sys.int8fixeddecimaldiv` point to it.

tanscorpio7 (approved):

contrib/babelfishpg_common/sql/money/fixeddecimal--1.1.0_base_parallel.sql
ayushdsh: missed this ... a very good catch!

tanscorpio7 (approved):

jsudrik (approved):
Approved.

PR #3805, opened 2025-06-05
Optimize the antlr g4 definition to get better performance

forestkeeper:
Rewrite the antlr rule for column_name, skip the nested ()? structure, rewrite it into a clean antlr rule options, it'll save 30 ~ 40% of the time in some extreme cases.

Meanwhile, such rewrite will ignore those column text like '..a.b', 'a..b' or '.a..b' previously regard as valid by antlr parser, after this change, those column text will be invalid from antlr parser, it'll change the error message for such column texts. But this commit will not change the behavior for such column text, because previously, it was regard as invalid by gram.y parser.

Task: BABEL-5548

### Check List

### DIFF
--- a/contrib/babelfishpg_tsql/antlr/TSqlParser.g4
+++ b/contrib/babelfishpg_tsql/antlr/TSqlParser.g4
@@ -5266,26 +5266,15 @@ collation
     ;
 
 full_column_name
-// standard
-    : column_name=id                                     // column
-    | column=id DOT geospatial_col          // column.geospatial
-    | tablename=id DOT column_name=id                      // table.column_name
-    | table=id DOT column=id DOT geospatial_col          // table.column.geospatial
+    : column_name=id                                    // column
+    | column=id DOT geospatial_col                      // column.geospatial
+    | tablename=id DOT column_name=id                   // table.column_name
+    | table=id DOT column=id DOT geospatial_col         // table.column.geospatial
     | schema=id DOT tablename=id DOT column_name=id                      // schema.tablename.column_name
     | schema=id DOT table=id DOT column=id DOT geospatial_col   // schema.table.column.geospatial
     | server=id DOT schema=id DOT tablename=id DOT column_name=id               // server.schema.tablename.column_name
-// geospatial_col
-// standard partial with missing IDs
-//    | DOT column_name=id
-//    | DOT DOT column_name=id
-//    | DOT DOT DOT column_name=id
-    | DOT tablename=id DOT column_name=id
-//    | DOT DOT tablename=id DOT column_name=id
-//    | DOT schema=id DOT tablename=id DOT column_name=id
-// geospatial_col partial with missing IDs
-//    | DOT table=id DOT column=id DOT geospatial_col                      // .t.c.g
-//    | DOT DOT column=id DOT geospatial_col                               // ..c.g
-//    | DOT column=id DOT geospatial_col                               // .c.g
+    | DOT tablename=id DOT column_name=id                                       // .tablename.column_name
+    | DOT DOT tablename=id DOT column_name=id                                   // ..tablename.column_name
     ;
 
 column_name_list_with_order
--- a/contrib/babelfishpg_tsql/antlr/TSqlParser.g4
+++ b/contrib/babelfishpg_tsql/antlr/TSqlParser.g4
@@ -5274,8 +5274,6 @@ full_column_name
     | schema=id DOT table=id DOT column=id DOT geospatial_col   // schema.table.column.geospatial
     | server=id DOT schema=id DOT tablename=id DOT column_name=id               // server.schema.tablename.column_name
     | DOT tablename=id DOT column_name=id                                       // .tablename.column_name
-    | DOT table=id DOT column=id DOT geospatial_col                             // .tablename.column_name.geospatial
-    | DOT DOT column=id DOT geospatial_col                                      // ..column.geospatial
     | DOT DOT tablename=id DOT column_name=id                                   // ..tablename.column_name
     ;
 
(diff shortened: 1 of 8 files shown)


contrib/babelfishpg_tsql/antlr/TSqlParser.g4
```diff
@@ -5264,10 +5264,19 @@ external_name
...
+    : column_name=id                                    // column
+    | column=id DOT geospatial_col                      // column.geospatial
+    | tablename=id DOT column_name=id                   // table.column_name
+    | table=id DOT column=id DOT geospatial_col         // table.column.geospatial
+    | schema=id DOT tablename=id DOT column_name=id                      // schema.tablename.column_name
+    | schema=id DOT table=id DOT column=id DOT geospatial_col   // schema.table.column.geospatial
+    | server=id DOT schema=id DOT tablename=id DOT column_name=id               // server.schema.tablename.column_name
+    | DOT tablename=id DOT column_name=id                                       // .tablename.column_name
+    | DOT table=id DOT column=id DOT geospatial_col                             // .tablename.column_name.geospatial
+    | DOT DOT column=id DOT geospatial_col                                      // ..column.geospatial
```
Anikait143: I think we could also skip these 2 rules

contrib/babelfishpg_tsql/antlr/TSqlParser.g4
```diff
@@ -5264,10 +5264,19 @@ external_name
...
+    | DOT DOT tablename=id DOT column_name=id                                   // ..tablename.column_name
```
Anikait143, line 5277: We missed to add these rules here:
1. `DOT schema=id DOT tablename=id DOT column_name=id`
2. `server=id DOT DOT tablename=id DOT column_name=id`

Although seems like they are invalidated by bison but TSQL supports these syntaxes

test/JDBC/expected/BABEL-2724.out
```diff
@@ -36,23 +36,23 @@ select .a from t2724;
 go
 ~~ERROR (Code: 33557097)~~
 
-~~ERROR (Message: syntax error at or near ".")~~
+~~ERROR (Message: syntax error near 'from' at line 2 and character position 10)~~
```
Anikait143, line 39: Can we somehow have a more informative error message here? The last one seems to describe the issue more efficiently.

test/JDBC/input/BABEL-5548.sql
```diff
@@ -0,0 +1,45 @@
...
+GO
+
+select .GEOSPATIALPOINTEMPTYdt.geom.STAsText() from GEOSPATIALPOINTEMPTYdt
+GO
+
+select ..geom.STAsText() from GEOSPATIALPOINTEMPTYdt
+GO
+
+select dbo.GEOSPATIALPOINTEMPTYdt.geom.STAsText() from GEOSPATIALPOINTEMPTYdt
+GO
```
Anikait143, line 39: These cases with STAsText() would not match the `full_column_name` rule. Instead let's use a function from `geospatial_col` list such as `STX`.

test/JDBC/expected/sys-suser_sname-vu-verify.out
```diff
@@ -18,7 +18,7 @@ SELECT SYS.SUSER_SNAME(0x0P)
 GO
 ~~ERROR (Code: 33557097)~~
 
-~~ERROR (Message: syntax error near '0x0' at line 1 and character position 23)~~
+~~ERROR (Message: syntax error near '.' at line 1 and character position 10)~~
```
Anikait143, line 21: I am not able to understand why this got affected. Is this because of some internal rewriting?

contrib/babelfishpg_tsql/antlr/TSqlParser.g4
forestkeeper: Those two rules actually works, I have test cases for those two.

contrib/babelfishpg_tsql/antlr/TSqlParser.g4
forestkeeper: I have test cases to prove those 2 rules will error out by bison parser, so we don't waste our CPU on this .

The server=id DOT DOT tablename=id DOT column_name=id rule will waste too much CPU, and not worth to add it here.

test/JDBC/expected/BABEL-2724.out
forestkeeper: We can't, if we wanted to make bison parser to parse this, we'll waste around 20% parse time in there.

test/JDBC/input/BABEL-5548.sql
forestkeeper: ack

contrib/babelfishpg_tsql/antlr/TSqlParser.g4
Anikait143: Ack, can we have a Jira created for this that we don't support these two rules while TSQL does?

forestkeeper: ..table.column -> SQL Server supported, but we don't, looks like we should create a JIRA but low priority.
.dbo..column -> SQL Server doesn't support it as well.

test/JDBC/expected/sys-suser_sname-vu-verify.out
forestkeeper: Yes, it got affected because we changed a bit about the full_column_rule, the result is the same, AKA : error out, but the message will be changed.

Anikait143 (approved):

lejaokri (approved):
just merging this one as per discussion with Zhibai

PR #3810, opened 2025-06-06
charindex support for varbinary and enhancement

herambh-shah:
### CHARINDEX Function Specification -

_**Syntax**:_
CHARINDEX(expressionToFind, expressionToSearch, start_position)

_**expressionToFind**:_
-Must be either VARCHAR or VARBINARY type
-No other data types are directly accepted

_**expressionToSearch:**_
-Can accept any data type
-Must be implicitly convertible to match the type of expressionToFind

_**start_position**_ (optional):  Starting position for the search, If omitted, defaults to 1

Through this implementation, we ensure CHARINDEX supports: 
-All valid combinations of binary and character data types
-Proper type conversion between expressionToFind and expressionToSearch

### Issue resolved -
BABEL - 3659 & BABEL - 5912

### Key Changes -
"This Pull Request includes 3 main enhancements to the CHARINDEX function:

- [ ] Added Support for Data Types - VAR[BINARY] 
- [ ] Performance Improvements -
      1. Reimplemented system function from SQL to C for text argument of charindex
      2. Improved execution efficiency by moving from SQL-based to C-based implementation
- [ ] Support Mixed type arguments also. For eg - charindex( varbinary , int )

### Implementation details  -
1.  New System Functions to support VAR[BINARY] args -

- [ ] charindex(varbinary , anyelement )  - calls **charindex_helper**( varbinary , anyelement :: varbinary ) 
- [ ] charinex( binary , anyelement ) -  calls **charindex_helper**( varbinary , anyelement ::binary )
- [ ] charindex_helper(varbinary , anyelement ) -  Calls **tsql_charindex_binary C function**
- [ ] charindex_helper( binary , anyelement ) -> Calls **tsql_charindex_binary C function**
- [ ] charindex( varbinary , varchar ) -> Prevents implicit VARCHAR to VARBINARY conversion
- [ ] charindex(binary , varchar ) -> Prevents implicit VARCHAR to BINARY conversion

2. Added tsql_charindex_char C function to convert charindex with VAR[CHAR] args.

3.  As said above, we created sys function with anyelement type to support mixed arguments.

* **Use case based**
* **Boundary conditions**
* **Arbitrary inputs**
* **Negative test cases**
* **Dependent object Tests**
* **large Input Tests**
* **Collation Tests**
* **Mixed Input tests**
* **NULL Tests**

### Check List


contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -5107,3 +5109,127 @@ get_bbf_pivot_tuplestore(const char 	*sourcetext,
...
+{
+    bytea *expressionToFind;
+    bytea *expressionToSearch;
+    int32 start_location;
+    int result = 0;
+    
+    /* Check for NULL pattern and throw error */
+    if (PG_ARGISNULL(0))
+        ereport(ERROR,
+                (errcode(ERRCODE_NULL_VALUE_NOT_ALLOWED),
```
forestkeeper: Error code ERRCODE_NULL_VALUE_NOT_ALLOWED looks not match the err msg ?

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -5107,3 +5109,127 @@ get_bbf_pivot_tuplestore(const char 	*sourcetext,
...
+    Oid collation;
+
+    /* Handle NULL inputs */
+    if (PG_ARGISNULL(0) || PG_ARGISNULL(1))
+        PG_RETURN_NULL();
+
+    /* Get function arguments */
+    expressionToFind = PG_GETARG_TEXT_PP(0);
+    expressionToSearch = PG_GETARG_TEXT_PP(1);
+    start_location = PG_GETARG_INT32(2);
```
ahmed-shameem: Let's have NULL check for 3rd arg as well. Else, while fetching the value like this, server may crash. Also add such test case.

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -5107,3 +5109,127 @@ get_bbf_pivot_tuplestore(const char 	*sourcetext,
...
+    
+    /* Handle empty search pattern case */
+    if (VARSIZE_ANY_EXHDR(expressionToFind) == 0)
+        PG_RETURN_INT32(0);
+    
+    /* Get collation from function call */
+    collation = PG_GET_COLLATION();
+
+    /* If no collation specified, use default */
+    if (!OidIsValid(collation))
```
ahmed-shameem, line 5202: Ideally this should not happen. We should have error condition here instead of assigning DEFAULT COLLATION. Let's throw error llike check_collation_set function

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -5107,3 +5109,127 @@ get_bbf_pivot_tuplestore(const char 	*sourcetext,
...
+        /* Search from the beginning of the string using textpos with collation */
+        result = DatumGetInt32(DirectFunctionCall2Coll(textpos,
+                                                     collation,    
+                                                     PointerGetDatum(expressionToSearch),
+                                                     PointerGetDatum(expressionToFind)));
+    }
+    else
+    {
+        /* Get substring starting from start_location */
+        text *substr = DatumGetTextP(DirectFunctionCall2(text_substr_no_len,
```
ahmed-shameem: Let's use DatumGetTextPP instead

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -5107,3 +5109,127 @@ get_bbf_pivot_tuplestore(const char 	*sourcetext,
...
+        int pos = DatumGetInt32(DirectFunctionCall2Coll(textpos,
+                                                      collation,
+                                                      PointerGetDatum(substr),
+                                                      PointerGetDatum(expressionToFind)));
+        
+        /* Calculate final position if found */
+        if (pos > 0)
+            result = pos + start_location - 1;
+        
+        pfree(substr);
```
ahmed-shameem: Let's have a sanity check whether substr is NULL or not. Also, what will be the impact on pos if substr is NULL?

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -5107,3 +5109,127 @@ get_bbf_pivot_tuplestore(const char 	*sourcetext,
...
+    start_location = PG_GETARG_INT32(2);
+    
+    /* Handle empty pattern */
+    if (VARSIZE_ANY_EXHDR(expressionToFind) == 0)
+        PG_RETURN_INT32(0);
+
+    if (start_location <= 0)
+    {
+        /* Use byteapos directly */
+        result = DatumGetInt32(DirectFunctionCall2(byteapos,
```
ahmed-shameem, line 5287: Same comments for tsql_charindex_char apply here as well

contrib/babelfishpg_tsql/sql/sys_functions.sql
```diff
@@ -915,28 +915,25 @@ $BODY$
...
-	END
-END;
-$BODY$
-STRICT
-LANGUAGE SQL IMMUTABLE;
+-- This function returns substring first position in expressionToSearch
+CREATE OR REPLACE FUNCTION sys.charindex(expressionToFind TEXT,
+                                        expressionToSearch TEXT,
+                                        start_location INTEGER DEFAULT 0)
+RETURNS INTEGER AS 'babelfishpg_tsql', 'tsql_charindex_char'
```
ahmed-shameem: return type should be BIGINT if expressionToSearch has an nvarchar(max), varbinary(max), or varchar(max) data type. Else INT should be returned. How do we handle that? Although the underlying textpos function returns INT, I do not see any value in updating the return type as the limitation is coming from PG. Let's think about it

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -5107,3 +5109,127 @@ get_bbf_pivot_tuplestore(const char 	*sourcetext,
...
+                                                      PointerGetDatum(expressionToFind)));
+        
+        /* Calculate final position if found */
+        if (pos > 0)
+            result = pos + start_location - 1;
+        
+        pfree(substr);
+    }
+    
+    PG_RETURN_INT32(result);
```
ahmed-shameem, line 5251: If we want to return BIGINT, will have to update these calls as well

test/JDBC/input/functions/string_functions/charindex-vu-prepare.sql
```diff
@@ -0,0 +1,83 @@
...
+    ID INT,
+    Content NVARCHAR(100) COLLATE Japanese_CI_AS
+);
+INSERT INTO charindex_tests.JapaneseTest VALUES(1, N'こんにちは'), (2, N'テスト'),(3, N'日本語');
+GO
+
+CREATE TABLE charindex_tests.UnicodeTest (
+    ID INT,
+    Content_CI NVARCHAR(100) COLLATE Latin1_General_CI_AS,  -- Case Insensitive
+    Content_CS NVARCHAR(100) COLLATE Latin1_General_CS_AS   -- Case Sensitive
```
ahmed-shameem, line 71: Good test coverage! Lets have columns having collation CI_AI and CS_AI. IIRC, there may been some issues with accent insensitive collations. The CI_AI/ CS_AI collation should be something OTHER than what is mentioned in: https://github.com/babelfish-for-postgresql/babelfish-for-postgresql/releases/tag/BABEL_4_2_0__PG_16_3#:~:text=Added%20support%20for%20LIKE,traditional_spanish_ci_ai

test/JDBC/input/functions/string_functions/charindex-vu-verify.sql
```diff
@@ -0,0 +1,200 @@
+-- Basic syntax: CHARINDEX(substring, string [, start_location])
+SELECT CHARINDEX('abc', 'xxabcxx');  -- Returns 3
```
ahmed-shameem, line 6: Let's have test where 3rd arg is -ve

test/JDBC/input/functions/string_functions/charindex-vu-verify.sql
```diff
@@ -0,0 +1,200 @@
...
+GO
+
+-- Case sensitive (depends on collation)
+SELECT CHARINDEX('ABC', 'xxabcxx');  -- Returns 0 or 3 depending on collation
+GO
+SELECT CHARINDEX('abc', 'xxABCxx');  -- Returns 0 or 3 depending on collation
+GO
+
+-- Force case sensitivity
+SELECT CHARINDEX('ABC', 'xxabcxx' COLLATE Latin1_General_CS_AS);  -- Returns 0
```
ahmed-shameem, line 40: Add test where collation of both args are different, we expect to throw error

test/JDBC/input/functions/string_functions/charindex-vu-verify.sql
```diff
@@ -0,0 +1,200 @@
...
+GO
+
+-- Returns first occurrence test
+SELECT CHARINDEX('a', 'aaa');  -- Returns 1
+GO
+SELECT CHARINDEX('a', 'aaa', 2);  -- Returns 2
+GO
+
+-- Special character
+SELECT CHARINDEX(' ', 'Hello World');  -- Returns 6
```
ahmed-shameem, line 79: Let's check for wildcard characters like %, [], - etc

test/JDBC/input/functions/string_functions/charindex-vu-verify.sql
ahmed-shameem: Also add another test like select charindex('abc' collate DATABASE_DEFAULT, 'abc' collate latin1_general_ci_ai). This should also throw error. 
Let's have explicit collation on both args with same collations as well

test/JDBC/input/functions/string_functions/charindex-vu-verify.sql
ahmed-shameem, line 79: What is the expressionToSearch is empty?

test/JDBC/input/functions/string_functions/charindex-vu-verify.sql
```diff
@@ -0,0 +1,200 @@
+-- Basic syntax: CHARINDEX(substring, string [, start_location])
+SELECT CHARINDEX('abc', 'xxabcxx');  -- Returns 3
+GO
+SELECT CHARINDEX('abc', 'xxabcxx', 1);  -- Returns 3
+GO
+SELECT CHARINDEX('abc', 'xxabcxx', 4);  -- Returns 0
```
ahmed-shameem, line 10: What if the 3rd arg is NULL?

ahmed-shameem (changes requested):

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -5107,3 +5109,127 @@ get_bbf_pivot_tuplestore(const char 	*sourcetext,
...
+    bytea *expressionToFind;
+    bytea *expressionToSearch;
+    int32 start_location;
+    int result = 0;
+    
+    /* Check for NULL pattern and throw error */
+    if (PG_ARGISNULL(0))
+        ereport(ERROR,
+                (errcode(ERRCODE_NULL_VALUE_NOT_ALLOWED),
+                errmsg("Argument data type is invalid for argument 2 of charindex function.")));
```
rohit01010: Why are we throwing invalid argument data type error when pattern is NULL, can you test following query
```
select charindex(cast(NULL as varbinary), 0x12345678)
go
```

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -5107,3 +5109,127 @@ get_bbf_pivot_tuplestore(const char 	*sourcetext,
...
+{
+    text *expressionToFind;
+    text *expressionToSearch;
+    int32 start_location;
+    int result = 0;
+    Oid collation;
+
+    /* Handle NULL inputs */
+    if (PG_ARGISNULL(0) || PG_ARGISNULL(1))
+        PG_RETURN_NULL();
```
rohit01010: The corresponding SQL function which uses this is already marked as STRICT which returns NULL on NULL argument, why do we need to check explicitly here as well ?

contrib/babelfishpg_tsql/sql/sys_functions.sql
```diff
@@ -915,28 +915,25 @@ $BODY$
...
+                                        expressionToSearch VARBINARY,
+                                        start_location INTEGER DEFAULT 0)
+RETURNS INTEGER AS 'babelfishpg_tsql', 'tsql_charindex_binary'
+LANGUAGE C IMMUTABLE PARALLEL SAFE;
+
+CREATE OR REPLACE FUNCTION sys.charindex(expressionToFind BINARY,
+                                        expressionToSearch BINARY,
+                                        start_location INTEGER DEFAULT 0)
+RETURNS INTEGER AS 'babelfishpg_tsql', 'tsql_charindex_binary'
+LANGUAGE C IMMUTABLE PARALLEL SAFE;
```
rohit01010: can you test charindex with UDTs created on binary and varbinary. For example, 
```
declare @a varbinary_udt = 0x656667
declare @b varbinary_udt = 0x707172656667
select charindex(@a, @b)
go
```

rohit01010 (changes requested):

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--5.2.0--5.3.0.sql
```diff
@@ -463,6 +463,39 @@ FROM sys.babelfish_namespace_ext sch
 WHERE sch.nspname = t.typnamespace::regnamespace::name
 	AND t.typtypmod = -1
 	AND t.typtype = 'd';
+DO $$
+DECLARE
+    exception_message text;
+BEGIN
+    ALTER FUNCTION sys.charindex RENAME TO charindex_deprecated_in_5_3_0;
```
rohit01010: Use `sys.charindex(PG_CATALOG.TEXT, PG_CATALOG.TEXT, INTEGER)` explicitly, else it will raise error 'function name "sys.charindex" is not unique' if we are trying to upgrade from a lower version also has this fix to this version. Although we had added the exception handling which will convert that error to warning, but its better not to throw such incorrect warnings.

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--5.2.0--5.3.0.sql
```diff
@@ -463,6 +463,39 @@ FROM sys.babelfish_namespace_ext sch
...
+EXCEPTION WHEN OTHERS THEN
+    GET STACKED DIAGNOSTICS
+    exception_message = MESSAGE_TEXT;
+    RAISE WARNING '%', exception_message;
+END;
+$$;
+
+CALL sys.babelfish_drop_deprecated_object('function', 'sys', 'charindex_deprecated_in_5_3_0');
```
rohit01010: As we are only updating the body, language, strictness and parallel safety of the function and not the definition(i.e. the number of arguments or argument types or return type), i do not see any point in deprecating the function. so why are we deprecating this function ?

test/JDBC/expected/charindex-vu-prepare.out
```diff
@@ -0,0 +1,105 @@
...
+CREATE TABLE charindex_tests.TestFixedNChar (col1 nchar(20))
+INSERT INTO charindex_tests.TestFixedNChar VALUES (N'Hello World')
+GO
+~~ROW COUNT: 1~~
+
+
+-- Create tables with specific collations
+CREATE TABLE charindex_tests.ChineseTest (
+    ID INT,
+    Content NVARCHAR(100) COLLATE Chinese_PRC_CI_AS
```
rohit01010: Lets also test other string datatypes columns such as VARCHAR, CHAR, NCHAR, with specific collation.

test/JDBC/expected/charindex-vu-prepare.out
```diff
@@ -0,0 +1,105 @@
...
+-- Insert test data
+INSERT INTO charindex_tests.SpecialBinary (BinaryData, VarbinaryData, Description)
+VALUES 
+    (CAST('Hello123' AS BINARY(50)), CAST('Hello123World' AS VARBINARY(MAX)), 'Basic test'),
+    (0x48656C6C6F313233, 0x48656C6C6F313233576F726C64, 'Hex encoded Hello123World'),
+    (CAST(REPLICATE('A', 50) AS BINARY(50)), CAST(REPLICATE('A', 1000) AS VARBINARY(MAX)), 'Repeated characters'),
+    ( CAST('Test©®™' AS BINARY(50)), CAST('Special©Characters®™' AS VARBINARY(MAX)), 'Special characters' );
+GO
+~~ROW COUNT: 4~~
+
```
rohit01010, line 133: Lets also add tests for dependent objects(views, procedures, functions, computed columns, constraint) on charindex, to tests its behaviour in the upgrade scenario.

test/JDBC/upgrade/latest/schedule
```diff
@@ -628,3 +628,4 @@ babel_test_numeric_int8_oper
 Test_cast_binary_bytea
 numeric_money
 MONEY_OBJECT
+charindex
```
rohit01010, line 658: Lets add charindex to schedule files of other versions to test the charindex tests in upgrades from other versions as well.

contrib/babelfishpg_tsql/runtime/functions.c
herambh-shah: Thanks @forestkeeper for pointing this out. I believe we should change it to "NULL" instead of throwing error.

herambh-shah: And as per our discussion for cases like charindex(NULL, varbinary) where it throws error, will try to handle it in analyzer.

contrib/babelfishpg_tsql/runtime/functions.c
herambh-shah: updated
 the sys function to STRICT so that we wont have to handle it explicitly.

contrib/babelfishpg_tsql/runtime/functions.c
herambh-shah: I've added a "if" check in tsql_charindex_char function to check invalid collation.

contrib/babelfishpg_tsql/runtime/functions.c
herambh-shah: I think we can skip this because DatumGetTextPP will never return NULL value.

contrib/babelfishpg_tsql/runtime/functions.c
herambh-shah: got it.

test/JDBC/input/functions/string_functions/charindex-vu-verify.sql
herambh-shah: Added !

test/JDBC/input/functions/string_functions/charindex-vu-verify.sql
herambh-shah: yes , added . Thanks

herambh-shah: It gives "0". Added more test case for spaces and empty string
SELECT CHARINDEX('Hello World', '');  -- return 0
SELECT CHARINDEX('', '');   -- return 0
SELECT CHARINDEX('              ', '    '); (10 spaces , 3 spaces).  -- return 0

test/JDBC/input/functions/string_functions/charindex-vu-verify.sql
herambh-shah: it gives NULL.

contrib/babelfishpg_tsql/runtime/functions.c
herambh-shah: I've removed above condition.
But in sql server, this is the behavior, I observed for NULL with varbinary arguments -
CASE 1: charindex(NULL, var[binary] ) -> **throws error**
CASE 2: charindex(d , var[binary]) -> _where col d is having NULL values_ -> **gives NULL**
when we provide constant NULL value in charindex function along with varbinary argument, it seems sql server throws error
But if NULL value is given through a column, it returns NULL.
It might be the case with other functions also.

contrib/babelfishpg_tsql/runtime/functions.c
herambh-shah: yes, we actually don't need it . I was following other function convection and thought should use it as a pre-check. Added STRICT in sys definition.

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--5.2.0--5.3.0.sql
herambh-shah: ohkk. I didn't know it before. I thought whenever we update any sys_functions, we need to add it in upgrade file also. Thanks for noticing it. I will remove it from upgrade file.

contrib/babelfishpg_tsql/sql/sys_functions.sql
herambh-shah: yes, I also thought about it after reading charindex docmentation but not sure about how to handle it and create sys function accordingly.

contrib/babelfishpg_tsql/runtime/functions.c
herambh-shah: yes

test/JDBC/expected/charindex-vu-prepare.out
herambh-shah: Added. Thanks

test/JDBC/expected/charindex-vu-prepare.out
herambh-shah: Added .

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--5.2.0--5.3.0.sql
herambh-shah: Got it. updated as per above suggestion. Thanks

contrib/babelfishpg_tsql/runtime/functions.c
ahmed-shameem: DatumGetTextPP can return pointer to null. Here you are obtaining the substring using a function, what if that function returns NULL itself? There are case where it will happen

test/JDBC/input/functions/string_functions/charindex-vu-verify.sql
ahmed-shameem: Kindly add tests for the same

test/JDBC/input/functions/string_functions/charindex-vu-verify.sql
ahmed-shameem: Kindly add tests for the same

test/JDBC/input/functions/string_functions/charindex-vu-verify.sql
herambh-shah: I was not able to add test case with different collation as during default collation PR check, it seems it is changing the collation type of input and thus creating flaky test case for us.
Else I've added all other scenerios

contrib/babelfishpg_tsql/sql/sys_functions.sql
herambh-shah: Added

contrib/babelfishpg_tsql/runtime/functions.c
herambh-shah: instead of throwing error message here, I've updated sys function to "STRICT"

contrib/babelfishpg_tsql/runtime/functions.c
herambh-shah: ok added a sanity check on substr

forestkeeper: I actually have one more question, what makes it so necessary to free substr at this point ?

herambh-shah: It seems DirectFunctionCall2(text_substr_no_len,..) creates new text object for substr. So that's why I've used pfree(substr) to free the allocated memory.

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -5107,3 +5109,152 @@ get_bbf_pivot_tuplestore(const char 	*sourcetext,
...
+        /* Find position in substring */
+        pos = DatumGetInt32(DirectFunctionCall2(byteapos,
+                                            PointerGetDatum(substr),
+                                            PointerGetDatum(expressionToFind)));
+        
+        /* Adjust position if found */
+        if (pos > 0)
+            result = pos + start_location - 1;
+        
+        pfree(substr);
```
forestkeeper: Why have to free substr at this point ? From my understand PG will release function memory context right after the function execution.

herambh-shah: Hi @forestkeeper 
I thought I memory that we allocated gets free only after query/transantion gets ended.
Since the memory for substr has been allocated in text_substring PG function itself before returning the string. 
so I thought it is better to free it in charindex function itself as we don't need it anymore.

herambh-shah: Alternatively, we can use xpfree whose definition written in the same file -
```
#define xpfree(var_) \
	do { \
		if (var_ != NULL) \
		{ \
			pfree(var_); \
			var_ = NULL; \
		} \
	} while (0)
```
This way we dont have to do sanity check on substr also?
what do you suggest @forestkeeper @rohit01010  ?

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -5107,3 +5109,152 @@ get_bbf_pivot_tuplestore(const char 	*sourcetext,
...
+    Oid collation;
+
+    /* Handle empty search pattern case */
+    if (VARSIZE_ANY_EXHDR(expressionToFind) == 0)
+        PG_RETURN_INT32(0);
+
+    /* Get collation from function call */
+    collation = PG_GET_COLLATION();
+
+    /* If no collation specified, use default */
```
rohit01010: This comment seems irrelevant

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -5107,3 +5109,152 @@ get_bbf_pivot_tuplestore(const char 	*sourcetext,
...
+    /* final result */
+    int result = 0;
+
+    /* Check if conversion failed - Should never happen */
+    if (expressionToSearch == NULL)
+    {
+        ereport(ERROR,
+                (errcode(ERRCODE_INTERNAL_ERROR),
+                errmsg("Failed to convert input to binary type")));
+    }
```
rohit01010, line 5276: why is this required ?

contrib/babelfishpg_tsql/runtime/functions.c
rohit01010: Yeah using xpfree looks safe to me

contrib/babelfishpg_tsql/sql/sys_functions.sql
```diff
@@ -915,28 +915,91 @@ $BODY$
...
+$$
+LANGUAGE plpgsql IMMUTABLE STRICT PARALLEL SAFE;
+
+CREATE OR REPLACE FUNCTION sys.charindex(expressionToFind sys.VARCHAR,
+                                        expressionToSearch ANYELEMENT,
+                                        start_location INTEGER DEFAULT 0)
+RETURNS INTEGER
+AS $$
+BEGIN
+    RETURN sys.charindex_helper(expressionToFind, expressionToSearch::VARCHAR, start_location);
```
rohit01010, line 994: why we are using VARCHAR (i.e. PG_CATALOG.VARCHAR) instead of sys.VARCHAR?

contrib/babelfishpg_tsql/sql/sys_functions.sql
```diff
@@ -915,28 +915,91 @@ $BODY$
...
+$$
+LANGUAGE plpgsql IMMUTABLE STRICT PARALLEL SAFE;
+
+CREATE OR REPLACE FUNCTION sys.charindex(expressionToFind sys.VARCHAR,
+                                        expressionToSearch sys.VARCHAR,
+                                        start_location INTEGER DEFAULT 0)
+RETURNS INTEGER
+AS $$
+BEGIN
+    RETURN sys.charindex_helper(expressionToFind, expressionToSearch::VARCHAR, start_location);
```
rohit01010, line 1027: why we are using VARCHAR (i.e. PG_CATALOG.VARCHAR) instead of sys.VARCHAR?

contrib/babelfishpg_tsql/sql/sys_functions.sql
```diff
@@ -915,28 +915,91 @@ $BODY$
...
+END;
+$$
+LANGUAGE plpgsql IMMUTABLE STRICT PARALLEL SAFE;
```
rohit01010, line 1030: why we need this definition we can just make it a C function which charindex(varchar, anyelement) function, hence we can remove the TEXT, TEXT definition

test/JDBC/upgrade/latest/schedule
herambh-shah: Added

contrib/babelfishpg_tsql/sql/sys_functions.sql
herambh-shah: as discussed, some of the collation test case are failing if we don't define VARCHAR here. 
collation information is not getting passed if we defined t as sys.VARCHAR.

contrib/babelfishpg_tsql/sql/sys_functions.sql
herambh-shah: Same question as below.

contrib/babelfishpg_tsql/runtime/functions.c
herambh-shah: In cast, implicit type casting failed to binary. It's just for safety check.

contrib/babelfishpg_tsql/sql/sys_functions.sql
herambh-shah: as discussed before, if we dont keep this, then we were getting error that - **could not find one candidate for ( VARCHAR, VARCHAR ) input.**

contrib/babelfishpg_tsql/sql/sys_functions.sql
```diff
@@ -962,28 +962,91 @@ $BODY$
...
+CREATE OR REPLACE FUNCTION sys.charindex(expressionToFind sys.BBF_BINARY,
+                                        expressionToSearch sys.VARCHAR,
+                                        start_location INTEGER DEFAULT 0)
+RETURNS INTEGER
+AS $$
+BEGIN
+    RAISE EXCEPTION 'Implicit Casting from varchar to binary is not allowed'; 
+END;
+$$
+LANGUAGE plpgsql IMMUTABLE STRICT PARALLEL SAFE;
```
rohit01010, line 1019: function call charindex(sys.varbinary/sys.binary, sys.nvarchar) will also hit these functions. throwing error 
'Implicit Casting from varchar to binary is not allowed' might be misleading for it. Lets change the error message to 'Implicit Casting from (n)varchar to binary is not allowed'

contrib/babelfishpg_tsql/sql/sys_functions.sql
```diff
@@ -962,28 +962,91 @@ $BODY$
...
+                                        expressionToSearch sys.BBF_BINARY,
+                                        start_location INTEGER DEFAULT 0)
+RETURNS INTEGER AS 'babelfishpg_tsql', 'tsql_charindex_binary'
+LANGUAGE C IMMUTABLE STRICT PARALLEL SAFE;
+
+CREATE OR REPLACE FUNCTION sys.charindex_helper(expressionToFind TEXT,
+                                        expressionToSearch TEXT,
+                                        start_location INTEGER DEFAULT 0)
+RETURNS INTEGER AS 'babelfishpg_tsql', 'tsql_charindex_char'
+LANGUAGE C IMMUTABLE STRICT PARALLEL SAFE;
```
rohit01010, line 1048: shouldn't the argument type be sys.varchar instead of TEXT here ?

contrib/babelfishpg_tsql/sql/sys_functions.sql
rohit01010: since we want to use sys.charindex_helper(TEXT, TEXT, INT) function, shouldn't we cast the arguments to TEXT instead ?

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--5.3.0--5.4.0.sql
```diff
@@ -307,6 +307,104 @@ $$;
...
+$$
+LANGUAGE plpgsql IMMUTABLE STRICT PARALLEL SAFE;
+
+CREATE OR REPLACE FUNCTION sys.charindex(expressionToFind sys.BBF_VARBINARY,
+                                        expressionToSearch sys.VARCHAR,
+                                        start_location INTEGER DEFAULT 0)
+RETURNS INTEGER
+AS $$
+BEGIN
+   RAISE EXCEPTION 'Implicit Casting from varchar to varbinary is not allowed';
```
rohit01010, line 363: Is similar handling required for other datatypes which are not implicitly castable to sys.varbinary/sys.binary/sys.varchar ?

contrib/babelfishpg_tsql/runtime/functions.c
rohit01010: Do we still need this check ?

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--5.3.0--5.4.0.sql
herambh-shah: yes It is possible implicit casting from one type to another is restricted but we cannot create sys function with that many combination.
The reason I have added for var[char]-> var[binary] is because it is mandatory to have **var[char] , var[binary]** as **expressiontoFind** in charindex function. Hence we should have at least error message for this types as we may expects users to run charindex on these types.

contrib/babelfishpg_tsql/sql/sys_functions.sql
herambh-shah: yes changed it

contrib/babelfishpg_tsql/sql/sys_functions.sql
herambh-shah: changed. Thanks

PR #3814, opened 2025-06-09
Update function signature for OBJECT_DEFINITION

shardgupta:
### Description
Currently OBJECT_DEFINITION function return type is defined as sys.NVARCHAR(4000). This limits the object definition being return to 4000 bytes pair. This PR removes this limitation. 

### Issues Resolved

Task: BABEL-5057

* **Use case based -** - Added objects with byte pair size greater than 4000.

* **Minor version upgrade tests -** - Added

* **Major version upgrade tests -** - Added

* **Performance tests -** - NA


contrib/babelfishpg_tsql/sql/sys_functions.sql
```diff
@@ -4390,11 +4390,11 @@ RETURNS sys.SYSNAME AS
...
 LANGUAGE C STABLE;
 
-CREATE OR REPLACE FUNCTION OBJECT_DEFINITION(IN object_id INT)
-RETURNS sys.NVARCHAR(4000)
+CREATE OR REPLACE FUNCTION sys.OBJECT_DEFINITION(IN object_id INT)
+RETURNS sys.NVARCHAR
 AS $$
 DECLARE
-    definition sys.nvarchar(4000);
+    definition sys.NVARCHAR;
```
Yvinayak07, line 4410: As `NVARCHAR_MAX_SCALE` is `4000`, How are we increasing the limitation of 4000byte pair by changing NVARCHAR(4000) to NVARCHAR?

test/JDBC/input/BABEL_OBJECT_DEFINITION-vu-prepare.mix
```diff
@@ -87,3 +87,342 @@ GO
 
 CREATE FUNCTION object_definition_function() RETURNS [VARCHAR](8000) AS BEGIN DECLARE @storeOriginalQuery_var [VARCHAR](8000) RETURN @storeOriginalQuery_var END
 GO
+
+-- Test OBJECT_DEFINITION with objects having definitions larger than 4000 characters
+-- Create a stored procedure with a large definition
+CREATE PROCEDURE dbo.LargeProc
```
shalinilohia50: Lets rename these large procedure/view/function names as `object_definition_Largeproc/view/func` to comply with the existing objects in this test. It helps in debugging the failures in future if the names are distinct and descriptive.

contrib/babelfishpg_tsql/sql/sys_functions.sql
shalinilohia50: That's in T-SQL, where we handle typmod for this datatype in parser to not exceed 4000. 
Here, we are updating the definition in PG.

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--5.2.0--5.3.0.sql
```diff
@@ -621,6 +621,32 @@ WHERE sch.nspname = t.typnamespace::regnamespace::name
 	AND t.typtypmod = -1
 	AND t.typtype = 'd';
 
+CREATE OR REPLACE FUNCTION sys.OBJECT_DEFINITION(IN object_id INT)
+RETURNS sys.NVARCHAR
```
KushaalShroff, line 625: I think we need to deprecate the older version in the upgrade script and replace it with the new one.

Please also add upgrade testing with dependent objects

KushaalShroff (approved):

PR #3817, opened 2025-06-09
Updated escape logic in quotename

herambh-shah:
### Description

Issue Description: 
1. In Babelfish, the QUOTENAME function currently produces incorrect results for many queries due to flaws in its escape character handling logic. The function fails to properly escape characters that match the right delimiter.
2. Also handle Unicode character handling in delimeter.
3. Changes logic to accept fetch first character from delimeter string and use it as delimeter

Solution: 
1. The escape logic has been updated to properly handle special characters. Now, when QUOTENAME encounters any character within the input string that matches the right delimiter, it correctly escapes that character.
2. Updated logic to check byte size instead of len check.

### Issues Resolved

BABEL - 2311

* **Use case based -**
* **Boundary conditions -**
* **Arbitrary inputs -**

### Check List


pranavJ23:
Just a thought, BBF returns NULL when quote_character is not specified. As per the tsql documentation if quote_character is not specified, brackets are used. So should we address this behaviour too and add test case related to this scenario ?
query: 
```select quotename('hello','')```

**Comment by herambh-shah** [comment]
> Just a thought, BBF returns NULL when quote_character is not specified. As per the tsql documentation if quote_character is not specified, brackets are used. So should we address this behaviour too and add test case related to this scenario ? query: `select quotename('hello','')`



> Just a thought, BBF returns NULL when quote_character is not specified. As per the tsql documentation if quote_character is not specified, brackets are used. So should we address this behaviour too and add test case related to this scenario ? query: `select quotename('hello','')`

**Comment by herambh-shah** [comment]
> yeah I debug the code and since the 2nd argument of quotename is of type "char". Hence , it is also considering it as delimeter. I will update it in code. Thanks @pranavJ23 for noticing it

File: test/JDBC/expected/babel_function_string-vu-verify.out [hunk 1]
```diff
@@ -385,6 +385,78 @@ nvarchar
...
+
+SELECT QUOTENAME('"KEY"', '[')
+GO
+~~START~~
+nvarchar
+["KEY"]
+~~END~~
+
+
+SELECT QUOTENAME('"KEY"', '')
```
Deepesh125, line 452: can you please add some test cases for multibyte characters? with various combination with length

test/JDBC/input/datatypes/Datetime_system_functions.sql
```diff
@@ -974,7 +974,7 @@ GO
 
 -- Test 15: Testing comparison with local time
 SELECT
-    CASE WHEN ABS(DATEDIFF(HOUR, SYSUTCDATETIME(), SYSDATETIME())) = 
+    CASE WHEN ABS(DATEDIFF(HOUR, SYSUTCDATETIME(), SYSDATETIME())) >
```
Deepesh125, line 977: why is this needed?

contrib/babelfishpg_tsql/src/string.c
```diff
@@ -230,17 +234,16 @@ quotename(PG_FUNCTION_ARGS)
...
 	for (int i = 0; i < strlen(input_string); i++)
 	{
-		switch (input_string[i])
+        if (input_string[i] == right_delim)
+        {
+            /* Escape the delimiter character by doubling it */
+            buf[buf_i++] = input_string[i];
+            buf[buf_i++] = input_string[i];
+        }
+        else
```
rohit01010: NIT: fix the indentation

test/JDBC/input/datatypes/Datetime_system_functions.sql
Anikait143: This was a flaky test, so I requested for this slight fix.

contrib/babelfishpg_tsql/src/string.c
herambh-shah: Fixed

herambh-shah:
Updated logic a little bit to handle below cases -
1. Unicode character in delimeter 
   for eg - `SELECT quotename('Москва}test', 'Ğ🚀🌟')`
2. if size(deliemeter) >1 , consider first character-
`SELECT quotename('Москва}test', '}\^{')`
`Returns - {'Москва}}test'} `

test/JDBC/expected/babel_function_string-vu-verify.out
herambh-shah: Added

contrib/babelfishpg_tsql/src/string.c
```diff
@@ -199,8 +207,8 @@ quotename(PG_FUNCTION_ARGS)
 		case '`':
 		case '\'':
 		case '"':
```
pranavJ23, line 209: can we add a test case for these delimiters with multiple occurences ? 
like :
```
select quotename('user''s "custom" name','[')
select quotename('user''s "custom" name','''')
```
and let's also verify the result with tsql

otherwise changes looks good to me

rohit01010 (approved):

pranavJ23 (approved):

forestkeeper (approved):

PR #3820, opened 2025-06-10
Add support for OPENXML with WITH clause

harshdubey166:
### Description

OPENXML provides a rowset view over an XML document. Since  OPENXML is a rowset provider and it returns a set of rows, we can use OPENXML in the FROM clause of a T-SQL statement just as we can use any other table, view, or table-valued function. The WITH clause in OPENXML provides a rowset format (and additional mapping information as required) by using either SchemaDeclaration or specifying an existing TableName.
Currently Babelfish does not have support for openxml. So the primary objective is to add support for OPENXML with WITH clause.
The syntax of openxml is as follows :
```
OPENXML ( idoc int [ in ] , rowpattern nvarchar [ in ], [ flags byte [ in ] ] )
    [ WITH ( SchemaDeclaration | TableName ) ]
```
Arguments

* idoc
    * The document handle of the internal representation of an XML document. The internal representation of an XML document is created by calling sp_xml_preparedocument.
* rowpattern
    * The XPath pattern used to identify the nodes to be processed as rows. The nodes come from the XML document whose handle is passed in the idoc parameter.
* flags
    * Indicates the mapping used between the XML data and the relational rowset, and how the spill-over column is filled. flags is an optional input parameter, and can be one of the following values.

The WITH clause provides a rowset format (and additional mapping information as required) by using either SchemaDeclaration or specifying an existing TableName. If the optional WITH clause isn't specified, the results are returned in an edge table format. Edge tables represent the fine-grained XML document structure (such as element/attribute names, the document hierarchy, the namespaces, PIs, and so on) in a single table.

Engine PR : https://github.com/babelfish-for-postgresql/postgresql_modified_for_babelfish/pull/587
Signed-off-by: Harsh Dubey [harshdu@amazon.com](mailto:harshdu@amazon.com)
Authored-by: Harsh Dubey [harshdu@amazon.com](mailto:harshdu@amazon.com)

**Issues Resolved**

BABEL-3635, BABEL-6045

### Check List


contrib/babelfishpg_tsql/src/backend_parser/gram-tsql-rule.y
```diff
@@ -1779,13 +1779,129 @@ table_ref:	relation_expr tsql_table_hint_expr
 					 */
 					$$ = (Node *) $1;
 				}
+			| TSQL_APPLY openxml_expr
+				{
+					/*
+					 * This case handles openxml cross/outer apply
+					 */
+					$$ = (Node *) $2;
```
HarshLunagariya, line 1801: Please add test case for APPLY.

contrib/babelfishpg_tsql/src/backend_parser/gram-tsql-rule.y
```diff
@@ -1779,13 +1779,129 @@ table_ref:	relation_expr tsql_table_hint_expr
...
+openxml_column_el:
+			ColId Typename
+				{
+					RangeTableFuncCol *fc = makeNode(RangeTableFuncCol);
+
+					fc->colname = $1;
+					fc->typeName = $2;
+					/* Set a default XPath expression based on column name */
+					/* fc->colexpr = (Node *) makeStringConst(psprintf("@%s", $1), @1); */
+					/* fc->colexpr = makeStringConst($1, @1); */
```
HarshLunagariya: Any specific reason to comment it ?, Let's remove it if it's not required.

contrib/babelfishpg_tsql/src/backend_parser/parser.c
```diff
@@ -266,6 +266,8 @@ pgtsql_base_yylex(YYSTYPE *lvalp, YYLTYPE * llocp, core_yyscan_t yyscanner)
 				case '(':
 					cur_token = WITH_paren;
 					break;
+				case TABLE:
+					cur_token = WITH_table;
```
HarshLunagariya, line 270: Why do we do this? To avoid any SR conflicts?

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -6526,6 +6532,170 @@ pltsql_exprTypmod(Plan *plan, Node *expr)
...
+		/* Create RangeTableFuncCol nodes from the table's columns */
+		for (i = 0; i < tupdesc->natts; i++)
+		{
+			Form_pg_attribute att = TupleDescAttr(tupdesc, i);
+			RangeTableFuncCol *fc;
+			char *colname;
+			
+			/* Skip dropped and generated columns */
+			if (att->attisdropped || att->attgenerated)
+				continue;
```
HarshLunagariya: Why are we skipping attgenerated? There might some feature relied upon this PG functionality not now but maybe in future. What if we don't skip it?

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -6526,6 +6532,170 @@ pltsql_exprTypmod(Plan *plan, Node *expr)
...
+			/* Create a column definition */
+			fc = makeNode(RangeTableFuncCol);
+			fc->colname = pstrdup(colname);
+			
+			/* Create a TypeName node for the column type */
+			fc->typeName = makeTypeNameFromOid(att->atttypid, att->atttypmod);
+			
+			/* Set the column expression to the generated XPath */
+			fc->colexpr = (Node *) makeFuncCall(list_make2(makeString("sys"), makeString("tsql_openxml_get_colpattern")), list_make2(makeStringConst(colname, -1), flag), COERCE_EXPLICIT_CALL, -1);
+			fc->coldefexpr = NULL;
```
HarshLunagariya, line 7046: This means that table schema's default expression shouldn't get picked up in the result set generated by OPENXML(). Please make sure we have a testcase for this where we are picking up schema from table having column with default value/expr or even constriants.

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -6526,6 +6532,170 @@ pltsql_exprTypmod(Plan *plan, Node *expr)
...
+ * The function ensures each column has a proper col expression using the tsql_openxml_get_colpattern function.
+ */
+static void
+transformOpenxml_expr(ParseState *pstate, Openxml_expr *expr)
+{
+
+	/* If a table reference is provided, get the columns from that table */
+	if (expr->table_ref != NULL)
+	{
+		expr->columns = fetch_table_schema(expr->table_ref, expr->tsql_flag);
```
HarshLunagariya: Should we add assert for  `(expr->columns != NULL)` here?

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -6526,6 +6532,170 @@ pltsql_exprTypmod(Plan *plan, Node *expr)
...
+		ListCell *lc;
+		foreach(lc, expr->columns)
+		{
+			RangeTableFuncCol *fc = (RangeTableFuncCol *) lfirst(lc);
+			
+			/* If no column expression is provided, generate one based on the flag */
+			if (fc->colexpr == NULL)
+			{
+				/* To get original column name, utilize location of ColumnDef and query string. For colexpr, we need orignal name of columns (no downcase or uppercase) */
+				const char *column_name_start = pstate->p_sourcetext + fc->location;
```
HarshLunagariya: fc->location will always known , right? It won't be -1 in any case.

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -4290,3 +4309,746 @@ sp_reset_connection_internal(PG_FUNCTION_ARGS)
...
+			/* For flag 2, use colname */
+			xpath_expr = pstrdup(colname);
+			break;
+		case 3:
+			/* For flag 3, use colname|@colname */
+			xpath_expr = psprintf("%s|@%s", colname, colname);
+			break;
+		default:
+			/* Default to @colname for unknown flags */
+			xpath_expr = psprintf("@%s", colname);
```
HarshLunagariya, line 5113: can colname contain '|' or '@' ? In that case, does xpath_expr need to be escaped?

jsudrik: Looks like in PG, colname can not contain 'I' or '@'. In SQL Server, colname can contain '@', but not '|'. Hope this helps.

contrib/babelfishpg_tsql/src/backend_parser/gram-tsql-rule.y
harshdubey166: Yes, this is redundant

contrib/babelfishpg_tsql/src/backend_parser/parser.c
harshdubey166: Yes, we are doing this to avoid any shift reduce conflicts

contrib/babelfishpg_tsql/sql/sys_functions.sql
```diff
@@ -4029,6 +4029,16 @@ CREATE OR REPLACE FUNCTION sys.openjson_with(json_string text, path text, VARIAD
 RETURNS SETOF RECORD
 AS 'babelfishpg_tsql', 'tsql_openjson_with' LANGUAGE C STRICT IMMUTABLE PARALLEL SAFE;
 
+CREATE OR REPLACE FUNCTION sys.tsql_openxml_get_xmldoc(int)
+RETURNS xml
+AS 'babelfishpg_tsql', 'tsql_openxml_get_xmldoc'
+LANGUAGE C STRICT IMMUTABLE PARALLEL SAFE;
```
Deepesh125: why parallel safe? IIRC, we store doc using temp table so how will it be accessing through parallel worker which is essentially different backend

contrib/babelfishpg_tsql/sql/sys_functions.sql
```diff
@@ -4029,6 +4029,16 @@ CREATE OR REPLACE FUNCTION sys.openjson_with(json_string text, path text, VARIAD
...
+
+CREATE OR REPLACE FUNCTION sys.tsql_openxml_get_colpattern(text,int)
```
Deepesh125, line 4162: Please describe use of both this helper

contrib/babelfishpg_tsql/src/tsqlIface.cpp
```diff
@@ -1253,6 +1253,16 @@ class tsqlCommonMutator : public TSqlParserBaseListener
...
 	}
 
+	void exitOpen_xml(TSqlParser::Open_xmlContext *ctx) override
+	{
+		if(ctx->table_name())
+		{
+			rewritten_query_fragment.emplace(std::make_pair(ctx->table_name()->start->getStartIndex(),
+											 std::make_pair("", "TABLE ")));
+
+		}
```
Deepesh125, line 1310: Please explain this change

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -6526,6 +6532,147 @@ pltsql_exprTypmod(Plan *plan, Node *expr)
...
+	 * Set the document expression to retrieve the XML document using the document handle. This creates a function call to tsql_openxml_get_xmldoc
+	 * which retrieves the  previously prepared XML document based on the document ID from sp_xml_preparedocument.
+	 */
+	rtf->docexpr = (Node *) makeFuncCall(list_make2(makeString("sys"), makeString("tsql_openxml_get_xmldoc")), list_make1(tsql_docid_node), COERCE_EXPLICIT_CALL, -1);
+
+	if (rtf->columns != NIL)
+	{
+		Node  *first_col = linitial(rtf->columns);
+
+		if (IsA(first_col, RangeVar))
```
Deepesh125: better would be to check length of columns is one and only node is range var.

contrib/babelfishpg_tsql/src/backend_parser/gram-tsql-rule.y
```diff
@@ -1779,13 +1779,129 @@ table_ref:	relation_expr tsql_table_hint_expr
...
 		;
 
+openxml_expr: OPENXML '(' a_expr ',' a_expr ')' opt_alias_clause
+				{
+					RangeTableFunc *n = makeNode(RangeTableFunc);
+					n->docexpr = NULL;
+					n->rowexpr = $5;
+					n->location = @1;
+					/* Default flag is 0 when not specified */
+					n->namespaces = list_make2((Node*)$3, makeIntConst(0, @1));
```
Deepesh125: can we try to use docexpr for $3?

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -6526,6 +6532,147 @@ pltsql_exprTypmod(Plan *plan, Node *expr)
...
+			foreach(lc, rtf->columns)
+			{
+				RangeTableFuncCol *fc = (RangeTableFuncCol *) lfirst(lc);
+
+				/* If no column expression is provided, generate one based on the flag */
+				if(fc->colexpr == NULL && tsql_flag != NULL)
+				{
+					/* To get original column name, utilize location of ColumnDef and query string. For colexpr, we need orignal name of columns (no downcase or uppercase) */
+					const char *column_name_start = pstate->p_sourcetext + fc->location;
+					char	   *original_name = extract_identifier(column_name_start, NULL);
```
Deepesh125: ensure we add test cases for this

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -4290,3 +4309,730 @@ sp_reset_connection_internal(PG_FUNCTION_ARGS)
...
+ *   0,1: Use attribute access (@colname)
+ *   2: Use element access (colname)
+ *   3: Try both element and attribute (colname|@colname)
+ */
+Datum
+tsql_openxml_get_colpattern(PG_FUNCTION_ARGS)
+{
+	char *xpath_expr;
+	int flag = PG_GETARG_INT32(1);
+	char *colname = text_to_cstring(PG_GETARG_TEXT_PP(0));
```
Deepesh125: check if colname is not null

contrib/babelfishpg_tsql/src/hooks.c
Deepesh125: please confirm expected TSQL behaviour as well.

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -6526,6 +6532,147 @@ pltsql_exprTypmod(Plan *plan, Node *expr)
...
+		{
+			Form_pg_attribute att = TupleDescAttr(tupdesc, i);
+			RangeTableFuncCol *fc;
+			char *colname;
+			
+			/* Skip dropped columns */
+			if (att->attisdropped)
+				continue;
+				
+			colname = NameStr(att->attname);
```
Deepesh125: please add test cases where length of table name or column name is > 64 bytes

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -4290,3 +4309,730 @@ sp_reset_connection_internal(PG_FUNCTION_ARGS)
...
+				BTEqualStrategyNumber, F_INT4EQ,
+				Int32GetDatum(document_id));
+	
+	scan = table_beginscan_catalog(relation, 1, skey);
+	tuple = heap_getnext(scan, ForwardScanDirection);
+	
+	if (HeapTupleIsValid(tuple))
+	{
+		/* Get the XML document from column 6 (doc) */
+		result = heap_getattr(tuple, 6, RelationGetDescr(relation), &isnull);
```
Deepesh125: use macro instead

Deepesh125 (changes requested):

contrib/babelfishpg_tsql/sql/sys_functions.sql
harshdubey166: The function tsql_openxml_get_colpattern set the colpattern(if not given) according to flags and the function tsql_openxml_get_xmldoc fetches the xml document using doc_id.

contrib/babelfishpg_tsql/src/tsqlIface.cpp
harshdubey166: This change is done to avoid any shift reduce conflicts during parsing stage of openxml for the case of openxml WITH tablename.

contrib/babelfishpg_tsql/src/hooks.c
harshdubey166: Yes, we have added test cases for that.

contrib/babelfishpg_tsql/src/hooks.c
harshdubey166: Yes, we should not skip the attgenerated columns. Fixed that.

contrib/babelfishpg_tsql/src/hooks.c
harshdubey166: This approach has been changed.

contrib/babelfishpg_tsql/src/hooks.c
harshdubey166: Yes, fc->location will always be known in the pretransform openxml function because these these columns are explicitly written by user in the openxml syntax.

contrib/babelfishpg_tsql/src/hooks.c
harshdubey166: Yes, added those testcases.

contrib/babelfishpg_tsql/sql/sys_functions.sql
harshdubey166: Yes, it shouldn't be parallel safe since we are using temp tables. Corrected that.

contrib/babelfishpg_tsql/src/procedures.c
harshdubey166: Added that check also.

contrib/babelfishpg_tsql/src/hooks.c
harshdubey166: Added those testcases.

contrib/babelfishpg_tsql/src/backend_parser/gram-tsql-rule.y
harshdubey166: Yes, added that.

contrib/babelfishpg_tsql/sql/sys_functions.sql
rohit01010: Lets add comments on top of these functions, with some description.

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -4923,4 +4926,137 @@ sp_xml_removedocument(PG_FUNCTION_ARGS)
 	delete_xml_handle_entry(doc_handle);
 
 	PG_RETURN_VOID();
+}
+
+/*
+ * Function to retrieve XML document from temporary table using document ID
+ */
+#define XML_HANDLE_DOC_COLUMN_NUM 6
```
rohit01010: Lets rename this to `Anum_<table-name>_<colname>`

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -4923,4 +4926,137 @@ sp_xml_removedocument(PG_FUNCTION_ARGS)
...
+
+	if (!table_exists)
+	{
+		ereport(ERROR,
+				(errcode(ERRCODE_UNDEFINED_OBJECT),
+				 errmsg("Could not find prepared statement with handle %d", document_id)));
+	}
+
+	ScanKeyInit(&skey[0],
+				1,  /* Column number */
```
rohit01010: Lets define constant `Anum_<table-name>_<colname>` for Document id column and use that here.

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -4923,4 +4926,137 @@ sp_xml_removedocument(PG_FUNCTION_ARGS)
...
+	
+	if (HeapTupleIsValid(tuple))
+	{
+		/* Get the XML document from column 6 (doc) */
+		result = heap_getattr(tuple, XML_HANDLE_DOC_COLUMN_NUM, RelationGetDescr(relation), &isnull);
+		
+		if (!isnull)
+		{
+			/* Make a copy of the value */
+			result = datumCopy(result, false, -1);
```
rohit01010: why do we need copy ?

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -4923,4 +4926,137 @@ sp_xml_removedocument(PG_FUNCTION_ARGS)
...
+			found = true;
+		}
```
rohit01010: what will be the output when NULL document is passed to openxml ?

contrib/babelfishpg_tsql/src/procedures.c
rohit01010: Also move this constant to the top

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -4923,4 +4926,137 @@ sp_xml_removedocument(PG_FUNCTION_ARGS)
...
+	relation_close(relation, AccessShareLock);
+	
+	/* If we found the document, return it */
+	if (found)
+		PG_RETURN_DATUM(result);
+	
+	/* If we didn't find the handle , throw an error */
+	ereport(ERROR,
+			(errcode(ERRCODE_UNDEFINED_OBJECT),
+			 errmsg("Could not find prepared statement with handle %d", document_id)));
```
rohit01010: Need to update the error message.

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -6689,6 +6696,176 @@ pltsql_exprTypmod(Plan *plan, Node *expr)
...
+			
+			if (!isnull)
+			{
+				attoptions = DatumGetArrayTypeP(datum);
+				colname = get_value_by_name_from_array(attoptions, ATTOPTION_BBF_ORIGINAL_NAME);
+			}
+			else
+			{
+				colname = NameStr(att->attname);
+			}
```
rohit01010, line 7035: Lets test this by creating table from psql side

rohit01010 (changes requested):

contrib/babelfishpg_tsql/src/procedures.c
harshdubey166: Yes renamed that and moved to top.

contrib/babelfishpg_tsql/src/procedures.c
harshdubey166: The datumCopy() is necessary because heap_getattr() returns a pointer to data within the heap tuple's memory. After table_endscan() and relation_close() are called, the original tuple memory becomes invalid and may be freed. Without datumCopy(), the function would return a dangling pointer to freed memory, causing crashes or data corruption when the XML document is accessed later.

contrib/babelfishpg_tsql/src/procedures.c
harshdubey166: Updated that.

contrib/babelfishpg_tsql/src/procedures.c
harshdubey166: If NULL doc is passed then output is NULL. In our case also if no doc is present, then no rows are returned.

contrib/babelfishpg_tsql/src/procedures.c
harshdubey166: Updated that.

contrib/babelfishpg_tsql/src/hooks.c
harshdubey166: Tested and verified that.

test/JDBC/upgrade/17_6/schedule
```diff
@@ -610,6 +610,7 @@ BABEL-USER-GRANT
 test_conv_money_to_varchar
 sp_xml_preparedocument
 sp_xml_removedocument
+openxml_with_clause
```
Yvinayak07: Do we need schedule testing? if yes, then let's also add in other versions of 17.

contrib/babelfishpg_tsql/src/backend_parser/gram-tsql-rule.y
harshdubey166: Yes we can also use that. Made changes.

test/JDBC/upgrade/17_6/schedule
harshdubey166: Added that.

contrib/babelfishpg_tsql/sql/sys_functions.sql
harshdubey166: Added

rohit01010 (approved):

Yvinayak07 (approved):

rohit01010 (approved):

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -6927,6 +6958,242 @@ pltsql_exprTypmod(Plan *plan, Node *expr)
...
+								 RelationGetDescr(attrel), &isnull);
+			
+			if (!isnull)
+			{
+				attoptions = DatumGetArrayTypeP(datum);
+				colname = get_value_by_name_from_array(attoptions, ATTOPTION_BBF_ORIGINAL_NAME);
+			}
+			else
+			{
+				colname = NameStr(att->attname);
```
rohit01010: shouldn't this be as follows ?
```
colname = pstrdup(NameStr(att->attname));
```

rohit01010 (approved):

contrib/babelfishpg_tsql/src/hooks.c
harshdubey166: Updated that.

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -212,6 +212,7 @@ void	   *get_servername_internal(void);
 void	   *get_servicename_internal(void);
 void	   *get_language(void);
 void	   *get_host_id(void);
+void        extract_namespaces_from_xml(xmltype *ns_data, char ***ns_names, char ***ns_uris, int *ns_count);
```
Deepesh125: nit: indentation

Deepesh125: instead we should define this as an extern in some header

test/python/expected/upgrade_validation/expected_dependency.out
```diff
@@ -715,6 +714,8 @@ Function sys.tsql_get_constraintdef(oid)
 Function sys.tsql_get_expr(text,oid)
 Function sys.tsql_get_functiondef(oid)
 Function sys.tsql_get_returntypmodvalue(oid)
+Function sys.tsql_openxml_get_colpattern(text,integer)
+Function sys.tsql_openxml_get_xmldoc(integer)
```
Deepesh125: is it possible to add some basic tests for these functions?

Deepesh125 (dismissed):

test/python/expected/upgrade_validation/expected_dependency.out
harshdubey166: Added them

contrib/babelfishpg_tsql/runtime/functions.c
harshdubey166: We have defined this as an extern in hooks.c file

Deepesh125: we should define extern in header file when sharing declarations across multiple source files.

harshdubey166: Updated that.

Deepesh125 (approved):

PR #3830, opened 2025-06-13
Fixed issue of incorrect splitting of response buffer into multiple packets for PLP Data

rohit01010:
### Description
Currently, in Babelfish for large response of nvarchar(max) type while splitting the response buffer into multiple packets the last PLP chunk is getting splitting directly and the next packet starts with rest of the part of PLP chunk. Instead if the last PLP chunk size cannot be accommodated into available buffer, the PLP chunk size should have been reduced to available buffer and next packet should start with new PLP chunk. This PR will fix this issue.

### Issues Resolved

BABEL-5875

* **Use case based -** YES

* **Arbitrary inputs -** NA

* **Performance tests -** NA

* **Tooling impact -** NA

* **Client tests -** NA

### Check List


contrib/babelfishpg_tds/src/backend/tds/tdscomm.c
```diff
@@ -58,6 +58,19 @@ static void SocketSetNonblocking(bool nonblocking);
...
+GetAvailableBufferSize()
+{
+	return TdsBufferSize - TdsSendCur;
+}
+
+void
+FlushBuffer()
+{
+	SocketSetNonblocking(false);
+	InternalFlush(false);
```
Anikait143: What if `InternalFlush` returns trouble? we should also check and handle that appropriately.

contrib/babelfishpg_tds/src/backend/tds/tdstypeio.c
```diff
@@ -2820,6 +2821,26 @@ TdsSendPlpDataHelper(char *data, int len)
...
+			 * If PLP CHUNK block cannot be stored in available buffer, 
+			 * then set PLP CHUNK LEN to size of available buffer.
+			 */
+			if ((plpChunckLen + sizeof(uint32_t)) > amount)
+			{
+				/*
+				 * If available buffer size cannot accomodate smallest chunk i.e. of size 1 byte
+				 * then flush the buffer and try again, else set PLP CHUNK LEN to size of available buffer.
+				 */
+				if (amount < (1 + sizeof(uint32_t)))
```
Anikait143: Let's have a test case where we end up hitting this condition and verify the behaviour

test/dotnet/input/InsertBulk/insertBulkWithNVarcharCol.txt
Anikait143, line 1: Let's have test case where we insert a row with pretty large data which would exceed PLP_CHUNK_LEN and then verify how packets split in that case with TSQL. Consider these scenarios: 

(Suppose we have a data of size 32001)
1. What if buffer split happens after 32000 bytes data but before the last byte is written?
2. What if buffer split happens before 32000 bytes data is written?

Anikait143 (commented):
General Question: Have we seen this behaviour/issue with every datatype which returns PLP or is it only specific to NVARCHAR(MAX)?

test/dotnet/input/InsertBulk/insertBulkWithNVarcharCol.txt
```diff
@@ -0,0 +1,110 @@
+# NVARCHAR(MAX), VARCHAR(MAX) datatype with huge rows
```
KushaalShroff, line 1: This is not insertbulk test case, why are we categorizing it as one?

contrib/babelfishpg_tds/src/backend/tds/tdstypeio.c
```diff
@@ -2820,6 +2821,26 @@ TdsSendPlpDataHelper(char *data, int len)
 			/* Either data is "0" or no more data to send */
 			if (plpChunckLen == 0)
 				break;
+			
+			amount = GetAvailableBufferSize();
```
KushaalShroff: Please rename amount variable to something more appropriate like availableSpace

contrib/babelfishpg_tds/src/backend/tds/tdstypeio.c
```diff
@@ -2820,6 +2821,26 @@ TdsSendPlpDataHelper(char *data, int len)
...
+			/*
+			 * If PLP CHUNK block cannot be stored in available buffer, 
+			 * then set PLP CHUNK LEN to size of available buffer.
+			 */
+			if ((plpChunckLen + sizeof(uint32_t)) > amount)
```
KushaalShroff: Please justify the + sizeof(uint32_t)) in comments, readers wont understand its len metadata and that we do not want it to cross packet boundaries

contrib/babelfishpg_tds/src/backend/tds/tdstypeio.c
```diff
@@ -2830,6 +2851,7 @@ TdsSendPlpDataHelper(char *data, int len)
 				return rc;
 
 			tempOffset += plpChunckLen;
+			plpChunckLen = PLP_CHUNCK_LEN;
```
KushaalShroff: init plpChunckLen = PLP_CHUNCK_LEN only once at the top of the loop

contrib/babelfishpg_tds/src/backend/tds/tdstypeio.c
KushaalShroff: We can also add more debug logs for PLP chunks and make use of TdsErrorContext->err_text appropriately

contrib/babelfishpg_tds/src/backend/tds/tdstypeio.c
KushaalShroff: ++ to @Anikait143, We would want to test this properly.

contrib/babelfishpg_tds/src/backend/tds/tdscomm.c
KushaalShroff: We arent handling that while calling TdsPutbytes either. It transitively can get EOF response and its being widely used in the codebase without proper error handling -- deducing that EOF here is just defensive coding? @rohit01010 Can you please verify

test/dotnet/input/InsertBulk/insertBulkWithNVarcharCol.txt
KushaalShroff: I am not sure what we mean by buffer split here. I think adding test cases for scenarios where the available space in buffer is between 1-4 bytes with another PLP CHUNK remaining to flush would be appropriate for this fix (Unless we are testing for the PLP feature at length)

KushaalShroff (changes requested):

test/dotnet/input/InsertBulk/insertBulkWithNVarcharCol.txt
rohit01010: For large data whose size exceeds PLP_CHUNK_LEN, lets say its size is data_size. The behaviour would be the data will be distributed into multiple CHUNKS each of size PLP_CHUNK_LEN, and once it reaches the stage where the available buffer size is less than PLP_CHUNK_LEN then a chunk of what size that can be fit in available buffer size will be used, and from next packet the remaining data will again gets split to chunks of size PLP_CHUNK_LEN, and so on, until it reaches where the remaining size of data is less than PLP_CHUNK_LEN, then whatever size of data is remaining that size of chunk will be used.

@Anikait143 This above explanation should answer your query.

test/dotnet/input/InsertBulk/insertBulkWithNVarcharCol.txt
rohit01010: The specific kind of scenario where we were hitting the issue is captured during the insertbulk testing in dotnet, where we are reading the response of "select * from sourceTable" using SqlClient.SqlDataReader. Without the fix from this PR, this test added will also fail with same error. Hence added insertBulk test for the same.

contrib/babelfishpg_tds/src/backend/tds/tdscomm.c
rohit01010: There are multiple functions which underlying uses InternalFlush such as `TdsPutbytes, TdsPutInt8, TdsPutInt64LE` etc, at some of the places EOF returned by this function is being returned further to the parent function, and at some of the places we do not return. For this case even if it is a defensive coding, added similar check and returned EOF.

contrib/babelfishpg_tds/src/backend/tds/tdstypeio.c
rohit01010: Added test case which hits the above mentioned scenario. And we are matching the behaviour of TSQL in that case as well.

contrib/babelfishpg_tds/src/backend/tds/tdstypeio.c
rohit01010: > Please rename amount variable to something more appropriate like availableSpace

Renamed.

rohit01010: > We can also add more debug logs for PLP chunks and make use of TdsErrorContext->err_text appropriately

What kind of debug logs we want to print here ?

contrib/babelfishpg_tds/src/backend/tds/tdstypeio.c
rohit01010: Added the same in the comment.

contrib/babelfishpg_tds/src/backend/tds/tdstypeio.c
rohit01010: Updated

rohit01010:
> Have we seen this behaviour/issue with every datatype which returns PLP or is it only specific to NVARCHAR(MAX)?

Yes similar behaviour difference in response packets were observed in other datatypes which returns PLP, Added tests for other datatypes which returns PLP.

Anikait143 (approved):

contrib/babelfishpg_tds/src/backend/tds/tdstypeio.c
rohit01010: Added Debug logs.

Anikait143 (approved):

Anikait143 (approved):

KushaalShroff (approved):

PR #3840, opened 2025-06-17
Do not coerce float/real datatypes in pre_transform_hook

KushaalShroff:
**ISSUE:** When lower precedence datatypes (to real/float) are grouped on one side and then union'ed with real or float datatypes then "not equal" values were being grouped together. This was found during real and float testing when "not equal" real values were being grouped together and the root cause helped us extrapolate issues with numeric/int/smallint/tinyint as well.
**NOTE:** In Fact, for Numeric unioned REAL/FLOAT we were also observing crashes due to the early coercion and precision mismatch during aggregation.

**Root Cause**: The coerce_to_target_type during pre_transform_hook is applying the implicit cast too early -- even before grouping.
When the REAL(or other lower precedence type) values are cast to FLOAT(Or Higher precedence types) during the coercion, there might be precision differences. The cast types in PG from lower to high precedence types are defined as "IMPLICIT" which gives the freedom to optimizer to evaluate on the go. In case of "ASSIGNED" cast types optimizer would have grouped first and then applied cast -- it prevents the automatic type conversion during the grouping phase -- which would be the expected outcome.

**Solution:** Prevent the cast from happening too early for REAL/FLOAT, thus, maintaining the original precision for grouping and only cast for the final result.

### Check List


contrib/babelfishpg_tsql/src/tsql_analyze.c
```diff
@@ -413,7 +413,18 @@ fix_setop_typmods(ParseState *pstate, Query *qry)
...
+
+		/*
+		 * If there was an aggregate in one of the unions then the grouping is
+		 * required to happen before coercing to common type. For most of the common
+		 * types it is OK to coerce first since the type representation doesn't
+		 * change but for float types we need to avoid losing precision for aggregation.
+		 */
+		if (common_type == FLOAT8OID || common_type == FLOAT4OID)
+		{
+				topColCollations = lappend_oid(topColCollations, -1);
```
Deepesh125: It should be InvalidOid for FLOAT data types. 

Can you also fix typo in the comment on the line 348?

Deepesh125 (approved):

PR #3855, opened 2025-06-19
fix for char and nchar function correct handling

pranavJ23:
### Description

Function char throws error if the input integer is not in range 0-255, where as tsql returns NULL instead of error.
Function nchar should throw NULL if the the input integer is not in range 0-65535 
Fix for nchar() function so that if input type is coercible to int then it should execute successfully without throwing error

### Issues Resolved

[BABEL-4810] [BABEL-4812][BABEL-5933]

* **Use case based -**

* **Boundary conditions -**

* **Arbitrary inputs -**

* **Negative test cases -**

* **Performance tests -**

* **Tooling impact -**

* **Client tests -**

### Check List


contrib/babelfishpg_common/sql/upgrades/babelfish_common_helper--5.2.0--5.3.0.sql
```diff
@@ -145,6 +145,50 @@ BEGIN
     END IF;
 END $$;
 
+create or replace function sys.cht(x in int) returns sys.varchar
```
Anikait143: Old function should be deprecated since we have changed function signature

contrib/babelfishpg_common/sql/string_operators.sql
```diff
@@ -111,14 +111,14 @@ EXTENSION PACK function CHAR(x)
     end if;
 END;
 $body$
-language plpgsql STABLE;
+LANGUAGE plpgsql IMMUTABLE STRICT PARALLEL SAFE;
```
Anikait143, line 117: We should mark all the functions as `STABLE` since the underlying helper function depends on `encoding` settings

Anikait143, line 117: Have we verified that it is really `PARALLEL SAFE`?

contrib/babelfishpg_common/sql/string_operators.sql
```diff
@@ -97,7 +97,7 @@ CREATE OPERATOR sys.+ (
     FUNCTION = sys.babelfish_concat_wrapper
 );
 
-create or replace function sys.CHAR(x in int)returns char
+create or replace function sys.cht(x in int) returns sys.varchar
```
Anikait143: Return type should be `sys.bpchar`

contrib/babelfishpg_common/sql/string_operators.sql
```diff
@@ -97,7 +97,7 @@ CREATE OPERATOR sys.+ (
...
 AS
 $body$
 BEGIN
```
Anikait143, line 103: What if input is `0`? (Same for other functions as well)

test/JDBC/expected/ascii.out
```diff
@@ -127,9 +127,10 @@ int#!#int#!#int#!#int
 -- following throws error in babelfish
 SELECT ASCII(CHAR(0)) AS NullChar
 GO
-~~ERROR (Code: 33557097)~~
-
-~~ERROR (Message: null character not permitted)~~
+~~START~~
+int
+<NULL>
+~~END~~
```
Anikait143: I think this not the correct expectation, shouldn't it be `0`?

contrib/babelfishpg_tsql/src/tsqlIface.cpp
```diff
@@ -3037,12 +3037,14 @@ class tsqlMutator : public TSqlParserBaseListener
...
-	    else
-		stream.setText(ctx->start->getStartIndex(), " chr");
+		int startIndex = proc->start->getStartIndex();
+
+		if (proc->DOUBLE_QUOTE_ID())
+			stream.setText(startIndex, "\"cht\" ");
+		else if (proc->SQUARE_BRACKET_ID())
+			stream.setText(startIndex, "[cht] ");		
+		else
+			stream.setText(startIndex, " cht");
```
Anikait143: Would there be any case where `DOUBLE_QUOTE_ID` and `SQUARE_BRACKET_ID` are valid for function calls?

contrib/babelfishpg_tsql/src/tsqlIface.cpp
```diff
@@ -3181,12 +3183,14 @@ class tsqlMutator : public TSqlParserBaseListener
...
-	    else
-		stream.setText(ctx->start->getStartIndex(), " chr");
+		int startIndex = proc->start->getStartIndex();
+
+		if (proc->DOUBLE_QUOTE_ID())
+		        stream.setText(startIndex, "\"cht\" ");
+		else if (proc->SQUARE_BRACKET_ID())
+		        stream.setText(startIndex, "[cht] ");		
+		else
+		        stream.setText(startIndex, " cht");
```
Anikait143: It's better to have proper text replacement maybe use underscore instead of space for new function name.

contrib/babelfishpg_common/sql/string_operators.sql
```diff
@@ -111,14 +111,14 @@ EXTENSION PACK function CHAR(x)
...
 END;
 $body$
-language plpgsql STABLE;
+LANGUAGE plpgsql IMMUTABLE STRICT PARALLEL SAFE;
 
 CREATE OR REPLACE FUNCTION sys.nchar(IN x INTEGER) RETURNS sys.nvarchar
 AS
 $body$
 BEGIN
     --- 1114111 is 0x10FFFF - max value permitted as specified by documentation
```
Anikait143: Let's also correct the comment here

contrib/babelfishpg_common/sql/string_operators.sql
```diff
@@ -132,7 +132,7 @@ AS
...
-    if x::integer between 1 and 1114111 then
+    if x::integer between 1 and 65535 then
         return(select chr(x::integer))::sys.nvarchar;
     else
         return null;
```
Anikait143: Why do we need a separate function for `varbinary` input?

contrib/babelfishpg_common/sql/string_operators.sql
```diff
@@ -111,14 +111,14 @@ EXTENSION PACK function CHAR(x)
     end if;
 END;
 $body$
-language plpgsql STABLE;
+LANGUAGE plpgsql IMMUTABLE STRICT PARALLEL SAFE;
 
 CREATE OR REPLACE FUNCTION sys.nchar(IN x INTEGER) RETURNS sys.nvarchar
```
Anikait143: Return type should be `nchar`

Anikait143 (changes requested):
Please add following test case:
1. Tests with schema.function_call
2. Test with different collation settings
3. Test for multibyte character input integer value.
4. Test with edge case values
5. Test with datatype inputs that can be implicitly casted to `int`

contrib/babelfishpg_common/sql/string_operators.sql
pranavJ23: yes, this edge case need to be handled since pg function 'chr' expects the input b/w 1-255 where as the tsql expects input b/w 0-255. Thanks

contrib/babelfishpg_common/sql/string_operators.sql
pranavJ23: IMO nvarchar might be a better return function since nchar using fixed length whereas the output will be of length = 1 always, so it might allocate extra space in case of nchar. Else anything can be used, not an issue.

contrib/babelfishpg_common/sql/string_operators.sql
pranavJ23: Since the test case is passing in the JDBC parallel query mode, so we can say that it's parallel safe ?

pranavJ23: with stable we are not able to use the CHAR function in computed column, where as in tsql the char can be used in computed column.
Since the encoding in babelfish_db can' t be changed once the db is formed, hence we can mark the function as immutable

contrib/babelfishpg_common/sql/string_operators.sql
pranavJ23: IMO varchar might be a better return function since bpchar using fixed length whereas the output will be of length = 1 always, so it might allocate extra space in case of bpchar. Else anything can be used, not an issue.

contrib/babelfishpg_common/sql/string_operators.sql
pranavJ23: For nchar : it's not a built in dataype but if we do not define the direct match sql function for nchar then it will consider nchar as datatype and hence we will be getting the error as
```The string size for the given CHAR/NCHAR data is not defined. Please use an explicit CAST or CONVERT to CHAR(n)/NCHAR(n)```
So we need to define the explicit varbinary function for nchar too

File: contrib/babelfishpg_common/sql/string_operators.sql (hunk 8 above)
Comment (tanscorpio7) [reply contrib/babelfishpg_common/sql/string_operators.sql]: Should this function be dependant upon the input collation ?

**Review by tanscorpio7 — CHANGES_REQUESTED** [review]

File: contrib/babelfishpg_common/sql/upgrades/babelfish_common_helper--5.2.0--5.3.0.sql [hunk 11]
```diff
@@ -145,6 +145,75 @@ BEGIN
     END IF;
 END $$;
 
+DO $$
+DECLARE
+    exception_message text;
+BEGIN
+    ALTER FUNCTION sys.char(x in int) RENAME TO chr_deprecated_5_2_0;
```
Anikait143: Deprecated version 5.3.0, please correct the name everywhere.

contrib/babelfishpg_common/sql/upgrades/babelfish_common_helper--5.2.0--5.3.0.sql
```diff
@@ -145,6 +145,75 @@ BEGIN
...
+    exception_message = MESSAGE_TEXT;
+    RAISE WARNING '%', exception_message;
+END;
+$$;
+
+DO $$
+DECLARE
+    exception_message text;
+BEGIN
+   ALTER FUNCTION sys.nchar(In x INTEGER) RENAME TO nchar_int_deprecated_5_2_0;
```
Anikait143: We should also drop the deprecated functions. Refer: https://github.com/babelfish-for-postgresql/babelfish_extensions/blob/e6e094c8dc226e983ca03eeebb13ab4ad117ca73/contrib/babelfishpg_common/sql/upgrades/spatial_types--5.2.0--5.3.0.sql#L55

contrib/babelfishpg_tsql/src/tsqlIface.cpp
pranavJ23: Yes we do support the query like :
```
select [char](255)
```
adding tests for it

contrib/babelfishpg_common/sql/string_operators.sql
pranavJ23: From the pg_proc catalog too, we have verified that chr() function is parallel safe and since we are using chr() underlying sys.char(), so we can say it's parallel safe

test/JDBC/expected/ascii.out
pranavJ23: yes, it should be 0, addign a condition to handle the case when input is zero

contrib/babelfishpg_tsql/src/tsqlIface.cpp
```diff
@@ -3176,17 +3204,44 @@ class tsqlMutator : public TSqlParserBaseListener
...
+
+		if(schema && !schNameStr.empty() && pg_strcasecmp(schNameStr.c_str(), "sys") != 0)
+			return;
+
+		if (proc->DOUBLE_QUOTE_ID())
+			stream.setText(startIndex, "\"ncht_\"");
+		else if (proc->SQUARE_BRACKET_ID())
+			stream.setText(startIndex, "[ncht_]");		
+		else
+			stream.setText(startIndex, "ncht_");
```
tanscorpio7: Looks like this block is duplicated 4 times. Can we create one helper function for this ?

tanscorpio7 (changes requested):

Anikait143 (approved):

tanscorpio7 (approved):

shardgupta (approved):

PR #3866, opened 2025-06-21
Fixed error while using PosgreSQL reserved keyword in cursor operations

shardgupta:
### Description
During the cursor prepare, execution flow for statement is not passed through antlr parser. This causes issue when query contains PosgreSQL reserved keywords. This PR addresses this issue, along with any potential issue with query which is not compatible with PostgreSQL

Task: BABEL-5743

Signed-off-by: Shard Gupta <shardga@amazon.com>

### DIFF
--- a/contrib/babelfishpg_tsql/src/cursor.c
+++ b/contrib/babelfishpg_tsql/src/cursor.c
@@ -1452,7 +1452,7 @@ execute_sp_cursoropen_common(int *stmt_handle, int *cursor_handle, const char *s
 	PLtsql_stmt_execsql *parse_result;
 	PLtsql_function *func;
 	char *stmt_copy;
-	const char *tsql_stmt = stmt; /* Use this for potentially modified statement */
+	char *tsql_stmt = NULL; /* Use this for potentially modified statement */
 
 	/*
 	 * Connect to SPI manager. should be handled in the same way with
@@ -1493,7 +1493,7 @@ execute_sp_cursoropen_common(int *stmt_handle, int *cursor_handle, const char *s
 			 * and PLTSQL_STMT_RETURN being last. In case of empty statement only PLTSQL_STMT_RETURN node will be present.
 			 * So total number of nodes should be 3 for valid cursor execution. Actual query statement will be at second 
 			 * position of type PLTSQL_STMT_EXECSQL.
-			 
+			 *
 			 * This is defensive code, where we only assign the tsql_stmt variable to parsed query,
 			 * if the cmd_type is PLTSQL_STMT_EXECSQL. There might be other types of cmd_type like
 			 * PLTSQL_STMT_EXECSQL (for procedures), for them we will keep the old behavior.
@@ -1513,14 +1513,12 @@ execute_sp_cursoropen_common(int *stmt_handle, int *cursor_handle, const char *s
 		}
 
 		/* prepare plan and insert a cursor entry */
-		plan = SPI_prepare_cursor(tsql_stmt, nBindParams, boundParamsOidList, cursor_options);
+		plan = SPI_prepare_cursor(tsql_stmt?tsql_stmt:stmt, nBindParams, boundParamsOidList, cursor_options);
 		if (plan == NULL)
-		{
-			/* Free memory if tsql_stmt is different from stmt */
-			if (tsql_stmt != stmt)
-				pfree((void *)tsql_stmt);
 			return 1;			/* procedure failed */
-		}
+		/* Free memory if tsql_stmt is not null */
+		if (tsql_stmt)
+			pfree((void *)tsql_stmt);
 
 		if (save_plan)
 		{
@@ -1535,10 +1533,6 @@ execute_sp_cursoropen_common(int *stmt_handle, int *cursor_handle, const char *s
 
 			SPI_keepplan(plan);
 		}
-
-		/* Free memory if tsql_stmt is different from stmt */
-		if (tsql_stmt != stmt)
-			pfree((void *)tsql_stmt);
 	}
 	else						/* !prepare */
 	{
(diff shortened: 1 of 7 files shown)


contrib/babelfishpg_tsql/src/cursor.c
```diff
@@ -1449,6 +1449,20 @@ execute_sp_cursoropen_common(int *stmt_handle, int *cursor_handle, const char *s
...
+	char *stmt_copy;
+
+	/*
+	 * Parse the function's text
+	 */
+	if(stmt)
+	{
+		stmt_copy = pstrdup(stmt);
+		function = pltsql_compile_inline(stmt_copy, NULL);
+		parse_result = (PLtsql_stmt_execsql *) lsecond(function->action->body);
```
tanscorpio7: Let us only do this if this block contains a single statement and the statement type is `PLtsql_stmt_execsql`. Otherwise we keep the existing behaviour.

contrib/babelfishpg_tsql/src/cursor.c
```diff
@@ -1449,6 +1449,20 @@ execute_sp_cursoropen_common(int *stmt_handle, int *cursor_handle, const char *s
...
 	MemoryContext oldcontext;
 	MemoryContext savedPortalCxt;
+	PLtsql_stmt_execsql *parse_result;
+	PLtsql_function *function;
+	char *stmt_copy;
+
+	/*
+	 * Parse the function's text
+	 */
+	if(stmt)
```
tanscorpio7: 1. We should only do this when prepare = true.
2. Merge this block with the prepare if block already below.

tanscorpio7 (changes requested):

tanscorpio7 (approved):

contrib/babelfishpg_tsql/src/cursor.c
```diff
@@ -1468,6 +1471,42 @@ execute_sp_cursoropen_common(int *stmt_handle, int *cursor_handle, const char *s
 
 	if (prepare)
 	{
+		/*
+		 * This entire block is to parse the statement by antlr and use the resultant
+		 * statement to sent to Postgres cursor execution. This is necessary in some use case,
```
KushaalShroff: Nit: use the statement for cursor execution.

contrib/babelfishpg_tsql/src/cursor.c
```diff
@@ -1468,6 +1471,42 @@ execute_sp_cursoropen_common(int *stmt_handle, int *cursor_handle, const char *s
...
+		 * for example when we have PostgreSQL reversed keywords in query which is valid in TSQL.
+		 * Antlr parser will add the quotes at necessary places in query so that Postgres engine
+		 * can resolve this query correctly.
+		 */
+		if (stmt)
```
KushaalShroff, line 1482: Is it possible for white spaced empty stmt to reach here?
If yes, does pltsql_compile_inline handle that safely? Maybe we can add negative test case for this

contrib/babelfishpg_tsql/src/cursor.c
```diff
@@ -1468,6 +1471,42 @@ execute_sp_cursoropen_common(int *stmt_handle, int *cursor_handle, const char *s
...
+			/*
+			 * Check the node list of type PLtsql_stmt_type. Cursor only support single statement.
+			 * If there are more than 1 statement we will through the error. func->action-body
+			 * returned by pltsql_compile_inline contains two default nodes, PLTSQL_STMT_INIT being first
+			 * and PLTSQL_STMT_RETURN being last. So total number of nodes should be 3 for cursor. Actual
+			 * query statement will be at second position of type PLTSQL_STMT_EXECSQL.
+
+			 * This is defensive code, where we only reassign the stmt variable to parsed query,
+			 * if the cmd_type is PLTSQL_STMT_EXECSQL. There might be other types of cmd_type like
+			 * PLTSQL_STMT_EXECSQL (for procedures), for them we will keep the old behavior.
```
KushaalShroff, line 1499: Is there a case where other stmt types need ANTLR parsing? If yes, Do we plan to fix this in future?
(Might me better to add FIXME if we are planning for it)

Update: I see the test case below for exec procedure -- is that tsql behaviour? Then should we preempt here and throw error in else condition?

contrib/babelfishpg_tsql/src/cursor.c
```diff
@@ -1468,6 +1471,42 @@ execute_sp_cursoropen_common(int *stmt_handle, int *cursor_handle, const char *s
...
+			 */
+			if((( (PLtsql_stmt *) lsecond(func->action->body))->cmd_type == PLTSQL_STMT_EXECSQL)
+				&& list_length(func->action->body) == 3)
+			{
+				parse_result = (PLtsql_stmt_execsql *) lsecond(func->action->body);
+				stmt = pstrdup(parse_result->sqlstmt->query);
```
KushaalShroff: 1. stmt is declared as a const char *. It's best practice not to reassign a const pointer. We can either think of changing the const declaration or by having another local tsql_stmt for this case.

2. When overriding a pointer that refers to palloc'ed memory, we should ensure the original memory is freed before assignment to prevent memory leaks. Here since it is const, I assume we might get error while freeing it. If we introduce a new variable for the rewritten statement, this problem is avoided.

KushaalShroff (changes requested):

contrib/babelfishpg_tsql/src/cursor.c
shardgupta: Thanks for catching it. Empty statement is handled correctly by pltsql_compile_inline but it would crash during referencing the second list node. Fixed and added test for that.

contrib/babelfishpg_tsql/src/cursor.c
shardgupta: TSQL can handle procedure with single select statement as cursor. But it is not being handled in Babelfish. I have put a comment as fixme to revisit it in future along with other statement types.

KushaalShroff (approved):

PR #3882, opened 2025-06-26
Fixed error while using PosgreSQL reserved keyword in cursor operations

shardgupta:
During the cursor prepare, execution flow for statement is not passed through antlr parser. This causes issue when query contains PosgreSQL reserved keywords. This PR addresses this issue, along with any potential issue with query which is not compatible with PostgreSQL

Task: BABEL-5743

Signed-off-by: Shard Gupta shardga@amazon.com


jsudrik (approved):
Cherrypick of the identical changes in 5x branch. Approved.

PR #3918, opened 2025-07-03
Rectified sp_helplogins stored proc for certain cases

ayushdsh:
Currently, there are certain issues with sp_helplogins -

1. db_owner is not treated as a special role with special priv leading  to incorrect results
2. Using INTs instead of BIGINTs as the intermediate value for SIDs since we are converting it to binary and size of the intermediate type matters
3. Filter for the second result was missing from SQL for when loginname is not null.

And removed the branching to merge the if..else

### Check List


contrib/babelfishpg_tsql/sql/babelfishpg_tsql.sql
```diff
@@ -3760,23 +3760,28 @@ LANGUAGE pltsql
...
 DECLARE @is_sysadmin BIT
+DECLARE @is_securityadmin BIT
 BEGIN
 
-    IF is_srvrolemember('securityadmin') = 0 
+    SET @current_userid = suser_id();
+    SET @is_sysadmin = pg_has_role(@current_userid, cast('sysadmin' as text), 'MEMBER');
+    SET @is_securityadmin = pg_has_role(@current_userid, cast('securityadmin' as text), 'MEMBER');
+
+    IF @is_securityadmin = 0 AND @is_sysadmin = 0
```
ayushdsh: This was changed because currently, `is_srvrolemember('securityadmin')` is impl like below

```
has_role = (pg_has_role(login::TEXT, role::TEXT, 'MEMBER')
				OR ((login COLLATE sys.database_default NOT IN ('sysadmin', 'securityadmin', 'dbcreator'))
					AND pg_has_role(login::TEXT, 'sysadmin'::TEXT, 'MEMBER')));
```
Here the problem with the function call `pg_has_role` is we are passing the login name instead of oid. This leads to an error in PG like below - 

```
LOG:  Unmapped error found. Code: 67137668, Message: role "gad\new_user" does not exist, File: acl.c
ERROR:  role "gad\new_user" does not exist
```

This happens because of an incorrect (maybe intentional) mapping of window logins in pg_roles. The correction here is to use oid of the login instead of the name

```
pg_has_role(suser_id(), 'securityadmin'::TEXT, 'MEMBER')
```

The problem is we are passing a login and `pg_has_role(suser_id(login)...` does not also behave correctly because of the same reason above.

Hence, I've used the internal impl of `is_srvrolemember` and replace `login` with `oid` using `suser_id()`


Evidence - 

```
--- /local/home/ayushdsh/MergeMonkeyScratch_Rel17-X-Dev/src/ManfredBabelIntegrationTests/adtest/output/login_token.out
+++ /local/home/ayushdsh/MergeMonkeyScratch_Rel17-X-Dev/src/ManfredBabelIntegrationTests/adtest/expected/login_token.out
@@ -55,6 +55,28 @@
 
 -- Case 1.2: AD user has a login, login is a member of sysadmin
 -- tsql aduser=User adpassword=Babel@123
+select pg_has_role(suser_id(), cast('sysadmin' as text), 'MEMBER') as is_sysadmin;
+~~START~~
+is_sysadmin
+-----------
+1
+
+(1 rows affected)
+
+~~END~~
+
+
+select is_srvrolemember('sysadmin') as is_sysadmin;
+~~START~~
+is_sysadmin
+-----------
+0
+
+(1 rows affected)
+
+~~END~~
+
+
```
Below you can see that the AD user 'User' has sysadmin permissions, however, `is_srvrolemember()` returns 0

contrib/babelfishpg_tsql/sql/babelfishpg_tsql.sql
```diff
@@ -3816,9 +3821,19 @@ BEGIN
...
-                    AND UExt2.login_name = @current_username
+                    AND (
+                        (UExt2.type IN ('S', 'R') AND UExt2.login_name = @current_username) OR
+                        -- Babelfish stores login information differently depending on the type of login:
+                        -- - For SQL Server logins (type = 'S') and roles (type = 'R'), the `login_name` in UExt4
+                        --   directly matches the current username (`@current_username`).
+                        -- - For Windows logins (type = 'U'), Babelfish maps the Windows account to an internal role 
+                        --   and stores the original Windows login in `LExt.orig_loginname` instead of `UExt4.login_name`.
+                        --   eg. for login gad\user it is mapped to user@GAD.COM in bbf_user catalog
+                        (UExt2.type = 'U' AND LExt.orig_loginname = @current_username)
```
ayushdsh: This was changed to account for group and individual AD users. More details in the comments

contrib/babelfishpg_tsql/sql/babelfishpg_tsql.sql
anju15bharti: From a GAD perspective, the query
`select pg_has_role(suser_id(), cast('sysadmin' as text), 'MEMBER') as is_sysadmin;
`
works correctly when the individual login exists and is a member of the sysadmin role. However, it fails if the individual login does not exist but the group login does and the group is a member of sysadmin.

To handle this properly under GAD, we should use `session_user` instead of `suser_id()`.

These GAD-related changes should be applied internally only, as GAD is not supported in the open-source version.

contrib/babelfishpg_tsql/sql/babelfishpg_tsql.sql
```diff
@@ -3760,23 +3760,28 @@ LANGUAGE pltsql
...
+    SET @is_sysadmin = pg_has_role(@current_userid, cast('sysadmin' as text), 'MEMBER');
+    SET @is_securityadmin = pg_has_role(@current_userid, cast('securityadmin' as text), 'MEMBER');
+
+    IF @is_securityadmin = 0 AND @is_sysadmin = 0 
     BEGIN
         RAISERROR('User does not have permission to perform this action.', 16, 1);
 		RETURN 0;
     END
 
     SET @current_username = LOWER(sys.suser_name());
```
anju15bharti: The value of `@current_username` for GAD purposes should be derived from session_user, which is an internal-specific change.
Please ensure that GAD-related scenarios are handled only within the internal environment and not in OSS.
This PR should include only the non-GAD session changes.

anju15bharti (changes requested):

contrib/babelfishpg_tsql/sql/babelfishpg_tsql.sql
HarshLunagariya: If we can come up with something which works with both correctly will be great as well.

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--5.2.0--5.3.0.sql
```diff
@@ -475,7 +475,24 @@ BEGIN
         WHERE UExt.type != 'R' AND  
             UExt.orig_username != 'guest' AND 
             has_dbaccess(UExt.database_name) = 1 AND
-            LExt.orig_loginname = @input_loginname
+            LExt.orig_loginname = @input_loginname AND
+            (
+                @is_sysadmin = 1 OR
+                LExt.orig_loginname = @current_username OR
+                ISNULL(UExt.login_name, '') = '' OR
```
HarshLunagariya: Why is this condition needed?

contrib/babelfishpg_tsql/sql/babelfishpg_tsql.sql
```diff
@@ -3776,7 +3776,7 @@ BEGIN
     BEGIN    
         SELECT DISTINCT
             CAST(LExt.orig_loginname AS sys.SYSNAME) AS LoginName,
-            CAST(CAST(Base.oid AS BIGINT) AS sys.varbinary(85)) AS SID,
+            CAST(CAST(Base.oid AS INT) AS sys.varbinary(85)) AS SID,
             CAST(LExt.default_database_name AS SYS.SYSNAME) AS DefDBName,
```
HarshLunagariya: Why are we not keeping it BIGINT only?

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--5.2.0--5.3.0.sql
ayushdsh: This is the same condition we've used above in the non-parameter version of the proc. These conditions just control what should be visible to the user based on their priv .. this was a miss in the previous PR and should've been present in the initial PR itself

contrib/babelfishpg_tsql/sql/babelfishpg_tsql.sql
ayushdsh: I think using session_user in our internal codebase will be simpler rather than the above conditions (check securityadmin and sysadmin). Also perf wise its a little better

contrib/babelfishpg_tsql/sql/babelfishpg_tsql.sql
```diff
@@ -3889,7 +3889,24 @@ BEGIN
...
+                EXISTS (
+                    SELECT 1 
+                    FROM pg_catalog.pg_auth_members AS Authmbr
+                    INNER JOIN pg_catalog.pg_roles AS PGR1 ON PGR1.oid = Authmbr.roleid
+                    INNER JOIN pg_catalog.pg_roles AS PGR2 ON PGR2.oid = Authmbr.member
+                    INNER JOIN sys.babelfish_authid_user_ext AS UExt1 ON PGR1.rolname = UExt1.rolname
+                    INNER JOIN sys.babelfish_authid_user_ext AS UExt2 ON PGR2.rolname = UExt2.rolname
+                    WHERE UExt1.orig_username IN ('db_owner', 'db_securityadmin', 'db_accessadmin') 
+                    AND UExt2.database_name = UExt.database_name -- filter to check if the processing db is equal to the outer query db, since we want to find if the user is a member of the roles in the outer db
+                    AND UExt2.login_name = @current_username
```
anju15bharti: Can't we combine the filters for both the if and else cases, since they are the same? Only the value against which the privilege is checked differs.

test/JDBC/expected/z_sp_helplogins-vu-verify.out
```diff
@@ -207,6 +220,62 @@ testloginwithsecurityadmin#!#sp_helplogins_db1#!#userof_testloginwithsecurityadm
...
+
+
+-- HOWEVER if we explicitly query for other logins with users, 
+-- the corresponding users should not be visible
+-- ignore_columns 2
+EXEC sp_helplogins 'sp_helplogins_testlogin'
+GO
+~~START~~
+varchar#!#varchar#!#varchar#!#char#!#char
+sp_helplogins_testlogin#!#master#!#English#!#yes  #!#no
```
anju15bharti, line 248: the value for `Auser` should be no in the cases when user is not able to see the corresponding users.

test/JDBC/expected/z_sp_helplogins-vu-verify.out
```diff
@@ -53,6 +53,12 @@ GO
 alter login testloginindb1 with password = '12345678'
 GO
 
+alter login sp_helplogin_loginwithusermemberofdbowner with password = '12345678'
```
anju15bharti, line 56: Let's also add testcases it shows SID displayed is correct.

anju15bharti (changes requested):

test/JDBC/expected/z_sp_helplogins-vu-verify.out
ayushdsh: discussed offline on how this change will have to be a change in the test framework

test/JDBC/expected/z_sp_helplogins-vu-verify.out
ayushdsh: Created a issue for tracking this - BABEL-5985

contrib/babelfishpg_tsql/sql/babelfishpg_tsql.sql
ayushdsh: If we want to combine both, we need to have an additional filter like - 
`@loginname IS NOT NULL AND sys.RTRIM(@loginname) = @current_username`.

This is not much, but it adds a little overhead of RTRIM everytime the filter executed. I measured a difference of 50-70ms overhead with this.

anju15bharti: What if we keep setting `input_loginname` as is and integrate the rest of the logic around it?

HarshLunagariya (approved):

HarshLunagariya (changes requested):
Why is latest-latest mvu test is failing? Overall changes looks fine to me.

ayushdsh:
the mVU test failure is expected because its a upgrade from 17.6 (yak) itself .. so since there's no upgrade path the verify runs on previous version of proc.

anju15bharti (approved):
Update the PR title and remove the mention of GAD, as it is not addressed in this PR

jsudrik (approved):
Approved.

PR #3972, opened 2025-07-17
Introduce additional failure mechanism for broken views as safety check

R4hul04:
### Description

This change enhances the broken view handling mechanism by adding an additional layer of protection. While our primary view repair mechanism should catch and fail when attempting to repair broken views (with errors like "relation does not exist"), this change adds a failsafe mechanism to ensure broken views always fail explicitly rather than potentially returning NULL values. The expected scenario is that this view function will never ever be reached during execution but if due to some uncertainty it does then this will guarantee error instead of incorrect results.

This PR includes following changes:
1. Added new system function `sys.babelfish_broken_view_function()` in `sys_functions.sql` that raises an explicit error
2. Updated `create_dummy_view_query_for_broken_view()` to include a WHERE clause that triggers this function
3. Updated `is_dummy_view()` to recognize dummy view should have above WHERE func() = 1 in it's query

### Issues Resolved

BABEL - 5974

Signed-off: Rahul Parande rparande@amazon.com
* **Use case based -**
```
1> create table t(id int, id1 int)
2> go
1> create view v1 as select id from t
2> go
1> drop table t
2> go
1> select * from v1
2> go
Cannot reference a broken view in query
```

```
1> create function f1() returns table as return (select * from t);
2> go
1> create view test_f as select * from f1()
2> go
1> drop function f1()
2> go
1> select * from test_f
2> go
Cannot reference a broken view in query
```

* **Boundary conditions -**

* **Arbitrary inputs -**

* **Negative test cases -**

* **Performance tests -**

* **Tooling impact -**

* **Client tests -**

### Check List


contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -7012,6 +7016,41 @@ is_dummy_view(Oid viewOid)
...
+						is_dummy = true;
+					else
+						is_dummy = false;
+				}
+				else
+					is_dummy = false;
+			}
+			else
+				is_dummy = false;
+		}
```
tanscorpio7: I think we can skip this check during rebuild. The original check for `is_dummy` is already very tight.

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -7285,6 +7331,55 @@ create_dummy_view_query_for_broken_view(Oid viewOid)
...
+		constExpr = makeConst(INT4OID,
+							  -1,
+							  InvalidOid,
+							  sizeof(int32),
+							  Int32GetDatum(1),
+							  false,
+							  true);
+							
+		opname = list_make1(makeString("="));
+		opoid = LookupOperName(NULL, opname, INT4OID, INT4OID, true, -1);
```
tanscorpio7: noError should be false ?

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -7285,6 +7331,55 @@ create_dummy_view_query_for_broken_view(Oid viewOid)
 		targetList = lappend(targetList, te);
 	}
 	dummyQuery->targetList = targetList;
+
+	/* This WHERE clause serves as an additional safety mechanism to ensure
+	 * broken views fail explicitly even if user somehow bypass the primary
+	 * view repair mechanism. While the view repair process should typically
+	 * fail earlier, this acts as extra check against accessing broken views. 
+	 */
+
```
tanscorpio7, line 7300: Please test in your local build that this works as expected. Temporariliy disable the rebuild logic and try selecting from a broken view.

tanscorpio7 (changes requested):

contrib/babelfishpg_tsql/src/hooks.c
R4hul04: Should we not look for our specific function `babelfish_broken_view_function` in the query? This will guarantee us that this dummy view was made by us? Because it might be possible user would also have created dummy view, which will go through extra steps of trying to repair the user side dummy view

R4hul04: Ah, the user dummy query won't go inside repair logic as it isn't marked as broken 👍

contrib/babelfishpg_tsql/src/hooks.c
R4hul04: Added to PR description, couple of tests.

contrib/babelfishpg_tsql/src/hooks.c
R4hul04: I don't think it will matter, since it's an impossible case. Still changed the value to false

forestkeeper (approved):

PR #3987, opened 2025-07-24
Solving Crash in FOR JSON AUTO/PATH

herambh-shah:
### Description

Bug Description: A crash has been identified in the FOR JSON AUTO implementation, specifically in the pl_handler.c file. The crash occurs within the strcmp() function called from the buildJsonEntry() function when processing column names.

Root Cause:

The code fails to check if entries in the target list are marked as "resjunk" (which should be skipped during processing).
There's no validation to ensure column names are not NULL before processing.
### Issues Resolved

BABEL-5910

* **Use case based -**
* **Boundary conditions -**
* **Arbitrary inputs -**
* **Negative test cases -**

### Check List

### DIFF
--- a/contrib/babelfishpg_tsql/src/pl_handler.c
+++ b/contrib/babelfishpg_tsql/src/pl_handler.c
@@ -1639,7 +1639,7 @@ buildJsonEntry(int nestLevel, char* tableAlias, TargetEntry* te)
 	sprintf(nest, "%d", nestLevel);
 	// Adding JSONAUTOALIAS prevents us from modifying
 	// a column more than once
-	if(!strcmp(te->resname, "\?column\?")) {
+	if(te->resname == NULL || !strcmp(te->resname, "\?column\?")) {
 		ereport(ERROR,
 				(errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
 					errmsg("column expressions and data sources without names or aliases cannot be formatted as JSON text using FOR JSON clause. Add alias to the unnamed column or table")));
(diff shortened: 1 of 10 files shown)


contrib/babelfishpg_tsql/src/tsql_for/forjson.c
```diff
@@ -496,6 +496,12 @@ tsql_row_to_json(JsonbValue* jsonbArray, Datum record, bool include_null_values)
 
 		colname = NameStr(att->attname);
 
+		// Before using colname, let's check colname is NULL or not
+		if (!colname)
+			ereport(ERROR,
+					(errcode(ERRCODE_NULL_VALUE_NOT_ALLOWED),
+					errmsg("column name is NULL")));
```
forestkeeper: Where is the test case for this error ?

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -1639,7 +1639,7 @@ buildJsonEntry(int nestLevel, char* tableAlias, TargetEntry* te)
 	sprintf(nest, "%d", nestLevel);
 	// Adding JSONAUTOALIAS prevents us from modifying
 	// a column more than once
-	if(!strcmp(te->resname, "\?column\?")) {
+	if(te->resname == NULL || !strcmp(te->resname, "\?column\?")) {
```
tanscorpio7, line 1642: There are some more places doing this. Please search for `?column` in code base.

contrib/babelfishpg_tsql/src/tsql_for/forjson.c
tanscorpio7: Why not merge it into the below check ?

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -1659,10 +1659,27 @@ static void modifyColumnEntries(List* targetList, forjson_table **tableInfoArr,
 	int i = 0;
 	int currMax = 0;
 	ListCell* lc;
+	
+	// Add null check for input parameters
+	if (!targetList || !tableInfoArr || !colnameAlias)
+		return;
+
```
tanscorpio7: Could you explain why we are returning and not throwing an error here ?

Also comments should be /* ... */ format.

tanscorpio7 (changes requested):

contrib/babelfishpg_tsql/src/tsql_for/forjson.c
herambh-shah: @forestkeeper  we cannot have a test case where the column name is NULL. I just added it as a safety measure.
However, I found a new crash in FOR JSON PATH where if the column is empty, `tsql_row_to_json` is crashing. Hence, I added a check for this case with test cases

herambh-shah: @tanscorpio7 combined both checks

contrib/babelfishpg_tsql/src/pl_handler.c
herambh-shah: I checked and found only one more place in the `tsql_auto_row_to_json` function with a similar condition. However, we don't need to add such a check there since it's a FOR JSON AUTO function, and we already added a check for empty column names in the pl_handler.c file when we are processing column name for **FOR JSON AUTO**.

contrib/babelfishpg_tsql/src/pl_handler.c
herambh-shah: I just realized this checks would be redundant as they are already being performed in the handleForJsonAuto function before these values are passed down. So removed it

forestkeeper (approved):

PR #3994, opened 2025-07-27
fixing revoke permission from object behaviour

pranavJ23:
### Description
Problem : When we grant some permissions on schema and object in that schema to a user, in that case regardless of permission that we are trying to revoke, in babelfish we never really execute the revoke and just update the catalog.
Fixed the issue by adding a permission check in the privilege check function so that we execute revoke when the schema permission and object permission are not common.
Secondly, we are revoking the permission which are already not granted to the schema, if any permission granted to schema and object both then we won't revoke it, for that we are updating the grant-privileges list in the query tree itself before execution stage.

### Issues Resolved

[BABEL-5904]

### Check List


tanscorpio7:
Could you please describe the problem in more detail in the description ?

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -3869,7 +3869,7 @@ bool
 privilege_exists_in_bbf_schema_permissions(const char *schema_name,
 							const char *object_name,
 							const char *grantee,
-							const char *object_type)
+							const char *object_type, int curr_permission)
```
tanscorpio7: What is `curr_permission` and how is `privilege_exists_in_bbf_schema_permissions` using it ?

Also if we want to use some default value like `-1` it should be defined as a MACRO

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -4706,6 +4706,7 @@ bbf_ProcessUtility(PlannedStmt *pstmt,
 					bool exec_pg_command = false;
 					ListCell   *lc;
 					ListCell	*lc1;
+					List *filtered_list = NIL;
```
tanscorpio7: Do we reach here for INTERNAL GRANT ON SCHEMA commands ? If not then how are they getting filtered out ?

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -4771,19 +4772,39 @@ bbf_ProcessUtility(PlannedStmt *pstmt,
...
 									throw_error_for_fixed_db_role(rol_spec->rolename, dbname);
-									/*
-									 * If permission on schema exists, don't revoke any permission from the object.
-									 */
-									if (!exec_pg_command && !privilege_exists_in_bbf_schema_permissions(logical_schema, PERMISSIONS_FOR_ALL_OBJECTS_IN_SCHEMA, rol_spec->rolename, OBJ_SCHEMA))
+									/* If permission on schema exists, don't revoke any permission from the object. */
+									if (exec_pg_command == false && privilege_exists_in_bbf_schema_permissions(logical_schema, PERMISSIONS_FOR_ALL_OBJECTS_IN_SCHEMA, rol_spec->rolename, OBJ_SCHEMA, privilege))
+									{
+										exec_pg_command = false;
+									}
```
tanscorpio7: What is this IF block doing ? `exec_pg_command` has to be false to enter it and we set it to false again.

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -4798,6 +4819,7 @@ bbf_ProcessUtility(PlannedStmt *pstmt,
 					const char *obj_type = NULL;
 					Oid func_oid = LookupFuncWithArgs(OBJECT_ROUTINE, ob, true);
 					const char *func_args = NULL;
+					List *filtered_list = NIL;
```
tanscorpio7: Please add an empty line after end of all declarations.

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -4879,18 +4901,36 @@ bbf_ProcessUtility(PlannedStmt *pstmt,
...
+					{
+						if(filtered_list == NIL)
+							return;
+						list_free(grant->privileges);
+						grant->privileges = filtered_list;
+						pstmt->utilityStmt = (Node*)grant;
+					}
 					if (exec_pg_command)
 						call_prev_ProcessUtility(pstmt, queryString, readOnlyTree, context, params, queryEnv, dest, qc);
+					list_free(filtered_list);
```
tanscorpio7: Should we bring out this entire block one nesting level outside ? looks like duplicate code.

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -4771,19 +4772,39 @@ bbf_ProcessUtility(PlannedStmt *pstmt,
...
+										exec_pg_command = false;
+									}
+									else if (!privilege_exists_in_bbf_schema_permissions(logical_schema, PERMISSIONS_FOR_ALL_OBJECTS_IN_SCHEMA, rol_spec->rolename, OBJ_SCHEMA, privilege))
+									{
+										/* 
+										 * If the privilege is not common to schema and object then 
+										 * execute_pg_command true and append the privilege to filtered list 
+										 */
 										exec_pg_command = true;
+										filtered_list = lappend(filtered_list, ap);
```
tanscorpio7: Can we instead remove the priveleges which do not require any PG revoke using `list_delete_cell` ?

tanscorpio7 (changes requested):

contrib/babelfishpg_tsql/src/pl_handler.c
pranavJ23: We don't need any filtered list for case of grant on schema, we need filtered list in case of REVOKE query, because in case of revoke only we need to check weather user already has privilege(let's say A privilege) on schema has that that we are revoking from user on a object(let's say table T) in that schema. So if user has privilege on schema then we don't need to revoke the privilege from user on object.
In case of grant it won't matter if user has the privilege on schema or not, irrespective of that we will be granting the permission.

contrib/babelfishpg_tsql/src/pl_handler.c
pranavJ23: Yes can remove it

contrib/babelfishpg_tsql/src/catalog.c
pranavJ23: The curr_permission is the permission that we are trying to revoke from the user, in function privilege_exists_in_bbf_schema_permissions we are checking if schema already has that same permission(i.e curr_permission) or not by bitwise & operator.

contrib/babelfishpg_tsql/src/pl_handler.c
pranavJ23: I think we can't directly use list_delete_cell over grant->privileges while iterating over it, since it might be creating inconsistency i.e deleting from list while looping over same list.(tried doing the same getting inconsistent result)

tanscorpio7: You can try using `foreach_delete_current`

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -3964,7 +3964,29 @@ privilege_exists_in_bbf_schema_permissions(const char *schema_name,
 
 	tuple_bbf_schema = systable_getnext(scan);
 	if (HeapTupleIsValid(tuple_bbf_schema))
-		catalog_entry_exists = true;
+    {
+		if (curr_permission == -1)
```
tanscorpio7: MACRO instead of -1

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -3964,7 +3964,29 @@ privilege_exists_in_bbf_schema_permissions(const char *schema_name,
...
+			bool isnull;
+			int sch_permission;
+
+			datum = heap_getattr(tuple_bbf_schema, Anum_bbf_schema_perms_permission, RelationGetDescr(bbf_schema_rel), &isnull);
+			sch_permission = DatumGetInt32(datum);
+
+			if (!isnull && ((sch_permission & curr_permission) == curr_permission))
+				catalog_entry_exists = true;
+			else if (!isnull && ((sch_permission & curr_permission) != curr_permission))
+				catalog_entry_exists = false;
```
tanscorpio7: 1. Should not access datum before first checking isnull.
2. else if block is redundant

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -3964,7 +3964,29 @@ privilege_exists_in_bbf_schema_permissions(const char *schema_name,
 
 	tuple_bbf_schema = systable_getnext(scan);
 	if (HeapTupleIsValid(tuple_bbf_schema))
```
tanscorpio7: We need to convert this into while loop now. It is possible the the first tuple does not match the `curr_permission`.

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -5048,33 +5048,35 @@ bbf_ProcessUtility(PlannedStmt *pstmt,
...
+									{
+										/* 
+										 * If the privilege is not common to schema and object then 
+										 * execute_pg_command true and append the privilege to filtered list 
+										 */
 										exec_pg_command = true;
-
+									}
+									else
+										foreach_delete_current(grant->privileges, lc1);
```
tanscorpio7: `grant->privileges = foreach_delete_current(grant->privileges, lc1);`

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -5156,20 +5158,33 @@ bbf_ProcessUtility(PlannedStmt *pstmt,
...
+
 								/* Update the privilege in the catalog. */
 								update_privileges_of_object(logicalschema, funcname, privilege, rol_spec->rolename, obj_type, false);
 							}
 						}
 					}
+				}
+
+				if (((grant->objtype == OBJECT_FUNCTION) || (grant->objtype == OBJECT_PROCEDURE) || (grant->objtype == OBJECT_TABLE)))
+				{
```
tanscorpio7: Do we really need the check for object type ? Wouldn't `exec_pg_command` be sufficient.

tanscorpio7 (changes requested):

contrib/babelfishpg_tsql/src/catalog.c
pranavJ23: I think we do not need WHILE loop here since the schema permission is stored in interger bit as:
eg query: grant select,insert on schema::sch to l1
```
babelfish_db=# select * from sys.babelfish_schema_permissions;
 dbid | schema_name | object_name | permission |  grantee  | object_type | function_args | grantor 
------+-------------+-------------+------------+-----------+-------------+---------------+---------
    1 | sch         | ALL         |          3 | master_l1 | s           |               | 
(1 row)
```

So we are storing the sch_permission as a|b(here a = 1 for select and b = 2 for insert) so a|b = 3 
and we are simply doing the & operation with curr_permission. By fact we can say (a|b)&a = a always so if schema has same permission then condition become true else false

tanscorpio7: I would say we should still have a while loop since we are not searching using a unique key (grantee == NULL case). In future if we change the row format in this catalog table, it will hard to catch this issue with this function.

pranavJ23: Make sense, will update that
thanks!

contrib/babelfishpg_tsql/src/catalog.c
tanscorpio7: Let us follow the indentation and move the new arg to new line.

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -3964,7 +3964,30 @@ privilege_exists_in_bbf_schema_permissions(const char *schema_name,
...
-		catalog_entry_exists = true;
+    {
+		if (curr_permission == INVALID_PERMISSION)
+		{
+			catalog_entry_exists = true;
+		}
+
```
tanscorpio7: Stray extra line

contrib/babelfishpg_tsql/src/catalog.c
tanscorpio7: We are still converting this to a while loop right ?

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -5012,81 +5013,120 @@ bbf_ProcessUtility(PlannedStmt *pstmt,
...
-							/* Don't add/update an entry, if the permission is granted on column list.*/
-							if (ap->cols == NULL)
+							AccessPriv *ap = (AccessPriv *) lfirst(lc1);
+							AclMode privilege = string_to_privilege(ap->priv_name);
+							if (grant->is_grant)
 							{
-								foreach(lc, grant->grantees)
+								exec_pg_command = true;
+								/* Don't add/update an entry, if the permission is granted on column list.*/
+								if (ap->cols == NULL)
```
tanscorpio7, line 5199: Do we need column level handling for ALL case ?

contrib/babelfishpg_tsql/src/pl_handler.c
tanscorpio7: @pranavJ23

test/JDBC/expected/single_db/GRANT_SCHEMA.out
```diff
@@ -6664,3 +6664,121 @@ go
...
+-- Grant table and schema permissions
+grant select, insert on babel_test_s1.t1 to babel_test_u1;
+go
+
+grant select on schema::babel_test_s1 to babel_test_u1;
+go
+
+-- Revoke all permissions from table
+revoke all on babel_test_s1.t1 from babel_test_u1;
+go
```
tanscorpio7, line 6761: Please print the acl_privs column from pg_class before and after this statement.

tanscorpio7 (changes requested):

contrib/babelfishpg_tsql/src/pl_handler.c
pranavJ23: We are not supporting revoke/grant all on column in bbf, instead perform grant/revoke on object currently

contrib/babelfishpg_tsql/src/pl_handler.c
pranavJ23: moved it inside the object type check

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -5097,61 +5099,137 @@ bbf_ProcessUtility(PlannedStmt *pstmt,
...
+								{
+									privileges = (char **) palloc0(8 * sizeof(char *));
+									privileges[0] = "insert";
+									privileges[1] = "select";
+									privileges[2] = "update";
+									privileges[3] = "delete";
+									privileges[4] = "references";
+									privileges[5] = "truncate";
+									privileges[6] = "maintain";
+									privileges[7] = "trigger";
```
tanscorpio7: These allocation should be pstrdup("insert") I think. Please run valgrind with your changes to be sure.

tanscorpio7 (approved):

shalinilohia50 (approved):

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -3963,8 +3964,29 @@ privilege_exists_in_bbf_schema_permissions(const char *schema_name,
 	}
 
 	tuple_bbf_schema = systable_getnext(scan);
-	if (HeapTupleIsValid(tuple_bbf_schema))
-		catalog_entry_exists = true;
+	while (HeapTupleIsValid(tuple_bbf_schema))
+    {
+		if (curr_permission == INVALID_PERMISSION)
```
shalinilohia50, line 3977: What does INVALID_PERMISSION signify? Perhaps we add a comment for better understanding of implementation in future.

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -3963,8 +3964,29 @@ privilege_exists_in_bbf_schema_permissions(const char *schema_name,
...
+			catalog_entry_exists = true;
+		else
+		{
+			/* find the permission corresponding to tuple_bbf_schema */
+			Datum datum;
+			bool isnull;
+			int sch_permission = INVALID_PERMISSION;
+
+			datum = heap_getattr(tuple_bbf_schema, Anum_bbf_schema_perms_permission, RelationGetDescr(bbf_schema_rel), &isnull);
+			if(isnull)
```
shalinilohia50: Nit: `if (isnull)`

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -5097,61 +5099,137 @@ bbf_ProcessUtility(PlannedStmt *pstmt,
...
+								else
+								{
+									privileges = (char **) palloc0(8 * sizeof(char *));
+									privileges[0] = "insert";
+									privileges[1] = "select";
+									privileges[2] = "update";
+									privileges[3] = "delete";
+									privileges[4] = "references";
+									privileges[5] = "truncate";
+									privileges[6] = "maintain";
```
shalinilohia50: Babelfish doesn't support `truncate, maintain, trigger` permissions on objects.

pranavJ23: Yes, but pg does support so to handle case as: 
when we do grant all on obj to user 
and then revoke all on obj from user.
We have granted as per pg i.e granted truncate, maintain and trigger as well. but while revoking we are revoking less perms.
So while dropping user we will be getting error i.e some objs depend upon it

contrib/babelfishpg_tsql/src/catalog.c
pranavJ23: invalid permission signifying we are not comparing any curr_permission(sending -1) with schema permission.
Sure, will add one comment
thanks!

jsudrik (approved):
Approved.

PR #4015, opened 2025-08-07
Ownership Chaining: Support relation references inside views

HarshLunagariya:
### Description

This commit implements the ownership chaining support for table/view object 
references inside view. The basic idea of Ownership chaining is that when 
one object references another, and both have the same owner, then permissions 
are only checked when the top-level object is accessed.

The scope of the current commit is to only support scenarios when parent 
object is view and child objects are table/view. "[parent] view -> [child] 
function" scenarios is out-of-scope for this effort.

Implementation:
- Add infrastructure changes to support ownership tracking:
  * insideView flag in RTEPermissionInfo for view context tracking
  * parentOwnerId and insideView fields in FuncExpr (for future use)
  * walk_view_rule_hook for extension-specific view rule processing and 
     apply view context marking in ApplyRetrieveRule

- Add walker to traverse view definition parse trees:
  * Mark relations and functions inside views for permission checking
  * Set checkAsUser to view_owner for owned relations to enable chaining
  * Track view context using insideView flag
  * Only applies to user-defined objects created in non-shared schemas in TSQL mode

This enables proper ownership chaining behavior where tables and 
views owned by the view owner can be accessed through the view.

Note: While infrastructure for function ownership chaining is included, 
the actual implementation of function ownership chaining is out of scope 
for this commit.

### Issues Resolved

BABEL-6026

* **Use case based -**

* **Boundary conditions -**

* **Arbitrary inputs -**

* **Negative test cases -**

* **Performance tests -**

* **Tooling impact -**

* **Client tests -**

### Check List


contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -7634,4 +7645,112 @@ find_all_view_references(Node *node, List **view_oids)
 									find_view_references_walker,
 									(void *) &context,
 									QTW_EXAMINE_RTES_BEFORE);
-}
\ No newline at end of file
+}
```
tanscorpio7: Wondering If we should move this to another file. hooks.c is already to big :)

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -7634,4 +7645,112 @@ find_all_view_references(Node *node, List **view_oids)
...
+	char *physical_schemaname = NULL;
+	if (node == NULL)
+		return false;
+
+	if (IsA(node, Query))
+    {
+        Query *query = (Query *) node;
+        ListCell *lc;
+        
+        foreach(lc, query->rtable)
```
tanscorpio7: The structure for this walker can be kept similar to `setRuleCheckAsUser_walker`.

tanscorpio7 (changes requested):

contrib/babelfishpg_tsql/src/hooks.c
HarshLunagariya: Yup, Makes sense.

contrib/babelfishpg_tsql/src/hooks.c
HarshLunagariya: Took a look there. Both walker has a bit different requirements. `setRuleCheckAsUser_walker` only needs to walk over rtes on other hand our walker needs to walk over `FuncExpr` as well -- If we consider subsequent implementation of ownership chaining for function/procedure. So, keeping in this form serves the purpose.

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -117,6 +117,11 @@ typedef enum PltsqlInitPrivsOptions
 	ERROR_INIT_PRIVS
 } PltsqlInitPrivsOptions;
 
+/* Used to mark parsenode inside/outside view to apply ownership chaining logic */
+#define PNODE_UNMARKED		0
+#define PNODE_INSIDE_VIEW	1
+#define PNODE_OUTSIDE_VIEW	2
```
KushaalShroff: should be an enum

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -7634,4 +7645,102 @@ find_all_view_references(Node *node, List **view_oids)
...
+            RTEPermissionInfo *perminfo = NULL;
+
+			/* Process tables and view subqueries */
+            if (OidIsValid(rte->relid) &&
+				(rte->rtekind == RTE_RELATION ||
+				 (rte->rtekind == RTE_SUBQUERY && rte->relkind == RELKIND_VIEW)))
+            {
+                nspid = get_rel_namespace(rte->relid);
+                physical_schemaname = get_namespace_name(nspid);
+                perminfo = getRTEPermissionInfo(query->rteperminfos, rte);
```
Deepesh125: we need to check right set of permissions SELECT ,  INSERT or UPDATE. Do we need to apply ownership chaining in case of INSERT / UPDATE? That might be through INSTEAD OF TRIG, we need to check it out

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -3068,6 +3068,40 @@ bbf_ProcessUtility(PlannedStmt *pstmt,
 			{
 				ViewStmt *stmt = (ViewStmt *) parsetree;
 
+				/*
+				 * By default, TSQL view should be created with security_invoker
+				 * property. It adds security_invoker option to the options list.
+				 */
+				if (sql_dialect == SQL_DIALECT_TSQL && IS_TDS_CLIENT() && (stmt->createOrAlter || stmt->replace == false))
```
KushaalShroff: Why do we need both conditions: sql_dialect == SQL_DIALECT_TSQL && IS_TDS_CLIENT()? I think here we should only have SQL_DIALECT_TSQL check
what are the interoperable use cases where we restrict adding security_invoker?

KushaalShroff: Have we documented the expected behaviour shift for old vs new views?

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -3068,6 +3068,40 @@ bbf_ProcessUtility(PlannedStmt *pstmt,
 			{
 				ViewStmt *stmt = (ViewStmt *) parsetree;
 
+				/*
+				 * By default, TSQL view should be created with security_invoker
+				 * property. It adds security_invoker option to the options list.
+				 */
```
KushaalShroff, line 3138: Why only for stmt->replace == false?

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -7634,4 +7650,126 @@ find_all_view_references(Node *node, List **view_oids)
...
+				 rte->relkind == RELKIND_VIEW))
+            {
+                nspid = get_rel_namespace(rte->relid);
+                physical_schemaname = get_namespace_name(nspid);
+                perminfo = getRTEPermissionInfo(query->rteperminfos, rte);
+				nonDMLpermscheck = perminfo->requiredPerms & ~(ACL_SELECT | ACL_INSERT | ACL_UPDATE | ACL_DELETE);
+                
+                if (physical_schemaname && !is_shared_schema(physical_schemaname))
+                {
+                    if (get_rel_owner(rte->relid) == *context && nonDMLpermscheck == 0)
```
KushaalShroff: nit: nonDMLpermscheck is bool
Let be consistent by using nonDMLpermscheck == false

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -7634,4 +7650,126 @@ find_all_view_references(Node *node, List **view_oids)
...
+
+			/* Process tables and view subqueries */
+            if (OidIsValid(rte->relid) &&
+				(rte->rtekind == RTE_RELATION ||
+				 rte->relkind == RELKIND_VIEW))
+            {
+                nspid = get_rel_namespace(rte->relid);
+                physical_schemaname = get_namespace_name(nspid);
+                perminfo = getRTEPermissionInfo(query->rteperminfos, rte);
+				nonDMLpermscheck = perminfo->requiredPerms & ~(ACL_SELECT | ACL_INSERT | ACL_UPDATE | ACL_DELETE);
```
KushaalShroff: Which ACLs are we supporting?
please add comments

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -7634,4 +7650,126 @@ find_all_view_references(Node *node, List **view_oids)
...
+                            	 mark_nodes_inside_view_walker,
+                            	 (void *) context,
+                            	 0);
+	}
+	else if (IsA(node, FuncExpr))
+    {
+        FuncExpr *funcexpr = (FuncExpr *) node;
+		nspid = get_func_namespace(funcexpr->funcid);
+		physical_schemaname = get_namespace_name(nspid);
+		if (physical_schemaname && !is_shared_schema(physical_schemaname))
```
KushaalShroff: What happens if this was executed through sql APIs for sp_prepare/exec, they are defined in sys, right?

So ownership chaining wont work?

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -7634,4 +7650,126 @@ find_all_view_references(Node *node, List **view_oids)
...
+	if (!IS_TDS_CLIENT() || InSecurityRestrictedOperation())
+		return;
+
+	mark_nodes_inside_view_walker((Node *)query,
+								  &view_owner);
+
+}
+
+static void
+tsql_handle_target_view_hook(RTEPermissionInfo *new_perminfo, RangeTblEntry *view_rte)
```
KushaalShroff: Lets add comment on top of this hook function

test/JDBC/input/ownership/ownership_chaining-vu-prepare.mix
```diff
@@ -0,0 +1,559 @@
...
+-- CASE 2.1: Subquery in SELECT clause
+create view v3014_2_1 as select t3014_1.a, (SELECT TOP 1 a FROM t3014_2 b2 WHERE b2.a = t3014_1.a) AS subquery_data FROM t3014_1;
+go
+
+-- CASE 2.2: Subquery in WHERE clause
+create view v3014_2_2 as SELECT * FROM t3014_1 t1 WHERE t1.a IN (SELECT a FROM t3014_2 WHERE a > 1);
+go
+
+-- CASE 2.3: Subquery in FROM clause
+create view v3014_2_3 as SELECT sq.* FROM (SELECT t1.a, t2.a as a2 FROM t3014_1 t1 LEFT JOIN t3014_2 t2 ON t1.a = t2.a) sq;
```
KushaalShroff, line 104: Do we have tests with selecting * from function?

KushaalShroff (changes requested):

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -3068,6 +3068,40 @@ bbf_ProcessUtility(PlannedStmt *pstmt,
...
+				{
+					ListCell   *lc;
+					bool        security_invoker_found = false;
+
+					/* Check if security_invoker option already exists */
+					foreach(lc, stmt->options)
+					{
+						DefElem    *defel = (DefElem *) lfirst(lc);
+
+						if (strcmp(defel->defname, "security_invoker") == 0)
```
Deepesh125, line 3147: Can we ever get this security_invoker option for TSQL created views?

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -3068,6 +3068,40 @@ bbf_ProcessUtility(PlannedStmt *pstmt,
...
+				 */
+				if (sql_dialect == SQL_DIALECT_TSQL && IS_TDS_CLIENT() && (stmt->createOrAlter || stmt->replace == false))
+				{
+					ListCell   *lc;
+					bool        security_invoker_found = false;
+
+					/* Check if security_invoker option already exists */
+					foreach(lc, stmt->options)
+					{
+						DefElem    *defel = (DefElem *) lfirst(lc);
```
Deepesh125: Nitpick: May be better to use foreach_node?

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -3068,6 +3068,40 @@ bbf_ProcessUtility(PlannedStmt *pstmt,
...
+
+						if (strcmp(defel->defname, "security_invoker") == 0)
+						{
+							/* Modify existing option to true */
+							defel->arg = (Node *) makeInteger(1);
```
Deepesh125: shouldn't be boolean?

Deepesh125: Do we also want to do this for security restricted operations? ie internal commands?

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -3068,6 +3068,40 @@ bbf_ProcessUtility(PlannedStmt *pstmt,
...
+							security_invoker_found = true;
+							break;
+						}
+					}
+
+					/* If security_invoker option not found, add it */
+					if (!security_invoker_found)
+					{
+						DefElem *new_option;
+						new_option = makeDefElem("security_invoker",
```
Deepesh125, line 3160: From gram.y
```
$$ = makeDefElem("security", (Node *) makeBoolean(false), @1);
```

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -614,6 +623,11 @@ InstallExtendedHooks(void)
 	prev_pre_QueryRewrite_hook = pre_QueryRewrite_hook;
 	pre_QueryRewrite_hook = repair_broken_views;
 
+	prev_walk_view_rule_hook = walk_view_rule_hook;
```
Deepesh125: where are we actually using this prev_ hook?

test/JDBC/expected/ownership_chaining-vu-prepare.out
```diff
@@ -0,0 +1,559 @@
...
+go
+
+-- CASE 1.4 nesting case - two-level
+create view v3014_1_4 as select * from v3014_b;
+go
+
+
+-- CASE 2 Subquery/CTE inside views
+-- CASE 2.1: Subquery in SELECT clause
+create view v3014_2_1 as select t3014_1.a, (SELECT TOP 1 a FROM t3014_2 b2 WHERE b2.a = t3014_1.a) AS subquery_data FROM t3014_1;
```
Deepesh125, line 96: select case (select sum(a) FROM t3014_2 b2 WHERE b2.a = t3014_1.a) when 2 then .... else .... from t3014_1

test/JDBC/expected/ownership_chaining-vu-prepare.out
```diff
@@ -0,0 +1,559 @@
...
+-- CASE 2.3: Subquery in FROM clause
+create view v3014_2_3 as SELECT sq.* FROM (SELECT t1.a, t2.a as a2 FROM t3014_1 t1 LEFT JOIN t3014_2 t2 ON t1.a = t2.a) sq;
+go
+
+-- CASE 2.4: Correlated subquery in WHERE
+create view v3014_2_4 as SELECT * FROM t3014_1 t1 WHERE EXISTS (SELECT 1 FROM t3014_2 t2 WHERE t2.a = t1.a);
+go
+
+-- CASE 2.5: Multiple nested subqueries
+create view v3014_2_5 as SELECT * FROM t3014_1 t1 WHERE a IN (SELECT a FROM t3014_2 WHERE a IN (SELECT a FROM t3014_1 WHERE a < 3));
```
Deepesh125, line 112: EXIST or NOT EXIST sublink

contrib/babelfishpg_tsql/src/hooks.c
HarshLunagariya: Added

contrib/babelfishpg_tsql/src/hooks.c
HarshLunagariya: Still keeping it as MACROS as we are using this to assign them in parsenode fields defined in engine. If we want to define it as enum then we need to include it in header in engine.

contrib/babelfishpg_tsql/src/pl_handler.c
HarshLunagariya: > Have we documented the expected behaviour shift for old vs new views?

Yes, that's planned.

HarshLunagariya: > Why do we need both conditions: sql_dialect == SQL_DIALECT_TSQL && IS_TDS_CLIENT()? I think here we should only have SQL_DIALECT_TSQL check what are the interoperable use cases where we restrict adding security_invoker?

As discussed, Keeping the PG endpoint as it is even in TSQL dialect.

contrib/babelfishpg_tsql/src/pl_handler.c
HarshLunagariya: Addressed

contrib/babelfishpg_tsql/src/pl_handler.c
HarshLunagariya: Addressed.

contrib/babelfishpg_tsql/src/pl_handler.c
HarshLunagariya: Nice suggestion, I was not aware about this. Addressed.

contrib/babelfishpg_tsql/src/pl_handler.c
HarshLunagariya: No

contrib/babelfishpg_tsql/src/hooks.c
HarshLunagariya: Made it AclMode. and modified it to `nonDMLpermscheck == ACL_NO_RIGHTS`

contrib/babelfishpg_tsql/src/pl_handler.c
HarshLunagariya: (stmt->createOrAlter && !stmt->replace) captures the TSQL "CREATE OR ALTER view" stmt and
(!stmt->createOrAlter && !stmt->replace) will capture the TSQL "CREATE view" stmt, combination of both condition simplifies into !(stmt->replace). So only relying on !(stmt->replace) now. Since we don't want to change the older views.

contrib/babelfishpg_tsql/src/hooks.c
HarshLunagariya: Added.

contrib/babelfishpg_tsql/src/hooks.c
HarshLunagariya: Added.

contrib/babelfishpg_tsql/src/hooks.c
HarshLunagariya: It will work, I have already added the testcases. This logic will be applied when given function is referenced as child object inside the view. But it is not valid scenario for the views `create view v1 as select exec sp_prepare/exec`.

HarshLunagariya:
```
-- CASE 7.3: Prepare/Execute with ownership chain
-- Prepare statement
DECLARE @handle int;
DECLARE @stmt nvarchar(max) = N'SELECT * FROM v3014_7_3 WHERE a = @p1';
EXEC sp_prepare @handle OUTPUT, N'@p1 int', @stmt;
GO
~~START~~
int
~~END~~

-- Execute prepared statement multiple times
EXEC sp_execute 1, 1;
GO
~~START~~
int
~~END~~

EXEC sp_execute 1, 2;
GO
~~START~~
int
~~END~~

-- Deallocate prepared statement
EXEC sp_unprepare 1;
GO

-- CASE 7.4: Dynamic SQL with ownership chain
DECLARE @sql nvarchar(max);
SET @sql = N'SELECT * FROM v3014_7_4 WHERE a = @id';
EXEC sp_executesql @sql, N'@id int', 1;
GO
~~START~~
int
~~END~~

```

contrib/babelfishpg_tsql/src/hooks.c
HarshLunagariya: Removed them.

test/JDBC/expected/ownership_chaining-vu-prepare.out
HarshLunagariya: Added.

test/JDBC/expected/ownership_chaining-vu-prepare.out
HarshLunagariya: Added.

contrib/babelfishpg_tsql/src/pl_handler.c
tanscorpio7: @pranavJ23 @HarshLunagariya.
Will dump restore in MVU change existing views to invoker ? Also should we throw an error if we find security definer option here since that is an unexpected scenario ?

contrib/babelfishpg_tsql/src/hooks.c
tanscorpio7: Also we should use relcache here instead of doing multiple syscache lookups.

contrib/babelfishpg_tsql/src/hooks.c
tanscorpio7: @pranavJ23 ☝️

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -7688,3 +7700,137 @@ find_all_view_references(Node *node, List **view_oids)
...
+                if (physical_schemaname && !is_shared_schema(physical_schemaname))
+                {
+                    if (get_rel_owner(rte->relid) == *context && nonDMLpermscheck == ACL_NO_RIGHTS)
+                    {
+                        perminfo->checkAsUser = *context;
+                    }
+                    perminfo->insideView = PNODE_INSIDE_VIEW;
+                }
+                if (physical_schemaname)
+                    pfree(physical_schemaname);
```
tanscorpio7: Should be outside the top if block.

contrib/babelfishpg_tsql/src/pl_handler.c
tanscorpio7: I think just having sql_dialect check and TDS check is sufficient. I do not understand the exact need of `stmt->replace` & `InSecurityRestrictedOperation()` here

tanscorpio7 (changes requested):

contrib/babelfishpg_tsql/src/pl_handler.c
HarshLunagariya: No, Keeping it as it is since we decided not to change the configuration of existing views.

contrib/babelfishpg_tsql/src/hooks.c
pranavJ23: Not sure if we should move it outside the loop, because we should free the physical_schemaname before allocating it again inside the loop itself

contrib/babelfishpg_tsql/src/pl_handler.c
pranavJ23: Just an observation: since we are removing the check ```stmt->replace```, as per the suggestion above: https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/4015/files#r2325291807 
If we do create or alter view then the existing views will also change to invoker

tanscorpio7: Even with that check, CREATE OR ALTER would have changed existing views to invoker right ?

pranavJ23: Right, alter view would have definer(with check stmt->replace) instead of definer (not create or alter, my fault)
the testfile output [here](https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/4015/files#diff-73874d5e73d54d043be4ce48051ce2786e0dd87638f4500767d5fd26eca42d53R48)()
was definer with check !stmt->replace but now it's invoker, which makes sense

tanscorpio7 (approved):

contrib/babelfishpg_tsql/src/hooks.c
Deepesh125: @pranavJ23 / @HarshLunagariya  We have defined very same macros in contrib/babelfishpg_tsql/src/pltsql_permissions.h and including the same file here then why do we need to define these macros?

Deepesh125 (dismissed):

contrib/babelfishpg_tsql/src/hooks.c
pranavJ23: yes redundant, removing it. Thanks!

Deepesh125 (dismissed):

KushaalShroff (approved):

PR #4020, opened 2025-08-14
Linestring [Part 1]: Parser changes and TDS sender support

Mohitraj-g:
#### **LINESTRING SUPPORT**  

This PR implements LINESTRING geometry type support in Babelfish, expanding beyond the current POINT-only implementation.

#### **Key Changes**:
1. Parser Modifications
   * Added LINESTRING syntax support
   * Implemented handling of varying point dimensions within LINESTRING
   * Added normalization of points to highest dimension using NaN values

2. TDS Layer Enhancements
   * TDS Sender:
   * used cast functions through plugin pointer

3. STAsBinary modification for empty linestring

4. Character to geometry/geography CASTs
5. Geometry/Geography  to character CASTs

### Issues Resolved

BABEL-5766
BABEL-5767
BABEL-5978

* **Use case based -** Yes

* **Arbitrary inputs -** Yes

* **Performance tests -**

* **Tooling impact -**

* **Client tests -**

### Check List


contrib/babelfishpg_common/sql/geography.sql
```diff
@@ -272,6 +272,29 @@ CREATE OR REPLACE FUNCTION sys.Geography__STPointFromText(sys.NVARCHAR, integer)
...
+		ELSIF $1 IS NULL THEN
+			RETURN NULL;
+		END IF;
+		geom = (SELECT sys.geogfromtext_helper($1::text, $2));
+		Geomtype = (SELECT sys.ST_GeometryType(geom));
+
+		IF Geomtype = 'ST_LineString' THEN
+			RETURN geom;
+		ELSE
+			RAISE EXCEPTION '% is not supported', Geomtype;
```
rishabhtanwar29: What if POINT geometry is passed to this function? Will we throw "not supported" in that case or should we be throwing an invalid function usage error?

contrib/babelfishpg_common/sql/geometry.sql
```diff
@@ -267,6 +267,29 @@ CREATE OR REPLACE FUNCTION sys.Geometry__STPointFromText(sys.NVARCHAR, integer)
...
+		ELSIF $1 IS NULL THEN
+			RETURN NULL;
+		END IF;
+		geom = (SELECT sys.geomfromtext_helper($1::text, $2));
+		Geomtype = (SELECT sys.ST_GeometryType(geom));
+
+		IF Geomtype = 'ST_LineString' THEN
+				RETURN geom;
+		ELSE
+			RAISE EXCEPTION '% is not supported', Geomtype;
```
rishabhtanwar29: Same as above.

contrib/babelfishpg_common/src/geo.c
```diff
@@ -189,3 +189,295 @@ rewrite_point_dim_query(POINT coord)
...
+
+/**
+ * Initialize a PointArray structure with default capacity
+ * Allocates memory for points array with initial capacity of 1024 points.
+ * Sets count to 0 and reports error if memory allocation fails.
+ */
+void 
+init_point_array(PointArray *pa) 
+{
+    pa->capacity = 1024;
```
rishabhtanwar29: Why 1024? Do we think a line in general has these many points? To me it seems to be a pretty high estimate, I think we can go with a smaller value like 64 or maybe less since line geometries are going to have few points in general?

contrib/babelfishpg_common/src/geo.c
```diff
@@ -189,3 +189,295 @@ rewrite_point_dim_query(POINT coord)
...
+        /* Add comma between points, except after the last point */
+        if (i < pa->count - 1) 
+            appendStringInfoString(&output, ", ");
+    }
+
+    /* Close parenthesis */
+    appendStringInfoChar(&output, ')');
+    
+    /* Create a copy of the string that will survive after we free output */
+    res = pstrdup(output.data);
```
rishabhtanwar29: Even StringInfo's data is `palloc'ed` so we don't need to create a copy and can directly return it.

contrib/babelfishpg_common/src/geo.c
```diff
@@ -189,3 +189,295 @@ rewrite_point_dim_query(POINT coord)
...
+        if (FLAGS_GET_Z(p.flags) && FLAGS_GET_M(p.flags)) 
+        {
+            /* Point has both Z and M flags */
+            if (!isnan(p.z) && !isnan(p.m)) 
+            {
+                /* Both Z and M are not NaN */
+                appendStringInfo(&output, " %s", FLOAT8_TO_CSTRING(p.z));
+                appendStringInfo(&output, " %s", FLOAT8_TO_CSTRING(p.m));
+            }
+            else if (!isnan(p.z) && isnan(p.m))
```
rishabhtanwar29: I think checking `isnan(p.m)` is redundant here.

contrib/babelfishpg_common/src/geo.c
```diff
@@ -189,3 +189,295 @@ rewrite_point_dim_query(POINT coord)
...
+                /* Both Z and M are not NaN */
+                appendStringInfo(&output, " %s", FLOAT8_TO_CSTRING(p.z));
+                appendStringInfo(&output, " %s", FLOAT8_TO_CSTRING(p.m));
+            }
+            else if (!isnan(p.z) && isnan(p.m)) 
+            {
+                /* Only Z is not NaN */
+                appendStringInfo(&output, " %s", FLOAT8_TO_CSTRING(p.z));
+            }
+            else if (isnan(p.z) && !isnan(p.m))
```
rishabhtanwar29: Similarly checking `isnan(p.z)` seems redundant here.

contrib/babelfishpg_common/src/geo.c
```diff
@@ -189,3 +189,295 @@ rewrite_point_dim_query(POINT coord)
...
+        /* Add comma between points, except after the last point */
+        if (i < pa->count - 1) 
+            appendStringInfoString(&output, ", ");
+    }
+    
+    /* Close parenthesis */
+    appendStringInfoChar(&output, ')');
+
+    /* Create a copy of the string that will survive after we free output */
+    result = pstrdup(output.data);
```
rishabhtanwar29: Same as above, can directly return output.data without copy.

contrib/babelfishpg_common/src/geo.c
```diff
@@ -189,3 +189,295 @@ rewrite_point_dim_query(POINT coord)
...
+    }
+
+    /* Close parenthesis */
+    appendStringInfoChar(&output, ')');
+    
+    /* Create a copy of the string that will survive after we free output */
+    res = pstrdup(output.data);
+    
+    /* Clean up the StringInfoData */
+    pfree(output.data);
```
rishabhtanwar29: Should we free pointArray as well?

contrib/babelfishpg_common/src/spatialtypes.c
```diff
@@ -136,6 +145,20 @@ EMPTY_COORD[] = {
     0xff, 0xff, 0xff, 0xff
 };
 
+/* Linestring instances containing more than 2 points have trailing metadata of 22 bytes */
+static const uint8 
+line_end_metada[] = {
```
rishabhtanwar29: nit: meta**data**
Question: What does these bytes represent?

contrib/babelfishpg_common/src/spatialtypes.c
```diff
@@ -330,6 +366,81 @@ GetGeometryTypeName(FunctionCallInfoBaseData *fcinfo, Datum geom_datum)
...
+    float8 lat;
+    int npoints;
+    int i;
+    Datum point;
+
+    /* Initialize function call info */
+    InitFunctionCallInfoData(*fcinfo_local, NULL, 2, InvalidOid, NULL, NULL);
+    
+    /* Get geometry type name using helper function */
+    geom_type = GetGeometryTypeName(fcinfo_local, geom_datum);
```
rishabhtanwar29, line 394: Do we have function in postgis which returns geometry type as Enum/integer instead of string? I think `check_geom_type` function is getting a bit expensive with inclusion of new geometries.

contrib/babelfishpg_common/src/spatialtypes.c
```diff
@@ -1103,173 +1154,350 @@ geography_from_bytea(PG_FUNCTION_ARGS)
...
-    /* Determine if the geometry has an SRID */
     geom_data->has_srid = geom_data->srid_flag & SRID_MASK;
     
-    /* Determine the TSQL dimensionality based on the PostGIS's dimension flags */
     switch (geom_data->srid_flag & DIMENSION_MASK) 
     {
-        case POSTGIS_DIM_XY:  /* XY or Empty */
-            if (geom_data->is_empty) 
+        case POSTGIS_DIM_XY:
+            switch (geom_data->is_empty)
```
rishabhtanwar29: I think using `switch-case` for a bool type is an overkill.

contrib/babelfishpg_common/src/spatialtypes.c
```diff
@@ -1103,173 +1154,350 @@ geography_from_bytea(PG_FUNCTION_ARGS)
...
+                    handle_xy_dimension(geom_data);
+                    break;
             }
             break;
-            
-        case POSTGIS_DIM_XYZ:  /* XYZ (3D) */
-            geom_data->geom_type = 0x0D;
-            geom_data->coord_size = COORD_SIZE_XYZ;
+        case POSTGIS_DIM_XYZ:
+            handle_xyz_dimension(geom_data);
```
rishabhtanwar29: Should we inline all these `handle_*` functions here considering there is no other caller than this function itself?

contrib/babelfishpg_common/src/spatialtypes.c
```diff
@@ -136,6 +145,20 @@ EMPTY_COORD[] = {
...
+static const uint8 
+line_end_metada[] = {
+    0x01, 0x00, 0x00, 0x00, 0x01, 0x00, 0x00, 0x00,
+    0x00, 0x01, 0x00, 0x00, 0x00, 0xff, 0xff, 0xff,
+    0xff, 0x00, 0x00, 0x00, 0x00, 0x02
+};
+
+/* NAN format used by TSQL */
+static const uint8 
+SPECIFIC_NAN[8] = {
```
rishabhtanwar29: nit: use `COORD_SIZE` instead of hardcoded `8`

contrib/babelfishpg_common/src/spatialtypes.h
```diff
@@ -0,0 +1,23 @@
...
+extern Datum charTogeog(PG_FUNCTION_ARGS);
+extern Datum geography_point(PG_FUNCTION_ARGS);
+extern Datum st_as_binary_geometry(PG_FUNCTION_ARGS);
+extern Datum st_as_binary_geography(PG_FUNCTION_ARGS);
+extern Datum st_as_text(PG_FUNCTION_ARGS);
+extern Datum geometry_astext(PG_FUNCTION_ARGS);
+extern Datum geometry_asbpchar(PG_FUNCTION_ARGS);
+
+/* Function to rewrite geospatial data */
+extern text* geo_wkt_rewrite(text *input_text);
```
rishabhtanwar29: I think we just need to expose `bytea_from_geometry` and `bytea_from_geography`, rest of them can be removed.

test/JDBC/expected/Test-Linestring-vu-verify.out
rishabhtanwar29, line 1: There are so many tests which makes it hard to review them. Having a test plan would be more suitable for this.

test/python/expected/upgrade_validation/expected_dependency.out
```diff
@@ -455,6 +455,7 @@ Function sys.geography(sys.bpchar)
 Function sys.geography(sys.geography,integer,boolean)
 Function sys.geography(text,integer,boolean)
 Function sys.geography__stflipcoordinates(sys.geography)
+Function sys.geography__stlinefromtext(sys.nvarchar,integer)
```
rishabhtanwar29, line 457: Let's add dependency test for stlinefromtext for both geometry and geography.

contrib/babelfishpg_common/src/geo.c
Mohitraj-g: Yes, added since it's also a terminal function in the flow

contrib/babelfishpg_common/src/spatialtypes.c
Mohitraj-g: These trailing 22 bytes are required by T-SQL's spatial format but their exact meaning is unclear however they are constant for all linestring with >2 points and doesn't have any identified use case. The pattern only changes for different geometry types.

contrib/babelfishpg_common/src/spatialtypes.c
Mohitraj-g: Upon investigation, no such function exists in PostGIS.

contrib/babelfishpg_common/sql/geography.sql
```diff
@@ -267,7 +267,30 @@ CREATE OR REPLACE FUNCTION sys.Geography__STPointFromText(sys.NVARCHAR, integer)
 		IF Geomtype = 'ST_Point' THEN
 			RETURN geom;
 		ELSE
-			RAISE EXCEPTION '% is not supported', Geomtype;
+			RAISE EXCEPTION 'Invalid function usage : Expected "POINT"';
+		END IF;
+	END;
+	$$ LANGUAGE plpgsql IMMUTABLE PARALLEL SAFE;
+
+CREATE OR REPLACE FUNCTION sys.Geography__STLineFromText(sys.NVARCHAR, integer)
```
Yvinayak07: What does the second argument `integer` signifies? is it SRID, if yes then can we specify the argument name.
ref. function `sys.GeomPoint_helper`.

contrib/babelfishpg_tds/src/backend/tds/tdstypeio.c
```diff
@@ -4385,146 +4363,42 @@ int
 TdsSendSpatialHelper(FmgrInfo *finfo, Datum value, void *vMetaData, int TdsInstr)
```
Yvinayak07, line 4363: Can we also update the comment as we are not using `GSERIALIZED format` to store bytes.

contrib/babelfishpg_tds/src/backend/tds/tdstypeio.c
```diff
@@ -4385,146 +4363,42 @@ int
...
+    /* Set up function call info for spatial conversion functions */
+    LOCAL_FCINFO(fcinfo, 1);
+    InitFunctionCallInfoData(*fcinfo, NULL, 1, InvalidOid, NULL, NULL);
+    fcinfo->args[0].value = value;
+    fcinfo->args[0].isnull = false;
+    
+    /* Convert spatial data to binary format based on type */
+    if (TdsInstr == (int)INSTR_TDS_DATATYPE_GEOMETRY)
+        bytea_datum = pltsql_plugin_handler_ptr->sql_bytea_from_geometry(fcinfo);
     else
```
Yvinayak07, line 4384: Are we sure that `TdsSendSpatialHelper` is invoked only for `INSTR_TDS_DATATYPE_GEOMETRY` and `INSTR_TDS_DATATYPE_GEOGRAPHY`, as we left the else condition for Geography.

contrib/babelfishpg_tds/src/backend/tds/tdstypeio.c
```diff
@@ -4385,146 +4363,42 @@ int
...
     destBuf = TdsEncodingConversion(buf, len, PG_UTF8, col->encoding, &actualLen);
 
     TDSInstrumentation(TdsInstr);
 
     rc = TdsSendPlpDataHelper(destBuf, actualLen);
 
-    pfree(destBuf);
+    /* Clean up allocated memory if encoding conversion created a new buffer */
+    if (destBuf != buf)
+        pfree(destBuf);
```
Yvinayak07, line 4400: Why can't we free both (destBuf and buf) the `char *` is they are not NULL, even if they are equal?

contrib/babelfishpg_tds/src/backend/tds/tdstypeio.c
```diff
@@ -4385,146 +4363,42 @@ int
...
-    unsigned char *itr;
+                len = 0,                /* Number of bytes in the binary data */
+                actualLen;              /* Number of bytes after encoding conversion */
 
+    char        *destBuf,               /* Buffer after encoding conversion */
+                *buf = NULL;            /* Raw binary data from spatial object */
+    Datum       bytea_datum;            /* Datum containing binary spatial data */
+    bytea       *bytea_result;          /* Binary result from spatial conversion */
     TdsColumnMetaData *col = (TdsColumnMetaData *) vMetaData;
-    GSERIALIZED *gser = (GSERIALIZED *)PG_DETOAST_DATUM(value);    /* Used to Store the bytes in the Format which is stored in PostGIS */
```
Yvinayak07, line 4403: as we are removing this `GSERIALIZED` struct usage can we also remove its structure:
```
typedef struct
{
    uint32_t size; /* For PgSQL use only, use VAR* macros to manipulate. */
    uint8_t srid[3]; /* 24 bits of SRID */
    uint8_t gflags; /* HasZ, HasM, HasBBox, IsGeodetic */
    uint8_t data[1]; /* See gserialized.txt */
} GSERIALIZED;
```

contrib/babelfishpg_common/src/spatialtypes.h
```diff
@@ -0,0 +1,5 @@
+#include "postgres.h"
```
Yvinayak07, line 12: As we have created new header file could we also add comment on top of this header file for what pupose we will use it in future or just check once if we can add this content in `geo_data.h`.

contrib/babelfishpg_common/src/babelfishpg_common.h
```diff
@@ -103,5 +103,7 @@ typedef struct common_utility_plugin
 	TimeADT		(*initializeToDefaultTime) (int32 typmod);
 	Timestamp	(*roundoff_datetime) (Timestamp timestamp);
 	void		(*UpdateToNextDayHelper) (struct pg_tm *tm);
-	void 		(*handle_type_and_collation) (struct Node *node, Oid typid, Oid collationid);
+  void 		(*handle_type_and_collation) (struct Node *node, Oid typid, Oid collationid);
```
Yvinayak07: NIT: let's correct the indentation for `(*handle_type_and_collation)` so it doesn't appear in diff.

contrib/babelfishpg_common/src/geo.c
```diff
@@ -189,3 +188,287 @@ rewrite_point_dim_query(POINT coord)
 
     return output.data; 
 }
+
+/**
+ * Initialize a PointArray structure with default capacity
+ * Allocates memory for points array with initial capacity of 1024 points.
```
Yvinayak07: let's update the comment ad we have changed the capacity here.

contrib/babelfishpg_common/src/geo.c
```diff
@@ -189,3 +188,287 @@ rewrite_point_dim_query(POINT coord)
...
+        if ((FLAGS_GET_Z(p.flags) && isnan(p.z)) ||
+            (FLAGS_GET_M(p.flags) && isnan(p.m)))
+        {
+            ereport(ERROR,
+                    (errcode(ERRCODE_INVALID_PARAMETER_VALUE),
+                        errmsg("Invalid coordinate value (NaN)")));
+        }
+
+        if (FLAGS_GET_Z(p.flags) && FLAGS_GET_M(p.flags)) 
+            has_zm = 1;
```
Yvinayak07: I think we don't need these variables `has_zm` , `has_z`, and `has_m` as we can directly return `ZM` , `M` and `Z` respectively from here. if needed then change the type to `bool`.

contrib/babelfishpg_common/src/geo.c
```diff
@@ -189,3 +188,287 @@ rewrite_point_dim_query(POINT coord)
...
+        POINT p = pa->points[i];
+
+        /* Check for NaN values in Z nd M coordinate if present */
+        if ((FLAGS_GET_Z(p.flags) && isnan(p.z)) ||
+            (FLAGS_GET_M(p.flags) && isnan(p.m)))
+        {
+            ereport(ERROR,
+                    (errcode(ERRCODE_INVALID_PARAMETER_VALUE),
+                        errmsg("Invalid coordinate value (NaN)")));
+        }
```
Yvinayak07, line 257: What is could see in `transform_points`, `Z` and `M` points could be NaN. so this condition should be for `Z` and `M` or `X` and `Y`?

contrib/babelfishpg_common/src/geo.c
```diff
@@ -189,3 +188,287 @@ rewrite_point_dim_query(POINT coord)
...
+                if (!FLAGS_GET_Z(p->flags)) p->z = NAN;
+                FLAGS_SET_Z(p->flags, 1);
+                FLAGS_SET_M(p->flags, 0);
+                break;
+            case M:
+                if (!FLAGS_GET_M(p->flags)) p->m = NAN;
+                FLAGS_SET_Z(p->flags, 0);
+                FLAGS_SET_M(p->flags, 1);
+                break;
+            case XY:
```
Yvinayak07: This case look redundant instead if we can place an if condition above where we can have early exit for `XY`.

contrib/babelfishpg_tds/src/backend/tds/tdstypeio.c
Mohitraj-g:
```
int
TdsSendTypeGeometry(FmgrInfo *finfo, Datum value, void *vMetaData)
{
    int TdsInstr = (int)INSTR_TDS_DATATYPE_GEOMETRY;
    return TdsSendSpatialHelper(finfo, value, vMetaData, TdsInstr);
}

int
TdsSendTypeGeography(FmgrInfo *finfo, Datum value, void *vMetaData)
{
    int TdsInstr = (int)INSTR_TDS_DATATYPE_GEOGRAPHY;
    return TdsSendSpatialHelper(finfo, value, vMetaData, TdsInstr);
}
```
yes these are the only conditions

contrib/babelfishpg_tds/src/backend/tds/tdstypeio.c
Mohitraj-g: we are using GSERIALIZED structure in `TdsTypeSpatialToDatum` so we will remove this when we support TDS receiver for linestring.

contrib/babelfishpg_common/src/geo.c
Mohitraj-g: Z and M can be NAN but only for internal use and we omit them while showing the result hence it's for Z and M only.

contrib/babelfishpg_common/src/geo.c
Mohitraj-g: added an early exit but still need this because when using a switch statement with an enum, there is a compile error if we don't handle all cases.

contrib/babelfishpg_tds/src/backend/tds/tdstypeio.c
Mohitraj-g: freeing buf leads to a server crash hence not freed.

contrib/babelfishpg_common/src/spatialtypes.h
Mohitraj-g: We are mainly using this on tds side so we don't want to expose more functions than required which are there in geo_data.h

contrib/babelfishpg_common/src/geo.c
Yvinayak07: can we use default in that case.

Mohitraj-g: Yes done

contrib/babelfishpg_common/sql/geography.sql
```diff
@@ -267,7 +267,30 @@ CREATE OR REPLACE FUNCTION sys.Geography__STPointFromText(sys.NVARCHAR, integer)
 		IF Geomtype = 'ST_Point' THEN
 			RETURN geom;
 		ELSE
-			RAISE EXCEPTION '% is not supported', Geomtype;
+			RAISE EXCEPTION 'Invalid function usage : Expected "POINT"';
```
rishabhtanwar29: Do we get the same as per TSQL? If not then we should throw the same as TSQL.

contrib/babelfishpg_common/src/spatialtypes.h
rishabhtanwar29, line 1: Nit: lets add header guards, i.e.:
```
#ifndef SPATIALTYPES_H
#define SPATIALTYPES_H
...
#endif SPATIALTYPES_H
```

contrib/babelfishpg_common/sql/geography.sql
Mohitraj-g: Added

Yvinayak07 (approved):

rishabhtanwar29 (approved):

PR #4036, opened 2025-08-22
Fix column-level permissions loss during ALTER VIEW operations

R4hul04:
### Description

This PR fixes an issue where column-level permissions were being lost when a view was modified using ALTER VIEW. While view-level ACLs were correctly preserved, column-level ACLs stored in pg_attribute.attacl were not being carried over to the new view definition.

#### Changes:
- Added a `ColumnAclInfo` structure to store column permission information
- `get_old_view_column_acls()` to retrieve column-level ACLs before view alteration
- `restore_view_column_acls()` to restore column-level ACLs after view recreation

### Issues Resolved

BABEL-5936

Signed off: Rahul Parande rparande@amazon.com
* **Use case based -**

* **Boundary conditions -**

* **Arbitrary inputs -**

* **Negative test cases -**

* **Performance tests -**

* **Tooling impact -**

* **Client tests -**

### Check List

### DIFF
--- a/contrib/babelfishpg_tsql/src/pl_handler.c
+++ b/contrib/babelfishpg_tsql/src/pl_handler.c
@@ -154,6 +154,8 @@ static void set_pgtype_byval(List *name, bool byval);
 static void pltsql_proc_get_oid_proname_proacl(AlterFunctionStmt *stmt, ParseState *pstate, Oid *oid, Acl **acl, bool *isSameFunc, bool is_proc);
 static Acl *get_old_view_acl(Oid oldViewOid);
 static void pg_class_update_acl(Oid newViewOid, Acl *oldViewAcl);
+static List *get_old_view_column_acls(Oid oldViewOid);
+static void restore_view_column_acls(Oid newViewOid, List *column_acls);
 static bool pltsql_truncate_identifier(char *ident, int len, bool warn);
 static Name pltsql_cstr_to_name(char *s, int len);
 extern void pltsql_add_guc_plan(CachedPlanSource *plansource);
@@ -193,8 +195,6 @@ static void bbf_set_tran_isolation(char *new_isolation_level_str);
 static void gen_command_grant_revoke_priv_to_role(StringInfo query, const char *rolename,
 							bool is_grant, Oid login_oid);
 
-static List *get_old_view_column_acls(Oid oldViewOid);
-static void restore_view_column_acls(Oid newViewOid, List *column_acls);
 typedef struct ColumnAclInfo
 {
     AttrNumber  attnum;    /* Column number */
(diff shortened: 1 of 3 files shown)


contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -5349,6 +5363,104 @@ pg_class_update_acl(Oid newViewOid, Acl *oldViewAcl)
...
+	
+	foreach(lc, column_acls)
+	{
+		ColumnAclInfo *info = (ColumnAclInfo *) lfirst(lc);
+		Relation pg_attribute_rel;
+		HeapTuple attTup;
+		AttrNumber attnum = InvalidAttrNumber;
+		
+		/* Find the corresponding column in the new view by name */
+		attnum = get_attnum(newViewOid, info->attname);
```
forestkeeper: You already have info->att_num, why do you have to get attnum again from another api ?

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -5349,6 +5363,104 @@ pg_class_update_acl(Oid newViewOid, Acl *oldViewAcl)
...
+		if (att->attnum <= 0)
+			continue;
+			
+		aclDatum = SysCacheGetAttr(ATTNAME, tuple, 
+								Anum_pg_attribute_attacl, &isNull);
+		
+		info = palloc(sizeof(ColumnAclInfo));
+		info->attnum = att->attnum;
+		info->attname = pstrdup(NameStr(att->attname));
+		info->acl = isNull ? NULL : DatumGetAclPCopy(aclDatum);
```
forestkeeper, line 5391: Let me ask : is attname information not in the acl Datum ?

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -5349,6 +5363,104 @@ pg_class_update_acl(Oid newViewOid, Acl *oldViewAcl)
...
+	
+	pg_attribute_rel = table_open(AttributeRelationId, AccessShareLock);
+	
+	ScanKeyInit(&skey,
+				Anum_pg_attribute_attrelid,
+				BTEqualStrategyNumber, F_OIDEQ,
+				ObjectIdGetDatum(oldViewOid));
+	
+	scan = systable_beginscan(pg_attribute_rel, AttributeRelidNumIndexId,
+							true, NULL, 1, &skey);
```
tanscorpio7: Can we use `SearchSysCacheList1(ATTNUM ....` instead of systable scan. `SysCacheGetAttr` should be used on sys cache tuple.

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -5349,6 +5363,104 @@ pg_class_update_acl(Oid newViewOid, Acl *oldViewAcl)
...
+	while (HeapTupleIsValid(tuple = systable_getnext(scan)))
+	{
+		Form_pg_attribute att = (Form_pg_attribute) GETSTRUCT(tuple);
+		Datum aclDatum;
+		bool isNull;
+		ColumnAclInfo *info;
+		
+		/* Skip system columns */
+		if (att->attnum <= 0)
+			continue;
```
tanscorpio7: `InvalidAttrNumber` instead of 0

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -5349,6 +5363,104 @@ pg_class_update_acl(Oid newViewOid, Acl *oldViewAcl)
...
+			
+		aclDatum = SysCacheGetAttr(ATTNAME, tuple, 
+								Anum_pg_attribute_attacl, &isNull);
+		
+		info = palloc(sizeof(ColumnAclInfo));
```
tanscorpio7: (ColumnAclInfo *) palloc0(sizeof(ColumnAclInfo)), new nodes should always be initialised with all zeroes and explicit type cast if type is a pointer.

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -71,6 +71,7 @@
 #include "utils/plancache.h"
 #include "utils/ps_status.h"
 #include "utils/queryenvironment.h"
+#include "utils/fmgroids.h"
```
tanscorpio7: Curious what required this include, this file is a generated file and only contains build int functions oids, did not see any use of them in this PR.

tanscorpio7 (changes requested):

contrib/babelfishpg_tsql/src/pl_handler.c
R4hul04: Removing this line as we don't need `attnum` because anyway we search the column by name. I missed removing this line. My thought process previously I was using attnum (position) + colname to match old view col to new view col.

I was trying to find the attribute number (column position) in the new view based on the column name and attnum from the old view. The reason for this was:

1. `info->attnum` contains the column position in the original/old view
2. `attnum` we're retrieving here is meant to be the corresponding column position in the new view

When altering a view, the new definition might:
- Reorder columns
- Remove existing columns
- Add new columns

Because of these possibilities, a column named "x" might be at position 1 in the old view but position 3 in the new view.
But now using only col name instead of attnum as anyhow goal is to match old and new col which can be done by its col name

contrib/babelfishpg_tsql/src/pl_handler.c
R4hul04:
```
typedef struct AclItem
{
    Oid     ai_grantee;    /* User/role receiving permission */
    Oid     ai_grantor;    /* User/role granting permission */
    AclMode ai_privs;      /* Permission bitmask */
} AclItem;
```
It does not contain column name information. System maintains relationship between ACLs and objects by storing them in catalog table (pg_attribute.attacl for columns), but the ACL itself doesn't know which object it belongs to.

contrib/babelfishpg_tsql/src/pl_handler.c
R4hul04: The include is required for the F_OIDEQ macro (OID 184) used in the ScanKeyInit() call when setting up the pg_attribute index scan in `get_old_view_column_acls()`

```
	ScanKeyInit(&skey,
				Anum_pg_attribute_attrelid,
				BTEqualStrategyNumber, F_OIDEQ,
				ObjectIdGetDatum(oldViewOid));
```

contrib/babelfishpg_tsql/src/pl_handler.c
R4hul04: Changed to 
`catlist = SearchSysCacheList1(ATTNUM, ObjectIdGetDatum(oldViewOid));`
`aclDatum = SysCacheGetAttr(ATTNUM, tuple, Anum_pg_attribute_attacl, &isNull);`

contrib/babelfishpg_tsql/src/pl_handler.c
R4hul04: Removed this include as we are using `SearchSysCacheList1` now which doesn't require `F_OIDEQ`

test/JDBC/expected/alter-view-acl.out
```diff
@@ -181,6 +233,9 @@ AS
     SELECT id, name FROM test_schema.base_table;
 GO
 
+ALTER VIEW test_schema.colacl_view1 AS SELECT 2 AS a;
+GO
+
```
tanscorpio7, line 518: Let us also test some cases where 
1. View is created with 3 columns and then altered to have 4+ columns. The new 4 columns can include same name column as previous view, different name column, same name but different case column from before. Then we can check final permission again. 
2. We can also repeat this test case with the same ALTER VIEW command but this time with explicit column aliases.
3. Please also print the before and after output of privileges on column using system views i.e. information_schema.

tanscorpio7 (changes requested):

tanscorpio7 (approved):

forestkeeper (approved):
LGTM

PR #3739, opened 2025-05-07
Reimplement ISNUMERIC() function in C language to improve performance

sumitj824:
### Description
To improve the performance of queries with ISNUMERIC() predicates, this commit 
reimplements the ISNUMERIC() function from PL/pgSQL to C language. The new implementation 
uses InputFunctionCallSafe for safer and more efficient type checking and conversion. 

Additionally, the money data type's input function has been updated to report 
invalid input as a "soft" error for InputFunctionCallSafe, avoiding unnecessary transaction aborts. 
This commit also changes the ISNUMERIC() function attributes to IMMUTABLE and 
PARALLEL SAFE for better query optimization opportunities.

Signed-off-by: Sumit Jaiswal <sumiji@amazon.com>

### Issues Resolved

Task : BABEL-5696

### Check List
- [x] Commits are signed per the DCO using --signoff 

By submitting this pull request, I confirm that my contribution is under the terms of the Apache 2.0 and PostgreSQL licenses, and grant any person obtaining a copy of the contribution permission to relicense all or a portion of my contribution to the PostgreSQL License solely to contribute all or a portion of my contribution to the PostgreSQL open source project.

For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/babelfish-for-postgresql/babelfish_extensions/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #4037, opened 2025-08-24
Re-Implementing Isdate() function 

pranavJ23:
### Description
Re-Implementing the ISDATE() function for babelfish.
1. Casting the input type for isdate() function to datetime instead of date datatype
2. Added support for UDT handling.
3. Added sufficient checks to allow only datetime, smalldatetime as valid input datetime datatypes(input in quotes is also valid as per isdate() function definition).
4. Added handling to ensure that extra/invalid delimiters should not be ignored and should be treated as invalid inputs

### Issues Resolved

[BABEL-970]

1. Basic isdate() tests
2. Test with UDT 
3. Test with different datatypes like datetime2, datetimeoffset
4. NULL input test
5. Empty input test
6. Unknown datatype(eg input :'10/10/10') test
7. Test when input syntax is not correct(eg: '10/10/10/') i.e excess delimiter etc.
8. Using isdate() function over a table column.

### Check List


contrib/babelfishpg_tsql/sql/sys_functions.sql
```diff
@@ -1637,15 +1637,65 @@ $$
...
+    basetype oid;
+begin
+
+    arg_datatype_oid := pg_typeof(v)::oid;
+    arg_datatype := sys.translate_pg_type_to_tsql(arg_datatype_oid);
+
+    IF arg_datatype IS NULL THEN
+        basetype := sys.bbf_get_immediate_base_type_of_UDT(arg_datatype_oid);
+        arg_datatype := sys.translate_pg_type_to_tsql(basetype);
+    END IF;
```
Anikait143: Could arg_datatype still be `NULL` after this?

contrib/babelfishpg_tsql/sql/sys_functions.sql
```diff
@@ -1637,15 +1637,65 @@ $$
...
+
+    IF arg_datatype IN ('date','time','datetime2','datetimeoffset','text','ntext','image') THEN
+        RAISE EXCEPTION USING 
+        ERRCODE = 'invalid_parameter_value',
+        MESSAGE = format('Argument data type %s is invalid for argument 1 of ISDATE function.', arg_datatype);
+    END IF;
```
Anikait143: Shouldn't we return `0` in such case?

contrib/babelfishpg_tsql/sql/sys_functions.sql
```diff
@@ -1637,15 +1637,65 @@ $$
...
+    IF NOT (arg_datatype IN ('datetime', 'smalldatetime','varchar','sys.varchar','char','nchar')) THEN
+        return 0;
+    END IF;
+
+    if length(v::sys.varchar) = 0 then
+        return 0;
+    end if;
+
+    if v is NULL THEN
+        return 0;
```
Anikait143: This should be the very first check in the function

contrib/babelfishpg_tsql/sql/sys_functions.sql
```diff
@@ -1637,15 +1637,65 @@ $$
 STRICT
 LANGUAGE plpgsql;
 
-create or replace function sys.isdate(v text)
+create or replace function sys.isdate(IN v anyelement)
```
Anikait143: what would happen in case of cstring input?

contrib/babelfishpg_tsql/sql/sys_functions.sql
```diff
@@ -1637,15 +1637,65 @@ $$
...
+        return 1;
+    end if;
+    EXCEPTION 
+        WHEN invalid_parameter_value THEN
+            RAISE;
+        WHEN others THEN
+            RETURN 0;
+end
+$body$
+language 'plpgsql' STABLE;
```
Anikait143: This should be marked as `IMMUTABLE PARALLEL SAFE`

contrib/babelfishpg_tsql/sql/sys_functions.sql
```diff
@@ -1637,15 +1637,65 @@ $$
...
+
+create or replace function sys.isdate(IN v sys.varchar)
```
Anikait143: Why do we need a separate function for varchar? This can be handled in ANYELEMENT definition.

contrib/babelfishpg_tsql/sql/sys_functions.sql
```diff
@@ -1654,6 +1704,18 @@ end
...
+returns integer as
+$body$
+begin
+    RAISE EXCEPTION USING 
+    ERRCODE = 'invalid_parameter_value',
+    MESSAGE = 'Argument data type (n)text is invalid for argument 1 of ISDATE function.';
+    return 0;
+end;
+$body$
+language plpgsql stable;
```
Anikait143: This should be marked as `IMMUTABLE PARALLEL SAFE`

test/python/expected/upgrade_validation/expected_dependency.out
```diff
@@ -522,6 +522,8 @@ Function sys.is_rolemember(sys.sysname,sys.sysname)
 Function sys.is_rolemember_internal(sys.sysname,sys.sysname)
 Function sys.is_srvrolemember(sys.sysname,sys.sysname)
 Function sys.is_table_type(oid)
+Function sys.isdate(anyelement)
+Function sys.isdate(sys."varchar")
```
Anikait143: Let's add dependency tests

contrib/babelfishpg_common/src/datetime.c
Anikait143, line 1: Why do we need these changes as part of this fix? Have we considered wide scope use?

contrib/babelfishpg_tsql/sql/sys_functions.sql
pranavJ23: Tsql throw error instead of error in such case, so I throw error to be consistent.

contrib/babelfishpg_tsql/sql/sys_functions.sql
pranavJ23: Right, will change it

contrib/babelfishpg_tsql/sql/sys_functions.sql
pranavJ23: In case of a string eg: '10-10-10' we want to check if it's a valid date and then returns 0/1 accordingly where as in the case eg: select isdate(cast('10-10-10' as text)) we will simply throw error as the input datatype is text.
So to differentiate between the both we need sys.varchar function separately so that if the input datatype is unknown then we should hit function with sys.varchar as input  not text as input

contrib/babelfishpg_common/src/datetime.c
pranavJ23: We are implementing these changes in the case of datetime.
Looking into the Tsql results if we have delimiter at the end, start or consecutive delimiter then we should throw error eg:

```
select cast('10/10/10/' as datetime)
> throw error
select cast('10/10//10' as datetime)
> throw error
select cast('/10/10/10' as datetime)
> throw error
```

where as currently in babelfish we just ignore these extra delimiters and give output.
Following Tsql behaviour we should restrict it and throw error.

contrib/babelfishpg_tsql/sql/sys_functions.sql
pranavJ23: it will hit the isdate() function with input type as sys.varchar

contrib/babelfishpg_tsql/sql/sys_functions.sql
pranavJ23: I don't think that it will be null since we checked if UDT then converting to basetype.

contrib/babelfishpg_tsql/sql/sys_functions.sql
```diff
@@ -1684,22 +1684,88 @@ $$
...
+        RAISE EXCEPTION USING 
+        ERRCODE = 'invalid_parameter_value',
+        MESSAGE = format('Argument data type %s is invalid for argument 1 of ISDATE function.', arg_datatype);
+    END IF;
+
+    IF NOT (arg_datatype IN ('datetime', 'smalldatetime','varchar','sys.varchar','char','nchar')) THEN
+        return 0;
+    END IF;
+
+    if length(v::sys.varchar) = 0 then
```
rohit01010: lets prepend `pg_catalog` before length function.

test/JDBC/expected/babel_function.out
```diff
@@ -484,6 +484,79 @@ int
...
+go
+~~ERROR (Code: 33557097)~~
+
+~~ERROR (Message: syntax error near '10' at line 1 and character position 14)~~
+
+select isdate(10@20@12);
+go
+~~ERROR (Code: 33557097)~~
+
+~~ERROR (Message: syntax error near '10' at line 1 and character position 14)~~
```
rohit01010: Shouldn't the argument in quotes ?

test/JDBC/expected/Datetime_system_functions.out
```diff
@@ -9060,10 +9060,9 @@ BEGIN CATCH
     SELECT 'PASS: ISDATE rejected multiple parameters' AS MultipleArgumentsTest;
 END CATCH
 GO
-~~START~~
-varchar
-PASS: ISDATE rejected multiple parameters
-~~END~~
+~~ERROR (Code: 33557097)~~
+
+~~ERROR (Message: The isdate function requires 1 argument)~~
```
rohit01010, line 9065: what was the error returned earlier and what was its error code for following query ?
```
SELECT ISDATE('2023-01-01', 'extra_arg');
```

contrib/babelfishpg_common/src/datetime.c
```diff
@@ -312,11 +331,23 @@ clean_input_str(char *str, bool *contains_extra_spaces, DateTimeContext context)
...
 
 		last_non_space = i;
+		last_char = str[i];
 		i++;
 	}
 
 	result[j] = '\0';
 
+	/* Additional validation checks */
+	if (context == DATE_TIME && (has_trailing_delimiter ||
```
rohit01010: why do we only need this validation only for `DATE_TIME` and not for `DATE_TIME2` ?

contrib/babelfishpg_tsql/sql/sys_functions.sql
```diff
@@ -1684,22 +1684,88 @@ $$
...
+
+create or replace function sys.isdate(IN v sys.varchar)
+returns integer
+as
+$body$
+begin
+
+    if length(v::sys.varchar) = 0 then
+        return 0;
+    end if;
```
rohit01010: v is already of type sys.varchar, no need of casting it again to sys.varchar.

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--5.3.0--5.4.0.sql
```diff
@@ -86,6 +86,109 @@ END
...
+returns integer as
+$body$
+begin
+    RAISE EXCEPTION USING 
+    ERRCODE = 'invalid_parameter_value',
+    MESSAGE = 'Argument data type (n)text is invalid for argument 1 of ISDATE function.';
+    return 0;
+end;
+$body$
+language plpgsql IMMUTABLE PARALLEL SAFE;
```
rohit01010: We need to deprecate and recreate this as this will lead to errors in existing views which are dependent on `sys.isdate(v text)`.

test/JDBC/upgrade/latest/schedule
```diff
@@ -644,3 +644,4 @@ babel_numeric_fixeddecimal_oper
 babel_numeric_money_oper
 babel_numeric_smallmoney_oper
 datetime_roundoff
+babel_datetimeoffset
```
rohit01010, line 653: Lets add this in other schedule files as well

rohit01010 (changes requested):

contrib/babelfishpg_common/src/datetime.c
pranavJ23: Thanks for pointing out. It can be done for datetime2 as well, since isdate is casted to datetime so did it for datetime only as of now.

contrib/babelfishpg_tsql/sql/sys_functions.sql
Deepesh125: Since we have already introduced separate function to handle TEXT type, I am not sure if we really want one separate function with sys.varchar. If you are concerned about string literal input like '10-10-10' then it should have been take care somewhere in the parser and should have appropriate node where we treat it as sys.varchar and hence anyelement will catch it. So seems like we can get rid of this sys.varchar specific function. please correct me if I am wrong. 

@rohit01010 should be aware about handling of string literal.

Deepesh125 (changes requested):
This is similar functionality to ISNUMERIC. For ISNUMERIC, we experienced performance related issue when we implemented it as a plpgsql function initially and later we had to re-work to implement it in C function (check https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/3739/files) 

Have we considered any of performance implication side before choosing to implement in plpgsql?

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -2289,7 +2290,76 @@ isnumeric(PG_FUNCTION_ARGS)
...
+
+    /* Sanity check */
+    if (!OidIsValid(argtypeid))
+        ereport(ERROR,
+                (errcode(ERRCODE_INVALID_PARAMETER_VALUE),
+                 errmsg("could not determine input data type")));
+
+    /* Check for invalid types */
+    if (argtypeid == DATEOID ||
+        argtypeid == TIMEOID ||
```
tanyagupta17: Could you please also add handling for geometry/geography/sql_variant/xml, they don't seem to be supported either.

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -2289,7 +2290,76 @@ isnumeric(PG_FUNCTION_ARGS)
...
 
+Datum
+isdate(PG_FUNCTION_ARGS)
+{
+    Oid         argtypeid;
+    char        *value_str;
+    bool        result = false;
+
+    if (PG_ARGISNULL(0))
+        PG_RETURN_INT32(result ? 1 : 0);
```
tanyagupta17: Can't we simply return 0 here, as result will always be false in this step?

contrib/babelfishpg_tsql/runtime/functions.c
pranavJ23: For this I will be applying a check if not datetime/smalldatetime/(n)char/(n)varchar

contrib/babelfishpg_tsql/runtime/functions.c
pranavJ23: C implementation function is still under progress, since implementation shifted from SQL to C recently.
Will send it for review once done with the changes

test/JDBC/expected/Datetime_system_functions.out
pranavJ23:
```
The isdate function requires 1 argument
```
earlier this was the error

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -2296,7 +2297,92 @@ isnumeric(PG_FUNCTION_ARGS)
...
+	{
+		Oid typoutput;
+		bool typisvarlena;
+		getTypeOutputInfo(argtypeid, &typoutput, &typisvarlena);
+		value_str = OidOutputFunctionCall(typoutput, PG_GETARG_DATUM(0));
+	}
+
+
+	/* Check for empty string */
+	if (strlen(value_str) == 0)
```
tanyagupta17: Would this also take care of whitespaces or char which by default has 30 as its typmod, expected results:
```
1> select isdate('     ')
2> go
           
-----------
          0

```

```
1> select isdate(cast('' as char))
2> go
           
-----------
          0
```
You can use this reference [PR](https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/4094/files#diff-14053c36507852e9e6bec9aaf6879220d4e22efea701990e9bfc5ac6289523d4R2239) for this handling

test/JDBC/jdbc_schedule
```diff
@@ -636,3 +636,6 @@ ignore#!#view_sec_setting_before_17_6-or-16_10-vu-cleanup
 ignore#!#mathematical_func-before_17_7-or-16_11-vu-prepare
 ignore#!#mathematical_func-before_17_7-or-16_11-vu-verify
 ignore#!#mathematical_func-before_17_7-or-16_11-vu-cleanup
+ignore#!#babel_datetimeoffset_before_16_10-or-17_6-vu-prepare
+ignore#!#babel_datetimeoffset_before_16_10-or-17_6-vu-verify
+ignore#!#babel_datetimeoffset_before_16_10-or-17_6-vu-cleanup
```
tanyagupta17: NIT: Please add an extra line here

contrib/babelfishpg_common/src/datetime.c
```diff
@@ -312,11 +331,23 @@ clean_input_str(char *str, bool *contains_extra_spaces, DateTimeContext context)
...
+	/* Additional validation checks */
+	if ((context == DATE_TIME_2 || context == DATE_TIME) && (has_trailing_delimiter ||
+		has_leading_delimiter ||
+		consecutive_delimiters))
+	{
+		pfree(result);
+		ereport(ERROR,
+				(errcode(ERRCODE_INVALID_DATETIME_FORMAT),
+				errmsg("invalid input syntax for type %s: \"%s\"", context_str, str)));
+	}
```
rohit01010, line 353: Why do we need check on context here ? are there any cases where different value of context can have `trailing_delimiter` or `leading_delimiter` or `consecutive_delimiters` ?

contrib/babelfishpg_common/src/datetime.c
```diff
@@ -210,10 +218,21 @@ clean_input_str(char *str, bool *contains_extra_spaces, DateTimeContext context)
 					(errcode(ERRCODE_INVALID_DATETIME_FORMAT),
 					errmsg("invalid input syntax for type %s: \"%s\"", context_str, str)));
 		}
+
+		/* Check for consecutive delimiters */
+		if ((str[i] == '/' || str[i] == '-' || str[i] == '.') &&
+			(last_char == '/' || last_char == '-' || last_char == '.'))
+		{
+			consecutive_delimiters = true;
+		}
```
rohit01010, line 229: should we break out of the loop here ? Or can we just throw error directly here instead of processing the rest of the string ?

contrib/babelfishpg_common/src/datetime.c
```diff
@@ -183,9 +183,17 @@ clean_input_str(char *str, bool *contains_extra_spaces, DateTimeContext context)
 	int i = 0, j = 0;
 	int last_non_space = -1;
 	int num_colons = 0;
+	bool has_trailing_delimiter = false;
+	bool has_leading_delimiter = false;
+	bool consecutive_delimiters = false;
```
rohit01010, line 189: Are these scenario's sufficient ? can you please test following query
```
select isdate(':12/12/12')
go
```

contrib/babelfishpg_common/src/datetime.c
```diff
@@ -312,11 +331,23 @@ clean_input_str(char *str, bool *contains_extra_spaces, DateTimeContext context)
...
 	}
 
 	result[j] = '\0';
 
+	/* Additional validation checks */
+	if ((context == DATE_TIME_2 || context == DATE_TIME) && (has_trailing_delimiter ||
+		has_leading_delimiter ||
+		consecutive_delimiters))
+	{
+		pfree(result);
```
rohit01010: lets add a sanity check 
```
if (result)
	pfree(result);
```

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -2296,7 +2297,92 @@ isnumeric(PG_FUNCTION_ARGS)
...
+		(*common_utility_plugin_ptr->is_tsql_varchar_datatype) (argtypeid) ||
+		(*common_utility_plugin_ptr->is_tsql_nchar_datatype) (argtypeid) ||
+		(*common_utility_plugin_ptr->is_tsql_bpchar_datatype) (argtypeid) ||
+		 argtypeid == VARCHAROID || argtypeid == BPCHAROID || argtypeid == INT4OID))
+	{
+		ereport(ERROR,
+				(errcode(ERRCODE_INVALID_PARAMETER_VALUE),
+				 errmsg("Argument data type %s is invalid for argument 1 of ISDATE function.",
+						format_type_be(argtypeid))));
+	}
```
rohit01010, line 2385: Can you please point me to the tests added for isdate function called on all the datatypes.

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -2296,7 +2297,92 @@ isnumeric(PG_FUNCTION_ARGS)
...
+
+	/* Get the string representation from input datum */
+	if (argtypeid == VARCHAROID || argtypeid == BPCHAROID ||
+		(*common_utility_plugin_ptr->is_tsql_nvarchar_datatype) (argtypeid) ||
+		(*common_utility_plugin_ptr->is_tsql_varchar_datatype) (argtypeid) ||
+		(*common_utility_plugin_ptr->is_tsql_nchar_datatype) (argtypeid) ||
+		(*common_utility_plugin_ptr->is_tsql_bpchar_datatype) (argtypeid))
+		value_str = text_to_cstring(PG_GETARG_TEXT_PP(0));
```
rohit01010: Currently nvarchar/nchar/varchar/bpchar/text all are created using varlena, so using `PG_GETARG_TEXT_PP(0)` will work for now, but can it cause any issue in future or should we be using different functions `PG_GETARG_VARCHAR_PP` or `PG_GETARG_BPCHAR_PP` for different argtypeids ?

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -2296,7 +2297,92 @@ isnumeric(PG_FUNCTION_ARGS)
...
+		bool typisvarlena;
+		getTypeOutputInfo(argtypeid, &typoutput, &typisvarlena);
+		value_str = OidOutputFunctionCall(typoutput, PG_GETARG_DATUM(0));
+	}
+
+
+	/* Check for empty string */
+	if (strlen(value_str) == 0)
+	{
+		pfree(value_str);
```
rohit01010: NIT: Lets add a sanity check that value_str is not NULL before freeing it.

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -2296,7 +2297,92 @@ isnumeric(PG_FUNCTION_ARGS)
...
+		result = true;
+	}
+	PG_CATCH();
+	{
+		result = false;
+		FlushErrorState();
+	}
+	PG_END_TRY();
+
+	pfree(value_str);
```
rohit01010: NIT: Lets add a sanity check that value_str is not NULL before freeing it.

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--5.3.0--5.4.0.sql
```diff
@@ -33,6 +34,7 @@ end
 $$
 LANGUAGE plpgsql;
 
+
```
rohit01010: NIT: unncessary change

PR #4057, opened 2025-09-03
Support Same-DB Ownership Chaining for object references inside Procedure/Functions

HarshLunagariya:
### Description

This commit implements the ownership chaining support for object
references inside function/procedure. The basic idea of Ownership chaining is that when
one object references another, and both have the same owner, then permissions
are only checked when the top-level object is accessed.

Core Functionality:
 Added support for ownership chaining within the same database for procedures and functions

Key Components Modified:
1. Added ownership chain validation in permission checks
2. Walker function to mark relations and functions inside view definitions

Implementation:

Walker function to mark relations and functions inside view definitions

Walks through the query parse tree to:
1. Mark relations and functions as being inside a view context
2. For relations:
   - Set checkAsUser to the view_owner when it matches the relation's owner,
     enabling permission checking to pass at the executor stage (ownership chaining)
3. For procedures/functions:
   - Store the view_owner in the parentOwnerId field to support
     procedure/function-specific ownership chaining logic
     during permission checks at the executor stage

### Issues Resolved

* **Use case based -**

* **Boundary conditions -**

* **Arbitrary inputs -**

* **Negative test cases -**

* **Performance tests -**

* **Tooling impact -**

* **Client tests -**

### Check List


contrib/babelfishpg_tsql/src/pltsql_utils.c
```diff
@@ -2983,3 +2984,99 @@ downcase_truncate_split_object_name(char *four_part_object_name, char **server_n
...
+ */
+Oid
+get_rel_owner(Oid relid)
+{
+    HeapTuple   tuple;
+    Oid         owner;
+
+    /* Get relation tuple from pg_class */
+    tuple = SearchSysCache1(RELOID, ObjectIdGetDatum(relid));
+    if (!HeapTupleIsValid(tuple))
```
tanscorpio7, line 3001: Let us add asserts for input not being InvalidOid. Might be needed in other functions as well.

Also let's evaluate if storing the function and relation owners inside a new field in parse nodes strcutures will help performance.

@HarshLunagariya would also like to know your thoughts on this.

contrib/babelfishpg_tsql/src/pltsql_utils.c
```diff
@@ -2983,3 +2984,99 @@ downcase_truncate_split_object_name(char *four_part_object_name, char **server_n
...
+ */
+static Oid
+get_current_func_owner(void)
+{
+
+	if (sql_dialect != SQL_DIALECT_TSQL)
+		return InvalidOid;
+
+	if (pltsql_non_tsql_proc_entry_count > 0 || pltsql_sys_func_entry_count > 0)
+		return InvalidOid;
```
tanscorpio7: Let use `pltsql_support_tsql_transactions()` instead

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -1039,13 +1039,27 @@ pltsql_ExecFuncProc_AclCheck(Oid funcid)
 			 */
 			if (nspname != NULL && !is_shared_schema(nspname) &&
 				!is_schema_from_db(schema_id, get_cur_db_id()))
+			{
 				userid = GetSessionUserId();
+			}
+			else
+			{
```
tanscorpio7, line 1046: This block only has check for `IS_TDS_CLIENT() && !InSecurityRestrictedOperation()`. Let's find out the why there is not dialect check here ? based on the answer we might have to remove dialect checks from all functions being used inside this block.

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -1039,13 +1039,27 @@ pltsql_ExecFuncProc_AclCheck(Oid funcid)
...
+				/*
+				 * Ownership Chaining Logic for object references inside function/procedures.
+				 * Only applicable for same-db ownership chaining cases.
+				 */
+				if(IsA(expr, FuncExpr) &&
+				   is_valid_func_ownership_chain(expr, get_func_owner(((FuncExpr *)expr)->funcid)))
+				{
+					return ACLCHECK_OK;
+				}
+			}
```
tanscorpio7, line 1061: I think this we should first check if permission are enough without ownership chaining. If we do not have the required permission, only then enter ownership change code. This will reduce the additional sys cache lookups.

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -7919,4 +7922,92 @@ gen_command_grant_revoke_priv_to_role(StringInfo query, const char *rolename,
...
+static void
+mark_outside_view(Query *query)
+{
+
+	if (!IS_TDS_CLIENT() || InSecurityRestrictedOperation())
+		return;
+
+	mark_outside_view_ref_walker((Node *)query, NULL);
+}
+
```
tanscorpio7: Should all of this code go into pltsql_permissions.c ?

tanscorpio7 (changes requested):

contrib/babelfishpg_tsql/src/pltsql_utils.c
HarshLunagariya: > Also let's evaluate if storing the function and relation owners inside a new field in parse nodes strcutures will help performance.

For each parsenode, Owner will be fetched once at the time of permission check. So I don't there will be much of a performance gain.

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -1039,13 +1039,27 @@ pltsql_ExecFuncProc_AclCheck(Oid funcid)
...
+			else
+			{
+				/*
+				 * Ownership Chaining Logic for object references inside function/procedures.
+				 * Only applicable for same-db ownership chaining cases.
+				 */
+				if(IsA(expr, FuncExpr) &&
+				   is_valid_func_ownership_chain(expr, get_func_owner(((FuncExpr *)expr)->funcid)))
+				{
+					return ACLCHECK_OK;
```
Deepesh125, line 1059: free nspname

contrib/babelfishpg_tsql/src/pltsql_utils.c
```diff
@@ -2983,3 +2984,99 @@ downcase_truncate_split_object_name(char *four_part_object_name, char **server_n
...
+	* Fetch the top procedure excution state from execution state call stack
+	* and get the owner of that procedure. Top entry in stack will have
+	* fn_oid and fn_owner value set.
+	*/
+	if (!exec_state_call_stack ||
+		!exec_state_call_stack->estate ||
+		!exec_state_call_stack->estate->func)
+		return InvalidOid;
+
+	return exec_state_call_stack->estate->func->fn_oid;
```
Deepesh125, line 3057: is this really function onwer? I think we need to update function name

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -1112,14 +1126,18 @@ pltsql_ExecutorStart(QueryDesc *queryDesc, int eflags)
...
-						if (OidIsValid(perminfo->checkAsUser))
+						if (OidIsValid(perminfo->checkAsUser) && !isTempNamespace(schema_id))
 						{
 							Oid loginId = get_login_for_user(perminfo->checkAsUser, nspname);
 							if (OidIsValid(loginId))
 								perminfo->checkAsUser = loginId;
 						}
 						else
 							perminfo->checkAsUser = GetSessionUserId();
+						if (is_valid_func_ownership_chain(perminfo, get_rel_owner(relOid)))
```
Deepesh125: nit: get_rel_owner can be reused

contrib/babelfishpg_tsql/src/pltsql_utils.c
tanscorpio7: Got it.
@pranavJ23 Lets validate the for SELECT f() from table, this permission check code path is only called once and not for each row.

contrib/babelfishpg_tsql/src/hooks.c
pranavJ23: We need to have a dialect check where ever we are trying to (re)parse, execute a query in t-sql syntax, since we are checking the user permission when we are connected with bbf client(i.e tds_client) so the IS_TDS_CLIENT() check is sufficient since regardless of dialect we need to do permission check.
Yes we should remove the dialect check from functions called from here since we don't need tsql check(for performing parsing/ executing any tsql query).

contrib/babelfishpg_tsql/src/hooks.c
pranavJ23: right, will add it

contrib/babelfishpg_tsql/src/pltsql_utils.c
pranavJ23: addressed

contrib/babelfishpg_tsql/src/hooks.c
pranavJ23: addressed

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -1039,13 +1039,32 @@ pltsql_ExecFuncProc_AclCheck(Oid funcid)
...
 			if (nspname != NULL && !is_shared_schema(nspname) &&
 				!is_schema_from_db(schema_id, get_cur_db_id()))
+			{
 				userid = GetSessionUserId();
+			}
+			else
+			{
+				/* If user already has the permission then return */
+				if (object_aclcheck(ProcedureRelationId, funcid, userid, ACL_EXECUTE) == ACLCHECK_OK)
+					return ACLCHECK_OK;
```
Deepesh125, line 1049: free nspname

contrib/babelfishpg_tsql/src/pltsql_utils.c
```diff
@@ -2983,3 +2984,101 @@ downcase_truncate_split_object_name(char *four_part_object_name, char **server_n
...
+}
+
+extern bool
+is_valid_func_ownership_chain(void *expr, Oid objectOwnerId)
+{
+	Oid top_func = InvalidOid;
```
Deepesh125: I think we should avoid top function, it may confuse readers. For example, f1 -> f2 -> f3 then f1 might be confused as top function while checking f3. Somthing like immediate_parent_func would make more sense

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -1039,13 +1039,32 @@ pltsql_ExecFuncProc_AclCheck(Oid funcid)
...
+				/*
+				 * Ownership Chaining Logic for object references inside function/procedures.
+				 * Only applicable for same-db ownership chaining cases.
+				 */
+				if (IsA(expr, FuncExpr) &&
+				   is_valid_func_ownership_chain(expr, get_func_owner(((FuncExpr *)expr)->funcid)))
```
Deepesh125, line 1055: IIUC, there will be two cases here
1. Function which are marked inside view. 
2. Function which are marked outside view 

For function marked outside of view, I can understand how will it behave. But what about function marked inside view and has a parent owner set? How are we checking permission in that case?

Deepesh125 (changes requested):

Deepesh125 (dismissed):

contrib/babelfishpg_tsql/src/hooks.c
tanscorpio7: @pranavJ23 Let us push `get_func_owner` to inside `is_valid_func_ownership_chain`. In many cases `is_valid_func_ownership_chain` will exit even before requiring the owner info. We should only do sys cache lookup for owner when we really need it.

tanscorpio7 (changes requested):

tanscorpio7 (approved):

Deepesh125 (approved):

PR #3934, opened 2025-07-04
Implement bidirectional cast/convert for char/nchar ↔ binary/varbinary

bilalkah:
### Description

This PR introduces support for casting/converting between `char`/`nchar` types and `binary`/`varbinary`, improving compatibility with T-SQL-style conversions in Babelfish.

1. **What is the change?**  
Previously, Babelfish did not support direct casts or converts between `char`/`nchar` and `binary`/`varbinary` types.  
With this change, the system can now handle such casts bidirectionally, enabling more seamless T-SQL compatibility for users.

2. **Why was the change made?**  
This change was driven by the need to increase compatibility with T-SQL’s casting behavior, as part of issue [#3835](https://github.com/babelfish-for-postgresql/babelfish_extensions/issues/3835).  
T-SQL workloads often rely on implicit and explicit casts between string and binary types, so supporting these casts is essential for broader T-SQL compatibility.

3. **How was the code changed?**  
- `babelfish_conv_helper_to_nvarchar` covers conversions to `nvarchar`. 
- Implemented and registered cast functions from `nchar` to `binary`/`varbinary`.  
- Implemented and registered cast functions from `binary`/`varbinary` to `char`/`nchar`.  
- Adjust `bpchar`/`varchar` to `binary` cast functions and use encoding.
- Updated conversion helper logic.  
- Added test cases covering each cast direction.
- Add missing support for ```TRY_CAST```, ```CONVERT``` and ```TRY_CONVERT``` for ```(nvar)char```-```var(binary)```.

---

### Issues Resolved

- [#3835](https://github.com/babelfish-for-postgresql/babelfish_extensions/issues/3835)

---

### Test Scenarios Covered

- **Use case based**  
  Casting between `nchar`, `char`, `binary`, and `varbinary` using realistic data inputs, validating correctness and encoding behavior.

- **Boundary conditions**  
  Tests include max-length inputs for each data type and edge-case conversions (e.g., empty strings, NULLs).

- **Arbitrary inputs**  
  Casts with uncommon but valid byte sequences and multi-byte characters were tested.

- **Negative test cases**  
  Tested unsupported or invalid conversions (e.g., improperly encoded binaries) to confirm expected error handling.

---

### Check List

- [x] Commits are signed per the DCO using --signoff

[...]

PR #4058, opened 2025-09-03
datetime2fromparts p_fractions to allow ranges between 32786 and 9999999

bhutraraghav:
### Description

Function DATETIME2FROMPARTS gives the following error when value of p_fractions is passed exceeding SMALLINT range

```SELECT DATETIME2FROMPARTS(2026, 6, 17, 13, 5, 40, 39767, 5)```

```SQL Error [33557097] [S0001]: Error while trying to cast to SMALLINT data type.```

### Check List
- [yes] Commits are signed per the DCO using --signoff 

### CODE REVIEW

**Comment by bhutraraghav** [comment]
@rohit01010 can you take a look at this pr?

File: contrib/babelfishpg_tsql/sql/sys_functions.sql [hunk 1]
```diff
@@ -482,7 +482,7 @@ BEGIN
 
    v_calc_seconds := pg_catalog.format('%s.%s',
                             floor(p_seconds)::SMALLINT,
-                            substring(rpad(lpad(v_fractions, v_precision, '0'), 7, '0'), 1, v_precision))::NUMERIC;
+                            substring(rpad(lpad(v_fractions, v_precision, '0'), 7, '0'), 1, LEAST(v_precision, 6)))::NUMERIC;
```
rohit01010: why do we need this change ? Can you provide any test case where the results have been fixed due to this change ?

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.2.0--2.3.0.sql
rohit01010, line 1: we should not modify any of the previous upgrade script. Adding the modified function definition only in the latest upgrade script is required.

rohit01010 (changes requested):

contrib/babelfishpg_tsql/sql/sys_functions.sql
bhutraraghav: Select datetime2fromparts(3, 2022, 12, 31, 23, 59, 59, 9999999, 7);
this returns '2023-01-01 00:00:00'
instead of '2022-12-31 23:59:59.999999'
because make_timestamp() used over here rounds of the precision to microseconds

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.2.0--2.3.0.sql
bhutraraghav: The updated test workflow were failing for Version Upgrade Tests.
Adding this to the latest upgrade script only

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--5.4.0--5.5.0.sql
```diff
@@ -552,6 +552,88 @@ LANGUAGE plpgsql
...
 
+CREATE OR REPLACE FUNCTION sys.datetime2fromparts(IN p_year NUMERIC,
+                                                                IN p_month NUMERIC,
+                                                                IN p_day NUMERIC,
+                                                                IN p_hour NUMERIC,
+                                                                IN p_minute NUMERIC,
+                                                                IN p_seconds NUMERIC,
+                                                                IN p_fractions NUMERIC,
+                                                                IN p_precision NUMERIC)
+RETURNS sys.DATETIME2
```
Yvinayak07: As per TSQL doc arguments seems to be int type instead of numeric. Could we try by making that change.

contrib/babelfishpg_tsql/sql/sys_functions.sql
Yvinayak07: Could we test :
```
select datetime2fromparts(2011, 8, 15, 14, 23, 44, 5, 1);
```

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--5.4.0--5.5.0.sql
```diff
@@ -552,6 +552,88 @@ LANGUAGE plpgsql
...
+
+      RAISE USING MESSAGE := pg_catalog.format('Error while trying to cast to %s data type.', v_err_message),
+                  DETAIL := pg_catalog.format('Source value is out of %s data type range.', v_err_message),
+                  HINT := pg_catalog.format('Correct the source value you are trying to cast to %s data type and try again.',
+                                 v_err_message);
+END;
+$BODY$
+LANGUAGE plpgsql
+IMMUTABLE
+RETURNS NULL ON NULL INPUT;
```
Yvinayak07, line 1617: Let's not mark this `NULL ON NULL INPUT` as if precision (the last argument) is NULL we should throw error instead of giving NULL.
e.g.:
```
select datetime2fromparts(2011, 8, 15, 14, 23, 44, 5, NULL);
```

Yvinayak07 (commented):
Could we also take a look at test failures in github actions.
thanks!

contrib/babelfishpg_tsql/sql/sys_functions.sql
bhutraraghav: `select datetime2fromparts(2011, 8, 15, 14, 23, 44, 5, 1);`
output
`2011-08-15 14:23:44.5`

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--2.2.0--2.3.0.sql
rohit01010: > The updated test workflow were failing for Version Upgrade Tests.

Can you brief me on the failure, it could happen that the failure is due to some other reason ?

contrib/babelfishpg_tsql/sql/sys_functions.sql
rohit01010: > Select datetime2fromparts(3, 2022, 12, 31, 23, 59, 59, 9999999, 7);
this returns '2023-01-01 00:00:00'
instead of '2022-12-31 23:59:59.999999'
because make_timestamp() used over here rounds of the precision to microseconds

Here for query `datetime2fromparts(3, 2022, 12, 31, 23, 59, 59, 9999999, 7)` won't result `2023-01-01 00:00:00` makes more sense and is close to given input than `2022-12-31 23:59:59.999999` ? 

Here compared to truncation of the last digit, rounding gives better results. So IMO we should keep the existing behaviour for this. @bhutraraghav Let me know your thoughts on this ?

rohit01010: Lets create a new test file specifically for `datetime2fromparts` and add all these tests in there. Also since this PR has upgrade script changes, we need to test these changes in upgrade test frameworks. 
For adding tests in upgrade test frameworks, you need to create three files
* `<test_name>-vu-prepare.[sql/mix/txt]` -- Add all DDLs here. For example, create tables, views, function which you want to use to test your change(in this case create view/function/procedure that are using datetime2fromparts in it).
* `<test_name>-vu-verify.[sql/mix/txt]` -- Add all the non-DDLs queries. For example, you can add all the select queries here.
* `<test_name>-vu-cleanup.[sql/mix/txt]` -- Add all the delete/drop queries here which will cleanup all the testing objects created by you so that it will not interfere with other tests.

Once you create these test files add a entry `<test_name>` without the suffix(-vu-*), in all the  `upgrade/<version>/schedule` files.
To understand more about Version Upgrade Test Framework, refer: https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/3934#discussion_r2413332402

rohit01010 (changes requested):

contrib/babelfishpg_tsql/sql/sys_functions.sql
bhutraraghav: @rohit01010 I tested this on SQL Server and confirmed that it rounds, not truncates:
```
SELECT DATETIME2FROMPARTS(2022, 12, 31, 23, 59, 59, 9999999, 7) AS Original,
       CAST(DATETIME2FROMPARTS(2022, 12, 31, 23, 59, 59, 9999999, 7) AS DATETIME2(6)) AS Precision6;
```
Result:
```
Original                     Precision6
2022-12-31 23:59:59.9999999  2023-01-01 00:00:00.000000
```
SQL Server rounds the 7th fractional digit when it can't be represented, which causes the rollover to the next day in this edge case. You're correct that 2023-01-01 00:00:00 is mathematically closer to 2022-12-31 23:59:59.9999999 than truncating to 2022-12-31 23:59:59.999999.

I also verified this behavior by inserting datetime2(7) values into Babelfish and confirmed that the last digit is rounded off (not truncated) when stored, consistent with SQL Server's behavior.

I agree we should keep the existing rounding behavior to maintain SQL Server compatibility. Thanks for catching this!

bhutraraghav: @rohit01010 Regarding the tests for datetime2fromparts (which was introduced in version 15_2), these are the existing tests:

1. datetime2fromparts: Used for upgrade tests for source versions before 15_2.
2. datetime2fromparts-after-15_2: Used for upgrade tests after 15_2 and for latest version tests.
3. There are additional tests present in: test/JDBC/input/datatypes/Datetime_system_functions.sql

Given this, should I:
- move all the tests from Datetime_system_functions.sql into datetime2fromparts-after-15_2, or
- keep them as they are?

My initial thought is that we could keep the test structure as it is, but happy to go with your recommendation.

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--5.4.0--5.5.0.sql
bhutraraghav: done

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--5.4.0--5.5.0.sql
bhutraraghav: done

contrib/babelfishpg_tsql/sql/sys_functions.sql
rohit01010: Lets keep the tests in Datetime_system_functions.sql as it is, and add new tests that cover scenarios related to your changes in the existing test files `datetime2fromparts-vu-*` and `datetime2fromparts-after-15_2-vu-*`.

bhutraraghav:
Hi @rohit01010 
can you take a look at this pr?
Thanks

contrib/babelfishpg_tsql/sql/sys_functions.sql
```diff
@@ -472,8 +485,8 @@ BEGIN
        (p_hour::SMALLINT NOT BETWEEN 0 AND 23) OR
        (p_minute::SMALLINT NOT BETWEEN 0 AND 59) OR
        (p_seconds::SMALLINT NOT BETWEEN 0 AND 59) OR
-       (p_fractions::SMALLINT NOT BETWEEN 0 AND 9999999) OR
-       (p_fractions::SMALLINT != 0 AND char_length(v_fractions) > v_precision))
+       (p_fractions::INT NOT BETWEEN 0 AND 9999999) OR
+       (p_fractions::INT != 0 AND char_length(v_fractions) > v_precision))
```
rohit01010: `p_fractions` is already of type INT, we do not need to cast it to INT again.

rohit01010: we can also remove redundant casting on other variables.

contrib/babelfishpg_tsql/sql/sys_functions.sql
```diff
@@ -442,14 +442,14 @@ RETURNS INTEGER
...
-                                                                IN p_fractions NUMERIC,
-                                                                IN p_precision NUMERIC)
+CREATE OR REPLACE FUNCTION sys.datetime2fromparts(IN p_year INT,
+                                                                IN p_month INT,
+                                                                IN p_day INT,
+                                                                IN p_hour INT,
+                                                                IN p_minute INT,
+                                                                IN p_seconds INT,
+                                                                IN p_fractions INT,
+                                                                IN p_precision INT)
```
rohit01010, line 458: Do we have any tests where parameters passed to datetime2fromparts are of type other than INT, like decimal numbers, float, real, etc. If we do not have those tests lets add them and check whether changing argument type of this function `datetime2fromparts` from NUMERIC to INT is not regressing behaviour of datetime2fromparts with argument of other datatypes.

bhutraraghav: Added

contrib/babelfishpg_tsql/sql/sys_functions.sql
bhutraraghav: Done

rohit01010 (approved):

bhutraraghav:
@rohit01010 do we need another approval to merge this ?, if yes can you ask someone to take a look. Thanks

rohit01010:
Hi @bhutraraghav, I am waiting on approvals from remaining reviewers: @tanyagupta17, @Yvinayak07. 
@Yvinayak07 @tanyagupta17 Lets try to complete the review soon.

contrib/babelfishpg_tsql/sql/sys_functions.sql
```diff
@@ -683,15 +694,28 @@ CREATE OR REPLACE FUNCTION sys.datetime2fromparts(IN p_year TEXT,
...
 AS
 $BODY$
 DECLARE
     v_err_message VARCHAR;
 BEGIN
-    RETURN sys.datetime2fromparts(p_year::NUMERIC, p_month::NUMERIC, p_day::NUMERIC,
-                                                p_hour::NUMERIC, p_minute::NUMERIC, p_seconds::NUMERIC,
-                                                p_fractions::NUMERIC, p_precision::NUMERIC);
+    IF p_precision IS NULL THEN
+        RAISE invalid_parameter_value USING
```
tanyagupta17, line 710: Do we have tests where parameters passed to datetime2fromparts are of type other than text? If not could you please add tests where we can hit our newly added conditions and verify it. 
One example could be like this, please feel free to add different types of test cases. 
```
SELECT datetime2fromparts(cast('2023' as text), cast('6' as text), cast('15' as text), cast('10' as text), cast('30' as text), cast('45' as text), cast('123456' as text), 6) AS Test1_All_Text_Valid;

```

contrib/babelfishpg_tsql/sql/sys_functions.sql
```diff
@@ -683,15 +694,28 @@ CREATE OR REPLACE FUNCTION sys.datetime2fromparts(IN p_year TEXT,
                                                                 IN p_seconds TEXT,
                                                                 IN p_fractions TEXT,
                                                                 IN p_precision TEXT)
-RETURNS TIMESTAMP WITHOUT TIME ZONE
+RETURNS sys.DATETIME2
```
tanyagupta17, line 703: Could we please also add test cases to verify that changing the return type is not causing any regressions?

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--5.4.0--5.5.0.sql
```diff
@@ -564,6 +564,151 @@ END;
 $$ 
 LANGUAGE plpgsql IMMUTABLE STRICT;
 
+ALTER FUNCTION sys.datetime2fromparts(NUMERIC, NUMERIC, NUMERIC, NUMERIC, NUMERIC, NUMERIC, NUMERIC, NUMERIC) RENAME TO datetime2fromparts_deprecated_in_5_5_0;
```
Yvinayak07: As we have moved to next release we need to move these changes to next upgrade file i.e `babelfishpg_tsql--5.5.0--5.6.0.sql`
Copied!
thanks!

bhutraraghav: Done

contrib/babelfishpg_tsql/sql/sys_functions.sql
bhutraraghav: done, added some tests for return type

contrib/babelfishpg_tsql/sql/sys_functions.sql
bhutraraghav: Done, added

bhutraraghav:
Hi @rohit01010 , can you assign this PR to someone else for review if @tanyagupta17 @Yvinayak07 are not available?
Thanks

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--5.5.0--5.6.0.sql
```diff
@@ -44,6 +44,151 @@ BEGIN
 END;
 $$;
 
+ALTER FUNCTION sys.datetime2fromparts(NUMERIC, NUMERIC, NUMERIC, NUMERIC, NUMERIC, NUMERIC, NUMERIC, NUMERIC) RENAME TO datetime2fromparts_deprecated_in_5_6_0;
```
rohit01010: Let's add an exception handling block around this renaming to avoid upgrade failures if previous version does not had this definition.
```
DO $$
DECLARE
    exception_message text;
BEGIN
    ALTER FUNCTION sys.datetime2fromparts(NUMERIC, NUMERIC, NUMERIC, NUMERIC, NUMERIC, NUMERIC, NUMERIC, NUMERIC) RENAME TO datetime2fromparts_deprecated_in_5_6_0;
EXCEPTION WHEN OTHERS THEN
    GET STACKED DIAGNOSTICS
    exception_message = MESSAGE_TEXT;
    RAISE WARNING '%', exception_message;
END;
$$;
```

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--5.5.0--5.6.0.sql
```diff
@@ -44,6 +44,151 @@ BEGIN
...
+      RAISE USING MESSAGE := pg_catalog.format('Error while trying to cast to %s data type.', v_err_message),
+                  DETAIL := pg_catalog.format('Source value is out of %s data type range.', v_err_message),
+                  HINT := pg_catalog.format('Correct the source value you are trying to cast to %s data type and try again.',
+                                 v_err_message);
+END;
+$BODY$
+LANGUAGE plpgsql
+IMMUTABLE;
+
+ALTER FUNCTION sys.datetime2fromparts(TEXT, TEXT, TEXT, TEXT, TEXT, TEXT, TEXT, TEXT) RENAME TO datetime2fromparts_text_deprecated_in_5_6_0;
```
rohit01010: Similarly also add an exception handling block around this renaming to avoid upgrade failures if previous version does not had this definition.
```
DO $$
DECLARE
    exception_message text;
BEGIN
    ALTER FUNCTION sys.datetime2fromparts(TEXT, TEXT, TEXT, TEXT, TEXT, TEXT, TEXT, TEXT) RENAME TO datetime2fromparts_text_deprecated_in_5_6_0;
EXCEPTION WHEN OTHERS THEN
    GET STACKED DIAGNOSTICS
    exception_message = MESSAGE_TEXT;
    RAISE WARNING '%', exception_message;
END;
$$;
```

rohit01010:
@bhutraraghav Kindly address above comments. Also, informed @tanyagupta17 @Yvinayak07 they will review this PR soon.

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--5.5.0--5.6.0.sql
bhutraraghav: Done

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--5.5.0--5.6.0.sql
bhutraraghav: Done

Yvinayak07 (approved):

rohit01010 (approved):

tanyagupta17 (approved):

jsudrik (approved):
Approved.

PR #4065, opened 2025-09-09
Fixed error when parallel worker tries to access rel with oid > INT32_MAX

Deepesh125:
### Description

As part of the commit 6da7ff0, we introduced logic to skip permission checking of Babelfish temp table on Parallel
workers. But logic will run into an error if Oid is greated than INT32_MAX because Bitmapset was used to communicate
Oids from leader node to parallel worker. This Bitmapset can only work with signed integer meaning if Oid value is greater
than INT32_MAX then it will be rounded off to negative value in signed integer and Bitmapset will throw an error that
it can't handle negative values.

In order to fix this overflow issue, we create one more bitmapset especially for Oid > INT32_MAX and communicate it with
parallel worker. Construction of bitmapset now would look like this, 
```
if relid <= INT32_MAX
    add relid to temp_relids
else
    add (relid - INT32_MAX - 1) to temp_relids_beyond_int32
 ```

### Issues Resolved
BABEL-6079

### Check List

### DIFF
--- a/contrib/babelfishpg_tsql/src/bbf_parallel_query.c
+++ b/contrib/babelfishpg_tsql/src/bbf_parallel_query.c
@@ -24,15 +24,25 @@
 
 /*
  * temp_relids - maintains relid list of temp table shared by leader node. This should be
-* strictly accessed within parallel worker context.
+ * strictly accessed within parallel worker context.
+ * Here, we keep two bitmapset, one for relids <= INT32_MAX and another for relids > INT32_MAX
+ * because OID is unsigned int and bitmapset will not be handle OID > INT32_MAX.
+ * Overall,
+ * 	if relid <= INT32_MAX
+ * 		add relid to temp_relids
+ * 	else
+ * 		add (relid - INT32_MAX) to temp_relids_beyond_in32
  */
 static Bitmapset   *temp_relids = NULL;
+static Bitmapset   *temp_relids_beyond_in32 = NULL;
 
 /*
  * string representation of list of temp table oids computed by Leader node to be shared
- * with parallel worker.
+ * with parallel worker. Check comment above temp_relids_beyond_in32 for reason to
+ * keep two bitmapset string.
  */
 static char		   *temp_relids_str = NULL;
+static char		   *temp_relids_beyond_int32_str = NULL;
 
 /*
  * In Babelfish, Any user under given session should be able access the temp tables. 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: contrib/babelfishpg_tsql/src/bbf_parallel_query.c [hunk 1]
```diff
@@ -105,23 +117,29 @@ bbf_ExecInitParallelPlan(EState *estate, ParallelContext *pcxt, bool estimate)
...
 				}
-				temp_relids_local = bms_add_member(temp_relids_local, rte->relid);
+				if (rte->relid <= (Oid) INT32_MAX)
+					temp_relids_local = bms_add_member(temp_relids_local, rte->relid);
+				else
+					temp_relids_beyond_int32_local = bms_add_member(temp_relids_beyond_int32_local, (rte->relid - (Oid) INT32_MAX - 1U));
 			}
 		}
 		temp_relids_str = bmsToString(temp_relids_local);
+		temp_relids_beyond_int32_str = bmsToString(temp_relids_beyond_int32_local);
```
surendravishnoi, line 128: what if temp_relids_beyond_int32_local is NULL? what will bmsTostring return?

Deepesh125: it will be empty representation of bms like `(b)`

surendravishnoi (approved):

contrib/babelfishpg_tsql/src/bbf_parallel_query.c
```diff
@@ -105,23 +117,29 @@ bbf_ExecInitParallelPlan(EState *estate, ParallelContext *pcxt, bool estimate)
 									get_relkind_objtype(get_rel_relkind(perminfo->relid)),
 									get_rel_name(perminfo->relid));
 				}
-				temp_relids_local = bms_add_member(temp_relids_local, rte->relid);
+				if (rte->relid <= (Oid) INT32_MAX)
+					temp_relids_local = bms_add_member(temp_relids_local, rte->relid);
+				else
+					temp_relids_beyond_int32_local = bms_add_member(temp_relids_beyond_int32_local, (rte->relid - (Oid) INT32_MAX));
```
lejaokri: nit: did we check for other calls to `bms_*` in bbf code to see if they have the same issue?

Deepesh125: yes, already scanned through. this is the only occurrence

Issue #4028, opened 2025-08-18
[Bug]: float/real accept NaN/Infinity in Babelfish (but T-SQL doesn’t) — intended?


bilalkah:
### What happened?

In T-SQL, there’s no textual representation for non-finite floating values (NaN, Infinity, -Infinity). Attempts to cast such strings to float/real should fail. In Babelfish, the same inputs are accepted when targeting real (float4) or float (float8), because PostgreSQL’s float I/O routines allow these special values. Is this divergence intentional? If not, I can prepare a fix.

### Case examples
```sql
SELECT CAST('NaN' AS float);
GO
SELECT CAST('Infinity' AS float);
GO
DECLARE @f FLOAT='infinify';
select @f;
GO
```
They all pass in babelfish and fail in T-SQL. Additionally, Also, if you connect babelfish via SSMS, they also fail. Probably SSMS does additional check as well. 

### Why this happens
As I see, sys.FLOAT and sys.REAL are domain types of float4 and float8, therefore when an inputs comes or when a cast happens they use underlying types behavior.

### Result

So as a result, we can adjust the real/float behaviour with additional helper function in the following example for real datatype
```c
Datum
varchar2float4(PG_FUNCTION_ARGS)
{
	VarChar    *source = PG_GETARG_VARCHAR_PP(0);
	char 	   *num;

	if (varcharTruelen(source) == 0)
		PG_RETURN_FLOAT4(0);

	num = varchar2cstring(source);

	/* Validate the input string for T-SQL float literals */
	if (!is_valid_tsql_float(num))
	{
		pfree(num);
		ereport(ERROR,
			(errcode(ERRCODE_INVALID_PARAMETER_VALUE),
			errmsg("Error converting data type varchar to real.")));
	}

	return cstring2float4(num);
}
```
If this divergence is intended, please confirm and consider documenting it as a known difference. Otherwise, I’m happy to submit a PR implementing the stricter T-SQL behavior described above. We can also introduce a GUC, default on:
```ini
babelfish.tsql_float_literals_strict = on
```
When off, keep current PostgreSQL behavior.

### Version

BABEL_5_X_DEV (Default)

### Extension

None

### Which flavor of Linux are you using when you see the bug?

Ubuntu (Default)

### Relevant log output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct.

Yvinayak07:
[...]

PR #4075, opened 2025-09-10
Validate inputs for FLOAT/REAL data types

bilalkah:
### Description

T-SQL parity for `real/float`: reject special literals; add safe casts

- What: String→`real/float` casts now reject `NaN`/`Infinity`/`inf` (T-SQL doesn’t allow them). `TRY_CAST` returns `NULL` instead of accepting those values. Empty strings cast to `0`.
- Why: Align Babelfish with T-SQL behavior; avoid downstream errors.
- How (high level):
 - Added validator `is_valid_tsql_float` and used it in `varchar2float4/8`, `bpchar2float4/8`.
 - Registered explicit casts for `pg_catalog.VARCHAR/BPCHAR/TEXT → FLOAT4/FLOAT8` via new `sys.*2float*` functions.
 - Parser/coercion: route `TRY_CAST(... AS real|float(p));` `float(p)` uses `real` for `1..24`, `float8` otherwise.
 - Exposed `is_tsql_real_datatype` / `is_tsql_float_datatype` via common utility plugin.
 - Added upgrade scripts to install new casts/functions.

### Issues Resolved

https://github.com/babelfish-for-postgresql/babelfish_extensions/issues/4028

- CAST('' AS real) → 0; CAST('Infinity' AS float) → error
- TRY_CAST('NaN' AS real) → NULL
- TRY_CAST('123' AS float(24)) → real; float(25) → float8
- Whitespace/exponent forms accepted; non-numeric strings error (or NULL via TRY_CAST)

### Check List


contrib/babelfishpg_common/sql/bpchar.sql
```diff
@@ -210,6 +210,14 @@ LANGUAGE C IMMUTABLE STRICT PARALLEL SAFE;
 CREATE CAST (sys.BPCHAR AS FLOAT4)
 WITH FUNCTION sys.bpchar2float4(sys.BPCHAR) AS IMPLICIT;
 
+CREATE OR REPLACE FUNCTION sys.bpchar2float4(pg_catalog.BPCHAR)
```
Yvinayak07, line 216: As we are creating all new casts here, so where was this hit earlier?

bilalkah: On the T-SQL side, these casts were handled through I/O:
```c
/*  string -> float8 via I/O */
{TSQL_CAST_WITHOUT_FUNC_ENTRY, "pg_catalog", "text",    "pg_catalog", "float8", NULL, 'i', 'i'},
{TSQL_CAST_WITHOUT_FUNC_ENTRY, "pg_catalog", "bpchar",  "pg_catalog", "float8", NULL, 'i', 'i'},
{TSQL_CAST_WITHOUT_FUNC_ENTRY, "pg_catalog", "varchar", "pg_catalog", "float8", NULL, 'i', 'i'},

/*  string -> float4 via I/O */
{TSQL_CAST_WITHOUT_FUNC_ENTRY, "pg_catalog", "text",    "pg_catalog", "float4", NULL, 'i', 'i'},
{TSQL_CAST_WITHOUT_FUNC_ENTRY, "pg_catalog", "bpchar",  "pg_catalog", "float4", NULL, 'i', 'i'},
{TSQL_CAST_WITHOUT_FUNC_ENTRY, "pg_catalog", "varchar", "pg_catalog", "float4", NULL, 'i', 'i'},
```

On the PostgreSQL side, it also happens via I/O (COERCION_PATH_COERCEVIAIO), which directly calls float4in.

The problem is that the behavior is not the same as in T-SQL. PostgreSQL accepts special values like "infinity" and "NaN", while SQL Server does not. Since we cannot validate these values beforehand, all casts from string types to float/real were redirected to custom casts so we can perform validation before inserting.

Yvinayak07 (commented):
And let's fix the test cases first.

bilalkah:
> And let's fix the test cases first.

@Yvinayak07 I don't understand. What do you mean?

bilalkah:
Can you please make a review? @Yvinayak07

bilalkah:
Hi @Yvinayak07 ,
Can you please help me understand the upgrade test from version X_Y to latest? As far as I understand, in such test cases, the result should include my changes since it has been upgraded to latest. However, the results are saying that. For example, I need to provide old results in test/JDBC/upgrade/17_7/schedule. Why is that?

PR #4091, opened 2025-09-14
Added support OPENXML without WITH clause 

rohit01010:
### Description

This PR adds support for OPENXML without WITH Clause which provides an edge table which represent XML document structure in a single table.
The syntax of openxml without WITH clause is as follows :
```
OPENXML ( idoc int [ in ] , rowpattern nvarchar [ in ] [, flags byte [ in ] ] )
```
This support is added by implementing a C function `OPENXML_SIMPLE` which uses libxml2 APIs to parse the XML document and evaluate given XML document against row pattern. Also added function `assign_ids` to maintain unique ids for each nodes in XML document, and function `add_node_details` which traverse over all the evaluated XML nodes and fetches each node details and populate them in the form of rows.

Engine PR: https://github.com/babelfish-for-postgresql/postgresql_modified_for_babelfish/pull/622

Authored-by: Rohit Bhagat [rohitbgt@amazon.com](mailto:rohitbgt@amazon.com)
Signed-off-by: Rohit Bhagat [rohitbgt@amazon.com](mailto:rohitbgt@amazon.com)

### Issues Resolved

BABEL-6046

* **Use case based -** YES

* **Arbitrary inputs -** YES

* **Performance tests -** NA

* **Tooling impact -** NA

* **Client tests -** NA

### Check List


contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -4923,4 +4926,90 @@ sp_xml_removedocument(PG_FUNCTION_ARGS)
...
+
+	if (xml_data == NULL && ns_data == NULL)
+		return;
+
+	if (xml_data)
+		*xml_data = NULL;
+	if (ns_data)
+		*ns_data = NULL;
+
+	if (xml_handle_temp_table_name != NULL)
```
tanyagupta17, line 4965: Could we add some comments explaining the different steps

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -114,6 +114,8 @@ bool		sp_describe_first_result_set_inprogress = false;
 char	   *orig_proc_funcname = NULL;
 static bool is_supported_case_sp_describe_undeclared_parameters = true;
 
+#define 			 XML_HANDLE_DOC_COLUMN_NUM 6
+#define 			 XML_HANDLE_NAMESPACE_COLUMN_NUM 7
```
tanyagupta17: Could we rename these to something like `anum-table_name-col_name` and add a comment on top of this.

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -5123,3 +5167,671 @@ get_bbf_pivot_tuplestore(const char 	*sourcetext,
...
+        if (doc)
+            xmlFreeDoc(doc);
+        return;
+    }
+
+	/*
+	 * Allocate memory for namespace names and URIs
+	 */
+	*ns_names = (char **) palloc0((*ns_count) * sizeof(char *));
+    *ns_uris = (char **) palloc0((*ns_count) * sizeof(char *));
```
tanyagupta17, line 5232: Could you please look into pfree() of these args appropriately?

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -5123,3 +5167,671 @@ get_bbf_pivot_tuplestore(const char 	*sourcetext,
...
+	for (xmlNodePtr cur = node->children; cur != NULL; cur = cur->next)
+	{
+		populate_xml_nodes(cur);
+	}
+}
+
+/*
+ * For a given XML Document Node, 
+ * function assign_ids prepares a hash table mapping for each xmlNodePtr to a unique ID.
+ */
```
tanyagupta17: Nit: Please keep the commenting style consistent

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -5123,3 +5167,671 @@ get_bbf_pivot_tuplestore(const char 	*sourcetext,
...
+	}
+}
+#endif							/* USE_LIBXML */
+
+Datum
+openxml_simple(PG_FUNCTION_ARGS)
+{
+#ifdef USE_LIBXML
+    int        idoc = PG_GETARG_INT32(0);
+    text      *xpath_expr_text = PG_GETARG_TEXT_PP(1);
```
tanyagupta17: Could we add a null check for this

rohit01010: Done

contrib/babelfishpg_tsql/runtime/functions.c
rohit01010: Updated

contrib/babelfishpg_tsql/src/procedures.c
rohit01010: Done

contrib/babelfishpg_tsql/src/procedures.c
rohit01010: Done

contrib/babelfishpg_tsql/runtime/functions.c
rohit01010: Done

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -4923,4 +4929,100 @@ sp_xml_removedocument(PG_FUNCTION_ARGS)
...
+	}
+	else
+	{
+		ereport(ERROR,
+				(errcode(ERRCODE_UNDEFINED_OBJECT),
+				 errmsg("Could not find prepared statement with handle %d.", document_id)));
+	}
+
+	table_endscan(scan);
+	relation_close(relation, AccessShareLock);
```
HarshLunagariya, line 5030: Should be RowExclusiveLock

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -234,6 +243,41 @@ extern bool inited_ht_tsql_cast_info;
...
+	/* strictness argument passed to pg_xml_init */
+	PgXmlStrictness strictness;
+	/* current error status and accumulated message, if any */
+	bool		err_occurred;
+	StringInfoData err_buf;
+	/* previous libxml error handling state (saved by pg_xml_init) */
+	xmlStructuredErrorFunc saved_errfunc;
+	void	   *saved_errcxt;
+	/* previous libxml entity handler (saved by pg_xml_init) */
+	xmlExternalEntityLoader saved_entityfunc;
```
HarshLunagariya: Have we refined from here from engine? If yes please add the reference/file location from engine.

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -5123,3 +5167,766 @@ get_bbf_pivot_tuplestore(const char 	*sourcetext,
...
+		/*
+		 * For each of the namespace declaration in the node, create a new attribute.
+		 */
+		for (xmlNs *cur = node->nsDef; cur != NULL; cur = cur->next)
+		{
+			if (cur->prefix == NULL)	// Default namespace declaration
+				xmlNewNsProp(node, xmlNewNs(NULL, NULL, BAD_CAST "xmlns"), BAD_CAST "xmlns", BAD_CAST cur->href);
+			else
+				xmlNewNsProp(node, xmlNewNs(NULL, NULL, BAD_CAST "xmlns"), BAD_CAST cur->prefix, BAD_CAST cur->href);
+		}
```
HarshLunagariya: What if these APIs throws error? IMO we should show them as internal error. Does it make sense to wrap it with try/catch block and rethrow the internal error and print the API error in the logs?

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -5123,3 +5167,766 @@ get_bbf_pivot_tuplestore(const char 	*sourcetext,
...
+
+	if (key == NULL)
+		return -1;
+
+	hinfo = (ht_xmlNode2Id_entry_t *) hash_search(ht_xmlNode2Id,
+												  &key,
+												  HASH_FIND,
+												  &found);
+	if (!found)
+		return -1;
```
HarshLunagariya, line 5425: Should it be the error case? There always be an entry, right?

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -5123,3 +5167,766 @@ get_bbf_pivot_tuplestore(const char 	*sourcetext,
...
+			nulls[3] = false;
+			values[3] = PointerGetDatum((VarChar *) cstring_to_text("#comment"));
+		}
+		else
+		{
+			if (node->name != NULL)
+			{
+				nulls[3] = false;
+				values[3] = PointerGetDatum((VarChar *) cstring_to_text((const char *) node->name));
+			}
```
HarshLunagariya, line 5499: Can node->name be NULL?

contrib/babelfishpg_tsql/runtime/functions.c
rohit01010: There are few nodes such as `XML_DOCUMENT_NODE` to which we do not assign any id. For such nodes, NULL should be returned instead of error. Refer following example
```
-- Simple XML Document
DECLARE @xml nvarchar(1000) = '<root>Hello World</root>';
DECLARE @handle INT;
EXEC sp_xml_preparedocument @handle OUTPUT, @xml;
SELECT * FROM OPENXML(@handle, '/') ORDER BY id;
EXEC sp_xml_removedocument @handle
GO
~~START~~
bigint#!#bigint#!#int#!#nvarchar#!#nvarchar#!#nvarchar#!#nvarchar#!#bigint#!#ntext
0#!#<NULL>#!#1#!#root#!#<NULL>#!#<NULL>#!#<NULL>#!#<NULL>#!#<NULL>
2#!#0#!#3#!##text#!#<NULL>#!#<NULL>#!#<NULL>#!#<NULL>#!#Hello World
~~END~~
```
Here the parent id of root element is NULL, the parent of root element is document node whose id we have not assigned and this check will help us to get above behaviour.

contrib/babelfishpg_tsql/runtime/functions.c
rohit01010: No node->name will not be NULL, these NULL checks are just sanity checks to avoid any unexpected failure if in any case node->name became NULL.

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--5.3.0--5.4.0.sql
```diff
@@ -223,6 +223,23 @@ GRANT EXECUTE ON PROCEDURE sys.sp_xml_removedocument(
...
+                                       rowpattern TEXT, 
+                                       flags INTEGER DEFAULT 0)
+RETURNS table (
+  id sys.BIGINT,
+  parentid sys.BIGINT,
+  nodetype sys.INT,
+  localname sys.NVARCHAR,
+  prefix sys.NVARCHAR,
+  namespaceuri sys.NVARCHAR,
+  datatype sys.NVARCHAR,
```
Deepesh125, line 236: make sure that typmod is correct/expected in this case?

contrib/babelfishpg_tsql/runtime/functions.c
rohit01010: Done

contrib/babelfishpg_tsql/src/procedures.c
rohit01010: Fixed

contrib/babelfishpg_tsql/runtime/functions.c
rohit01010: APIs xmlNewNs and xmlNewNsProp does not throw. These APIs return NULL instead, if arguments are invalid or a memory allocation failed. Added handling to catch such scenario's.

contrib/babelfishpg_tsql/src/backend_parser/gram-tsql-rule.y
```diff
@@ -1779,13 +1779,51 @@ table_ref:	relation_expr tsql_table_hint_expr
 					 */
 					$$ = (Node *) $1;
 				}
+			| TSQL_APPLY openxml_expr
```
Deepesh125, line 1796: what is different here? have we added test cases for the same?

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -4923,4 +4929,103 @@ sp_xml_removedocument(PG_FUNCTION_ARGS)
...
+	/* 
+	 * Check if the xml_handle_temp_table exists using ENR lookup,
+	 * if found get its relation descriptor.
+	 */
+	if (xml_handle_temp_table_name != NULL)
+	{
+		enr = get_ENR(currentQueryEnv, xml_handle_temp_table_name, true);
+		if (enr)
+		{
+			relation = relation_open(enr->md.reliddesc, AccessShareLock);
```
Deepesh125, line 4970: Can you confirm with @ayushdsh or @lejaokri whether any kind of locking needed on temp table to read the data? Because my understanding is that we should not care because temp table anyway can't be accessed from concurrent session

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -4923,4 +4929,103 @@ sp_xml_removedocument(PG_FUNCTION_ARGS)
...
+	Relation               		relation;
+	ScanKeyData            		skey[1];
+	TableScanDesc      		    scan;
+	HeapTuple              		tuple;
+	Datum                  		datum;
+	bool                   		isnull;
+	bool              	        table_exists = false;
+
+	if (xml_data == NULL && ns_data == NULL)
+		return;
```
Deepesh125, line 4951: I am not sure what case we are handling here? IMO, we should initialise these variables to NULL from wherever this function is being called.

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -5123,3 +5168,783 @@ get_bbf_pivot_tuplestore(const char 	*sourcetext,
...
+    if (doc == NULL)
+		return;
+
+	/*
+	 * Get namespace declaration count
+	 */
+	root = xmlDocGetRootElement(doc);
+    for (xmlNs *cur = root->nsDef; cur != NULL; cur = cur->next)
+	{
+		if (cur->prefix)	// Ignore default namespace declaration
```
Deepesh125, line 5215: please explain meaning of this comment

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -5123,3 +5168,783 @@ get_bbf_pivot_tuplestore(const char 	*sourcetext,
...
+	doc = xml_parse_wrapper(ns_data, XMLOPTION_DOCUMENT, false, GetDatabaseEncoding(), NULL, NULL, NULL);
+
+    if (doc == NULL)
+		return;
+
+	/*
+	 * Get namespace declaration count
+	 */
+	root = xmlDocGetRootElement(doc);
+    for (xmlNs *cur = root->nsDef; cur != NULL; cur = cur->next)
```
Deepesh125, line 5213: root might be NULL?

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -5123,3 +5168,783 @@ get_bbf_pivot_tuplestore(const char 	*sourcetext,
...
+	*ns_names = (char **) palloc0((*ns_count) * sizeof(char *));
+    *ns_uris = (char **) palloc0((*ns_count) * sizeof(char *));
+
+	/*
+	 * Store namespace names and URIs in ns_names and ns_uris
+	 */
+    index = 0;
+    for (xmlNs *cur = root->nsDef; cur != NULL; cur = cur->next)
+    {
+        if (cur->prefix)
```
Deepesh125, line 5240: do we need any check on index? like index < ns_count?

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -5123,3 +5168,783 @@ get_bbf_pivot_tuplestore(const char 	*sourcetext,
...
+        {
+            (*ns_names)[index] = (char *) pstrdup((const char *) cur->prefix);
+            (*ns_uris)[index] = cur->href ? (char *) pstrdup((const char *) cur->href) : NULL;
```
Deepesh125, line 5243: these cur->prefix and cur->href will always be NULL terminated? if not sure then better to use pnstrdup

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -5123,3 +5168,783 @@ get_bbf_pivot_tuplestore(const char 	*sourcetext,
...
+     */
+    get_xml_data_and_namespace_data(document_id, &xmldata, &ns_data);
+
+	if (xmldata == NULL)
+		goto done;
+
+    extract_namespaces_from_xml(ns_data, &ns_names, &ns_uris, &ns_count);
+
+	datastr = VARDATA(xmldata);
+	len = VARSIZE(xmldata) - VARHDRSZ;
```
Deepesh125: why dont we use VARSIZE_ANY_EXHDR?

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -5123,3 +5168,783 @@ get_bbf_pivot_tuplestore(const char 	*sourcetext,
...
+		ereport(ERROR,
+				(errcode(ERRCODE_NULL_VALUE_NOT_ALLOWED),
+				 errmsg("XPath expression cannot be null")));
+
+	xpath_expr_text = PG_GETARG_TEXT_PP(1);
+	xpath_len = VARSIZE_ANY_EXHDR(xpath_expr_text);
+	if (xpath_len == 0)
+		ereport(ERROR,
+				(errcode(ERRCODE_DATA_EXCEPTION),
+				errmsg("empty XPath expression")));
```
Deepesh125: Is this expected tsql?

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -5123,3 +5168,783 @@ get_bbf_pivot_tuplestore(const char 	*sourcetext,
...
+
+	string = pg_xmlCharStrndup_wrapper(datastr, len);
+	xpath_expr = pg_xmlCharStrndup_wrapper(VARDATA_ANY(xpath_expr_text), xpath_len);
+
+	/*
+	 * In a UTF8 database, skip any xml declaration, which might assert
+	 * another encoding.  Ignore parse_xml_decl_wrapper() failure, letting
+	 * xmlCtxtReadMemory() report parse errors.
+	 */
```
Deepesh125: Babelfish will always use UTF8

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -5123,3 +5168,783 @@ get_bbf_pivot_tuplestore(const char 	*sourcetext,
...
+
+	xpath_expr_text = PG_GETARG_TEXT_PP(1);
+	xpath_len = VARSIZE_ANY_EXHDR(xpath_expr_text);
+	if (xpath_len == 0)
+		ereport(ERROR,
+				(errcode(ERRCODE_DATA_EXCEPTION),
+				errmsg("empty XPath expression")));
+
+	string = pg_xmlCharStrndup_wrapper(datastr, len);
+	xpath_expr = pg_xmlCharStrndup_wrapper(VARDATA_ANY(xpath_expr_text), xpath_len);
```
Deepesh125, line 5786: this pg_xmlCharStrndup would return palloc'd copy. Make sure we free that memory

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -5123,3 +5168,783 @@ get_bbf_pivot_tuplestore(const char 	*sourcetext,
...
+
+	/*
+	 * In a UTF8 database, skip any xml declaration, which might assert
+	 * another encoding.  Ignore parse_xml_decl_wrapper() failure, letting
+	 * xmlCtxtReadMemory() report parse errors.
+	 */
+	if (GetDatabaseEncoding() == PG_UTF8)
+		parse_xml_decl_wrapper((xmlChar *) string, &xmldecl_len, NULL, NULL, NULL);
```
Deepesh125: This function may return non-zero in the case of error or invalid xml. We should catch it and take appropriate action before continuing

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -5123,3 +5168,783 @@ get_bbf_pivot_tuplestore(const char 	*sourcetext,
...
+
+/*
+ * populate_xml_nodes 
+ * 		Recursively traverse the XML tree and populate xml_nodes_list
+ */
+static void 
+populate_xml_nodes(xmlNode *node)
+{
+	if (node->type == XML_TEXT_NODE && xmlIsBlankNode(node))
+		return;  // skip whitespace-only text node
```
Deepesh125, line 5305: can you give an example?

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -5123,3 +5168,783 @@ get_bbf_pivot_tuplestore(const char 	*sourcetext,
...
+		{
+			entry->id = (*cur == root) ? 0 : counter;
+			counter++;
+		}
+	}
+
+	/*
+	 * Free the temporary list of XML nodes as it is no longer needed.
+	 */
+	destroy_vector(xml_nodes_list);
```
Deepesh125, line 5402: what about hash table

Deepesh125 (changes requested):

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--5.3.0--5.4.0.sql
rohit01010: Yes, the typmod is correct here.

contrib/babelfishpg_tsql/runtime/functions.c
rohit01010: No, root cannot be NULL here, as the `ns_data` is parsed with option `XMLOPTION_DOCUMENT` so the result doc pointer will always have root node, else an error will be thrown during parsing in function `xml_parse_wrapper`.

contrib/babelfishpg_tsql/runtime/functions.c
rohit01010: Default namespace declarations are ignored as they do not have any prefix hence they cannot be registered in XPATH. 
```
'<root xmlns="http://example.com/default"></root>'
```
In above example, we can see there is no prefix for `namespace-uri` `"http://example.com/default"`. Such namespace are called default namespace.

contrib/babelfishpg_tsql/runtime/functions.c
rohit01010: Not required, as index will always be less than ns_count for cur != NULL.

contrib/babelfishpg_tsql/runtime/functions.c
rohit01010: cur->prefix and cur->href are of type (xmlChar *), where xmlChar is defined as  `unsigned char`. So its just like other (char *) type strings which are always NULL terminated. Also we cannot use pnstrdup here as we do not know the size of cur->prefix and cur->href.

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -5123,3 +5168,783 @@ get_bbf_pivot_tuplestore(const char 	*sourcetext,
...
+	}
+}
+
+/*
+ * populate_xml_nodes 
+ * 		Recursively traverse the XML tree and populate xml_nodes_list
+ */
+static void 
+populate_xml_nodes(xmlNode *node)
+{
```
Deepesh125, line 5299: if node == NULL then return

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -234,6 +243,42 @@ extern bool inited_ht_tsql_cast_info;
 extern bool inited_ht_tsql_datatype_precedence_info;
 extern PLtsql_execstate *get_outermost_tsql_estate(int *nestlevel);
 extern char *replace_special_chars_fts_impl(char *input_str);
+extern void get_xml_data_and_namespace_data(int document_id, xmltype **xml_data, xmltype **ns_data);
+
+#ifdef USE_LIBXML
+MemoryContext TransMemoryContext = NULL;
+HTAB	     *ht_xmlNode2Id = NULL;
+static bool   inited_ht_xmlNode2Id = false;
+DynaVec 	 *xml_nodes_list = NULL;
```
Deepesh125: why global list?

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -5123,3 +5168,783 @@ get_bbf_pivot_tuplestore(const char 	*sourcetext,
...
+{
+	if (node->type == XML_TEXT_NODE && xmlIsBlankNode(node))
+		return;  // skip whitespace-only text node
+
+	/*
+	 * Add the current node to the list if it is not a Document node.
+	 * Document node is not added to the list as it is not required for OpenXML processing.
+	 */
+	if (node->type != XML_DOCUMENT_NODE)
+		vec_push_back(xml_nodes_list, &node);
```
Deepesh125, line 5312: why &node?

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -5123,3 +5168,783 @@ get_bbf_pivot_tuplestore(const char 	*sourcetext,
...
+	for (i = 1; i <= xml_nodes_list_size; i++)
+	{
+		ht_xmlNode2Id_entry_t *entry;
+		bool                   found = false;
+		xmlNode              **cur = (xmlNode **) vec_at(xml_nodes_list, i-1);
+
+		entry = hash_search(ht_xmlNode2Id, cur, HASH_ENTER, &found);
+		if (!found)
+		{
+			entry->id = (*cur == root) ? 0 : counter;
```
Deepesh125, line 5394: is this the right way to compare xmlNode? seems like we are just checking address

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -5123,3 +5168,783 @@ get_bbf_pivot_tuplestore(const char 	*sourcetext,
...
+populate_xml_nodes(xmlNode *node)
+{
+	if (node->type == XML_TEXT_NODE && xmlIsBlankNode(node))
+		return;  // skip whitespace-only text node
+
+	/*
+	 * Add the current node to the list if it is not a Document node.
+	 * Document node is not added to the list as it is not required for OpenXML processing.
+	 */
+	if (node->type != XML_DOCUMENT_NODE)
```
Deepesh125, line 5311: what about XML_DOCUMENT_TYPE_NODE?

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -5123,3 +5168,783 @@ get_bbf_pivot_tuplestore(const char 	*sourcetext,
...
+		|| node->type == XML_PI_NODE)
+	{
+		/*
+		 * Initialize all values to NULL and nulls to true
+		 */
+		memset(values, 0, sizeof(values));
+		memset(nulls, true, sizeof(nulls));
+
+		node_id = lookup_xmlNode_id(node);
+		if (node_id != -1)
```
Deepesh125, line 5462: previous populate_xml should have visited all the nodee, right? then how can we end up here?

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -5123,3 +5168,783 @@ get_bbf_pivot_tuplestore(const char 	*sourcetext,
...
+
+				nulls[8] = false;
+				values[8] = PointerGetDatum(cstring_to_text((const char *) ptr));
+			}
+		}
+
+		if (!nulls[0] && !bms_is_member(DatumGetInt64(values[0]), *xml_visited_nodes_set))
+		{
+			tuplestore_putvalues(tupstore, tupdesc, values, nulls);
+			*xml_visited_nodes_set = bms_add_member(*xml_visited_nodes_set, DatumGetInt64(values[0]));
```
Deepesh125, line 5624: what if node is already visited? I understand it might not be possible but should we have safeguard to return early before processing next?

Deepesh125 (changes requested):

contrib/babelfishpg_tsql/runtime/functions.c
rohit01010: For `XML_DOCUMENT_NODE` we are sure that an id should not be assigned to it, as parent of root node in the final output should be NULL. But for all the other nodes we are unsure, as in some cases, some of these nodes among `XML_DOCUMENT_TYPE_NODE, XML_DOCUMENT_FRAG_NODE, XML_HTML_DOCUMENT_NODE, XML_ELEMENT_DECL, XML_ATTRIBUTE_DECL, XML_ENTITY_DECL, XML_NAMESPACE_DECL, XML_XINCLUDE_START, XML_XINCLUDE_END, XML_DOCB_DOCUMENT_NODE` have been assigned id which is identified by checking parentid for other nodes. So other than `XML_DOCUMENT_NODE` I had assigned id to all types of nodes.

contrib/babelfishpg_tsql/runtime/functions.c
rohit01010: As per the definition of `vec_push_back`, it expects address of element, instead of the element. Hence passed address of node using &node.

contrib/babelfishpg_tsql/runtime/functions.c
rohit01010: Hash table should be maintained till the entire execution of openxml, so that we can fetch the id of any node while adding details of some node in function `add_node_details`. The hash table is later freed in `openxml_simple` at the end during cleanup.

contrib/babelfishpg_tsql/runtime/functions.c
rohit01010: This is unlikely, it is just a sanity check to return NULL instead of -1, for any unexpected use case for some xml document.

contrib/babelfishpg_tsql/src/procedures.c
ayushdsh: That's right! For ENR temp tables, we can use NoLock for all operations since there's no concurrency. 

[Ref1](https://github.com/babelfish-for-postgresql/postgresql_modified_for_babelfish/blob/BABEL_5_X_DEV__PG_17_X/src/backend/access/common/relation.c#L67-L69), [Ref2](https://github.com/babelfish-for-postgresql/postgresql_modified_for_babelfish/blob/BABEL_5_X_DEV__PG_17_X/src/backend/storage/lmgr/lmgr.c#L335-L336)

contrib/babelfishpg_tsql/runtime/functions.c
rohit01010: TSQL has not documented anything about openxml behaviour with empty string XPATH. whereas my speculation is that it treats empty string XPATH as '/', but I am not very confident about. So kept PostgreSQL behaviour in this case.

test/JDBC/expected/openxml_without_with_clause-vu-verify.out
```diff
@@ -0,0 +1,1155 @@
...
+4#!#0#!#2#!#name#!#attr#!#http://attr.com#!#<NULL>#!#<NULL>#!#<NULL>
+5#!#4#!#3#!##text#!#<NULL>#!#<NULL>#!#<NULL>#!#<NULL>#!#test
+6#!#0#!#2#!#attr#!#xmlns#!#<NULL>#!#<NULL>#!#<NULL>#!#<NULL>
+7#!#6#!#3#!##text#!#<NULL>#!#<NULL>#!#<NULL>#!#<NULL>#!#http://attr.com
+8#!#0#!#3#!##text#!#<NULL>#!#<NULL>#!#<NULL>#!#<NULL>#!#content
+~~END~~
+
+
+-- XML Entities
+DECLARE @xml nvarchar(1000) = '<root>&lt;escaped&gt; &amp; &quot;quoted&quot; &apos;apostrophe&apos;</root>'
```
tanyagupta17, line 242: Could we have a test case like this in nested doc

test/JDBC/expected/openxml_without_with_clause-vu-verify.out
tanyagupta17, line 1: Were there any differences in results from TSQL?

test/JDBC/expected/openxml_without_with_clause-vu-verify.out
```diff
@@ -0,0 +1,1155 @@
...
+EXEC sp_xml_removedocument @handle
+GO
+~~START~~
+bigint#!#bigint#!#int#!#nvarchar#!#nvarchar#!#nvarchar#!#nvarchar#!#bigint#!#ntext
+2#!#0#!#1#!#target#!#<NULL>#!#<NULL>#!#<NULL>#!#<NULL>#!#<NULL>
+3#!#2#!#3#!##text#!#<NULL>#!#<NULL>#!#<NULL>#!#<NULL>#!#found
+~~END~~
+
+
+-- Wildcard Selection
```
tanyagupta17, line 874: Could we have regex in Xpath?

test/JDBC/expected/openxml_without_with_clause-vu-prepare.out
```diff
@@ -0,0 +1,23 @@
+CREATE PROCEDURE openxml_without_with_dep_proc1(@xml nvarchar(max))
```
tanyagupta17, line 1: Could we also have a test with trigger, if possible?

test/JDBC/expected/openxml_without_with_clause-vu-verify.out
```diff
@@ -0,0 +1,1155 @@
...
+</Customers>
+<Customers CustomerID="LILAS" ContactName="Carlos Gonzlez">
+   <Orders CustomerID="LILAS" EmployeeID="3" OrderDate=
+           "1996-08-16T00:00:00">
+      <OrderDetails OrderID="10283" ProductID="72" Quantity="3"/>
+   </Orders>
+</Customers>
+</ROOT>';
+EXEC sp_xml_preparedocument @handle OUTPUT, @xml;
+SELECT * FROM OPENXML(@handle, '/ROOT/Customers') ORDER BY id;
```
tanyagupta17, line 21: Could we have @handle as fixeddecimal/bigint/strings/bit etc? How would that be handled?

What if @xml is a UDT over string datatypes, are those cases handled?

test/JDBC/expected/openxml_without_with_clause-vu-verify.out
rohit01010: In Babelfish, Xpath should follow XPATH 1.0 rules. Ref: https://www.w3.org/TR/1999/REC-xpath-19991116/ for XPATH 1.0 syntax for Xpath.

test/JDBC/expected/openxml_without_with_clause-vu-verify.out
rohit01010: These are following differences that I found when compared with TSQL behaviour.
* XML declaration (for example `<?xml version="1.0" ?>`) is not represented as a node in libxml2's DOM tree, so entries corresponding to it will be not be shown in Babelfish. Although it seems in TSQL it creates a Processing-instruction for this declaration which gives some entries for `<?xml version="1.0" ?>`
* ids assigned to nodes might be different compared to TSQL, as there will be xml tree structure difference in internal implementation.

contrib/babelfishpg_tsql/runtime/functions.c
rohit01010: Libxml2 parser does not remove white spaces in some cases it instead create TEXT nodes for each of such white spaces. we are not interested in these nodes. So we skipped the whitespace-only text node. For example, 
```
'<root>      <child>Rohit</child>        </root>'
```
In above example, libxml2 parser will convert it to nodes as follows
element node: root,
text node:          (blank space)
element node: child
text node: Rohit
text node:          (blank space)

contrib/babelfishpg_tsql/runtime/functions.c
rohit01010: Libxml2 does not provide any API to check whether given node is root node or not. Hence this is the only way to compare whether cur node is root node or not.

contrib/babelfishpg_tsql/runtime/functions.c
rohit01010: This memory is palloc'd in `ExprContext` Memory context which is part of `ExecutorState` Memory context. This `ExecutorState` memory is created and destroyed at query level. So anyway this memory will be deallocated/freed after the function execution. Anyways added explicit pfree in the end.

contrib/babelfishpg_tsql/runtime/functions.c
rohit01010: Updated logic to use local list instead.

contrib/babelfishpg_tsql/runtime/functions.c
rohit01010: Added sanity check

contrib/babelfishpg_tsql/runtime/functions.c
rohit01010: Added a return before processing its attribute and child nodes for already visited node.

contrib/babelfishpg_tsql/runtime/functions.c
rohit01010: Took reference from PostgreSQL `xpath_internal` implementation. Updated it to `VARDATA_ANY` and `VARSIZE_ANY_EXHDR`.

contrib/babelfishpg_tsql/runtime/functions.c
rohit01010: Ack, removed the condition.

contrib/babelfishpg_tsql/runtime/functions.c
rohit01010: Added error handling if return code is non-zero.

contrib/babelfishpg_tsql/src/backend_parser/gram-tsql-rule.y
rohit01010: This case handles openxml cross/outer apply. Ref test cases: https://github.com/rohit01010/babelfish_extensions/blob/ba28fcb4511c3545ca4fcf1bfbfbf3d7b3d6e667/test/JDBC/expected/openxml_without_with_clause-vu-verify.out#L1157

contrib/babelfishpg_tsql/src/procedures.c
rohit01010: We cannot use NoLock here as whenever we do any operation which involve us opening relation `xml_handle_temp_table` inside a procedure following assert [Ref1](https://github.com/babelfish-for-postgresql/postgresql_modified_for_babelfish/blob/BABEL_5_X_DEV__PG_17_X/src/backend/access/common/relation.c#L67-L69)
```
Assert(lockmode != NoLock ||
		   IsBootstrapProcessingMode() ||
		   CheckRelationLockedByMe(r, AccessShareLock, true));
``` 
will fail, as the hook `pltsql_get_tsql_enr_from_oid_hook` will try to find the relation in currentQueryEnv, whereas the table may not exist in currentQueryEnv. Ideally it should have also checked parent queryEnvs as done in `get_ENR`.

ayushdsh: You're right Rohit. I overlooked at the pltsql_get_tsql_enr_from_oid_hook function. Currently, we acquire the default locks for each operation 

select -> access share
insert/delete -> row exclusive 

I think we can change to function pltsql_get_tsql_enr_from_oid_hook to look into the parent queryEnvs as well for a fast path and no contention in temp tables because of no possibility of concurrency. 

@lejaokri @tanscorpio7  any thoughts on this?

test/JDBC/expected/openxml_without_with_clause-vu-prepare.out
rohit01010: Added

test/JDBC/expected/openxml_without_with_clause-vu-verify.out
rohit01010: Added

contrib/babelfishpg_tsql/src/procedures.c
rohit01010: Took a reference from PostgreSQL `parse_xml_decl` function. Here `if (xml_data == NULL && ns_data == NULL)` is just for sanity check to avoid function execution if both of the variables are NULL.

test/JDBC/expected/openxml_without_with_clause-vu-verify.out
rohit01010: Added tests

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -5123,3 +5167,800 @@ get_bbf_pivot_tuplestore(const char 	*sourcetext,
...
+	if (PG_ARGISNULL(1))
+		ereport(ERROR,
+				(errcode(ERRCODE_NULL_VALUE_NOT_ALLOWED),
+				 errmsg("XPath expression cannot be null")));
+
+	xpath_expr_text = PG_GETARG_TEXT_PP(1);
+	xpath_len = VARSIZE_ANY_EXHDR(xpath_expr_text);
+	if (xpath_len == 0)
+		ereport(ERROR,
+				(errcode(ERRCODE_DATA_EXCEPTION),
```
HarshLunagariya: Can't we move this check earlier ? Even before executing get_xml_data_and_namespace_data & extract_namespaces_from_xml.

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -5123,3 +5167,800 @@ get_bbf_pivot_tuplestore(const char 	*sourcetext,
...
+	xmlerrcxt = pg_xml_init(PG_XML_STRICTNESS_ALL);
+
+	PG_TRY();
+	{
+		xmlInitParser();
+
+		ctxt = xmlNewParserCtxt();
+		if (ctxt == NULL || xmlerrcxt->err_occurred)
+			xml_ereport(xmlerrcxt, ERROR, ERRCODE_OUT_OF_MEMORY,
+						"could not allocate parser context");
```
HarshLunagariya, line 5805: Same error for `xmlerrcxt->err_occurred` case as well? Do we know various possibilities?

rohit01010: For all the APIs that we used from libxml2, will not directly throw error instead it just returns NULL. So here `ctxt == NULL` is sufficient for throwing the error. Although I took reference of this part from PostgreSQL `xpath_internal` function where they had this extra check on `xmlerrcxt->err_occurred` so followed the same here.

contrib/babelfishpg_tsql/runtime/functions.c
rohit01010: Moved before executing get_xml_data_and_namespace_data & extract_namespaces_from_xml.

tanyagupta17 (approved):

HarshLunagariya (approved):

contrib/babelfishpg_tsql/runtime/functions.c
Deepesh125: That why we need safety check to avoid memory overwrite

rohit01010: Added safety check

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -5130,3 +5174,800 @@ get_bbf_pivot_tuplestore(const char 	*sourcetext,
...
+										  "OpenXML Context",
+										  ALLOCSET_DEFAULT_SIZES);
+	}
+
+	if (ht_xmlNode2Id == NULL)	/* create hash table */
+	{
+		MemSet(&hashCtl, 0, sizeof(hashCtl));
+		hashCtl.keysize = sizeof(xmlNodePtr);
+		hashCtl.entrysize = sizeof(ht_xmlNode2Id_entry_t);
+		hashCtl.hcxt = TransMemoryContext;
```
Deepesh125: why don't we use same expr context for Htab?

contrib/babelfishpg_tsql/runtime/functions.c
Deepesh125: wouldn't it be better if we can have one structure with all the field? it will be better in terms of readability?

contrib/babelfishpg_tsql/runtime/functions.c
Deepesh125: okay, so we are relying on address comparison

Deepesh125 (changes requested):

contrib/babelfishpg_tsql/runtime/functions.c
rohit01010: Yes

contrib/babelfishpg_tsql/runtime/functions.c
rohit01010: Used CurrentMemoryContext which contains ExprContext.

contrib/babelfishpg_tsql/runtime/functions.c
rohit01010: - Hashtable is defined as global variable and we have defined xml_nodes_list as local variable now.
- Creating a structure for only Hashtable and xml_nodes_list, where xml_nodes_list is just a temporary list which is getting cleaned up immediately after hash table preparation.

Hence I don't think there is much value addition by creating a separate structure for this.

Deepesh125 (dismissed):

Deepesh125 (approved):
Can you please add bit more details on high level implementation in the description?

PR #4093, opened 2025-09-15
Updating Output(Returning) clause code in TSQL after EPQ

herambh-shah:
### Description

We are dealing with a critical crash bug in Babelfish's EPQ (EvalPlanQual) handling during concurrent UPDATE operations with OUTPUT clauses. The crash occurs when two sessions simultaneously modify the same row, where the column is of type "varchar" causing tuple alignment problems during EPQ re-evaluation. This is a memory corruption issue where PostgreSQL's EPQ mechanism fails to properly handle tuple structure changes when Babelfish processes OUTPUT clauses, leading to segmentation faults or assertion failures.

### The Problem
Babelfish was processing the OUTPUT clause (RETURNING clause) BEFORE EPQ evaluation, which created a race condition crash because:

1. Original Design Rationale: OUTPUT clause was moved before EPQ to ensure it executed before triggers
2. Race Condition: When EPQ re-evaluation occurred, the OUTPUT clause had already processed using stale tuple data 
3. Crash Point: EPQ tried to re-evaluate with changed tuple structure, but OUTPUT processing was already complete with wrong memory references

### The Solution
Align Babelfish OUTPUT clause execution with PostgreSQL's native RETURNING clause flow:

### Key Insights from Fix

- [ ]  Trigger Execution Order: Analysis revealed that OUTPUT clause doesn't actually need to run before triggers in Babelfish context
- [ ]  PostgreSQL Alignment: Following PostgreSQL's native RETURNING execution flow eliminates the race condition
- [ ]  EPQ Compatibility: Processing OUTPUT after EPQ ensures it always works with the correct, final tuple data
- [ ]  Memory Safety: Eliminates tuple alignment crashes by ensuring OUTPUT clause processes current data structure

### Related PR - 
Engine PR - https://github.com/babelfish-for-postgresql/postgresql_modified_for_babelfish/pull/623
### Issues Resolved

BABEL-4880

1. Manual testing to check concurrent update/insert/delete crash
2. Added comprehensive Java test cases that utilize multiple concurrent threads to simulate real-world race conditions involving simultaneous UPDATE, DELETE, and INSERT operations with OUTPUT clauses. These multi-threaded test scenarios are designed to reproduce the EPQ-related crash conditions, though the specific crash scenario has a low probability of occurrence due to the precise timing requirements needed to trigger the tuple alignment issue during concurrent modifications.

### Check List


test/JDBC/src/main/java/com/sqlsamples/JDBCBabelOutputConcurrency.java
```diff
@@ -0,0 +1,547 @@
...
+        cxns.add(conn1);
+        cxns.add(conn2);
+        
+        // Add Param2 column for NULL alignment test
+        try (Statement stmt = conn1.createStatement()) {
+            try {
+                stmt.execute("ALTER TABLE EmailPosts ADD Param2 VARCHAR(50)");
+            } catch (SQLException e) {
+                // Column might already exist, ignore
+            }
```
lejaokri, line 210: why not create table with this column right away?

test/JDBC/src/main/java/com/sqlsamples/JDBCBabelOutputConcurrency.java
```diff
@@ -0,0 +1,547 @@
...
+            session1Ready.await();
+            
+            conn.setAutoCommit(false);
+            Statement stmt = conn.createStatement();
+            
+            // This will be blocked until Session 1 commits, then EPQ will process it
+            stmt.execute("UPDATE EmailPosts SET LockProcessID = 7 " +
+                       "OUTPUT deleted.Param1 as old_param1, inserted.Param1 as new_param1, " +
+                       "deleted.LockProcessID as old_lock, inserted.LockProcessID as new_lock " +
+                       "WHERE EmailPostID = 1");
```
lejaokri: please use dummy table names and columns

lejaokri (changes requested):
Let's get a review from Avantika as well.

test/JDBC/src/main/java/com/sqlsamples/JDBCBabelOutputConcurrency.java
herambh-shah: yes we could do that also but I think it' ok to write it this way as Param2 column being used in one of the test case only.

PR #4054, opened 2025-09-01
Fix incorrect results of ASCII function with Binary and Varbinary arguments

sujanM14:
### Description
Earlier in Babelfish, ASCII function returns incorrect results when argument type is Binary and Varbinary
```
1> select ascii(0x65)
2> go
ascii 
-----------
48
```
This PR fixes the implementation of the ASCII function to correctly handle Binary and Varbinary inputs, aligning the output with t-sql behavior.
```
1> select ascii(0x65);
2> go
ascii      
-----------
        101
```
Signed-off-by: Sujan Mujawar [sujanmjr@amazon.com](mailto:sujanmjr@amazon.com)

### Issues Resolved

BABEL-5809

### Test Scenarios Covered ###
* **Use case based -** YES

* **Boundary conditions -** YES

* **Arbitrary inputs -** YES

* **Negative test cases -** YES

* **Minor version upgrade tests -**

* **Major version upgrade tests -**

* **Performance tests -**

* **Tooling impact -**

* **Client tests -**

### Check List
- [x] Commits are signed per the DCO using --signoff 

By submitting this pull request, I confirm that my contribution is under the terms of the Apache 2.0 and PostgreSQL licenses, and grant any person obtaining a copy of the contribution permission to relicense all or a portion of my contribution to the PostgreSQL License solely to contribute all or a portion of my contribution to the PostgreSQL open source project.

For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/babelfish-for-postgresql/babelfish_extensions/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #4101, opened 2025-09-18
Fix incorrect results of ASCII function with Binary and Varbinary arguments

sujanM14:
### Description
Earlier in Babelfish, ASCII function returns incorrect results when argument type is Binary and Varbinary
```
1> select ascii(0x65)
2> go
ascii 
-----------
48
```
This PR fixes the implementation of the ASCII function to correctly handle Binary and Varbinary inputs, aligning the output with t-sql behavior.
```
1> select ascii(0x65);
2> go
ascii      
-----------
        101
```
Cherry-pick PR : #4054 
Signed-off-by: Sujan Mujawar [sujanmjr@amazon.com](mailto:sujanmjr@amazon.com)

### Issues Resolved

BABEL-5809

* **Use case based -** YES

* **Arbitrary inputs -** YES

* **Performance tests -**

* **Tooling impact -**

* **Client tests -**

### Check List

### DIFF
--- a/contrib/babelfishpg_tsql/sql/sys_functions.sql
+++ b/contrib/babelfishpg_tsql/sql/sys_functions.sql
@@ -3335,6 +3335,50 @@ BEGIN
 END;
 $$ LANGUAGE plpgsql IMMUTABLE STRICT PARALLEL SAFE;
 
+-- wrapper functions for ascii --
+CREATE OR REPLACE FUNCTION sys.ascii(ANYELEMENT)
+RETURNS INTEGER
+AS $$
+DECLARE
+    arg_datatype text;
+    basetype oid;
+BEGIN
+    arg_datatype := sys.translate_pg_type_to_tsql(pg_typeof($1)::oid);
+    IF arg_datatype IS NULL THEN
+        -- for User Defined Datatype, use immediate base type to check for argument datatype validation
+        basetype := sys.bbf_get_immediate_base_type_of_UDT(pg_typeof($1)::oid);
+        arg_datatype := sys.translate_pg_type_to_tsql(basetype);
+    END IF;
+
+    -- restricting arguments with invalid datatypes for ascii function
+    IF arg_datatype IN ('image', 'sql_variant', 'xml', 'geometry', 'geography') THEN
+        RAISE EXCEPTION 'Argument data type % is invalid for argument 1 of ascii function.', arg_datatype;
+    END IF;
+    
+    IF arg_datatype IN ('binary', 'varbinary') THEN
+        IF len($1) = 0 THEN
+            RETURN NULL;
+        END IF;
+    ELSE
+        IF length($1::TEXT) = 0 THEN
+            RETURN NULL;
+        END IF;
+    END IF;
+    RETURN pg_catalog.ascii(CAST($1 AS sys.VARCHAR));
+END;
+$$ LANGUAGE plpgsql IMMUTABLE STRICT PARALLEL SAFE;
+
+CREATE OR REPLACE FUNCTION sys.ascii(TEXT)
+RETURNS INTEGER
+AS $$
+BEGIN
+    IF length($1) = 0 THEN
+        RETURN NULL;
+    END IF;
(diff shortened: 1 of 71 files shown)


tanyagupta17 (approved):

contrib/babelfishpg_tsql/sql/sys_functions.sql
```diff
@@ -3335,6 +3335,50 @@ BEGIN
...
+    END IF;
+    RETURN pg_catalog.ascii(CAST($1 AS sys.VARCHAR));
+END;
+$$ LANGUAGE plpgsql IMMUTABLE STRICT PARALLEL SAFE;
+
+CREATE OR REPLACE FUNCTION sys.ascii(TEXT)
+RETURNS INTEGER
+AS $$
+BEGIN
+    IF length($1) = 0 THEN
```
pranavJ23, line 3375: shouldn't we throw error in the case of text ?
eg query: 
```
select ascii(cast('abc' as text))
```
let's verify it once

contrib/babelfishpg_tsql/sql/sys_functions.sql
```diff
@@ -3335,6 +3335,50 @@ BEGIN
...
+        RAISE EXCEPTION 'Argument data type % is invalid for argument 1 of ascii function.', arg_datatype;
+    END IF;
+    
+    IF arg_datatype IN ('binary', 'varbinary') THEN
+        IF len($1) = 0 THEN
+            RETURN NULL;
+        END IF;
+    ELSE
+        IF length($1::TEXT) = 0 THEN
+            RETURN NULL;
```
pranavJ23, line 3364: I suggest to cast into sys.varchar instead of text, it would be better.
eg we want to consider 0x of length 0 but for following queries:
```
1> select length(cast(0x as varchar))
2> go
length     
-----------
          0

(1 rows affected)
1> select length(cast(0x as text))
2> go
length     
-----------
          2
```

contrib/babelfishpg_tsql/sql/sys_functions.sql
```diff
@@ -3335,6 +3335,50 @@ BEGIN
...
+        arg_datatype := sys.translate_pg_type_to_tsql(basetype);
+    END IF;
+
+    -- restricting arguments with invalid datatypes for ascii function
+    IF arg_datatype IN ('image', 'sql_variant', 'xml', 'geometry', 'geography') THEN
+        RAISE EXCEPTION 'Argument data type % is invalid for argument 1 of ascii function.', arg_datatype;
+    END IF;
+    
+    IF arg_datatype IN ('binary', 'varbinary') THEN
+        IF len($1) = 0 THEN
```
pranavJ23, line 3359: Do we have any query hitting this case ?

contrib/babelfishpg_tsql/sql/sys_functions.sql
sujanM14: yes, but 
```
1> select length(cast(0x as varchar))
2> select length(cast(0x00 as varchar))
3> go
length     
-----------
          0

(1 rows affected)
length     
-----------
          0

```

this resulting in same output for ascii which is incorrect. Thus using `len` for `binary` and `varbinary`. Whereas I will change typecast to `varchar` instead of `text`.

contrib/babelfishpg_tsql/sql/sys_functions.sql
sujanM14: yes for case where 
```
1> select ascii(0x);
2> go
           
-----------
       NULL
```

contrib/babelfishpg_tsql/sql/sys_functions.sql
sujanM14: Currently, TEXT inputs bypass the data type validation in ascii(ANYELEMENT) because they're being handled directly by the pg catalog ascii function. Thus it will require to create an extra wrapper function for ascii(text) which can be taken as improvement as discussed.

pranavJ23 (approved):

shardgupta (approved):

rohit01010 (approved):

PR #4121, opened 2025-09-24
Fix precision/scale of numeric columns in SELECT INTO with GROUP BY.

Yvinayak07:
### 1. Issues

If a GROUP BY clause is used in a SELECT statement, the target list during initialisation of TupleDesc will contains

- all target columns mentioned in SELECT statement
- columns from the GROUP BY clause (In this case column "id")
The columns from GROUP BY clause which is not there in target columns are marked as junk column, which will be filtered out later. During initialisation of TupleDesc we are computing and storing the correct typmod information for all the columns in target list

Later after TupleDesc is initialised, if any of the column is marked as junk, Junk filtration is done which recomputes the entire TupleDesc which leads to losing the typmod information of avgValue column.

Earlier we used to return typmod as `-1` if plan was NULL for `T_var` node whereas correct typmod was preserved in `var->vartypmod`. So we fixed that portion in thi PR.

### Issues Resolved

Task: BABEL-6081

* **Use case based -**

* **Boundary conditions -**

* **Arbitrary inputs -**

* **Negative test cases -**

* **Performance tests -**

* **Tooling impact -**

* **Client tests -**

### Check List


KushaalShroff (dismissed):

contrib/babelfishpg_tsql/src/pltsql_coerce.c
```diff
@@ -1434,8 +1434,10 @@ resolve_numeric_typmod_from_exp(Plan *plan, Node *expr, bool *found)
 				/* if varno is INNER_VAR or OUTER_VAR then we need plan, else we cannot find typmod, hence set found as false and return -1 */
 				if (plan == NULL && (var->varno == INNER_VAR || var->varno == OUTER_VAR))
 				{
-					if (found != NULL) *found = false;
-					return -1;
+					if (var->vartypmod == -1)
+						if (found != NULL) *found = false;
+
+					return var->vartypmod;
```
rohit01010: This will fix cases where plan is not required for `resolve_numeric_typmod_from_exp` to find the typmod. But in case where plan is required for `resolve_numeric_typmod_from_exp` to find the typmod, how are we handling those cases. For example, following case should still store NULL for precision and scale.
```
CREATE TABLE BABEL_5454_T10 (id NUMERIC(4, 2));
GO
INSERT INTO BABEL_5454_T10 VALUES (1.0)
GO

WITH 
cte1 AS (
    SELECT id, CAST(id AS INT) AS int_id
    FROM BABEL_5454_T10
),
cte2 AS (
    SELECT DISTINCT int_id
    FROM cte1
)
SELECT 
    c1.id * avg(c2.int_id) AS product
INTO
    rohit_table
FROM cte1 c1
JOIN cte2 c2 ON c1.int_id = c2.int_id
GROUP BY c1.id;
GO

select TABLE_CATALOG, TABLE_SCHEMA, TABLE_NAME, COLUMN_NAME, NUMERIC_PRECISION, NUMERIC_PRECISION_RADIX, NUMERIC_SCALE
from information_schema.columns 
where TABLE_NAME = 'rohit_table' order by COLUMN_NAME
GO
```
Can you please verify above query ?

Yvinayak07: We need `fixed length handling` as well here.

Yvinayak07: added test cases for that and even handled the case of UDT's.

rohit01010 (approved):

contrib/babelfishpg_tsql/src/pltsql_coerce.c
```diff
@@ -1357,6 +1357,8 @@ resolve_numeric_typmod_from_exp(Plan *plan, Node *expr, bool *found)
 				TargetEntry	*tle;
 				int		rettypmod;
 				bool		found_typmod;
+				int32 		fixlen_default_typmod;
+				Oid			immediate_base_type;
```
KushaalShroff: Why are these datatypes declared in this scope?

KushaalShroff (dismissed):

contrib/babelfishpg_tsql/src/pltsql_coerce.c
Yvinayak07: For fixed length we earlier used to have that condition inside if (plan) condition but now we need this for UDT's also but for immediate_base_type we could have kept it inside UDT condition only.

Yvinayak07: For fixed length we earlier used to have that condition inside if (plan) condition but now we need this for UDT's also but for immediate_base_type we could have kept it inside UDT condition only. 
addressed

KushaalShroff (approved):

PR #4174, opened 2025-10-12
Added tests for fix of sql_dialect check in ENRGetSystableScan() function for T-SQL temp tables

ayushdsh:
### Description
In ENRGetSystableScan, if the dialect is not TSQL_DIALECT and there is a scan initiated for a catalog (other than Sequence), we return false i.e. the tuple being scanned for does not exist in ENR. But this is not reliable in case this scan is initiated from a PG function since the dialect then would be PG_DIALECT and the scan in ENR would return false, even though the actual tuple being searched for is in ENR (in case it's a T-SQL temp table). This results in a "could not open a relation" error since the tuple also does not exist in pg_catalog.

Engine PR - https://github.com/babelfish-for-postgresql/postgresql_modified_for_babelfish/pull/632

### Issues Resolved
BABEL-6099

### Check List


tanscorpio7 (approved):

test/JDBC/expected/temp_table_dialect_check-before-15_12-or-16_8-vu-prepare.out
```diff
@@ -0,0 +1,80 @@
...
+~~END~~
+
+
+SELECT COUNT(REPLACE(Description, '55', '44')) FROM #tmpGoodsInfo;
+GO
+~~START~~
+int
+~~ERROR (Code: 33557097)~~
+
+~~ERROR (Message: could not open relation with OID 86045)~~
```
lejaokri, line 78: is this error message expected? Why does it complain with oid but not with relname does not exist?

ayushdsh: Yes. This is thrown at the time when during a plpgsql function, the backend tries to rebuild it relcache because it was invalidated. This is thrown from 

```
Relation
relation_open(Oid relationId, LOCKMODE lockmode)
{
	Relation	r;

	Assert(lockmode >= NoLock && lockmode < MAX_LOCKMODES);

	/* Get the lock before trying to open the relcache entry */
	if (lockmode != NoLock)
		LockRelationOid(relationId, lockmode);

	/* The relcache does all the real work... */
	r = RelationIdGetRelation(relationId);

	if (!RelationIsValid(r))
		elog(ERROR, "could not open relation with OID %u", relationId);
```
in `relation.c`

lejaokri: but where/how did we get the relationId? I see a drop temp table there, so we should not see relationId anymore if we removed it from ENR

ayushdsh: We do a `DROP IF EXISTS` first and then do a `SELECT INTO temp_table`. Then there comes a relcache invalidation and during the next `SELECT COUNT()`  , it tries to open the toast table to detoast the value while in `REPLACE` function which is a plpgsql function. And hence the error.

> I see a drop temp table there, so we should not see relationId anymore if we removed it from ENR
We do a `SELECT INTO`  to create a new temp table after it.

lejaokri: I get it now. This is before the fix

lejaokri (approved):

PR #4190, opened 2025-10-15
Fixing fluctuating errors and crash during self assignment of parameters.

tanyagupta17:
### Description

Problem: Previously, during variable self-assignment (e.g., SELECT @var = @var), Babelfish reused the same Datum reference for both source and target without creating a copy. This caused memory corruption, use-after-free errors, and crashes due to unsafe sharing across memory contexts.

Root Cause: The source and target pointed to the same memory location, which could be invalidated by context switches or garbage collection.

Solution: Introduced safe duplication using datumCopy() to ensure each variable has its own memory allocation in the correct context (estate->datum_context). This preserves data integrity and prevents invalid memory access.

### Issues Resolved

BABEL-6119, BABEL-5947

* **Use case based -** Yes

* **Arbitrary inputs -** Yes

* **Performance tests -** NA

* **Tooling impact -** NA

* **Client tests -** NA

### Check List


contrib/babelfishpg_tsql/src/pl_exec.c
```diff
@@ -10561,7 +10561,18 @@ pltsql_assign_var(PG_FUNCTION_ARGS)
...
+		bool		resTypByVal;
+		
+		/* Fetch the typlen and typbyval info for the arg type. */
+		get_typlenbyval(valtype, &valTypLen, &resTypByVal);
+		/* Copy the datum. */
+		data = datumCopy(data, resTypByVal, valTypLen);
+	}
+
+	/* We will reuse exec_assign_value function here provided in pl_exec.c */
 	exec_assign_value(estate, target, data, isNull, valtype, valtypmod);
```
Deepesh125, line 10576: Did we verify all the consumed of `exec_assign_value` to see if they are feeding datums correctly?

test/JDBC/expected/self-assign-var.out
```diff
@@ -0,0 +1,598 @@
...
+~~END~~
+
+
+-- JIRA query [BABEL-6119]
+CREATE TABLE babel_6119_t1 (ACC_ACCOUNT NVARCHAR(30) PRIMARY KEY,ACC_PROPERTY NVARCHAR(30)); 
+CREATE TABLE babel_6119_t2 (ORG_CODE NVARCHAR(30) PRIMARY KEY,ORG_NAME NVARCHAR(20));
+go
+
+INSERT INTO babel_6119_t1 VALUES ('1000013', 'ORG001'); 
+INSERT INTO babel_6119_t2 VALUES ('ORG001', 'Testing org1');
```
Deepesh125, line 315: can you also insert some values where data gets toasted and self assigning the variables work fine?

Deepesh125 (changes requested):

test/JDBC/expected/self-assign-var.out
rohit01010: +1. Also can we test "setting local variable to table column with toasted value, then updating/deleting that row from the table, and then checking the variable value or using that variable in some other operation".

contrib/babelfishpg_tsql/src/pl_exec.c
```diff
@@ -10561,7 +10561,18 @@ pltsql_assign_var(PG_FUNCTION_ARGS)
...
+	if (!isNull)
+	{
+		int16		valTypLen;
+		bool		resTypByVal;
+		
+		/* Fetch the typlen and typbyval info for the arg type. */
+		get_typlenbyval(valtype, &valTypLen, &resTypByVal);
+		/* Copy the datum. */
+		data = datumCopy(data, resTypByVal, valTypLen);
+	}
```
rohit01010, line 10573: For every local variable assignment we will be creating a copy. so if there are lots of self assignments of same variable to itself and each time we will be creating a copy, do we have mechanism to cleanup the old datum, to avoid memory bloating ?

test/JDBC/expected/self-assign-var.out
rohit01010, line 1: do we have tests as follows ?
```
select @v1 = @v2;
select @v2 = 'pqr';
select @v1; 
```
Here before our fix both @v1 and @v2 should have been pointing to the same memory and updating @v2 might corrupt data pointer by @v1. can you please add such tests if not already exists.

rohit01010, line 1: Have we added tests as follows ?
```
select @v1 = NULL
select @v2 = @v1
select @v1 = 'p'
select @v2
```

test/JDBC/expected/self-assign-var.out
tanyagupta17: Had added a few as per BABEL-5947 earlier, added more [here](https://github.com/amazon-aurora/babelfish_extensions/blob/e6d379018129849a51eff07f2f191bb31a698ed2/test/JDBC/expected/self-assign-var.out#L338)

test/JDBC/expected/self-assign-var.out
tanyagupta17: Yes such test cases are also fixed with this PR. Added tests [here](https://github.com/amazon-aurora/babelfish_extensions/blob/e6d379018129849a51eff07f2f191bb31a698ed2/test/JDBC/expected/self-assign-var.out#L88)

tanyagupta17: Added [here](https://github.com/amazon-aurora/babelfish_extensions/blob/e6d379018129849a51eff07f2f191bb31a698ed2/test/JDBC/expected/self-assign-var.out#L111)

contrib/babelfishpg_tsql/src/pl_exec.c
tanyagupta17: Yes, we have cleanup mechanism in exec_assign_value, there we are freeing the existing datum the variable is pointing to and assigning it to new datum that we passed as argument, so old datum will get cleaned up.

contrib/babelfishpg_tsql/src/pl_exec.c
tanyagupta17: Yes, verified. In other occurrences where we are consuming `exec_assign_value`, we are either making a copy internally using various pg functions or we are passing non-varlena type (int32, int64 etc) as third argument.

Deepesh125 (approved):

rohit01010 (approved):

PR #4192, opened 2025-10-15
Prevent data corruption in INSERT...EXECUTE with incomplete column lists

fizzlr:
Block INSERT...EXECUTE statements where the explicit column list has fewer columns than the target table. This prevents data corruption and server hang/crash issues in such cases.

### Description
INSERT...EXECUTE statements with explicit column lists that specify fewer columns than the target table has were causing data corruption and potential server hangs/crashes.

Example:

CREATE TABLE target_table (col1 INT, col2 INT, col3 INT);
CREATE PROCEDURE proc1 AS BEGIN SELECT 1, 2, 3; END;

-- This would cause corruption:
INSERT INTO target_table (col1, col2) EXECUTE proc1;

Thus, added validation at two critical points to catch all INSERT...EXECUTE patterns:

1. Parse-Time Validation(pltsql_post_parse_analyze)

- Catches top-level statements: INSERT INTO table (cols) EXECUTE proc and INSERT INTO table (cols) EXECUTE @stmt
- Validates CallStmt (procedure calls) and ExecuteStmt (prepared statements)
- Checks: insert_col_count < table_col_count

2. Execution-Time Validation(pltsql_inline_handler)

- Catches dynamic SQL: EXEC('INSERT INTO table (cols) EXECUTE proc') and sp_executesql calls
- Uses codeblock->attrnos and codeblock->relation to identify INSERT context
- Same validation: insert_col_count < table_col_count

### Check List


contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -1135,6 +1135,47 @@ pltsql_post_parse_analyze(ParseState *pstate, Query *query, JumbleState *jstate)
...
+			else
+			{
+				insert_col_count = list_length(query->targetList);
+			}
+
+			/*
+			 * Block when INSERT specifies fewer columns than table has, as positional
+			 * mapping doesn't honor the explicit column list causing data corruption.
+			 */
+			if (insert_col_count < table_col_count)
```
forestkeeper, line 1170: This condition will block : 
```
create procedure P
as
select 'A', 'B'

go

create table T
(
  A char(1), 
  B char(1),
  I int default 10
);

insert into T(A, B)
execute P;

select *
from T;
```

But for now it's not a bug, so will cause regression.

Issue #4129, opened 2025-09-30
[Bug]: Conversion from Query-Based Assigned String types to SQLVARIANT fails.


berkayozkarhan:
### What happened?

Convert query-based assigned string type to sql variant type. It seems like it is blocking client, and also server, by entering forever loop.

```sql
DECLARE @result_sql_variant sql_variant;
SET @result_sql_variant = (
    SELECT TOP 1 value
    FROM STRING_SPLIT(N'A,B,C', N',')
);
SELECT @result_sql_variant AS Result,
       SQL_VARIANT_PROPERTY(@result_sql_variant, 'BaseType') AS BaseType,
       SQL_VARIANT_PROPERTY(@result_sql_variant, 'MaxLength') AS MaxLength;
```

<img width="786" height="883" alt="Image" src="https://github.com/user-attachments/assets/99eadd3e-2606-4f8c-9312-d40dabd37e26" />

When tried with direct assignment, it works fine.

```sql
-- Initialize nvarchar typed variable with N'A'
DECLARE      @testVar1 nvarchar(10);
SET          @testVar1 = N'A';
-- Initialize nvarchar typed variable with "SELECT TOP 1 value FROM STRING_SPLIT(N'A,B,C', N',')"
DECLARE      @testVar2 nvarchar(10);
SET          @testVar2 = (SELECT TOP 1 value FROM STRING_SPLIT(N'A,B,C', N','));
-- Initialize sql variant type variable with first variable
DECLARE      @result_sql_variant sql_variant;
SET          @result_sql_variant = ( @testVar1 );
-- Initialize sql variant type variable with second variable
-- SET          @result_sql_variant = ( @testVar2 );
-- Show the result
SELECT      @result_sql_variant AS Result;
```

<img width="662" height="456" alt="Image" src="https://github.com/user-attachments/assets/b88eae87-0c50-446d-8274-bbe9f10aa65f" />

Function name: nvarchar2sqlvariant

Issue: Header size comes with different size with query based assignment.

```sql
DECLARE @testVar1 nvarchar(10);
SET @testVar1 = N'A';

DECLARE @result_sql_variant sql_variant;
SET @result_sql_variant = ( @testVar1 );

SELECT @result_sql_variant as Result;
```

Result: The data comes with 4 byte sized header.

```sql
DECLARE @testVar1 nvarchar(10);
SET @testVar1 = (SELECT N'A');

DECLARE @result_sql_variant sql_variant;
[...]

PR #4198, opened 2025-10-17
Fix string types to sqlvariant conversion with query based initialization.

berkayozkarhan:
### Description

This PR fixes the bug in Babelfish where **implicit conversions from string types to the sqlvariant type** were being handled incorrectly during **query-based initialization.**

### Why was the change made 

The problem occurred with SQL queries containing implicit conversions from string types(specifically those starting with 'SELECT') during the initialization process of sqlvariant types. These conversions were not being processed correctly, leading to potential errors or incorrect behavior.

The root cause was about wrong calculation of the data's length in 'TdsSendTypeSqlvariant' function. The correct modifications have been done for char, varchar, nchar, nvarchar and binary types.

### How was the code changed?

* Data length calculated with appropriate macros
* Buffer start address calculated with appropriate macros
* Regress tests added for different size of inputs(for SHORT headers and others.)

### Issues Resolved

* Fixes Initializing sqlvariant type variables from string type variables with query based init
* Fixes client block bug occurs by initialization

**Related Issue:** https://github.com/babelfish-for-postgresql/babelfish_extensions/issues/4129

**Use case based**
* Verification of conversion from string types to sqlvariant type, during query based initialization
* Verification of client block bug fix.

**Boundary conditions**
* Verified with empty strings and long size strings.
* Verified conversion with literals and variables

**Arbitrary inputs**
* Covered conversions for all string types(char, nchar, varchar, nvarchar)

### Check List


tanscorpio7:
@berkayozkarhan is the pull request in progress ?

berkayozkarhan:
> @berkayozkarhan is the pull request in progress ?

It is ready. It needs to be reviewed.

berkayozkarhan:
@tanscorpio7 review please.

berkayozkarhan:
@tanscorpio7 Review please.

contrib/babelfishpg_common/src/sqlvariant.c
```diff
@@ -338,10 +389,19 @@ gen_sqlvariant_bytea_from_type_datum(size_t typcode, Datum data)
 
 	bytea	   *result;
 	size_t		result_len;
+	Datum 		normalized_data= (Datum)0;
 
 	if (IS_STRING_TYPE(typcode) || IS_BINARY_TYPE(typcode) || typcode == NUMERIC_T) /* varlena datatype */
 	{
-		data_len = VARSIZE_ANY(data);
```
tanscorpio7, line 344: Could you please check if using `VARSIZE_ANY_EXHDR` instead of `VARSIZE_ANY` fixes the issue.

tanscorpio7 (changes requested):

contrib/babelfishpg_common/src/sqlvariant.c
berkayozkarhan: It does not fix the issue. VARSIZE_ANY_EXHDR  macro calculates only data length without header. The main issue here is header is 1 byte when query based initialization occured. Normalizing it to 4 byte fixes the issue, the rest is same as before.

tanscorpio7:
@berkayozkarhan I did some digging into this and I think the type of datum we write should not be the problem instead we are reading the data incorrectly in `TdsSendTypeSqlvariant`
```
	else if (isBaseChar)
	{
		/*
		 * dataformat: totalLen(4B) + baseType(1B) + metadatalen(1B) +
		 * encodingLen(5B) + dataLen(2B) + data(dataLen)
		 */
		StringInfoData strbuf;
		int			actualDataLen = 0;	/* Number of bytes that would be
										 * needed to store given string in
										 * given encoding. */
		char	   *destBuf = NULL;

		dataLen -= VARHDRSZ;
		if (variantBaseType == VARIANT_TYPE_NCHAR ||
			variantBaseType == VARIANT_TYPE_NVARCHAR)
		{
			initStringInfo(&strbuf);
			TdsUTF8toUTF16StringInfo(&strbuf, buf + VARHDRSZ, dataLen);
			actualDataLen = strbuf.len;
		}
		else
		{
			/*
			 * TODO: [BABEL-1069] Remove collation related hardcoding from
			 * sql_variant sender for char class basetypes
			 */
			if (dataLen > 0)
				destBuf = TdsEncodingConversion(buf + VARHDRSZ, dataLen, PG_UTF8, PG_WIN1252, &actualDataLen);
			else
				/* We can not assume that buf would be NULL terminated. */
				actualDataLen = 0;
		}
```

The assumption that datum header will be of size `VARHDRSZ` is wrong. Instead we should use PostgreSQL provided macros for fetching the datalen and data from a datum. So,
1. `dataLen -= VARHDRSZ;` ---> `dataLen = VARSIZE_ANY_EXHDR(VARDATA_ANY(vlena) + variantHeaderLen);`
2. `buf + VARHDRS` --> `VARDATA_ANY(VARDATA_ANY(vlena) + variantHeaderLen)`
Plus any other place were we wrongly assume that header is of length VARHDRSZ.

pranavJ23:
Upon debugging it seems that the `destBuf` (variable as per code, data after corresponding encoding conversion) is getting corrupted due to the wrong assumption about header size.
Agree with @tanscorpio7  that we just need to change dataLen with function `VARSIZE_ANY_EXHDR` & move buf pointer ahead via `VARDATA_ANY`. 
Just one thing to add, we simply do it as : 
`dataLen -= VARHDRSZ;` --->  `dataLen = VARSIZE_ANY_EXHDR(buf);` and 
`buf + VARHDRS `-->  `VARDATA_ANY(buf)`  

Reason: at line 3803 we are already incrementing buf pointer as : 
```
buf += variantHeaderLen;
```
So in structure of buf we have header and data only, so to retrieve data we just need to move buf pointer according to the header size.
Note: we need to follow this for both nchar, nvarchar as well as varchar case.

berkayozkarhan:
> @berkayozkarhan I did some digging into this and I think the type of datum we write should not be the problem instead we are reading the data incorrectly in `TdsSendTypeSqlvariant`
> 
> ```
> 	else if (isBaseChar)
> 	{
> 		/*
> 		 * dataformat: totalLen(4B) + baseType(1B) + metadatalen(1B) +
> 		 * encodingLen(5B) + dataLen(2B) + data(dataLen)
> 		 */
> 		StringInfoData strbuf;
> 		int			actualDataLen = 0;	/* Number of bytes that would be
> 										 * needed to store given string in
> 										 * given encoding. */
> 		char	   *destBuf = NULL;
> 
> 		dataLen -= VARHDRSZ;
> 		if (variantBaseType == VARIANT_TYPE_NCHAR ||
> 			variantBaseType == VARIANT_TYPE_NVARCHAR)
> 		{
> 			initStringInfo(&strbuf);
> 			TdsUTF8toUTF16StringInfo(&strbuf, buf + VARHDRSZ, dataLen);
> 			actualDataLen = strbuf.len;
> 		}
> 		else
> 		{
> 			/*
> 			 * TODO: [BABEL-1069] Remove collation related hardcoding from
> 			 * sql_variant sender for char class basetypes
> 			 */
> 			if (dataLen > 0)
> 				destBuf = TdsEncodingConversion(buf + VARHDRSZ, dataLen, PG_UTF8, PG_WIN1252, &actualDataLen);
> 			else
> 				/* We can not assume that buf would be NULL terminated. */
> 				actualDataLen = 0;
> 		}
> ```
> 
> The assumption that datum header will be of size `VARHDRSZ` is wrong. Instead we should use PostgreSQL provided macros for fetching the datalen and data from a datum. So,
> 
> 1. `dataLen -= VARHDRSZ;` ---> `dataLen = VARSIZE_ANY_EXHDR(VARDATA_ANY(vlena) + variantHeaderLen);`
> 2. `buf + VARHDRS` --> `VARDATA_ANY(VARDATA_ANY(vlena) + variantHeaderLen)`
>    Plus any other place were we wrongly assume that header is of length VARHDRSZ.

I checked here. I think the problem is not here. It is not about the send, it is about acting like data has 4 bytes of header, but it is not.

In nvarchar2sqlvariant function, there are two types of data in our scenario.
1. Data comes from query based initialization
2. Data comes from literal expression

These two types differ in header size. From query based initialization, it comes with:
1 Byte Header + N Byte data

From literal expression, it comes with:
4 Byte Header + N Byte Data


If you ask about how you identify that problem is not in TDS, you can try these queries given below.


```sql

-- Initialize nvarchar typed variable with N'A'
DECLARE      @testVar1 nvarchar(10);
SET          @testVar1 = N'ABCDEF';

-- Initialize nvarchar typed variable with "SELECT TOP 1 value FROM STRING_SPLIT(N'A,B,C', N',')"
DECLARE      @testVar2 nvarchar(10);
SET          @testVar2 = (SELECT TOP N'ABCDEF');


-- Initialize sql variant type variable with first variable
DECLARE      @result_sql_variant sql_variant;
SET          @result_sql_variant = ( @testVar1 );

-- Initialize sql variant type variable with second variable
-- SET          @result_sql_variant = ( @testVar2 );

-- Show the result
SELECT      @result_sql_variant AS Result;

```

image


image (1)




When you initialize your sql variant variable with testVar2, you will see that the output is 'DEF'. This is because of it handles the data same as it's 4 byte header version. 3 byte offset shift occurs here, and the reason is it doesn't have any action for 1 byte header sized datas.


The other types are coming with 4 byte header, even though you initialize with query based init. But string types, there are two different data, as mentioned above.

In my opinion, the correct solution is normalizing the data as 4 byte header data. The problem may be occur before nvarchar2sqlvariant function, but not after. We may solve this by manipulating the data by setting it's header to 4 byte before nvarchar2sqlvariant function, but not after. It is not about send, the tds problem differs from these situation. If it would be in tds layer, you wouldn't success these two queries at the same time, but when your input size is greater than 3 bytes, it starts to generate output.

The summary is query based initialization doesn't fail all the time. There are some scenarios that it generates output, but in wrong way. If the size is greater than 3, it starts to generate input, with some missing characters.

Explanation:

1. Data comes from literal expression:
    4 byte header + ABCD
    Output: ABCD
    It starts from 4th byte of the data.

2. Data comes from query based initialization
    1 Byte Header + ABCD
    Output: D
    It starts from 4th byte of the data. As result, we see it like the data has been shifted. But it is not, ABC has been located at expected header's location. This is why the result starts from 'D' character.




```sql
DECLARE @result_sql_variant sql_variant;
SET @result_sql_variant = (
    SELECT TOP 1 value
    FROM STRING_SPLIT(N'ABCD,B,C', N',')
);
SELECT @result_sql_variant AS Result;

```


```sql
-- Initialize nvarchar typed variable with N'A'
DECLARE      @testVar1 nvarchar(10);
SET          @testVar1 = (SELECT N'ABCDEF');

-- Initialize sql variant type variable with first variable
DECLARE      @result_sql_variant sql_variant;
SET          @result_sql_variant = ( @testVar1 );

SELECT      @result_sql_variant AS Result;

-- The output will be 'DEF'

```

tanscorpio7:
Fixing it during write will not fix the issue for existing users who already have a non 4 byte sql variant datum in their tables. Some more advanatages of fixing it during read is
1. We do not use more space than required by keeping 1byte header as 1byte.
2. There are some other place that create a sql variant datum, we need to enforce 4 byte datum everywhere. This is more maintainence without any adavantage.
3. We can avoid low level code for datum conversion.

berkayozkarhan:
@tanscorpio7 @pranavJ23

You guys are right. There is no need to normalize it to 4 byte again, it waste of memory, it is waste of effort. I tried your solution, and it worked well. I will push the changes, thank you. The solution is simpler than I thought :)

Thanks.

contrib/babelfishpg_tds/src/backend/tds/tdstypeio.c
```diff
@@ -3851,12 +3851,12 @@ TdsSendTypeSqlvariant(FmgrInfo *finfo, Datum value, void *vMetaData)
...
 		char	   *destBuf = NULL;
 
-		dataLen -= VARHDRSZ;
+		dataLen = VARSIZE_ANY_EXHDR(VARDATA_ANY(vlena) + variantHeaderLen);
 		if (variantBaseType == VARIANT_TYPE_NCHAR ||
 			variantBaseType == VARIANT_TYPE_NVARCHAR)
 		{
 			initStringInfo(&strbuf);
-			TdsUTF8toUTF16StringInfo(&strbuf, buf + VARHDRSZ, dataLen);
+			TdsUTF8toUTF16StringInfo(&strbuf, VARDATA_ANY(VARDATA_ANY(vlena) + variantHeaderLen), dataLen);
```
tanscorpio7: Lets also verify the if any other lines in this function also need this change. For example
```
destBuf = TdsEncodingConversion(buf + VARHDRSZ, dataLen, PG_UTF8, PG_WIN1252, &actualDataLen);
```
&
```
dataLen = dataLen - VARHDRSZ;
```
Please aldo add test cases for the same, thanks.

tanscorpio7 (changes requested):

contrib/babelfishpg_tds/src/backend/tds/tdstypeio.c
berkayozkarhan: I've modified the code in accordance with your review. Please take a look.

Thanks.

tanscorpio7: Yes, but my point is we need similar change at other places in `TdsSendTypeSqlvariant`. Please review all uses of `VARHDRSZ` in this function.

pranavJ23: @berkayozkarhan can verify once if we get correct output for varchar, varbinary cases, after your changes ?
to test you can use:
```
DECLARE @v SQL_VARIANT
SET @v = (SELECT 'testaregood')
select @v
```

for varbinary:

```
DECLARE @v SQL_VARIANT
SET @v = (SELECT 0x45621495938)
select @v
```
If not we can do similar fixes for these cases as well.
And add the corresponding tests in testfiles along with the fix.
Let me know if you face any problem in adding tests cases.
Thanks!

contrib/babelfishpg_tds/src/backend/tds/tdstypeio.c
```diff
@@ -3851,12 +3851,12 @@ TdsSendTypeSqlvariant(FmgrInfo *finfo, Datum value, void *vMetaData)
 										 * given encoding. */
 		char	   *destBuf = NULL;
 
-		dataLen -= VARHDRSZ;
+		dataLen = VARSIZE_ANY_EXHDR(VARDATA_ANY(vlena) + variantHeaderLen);
 		if (variantBaseType == VARIANT_TYPE_NCHAR ||
```
berkayozkarhan, line 3855: Last case is we need to add varchar*char here.

pranavJ23: I don't completely follow up, why do we need to add varchar/ char in the if condition ?(if that's what you are suggesting).
We have else case to handle varchar, char cases.

pranavJ23: We just need to properly handle the actual data pointer here as well:
```
instead of 
destBuf = TdsEncodingConversion(buf + VARHDRSZ, dataLen, PG_UTF8, PG_WIN1252, &actualDataLen); 

change to 

destBuf = TdsEncodingConversion(VARDATA_ANY(VARDATA_ANY(vlena) + variantHeaderLen), dataLen, PG_UTF8, PG_WIN1252, &actualDataLen); 

```

berkayozkarhan: Sorry, I missed the part.

berkayozkarhan: Yes, you are right. The varchar evaluation is in different place, thanks :))

berkayozkarhan:
Which one is better:

`dataLen -= VARHDRSZ; ---> dataLen = VARSIZE_ANY_EXHDR(buf);`
`buf + VARHDRS --> VARDATA_ANY(buf)`

or

`dataLen -= VARHDRSZ; ---> dataLen = VARSIZE_ANY_EXHDR(VARDATA_ANY(vlena) + variantHeaderLen);`
`buf + VARHDRS --> VARDATA_ANY(VARDATA_ANY(vlena) + variantHeaderLen)`


I think 

`dataLen -= VARHDRSZ; ---> dataLen = VARSIZE_ANY_EXHDR(buf);`
`buf + VARHDRS --> VARDATA_ANY(buf)`

is better, because buf is already pointing **vlena + variantHeaderLen**. No need to create expression again.

tanscorpio7:
> Which one is better:
> 
> `dataLen -= VARHDRSZ; ---> dataLen = VARSIZE_ANY_EXHDR(buf);` `buf + VARHDRS --> VARDATA_ANY(buf)`
> 
> or
> 
> `dataLen -= VARHDRSZ; ---> dataLen = VARSIZE_ANY_EXHDR(VARDATA_ANY(vlena) + variantHeaderLen);` `buf + VARHDRS --> VARDATA_ANY(VARDATA_ANY(vlena) + variantHeaderLen)`
> 
> I think
> 
> `dataLen -= VARHDRSZ; ---> dataLen = VARSIZE_ANY_EXHDR(buf);` `buf + VARHDRS --> VARDATA_ANY(buf)`
> 
> is better, because buf is already pointing **vlena + variantHeaderLen**. No need to create expression again.

Both of them loog good to me.

berkayozkarhan:
Hello, @tanscorpio7 @pranavJ23 . Could you please check the tests in my last commit? I've added several tests, but they are failing. The tests are about binary string type. It outputs something different than sqlcmd output. I just tried to add them as they are in my local, but the pipeline generated different outputs. In sqlcmd, the commands are giving output as:

`
----------------------------------------------------
-- 37. Direct initialization with binary string from 'binary_string_test'
----------------------------------------------------
DECLARE @result_sql_variant sql_variant;
SET @result_sql_variant = (0x62696E6172795F737472696E675F74657374);
SELECT @result_sql_variant AS Result,
       SQL_VARIANT_PROPERTY(@result_sql_variant, 'BaseType') AS BaseType,
       SQL_VARIANT_PROPERTY(@result_sql_variant, 'MaxLength') AS MaxLength;
GO
`
In sqlcmd the output of above command is(on my machine): 62696E6172795F737472696E675F74657374

But when I run JDBC tests, it generates this output(on my machine)
~~START~~
sql_variant#!#sql_variant#!#sql_variant
[B@5b04476e#!#varbinary#!#65535`

But pipeline generates it as 
 ~~START~~
 sql_variant#!#sql_variant#!#sql_variant
+[B@3eee3e2b#!#varbinary#!#65535

Is it because of Java function outputs?

berkayozkarhan:
@tanscorpio7 @pranavJ23
One test if failing, is it OK? I could not understand why is it failing.

tanscorpio7:
LGTM! WIll wait for @pranavJ23 review. Thanks.

berkayozkarhan:
> LGTM! WIll wait for @pranavJ23 review. Thanks.

Could you also please review this PR: https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/4235
This is for another bug about conversion from string types to sqlvariant.

@tanscorpio7 @pranavJ23

pranavJ23 (approved):
Approved, 
@berkayozkarhan  please update the description with current approach.

berkayozkarhan:
> Approved, @berkayozkarhan please update the description with current approach.

Done. @pranavJ23

tanscorpio7 (approved):

PR #4219, opened 2025-10-24
Restrict non superuser to alter babelfish defined procedures/functions

sharathbp:
### Description
Currently non superuser and non owners can alter the procedures and function in sys schema. This will restrict non owners from altering sys schema procedures/function as they are created by babelfish.

### Issues Resolved

BABEL-6131

* **Use case based -**
Yes

* **Boundary conditions -**
Yes

* **Arbitrary inputs -**
Yes

* **Negative test cases -**
Yes

NA

NA

* **Performance tests -**
NA

* **Tooling impact -**
NA

* **Client tests -**
NA

### Check List


contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -5025,6 +5025,44 @@ bbf_ProcessUtility(PlannedStmt *pstmt,
...
+							{
+								Form_pg_proc procform = (Form_pg_proc) GETSTRUCT(tuple);
+
+								/* Check if the function is owned by bootstrap superuser */
+								if (procform->proowner ==  10)
+								{
+									ReleaseSysCache(tuple);
+									ereport(ERROR,
+											(errcode(ERRCODE_INSUFFICIENT_PRIVILEGE),
+											errmsg("cannot modify procedures owned by rdsadmin in sys schema")));
```
tanscorpio7: There is no `rdsadmin` in github.

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -5025,6 +5025,44 @@ bbf_ProcessUtility(PlannedStmt *pstmt,
...
+				CreateFunctionStmt 		*stmt = (CreateFunctionStmt *)parsetree;
+
+				/* Check if non-superuser is trying to create procedure in babelfish sys schemas */
+				if (!superuser())
+				{
+					char *schemaname = NULL;
+					char *objname = NULL;
+					
+					/* Extract schema name from function name */
+					DeconstructQualifiedName(stmt->funcname, &schemaname, &objname);
```
tanscorpio7: 1. Let's move this code block to a different function. Process utility hook is already too big to read.
2. `QualifiedNameGetCreationNamespace` instead of `DeconstructQualifiedName`
3. Lets keep the logic generic, any routine in sys namespace will be blocked irresepective of who owns it.

tanscorpio7 (changes requested):

contrib/babelfishpg_tsql/src/pl_handler.c
sharathbp: Its better to allow owner to alter/replace it right. In case we need to update it during upgrades

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -5412,6 +5419,27 @@ call_prev_ProcessUtility(PlannedStmt *pstmt,
...
 
+static void
+validate_sys_schema_permissions(List *funcname)
+{
+	char *objname = NULL;
+    Oid schemaOid;
+    Oid sysSchemaOid;
+
+    if (superuser())
+        return;
```
tanscorpio7: 1. Lets check the authenticated user or session user is superuser instead. There could be case where some other code path escaltes current user to superuser or descaltes superuser to lower.
2. Minor: Please check if we have sys schema cached somewhere and replace the `get_namespace_oid("sys", true);` call.
3. This will block CREATE, REPLACE of both procedure and functions, can we modify the error message to indicate that.

tanscorpio7 (changes requested):

tanscorpio7 (approved):

test/JDBC/expected/TEST-INTERNAL-PROC.out
rishabhtanwar29, line 1: What about ALTER PROCEDURE/FUNCTION? Is this already blocked or we think it's not harmful?

rishabhtanwar29 (approved):

PR #4246, opened 2025-11-05
Fix inconsistent index scans for binary/varbinary comparison operators and add cross-type support

RuchaSK1:
### Description

Currently, Babelfish has comparison operators for BINARY and VARBINARY types, but they are missing optimizer support clauses (RESTRICT, JOIN, NEGATOR).This causes the query planner to inconsistently choose between index scans and sequential scans even for highly selective queries.

Eg:
```
CREATE TABLE binary_test (
    binary_col BINARY(10)
);
GO
INSERT INTO binary_test (binary_col) 
    SELECT CAST(0x1 AS BINARY(10)) FROM generate_series(1, 3);
GO
~~ROW COUNT: 3~~
INSERT INTO binary_test (binary_col) 
    SELECT CAST(0x4 AS BINARY(10)) FROM generate_series(1, 3000000);
GO
~~ROW COUNT: 3000000~~
INSERT INTO binary_test (binary_col) 
    SELECT CAST(0x8 AS BINARY(10)) FROM generate_series(1, 3);
GO
~~ROW COUNT: 3~~
CREATE INDEX idx_binary_test ON binary_test(binary_col);
GO   SELECT COUNT(*) FROM binary_test WHERE binary_col < CAST(0x4 AS BINARY(10));
GO
~~START~~
int
3
~~END~~

-- Query with execution plan
SET BABELFISH_SHOWPLAN_ALL ON;
GO

SELECT binary_col FROM binary_test WHERE binary_col < CAST(0x4 AS BINARY(10));
GO
~~START~~
text
Query Text: SELECT binary_col FROM binary_test WHERE binary_col < CAST(0x4 AS [BINARY](10))
Seq Scan on binary_test (cost=0.00..53717.08 rows=1500003 width=32)
Filter: ((binary_col)::bbf_binary < '0x04000000000000000000'::bbf_binary)
~~END~~

~~START~~
text
Babelfish T-SQL Batch Parsing Time: 2.950 ms
~~END~~
```

Additionally, for cross-type comparisons between BINARY and VARBINARY types, Babelfish relies on implicit casts since no direct cross-type operators exist.While queries like `WHERE binary_col < varbinary_col` work, the implicit casting prevents index usage on the casted column, always resulting in sequential scans.

With this fix, the query planner will make consistent, cost-based decisions about index usage for both same-type and cross-type binary/varbinary comparisons.

### Issues Resolved

Task: BABEL-6098

Authored-by: Rucha Kulkarni [ruchask@amazon.com](mailto:ruchask@amazon.com)
Signed-off-by: Rucha Kulkarni [ruchask@amazon.com](mailto:ruchask@amazon.com)

* **Use case based -** Yes

* **Boundary conditions -**

* **Arbitrary inputs -** Yes

* **Performance tests -**

* **Tooling impact -**

* **Client tests -**

### Check List


contrib/babelfishpg_common/sql/binary.sql
```diff
@@ -375,9 +473,90 @@ CREATE OPERATOR sys.= (
...
+    NEGATOR = =,
+    JOIN = neqjoinsel
+);
+
+CREATE FUNCTION sys.varbinary_binary_gt(leftarg sys.bbf_varbinary, rightarg sys.bbf_binary)
+RETURNS boolean
+AS 'babelfishpg_common', 'varbinary_gt'
+LANGUAGE C IMMUTABLE STRICT PARALLEL SAFE;
+
+CREATE OPERATOR sys.> (
```
Yvinayak07, line 503: Why do we need add this operator explicitly for bbf_varbinary and bbf_binary the implicit cast should exist for this ,right?

contrib/babelfishpg_common/sql/binary.sql
```diff
@@ -375,9 +473,90 @@ CREATE OPERATOR sys.= (
...
+    LEFTARG = sys.bbf_varbinary,
+    RIGHTARG = sys.bbf_binary,
+    FUNCTION = sys.varbinary_binary_lt,
+    COMMUTATOR = >,
+    RESTRICT = scalarltsel,
+    NEGATOR = >=,
+    JOIN = scalarltjoinsel
+);
+
+CREATE FUNCTION sys.varbinary_binary_leq(leftarg sys.bbf_varbinary, rightarg sys.bbf_binary)
```
Yvinayak07, line 543: can't we use already existing function that using same underlying c function .
```
CREATE FUNCTION sys.binary_leq(leftarg sys.bbf_binary, rightarg sys.bbf_binary)
RETURNS boolean
AS 'babelfishpg_common', 'varbinary_leq'
```

contrib/babelfishpg_common/sql/binary.sql
```diff
@@ -353,11 +370,92 @@ CREATE OPERATOR sys.= (
...
+    LEFTARG = sys.bbf_binary,
+    RIGHTARG = sys.bbf_varbinary,
+    FUNCTION = sys.binary_varbinary_leq,
+    COMMUTATOR = >=,
+    RESTRICT = scalarlesel,
+    NEGATOR = >,
+    JOIN = scalarlejoinsel
 );
 
 alter OPERATOR family bbf_binary_ops USING btree add
```
Yvinayak07, line 457: Let's add a new code block for this that say:
```
ALTER OPERATOR FAMILY bbf_binary_ops USING btree ADD
            OPERATOR 1 sys.< (sys.bbf_binary, sys.bbf_varbinary),
            OPERATOR 2 sys.<= (sys.bbf_binary, sys.bbf_varbinary),
            OPERATOR 4 sys.>= (sys.bbf_binary, sys.bbf_varbinary),
            OPERATOR 5 sys.> (sys.bbf_binary, sys.bbf_varbinary);
```

contrib/babelfishpg_common/sql/upgrades/babelfish_common_helper--5.4.0--5.5.0.sql
```diff
@@ -93,3 +93,496 @@ CREATE OR REPLACE FUNCTION sys.binarysysnchar(sys.BBF_BINARY, integer, boolean)
...
+        RETURNS boolean AS 'babelfishpg_common', 'varbinary_gt'
+        LANGUAGE C IMMUTABLE STRICT PARALLEL SAFE;
+
+        CREATE OR REPLACE FUNCTION sys.binary_varbinary_geq(leftarg sys.bbf_binary, rightarg sys.bbf_varbinary)
+        RETURNS boolean AS 'babelfishpg_common', 'varbinary_geq'
+        LANGUAGE C IMMUTABLE STRICT PARALLEL SAFE;
+
+        CREATE OR REPLACE FUNCTION sys.bbf_binary_varbinary_cmp(sys.bbf_binary, sys.bbf_varbinary)
+        RETURNS int AS 'babelfishpg_common', 'varbinary_cmp'
+        LANGUAGE C IMMUTABLE STRICT PARALLEL SAFE;
```
Yvinayak07: We can create these functions outside of these conditional handlings? First of all let's check if we can use the already existing functions that have same underlying c functions

contrib/babelfishpg_common/sql/binary.sql
```diff
@@ -260,7 +260,9 @@ CREATE OPERATOR sys.= (
     RIGHTARG = sys.bbf_binary,
     FUNCTION = sys.binary_eq,
     COMMUTATOR = =,
-    RESTRICT = eqsel
+    RESTRICT = eqsel,
+    NEGATOR = <>,
+    JOIN = eqjoinsel
```
Yvinayak07, line 265: Is adding NEGATOR and JOIN field helping us in picking index? did we test it without this? (Isn't just adding to the operator family help us in picking index scan)

contrib/babelfishpg_common/sql/binary.sql
RuchaSK1: Implicit cast exists, but without explicit cross-type operators, PostgreSQL casts the column instead of the literal. Since the index is built on binary_col, not on (binary_col)::bbf_varbinary, applying a cast to the column prevents index usage. Explicit cross-type operators allow direct comparison without transforming the column.

Eg:
Without explicit > operator for bbf_varbinary and bbf_binary:

```
SELECT binary_col FROM BABEL_6098_binary_test 
WHERE binary_col > CAST(0x08 AS VARBINARY(10));
GO
~~START~~
text
Query Text: SELECT binary_col FROM BABEL_6098_binary_test 
WHERE binary_col > CAST(0x08 AS VARBINARY(10))
Seq Scan on babel_6098_binary_test
  Filter: ((binary_col)::bbf_varbinary > '0x08'::bbf_varbinary)
~~END~~
```

With explicit cross operator:

```
SELECT binary_col FROM BABEL_6098_binary_test 
WHERE binary_col > CAST(0x08 AS VARBINARY(10));
GO
~~START~~
text
Query Text: SELECT binary_col FROM BABEL_6098_binary_test 
WHERE binary_col > CAST(0x08 AS VARBINARY(10))
Index Only Scan using babel_6098_idx_binarybabel_60982612b34d3ea017566be3aba8d0cd159a on babel_6098_binary_test
  Index Cond: (binary_col > '0x08'::bbf_varbinary)
~~END~~
```

contrib/babelfishpg_common/sql/binary.sql
RuchaSK1: No, we can't reuse the existing function. PostgreSQL requires the function signature to exactly match the operator's argument types. If we use 

```
CREATE OPERATOR sys.<= (
    LEFTARG = sys.bbf_varbinary,
    RIGHTARG = sys.bbf_binary,
    FUNCTION = sys.binary_leq,
    COMMUTATOR = >=,
    RESTRICT = scalarlesel,
    NEGATOR = >,
    JOIN = scalarlejoinsel
);
```

We get error:

`ERROR:  function sys.binary_leq(bbf_varbinary, bbf_binary) does not exist`

So we need separate SQL wrapper functions for each type combination, even though they call the same underlying C function.

contrib/babelfishpg_common/sql/upgrades/babelfish_common_helper--5.4.0--5.5.0.sql
RuchaSK1: Yes we can move all function definitions outside IF block. 
But we cannot reuse existing functions since Postgres requires the function signature to exactly match the operator's argument types.

contrib/babelfishpg_common/sql/binary.sql
RuchaSK1: Adding to operator family is what primarily enables index scans for direct operators (>,< etc).
NEGATOR enables the planner to transform negated queries (e.g., NOT (col <= x) to col > x), which allows index scans to be used where applicable.
Eg:

Without Negator:

```
SELECT binary_col FROM BABEL_6098_binary_test 
WHERE NOT (binary_col <= CAST(0x08 AS VARBINARY(10)));
GO
~~START~~
text
Query Text: SELECT binary_col FROM BABEL_6098_binary_test 
WHERE NOT (binary_col <= CAST(0x08 AS VARBINARY(10)))
Seq Scan on babel_6098_binary_test
  Filter: (NOT ((binary_col)::bbf_binary <= '0x08'::bbf_varbinary))
~~END~~
```


With Negator:

```
SELECT binary_col FROM BABEL_6098_binary_test 
WHERE NOT (binary_col <= CAST(0x08 AS VARBINARY(10)));
GO
~~START~~
text
Query Text: SELECT binary_col FROM BABEL_6098_binary_test 
WHERE NOT (binary_col <= CAST(0x08 AS VARBINARY(10)))
Index Only Scan using babel_6098_idx_binarybabel_60982612b34d3ea017566be3aba8d0cd159a on babel_6098_binary_test
  Index Cond: (binary_col > '0x08'::bbf_varbinary)
~~END~~
```

JOIN clause provides selectivity estimates for join operations, helping the planner make better cost-based decisions. It doesn't enable index usage directly. If we want to keep changes minimal, we can remove JOIN since it's optional.

contrib/babelfishpg_common/sql/binary.sql
RuchaSK1: Added

contrib/babelfishpg_common/sql/upgrades/babelfish_common_helper--5.4.0--5.5.0.sql
```diff
@@ -93,3 +93,496 @@ CREATE OR REPLACE FUNCTION sys.binarysysnchar(sys.BBF_BINARY, integer, boolean)
 RETURNS sys.NCHAR
 AS 'babelfishpg_common', 'varbinarynchar'
 LANGUAGE C IMMUTABLE STRICT PARALLEL SAFE;
+
+-- complain if script is sourced in psql, rather than via ALTER EXTENSION
+\echo Use "ALTER EXTENSION ""babelfishpg_common"" UPDATE TO "5.5.0"" to load this file. \quit
+
+SELECT set_config('search_path', 'sys, '||current_setting('search_path'), false);
```
Deepesh125: plz run this script multiple times to verify correctness

test/JDBC/expected/babel_varbinary-vu-verify.out
```diff
@@ -737,8 +737,8 @@ GO
 ~~START~~
 text
 Query Text: Select varbinary_col from babel_varbinary_test_table2 where varbinary_col > cast(0x123 as varbinary(max)) + cast(0x123 as varbinary(max))
-Index Only Scan using babel_varbinary_test_indbabel_v93c04b51675320c0200287b60808b914 on babel_varbinary_test_table2
-  Index Cond: (varbinary_col > '0x01230123'::bbf_varbinary)
+Seq Scan on babel_varbinary_test_table2
+  Filter: ((varbinary_col)::bbf_varbinary > '0x01230123'::bbf_varbinary)
```
Deepesh125, line 741: plz explain this diff?

test/JDBC/expected/binary-index-vu-verify.out
```diff
@@ -72,24 +80,24 @@ Query Text: EXEC babel_3939_vu_prepare_p1
   Query Text: select count(*) from tab_binary where a > 0xBAADF00D
   ->  Aggregate
         ->  Seq Scan on tab_binary
-              Filter: ((a)::bbf_varbinary > '0xbaadf00d'::bbf_varbinary)
+              Filter: ((a)::bbf_binary > '0xbaadf00d'::bbf_varbinary)
```
Deepesh125, line 83: tab_binary(a) is already bbf_binary then why are we seeing explicit cast?

contrib/babelfishpg_common/sql/upgrades/babelfish_common_helper--5.4.0--5.5.0.sql
```diff
@@ -93,3 +93,496 @@ CREATE OR REPLACE FUNCTION sys.binarysysnchar(sys.BBF_BINARY, integer, boolean)
...
+    END IF;
+END;
+$$ LANGUAGE plpgsql;
+
+-- Add cross-type operators for bbf_varbinary op bbf_binary
+DO $$
+DECLARE 
+    bbf_varbinary_ops_c INT := (SELECT * FROM get_bbf_binary_ops_count('bbf_varbinary_ops'));
+BEGIN
+    RAISE NOTICE 'bbf_varbinary_ops operator count: %', bbf_varbinary_ops_c;
```
Yvinayak07: Do we really need this Raise notice condition? Let's remove this and from other places as well.

contrib/babelfishpg_common/sql/upgrades/babelfish_common_helper--5.4.0--5.5.0.sql
```diff
@@ -93,3 +93,496 @@ CREATE OR REPLACE FUNCTION sys.binarysysnchar(sys.BBF_BINARY, integer, boolean)
...
+$$ LANGUAGE plpgsql;
+
+-- Add cross-type operators for bbf_binary op bbf_varbinary
+DO $$
+DECLARE 
+    bbf_binary_ops_c INT := (SELECT * FROM get_bbf_binary_ops_count('bbf_binary_ops'));
+BEGIN
+    RAISE NOTICE 'bbf_binary_ops operator count: %', bbf_binary_ops_c;
+
+    IF bbf_binary_ops_c = 5 THEN
```
Yvinayak07: Do we really need this condition of `bbf_binary_ops_c = 5`? from what i could understand it is if we don't have `CREATE OPERATOR sys.=`.  as the below operator already exsists.
```
CREATE OPERATOR sys.= (
            LEFTARG = sys.bbf_binary,
            RIGHTARG = sys.bbf_varbinary,
            FUNCTION = sys.binary_varbinary_eq,
            COMMUTATOR = =,
            NEGATOR = <>,
            RESTRICT = eqsel,
            JOIN = eqjoinsel
        );
```

test/JDBC/expected/babel_varbinary-vu-verify.out
RuchaSK1: Table has 3,000,006 rows and the query selects 3,000,003 rows. For such low selectivity where almost all rows match, Seq Scan is more efficient than Index Scan. 

```
select count(*) from babel_varbinary_test_table2 where varbinary_col > cast(0x123 as varbinary(max)) + cast(0x123 as varbinary(max))
GO
~~START~~
int
3000003
~~END~~
```

Before our changes, the > operator for varbinary didn't have proper RESTRICT clause. Without it, PostgreSQL uses default selectivity estimate, which underestimated the actual selectivity, leading to incorrect Index Scan choice. After our changes, the planner gets accurate estimates and correctly chooses Seq Scan.

contrib/babelfishpg_common/sql/upgrades/babelfish_common_helper--5.4.0--5.5.0.sql
RuchaSK1: Agreed. Will remove RAISE NOTICE statements since they only help for debugging

test/JDBC/expected/binary-index-vu-verify.out
RuchaSK1: I'm not entirely sure why we see the explicit cast. The column a is stored as BINARY which is a domain over bbf_binary, so PostgreSQL is probably showing the conversion from domain to base type in the plan output. We can also see this pattern in the earlier output where it showed (a)::bbf_varbinary.

contrib/babelfishpg_common/sql/upgrades/babelfish_common_helper--5.4.0--5.5.0.sql
RuchaSK1: Yes, `bbf_binary_ops_c = 5` is for when the cross-type = operator doesn't exist. But since fresh installations go through intermediate upgrade scripts that add it, this condition won't occur. Will remove it.

contrib/babelfishpg_common/sql/upgrades/babelfish_common_helper--5.4.0--5.5.0.sql
```diff
@@ -2,6 +2,9 @@
 ---- Include changes related to other datatypes except spatial types here ----
 ------------------------------------------------------------------------------
 
+-- complain if script is sourced in psql, rather than via ALTER EXTENSION
+\echo Use "ALTER EXTENSION ""babelfishpg_common"" UPDATE TO "5.5.0"" to load this file. \quit
+
```
Yvinayak07: Why are we adding this?

contrib/babelfishpg_common/sql/upgrades/babelfish_common_helper--5.4.0--5.5.0.sql
```diff
@@ -94,6 +97,296 @@ RETURNS sys.NCHAR
 AS 'babelfishpg_common', 'varbinarynchar'
 LANGUAGE C IMMUTABLE STRICT PARALLEL SAFE;
 
+SELECT set_config('search_path', 'sys, '||current_setting('search_path'), false);
```
Yvinayak07: Why this line is explicilty added i don't think we need this in helper file and also remove trim line from below.

contrib/babelfishpg_common/sql/upgrades/babelfish_common_helper--5.4.0--5.5.0.sql
```diff
@@ -94,6 +97,296 @@ RETURNS sys.NCHAR
...
+alter OPERATOR sys.<= (sys.bbf_varbinary, sys.bbf_varbinary) 
+set (
+    RESTRICT = scalarlesel,
+    NEGATOR = >,
+    JOIN = scalarlejoinsel
+);
+
+-- Adding cross operators support for comparison between binary and varbinary
+-- Helper function
+CREATE OR REPLACE FUNCTION get_bbf_binary_ops_count(opffamily varchar)
```
Yvinayak07: We should append sys in front of function name here and whereever it is being used.

contrib/babelfishpg_common/sql/upgrades/babelfish_common_helper--5.4.0--5.5.0.sql
RuchaSK1: The set_config is required because NEGATOR/COMMUTATOR clauses in ALTER OPERATOR cannot be schema-qualified. PostgreSQL uses search_path to resolve them, so sys must be in the path.

```
dev-dsk-ruchask-1c-a99d43f3 % psql -h localhost -d babelfish_db
psql (17.6)
Type "help" for help.

babelfish_db=# alter OPERATOR sys.> (sys.bbf_binary, sys.bbf_binary) 
babelfish_db-# set (
babelfish_db(#     RESTRICT = scalargtsel,
babelfish_db(#     NEGATOR = <=,
babelfish_db(#     JOIN = scalargtjoinsel
babelfish_db(# );
ERROR:  operator does not exist: sys.bbf_binary <= sys.bbf_binary
babelfish_db=# SELECT set_config('search_path', 'sys, '||current_setting('search_path'), false);
      set_config      
----------------------
 sys, "$user", public
(1 row)

babelfish_db=# alter OPERATOR sys.> (sys.bbf_binary, sys.bbf_binary) 
babelfish_db-# set (
babelfish_db(#     RESTRICT = scalargtsel,
babelfish_db(#     NEGATOR = <=,
babelfish_db(#     JOIN = scalargtjoinsel
babelfish_db(# );
ALTER OPERATOR
babelfish_db=#

```

contrib/babelfishpg_common/sql/upgrades/babelfish_common_helper--5.4.0--5.5.0.sql
RuchaSK1: This is a standard safety header present in all upgrade scripts to prevent accidental direct execution in psql.

contrib/babelfishpg_common/sql/upgrades/babelfish_common_helper--5.4.0--5.5.0.sql
Yvinayak07: We call `babelfish_common_helper` and `spatial_types` from `babelfishpg_upgrades.in` file that already have this appended.

contrib/babelfishpg_common/sql/upgrades/babelfish_common_helper--5.4.0--5.5.0.sql
Yvinayak07: We call `babelfish_common_helper` and `spatial_types` from `babelfishpg_upgrades.in` file that already have this appended. We could check in the `babelfish_common_helper--5.4.0-5.5.0` where we removed this.

contrib/babelfishpg_common/sql/upgrades/babelfish_common_helper--5.4.0--5.5.0.sql
```diff
@@ -94,6 +97,296 @@ RETURNS sys.NCHAR
...
+        -- All cross-type operators already installed
+        NULL;
+
+    ELSE
+        RAISE EXCEPTION 'Unexpected operator count in bbf_varbinary_ops: % (expected 5, 6, or 10)', bbf_varbinary_ops_c;
+    END IF;
+END;
+$$ LANGUAGE plpgsql;
+
+DROP FUNCTION get_bbf_binary_ops_count(varchar);
```
Yvinayak07: Let's append `sys` here as well.

contrib/babelfishpg_common/sql/upgrades/babelfish_common_helper--5.4.0--5.5.0.sql
```diff
@@ -94,6 +97,296 @@ RETURNS sys.NCHAR
...
+        );
+
+        ALTER OPERATOR FAMILY bbf_binary_ops USING btree ADD
+            OPERATOR 1 sys.< (sys.bbf_binary, sys.bbf_varbinary),
+            OPERATOR 2 sys.<= (sys.bbf_binary, sys.bbf_varbinary),
+            OPERATOR 4 sys.>= (sys.bbf_binary, sys.bbf_varbinary),
+            OPERATOR 5 sys.> (sys.bbf_binary, sys.bbf_varbinary);
+
+    ELSIF bbf_binary_ops_c = 10 THEN
+        -- All cross-type operators already installed
```
Yvinayak07, line 276: Instead of NULL here we can have `Raise notice` saying that `operator of bbf_binary_ops is installed`. similarly for 
`bbf_varbinary_ops`.

Yvinayak07 (approved):

contrib/babelfishpg_common/sql/upgrades/babelfish_common_helper--5.4.0--5.5.0.sql
```diff
@@ -574,15 +367,29 @@ BEGIN
             OPERATOR 5 sys.> (sys.bbf_varbinary, sys.bbf_binary);
 
     ELSIF bbf_varbinary_ops_c = 10 THEN
-        RAISE NOTICE 'All cross-type operators already installed in bbf_varbinary_ops';
+        -- All cross-type operators already installed
+        RAISE NOTICE 'All operators of bbf_varbinary_ops are installed';
```
Deepesh125, line 371: do we really need this?

Deepesh125 (dismissed):

contrib/babelfishpg_common/sql/upgrades/babelfish_common_helper--5.4.0--5.5.0.sql
RuchaSK1: Not needed for functionality, but added for verification purposes. Can remove it if feels unnecessary.

Deepesh125 (approved):

PR #4275, opened 2025-11-16
Handling various whitespaces in isnumeric

tanyagupta17:
### Description

This commit handles various types of whitespaces in Babelfish for isnumeric(), ensuring compatibility with T-SQL behavior. T-SQL allows leading special whitespace to be trimmed but treats trailing special whitespace after numeric content as non-numeric. We have also handled arithmetic operations between these whitespaces and other characters. Following are the result:
*  1  - Contains only special whitespace (\t, \n, \v, \f, \r)
*  0  - Empty string or contains only spaces
* -1  - Contains other characters (needs numeric conversion)
*  0  - Contains other characters and ends with special whitespace

### Issues Resolved

BABEL-6174

* **Use case based -** Yes
 

* **Arbitrary inputs -** Yes

* **Performance tests -** NA

* **Tooling impact -** NA

* **Client tests -**NA

### Check List


contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -2294,6 +2295,64 @@ search_partition(PG_FUNCTION_ARGS)
...
+ */
+static int
+checkWhitespaceType(const char *str)
+{
+	size_t i;
+	size_t len;
+	bool has_special_ws = false;
+	bool has_other_char = false;
+
+	if (!str)
```
Yvinayak07: Do we have test case for this condition or is it a sanity check?

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -2369,9 +2429,18 @@ isnumeric(PG_FUNCTION_ARGS)
...
+	if (ws_check == 1)
+	{
+		/* Contains only special whitespace - return 1 */
+		pfree(value_str);
+		PG_RETURN_INT32(1);
+	}
+	else if (ws_check == 0)
 	{
+		/* Empty or only spaces - return 0 */
 		pfree(value_str);
```
Yvinayak07, line 2422: We should have an check that `value_str ` should not be NULL.

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -2294,6 +2295,64 @@ search_partition(PG_FUNCTION_ARGS)
...
+	if (!str)
+		return 0;
+
+	len = strlen(str);
+	if (len == 0)
+		return 0;
+
+	for (i = 0; i < len; i++)
+	{
+		if (str[i] == '\b')
```
Yvinayak07: What does '\b' stands for?

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -2294,6 +2295,64 @@ search_partition(PG_FUNCTION_ARGS)
...
+		}
+		else if (str[i] == ' ')
+		{
+			/* Space character - continue checking */
+			continue;
+		}
+		else
+		{
+			/* Found non-whitespace or non-special character */
+			has_other_char = true;
```
Yvinayak07, line 2328: We can remove `has_other_char` and directly return -1 from here, right?

test/JDBC/expected/babel_isnumeric-vu-verify.out
```diff
@@ -1092,3 +1092,1519 @@ int#!#int
 0#!#0
 ~~END~~
 
+
```
Yvinayak07: Could we test:
- select isnumeric($)
- select isnumeric('$$')
- select isnumeric(char(1) + char(1)) (and also adding  and subtracting other char in the same way)

Yvinayak07 (commented):
Could we check if we have marked isnumeric() strict? if not, then we should mark it strict.

contrib/babelfishpg_tsql/runtime/functions.c
tanyagupta17: Its a sanity check

contrib/babelfishpg_tsql/runtime/functions.c
tanyagupta17: '\b' represents Backspace, i.e. char(8) it has a different handling in sqlserver for this than other whitespaces

contrib/babelfishpg_tsql/runtime/functions.c
tanyagupta17: No, we need it to distinguish between leading and trailing special whitespaces.
TSQL's ISNUMERIC() only allows leading whitespace (tabs, newlines, etc.) to be trimmed, but trailing tabs/special whitespace characters make the string non-numeric
```
CHAR(9) + '123' returns 1 (leading tab is OK)
'123' + CHAR(9) returns 0 (trailing tab is NOT OK) 
```

contrib/babelfishpg_tsql/runtime/functions.c
tanyagupta17: We already have the condition above, so value_str can't be null right? I don't think we need the check again
```
	if (PG_ARGISNULL(0))
		PG_RETURN_INT32(0);
```

tanyagupta17 (commented):
> Could we check if we have marked isnumeric() strict? if not, then we should mark it strict.

We should not mark it as strict because Tsql's ISNUMERIC() returns 0 for NULL input, not NULL. And our C code implementation already handles it. Making it STRICT would change the behavior to return NULL instead of 0

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -215,6 +215,7 @@ void	   *get_language(void);
 void	   *get_host_id(void);
 
 Datum 		datepart_internal(char *field , Timestamp timestamp , float8 df_tz, bool general_integer_datatype);
+static int 	checkWhitespaceType(const char *str);
```
tanscorpio7: We do not need to declare static functions if they are only used after their definiton.

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -2294,6 +2295,43 @@ search_partition(PG_FUNCTION_ARGS)
 	PG_RETURN_INT32(result);
 }
 
+/* 
+ * Check whitespace type in string
+ * Returns:
+ *   1  - Contains only special whitespace (\t, \n, \v, \f, \r)
+ *   0  - Empty string or contains only spaces
+ *   -1 - Contains other characters (needs numeric conversion)
+ */
```
tanscorpio7, line 2303: How did we generate this list ? SQL Server opensource documentation or testing ?

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -2369,9 +2408,18 @@ isnumeric(PG_FUNCTION_ARGS)
 		value_str = OidOutputFunctionCall(typoutput, PG_GETARG_DATUM(0));
 	}
 
-	/* Handling empty string. */
-	if ((*common_utility_plugin_ptr->isEmptyOrWhitespace)(value_str))
+	/* Check whitespace type */
```
tanscorpio7, line 2410: IS_NUMERIC is defined as `ISNUMERIC returns 1 when the input expression evaluates to a valid numeric data type;` so we should also test casting numeric string with these special chars to numeric.

tanscorpio7 (changes requested):

contrib/babelfishpg_tsql/runtime/functions.c
tanyagupta17: Through testing, we found a divergence between T-SQL documentation and actual behavior. As per the TSQL doc, ISNUMERIC() should return 1 for any input that can be converted to a valid numeric data type. However, while empty strings can be converted to money/smallmoney (returning 0) in both Babelfish and SQL Server, ISNUMERIC() returns 0 for empty strings but returns 1 for special whitespace characters.

During our previous implementation in BABEL-6082, we had only considered empty strings and returned 0 for all cases. With this commit, we are handling all special whitespaces as well to match TSQL.

contrib/babelfishpg_tsql/runtime/functions.c
tanyagupta17: Added

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -2369,9 +2407,18 @@ isnumeric(PG_FUNCTION_ARGS)
...
+	ws_check = checkWhitespaceType(value_str);
+	
+	if (ws_check == 1)
+	{
+		/* Contains only special whitespace - return 1 */
+		pfree(value_str);
+		PG_RETURN_INT32(1);
+	}
+	else if (ws_check == 0)
```
tanscorpio7, line 2419: Wouldn't this lead to regression for example
```
SELECT ISNUMERIC(CHAR(92))
before ==> 1
after ==> 0
```
and this is a numeric type
```
1> SELECT cast('\' as money)
2> GO
                     
---------------------
                .0000

(1 rows affected)
```

Similarly `SELECT ISNUMERIC(',')`

tanscorpio7 (changes requested):

contrib/babelfishpg_tsql/runtime/functions.c
tanyagupta17: This regression was caused by BABEL-2187, where we blocked cases like '\0', nchar(160) [Non-breaking space] , CHAR(44) [Comma], CHAR(92) [Backslash] casting to money. 
This current PR has no role to play in this as we will be getting ws_check = -1 here and just continue with the rest of the code to check casting to numeric/money. So once we fix the casting of these values to money again, we will get the correct result. 

It is being tracked here BABEL-6270: #4342

tanscorpio7 (approved):

Yvinayak07 (approved):

PR #4292, opened 2025-11-25
Add remote procedure execution via TDS protocol with RPC out validation for linked servers

fizzlr:
Implement remote stored procedure execution for Babelfish linked servers through TDS RPC calls with security validations. Key changes include:
- Add RPC call handling and parameter marshaling in linked servers related files
- Extend catalog functions for remote procedure metadata lookup
- Implement result processing and error handling for remote calls
- Add RPC out server option validation for security
- Add SELECT-only statement validation for remote procedures
- Update parser to detect and route remote procedure calls

### Description

_What is the change?_
Currently, Babelfish does not support executing stored procedures on remote linked servers. With this change, users can now execute stored procedures on Babelfish linked servers using four-part naming syntax (e.g., EXEC [linked_server].[database].[schema].[procedure_name]). The execution is performed via TDS RPC calls to the remote server.

_Why was the change made?_
Linked server support is a critical feature for SQL Server compatibility, allowing users to execute queries and procedures across multiple database instances. Remote procedure execution enables distributed database applications and cross-server workflows. Without this capability, users migrating to Babelfish would need to refactor applications that rely on remote procedure calls.

_How was the code changed?_
- __Parser and executor enhancements__: Added detection logic to identify remote procedure calls from four-part names and route them to the appropriate handler

- __Linked server infrastructure__: Extended linked servers related files with RPC call handling, parameter marshaling, and TDS protocol communication

- __Catalog integration__: Enhanced catalog functions to look up remote procedure metadata

- __Security validations__: Implemented RPC out server option checking and SELECT-only statement validation to ensure secure remote execution

- __Result processing__: Added handling for remote procedure results, error propagation, and transaction coordination

### Issues Resolved
Jira: 4178

## Tests ##

RUNNING remote-proc-exec-vu-prepare
RUNNING remote-proc-exec-vu-verify
RUNNING remote-proc-exec-vu-cleanup
All tests executed successfully!
###########################################################################
###########################################################################
remote-proc-exec-vu-prepare: Passed! (869/120000ms OK)
remote-proc-exec-vu-verify:  Passed! (71685/120000ms OK)
remote-proc-exec-vu-cleanup: Passed! (134/120000ms OK)
###########################################################################
TOTAL TESTS:	3
TESTS PASSED:	3
TESTS FAILED:	0
###########################################################################
[INFO] Tests run: 4, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 73.386 s - in com.sqlsamples.TestQueryFile
[INFO]
[INFO] Results:
[INFO]
[INFO] Tests run: 4, Failures: 0, Errors: 0, Skipped: 0
[INFO]
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  01:14 min
[INFO] Finished at: 2025-12-15T19:14:55Z
[INFO] ------------------------------------------------------------------------

### Check List


contrib/babelfishpg_tsql/Makefile
```diff
@@ -95,9 +95,9 @@ PG_CXXFLAGS += -I$(ANTLR4_RUNTIME_INCLUDE_DIR)
 PG_CXXFLAGS += -std=c++17
 PG_CFLAGS += -g -Werror -Wfloat-conversion
 PG_CFLAGS += -fstack-protector-strong
-PG_CPPFLAGS += -I$(TSQLSRC) -I$(PG_SRC) -DFAULT_INJECTOR -Wfloat-conversion
+PG_CPPFLAGS += -I$(TSQLSRC) -I$(PG_SRC) -DFAULT_INJECTOR -DENABLE_TDS_LIB -Wfloat-conversion
```
forestkeeper: Hi, do we need to put this into here ? 
I think we have mentioned the operations in this doc : https://github.com/babelfish-for-postgresql/babelfish_extensions/blob/31b3d62300226de224581ff90cb473757811a4ad/INSTALLING.md.tmpl#L316-L320

contrib/babelfishpg_tsql/src/linked_servers.c
```diff
@@ -746,6 +748,11 @@ linked_server_establish_connection(char *servername, LinkedServerProcess * lspro
 	int	query_timeout = 0;
 	int	connect_timeout = 0;
 
+	/* LOG: Function entry with context - proves this function was called */
+	elog(LOG, "DEBUG_LINKED_SERVER: linked_server_establish_connection() called for server: %s (context: %s)",
```
forestkeeper: What's this log for ?

forestkeeper: It's exposing the server url in LOG info ?

contrib/babelfishpg_tsql/src/linked_servers.c
```diff
@@ -1126,8 +1172,9 @@ getOpenqueryTupdescFromMetadata(char *linked_server, char *query, TupleDesc *tup
 	{
 		if (lsproc)
 		{
-			LINKED_SERVER_DEBUG("LINKED SERVER: (Metadata) - Closing connections to remote server");
-			LINKED_SERVER_EXIT();
+			LINKED_SERVER_DEBUG("LINKED SERVER: (Metadata) - Closing connection to remote server");
+			LINKED_SERVER_CANCEL(lsproc);
```
forestkeeper: Why add a LINKED_SERVER_CANCEL in here ?

contrib/babelfishpg_tsql/src/linked_servers.c
```diff
@@ -1317,6 +1377,420 @@ openquery_imp(PG_FUNCTION_ARGS)
 	return (Datum) 0;
 }
 
+/*
+ * Map PostgreSQL/Babelfish type OID to TDS type for RPC parameter binding
+ */
+int
+get_tds_type_from_pg_oid(Oid pgtype)
```
forestkeeper, line 1461: We don't have similar function already in the code ?

contrib/babelfishpg_tsql/src/linked_servers.c
```diff
@@ -1317,6 +1377,420 @@ openquery_imp(PG_FUNCTION_ARGS)
...
+	
+	return 0;  /* Should not reach here */
+}
+
+/*
+ * Convert a Datum value to raw bytes suitable for TDS RPC parameter binding
+ * Returns palloc'd buffer containing the data
+ */
+void
+convert_datum_to_tds_bytes(Datum value, Oid valtype, int32 valtypmod, bool isnull,
```
forestkeeper, line 1654: We don't have existed function in the code ?

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -826,6 +830,408 @@ exec_run_dml_with_output(PLtsql_execstate *estate, PLtsql_stmt_push_result *stmt
...
+	TupleDesc tupdesc = NULL;
+	Tuplestorestate *tupstore = NULL;
+	Portal portal = NULL;
+	DestReceiver *receiver = NULL;
+	QueryCompletion qc;
+	uint64 processed = 0;
+	ListCell *lc;
+	
+	/* Build the full procedure name: database.schema.procedure */
+	full_proc_name = psprintf("%s.%s.%s",
```
forestkeeper, line 909: Is that a proper way to do a format like this ?

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -826,6 +830,408 @@ exec_run_dml_with_output(PLtsql_execstate *estate, PLtsql_stmt_push_result *stmt
...
+			int colcount_val = 0;
+			
+			elog(LOG, "SELECT-only validation: Opening connection for validation query");
+			
+			/* Open connection for validation - use dedicated variable */
+			linked_server_establish_connection(stmt->server_name, &validation_lsproc, false);
+			
+			/* Build query to fetch procedure definition */
+			initStringInfo(&validation_query);
+			appendStringInfo(&validation_query,
```
forestkeeper: SQL concation in here will leave a SQL inject leaking, pls try to use prepared stmt to do so.

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -826,6 +830,408 @@ exec_run_dml_with_output(PLtsql_execstate *estate, PLtsql_stmt_push_result *stmt
...
+			
+			/* Evaluate parameter expression */
+			val = exec_eval_expr(estate, p->expr, &isnull, &valtype, &valtypmod);
+			
+			/* Get TDS type for this parameter */
+			tds_type = get_tds_type_from_pg_oid(valtype);
+			
+			/* Convert Datum to raw bytes */
+			convert_datum_to_tds_bytes(val, valtype, valtypmod, isnull, 
+									   &param_data, &param_len);
```
forestkeeper, line 982: why not consider pltsql_exec_tsql_cast_value ?

forestkeeper: Seems doing the same thing.

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -826,6 +830,408 @@ exec_run_dml_with_output(PLtsql_execstate *estate, PLtsql_stmt_push_result *stmt
...
+				elog(LOG, "DEBUG_LINKED_SERVER: (RPC) - Fetched %d rows", rowcount);
+				
+				/* Finalize tuplestore */
+				tuplestore_donestoring(tupstore);
+				
+				/* Push active snapshot for SPI operations */
+				PushActiveSnapshot(GetTransactionSnapshot());
+				
+				/* Create a Portal to wrap the tuplestore for sending to client */
+				portal = SPI_cursor_open_with_args(NULL, "SELECT 1", 0, NULL, NULL, NULL, true, 0);
```
forestkeeper, line 1161: Why "select 1" , where's this from ?

contrib/babelfishpg_tsql/src/tsqlUnsupportedFeatureHandler.cpp
```diff
@@ -1497,8 +1497,8 @@ antlrcpp::Any TsqlUnsupportedFeatureHandlerImpl::visitFunc_proc_name_database_sc
 
 antlrcpp::Any TsqlUnsupportedFeatureHandlerImpl::visitFunc_proc_name_server_database_schema(TSqlParser::Func_proc_name_server_database_schemaContext *ctx)
 {
-	if (ctx->DOT().size() >= 3 && ctx->server) /* server.db.schema.funcname */
-		throw PGErrorWrapperException(ERROR, ERRCODE_FEATURE_NOT_SUPPORTED, "Remote procedure/function reference with 4-part object name is not currently supported in Babelfish", getLineAndPos(ctx));
+	// if (ctx->DOT().size() >= 3 && ctx->server) /* server.db.schema.funcname */
```
forestkeeper: Should just delete those code instead of commenting out.

contrib/babelfishpg_tsql/Makefile
thephantomthief: Yes, we do not build with linked server dependencies by default so we should not add this.

contrib/babelfishpg_tsql/Makefile
```diff
@@ -95,9 +95,9 @@ PG_CXXFLAGS += -I$(ANTLR4_RUNTIME_INCLUDE_DIR)
...
 
-SHLIB_LINK += -L$(ANTLR4_RUNTIME_LIB_DIR) $(ANTLR4_RUNTIME_LIB) -lcrypto $(ICU_LIBS)
+SHLIB_LINK += -L$(ANTLR4_RUNTIME_LIB_DIR) $(ANTLR4_RUNTIME_LIB) -lcrypto $(ICU_LIBS) -lsybdb
```
thephantomthief: Same here

contrib/babelfishpg_tsql/src/linked_servers.c
```diff
@@ -1297,16 +1350,23 @@ openquery_imp(PG_FUNCTION_ARGS)
...
 	}
 	PG_FINALLY();
 	{
 		if (lsproc)
 		{
-			LINKED_SERVER_DEBUG("LINKED SERVER: (OPENQUERY) - Closing connections to remote server");
-			LINKED_SERVER_EXIT();
+			LINKED_SERVER_DEBUG("LINKED SERVER: (OPENQUERY) - Closing connection to remote server");
+			LINKED_SERVER_CANCEL(lsproc);
+			LINKED_SERVER_CLOSE(lsproc);
```
thephantomthief: Why is this change needed?

contrib/babelfishpg_tsql/src/linked_servers.c
```diff
@@ -1317,6 +1377,420 @@ openquery_imp(PG_FUNCTION_ARGS)
...
+		if ((erc = LINKED_SERVER_RESULTS(validation_lsproc)) == FAIL)
+			ereport(ERROR,
+					(errcode(ERRCODE_FDW_UNABLE_TO_CREATE_EXECUTION),
+					 errmsg("Failed to get results from procedure definition query")));
+		
+		colcount = LINKED_SERVER_NUM_COLS(validation_lsproc);
+		
+		if (colcount > 0)
+		{
+			char bind_definition[65536] = {0x00};  /* Large buffer for procedure body */
```
thephantomthief: We need to change this. `definition` column in sys.sql_modules is nvarchar(max). There is no guarantee that it will fit within 65536 characters.

contrib/babelfishpg_tsql/src/linked_servers.c
```diff
@@ -1317,6 +1377,420 @@ openquery_imp(PG_FUNCTION_ARGS)
...
+				 errhint("Ensure the procedure exists and you have permission to view its definition")));
+	}
+	
+	elog(LOG, "SELECT-only validation: Analyzing procedure definition for %s (length: %zu bytes)",
+		 procedure_name, strlen(definition));
+	
+	/* Perform static analysis - case-insensitive search for forbidden keywords */
+	def_lower = lowerstr(definition);
+	
+	/* Check for DML operations */
```
thephantomthief: The way we are performing static analysis here doesn't seem conclusive. Can't we use our own ANTLR parser to check for DML/DDL?

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -826,6 +830,408 @@ exec_run_dml_with_output(PLtsql_execstate *estate, PLtsql_stmt_push_result *stmt
 	return rc;
 }
 
+#ifdef ENABLE_TDS_LIB
```
thephantomthief, line 834: Any specific reason we want to keep this code outside of linked_server.c?

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -826,6 +830,408 @@ exec_run_dml_with_output(PLtsql_execstate *estate, PLtsql_stmt_push_result *stmt
...
+	Tuplestorestate *tupstore = NULL;
+	Portal portal = NULL;
+	DestReceiver *receiver = NULL;
+	QueryCompletion qc;
+	uint64 processed = 0;
+	ListCell *lc;
+	
+	/* Build the full procedure name: database.schema.procedure */
+	full_proc_name = psprintf("%s.%s.%s",
+							 stmt->db_name ? stmt->db_name : "master",
```
thephantomthief: Assumption to default to master database is not correct. The default database is configurable.

thephantomthief (changes requested):

contrib/babelfishpg_tsql/Makefile
fizzlr: Will remove, it was added for local compilation purpose.

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -2807,6 +2813,28 @@ update_bbf_server_options(char *servername, char *optname, char *optvalue, bool
 				new_record[Anum_bbf_servers_def_connect_timeout - 1] = Int32GetDatum(timeout);
 			}
 		}
+		else if (optname && strlen(optname) == 7 && strncmp(optname, "rpc out", 7) == 0)
```
priyansx, line 2826: Lets start adding automated test cases for all new changes

contrib/babelfishpg_tsql/src/linked_servers.c
fizzlr: This specific one logs server name, will check if other logs log server url

contrib/babelfishpg_tsql/src/pl_exec-2.c
fizzlr: ack, will update with prepared statements to prevent sql injection

contrib/babelfishpg_tsql/src/linked_servers.c
```diff
@@ -1317,6 +1384,608 @@ openquery_imp(PG_FUNCTION_ARGS)
...
+			 * FreeTDS to provide the actual data length, avoiding truncation.
+			 */
+			if (LINKED_SERVER_NEXT_ROW(validation_lsproc) != NO_MORE_ROWS)
+			{
+				void *data = LINKED_SERVER_DATA(validation_lsproc, 1);
+				int data_len = LINKED_SERVER_DATA_LEN(validation_lsproc, 1);
+				
+				if (data && data_len > 0)
+				{
+					/* Allocate exactly the size needed based on actual data length */
```
thephantomthief: Why not use StringInfoBuf here

test/JDBC/expected/remote-proc-exec-vu-verify.out
```diff
@@ -0,0 +1,1205 @@
...
+
+
+-- Test 28: DATE with variable
+-- Expected: Should handle DATE variable
+DECLARE @dt DATE = '2024-01-01';
+EXEC bbf_rpe_server.master.dbo.sp_TestDate @p=@dt;
+GO
+~~ERROR (Code: 33557097)~~
+
+~~ERROR (Message: TDS client library error: DB #: 20114, DB Msg: When passing a SYBINTN, SYBDATETIMN, SYBMONEYN, or SYBFLTN parameter via dbrpcparam, it is necessary to specify the parameter's maximum or actual length so that DB-Library can recognize it as a SYINT1, SYBINT2, SYBINT4, SYBMONEY, SYBMONEY4, and so on, OS #: 0, OS Msg: Success, Level: 7)~~
```
thephantomthief: Is error expected here?

test/JDBC/expected/remote-proc-exec-vu-verify.out
```diff
@@ -0,0 +1,1205 @@
...
+
+-- Test 30: DATETIME with variables
+-- Expected: Should handle datetime variables
+DECLARE @dt1 DATETIME = '2024-06-15 09:30:45';
+DECLARE @dt2 DATETIME2 = '2024-06-15 09:30:45.123';
+EXEC bbf_rpe_server.master.dbo.sp_TestDateTime @p_datetime=@dt1, @p_datetime2=@dt2;
+GO
+~~ERROR (Code: 33557097)~~
+
+~~ERROR (Message: TDS client library error: Msg #: 33557097, Msg state: 1, Msg: timestamp out of range, Server: BABELFISH, Process: , Line: 1, Level: 16)~~
```
thephantomthief: Is error expected here?

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -826,6 +831,742 @@ exec_run_dml_with_output(PLtsql_execstate *estate, PLtsql_stmt_push_result *stmt
 	return rc;
 }
 
+#ifdef ENABLE_TDS_LIB
+/*
+ * Validate remote procedure definition using PostgreSQL parser.
```
thephantomthief: Why are we using PG parser to validate definition of remote procedure?

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -826,6 +831,742 @@ exec_run_dml_with_output(PLtsql_execstate *estate, PLtsql_stmt_push_result *stmt
...
+		int ret;
+		bool isnull;
+		char *data_source;
+		char *db_start;
+		char *db_end;
+		
+		SPI_connect();
+		
+		/* Query babelfish_servers catalog for data_source */
+		plan = SPI_prepare("SELECT data_source FROM sys.babelfish_servers WHERE name = $1",
```
thephantomthief: There is no sys.babelfish_servers catalog. Have we defined one for this feature?

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -843,6 +1584,70 @@ exec_stmt_exec(PLtsql_execstate *estate, PLtsql_stmt_exec *stmt)
...
+			ereport(ERROR,
+					(errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
+					 errmsg("Linked servers are not enabled")));
+
+		/* Check if remote procedure execution is enabled */
+		if (!pltsql_enable_remote_proc_exec)
+			ereport(ERROR,
+					(errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
+					 errmsg("Remote procedure execution via four-part names is not enabled"),
+					 errhint("Execute: SET babelfishpg_tsql.enable_remote_proc_exec = true")));
```
thephantomthief, line 1382: Have we added tests for this GUC?

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -2767,6 +2767,12 @@ update_bbf_server_options(char *servername, char *optname, char *optvalue, bool
 				else
 					new_record[Anum_bbf_servers_def_connect_timeout - 1] = Int32GetDatum(timeout);
 			}
+			else if(strlen(srvOptions_optname[i]) == 7 && strncmp(srvOptions_optname[i], "rpc out", 7) == 0)
+			{
+				/* Handle rpc out as boolean: "true" or "false" */
+				bool rpc_out_enabled = (strcmp(srvOptions_optvalue[i], "true") == 0);
```
thephantomthief: "yes" and "no" are not accepted?

test/JDBC/expected/remote-proc-exec-vu-verify.out
```diff
@@ -0,0 +1,1205 @@
...
+
+
+-- Test 32: TIME with variable
+-- Expected: Should handle TIME variable
+DECLARE @t TIME = '23:59:59';
+EXEC bbf_rpe_server.master.dbo.sp_TestTime @p=@t;
+GO
+~~START~~
+time#!#int
+06:49:30.0000000#!#6
```
thephantomthief: Isn't this output wrong?

test/JDBC/expected/remote-proc-exec-vu-verify.out
```diff
@@ -0,0 +1,1205 @@
...
+~~END~~
+
+
+-- Test 48: Three-part with database omitted (server..schema.procedure)
+-- Expected: Should use current/default database
+EXEC bbf_rpe_server..dbo.sp_NoParams;
+GO
+~~ERROR (Code: 33557097)~~
+
+~~ERROR (Message: relation "sys.babelfish_servers" does not exist)~~
```
thephantomthief: This shouldn't be happening

test/JDBC/expected/remote-proc-exec-vu-verify.out
```diff
@@ -0,0 +1,1205 @@
...
+~~END~~
+
+
+-- Test 50: Two-part (server...procedure)
+-- Expected: Should use defaults for database and schema
+EXEC bbf_rpe_server...sp_NoParams;
+GO
+~~ERROR (Code: 33557097)~~
+
+~~ERROR (Message: relation "sys.babelfish_servers" does not exist)~~
```
thephantomthief: This shouldn't be happening

contrib/babelfishpg_tsql/src/pl_exec-2.c
fizzlr: Fixed. Replaced string concatenation with sp_executesql and proper parameter binding (@proc_name, @schema_name parameters) to prevent SQL injection. The query template is now passed to sp_executesql which handles parameter substitution securely

contrib/babelfishpg_tsql/src/linked_servers.c
fizzlr: Done. Removed buffer binding entirely and now fetch data directly using LINKED_SERVER_DATA()/LINKED_SERVER_DATA_LEN(). This matches the pattern used in openquery_imp() and handles nvarchar(max) columns.

contrib/babelfishpg_tsql/Makefile
fizzlr: Removed.

contrib/babelfishpg_tsql/src/pl_exec-2.c
fizzlr: Fixed. Replaced hard-coded "master" with get_cur_db_name() which respects the user's current database context

test/JDBC/expected/remote-proc-exec-vu-verify.out
```diff
@@ -0,0 +1,1205 @@
...
+EXEC bbf_rpe_server.master.dbo.sp_NonExistent;
+GO
+~~ERROR (Code: 33557097)~~
+
+~~ERROR (Message: Could not fetch definition for procedure master.dbo.sp_nonexistent)~~
+
+
+-- Test 55: Procedure with INSERT statement (violates SELECT-only)
+-- Expected: Should throw error - INSERT not allowed
+EXEC bbf_rpe_server.master.dbo.sp_WithInsert;
```
thephantomthief: This should have thrown error right?

test/JDBC/expected/remote-proc-exec-vu-verify.out
```diff
@@ -0,0 +1,1205 @@
...
+~~END~~
+
+
+-- Test 56: Procedure with UPDATE statement (violates SELECT-only)
+-- Expected: Should throw error - UPDATE not allowed
+EXEC bbf_rpe_server.master.dbo.sp_WithUpdate;
+GO
+~~START~~
+int
+2
```
thephantomthief: Same here

test/JDBC/expected/remote-proc-exec-vu-verify.out
```diff
@@ -0,0 +1,1205 @@
...
+~~END~~
+
+
+-- Test 57: Procedure with DELETE statement (violates SELECT-only)
+-- Expected: Should throw error - DELETE not allowed
+EXEC bbf_rpe_server.master.dbo.sp_WithDelete;
+GO
+~~START~~
+int
+0
```
thephantomthief: Same here

contrib/babelfishpg_tsql/src/procedures.c
fizzlr: Added comprehensive automated test suite (100 test cases) into JDBC tests which is linked servers tests reside.

contrib/babelfishpg_tsql/Makefile
fizzlr: Removed.

contrib/babelfishpg_tsql/src/tsqlUnsupportedFeatureHandler.cpp
fizzlr: Removed.

contrib/babelfishpg_tsql/src/pl_exec-2.c
fizzlr: It's put here since it's tightly integrated with pltsql estate, var and etc. Moving it to linked_server.c would require exposing many internal PLtsql structures. However, if we do want to separate out linked servers from executor code, I can try to refactor and adjust.

contrib/babelfishpg_tsql/src/linked_servers.c
fizzlr: Added LINKED_SERVER_CANCEL before LINKED_SERVER_CLOSE to ensure proper cleanup of pending TDS operations and prevent connection leaks in error paths when metadata queries fail. This also prevent race condition of validation query and actual execution query for their tds connections, found out in local testing.

contrib/babelfishpg_tsql/src/linked_servers.c
fizzlr: Added LINKED_SERVER_CANCEL before LINKED_SERVER_CLOSE to ensure proper cleanup of pending TDS operations and prevent connection leaks in error paths when metadata queries fail. This also prevent race condition of validation query and actual execution query for their tds connections, found out in local testing.

contrib/babelfishpg_tsql/src/linked_servers.c
fizzlr: The current implementations are specific to TDS RPC parameter binding (converting PG types to TDS wire format for dbrpcparam). Tried existing type functions, but it throws errors and inconsistent to types required for freetds library

contrib/babelfishpg_tsql/src/linked_servers.c
fizzlr: The current implementations are specific to TDS RPC parameter binding (converting PG types to TDS wire format for dbrpcparam), which has different requirements than general type conversion.

contrib/babelfishpg_tsql/src/pl_exec-2.c
fizzlr: psprintf is PG standard memory-safe string formatting function and it's widely used in the codebase for dynamic string construction.

contrib/babelfishpg_tsql/src/pl_exec-2.c
fizzlr: "SELECT 1" is a dummy query required by the SPI_cursor_open_with_args() API to create a Portal. The actual result data comes from the remote procedure's tuplestore which is assigned to portal->holdStore. This is a workaround to use PG Portal mechanism for transmitting result sets to the client.

contrib/babelfishpg_tsql/src/linked_servers.c
fizzlr: Updated on this. Will look into the new comments on rev2

contrib/babelfishpg_tsql/src/pl_exec-2.c
fizzlr: pltsql_exec_tsql_cast_value and convert_datum_to_tds_bytes serve different purposes. The former performs T-SQL type casting during expression evaluation (Datum → Datum), while the latter serializes already-typed Datums to TDS wire format (Datum → raw bytes) for RPC parameter binding.

contrib/babelfishpg_tsql/src/linked_servers.c
fizzlr: Other places are using StringInfoBuf to build strings incrementally while here it's one time use for known size data. It also avoids the StringInfo's 1024-byte initial overhead. When the data is large, it also avoids the enlarging buffer multiple times.

test/JDBC/expected/remote-proc-exec-vu-verify.out
fizzlr: no, fixed it.

test/JDBC/expected/remote-proc-exec-vu-verify.out
fizzlr: fixed

contrib/babelfishpg_tsql/src/pl_exec-2.c
fizzlr: good catch. Changed from static analysis to antlr parser. so that it's fixing here using tsql parser.

contrib/babelfishpg_tsql/src/pl_exec-2.c
fizzlr: fixed by removing it. It should be doing the same things at this step similar to existing function. Using the establish connection function now

test/JDBC/expected/remote-proc-exec-vu-verify.out
fizzlr: right, fixed.

test/JDBC/expected/remote-proc-exec-vu-verify.out
fizzlr: right, fixed.

test/JDBC/expected/remote-proc-exec-vu-verify.out
fizzlr: yes, fixed by switching to tsql parser instead of static analysis

test/JDBC/expected/remote-proc-exec-vu-verify.out
fizzlr: yes, fixed by switching to tsql parser instead of static analysis

test/JDBC/expected/remote-proc-exec-vu-verify.out
fizzlr: yes, fixed by switching to tsql parser instead of static analysis

contrib/babelfishpg_tsql/src/linked_servers.c
```diff
@@ -1317,6 +1384,528 @@ openquery_imp(PG_FUNCTION_ARGS)
...
+		Oid datetime2_oid = (*common_utility_plugin_ptr->lookup_tsql_datatype_oid)("datetime2");
+		Oid float_oid = (*common_utility_plugin_ptr->lookup_tsql_datatype_oid)("float");
+		Oid real_oid = (*common_utility_plugin_ptr->lookup_tsql_datatype_oid)("real");
+		Oid varbinary_oid = (*common_utility_plugin_ptr->lookup_tsql_datatype_oid)("varbinary");
+		Oid binary_oid = (*common_utility_plugin_ptr->lookup_tsql_datatype_oid)("binary");
+		Oid uniqueidentifier_oid = (*common_utility_plugin_ptr->lookup_tsql_datatype_oid)("uniqueidentifier");
+		Oid bit_oid = (*common_utility_plugin_ptr->lookup_tsql_datatype_oid)("bit");
+		
+		if (pgtype == varchar_oid)
+			return SYBVARCHAR;
```
thephantomthief: Please avoid referencing direct FreeTDS type macros. Instead use or define macros for these in linked_server.h . This is necessary so that This is necessary to minimise code change if we ever decide to use another client library.

Also please ensure the code compiles with and without the ENABLE_TDS_LIB compiler flag.

contrib/babelfishpg_tsql/src/linked_servers.c
```diff
@@ -1317,6 +1384,528 @@ openquery_imp(PG_FUNCTION_ARGS)
...
+		Oid smalldatetime_oid = (*common_utility_plugin_ptr->lookup_tsql_datatype_oid)("smalldatetime");
+		Oid datetime2_oid = (*common_utility_plugin_ptr->lookup_tsql_datatype_oid)("datetime2");
+		Oid float_oid = (*common_utility_plugin_ptr->lookup_tsql_datatype_oid)("float");
+		Oid real_oid = (*common_utility_plugin_ptr->lookup_tsql_datatype_oid)("real");
+		Oid varbinary_oid = (*common_utility_plugin_ptr->lookup_tsql_datatype_oid)("varbinary");
+		Oid binary_oid = (*common_utility_plugin_ptr->lookup_tsql_datatype_oid)("binary");
+		Oid uniqueidentifier_oid = (*common_utility_plugin_ptr->lookup_tsql_datatype_oid)("uniqueidentifier");
+		Oid bit_oid = (*common_utility_plugin_ptr->lookup_tsql_datatype_oid)("bit");
+		
+		if (pgtype == varchar_oid)
```
thephantomthief: better to use switch ... case

contrib/babelfishpg_tsql/src/linked_servers.c
```diff
@@ -1317,6 +1384,528 @@ openquery_imp(PG_FUNCTION_ARGS)
...
+{
+	if (isnull)
+	{
+		*data_out = NULL;
+		*len_out = 0;
+		return;
+	}
+	
+	/* Check for Babelfish-specific types first */
+	if (common_utility_plugin_ptr && common_utility_plugin_ptr->lookup_tsql_datatype_oid)
```
thephantomthief: There seems to be multiple instances of duplicated code. I think we can combine multiple conditionals here

contrib/babelfishpg_tsql/src/linked_servers.c
thephantomthief: There seems to be multiple instances of duplicated code. I think we can combine multiple conditionals in this function

contrib/babelfishpg_tsql/src/linked_servers.c
```diff
@@ -1317,6 +1384,528 @@ openquery_imp(PG_FUNCTION_ARGS)
...
+				 errmsg("Could not fetch definition for remote procedure %s.%s.%s.%s",
+						server_name, database_name, schema_name, procedure_name),
+				 errhint("Ensure the procedure exists and you have permission to view its definition")));
+	}
+	
+	elog(LOG, "SELECT-only validation: Analyzing procedure definition for %s (length: %zu bytes)",
+		 procedure_name, strlen(definition));
+	
+	/* Use ANTLR parser for accurate T-SQL validation */
+	validate_remote_procedure_select_only_antlr(
```
thephantomthief: If we throw error inside this function, will we free `definition ` and `query.data`?

contrib/babelfishpg_tsql/src/linked_servers.c
```diff
@@ -1317,6 +1384,528 @@ openquery_imp(PG_FUNCTION_ARGS)
...
+	if (definition == NULL)
+	{
+		ereport(ERROR,
+				(errcode(ERRCODE_UNDEFINED_FUNCTION),
+				 errmsg("Could not fetch definition for remote procedure %s.%s.%s.%s",
+						server_name, database_name, schema_name, procedure_name),
+				 errhint("Ensure the procedure exists and you have permission to view its definition")));
+	}
+	
+	elog(LOG, "SELECT-only validation: Analyzing procedure definition for %s (length: %zu bytes)",
```
thephantomthief: We should probably make all these linked server debug logs instead of postgres LOG

contrib/babelfishpg_tsql/src/linked_servers.h
```diff
@@ -104,6 +106,20 @@ typedef DBPROCESS * LinkedServerProcess;
...
+					dbrpcparam(process, name, status, type, maxlen, datalen, value)
+#define LINKED_SERVER_RPC_SEND(process)		dbrpcsend(process)
+#define LINKED_SERVER_RPC_EXEC(process)			dbsqlok(process)
+
+/* RPC OUTPUT parameter retrieval macros */
+#define LINKED_SERVER_NUM_RETS(process)			dbnumrets(process)
+#define LINKED_SERVER_RET_NAME(process, retnum)		dbretname(process, retnum)
+#define LINKED_SERVER_RET_DATA(process, retnum)		dbretdata(process, retnum)
+#define LINKED_SERVER_RET_LEN(process, retnum)		dbretlen(process, retnum)
+#define LINKED_SERVER_RET_TYPE(process, retnum)		dbrettype(process, retnum)
```
thephantomthief, line 122: Don't we need corresponding definition in corresponding #else block like we have for LINKED_SERVER_CANCEL and LINKED_SERVER_CLOSE?

contrib/babelfishpg_tsql/src/procedures.c
thephantomthief: Did we check about this?

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -826,6 +831,503 @@ exec_run_dml_with_output(PLtsql_execstate *estate, PLtsql_stmt_push_result *stmt
...
+		
+		elog(LOG, "DEBUG_LINKED_SERVER: (RPC) - Remote procedure execution completed successfully");
+	}
+	PG_FINALLY();
+	{
+		/* Clean up parameter buffers if error occurred before RPC send completed */
+		if (param_data_buffers != NIL)
+		{
+			ListCell *buf_cell;
+			elog(LOG, "DEBUG_LINKED_SERVER: (CLEANUP) - Freeing parameter buffers in error path");
```
thephantomthief: Let's be consistent and use linked server debug logging here

contrib/babelfishpg_tsql/src/tsqlIface.cpp
```diff
@@ -848,6 +855,114 @@ clear_tables_info()
  * will be also moved. tsqlBuilder will focus on create Pltsql_stmt_* only.
  */
 
+////////////////////////////////////////////////////////////////////////////////
+// Remote Procedure SELECT-only Validator
+////////////////////////////////////////////////////////////////////////////////
+class RemoteProcedureSelectOnlyValidator : public TSqlParserBaseVisitor
```
thephantomthief, line 881: What have we decided on nested procedures? I think we won't be able to block those cases with this function.

test/JDBC/expected/BABEL-2455.out
```diff
@@ -320,7 +320,7 @@ select yourserver.master.dbo.f1(1);
 go
 ~~ERROR (Code: 33557097)~~
 
-~~ERROR (Message: Remote procedure/function reference with 4-part object name is not currently supported in Babelfish)~~
+~~ERROR (Message: improper qualified name (too many dotted names): ourserver.master.dbo.f1)~~
```
thephantomthief: This is wrong error please fix it.

test/JDBC/expected/openquery-vu-verify.out
```diff
@@ -3587,7 +3587,7 @@ GO
 text
 ~~ERROR (Code: 33557097)~~
 
-~~ERROR (Message: TDS client library error: DB #: 20003, DB Msg: server connection timed out, OS #: 0, OS Msg: Success, Level: 6)~~
+~~ERROR (Message: TDS client library error: DB #: 20047, DB Msg: DBPROCESS is dead or not enabled, OS #: 0, OS Msg: Success, Level: 9)~~
```
thephantomthief: Why is this error changing now?

test/JDBC/expected/remote-proc-exec-vu-verify.out
```diff
@@ -0,0 +1,1211 @@
...
+
+
+-- Test 15: NVARCHAR with special characters
+-- Expected: Should preserve special characters
+DECLARE @nstr NVARCHAR(100) = N'Special: !@#$%^&*()_+-={}[]|:;<>?,./';
+EXEC bbf_rpe_server.master.dbo.sp_TestNvarchar @p=@nstr;
+GO
+~~START~~
+varchar#!#int
+Special: !@#$%^&*()_+-={}[]|:;<>?,./#!#36
```
thephantomthief, line 171: These are not special characters for nvarchar, these are ascii characters which will work for varchar. Please tests for nvarchar specific chars too like emojis and kanji characters

test/JDBC/expected/remote-proc-exec-vu-verify.out
```diff
@@ -0,0 +1,1211 @@
...
+~~END~~
+
+
+-- Test 16: CHAR and NCHAR types
+-- Expected: Should handle fixed-length strings
+EXEC bbf_rpe_server.master.dbo.sp_TestChar @p_char='CHAR', @p_nchar=N'NCHAR';
```
thephantomthief, line 177: Same for nchar

test/JDBC/expected/remote-proc-exec-vu-verify.out
```diff
@@ -0,0 +1,1211 @@
...
+~~START~~
+varchar
+NULL value
+~~END~~
+
+
+-- Test 19: Large string (>8000 characters)
+-- Expected: Should handle large VARCHAR(MAX)
+DECLARE @large_str VARCHAR(MAX) = REPLICATE('A', 10000);
+EXEC bbf_rpe_server.master.dbo.sp_LargeString @p=@large_str;
```
thephantomthief, line 208: please add tests for nvarchar(max)

test/JDBC/expected/remote-proc-exec-vu-verify.out
```diff
@@ -0,0 +1,1211 @@
...
+~~END~~
+
+
+-- Test 23: VARBINARY parameter
+-- Expected: Should handle binary data
+EXEC bbf_rpe_server.master.dbo.sp_TestBinary @p=0xDEADBEEF;
+GO
+~~START~~
+binary#!#int
+DEADBEEF000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000#!#4
```
thephantomthief, line 257: Are the trailing zeros expected?

test/JDBC/expected/remote-proc-exec-vu-verify.out
```diff
@@ -0,0 +1,1211 @@
...
+
+
+-- Test 26: UNIQUEIDENTIFIER with variable
+-- Expected: Should handle GUID variable
+DECLARE @guid UNIQUEIDENTIFIER = '87654321-4321-4321-4321-CBA987654321';
+EXEC bbf_rpe_server.master.dbo.sp_TestUUID @p=@guid;
+GO
+~~ERROR (Code: 33557097)~~
+
+~~ERROR (Message: TDS client library error: DB #: 20114, DB Msg: When passing a SYBINTN, SYBDATETIMN, SYBMONEYN, or SYBFLTN parameter via dbrpcparam, it is necessary to specify the parameter's maximum or actual length so that DB-Library can recognize it as a SYINT1, SYBINT2, SYBINT4, SYBMONEY, SYBMONEY4, and so on, OS #: 0, OS Msg: Success, Level: 7)~~
```
thephantomthief: This is not expected right?

test/JDBC/expected/remote-proc-exec-vu-verify.out
thephantomthief: This does not look fixed

test/JDBC/expected/remote-proc-exec-vu-verify.out
```diff
@@ -0,0 +1,1211 @@
...
+
+
+-- Test 29: DATETIME and DATETIME2
+-- Expected: Should handle both datetime types
+EXEC bbf_rpe_server.master.dbo.sp_TestDateTime @p_datetime='2023-12-25 14:30:00', @p_datetime2='2023-12-25 14:30:00.1234567';
+GO
+~~START~~
+datetime#!#datetime2
+2023-12-25 14:30:00.0#!#2023-12-25 14:30:00.1234570
```
thephantomthief, line 324: Is extra precision expected?

test/JDBC/expected/remote-proc-exec-vu-verify.out
thephantomthief: Does not look fixed

test/JDBC/expected/remote-proc-exec-vu-verify.out
```diff
@@ -0,0 +1,1211 @@
...
+~~ERROR (Message: TDS client library error: Msg #: 33557097, Msg state: 1, Msg: timestamp out of range, Server: BABELFISH, Process: , Line: 1, Level: 16)~~
+
+
+-- Test 31: TIME parameter
+-- Expected: Should handle TIME type
+EXEC bbf_rpe_server.master.dbo.sp_TestTime @p='14:30:00';
+GO
+~~START~~
+time#!#int
+14:30:00.0000000#!#14
```
thephantomthief, line 346: Is extra precision expected?

test/JDBC/expected/remote-proc-exec-vu-verify.out
thephantomthief: Please address this.

test/JDBC/expected/remote-proc-exec-vu-verify.out
```diff
@@ -0,0 +1,1211 @@
...
+GO
+~~START~~
+int#!#varchar#!#varchar#!#numeric#!#bit#!#uniqueidentifier
+42#!#Test String#!#Unicode String#!#123.456#!#1#!#12345678-1234-1234-1234-123456789ABC
+~~END~~
+
+
+-- Test 40: Empty result set
+-- Expected: Should return 0 rows but proper column structure
+EXEC bbf_rpe_server.master.dbo.sp_EmptyResult;
```
thephantomthief, line 476: We should add a test where procedure returns more than one result set

test/JDBC/expected/remote-proc-exec-vu-verify.out
```diff
@@ -0,0 +1,1211 @@
...
+-- Test 57: Procedure with DELETE statement (violates SELECT-only)
+-- Expected: Should throw error - DELETE not allowed
+EXEC bbf_rpe_server.master.dbo.sp_WithDelete;
+GO
+~~ERROR (Code: 33557097)~~
+
+~~ERROR (Message: Remote procedure contains DELETE statement)~~
+
+
+-- Test 58: Procedure with dynamic SQL (violates SELECT-only)
```
thephantomthief: Please also add tests blocking CREATE, and other DDLs. Nested procedures too

test/JDBC/expected/remote-proc-exec-vu-verify.out
```diff
@@ -0,0 +1,1211 @@
...
+-- Test 68: Procedure with RETURN code (negative)
+-- Expected: Should capture return code = -1
+DECLARE @ret2 INT;
+EXEC @ret2 = bbf_rpe_server.master.dbo.sp_ReturnCode @value=-10;
+SELECT @ret2 AS return_code;
+GO
+~~START~~
+int
+0
+~~END~~
```
thephantomthief: These are capturing wrong return code. Please fix this.

test/JDBC/expected/remote-proc-exec-vu-verify.out
```diff
@@ -0,0 +1,1211 @@
...
+-- Expected: Should handle escaped quotes
+EXEC bbf_rpe_server.master.dbo.sp_TestVarchar @p='It''s a test';
+GO
+~~START~~
+varchar#!#int
+It's a test#!#11
+~~END~~
+
+
+-- Test 77: NVARCHAR with Unicode characters (if supported)
```
thephantomthief: We do support it. Why are we seeing `?` here?

test/JDBC/expected/remote-proc-exec-vu-verify.out
```diff
@@ -0,0 +1,1211 @@
...
+-- Expected: Should preserve Unicode
+DECLARE @unicode NVARCHAR(100) = N'Hello 世界 مرحبا';
+EXEC bbf_rpe_server.master.dbo.sp_TestNvarchar @p=@unicode;
+GO
+~~START~~
+varchar#!#int
+Hello ?? ?????#!#14
```
thephantomthief: Again, test says `Expected: Should preserve Unicode` but actual test output is not doing that.

test/JDBC/expected/remote-proc-exec-vu-verify.out
```diff
@@ -0,0 +1,1211 @@
...
+GO
+~~START~~
+varchar#!#int
+This should work#!#1
+~~END~~
+
+
+-- Test 82: SQL Injection attempt in procedure name (should fail safely)
+-- Expected: Should throw error - procedure not found (not SQL injection)
+EXEC bbf_rpe_server.master.dbo.[sp_NoParams'; DROP TABLE test; --];
```
thephantomthief: Please try injection in server name, database name and procedure name. Also, I don't think the procedure name injection is actually getting tested here. Having `'` in a procedure name is perfectly valid in T-SQL if it wrapped in `[]`

test/JDBC/expected/remote-proc-exec-vu-verify.out
```diff
@@ -0,0 +1,1211 @@
...
+~~END~~
+
+
+-- Test 95: NVARCHAR with emoji/symbols (if supported)
+-- Expected: Should preserve Unicode symbols
+EXEC bbf_rpe_server.master.dbo.sp_TestNvarchar @p=N'Test ★ ♥ ✓ ⚡';
+GO
+~~START~~
+varchar#!#int
+Test ? ? ? ?#!#12
```
thephantomthief: We should see the actual symbols

test/JDBC/expected/xml_exist-vu-verify.out
```diff
@@ -649,7 +649,7 @@ SELECT dbo.babel_5222_xml_exist_t1.XmlColumn.exist('/artists/artist/@name') FROM
 GO
 ~~ERROR (Code: 33557097)~~
 
-~~ERROR (Message: Remote procedure/function reference with 4-part object name is not currently supported in Babelfish)~~
+~~ERROR (Message: improper qualified name (too many dotted names): bo.babel_5222_xml_exist_t1.xmlcolumn.exist)~~
```
thephantomthief: This error message should not have been thrown

test/JDBC/expected/xml_value-vu-verify.out
```diff
@@ -1069,7 +1069,7 @@ SELECT dbo.babel_5223_xml_value_t1.XmlColumn.value('(/artists/artist/@name)[1]',
 GO
 ~~ERROR (Code: 33557097)~~
 
-~~ERROR (Message: Remote procedure/function reference with 4-part object name is not currently supported in Babelfish)~~
+~~ERROR (Message: improper qualified name (too many dotted names): bo.babel_5223_xml_value_t1.xmlcolumn.value)~~
```
thephantomthief: This error message should not have been thrown

test/JDBC/expected/remote-proc-exec-vu-verify.out
```diff
@@ -0,0 +1,1211 @@
+
+-- ========================================
```
thephantomthief, line 2: Please run these or add new tests for upgrade

contrib/babelfishpg_tsql/src/linked_servers.c
fizzlr: got it, I will actually remove them since they are not necessary

test/JDBC/expected/openquery-vu-verify.out
fizzlr: I noticed some old tests are referring to server options using number like 3 directly. Since now we add another server option rpc out, it crashed. I will need to fix relevant code.

contrib/babelfishpg_tsql/sql/linked_servers_tsql.sql
```diff
@@ -1,7 +1,8 @@
 CREATE TABLE sys.babelfish_server_options (
 	servername sys.SYSNAME NOT NULL PRIMARY KEY COLLATE "C",
 	query_timeout INT,
-	connect_timeout INT
+	connect_timeout INT,
+	rpc_out BOOLEAN DEFAULT FALSE
```
forestkeeper, line 5: 1. We can't change that directly, should use the upgrade script to alter the table add new columns
2. Also in this script, we should also use alter table after create table.

thephantomthief: We can modify this. This will only be run on a fresh provisioning of Babelfish database. This is not the upgrade script. The ALTER statement should be part of the upgrade script.

contrib/babelfishpg_tsql/src/linked_servers.c
```diff
@@ -47,45 +56,72 @@ Oid		tdsTypeToOid(int datatype);
...
 
+static int
+linked_server_msg_handler(LinkedServerProcess lsproc, int error_code, int state, int severity, char *error_msg, char *svr_name, char *proc_name, int line)
+{
 	/*
 	 * If error severity is greater than 10, we interpret it as a T-SQL error;
-	 * otheriwse, a T-SQL info
+	 * otherwise, a T-SQL info
 	 */
-	appendStringInfo(
```
forestkeeper, line 62: Why deleting those ?

contrib/babelfishpg_tsql/src/linked_servers.c
```diff
@@ -762,6 +806,9 @@ linked_server_establish_connection(char *servername, LinkedServerProcess * lspro
 					 errmsg("Error fetching foreign server with servername '%s'", servername)
 					 ));
 
+		/* LOG: Foreign server retrieved successfully */
+		elog(LOG, "DEBUG_LINKED_SERVER: Retrieved foreign server configuration for: %s", servername);
```
forestkeeper: Can we get rid of those logs ?

contrib/babelfishpg_tsql/src/linked_servers.c
```diff
@@ -1317,6 +1414,874 @@ openquery_imp(PG_FUNCTION_ARGS)
...
+	return false;
+}
+
+/*
+ * Extract nested EXEC/EXECUTE procedure calls from T-SQL definition
+ * Uses split_object_name() to parse multi-part names
+ * Returns List of NestedProcedureInfo structures
+ */
+static List *
+extract_nested_procedure_calls(const char *definition)
```
forestkeeper: Can you give an example for what this function will return ?

test/JDBC/expected/remote-proc-exec-vu-verify.out
```diff
@@ -0,0 +1,1379 @@
...
+-- SECTION 9: ERROR AND NEGATIVE TESTS
+-- ========================================
+-- Test 51: Invalid server name
+-- Expected: Should throw error - server not found
+EXEC invalid_server.master.dbo.sp_NoParams;
+GO
+~~ERROR (Code: 33557097)~~
+
+~~ERROR (Message: RPC out is not enabled for the specified server. Use sp_serveroption to enable it.)~~
```
forestkeeper: Error msg is not 'server not found' , so it's different from expected

test/JDBC/expected/remote-proc-exec-vu-verify.out
```diff
@@ -0,0 +1,1379 @@
...
+~~ERROR (Message: Remote database not found: database "invalid_db" does not exist)~~
+
+
+-- Test 53: Invalid schema name
+-- Expected: Should throw error - schema not found
+EXEC bbf_rpe_server.master.invalid_schema.sp_NoParams;
+GO
+~~ERROR (Code: 33557097)~~
+
+~~ERROR (Message: Could not fetch definition for remote procedure bbf_rpe_server.master.invalid_schema.sp_noparams)~~
```
forestkeeper: The error is not 'schema not found'

test/JDBC/expected/remote-proc-exec-vu-verify.out
```diff
@@ -0,0 +1,1379 @@
...
+~~ERROR (Message: Remote procedure contains INSERT statement)~~
+
+
+-- Test 56: Procedure with UPDATE statement (violates SELECT-only)
+-- Expected: Should throw error - UPDATE not allowed
+EXEC bbf_rpe_server.master.dbo.sp_WithUpdate;
+GO
+~~ERROR (Code: 33557097)~~
+
+~~ERROR (Message: Remote procedure contains UPDATE statement)~~
```
forestkeeper: Do you need change the error msg to Update not allowed ?

test/JDBC/expected/remote-proc-exec-vu-verify.out
```diff
@@ -0,0 +1,1379 @@
...
+
+-- Test 109: Nested procedure with DELETE in inner (should FAIL)
+-- Expected: Should fail with error about DELETE in nested procedure
+PRINT 'Test 109: Outer calls inner with DELETE - should fail';
+GO
+EXEC bbf_rpe_server.master.dbo.sp_Outer_CallsDelete;
+GO
+~~ERROR (Code: 33557097)~~
+
+~~ERROR (Message: Remote procedure contains DELETE statement)~~
```
forestkeeper: Can you add a test case like : 
insert into local table with exec from a remote server, thanks

contrib/babelfishpg_tsql/src/linked_servers.c
fizzlr: These statements were development-time debugging logs that were producing excessive output in production PG log files. Every single remote procedure call was generating 40+ LOG-level messages (connection parameters, query text, row-by-row data, etc.).

I converted them to a two-tier debug strategy:

- __19 retained__ as `LINKED_SERVER_DEBUG(...)` (maps to `elog(DEBUG1, ...)`) — visible only with `SET client_min_messages TO debug1`
- __46 removed entirely__ — redundant messages (e.g., "about to call function X" followed by "called function X successfully")

The `LINKED_SERVER_DEBUG` macro is defined in linked_servers.h and matches the existing pattern used by the OPENQUERY code path.

contrib/babelfishpg_tsql/src/linked_servers.c
fizzlr: Defined `LS_TYPE_*` abstraction macros in `linked_servers.h` (e.g., `LS_TYPE_VARCHAR`, `LS_TYPE_INT4`) with corresponding `#else` stubs, replacing all direct FreeTDS references like `SYBVARCHAR`.

contrib/babelfishpg_tsql/src/linked_servers.c
fizzlr: Consolidated `get_tds_type_from_pg_oid()` — Babelfish OIDs are runtime values so they use a single flat if-chain with category comments; standard PG OIDs use a proper switch/case.

contrib/babelfishpg_tsql/src/linked_servers.c
fizzlr: Consolidated 4 separate `if (common_utility_plugin_ptr ...)` blocks into a single block with all 23 Babelfish OID lookups in one place.

contrib/babelfishpg_tsql/src/linked_servers.c
fizzlr: Extracted `datum_text_to_tds_bytes()` and `datum_output_to_tds_bytes()` helpers, replacing duplicate code blocks.

contrib/babelfishpg_tsql/src/linked_servers.c
fizzlr: Added null-safe `pfree()` calls for both `definition` and `query.data` on all paths, and initialized `query.data = NULL` so `PG_FINALLY` can safely check it.

contrib/babelfishpg_tsql/src/linked_servers.c
fizzlr: Removed 46 `elog(LOG)` statements entirely and converted 19 retained logs to `LINKED_SERVER_DEBUG()`/`LINKED_SERVER_DEBUG_FINER()` macros (maps to `DEBUG1`/`DEBUG2`).

contrib/babelfishpg_tsql/src/linked_servers.h
fizzlr: Added complete `#else` stubs for all new macros: `LINKED_SERVER_RPC_INIT`, `LINKED_SERVER_RPC_PARAM`, `LINKED_SERVER_RPC_SEND`, `LINKED_SERVER_RPC_EXEC`, `LINKED_SERVER_RET_STATUS`, `LINKED_SERVER_NUM_RETS`, and all `LS_TYPE_*`/`LS_DB*` type macros.

contrib/babelfishpg_tsql/src/pl_exec-2.c
fizzlr: Converted all `elog(LOG, ...)` statements in the RPC execution path to `LINKED_SERVER_DEBUG(...)` macro calls for consistency

test/JDBC/expected/remote-proc-exec-vu-verify.out
fizzlr: Added `LINKED_SERVER_RET_STATUS` macro wrapping `dbretstatus()` and capture logic via `pltsql_proc_return_code` to correctly propagate return codes from remote procedures.

contrib/babelfishpg_tsql/src/linked_servers.c
fizzlr: Removed 46 development-time `elog(LOG)` statements; retained 19 as `LINKED_SERVER_DEBUG()` (only visible at `DEBUG1` level) for troubleshooting.

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -330,8 +330,8 @@ initTsqlSyscache()
 bool
 IsPLtsqlExtendedCatalog(Oid relationId)
 {
-	/* Skip during Babelfish restore */
-	if (!babelfish_dump_restore && (relationId == sysdatabases_oid ||
+	/* Skip during Babelfish restore or extension creation/upgrade */
+	if (!babelfish_dump_restore && !creating_extension && (relationId == sysdatabases_oid ||
```
forestkeeper, line 334: Why we need this change ?

contrib/babelfishpg_tsql/src/catalog.c
```diff
@@ -1423,6 +1423,42 @@ get_timeout_from_server_name(char *servername, int attnum)
...
+get_rpc_out_option(char *servername)
+{
+	Relation	bbf_servers_def_rel;
+	HeapTuple	tuple;
+	ScanKeyData	key;
+	TableScanDesc	scan;
+	bool		rpc_out_enabled = false;  /* Default to disabled */
+
+	bbf_servers_def_rel = table_open(get_bbf_servers_def_oid(),
+										 RowExclusiveLock);
```
forestkeeper: Why using RowExclusiveLock in here ?

forestkeeper: The query time out option use : https://github.com/amazon-aurora/babelfish_extensions/blob/f3bad22b8d8b8c4eee9cdbb8361418bb25f4a345/contrib/babelfishpg_tsql/src/catalog.c#L1402C12-L1402C27 accesssharelock

contrib/babelfishpg_tsql/src/linked_servers.c
```diff
@@ -22,6 +24,13 @@
 #define LINKED_SERVER_DEBUG(...)	elog(DEBUG1, __VA_ARGS__)
 #define LINKED_SERVER_DEBUG_FINER(...)	elog(DEBUG2, __VA_ARGS__)
 
+/*
+ * Note: UTF-8 to UTF-16 LE conversion is not needed here because FreeTDS
+ * handles the character encoding internally when we use XSYBNVARCHAR/XSYBNCHAR
+ * types. We send UTF-8 data and FreeTDS converts it to UTF-16 LE as needed
+ * for the TDS protocol wire format.
+ */
```
forestkeeper, line 33: Is this comment useful in here ?

contrib/babelfishpg_tsql/src/linked_servers.c
```diff
@@ -47,45 +56,72 @@ Oid		tdsTypeToOid(int datatype);
...
-
+	const char *clean_msg = error_msg;
+	
 	initStringInfo(&buf);
+	
+	/* Format based on common error codes */
+	switch (error_code)
+	{
+		case 201:  /* Missing or invalid parameter */
+			appendStringInfo(&buf, "Procedure parameter error: %s", clean_msg);
```
forestkeeper: Is the error code mapping behavior the same as SQL Server ?

forestkeeper: If not, we need to make sure the error code mapping is the same as SQL Server.

contrib/babelfishpg_tsql/src/linked_servers.c
```diff
@@ -1317,6 +1406,840 @@ openquery_imp(PG_FUNCTION_ARGS)
...
+ * Babelfish type OIDs are resolved at runtime via lookup_tsql_datatype_oid(),
+ * so they cannot be used in switch/case labels (C requires constant expressions).
+ * We check all Babelfish types in a single if-chain first, then fall through
+ * to a switch on standard PostgreSQL OIDs (compile-time constants).
+ */
+int
+get_tds_type_from_pg_oid(Oid pgtype)
+{
+	/* Check for Babelfish-specific types (runtime OIDs, cannot use switch) */
+	if (common_utility_plugin_ptr && common_utility_plugin_ptr->lookup_tsql_datatype_oid)
```
forestkeeper: Those code duplicates with code defined in typecodes.c

contrib/babelfishpg_tsql/src/linked_servers.c
```diff
@@ -1317,6 +1406,840 @@ openquery_imp(PG_FUNCTION_ARGS)
...
+
+/*
+ * Convert a Datum value to raw bytes suitable for TDS RPC parameter binding.
+ * Returns palloc'd buffer containing the data.
+ *
+ * Babelfish type OIDs are resolved at runtime, so they are checked in a
+ * single if-chain first. Standard PostgreSQL types use a switch statement.
+ */
+void
+convert_datum_to_tds_bytes(Datum value, Oid valtype, int32 valtypmod, bool isnull,
```
forestkeeper, line 1654: Consider to reuse the tds_typeio.h codes, not re-implement the convert datum code again.

contrib/babelfishpg_tsql/src/catalog.c
fizzlr: Thanks good find, it's changed recently in https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/4320/changes#diff-21156ac859506a645f7ccb05bfd9cde8e8c967aaa736531bdbca14de0f427044. That updated the lock of previous server option implementation. Will update.

contrib/babelfishpg_tsql/src/linked_servers.c
```diff
@@ -1168,6 +1285,9 @@ openquery_imp(PG_FUNCTION_ARGS)
 
 		LINKED_SERVER_DEBUG("LINKED SERVER: (OPENQUERY) - Executing query against remote server");
 
+		/* LOG: Query execution starting */
+		LINKED_SERVER_DEBUG("Executing query on remote server");
```
thephantomthief: These seem duplicated

PR #4319, opened 2025-12-16
Block ALTER .. OWNER .. of TSQL objects via PG port

shalinilohia50:
### Description
Block ALTER .. OWNER .. of TSQL objects via PG port except for the cases when the GUC babelfishpg_tsql.dump_restore or babelfishpg_tsql.enable_alter_owner_from_pg is set to true.

### Issue
Task: BABEL-6142


contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -2874,6 +2874,31 @@ bbf_ProcessUtility(PlannedStmt *pstmt,
 		}
 	}
 
+	/*
+	 * Block ALTER .. OWNER TO .. statements from PG dialect
+	 * executed on TSQL objects.
+	 */
+	if (sql_dialect == SQL_DIALECT_PG && !babelfish_dump_restore && !pltsql_enable_alter_owner_from_pg)
+	{
+		switch (nodeTag(parsetree))
```
rishabhtanwar29: Why can't we do this in the existing switch-case below?

contrib/babelfishpg_tsql/src/guc.c
```diff
@@ -1127,6 +1128,17 @@ define_custom_variables(void)
 							 PGC_USERSET,
 							 GUC_NOT_IN_SAMPLE | GUC_DISALLOW_IN_FILE | GUC_DISALLOW_IN_AUTO_FILE,
 							 NULL, NULL, NULL);
+	/*
+	 * Block ALTER .. OWNER .. from PG endpoint
+	 */
+	DefineCustomBoolVariable("babelfishpg_tsql.enable_alter_owner_from_pg",
+							 gettext_noop("Enables blocked ALTER .. OWNER .. statements from PG endpoint"),
```
rishabhtanwar29: nit: seems a bit confusing, we should rather say: `"Allows ALTER .. OWNER .. statements on Babelfish objects from PostgreSQL endpoint"`

contrib/babelfishpg_tsql/src/pltsql_utils.c
```diff
@@ -3093,3 +3093,137 @@ is_valid_func_ownership_chain(void *expr, Oid objectOwnerId)
...
+		}
+        default:
+            return; /* Allow other object types */
+    }
+
+    /* Get schema OID if we have schema name */
+    if (schema_name)
+        schema_oid = get_namespace_oid(schema_name, true);
+    else if (OidIsValid(schema_oid))
+        schema_name = get_namespace_name(schema_oid);
```
rishabhtanwar29: I think this is quite risky and not a full proof to get the schema name/oid, we should rather do it similar to how the actual ExecAlterOwnerStmt function does it, which is that first use `get_object_address` to get the object Oid and then use `get_object_namespace` to get the schema.

contrib/babelfishpg_tsql/src/pltsql_utils.c
```diff
@@ -3093,3 +3093,137 @@ is_valid_func_ownership_chain(void *expr, Oid objectOwnerId)
...
+    /* Check if it's a Babelfish schema */
+    if (schema_name && physical_schema_name_exists(schema_name))
+    {
+        /* Check if this is a shipped function being changed to sysadmin */
+        if (object_name && stmt->objectType == OBJECT_FUNCTION &&
+            is_ms_shipped((char*)object_name, OBJECT_TYPE_TSQL_SCALAR_FUNCTION, schema_oid))
+        {
+            /* Check if new owner is sysadmin */
+            if (stmt->newowner && stmt->newowner->rolename &&
+                strcmp(stmt->newowner->rolename, "sysadmin") == 0)
```
rishabhtanwar29: Let's use `get_role_name` instead.

contrib/babelfishpg_tsql/src/pltsql_utils.c
```diff
@@ -3093,3 +3093,137 @@ is_valid_func_ownership_chain(void *expr, Oid objectOwnerId)
...
+                return; /* Allow ALTER OWNER to sysadmin for shipped objects */
+            }
+        }
+		/* Check if this is a shipped stored procedure being changed to sysadmin */
+        else if (object_name && stmt->objectType == OBJECT_PROCEDURE &&
+                 is_ms_shipped((char*)object_name, OBJECT_TYPE_TSQL_STORED_PROCEDURE, schema_oid))
+        {
+            /* Check if new owner is sysadmin */
+            if (stmt->newowner && stmt->newowner->rolename &&
+                strcmp(stmt->newowner->rolename, "sysadmin") == 0)
```
rishabhtanwar29: Same as above, let's use `get_role_name`

contrib/babelfishpg_tsql/src/pltsql_utils.c
```diff
@@ -3093,3 +3093,137 @@ is_valid_func_ownership_chain(void *expr, Oid objectOwnerId)
...
+        {
+            if (stmt->relation->schemaname)
+                schema_name = stmt->relation->schemaname;
+            else
+                schema_oid = RangeVarGetRelid(stmt->relation, NoLock, true);
+        }
+
+        /* Get schema OID if we have schema name */
+        if (schema_name)
+            schema_oid = get_namespace_oid(schema_name, true);
```
rishabhtanwar29: What's the point of getting schema oid here if we are not going to use it?

contrib/babelfishpg_tsql/src/pltsql_utils.c
```diff
@@ -3093,3 +3093,137 @@ is_valid_func_ownership_chain(void *expr, Oid objectOwnerId)
...
+        else if (OidIsValid(schema_oid))
+            schema_name = get_namespace_name(schema_oid);
```
rishabhtanwar29: Instead evaluate the schema_name from oid above at the line number 3211.

contrib/babelfishpg_tsql/src/pltsql_utils.c
```diff
@@ -3093,3 +3093,137 @@ is_valid_func_ownership_chain(void *expr, Oid objectOwnerId)
...
+
+    /* Check if any command is AT_ChangeOwner */
+    foreach(lcmd, stmt->cmds)
+    {
+        AlterTableCmd *cmd = (AlterTableCmd *) lfirst(lcmd);
+        if (cmd->subtype != AT_ChangeOwner)
+            continue;
+
+        /* Extract schema from RangeVar */
+        if (stmt->relation)
```
rishabhtanwar29: This check should be outside of the for loop once.

contrib/babelfishpg_tsql/src/pltsql_utils.c
```diff
@@ -3093,3 +3093,137 @@ is_valid_func_ownership_chain(void *expr, Oid objectOwnerId)
...
+        schema_oid = get_namespace_oid(schema_name, true);
+    else if (OidIsValid(schema_oid))
+        schema_name = get_namespace_name(schema_oid);
+
+    /* Check if it's a Babelfish schema */
+    if (schema_name && physical_schema_name_exists(schema_name))
+    {
+        /* Check if this is a shipped function being changed to sysadmin */
+        if (object_name && stmt->objectType == OBJECT_FUNCTION &&
+            is_ms_shipped((char*)object_name, OBJECT_TYPE_TSQL_SCALAR_FUNCTION, schema_oid))
```
rishabhtanwar29: Can you please put a comment on why we need this special handling for ms shipped objects with some examples?

contrib/babelfishpg_tsql/src/pltsql_utils.c
shalinilohia50: `get_role_name` is a static function defined in tdsutils.c. Hence, fetching the new owner name like `stmt->newowner->rolename`.

contrib/babelfishpg_tsql/src/pltsql_utils.c
shalinilohia50: `get_role_name` is a static function defined in tdsutils.c. Hence, fetching the new owner name like `stmt->newowner->rolename`.

contrib/babelfishpg_tsql/src/pl_handler.c
shalinilohia50: We can, but I feel this is more neat to handle the block cases like we have done for views in past under the GUC `enable_create_alter_view_from_pg`

contrib/babelfishpg_tsql/src/pltsql_utils.c
rishabhtanwar29: So then let's use `get_rolespec_oid` and `get_sysadmin_oid` together for matching instead of name comparison.

contrib/babelfishpg_tsql/src/pl_handler.c
rishabhtanwar29: IMO, having several switch-case makes the code quite complex and highly unmaintainable as we keep adding newer and newer cases, so I still feel we should reuse the existing ones unless there is a strong reason not to do so.

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -2874,6 +2874,31 @@ bbf_ProcessUtility(PlannedStmt *pstmt,
 		}
 	}
 
+	/*
+	 * Block ALTER .. OWNER TO .. statements from PG dialect
+	 * executed on TSQL objects.
+	 */
+	if (sql_dialect == SQL_DIALECT_PG && !babelfish_dump_restore && !pltsql_enable_alter_owner_from_pg)
```
rishabhtanwar29: Also, I think we should have superuser check here to allow everything to superuser.

contrib/babelfishpg_tsql/src/pltsql_utils.c
```diff
@@ -3093,3 +3093,146 @@ is_valid_func_ownership_chain(void *expr, Oid objectOwnerId)
...
+            ObjectWithArgs *owa = (ObjectWithArgs *) stmt->object;
+            if (list_length(owa->objname) >= 1)
+                object_name = strVal(llast(owa->objname));
+
+            /*
+             * Special handling for MS shipped system objects - system functions/procedures
+             * created in dbo schemas, unlike other system objects. We allow changing their
+             * ownership to 'sysadmin' as part of internal SQL and upgrade scripts.
+             * One instance: 'ALTER PROCEDURE master_dbo.xp_qv OWNER TO sysadmin;'
+             * We should allow these ALTER OWNER statements for shipped objects.
```
rishabhtanwar29: I think having a superuser check at the top will make it redundant since these alter statements will only be in SQL/upgrade scripts which runs in superuser context.

contrib/babelfishpg_tsql/src/pltsql_utils.c
```diff
@@ -3093,3 +3093,146 @@ is_valid_func_ownership_chain(void *expr, Oid objectOwnerId)
...
+        return;
+
+    /* Extract schema from RangeVar */
+    if (stmt->relation)
+    {
+        if (stmt->relation->schemaname)
+            schema_name = stmt->relation->schemaname;
+        else
+        {
+            schema_oid = RangeVarGetRelid(stmt->relation, NoLock, true);
```
rishabhtanwar29: Let's take the same lock as AT_ChangeOwner would take so that there is no search_path related vulnerability.

contrib/babelfishpg_tsql/src/guc.c
rishabhtanwar29: I think you missed to update this?

contrib/babelfishpg_tsql/src/pltsql_utils.c
```diff
@@ -3093,3 +3093,146 @@ is_valid_func_ownership_chain(void *expr, Oid objectOwnerId)
...
+    Oid			schema_oid = InvalidOid;
+    char		*schema_name = NULL;
+    char		*object_name = NULL;
+    ObjectAddress	address;
+    Relation		relation = NULL;
+    Node			*object = stmt->object;
+
+    /* Only handle specific object types */
+    if (stmt->objectType != OBJECT_TYPE && stmt->objectType != OBJECT_SCHEMA &&
+        stmt->objectType != OBJECT_FUNCTION && stmt->objectType != OBJECT_PROCEDURE)
```
rishabhtanwar29, line 3112: We create a lot of other types of objects in babelfish, what about them? Like OBJECT_DOMAIN, OBJECT_OPERATOR, OBJECT_ROUTINE, OBJECT_COLLATION etc. Why do we want to restrict to only these?

shalinilohia50: Those are not user defined objects. We restrict only these objects mentioned in the [doc](https://quip-amazon.com/deUEAW2xBdaH/Altering-the-owner-of-Babelfish-Objects-should-be-blocked-via-PG-endpoint).

contrib/babelfishpg_tsql/src/pl_handler.c
shalinilohia50: After discussion with DBEs, we decided that we will block ALTER OWNER for all the users (rds_superuser/superuser).

To provide flexibility when needed, we've implemented a GUC that can bypass this restriction. When enabled, the GUC allows ALTER OWNER operations based on the user's actual privileges, following standard PostgreSQL permission checks.

contrib/babelfishpg_tsql/src/pltsql_utils.c
shalinilohia50: After discussion with DBEs, we decided that we will block ALTER OWNER for all the users (rds_superuser/superuser).
I have handled ms_shipped objects separately since its needed in the SQL/upgrade scripts.

contrib/babelfishpg_tsql/src/pl_handler.c
HarshLunagariya: +1

contrib/babelfishpg_tsql/src/pltsql_utils.c
HarshLunagariya: What if in future , We implement a new TSQL object type support using a different Postgresql object internally?

contrib/babelfishpg_tsql/src/pltsql_utils.c
```diff
@@ -3093,3 +3093,149 @@ is_valid_func_ownership_chain(void *expr, Oid objectOwnerId)
...
+			{
+				Oid newowner_oid = get_rolespec_oid(stmt->newowner, false);
+				if (OidIsValid(newowner_oid) && newowner_oid == get_sysadmin_oid())
+				{
+					if ((stmt->objectType == OBJECT_FUNCTION &&
+						is_ms_shipped((char*)object_name, OBJECT_TYPE_TSQL_SCALAR_FUNCTION, schema_oid)) ||
+						(stmt->objectType == OBJECT_PROCEDURE &&
+						is_ms_shipped((char*)object_name, OBJECT_TYPE_TSQL_STORED_PROCEDURE, schema_oid)))
+					{
+						pfree(schema_name);
```
HarshLunagariya: Any testcase is there?

shalinilohia50: As discussed with Rishabh, we are planning to allow all these ALTER OWNER statements for superuser.  These checks won't be required anymore.

contrib/babelfishpg_tsql/src/pltsql_utils.c
shalinilohia50: We might need to revisit this. Also, that type might need a different handling depending on the use case.

contrib/babelfishpg_tsql/src/pltsql_utils.c
```diff
@@ -3093,3 +3093,116 @@ is_valid_func_ownership_chain(void *expr, Oid objectOwnerId)
...
+            continue;
+
+        /* Check if it's a Babelfish schema */
+        if (schema_name && physical_schema_name_exists(schema_name))
+        {
+            ereport(ERROR,
+                    (errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
+                     errmsg("ALTER .. OWNER .. is blocked in PG dialect on TSQL objects. Please set babelfishpg_tsql.enable_alter_owner_from_pg to true to enable.")));
+        }
+        break;
```
rishabhtanwar29: I think we shouldn't break here since we want to traverse through all the AT_ChangeOwner commands and not just the first one.

contrib/babelfishpg_tsql/src/pltsql_utils.c
```diff
@@ -3093,3 +3093,116 @@ is_valid_func_ownership_chain(void *expr, Oid objectOwnerId)
...
+    /* Check if it's a Babelfish schema */
+    if (physical_schema_name_exists(schema_name))
+    {
+        pfree(schema_name);
+        ereport(ERROR,
+                (errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
+                 errmsg("ALTER .. OWNER .. is blocked in PG dialect on TSQL objects. Please set babelfishpg_tsql.enable_alter_owner_from_pg to true to enable.")));
+    }
+
+    if (schema_name)
```
rishabhtanwar29: nit (non-blocking): seems redundant due to null check at line 3147.

contrib/babelfishpg_tsql/src/guc.c
rishabhtanwar29: nit: still not been addressed 🙂

rishabhtanwar29 (approved):

HarshLunagariya (approved):

PR #4321, opened 2025-12-16
Fix to prevent dropping a login that owns a database

pranavJ23:
### Description

Currently we allow droplogin even when login is owner of database which leads to an inconsistent state.
If the login is owner of database then we should throw error. Added logic to throw error when login is owner of a database.
Added function is_database_owner which checks if the current login is owner of any database or not. If found to be owner of a database then return true else returns false.

### Issues Resolved

[BABEL-4376]

### Check List


contrib/babelfishpg_tsql/src/rolecmds.c
```diff
@@ -2286,6 +2286,47 @@ is_active_login(Oid role_oid)
...
+
+	db_rel = table_open(sysdatabases_oid, AccessShareLock);
+	scan = table_beginscan_catalog(db_rel, 0, NULL);
+
+	while ((tuple = heap_getnext(scan, ForwardScanDirection)) != NULL)
+	{
+		Datum owner_datum = heap_getattr(tuple, Anum_sysdatabases_owner,
+										 db_rel->rd_att, &is_null);
+
+		if (!is_null && strcmp(role_name, NameStr(*DatumGetName(owner_datum))) == 0)
```
tanscorpio7: CStringGetDatum can be used directly instead of `NameStr(*DatumGetName`

contrib/babelfishpg_tsql/src/rolecmds.c
```diff
@@ -2286,6 +2286,47 @@ is_active_login(Oid role_oid)
...
+	TableScanDesc	scan;
+	HeapTuple		tuple;
+	bool			is_null;
+	char			*role_name;
+	bool			is_owner = false;
+
+	/* get the role name from oid */
+	role_name = GetUserNameFromId(role_oid, true);
+	if (!role_name)
+		return false;
```
tanscorpio7: missing_ok should be false

tanscorpio7 (changes requested):

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -4445,6 +4445,11 @@ bbf_ProcessUtility(PlannedStmt *pstmt,
 							ereport(ERROR,
 									(errcode(ERRCODE_OBJECT_IN_USE),
 									 errmsg("Could not drop login '%s' as the user is currently logged in.", role_name)));
+														
+						if (is_database_owner(role_oid))
+							ereport(ERROR,
+									(errcode(ERRCODE_OBJECT_IN_USE),
+									 errmsg("Could not drop login '%s' as the login is owner of some database.", role_name)));
```
tanscorpio7: 1. Error message - `Login 'rolename' owns one or more database(s). Change the owner of the database(s) before dropping the login.`
2. Please point me to the test cases for this.

tanscorpio7 (changes requested):

contrib/babelfishpg_tsql/src/pl_handler.c
pranavJ23: test added here: https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/4321/files#diff-7a286a41987f2e6f3326205063c7569de7886ce9a912df79e390e9fd1dca04e5R124

tanscorpio7: There error message isn't this one in this test case.

pranavJ23: Updated the error message accordingly

shalinilohia50 (approved):

test/JDBC/expected/test_drop_login-vu-verify.out
```diff
@@ -100,6 +113,26 @@ GO
...
 
+-- tsql
+-- should throw error
+use master
+go
+drop login l1_owner
+go
+~~ERROR (Code: 33557097)~~
+
+~~ERROR (Message: Login 'l1_owner' owns one or more database(s). Change the owner of the database(s) before dropping the login.)~~
```
tanscorpio7, line 124: Can we also test 
1. ALTER AUTHORIZATION DATABASE OWNER ?
2. cleanup.sh script in JDBC i.e. sys.remove_babelfish()
3. Repeatable read isolation level where at the start of transaction login does not have any owner but some concurrent transaction makes it an owner.
```
-- txn1 (session 1)
BEGIN TRAN;
SET TRAN ISOLATION LEVEL REPEATABLE READ;
SELECT * FROM table1; -- this is necessary
-- execute concurrent txn2
DROP LOGIN l1;
COMMIT;

-- txn2 (session 2)
ALTER AUTHORIZATION some_db OWNER TO l1;
```

tanscorpio7 (changes requested):

test/JDBC/expected/test_drop_login-vu-verify.out
pranavJ23: I ran a transaction as:
```
BEGIN TRAN; SELECT * FROM t1; PRINT 'Transaction started at: ' + CONVERT(VARCHAR, GETDATE(), 120);DECLARE @i INT = 1;WHILE @i <= 100000 BEGIN     INSERT INTO t1(a) VALUES (@i);          IF @i % 1000 = 0     BEGIN         PRINT 'Loop iteration: ' + CAST(@i AS VARCHAR) +                ' at ' + CONVERT(VARCHAR, GETDATE(), 120);     END          SET @i = @i + 1; END PRINT ''; PRINT 'Loop completed at: ' + CONVERT(VARCHAR, GETDATE(), 120); SELECT 'l1 owns:' AS Status, name FROM sys.databases  WHERE owner_sid = SUSER_SID('l1'); DROP LOGIN l1; COMMIT;

```
in session 1 
and in session 2 ran during the time loop of transaction 1 was running.
```
ALTER AUTHORIZATION on database::db1 TO l1;
```
in session 2 I got the output as:
```
1> ALTER AUTHORIZATION on database::db1 TO l1;
2> go
1> 
```
In output of transaction ran in session 1 is:
```
Login 'l1' owns one or more database(s). Change the owner of the database(s) before dropping the login.
```
This is expected we are making the login l1 owner of database db1 in session2 before running drop login command.
If we do not run the session2 alter authorisation command then drop login is running without any error

tanscorpio7: If isolation level of txn1 is repeatable read then txn1 will not see commited changes from other parallel txns

pranavJ23: With the current method to scan catalog via ```table_beginscan_catalog```, we will be able to see the alter authorization changes via trnx2 in trnx1 even when trnx1 is ```repeatable read```, hence will not be able to drop the login in trnx1 as well. Reason being in the ```table_beginscan_catalog``` we are taking the catalog current snapshot(```GetCatalogSnapshot```). 
We make transaction1 isolated from transaction2 by using ```systable_beginscan``` function with snapshot as ```GetTransactionSnapshot()```, this will lead to successful drop of login in trnx1 but this will lead database to inconsistent state.

tanscorpio7: Makes sense, please add test cases for it. Thanks.

tanscorpio7 (approved):

PR #4334, opened 2025-12-18
Setting recurse to true for cases where we want to find the ENRs by traversing parent queryEnvs

ayushdsh:
### Description

* Function - `pltsql_get_tsql_enr_from_oid` is used in the below cases - 
    * lmgr -> within CheckRelationLockedByMe to check if the given relation is locked by a bknd. this is safe to go over parent queryEnvs.
    * lock -> when trying to acquire a lock on a rel. This is what causes the waiting when trying to operate on temp table in the parent queryEnv scope. we've also verified the behaviour with TSQL.
    * inval -> when deciding whether we want to put inval msg in the SI queue. This is also not a problem.
We will want to go recursively to search for ENR in all cases; hence setting recurse to true.

* Function - `object_name` is used for getting the name given obj_oid and db_oid. But in TSQL, temp tables do not have a object_id and hence object_name also does not work. Hence, in this case, we will keep the recurse = false.

### Issues Resolved

BABEL-5605, BABEL-6192

Engine PR: https://github.com/babelfish-for-postgresql/postgresql_modified_for_babelfish/pull/665

### Check List

### DIFF
--- a/contrib/babelfishpg_tsql/runtime/functions.c
+++ b/contrib/babelfishpg_tsql/runtime/functions.c
@@ -2793,7 +2793,7 @@ object_name(PG_FUNCTION_ARGS)
 	 * search in list of ENRs registered in the current query environment by
 	 * object_id
 	 */
-	enr = GetENRTempTableWithOid(object_id, true);
+	enr = GetENRTempTableWithOid(object_id, false);
 	if (enr != NULL && enr->md.enrtype == ENR_TSQL_TEMP)
 	{
 		PG_RETURN_VARCHAR_P((VarChar *) cstring_to_text(enr->md.name));
--- a/contrib/babelfishpg_tsql/runtime/functions.c
+++ b/contrib/babelfishpg_tsql/runtime/functions.c
@@ -2793,7 +2793,7 @@ object_name(PG_FUNCTION_ARGS)
 	 * search in list of ENRs registered in the current query environment by
 	 * object_id
 	 */
-	enr = GetENRTempTableWithOid(object_id);
+	enr = GetENRTempTableWithOid(object_id, true);
 	if (enr != NULL && enr->md.enrtype == ENR_TSQL_TEMP)
 	{
 		PG_RETURN_VARCHAR_P((VarChar *) cstring_to_text(enr->md.name));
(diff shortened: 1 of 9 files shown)


tanscorpio7 (dismissed):

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -2793,7 +2793,7 @@ object_name(PG_FUNCTION_ARGS)
 	 * search in list of ENRs registered in the current query environment by
 	 * object_id
 	 */
-	enr = GetENRTempTableWithOid(object_id);
+	enr = GetENRTempTableWithOid(object_id, true);
```
tanscorpio7: The comment above is stale now, please update it. Also verify that SQL Server supports object name for temp tables created in a different scope.

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -5333,7 +5333,7 @@ pltsql_is_local_only_inval_msg(const SharedInvalidationMessage *msg)
 static EphemeralNamedRelation
 pltsql_get_tsql_enr_from_oid(const Oid oid)
```
tanscorpio7, line 5334: Similar to `GetENRTempTableWithOid` are all occurneces of `pltsql_get_tsql_enr_from_oid` recursion safe ?

tanscorpio7 (changes requested):

tanscorpio7:
Please update the PR title and description.

contrib/babelfishpg_tsql/src/hooks.c
ayushdsh: Yes. Particularly, we use `pltsql_get_tsql_enr_from_oid` in -
1. lmgr -> within `CheckRelationLockedByMe` to check if the given relation is locked by a bknd. this is safe to go over parent queryEnvs.
2. lock -> when trying to acquire a lock on a rel. This is what causes the waiting when trying to operate on temp table in the parent queryEnv scope. we've also verified the behaviour with TSQL.
3. inval -> when deciding whether we want to put inval msg in the SI queue. This is also not a problem.

contrib/babelfishpg_tsql/runtime/functions.c
ayushdsh: Well, with TSQL, `object_id` also does not work for temp tables, so I doubt `object_name` would. But I can test it out.

tanscorpio7 (dismissed):

shardgupta (approved):

PR #4342, opened 2025-12-23
Handling valid characters in  string to money conversion

Vineetha125:
### Description

In Babelfish, the logic implemented for casting string to money added stricter validation for invalid characters. This change affecting certain cases in` ISNUMERIC() `and inadvertently led to rejecting some characters that T-SQL allows, such as:

- Backslash (\) - CHAR(92)
- Comma (,) - CHAR(44) 
- Non-breaking space - CHAR(160) 

This PR fixes the regression by modifying the money parsing logic to skip these characters, matching T-SQL behavior.

### Issues Resolved

BABEL-6270

* **Use case based -**  YES

* **Arbitrary inputs -** YES

* **Performance tests -** NA

* **Tooling impact -** NA

* **Client tests -** NA

### Check List


contrib/babelfishpg_money/fixeddecimal.c
```diff
@@ -529,21 +529,40 @@ scanfixeddecimal(const char *str, int *precision, int *scale, FunctionCallInfo *
...
+		
+		/* Skip ASCII whitespace */
+		if (isspace(c))
+		{
+			ptr++;
+			continue;
+		}
+		
+		/* Skip comma (thousand separator) and backslash */
+		if (c == ',' || c == '\\')
```
rohit01010: Can you please test following type of combinations
```
select cast('\\1234' as money)
go

select cast(',\,1234' as money)
go
```

contrib/babelfishpg_money/fixeddecimal.c
```diff
@@ -657,9 +676,24 @@ scanfixeddecimal(const char *str, int *precision, int *scale, FunctionCallInfo *
 			ptr++, vscale++;
 	}
 
-	/* consume any remaining space chars */
-	while (isspace((unsigned char) *ptr))
-		ptr++;
+	while (*ptr != '\0')
```
Yvinayak07: Do we have test cases for this condition being placed here?

Yvinayak07 (commented):
Could you also test (if this gets fixed by this PR):
- SELECT isnumeric(char(44));
- SELECT isnumeric(CHAR(92));

contrib/babelfishpg_money/fixeddecimal.c
Vineetha125: Added. Thanks

contrib/babelfishpg_money/fixeddecimal.c
Vineetha125: yes, added few test cases. Thanks

tanyagupta17 (commented):
Let's try to find if there is a pattern of acceptable combinations of symbols/characters while casting to money in TSQL?
As it seems to allow the following 
```
-- TSQL
1> select cast(',,,,,,,124' as money)
2> go
                     
---------------------
             124.0000

4> select cast('$,,123' as money)
5> go
                     
---------------------
             123.0000

1> select cast('$,123' as money)
2> go
                     
---------------------
             123.0000
```
But not this 
```
-- TSQL
1> select cast(',$123' as money)
2> go
Msg 235, Level 16, State 1, Server EC2AMAZ-T4MF7PS, Line 1
Cannot convert a char value to money. The char value has incorrect syntax.

1> select cast(',\\13454' as money);
2> go
Msg 235, Level 16, State 1, Server EC2AMAZ-T4MF7PS, Line 1
Cannot convert a char value to money. The char value has incorrect syntax.

1> select cast(',,,,,$,,124' as money)
2> go
Msg 235, Level 16, State 1, Server EC2AMAZ-T4MF7PS, Line 1
Cannot convert a char value to money. The char value has incorrect syntax.
1> select cast('$$$$$$123' as money)
2> go
Msg 235, Level 16, State 1, Server EC2AMAZ-T4MF7PS, Line 1
Cannot convert a char value to money. The char value has incorrect syntax.
```

contrib/babelfishpg_money/fixeddecimal.c
```diff
@@ -517,33 +571,147 @@ scanfixeddecimal(const char *str, int *precision, int *scale, FunctionCallInfo *
...
+			continue;
+		}
+
+		if (c == ',')
+		{
+			const char *check;
+			if (isdigit((unsigned char) ptr[1]))
+			{
+				ptr++;
+				continue;
```
pranavJ23: The failure in test framework(during install money extension)seems to be due to the failure during build extension:
```
babelfishpg_money--1.1.0.sql
fixeddecimal.c: In function 'scanfixeddecimal':
fixeddecimal.c:607:25: error: continue statement not within a loop
  607 |                         continue;
      |                         ^~~~~~~~
fixeddecimal.c:616:33: error: continue statement not within a loop
  616 |                                 continue;
      |                                 ^~~~~~~~
make: *** [<builtin>: fixeddecimal.o] Error 1
make: *** Waiting for unfinished jobs....
```
Fixing this should fix the failure during install extension

Vineetha125: Thanks

contrib/babelfishpg_money/fixeddecimal.c
```diff
@@ -580,7 +641,7 @@ scanfixeddecimal(const char *str, int *precision, int *scale, FunctionCallInfo *
 			}
 			integralpart = tmp;
 			/* skip thousand separator */
-			if (*ptr == ',')
+			while (*ptr == ',')
```
Yvinayak07: I think we can keep this if only as this condition itself is placed in a while loop.

contrib/babelfishpg_money/fixeddecimal.c
```diff
@@ -474,13 +474,53 @@ fixeddecimal2str(int64 val, char *buffer,
 	return ptr;
 }
 
+
+static inline bool
+is_money_space(unsigned char c)
```
rohit01010: why we are using unsigned char as argument type for `is_money_space` and `is_utf8_nbsp` function ?

contrib/babelfishpg_money/fixeddecimal.c
```diff
@@ -496,8 +536,22 @@ scanfixeddecimal(const char *str, int *precision, int *scale, FunctionCallInfo *
...
+			unsigned char c = (unsigned char)*check;
+			if (c == 0xA0 || (c == 0xC2 && (unsigned char)*(check+1) == 0xA0))
+			{
+				ereturn(escontext, (Datum) 0,
+						(errcode(ERRCODE_NUMERIC_VALUE_OUT_OF_RANGE),
+						errmsg("invalid characters found: cannot cast value \"%s\" to money",
+								str)));
+			}
+		}
+	}
```
rohit01010, line 554: can you please add some comments on this code block to understand what we are trying to do here ?

contrib/babelfishpg_money/fixeddecimal.c
```diff
@@ -496,8 +536,22 @@ scanfixeddecimal(const char *str, int *precision, int *scale, FunctionCallInfo *
 	 */
 
 	/* skip leading spaces */
-	while (isspace((unsigned char) *ptr))
```
rohit01010, line 499: Lets add comment with allowed string pattern which can be converted to `money/smallmoney`, to understand how we are breaking down and validating input string.

contrib/babelfishpg_money/fixeddecimal.c
```diff
@@ -517,33 +571,38 @@ scanfixeddecimal(const char *str, int *precision, int *scale, FunctionCallInfo *
...
-		!isdigit((unsigned char) *ptr) && 
+	if (!isdigit((unsigned char) *ptr) && 
 		*ptr != '.' && 
 		*ptr != '-' && 
-		*ptr != '+')
+		*ptr != ',' &&
+		*ptr != '+' &&
+		*ptr != '\0' &&
+		!is_money_space((unsigned char)*ptr) &&
+		!is_utf8_nbsp((const unsigned char *)ptr))
```
rohit01010: This line `ptr = skip_money_spaces(ptr);` should have already removed any spaces/non-breaking spaces from the start of the string, why do we again want to check the same thing ?

contrib/babelfishpg_money/fixeddecimal.c
```diff
@@ -651,15 +722,43 @@ scanfixeddecimal(const char *str, int *precision, int *scale, FunctionCallInfo *
...
+		
+		/* UTF-8 NBSP (0xC2 0xA0) - accepted ONLY at trailing position */
+		if (is_utf8_nbsp((const unsigned char *)ptr))
+		{
+			ptr += 2; 
+			continue;
+		}
+		
+		break;
+	}
```
rohit01010: why are we not using `skip_money_spaces` function here ?

contrib/babelfishpg_money/fixeddecimal.c
rohit01010: `is_money_space` is just a single line function and also it is not even being used at lot of places, why do we need to create a separate function for this ?

contrib/babelfishpg_money/fixeddecimal.c
```diff
@@ -474,13 +474,53 @@ fixeddecimal2str(int64 val, char *buffer,
...
+{
+    return isspace(c) || c == 0xA0;
+}
+
+static inline bool
+is_utf8_nbsp(const unsigned char *ptr)
+{
+    if (ptr[0] != 0xC2)
+        return false;
+    return ptr[1] == 0xA0;
```
rohit01010: can be refactored as `return (ptr[0] == 0xC2 && ptr[1] == 0xA0)`. Also, similar to `is_money_space`, this is just a single line function and also it is not even being used at lot of places, why do we need to create a separate function for this ?

contrib/babelfishpg_money/fixeddecimal.c
Vineetha125: Agreed. Removed the function as it is not required for single line check.

contrib/babelfishpg_money/fixeddecimal.c
Vineetha125: Removed the function. Thanks

contrib/babelfishpg_money/fixeddecimal.c
Vineetha125: Added. Thanks

contrib/babelfishpg_money/fixeddecimal.c
Vineetha125: Added.

contrib/babelfishpg_money/fixeddecimal.c
Vineetha125: Yes, your right. Removed the check.

contrib/babelfishpg_money/fixeddecimal.c
Vineetha125: Replaced this block of code with the existing` skip_money_spaces `function. Thanks

contrib/babelfishpg_money/fixeddecimal.c
Vineetha125: Removed `is_money_space` and `is_utf8_nbsp `functions. Thanks!

contrib/babelfishpg_money/fixeddecimal.c
```diff
@@ -474,13 +474,41 @@ fixeddecimal2str(int64 val, char *buffer,
...
+        }
+        
+        /* UTF-8 NBSP (0xC2 0xA0) */
+        if (c == 0xC2 && (unsigned char)ptr[1] == 0xA0)
+        {
+            ptr += 2;
+            continue;
+        }
+        
+        break;
```
Yvinayak07: We should modify this block to `if-else`  block in that way we will not be needing jump statement like `continue`.

contrib/babelfishpg_money/fixeddecimal.c
```diff
@@ -495,9 +523,36 @@ scanfixeddecimal(const char *str, int *precision, int *scale, FunctionCallInfo *
...
+	 /* skip leading spaces */
+	ptr = skip_money_spaces(ptr);
+
+    /* Rejects NBSP at leading position */ 
+	if (ptr > original && *ptr != '\0')
+	{
+		for (const char *check = original; check < ptr; check++)
+		{
+			unsigned char c = (unsigned char)*check;
+			if (c == 0xA0 || (c == 0xC2 && (unsigned char)*(check+1) == 0xA0))
```
Yvinayak07, line 546: Do we have test case for this condition? if yes, could you point that test?

contrib/babelfishpg_money/fixeddecimal.c
```diff
@@ -517,33 +572,36 @@ scanfixeddecimal(const char *str, int *precision, int *scale, FunctionCallInfo *
 		}
 	}
 
-	/* skip leading spaces */
-	while (isspace((unsigned char) *ptr))
-		ptr++;
+	// /* skip leading spaces */
```
Yvinayak07: NIT: please remove `//`.

contrib/babelfishpg_money/fixeddecimal.c
Vineetha125: Modified. Thanks

contrib/babelfishpg_money/fixeddecimal.c
rohit01010: Instead of skipping leading NBSP and then rechecking if there were any NBSP, why don't we just skip only whitespaces and just check whether the next character is NBSP or not ?

test/JDBC/expected/babel_string_to_money_reg-vu-verify.out
```diff
@@ -0,0 +1,1238 @@
...
+~~ERROR (Message: invalid characters found: cannot cast value " 123.45" to money)~~
+
+SELECT CAST(NCHAR(160) + NCHAR(160) + '99.99' AS MONEY);
+GO
+~~START~~
+money
+~~ERROR (Code: 293)~~
+
+~~ERROR (Message: invalid characters found: cannot cast value "  99.99" to money)~~
+
```
Vineetha125, line 172: > Do we have test case for this condition? if yes, could you point that test?

These are the test cases that addresses the condition (NBSP at leading position) , also we have : `SELECT ISNUMERIC(NCHAR(160) + '123');`. Thanks!

contrib/babelfishpg_money/fixeddecimal.c
Vineetha125: > Do we have test case for this condition? if yes, could you point that test?

Yes, we do have, I have pointed them.

contrib/babelfishpg_money/fixeddecimal.c
Vineetha125: That's a great point, and I initially considered the single-pass approach as well. However, the challenge is that SQL Server accepts cases where NBSP can appear mixed with regular whitespaces, like:
Example:
```
SELECT CAST(NCHAR(160) + ' ' AS MONEY);
GO
~~START~~
money
0.0000
```
In this example, NBSP is followed by a regular space, and SQL Server still treats it as a valid input. If we use a single-pass logic (skipping only whitespaces and then checking for NBSP), we would incorrectly throw an error for such valid cases.
The existing two-pass approach ensures we handle all the combinations that SQL Server accepts. Thanks

Yvinayak07 (approved):

rohit01010 (approved):

shardgupta (approved):

PR #4372, opened 2026-01-02
Fixing typmod for MIN()/MAX() on CHAR/NCHAR datatypes 

tanyagupta17:
### Description

Earlier in Babelfish, we were losing out on typmod information in case of MIN/MAX() function having CHAR/NCHAR as argument, which was leading to error and requiring explicit casting to Varchar/Nvarchar.

With this commit, we are using `pltsql_post_transform_expr_recurse` function to detect MIN/MAX aggregates on CHAR/NCHAR types and preserve the original column's typmod by adding a relabel node with the correct type modifier.

### Issues Resolved

BABEL-5688

* **Use case based -** Yes

* **Arbitrary inputs -** Yes

* **Performance tests -** NA

* **Tooling impact -** NA

* **Client tests -** NA

### Check List


test/JDBC/expected/TestDatatypeAggSort-before-16_12-or-17_8-vu-prepare.out
```diff
@@ -0,0 +1,165 @@
...
+	'2023-11-06 00:00:00.000 +0:00',
+	'2023-11-06 00:00:00.000'
+), (
+	NULL, NULL, NULL, NULL, NULL, NULL
+);
+go
+~~ROW COUNT: 9~~
+
+
+CREATE INDEX TestDatatypeAggSort_vu_prepare_char_idx ON TestDatatypeAggSort_vu_prepare_tbl (char_col)
```
tanscorpio7, line 66: Please add a tests cases for
1. Index scan still being used even with the explicit cast on top.
2. Explicit COLLATE clause on top of aggregate. `SELECT 1 WHERE 'ä' = (SELECT MAX(col1 COLLATE Latin1_General_CI_AI ) FROM t1)` where MAX(col1) = 'A'`

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -8387,6 +8387,43 @@ pltsql_post_transform_expr_recurse(ParseState *pstate, Node *expr)
 			}
 			break;
 		}
+		case T_Aggref:
+			{
+				Aggref		*aggref = (Aggref *) expr;
+				char		*aggFuncName = get_func_name(aggref->aggfnoid);
+
+				/* Handle MIN/MAX for char/nchar types to preserve typmod */
+				if (aggFuncName && strlen(aggFuncName) == 3 &&
```
tanscorpio7: 1. Please move this to a different function i.e. similar to `tsql_set_typmod_op_expr`.
2. Why is the RELABEL node special handling not need here considering nchar/nvarchar are also domain types ?

tanscorpio7 (changes requested):

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -8387,6 +8387,43 @@ pltsql_post_transform_expr_recurse(ParseState *pstate, Node *expr)
...
+					(strncmp(aggFuncName, "min", 3) == 0 || strncmp(aggFuncName, "max", 3) == 0) &&
+					((*common_utility_plugin_ptr->is_tsql_bpchar_datatype)(aggref->aggtype) ||
+					 (*common_utility_plugin_ptr->is_tsql_nchar_datatype)(aggref->aggtype)))
+				{
+					TargetEntry	*te = (TargetEntry *) linitial(aggref->args);
+					Oid			expr_type = exprType((Node *) te->expr);
```
Yvinayak07: I think we should have a check that `te` should not be NULL. Their could be a case where the `aggref->args` could be NULL. if not a check then we can have an assert check.

test/JDBC/expected/TestDatatypeAggSort-before-16_12-or-17_8-vu-prepare.out
tanyagupta17: Added in `string-aggregate-vu-*` files

contrib/babelfishpg_tsql/src/hooks.c
tanyagupta17: 1. Moved
2. I have added various tests from different datatypes to nchar/nvarchar/char etc and we don't seem to need special handling for RELABEL node, as we are able to find coercion path unlike binary where we add relable nodes

contrib/babelfishpg_tsql/src/hooks.c
tanyagupta17: Added check on `aggref->args`

tanscorpio7 (dismissed):

rohit01010 (approved):

Yvinayak07 (approved):

shardgupta (approved):

PR #4382, opened 2026-01-06
Fix pattern matching for like operator

Mohitraj-g:
### Description

```
DROP TABLE IF EXISTS DummyTable
GO

CREATE TABLE [dbo].[DummyTable](
 [ViewDetailsID] [varchar](50) NULL
) 

 GO  

  SELECT 1 FROM INFORMATION_SCHEMA.COLUMNS 
        WHERE TABLE_NAME = 'DummyTable' 
        AND COLUMN_NAME like 'ViewDetailsID ' 
```

16.11 returned 1, which is incorrect because it matches ViewDetailsID even though the pattern includes a trailing space. Now, we have fixed it to return no results, which is the expected behavior since the column name does not contain a trailing space. 

### Issues Resolved

BABEL-6223

* **Use case based -**

* **Boundary conditions -**

* **Arbitrary inputs -**

* **Negative test cases -**

* **Performance tests -**

* **Tooling impact -**

* **Client tests -**

### Check List


tanscorpio7:
Please explain the fix the in PR description

contrib/babelfishpg_tsql/src/collation.c
```diff
@@ -485,6 +485,10 @@ optimise_likenode(Node *node, OpExpr *op, like_ilike_info_t like_entry, coll_inf
 
 	Assert(ltypeId == rtypeId);
 
+	/* Always create a CollateExpr on top to match with op->inputcollid */
+	linitial(op->args) = (Node*) create_collate_expr(linitial(op->args), op->inputcollid);
+	lsecond(op->args) = (Node *) create_collate_expr(lsecond(op->args), op->inputcollid);
+
```
tanscorpio7, line 491: Please explain why are we deleting the COLLATE expression from arguments and then adding it back.

contrib/babelfishpg_tsql/src/collation.c
```diff
@@ -502,7 +506,8 @@ optimise_likenode(Node *node, OpExpr *op, like_ilike_info_t like_entry, coll_inf
 									InvalidOid,
 									coll_info_of_inputcollid.oid,
 									oprfuncid(optup)));
-		
+
+		ret = make_and_qual(ret, node);
```
tanscorpio7, line 510: Are there any reverese case ? where LIKE will return from some operation but equal to will not ?
col LIKE 'some_exact_pattern' --> true
col = 'some_exact_pattern' --> false

tanscorpio7 (changes requested):

contrib/babelfishpg_tsql/src/collation.c
Mohitraj-g: Yes, for example for varchar datatype:
Let's consider column new_test_like_index_vu_prepare_v_csas_col has an entry 'csas' --> (without any space)

If we use '=' operator with right operand with a space, the query would return true:
```
SELECT new_test_like_index_vu_prepare_v_csas_col 
FROM new_test_like_index_vu_prepare_v_csas 
WHERE new_test_like_index_vu_prepare_v_csas_col = 'csas ' --> (has a space)
```
Returns:
```
csas
```
As varchar removes the trailing spaces.

But for the LIKE operator
```
SELECT new_test_like_index_vu_prepare_v_csas_col 
FROM new_test_like_index_vu_prepare_v_csas 
WHERE new_test_like_index_vu_prepare_v_csas_col COLLATE latin1_general_cs_as LIKE 'csas ';
```
we do no expect any rows to be returned (TSQL behaviour).

contrib/babelfishpg_tsql/src/collation.c
Mohitraj-g: This enables us to have the correct collation information in the query plan. Else, the Const node/pattern has COLLATE 'default'. What we want is COLLATE 'input_collation'

Deepesh125: please check if we can optimize some code along with line 448

tanscorpio7 (approved):

PR #4410, opened 2026-01-12
Fix: Crash when dropping table variable related temp tables due to being in PG dialect

ayushdsh:
### Description

In BABEL-6227, nested calls corrupted the dialect to 'postgres' causing table variable temp table cleanup to fail. In BABEL-6239, nested calls left the dialect as 'postgres' causing subsequent TSQL queries to be parsed with the wrong parser.

This is because of using both direct assignment and set_config_option() function together in a nested call stack. The set_config_option() function is PGs GUC setting infrastructure which internally maintains a prev value stack. This does not work well, if caller functions have used direct assignment instead because it would mean that the value does not get registered in the internal GUC stack, leading to restoration of incorrect value for sql_dialect.

The fix for this is to use direct assignment in the culprit function which is using set_config_option. Previously, we've made the decision to use direct assignment instead of using set_config because of perf related issues.

### Issues Resolved

Task - BABEL-6227, BABEL-6239

### Check List


contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -6416,7 +6416,9 @@ pltsql_call_handler(PG_FUNCTION_ARGS)
 		 * validating a PL/tsql program because the validator function is not
 		 * written in PL/tsql, it's written in C.
 		 */
-		sql_dialect = SQL_DIALECT_TSQL;
+		set_config_option("babelfishpg_tsql.sql_dialect", "tsql",
+						  GUC_CONTEXT_CONFIG,
+						  PGC_S_SESSION, GUC_ACTION_SAVE, true, 0, false);
```
tanscorpio7: Changing this in fmgr.c is too risky. This issue only happens when there is a combinartion of set_config call and direct variable assignment. I think we should change the set_config call to direct assignment instead.

tanscorpio7 (dismissed):

contrib/babelfishpg_tsql/src/pl_handler.c
ayushdsh: > This issue only happens when there is a combinartion of set_config call and direct variable assignment.

Partially yes; but with direct assignment - in nested executions, we would still lose the dialect values. I will try to have an example added here.

tanscorpio7: I remeber seeing a similar issue in the past and we decided against changing the sql_dialect in fmgr.c i.e. inclined to wards direct assignment everywhere.
Could you add the code point where we are setting sql_dialect using set_config and it is creating this issue ?

ayushdsh: Yes. It is the `pltsql_truncate_identifier_func` in this particular case because col_length (plpgsql func) uses this function internally 

```
Datum
pltsql_truncate_identifier_func(PG_FUNCTION_ARGS)
{
	char	   *name = text_to_cstring(PG_GETARG_TEXT_PP(0));
	int			len = strlen(name);
	const char *saved_dialect = GetConfigOption("babelfishpg_tsql.sql_dialect", true, true);

	PG_TRY();
	{
		/* this is BBF help function. use BBF truncation logic */
		set_config_option("babelfishpg_tsql.sql_dialect", "tsql",
						  GUC_CONTEXT_CONFIG,
						  PGC_S_SESSION, GUC_ACTION_SAVE, true, 0, false);
		truncate_identifier(name, len, false);
	}
```

ayushdsh: > I remeber seeing a similar issue in the past and we decided against changing the sql_dialect in fmgr.c i.e. inclined to wards direct assignment everywhere.

I definitely agree with this being a dangerous change, however, don't you think this is the correct way to fix this. We should've ideally used PGs GUC setting infra instead of direct assignment because direct assignments will keep introducing new such issues as well ..

tanscorpio7: > We should've ideally used PGs GUC setting infra instead of direct assignment because direct assignments will keep introducing new such issues as well ..

We could have used either just not both together. We are two far into direct assignement for `sql_dialect` cannot go back now 😄 . Most hot code paths use it this way.

ayushdsh: > We could have used either just not both together.

Yes, exactly.

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -5848,27 +5848,21 @@ pltsql_truncate_identifier_func(PG_FUNCTION_ARGS)
...
-						  PGC_S_SESSION, GUC_ACTION_SAVE, true, 0, false);
+		sql_dialect = SQL_DIALECT_TSQL;
 		truncate_identifier(name, len, false);
 	}
 	PG_CATCH();
 	{
-		set_config_option("babelfishpg_tsql.sql_dialect", saved_dialect,
-						  GUC_CONTEXT_CONFIG,
-						  PGC_S_SESSION, GUC_ACTION_SAVE, true, 0, false);
+		sql_dialect = saved_dialect;
```
tanscorpio7, line 5861: 1. Let's change this to PG_TRY .. PG_FINALLY .. PG_END
2. Lets also fix this in `truncate_tsql_identifier`, `pltsql_text_name`, `pltsql_bpchar_name`, `pltsql_text_name(PG_FUNCTION_ARGS)`, `pltsql_bpchar_name(PG_FUNCTION_ARGS)`. These are all similar functions and will run into the same issue. Other occurences of `set_config('sql_dialect', ... ` can be analysed seperately.

tanscorpio7 (changes requested):

contrib/babelfishpg_tsql/src/pl_handler.c
ayushdsh: > Lets also fix this in truncate_tsql_identifier, pltsql_text_name, pltsql_bpchar_name, pltsql_text_name(PG_FUNCTION_ARGS), pltsql_bpchar_name(PG_FUNCTION_ARGS). These are all similar functions and will run into the same issue. 

Oh man. Missed this. Thanks for catching this!

> Let's change this to PG_TRY .. PG_FINALLY .. PG_END

Just curious semantically it won't matter right?

tanscorpio7: > Just curious semantically it won't matter right?

No, You can read more about it in these comments
```
* For the common case that the error recovery code and the cleanup in the
 * normal code path are identical, the following can be used instead:
 *
 *		PG_TRY();
 *		{
 *			... code that might throw ereport(ERROR) ...
 *		}
 *		PG_FINALLY();
 *		{
 *			... cleanup code ...
 *		}
 *      PG_END_TRY();
 *
 * The cleanup code will be run in either case, and any error will be rethrown
 * afterwards.
 *
 * You cannot use both PG_CATCH() and PG_FINALLY() in the same
 * PG_TRY()/PG_END_TRY() block.
 ```

Comment (ayushdsh) [reply contrib/babelfishpg_tsql/src/pl_handler.c:5861]: > You cannot use both PG_CATCH() and PG_FINALLY() in the same PG_TRY()/PG_END_TRY() block.

it makes sense that we might not need to .. but we "cannot" is interesting .. 

just a qq: won't these behave same

```
try
{
}
catch
{
    <do something>
    re_throw
}

<do the same thing as above>
```

```
try
{
}
finally
{
    <do something>
}
```

Comment (tanscorpio7) [reply contrib/babelfishpg_tsql/src/pl_handler.c:5861]: image

Comment (ayushdsh) [reply contrib/babelfishpg_tsql/src/pl_handler.c:5861]: Updated the PR. 
1. Using PG_TRY .. PG_FINALLY
2. updated the related functions with the same.

**Review by tanscorpio7 — APPROVED** [review]

File: contrib/babelfishpg_common/src/coerce.c [hunk 3]
```diff
@@ -251,22 +251,14 @@ pltsql_text_name(PG_FUNCTION_ARGS)
...
 				n = (*cstr_to_name_hook) (VARDATA_ANY(s), len);
 			}
-			PG_CATCH();
+			PG_FINALLY();
 			{
-				set_config_option("babelfishpg_tsql.sql_dialect", saved_dialect,
-								  GUC_CONTEXT_CONFIG,
-								  PGC_S_SESSION, GUC_ACTION_SAVE, true, 0, false);
-				PG_RE_THROW();
+				sql_dialect = saved_dialect;
```
surendravishnoi, line 259: why we have removed rethrow?

contrib/babelfishpg_common/src/coerce.c
```diff
@@ -251,22 +251,14 @@ pltsql_text_name(PG_FUNCTION_ARGS)
 			PG_TRY();
 			{
 				/* T-SQL casting. follow T-SQL truncation rule */
-				set_config_option("babelfishpg_tsql.sql_dialect", "tsql",
```
surendravishnoi, line 254: how do we manage risk here? new code can still add set config. Also, this in itself can introduce new issues.

contrib/babelfishpg_common/src/coerce.c
ayushdsh: we replaced PG_CATCH with PG_FINALLY, because we weren't doing anything with the error. the behaviour in this case would be the same.
we were doing this - 
```
try {}
catch{
    <do something>
    rethrow
}
<do the same thing>
```

we are replacing it by - 

```
try {}
finally{
    <do something>
}
```

if there's an error it gets thrown at PG_END_TRY()
```
#define PG_END_TRY(...)  \
		} \
		if (_do_rethrow##__VA_ARGS__) \
				PG_RE_THROW(); \
		PG_exception_stack = _save_exception_stack##__VA_ARGS__; \
		error_context_stack = _save_context_stack##__VA_ARGS__; \
	} while (0)
```

contrib/babelfishpg_common/src/coerce.c
ayushdsh: Discussed offline - I am writing all the functions we changed and verifying atleast that these functions don't have a set_config within their inner call stacks

ayushdsh: Went over the function changes, and turns out these are pretty primitive functions; identified no nested call chains doing more than computations (no set_config used)

surendravishnoi (approved):

PR #4438, opened 2026-01-15
fix #4434: DATEDIFF and DATEDIFF_BIG function support (weekday, dw, w) dateparts

### LINKED ISSUE #4434: [Bug]: Some dateparts are not allowed for function DATEDIFF
### What happened?

Some dateparts of funtion `DATEDIFF` are not allowed:
```sql
SELECT DATEDIFF(weekday, '2005-12-31 23:59:59.999', '2006-01-01 00:00:00.000');  -- error
SELECT DATEDIFF(dw, '2005-12-31 23:59:59.999', '2006-01-01 00:00:00.000');  -- error
SELECT DATEDIFF(w, '2005-12-31 23:59:59.999', '2006-01-01 00:00:00.000');  -- ok
```

### Version

BABEL_5_X_DEV (Default)

### Extension

babelfishpg_tsql (Default)

Ubuntu (Default)

### Relevant log output

```shell
syntax error at or near "weekday"
syntax error at or near "dw"
```

### Code of Conduct

Teletele-Lin:
This MR resolves the syntax error issue that occurred when specifying `weekday` or `dw` as the datepart parameter in the DATEDIFF and DATEDIFF_BIG functions.

* **Use case based -**
```sql
SELECT DATEDIFF(w, '2005-12-31 23:59:59.999999', '2006-01-01 00:00:00.000000');
SELECT DATEDIFF(weekday, '2005-12-31 23:59:59.999999', '2006-01-01 00:00:00.000000');
SELECT DATEDIFF(dw, '2005-12-31 23:59:59.999999', '2006-01-01 00:00:00.000000');

SELECT DATEDIFF_BIG(w, '2005-12-31 23:59:59.999999', '2006-01-01 00:00:00.000000');
SELECT DATEDIFF_BIG(weekday, '2005-12-31 23:59:59.999999', '2006-01-01 00:00:00.000000');
SELECT DATEDIFF_BIG(dw, '2005-12-31 23:59:59.999999', '2006-01-01 00:00:00.000000');
```

* **Boundary conditions -**

* **Arbitrary inputs -**

* **Negative test cases -**

* **Performance tests -**

* **Tooling impact -**

* **Client tests -**

### Check List


Teletele-Lin:
Due to data precision issues, the execution result of this SQL statement is inconsistent with SQL Server.
```sql
SELECT DATEDIFF_BIG(w, '2005-12-31 23:59:59.9999999', '2006-01-01 00:00:00.0000000'); 
```

contrib/babelfishpg_common/src/datetimeoffset.c
```diff
@@ -73,7 +73,13 @@ PG_FUNCTION_INFO_V1(varchar_datetimeoffset);
...
 PG_FUNCTION_INFO_V1(dateadd_datetimeoffset);
 
-#define DTK_NANO 32
+/*
+ * Token field definitions for time parsing and decoding.
+ *
+ * Note: Values 0 to 37 are already used in datetime.h.
+ * This token (DTK_NANO) is specific to T-SQL and is defined here.
+ */
+#define DTK_NANO 100
```
tanscorpio7: Can we skip this declaration since its already defined in datetime.h.

contrib/babelfishpg_tsql/src/backend_parser/gram-tsql-rule.y
```diff
@@ -4029,7 +4029,9 @@ datediff_arg:
 			| DAY_P									{ $$ = "day"; }
 			| TSQL_DD								{ $$ = "day"; }
 			| TSQL_D								{ $$ = "day"; }
-			| TSQL_W								{ $$ = "day"; }
+			| TSQL_W								{ $$ = "dow"; }
+			| TSQL_WEEKDAY							{ $$ = "dow"; }
+			| TSQL_DW								{ $$ = "dow"; }
```
tanscorpio7, line 4037: "dow" is used for returning the day of the week. https://www.postgresql.org/docs/8.1/functions-datetime.html#:~:text=dow

We should keep the mapping as day and only add the new keywords.

tanscorpio7 (changes requested):

contrib/babelfishpg_common/src/datetimeoffset.c
Teletele-Lin: `DTK_NANO` only used in datetime.c and datetimeoffset.c. The old value 32 is duplicated with `DTK_DOW`, so I think it should be changed because we will use `DTK_DOW` below.

contrib/babelfishpg_tsql/src/backend_parser/gram-tsql-rule.y
Teletele-Lin: weekday, dw and w  mean day of week, The original 'day' is inaccurate. It's a trick to implement datediff function.

contrib/babelfishpg_common/src/datetimeoffset.c
tanscorpio7: Can we define it in babelfish_extensions/contrib/babelfishpg_common/src/datetime.h ? We can avoid declaring the same MACRO twice.

The value 100 looks fine to me.

contrib/babelfishpg_tsql/src/backend_parser/gram-tsql-rule.y
tanscorpio7: Please share an example where date diff returns different results for day and weekday.

Also please add test cases to JDBC test framework.

Teletele-Lin: The result is the same, 'dow' more closely matches the actual meaning and remains consistent in the file. I'll add some cases later.

contrib/babelfishpg_tsql/src/backend_parser/gram-tsql-rule.y
```diff
@@ -4029,7 +4029,9 @@ datediff_arg:
 			| DAY_P									{ $$ = "day"; }
 			| TSQL_DD								{ $$ = "day"; }
 			| TSQL_D								{ $$ = "day"; }
-			| TSQL_W								{ $$ = "day"; }
+			| TSQL_W								{ $$ = "dow"; }
```
Deepesh125, line 4035: Same datediff_arg is also being used for `DATE_BUCKET`. How are we making sure that it doesn't show any unintended side effects? For example,  DATE_BUCKET would allow `w`, `weekday` and `dw` as a date part after your change whereas it should've thrown an error. Please verify it.

contrib/babelfishpg_tsql/src/backend_parser/gram-tsql-rule.y
Deepesh125: Looking at other diff in `timestamp_diff`, seems like end result would remain same before and after this change. So just curious -- why are we introducing new arg? And can't re-use "day" as it is being used before?

Deepesh125 (changes requested):

Teletele-Lin:
My JDBC test cases pass locally but fail on CI. How can I fix this?

tanscorpio7:
You may ignore the test in JDBC non default collation tests since your change is unrelated to collation. For example see: `echo "ignore#!#cast_binary_to_string-vu-prepare" >> jdbc_schedule`

contrib/babelfishpg_tsql/src/backend_parser/gram-tsql-rule.y
tanscorpio7: @Teletele-Lin

Teletele-Lin: > [@Teletele-Lin](https://github.com/Teletele-Lin)

Since TSQL_W means 'day of the week', mapping it to 'day' is inappropriate. Even though DTK_DOW and DTK_DAY are calculated in exactly the same way for the datediff function—which is why it was mapped to 'day' during implementation—changing this mapping itself is not an issue. This is because we have already made the necessary adaptations in the datetime.c file without re-implementing any functionality. Conversely, for the DATE_BUCKET function, its reuse of datediff_arg leads to a situation where an expected error fails to be thrown. Please refer to the reply above for details.

tanscorpio7:
https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/4635

PR #4478, opened 2026-01-22
BCP fix done token spanning packet boundaries

KushaalShroff:
### Description
Problem: When DONE token (13 bytes) was split across TDS packet boundaries, only the first byte (0xFD) would be checked, causing the remaining bytes in the next packet to be misinterpreted.
Actually we have always seen that done token gets truncated to fit in available space, I can see in some cases the done token is only 8 bytes at the end of the  message - packet marked with EOM. If we're at EOM with incomplete DONE token, that's actually acceptable.

Solution:
- Check if DONE token is present
- Only fetch additional bytes if NOT at EOM and buffer has < 13 bytes
- Accept incomplete DONE tokens at EOM (handles truncated tokens from optimizing clients)

### Check List


Deepesh125 (approved):

rishabhtanwar29 (approved):

jsudrik (approved):
Approved.

PR #4511, opened 2026-02-02
BABEL-2750 : ELEMENT support for FOR XML PATH 

JapleenKaur3:
## Description
Added support for ELEMENTS directive in FOR XML PATH clause.
JIRA : BABEL-2750

## Changes
- Added ELEMENTS XSINIL support for FOR XML PATH
- Added upgrade SQL for function signature changes
- Added backward compatibility for old 6-argument aggregate calls
- Added JDBC tests for PATH ELEMENTS scenarios

## Test Cases
- ELEMENTS XSINIL with NULL handling
- Various PATH scenarios with XSINIL
- Backward compatibility with existing views

## Related
- Similar to FOR XML RAW ELEMENTS support


contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--5.5.0--5.6.0.sql
```diff
@@ -59,6 +59,121 @@ $$
...
+END;
+$$;
+
+DO $$
+DECLARE
+    exception_message text;
+BEGIN
+    EXECUTE 'ALTER EXTENSION babelfishpg_tsql DROP AGGREGATE IF EXISTS sys.tsql_select_for_xml_agg_deprecated_in_5_6_0(anyelement, integer, text, boolean, text)';
+    DROP AGGREGATE IF EXISTS sys.tsql_select_for_xml_agg_deprecated_in_5_6_0(anyelement, integer, text, boolean, text);
+EXCEPTION WHEN OTHERS THEN
```
Deepesh125: please use babelfish_drop_deprecated_object, defined at the top of this script

contrib/babelfishpg_tsql/src/tsql_for/forxml.c
```diff
@@ -360,6 +379,25 @@ tsql_row_to_xml_path(StringInfo state, Datum record, const char *element_name, b
 				}
 			}
 		}
+		else if (xsinil)
+		{
+			/* XSINIL: Output NULL columns with xsi:nil="true" */
+			if (NameStr(att->attname)[0] != '@')
```
Deepesh125, line 443: what does this indicate?

Deepesh125 (changes requested):

contrib/babelfishpg_tsql/src/tsql_for/forxml.c
JapleenKaur3: The @ check ensures that we only apply xsi:nil="true" to element-centric columns. In FOR XML PATH, column names prefixed with @ are rendered as XML attributes. Since xsi:nil is only valid on XML elements and not on attributes, we skip attribute-centric columns when outputting NULL values in XSINIL mode.

contrib/babelfishpg_tsql/src/tsql_for/forxml.c
```diff
@@ -62,6 +62,8 @@ tsql_query_to_xml_sfunc(PG_FUNCTION_ARGS)
 		xsinil = PG_GETARG_BOOL(7);
 	}
 
+	(void)elements;
```
rohit01010: What is this for ?

contrib/babelfishpg_tsql/src/tsql_for/forxml.c
rohit01010: Let's add above details as a comment in the code.

test/JDBC/expected/forxml-path-elements-vu-verify.out
```diff
@@ -0,0 +1,1145 @@
...
+GO
+~~START~~
+ntext
+<Row/>
+~~END~~
+
+
+
+-- ============================================
+-- SECTION 9: Attribute-centric columns with ELEMENTS
```
rohit01010, line 319: Lets add more complex tests for this test case, such as follows
```
SELECT c.CustomerName AS "Customer/@Name", o.OrderID AS "Orders/@ID", o.TotalAmount AS "Order/@TotalAmount"
FROM forxml_path_elements_customers c
INNER JOIN forxml_path_elements_orders o ON c.CustomerID = o.CustomerID
ORDER BY o.OrderID
FOR XML PATH('Order'), ELEMENTS XSINIL;
GO
```

rohit01010: Use different JOINS, similar to tests added in `SECTION 18: Complex JOIN queries` with attribute centric columns

test/JDBC/upgrade/latest/schedule
```diff
@@ -677,4 +677,4 @@ babel_string_to_money_reg
 string_aggregate
 babel_convert_with_style
 forxml-raw-elements
-forxml-raw-elements-1gb-limit
```
rohit01010, line 681: Why have we removed `forxml-raw-elements-1gb-limit` ?

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--5.5.0--5.6.0.sql
JapleenKaur3: done

test/JDBC/upgrade/latest/schedule
JapleenKaur3: Fixed in the latest commit

contrib/babelfishpg_tsql/src/tsql_for/forxml.c
JapleenKaur3: added

contrib/babelfishpg_tsql/src/tsql_for/forxml.c
JapleenKaur3: It was initially added to suppress the compiler's "unused variable" warning, since FOR XML PATH doesn't use the elements flag. However, since elements is now being used in the tsql_row_to_xml_raw() call, the compiler will not give an unused variable warning. I will remove this line.

test/JDBC/expected/forxml-path-elements-vu-verify.out
JapleenKaur3: added more tests

contrib/babelfishpg_tsql/src/tsql_for/forxml.c
```diff
@@ -391,14 +400,20 @@ tsql_row_to_xml_path(StringInfo state, Datum record, const char *element_name, b
 			allnull = false;
 			if(NameStr(att->attname)[0] == '@')
 			{
-				appendStringInfo(state, "%s=\"%s\" ",
+				/* Add space before attribute (except for the very first content) */
+				if (state->len > 0 && state->data[state->len - 1] != ' ')
+					appendStringInfoChar(state, ' ');
+				appendStringInfo(state, "%s=\"%s\"",
 								 NameStr(att->attname)+1,
 								 map_sql_value_to_xml_value(colval, datatype_oid, true));
```
rohit01010: can we directly add space in appendStringInfo before attribute ? as follows ?
```
appendStringInfo(state, " %s=\"%s\"",
```
If not, can you please provide an example where this above suggested logic will fail.

contrib/babelfishpg_tsql/src/tsql_for/forxml.c
```diff
@@ -391,14 +400,20 @@ tsql_row_to_xml_path(StringInfo state, Datum record, const char *element_name, b
...
 			}
 			else
 			{
 				if (has_att_centric && first)
 				{
+					/* Remove trailing space if no attributes were added */
+					if (state->len > 0 && state->data[state->len - 1] == ' ')
+						state->len--;
```
rohit01010: can you please provide a test case, to understand why this is required ?

contrib/babelfishpg_tsql/src/tsql_for/forxml.c
JapleenKaur3: You are right. Looking at the code flow, the opening tag is always written before any @ attribute is processed, so the buffer always ends with either a letter or a closing quote at that point - never a space or empty. So directly adding space in the format string is safe. Updated.

contrib/babelfishpg_tsql/src/tsql_for/forxml.c
JapleenKaur3: After making your suggested change (space before attribute in format string), and the earlier change removing the trailing space from the xmlns line, the buffer never ends with a trailing space before > or /> is added. So the trim is no longer needed and I have removed it.

rohit01010 (approved):

rohit01010:
NIT: Can we change forxml-path-elements-before-17_9-or-18_3-* files to forxml-path-elements-before-17_10-or-18_4-* since we have bumped up the versions now?

Deepesh125 (approved):

PR #4577, opened 2026-02-15
Add hook to detect ENR to sys object dependencies and refactor find_object_in_enr hook

ayushdsh:
### Description

* Added a hook to detect if the dependency being created if from an ENR to a system object given the depender and referenced obejcts.
* Refactored the find_object_in_enr hook to also search in a given queryEnv apart from traversing all queryEnvs
* Added unit tests and sql tests for the ability to create ENR objects to non-ENR system objects dependencies

Engine PR - https://github.com/babelfish-for-postgresql/postgresql_modified_for_babelfish/pull/709

### Issues Resolved

Task: BABEL-6139

### Check List


contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -8316,15 +8324,19 @@ find_all_view_references(Node *node, List **view_oids)
...
-find_object_in_enr(Oid catalog_oid, Oid object_id)
+find_object_in_enr(Oid catalog_oid, Oid object_id, QueryEnvironment *qe)
 {
 	QueryEnvironment 		*queryEnv = currentQueryEnv;
 	ListCell         		*curlc;
 	EphemeralNamedRelation	enr;
 
+	if (qe)
+		queryEnv = qe;
+
```
tanscorpio7: Minor suggestion. Would `queryEnv = qe ? qe : currentQueryEnv;` improve readability ?

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -8421,10 +8433,51 @@ find_object_in_enr(Oid catalog_oid, Oid object_id)
 				break;
 		}
 		queryEnv = queryEnv->parentEnv;
+		if (qe)
+			break;
```
tanscorpio7, line 8436: One line comment `Do not recurse when caller provides the ENR`

tanscorpio7 (changes requested):

contrib/babelfishpg_tsql/src/guc.c
```diff
@@ -1266,6 +1266,20 @@ define_custom_variables(void)
 							 PGC_SUSET,
 							 GUC_NOT_IN_SAMPLE | GUC_NO_RESET_ALL | GUC_SUPERUSER_ONLY,
 							 NULL, NULL, NULL);
+
+	/* 
+	 * GUC to allow ENR to non-ENR dependency creation in pg_depend. In such cases, the entry goes into
+	 * the on-disk catalog instead of the in-memory ENR catalog. We should not be allowing this pattern,
+	 * and hence we will create a GUC which will by default disallow this pattern.
+	 */
+	DefineCustomBoolVariable("babelfishpg_tsql.allow_enr_to_non_enr_dependency",
```
surendravishnoi: why we need a GUC?

contrib/babelfishpg_tsql/src/hooks.c
surendravishnoi: what is the reasoning for this?

ayushdsh: so that we can use this function elsewhere as well. we have a usage which requires searching through all the queryEnvs (for ex. in lock.c, we try to see if we are locking an ENR object, for this we iterate over all envs); but there are cases where we need to only go through a given queryEnv (simply because the caller itself iterates over the queryEnvs) (for ex. recently added in find_associated_ENR fn in queryenvironment.c)

contrib/babelfishpg_tsql/src/guc.c
ayushdsh: Today, we support creation of dependency from an ENR object to a non-ENR non-system object. For ex.
```
create function p_test(@a int) returns int begin return @a * 100 end;
create table #t(a int check(p_test(a) > 100));
```
this add the pg_depend tuple in the on-disk catalog instead of ENR catalog. This pattern is incorrect because now we are leaking ENR things to on-disk (cleanup becomes difficult). 

Hence, now the change we've made disallows this. However, this means that for cx using this pattern is can be counted as a regression. Hence adding this GUC to allow this pattern in case cx faces an issue.

tanscorpio7 (dismissed):

tanscorpio7 (approved):

PR #4589, opened 2026-02-20
Fix to enable index scan with LIKE operator for non-const nodes

RuchaSK1:
### Description

Previously, LIKE operator index scan optimization only worked when the pattern was already a Const node. Expressions that could be simplified to a constant such as stable functions or type castswere left unreduced, forcing a sequential scan.
Additionally, view definitions after dump/restore lost the original column collation due to ::text casts, causing incorrect LIKE/ILIKE operator selection and skipping optimization entirely.

### Changes

Constant folding of RHS expressions : Apply eval_const_expressions() and evaluate_expr() to fold reducible pattern expressions down to a Const before prefix extraction. like_escape and remove_accents_internal are excluded to preserve ESCAPE clause and AI collation correctness.

Collation recovery after dump/restore : Unwrap RelabelType coercions on the left-hand side to recover the original column collation when ::text casts have stripped it.

NOT LIKE fix : Changed exact-match NOT LIKE optimization from incorrect AND to correct OR semantics: col <> pattern OR col NOT LIKE pattern.

Type reconciliation : Ensure the generated equality/range expressions always use the column's actual type rather than the LIKE-coerced TEXT type, so the planner can match the correct index operator class.

### Issues Resolved

Task: BABEL-5966

Authored-by: Rucha Kulkarni [ruchask@amazon.com](mailto:ruchask@amazon.com)
Signed-off-by: Rucha Kulkarni [ruchask@amazon.com](mailto:ruchask@amazon.com)

* **Use case based -** Yes

* **Arbitrary inputs -**

* **Negative test cases -**

* **Performance tests -**

* **Tooling impact -**

* **Client tests -**

### Check List


contrib/babelfishpg_tsql/src/collation.c
```diff
@@ -506,7 +550,7 @@ optimise_likenode(Node *node, OpExpr *op, like_ilike_info_t like_entry, coll_inf
 									InvalidOid,
```
kuntalghosh: [IMPORTANT] The logic for NOT LIKE optimization has been changed to use OR instead of AND, but lacks a comment explaining why this change matches SQL Server behavior.

Suggestion: Add a comment explaining the change:
```c
/* 
 * For NOT LIKE, we use OR to combine the equality check with the pattern check
 * to match SQL Server's behavior:
 * col NOT LIKE 'exact' => col <> 'exact' OR col NOT LIKE 'exact'
 */
ret = like_entry.is_not_match ? make_or_qual(ret, node) : make_and_qual(ret, node);
```

contrib/babelfishpg_tsql/src/collation.c
```diff
@@ -1053,6 +1097,43 @@ transform_likenode(Node *node, bool is_constraint)
```
kuntalghosh: [IMPORTANT] The view collation restoration code lacks error handling for invalid collations.

Suggestion: Add error handling for invalid collations:
```c
if (!OidIsValid(var_collation)) {
    ereport(DEBUG1,
            (errmsg("skipping collation restoration for view column with invalid collation")));
    return node;
}
```

kuntalghosh (commented):
Test comment - Ignore

contrib/babelfishpg_tsql/src/collation.c
```diff
@@ -401,7 +405,44 @@ optimise_likenode(Node *node, OpExpr *op, like_ilike_info_t like_entry, coll_inf
...
+	 * handles stable functions as a fallback.
+	 *
+	 * Peek through RelabelType wrappers to check for like_escape (ESCAPE
+	 * clause) and remove_accents_internal (AI mode). Evaluating these at
+	 * plan time loses escape and bracket pattern semantics.
+	 * For AI mode, the RelabelType unwrap below handles simple patterns.
+	 */
+
+	check_node = rightop;
+	while (IsA(check_node, RelabelType))
```
ahmed-shameem: Why do we need this? Is this needed for AI collations?

contrib/babelfishpg_tsql/src/collation.c
```diff
@@ -401,7 +405,44 @@ optimise_likenode(Node *node, OpExpr *op, like_ilike_info_t like_entry, coll_inf
...
+
+	if (!(IsA(check_node, FuncExpr) &&
+		  (strcmp(get_func_name(((FuncExpr *) check_node)->funcid),
+				 "like_escape") == 0 ||
+		   strcmp(get_func_name(((FuncExpr *) check_node)->funcid),
+				 "remove_accents_internal") == 0)))
+	{
+		rightop = eval_const_expressions(NULL, rightop);
+
+		if (!IsA(rightop, Const) && !IsA(rightop, Param) &&
```
ahmed-shameem: Why do we need these conditions?

contrib/babelfishpg_tsql/src/collation.c
```diff
@@ -466,9 +507,12 @@ optimise_likenode(Node *node, OpExpr *op, like_ilike_info_t like_entry, coll_inf
 	{
 		RelabelType	*relabel = (RelabelType *) leftop;
 		leftop = copyObject((Node*) relabel->arg);
-		prefix->consttype = rtypeId = ltypeId = exprType(leftop);
+		ltypeId = exprType(leftop);
```
ahmed-shameem, line 575: What is the necessity to make this change? Is there any test case which fails?

contrib/babelfishpg_tsql/src/collation.c
```diff
@@ -1053,6 +1097,43 @@ transform_likenode(Node *node, bool is_constraint)
...
+		/*
+		 * View definitions (after dump/restore) lose original column collation 
+		 * due to ::text casts, causing wrong LIKE/ILIKE operator selection. 
+		 * Unwrap coercions on leftop to find the original Var collation.
+		 * Skip if an explicit COLLATE clause is present.
+		 */
+
+		if (OidIsValid(like_entry.like_oid) && list_length(op->args) >= 2)
+		{
+					Node *left_arg = linitial(op->args);
```
ahmed-shameem: nit: indentation

contrib/babelfishpg_tsql/src/collation.c
```diff
@@ -1053,6 +1097,43 @@ transform_likenode(Node *node, bool is_constraint)
...
+					Node *right_arg = lsecond(op->args);
+			if (!IsA(left_arg, CollateExpr) && !IsA(right_arg, CollateExpr) && !contain_var_clause(right_arg))
```
ahmed-shameem, line 1174: Please explain this if condition

contrib/babelfishpg_tsql/src/collation.c
```diff
@@ -351,6 +355,7 @@ optimise_likenode(Node *node, OpExpr *op, like_ilike_info_t like_entry, coll_inf
 	Pattern_Prefix_Status pstatus;
 	int			collidx_of_cs_as;
 	CollateExpr *prefix_collate;
+	Node	   *check_node;
```
ahmed-shameem: On a very high level, I would rather prefer to call: simplify_function directly because the current implementation relies heavily on handling everything in our end. We can create a wrapper function over simplify_function (as this is static) and call that directly. Check: pattern_fixed_prefix_wrapper

test/JDBC/expected/like_op_index_scan-vu-prepare.out
```diff
@@ -0,0 +1,541 @@
+CREATE TABLE BABEL_5966_TestLikeIndex (
```
ahmed-shameem: This will pick db level collation. For AI collations to pick index, we need to create index from PG endpoint using the following:

```
create index <fully-qualified-table-name> (sys.remove_accents_internal(col_name));
```

check : test_like_index-vu-prepare

I would suggest not to have this table as in github we have 2 wfs. One for server default collation (CI_AS) and db level collation (CI_AI)

If you create the index from the TSQL endpoint only, CI_AS will pick index and NOT CI_AI, as evident from the expected output.

test/JDBC/expected/like_op_index_scan-vu-prepare.out
```diff
@@ -0,0 +1,541 @@
...
+);
+GO
+
+CREATE TABLE BABEL_5966_TestLikeCollation_CS_AI (
+    Id INT IDENTITY PRIMARY KEY,
+    Col VARCHAR(100) COLLATE SQL_Latin1_General_CP1_CS_AI
+);
+GO
+
+CREATE INDEX BABEL_5966_idx_ci_as ON BABEL_5966_TestLikeCollation_CI_AS(Col);
```
ahmed-shameem, line 25: Look at my comment on index creation for AI collations, these will never pick index scan

ahmed-shameem (changes requested):

contrib/babelfishpg_tsql/src/collation.c
RuchaSK1: Yes, needed for AI collations. When rightop is `RelabelType(remove_accents_internal(...))`, without the peek we can't detect remove_accents_internal inside. evaluate_expr would reduce it to a plain Const, losing AI semantics.

contrib/babelfishpg_tsql/src/collation.c
RuchaSK1: evaluate_expr actually executes the expression via PG executor. Without these checks, it crashes or produces wrong results for expressions that can't be evaluated at plan time.

contrib/babelfishpg_tsql/src/collation.c
RuchaSK1: Yes. Without this, BABEL_SCHEMATA-vu-verify crashes:
```
  use db_schemata_10;
  SELECT * FROM information_schema.schemata ORDER BY SCHEMA_NAME;
  TRAP: failed Assert("ltypeId == rtypeId")
```

rtypeId is set once at the top from original rightop. After evaluate_expr changes rightop to a Const, rtypeId becomes stale. Previously this code was never reached because rightop was never a Const for such queries. Now it is, and types mismatch.

contrib/babelfishpg_tsql/src/collation.c
RuchaSK1: These conditions check if it's safe to consider the column's collation for optimization:

1. `!IsA(left_arg, CollateExpr)`: If left has explicit COLLATE (eg Col COLLATE Latin1_General_CI_AI LIKE 'abc'), op->inputcollid already reflects it. Without this check, we would overwrite it with column's default collation, ignoring the explicit COLLATE.
2. `!IsA(right_arg, CollateExpr)`: If right has explicit COLLATE, evaluate_expr crashes on CollateExpr
3. `!contain_var_clause(right_arg)`: If pattern has column references (eg Col LIKE OtherCol), it cannot be reduced to a Const, so optimization won't apply.

contrib/babelfishpg_tsql/src/collation.c
RuchaSK1: Tested simplify_function approach. It only handles FuncExpr nodes. Rightop for basic queries like LIKE N'exact match' is: `RelabelType(CoerceToDomain(FuncExpr(...)))`

simplify_function cannot process RelabelType or CoerceToDomain. All test cases fall back to Seq Scan then.

We use eval_const_expressions + evaluate_expr because:
- eval_const_expressions handles immutable functions, OpExpr (e.g., string concatenation N'exact' + N' match'), nested type casts
- evaluate_expr handles stable functions (e.g., CONVERT, TRY_CONVERT) that eval_const_expressions leaves

test/JDBC/expected/like_op_index_scan-vu-prepare.out
RuchaSK1: Understood. Will remove BABEL_5966_TestLikeIndex and use only the explicit collation tables (CI_AS, CS_AS, CI_AI, CS_AI) which behave consistently across both workflows.

test/JDBC/expected/like_op_index_scan-vu-prepare.out
RuchaSK1: Agreed. Will remove AI indexes (idx_ci_ai, idx_cs_ai). Plain TSQL indexes can't match remove_accents_internal(col) used in AI queries.

contrib/babelfishpg_tsql/src/collation.c
ahmed-shameem: How do we make sure we are not leaving any other condition out? I mean handling this way can lead to cases which we are not handling explicitly. What happens then?

contrib/babelfishpg_tsql/src/collation.c
ahmed-shameem: It seems like we are putting some hack in our code, maybe in some cases that is unavoidable but in most of the conditions I see that if not done then this crashes. I am not comfortable with such explaination. Is there anyway to have a generic solution?

RuchaSK1: Yes, agreed. I've implemented a better approach now which removes the need for all the manual guards that solely arose from use of evaluate_expr. Replaced it with PG's built-in eval_const_expressions + estimate_expression_value - these handle safety checks internally. Any unhandled expression simply skips optimization with correct results.

The conditions in this block are unrelated to index optimization. They are for collation recovery needed because views lose column collation during storage due to ::text casts. They ensure we only recover when collation wasn't explicitly set and don't unnecessarily override collation. I think the test coverage for views with LIKE pattern was not much earlier and hence we couldn't catch this issue.

contrib/babelfishpg_tsql/src/collation.c
```diff
@@ -327,8 +372,9 @@ create_collate_expr(Node *arg, Oid collid)
  * function to remove the accents and optimize.
  * If the node is OpExpr and the collation is cs_as, then simply use optimization:
  *
- * Case 1: if the pattern is a constant stirng
- *		 col LIKE PATTERN -> col = PATTERN
+ * Case 1: if the pattern is a constant string
+ *		 col LIKE PATTERN -> col = PATTERN AND col LIKE PATTERN
+ *       col NOT LIKE PATTERN -> col <> PATTERN OR col NOT LIKE PATTERN
```
Deepesh125, line 417: With this OR clause, planner may loose an ability to choose index scan. Hence in most of the cases where we have correctness issue will be fixed with this but perf might get degraded. But I think tht should be okay

contrib/babelfishpg_tsql/src/collation.c
```diff
@@ -76,6 +79,48 @@ extern int	pattern_fixed_prefix_wrapper(Const *patt,
...
+
+	memset(&glob, 0, sizeof(PlannerGlobal));
+	glob.type = T_PlannerGlobal;
+	glob.boundParams = NULL;
+
+	memset(&root, 0, sizeof(PlannerInfo));
+	root.type = T_PlannerInfo;
+	root.glob = &glob;
+
+	return estimate_expression_value(&root, expr);
```
Deepesh125, line 161: comment over `fold_like_pattern_to_const` clearly says following, 
```
 * This function attempts to estimate the value of an expression for
 * planning purposes.  It is in essence a more aggressive version of
 * eval_const_expressions(): we will perform constant reductions that are
 * not necessarily 100% safe, but are reasonable for estimation purposes.
```

In essence -- runtime value might be different from planned value. So I think we should avoid using it blindly. Instead we should only try to const fold CONVERT/TRY_CONVERT selectively.

btw, what reason CONVERT/TRY_CONVERT function are marked STABLE in Babelfish?

Deepesh125 (changes requested):

contrib/babelfishpg_tsql/src/collation.c
RuchaSK1: Agreed. Replaced blind estimate_expression_value with a guarded approach. We now selectively fold only CONVERT/TRY_CONVERT, TRY_CAST, and string concatenation stable functions. Any other stable/volatile function skips folding.

These functions even though marked STABLE, have deterministic output for constant literals. I think they are marked STABLE likely because they perform internal catalog lookups and as per PG doc functions accessing tables should not be IMMUTABLE. Reference: https://www.postgresql.org/docs/current/xfunc-volatility.html#:~:text=It%20is%20generally%20unwise%20to%20select%20from%20database%20tables%20within%20an%20IMMUTABLE%20function%20at%20all

PR #4592, opened 2026-02-23
 [PART-1] Figure and Shape Parser support

Gopalverma062:
### Description

Fix geography validation bugs, replace ad-hoc CLR binary metadata parsing with spec-compliant implementation, and fix polygon Valid flag in properties byte.

**Currently**, Babelfish has the following issues in `spatialtypes.c`:

1. **Geography accepts out-of-range longitude values.** `geography::Point(45, 20000, 4326)` succeeds silently. T-SQL enforces longitude bounds of −15069 to +15069 degrees and rejects this input.

2. **Geometry/geography accepts Infinity coordinates.** Binary data containing `±Infinity` coordinate values passes all validation because only `isnan()` is checked, not `isinf()`. T-SQL rejects both NaN and Infinity.

3. **CLR binary metadata (Figure/Shape arrays) is parsed incorrectly.** The old `check_geom_end_metadata()` reads 4 bytes at a time across 5-byte figure entry boundaries. It works only because the attribute byte happens to produce the expected uint32 value when the following offset bytes are zeros. No bounds checking exists on metadata reads.

4. **Polygon properties byte missing Valid flag.** Babelfish outputs `0x00` for valid 2D polygons while T-SQL outputs `0x04` (V flag set). Linestrings already handled valid/invalid variants correctly.

5. **`set_dimension_flag()` prematurely guesses geometry type.** Properties byte `0x04` was hardcoded as `LINE_TYPE`, creating a latent bug where valid polygons from T-SQL would be misidentified.

**With this change:**

1. New `validate_longitude_range()` and `validate_not_inf_nan()` functions added and integrated into all validation paths (text input, binary input, direct point creation). Geography objects with longitude outside ±15069° are now rejected.

2. `check_nan_coordinates()` now checks both `isnan()` and `isinf()`. All geography validation paths also check for infinity.

3. New `parse_figures_and_shapes()` replaces `check_geom_end_metadata()`. Reads figure entries as 1+4 bytes and shape entries as 4+4+1 bytes per the T-SQL CLR spec (sections 2.1.3–2.1.4). Includes `CHECK_METADATA_BOUNDS` macro for buffer safety. Geometry type determined from Shape type byte — the single authoritative source.

4. Added `VALID_POLYGON_2D` through `VALID_POLYGON_3DM` constants. `determine_geom_dimensions()` now uses `is_valid` for polygons, matching the existing linestring pattern.

5. `set_dimension_flag()` no longer guesses `LINE_TYPE` for `0x04`. Falls through to `COMPLEX_GEOM_2D`, letting `parse_figures_and_shapes()` determine the actual type from Shape metadata.

6. `geography_from_bytea()` returns NULL for SRID = −1 input instead of crashing on length validation.

### Issues Resolved

- Geography silently accepts longitude values outside ±15069° range
- Geometry/geography silently accepts ±Infinity coordinates
- CLR Figure metadata parsed by reading 4 bytes across 5-byte entry boundaries (works by coincidence)
- CLR Shape metadata not parsed at all (geometry type guessed from byte patterns)
- Polygon properties byte missing Valid flag (byte 5 mismatch with T-SQL)
- `set_dimension_flag()` hardcodes `LINE_TYPE` for properties byte `0x04` (latent bug for valid polygons)
- NULL geography (SRID = −1) causes crash instead of returning NULL

* **Use case based -**
  - Point: 2D, 3DM, Empty (geometry)
  - LineString: 2-point, multi-point, Empty (geometry)
  - Polygon: invalid 2D, invalid 3D, invalid multi-ring, valid 2D, valid 3D, valid 3DM, self-intersecting, Empty (geometry)
  - Geography: Point, LineString, single-ring Polygon
  - Round-trip: geometry → varbinary → geometry for polygon

* **Boundary conditions -**
  - `geography::Point(45, 15069, 4326)` — longitude at positive boundary, must succeed
  - `geography::Point(45, -15069, 4326)` — longitude at negative boundary, must succeed
  - `geography::Point(90, 0, 4326)` — latitude at boundary, must succeed

* **Arbitrary inputs -**
  - 3-ring polygon (exterior + 2 interior rings)
  - Valid polygon (proper square) to verify V flag in properties byte
  - Self-intersecting polygon (invalid geometry)

* **Negative test cases -**
  - `geography::Point(45, 20000, 4326)` — longitude out of range, must error
  - `geography::Point(45, -20000, 4326)` — negative longitude out of range, must error
  - `geography::STGeomFromText('LINESTRING(0 0, 20000 45, 1 1)', 4326)` — longitude in linestring, must error
  - `geography::Point(91, 0, 4326)` — latitude out of range, must error
  - `CAST(0x00000000010C000000000000F07F0000000000004940 AS geometry)` — Infinity coordinate, must error
  - `CAST(0x00000000010C000000000000F87F0000000000004940 AS geometry)` — NaN coordinate, must error
  - `CAST(0xFFFFFFFF AS geography)` — SRID = −1, must return NULL

  N/A — no catalog or protocol changes

  N/A — no catalog or protocol changes

* **Performance tests -**
  N/A — validation adds negligible overhead (one extra isinf() call per coordinate, one memcpy for longitude extraction)

* **Tooling impact -**
  N/A — no changes to tooling

* **Client tests -**
  N/A — no client protocol changes

* **BABEL-6377**

### Check List
- [x] Commits are signed per the DCO using --gopalgv


contrib/babelfishpg_common/src/spatialtypes.c
```diff
@@ -87,6 +87,12 @@ static void load_functions();
 #define POLYGON_2DM         0x02  /* 2DM Polygon -  XYM dimensions*/
 #define POLYGON_3DM         0x03  /* 3DM Polygon - XYZM dimensions*/
```
rishabhtanwar29: I think we should prefix these with INVALID_

rishabhtanwar29:
Let's add jira task number in the description.

contrib/babelfishpg_common/src/spatialtypes.c
```diff
@@ -87,6 +87,12 @@ static void load_functions();
...
 
+/* Valid polygon variants — with V flag (bit 2 = 0x04) */
```
ahmed-shameem: What is this v flag?

contrib/babelfishpg_common/src/spatialtypes.c
```diff
@@ -127,13 +133,37 @@ static void load_functions();
...
+#define SHAPE_POLYGON            3
+#define SHAPE_MULTIPOINT         4
+#define SHAPE_MULTILINESTRING    5
+#define SHAPE_MULTIPOLYGON       6
+#define SHAPE_GEOMETRYCOLLECTION 7
+
+/* CLR binary metadata entry sizes */
+#define FIGURE_ENTRY_SIZE   5   /* 1 byte attribute + 4 bytes point_offset */
+#define SHAPE_ENTRY_SIZE    9   /* 4 bytes parent + 4 bytes figure_off + 1 byte type */
+#define COUNT_FIELD_SIZE    4   /* Size of NumFigures / NumShapes fields */
```
ahmed-shameem: Can this change later? Based on your comment, I assume that you calculated it using the given formula. If this value can change, lets' use the formula directly

contrib/babelfishpg_common/src/spatialtypes.c
```diff
@@ -1242,91 +1315,131 @@ set_dimension_flag(GeometryData *geom_data)
...
-        /* Complex geometry cases (linestring MP, polygon, and future geometry types) */
-        case COMPLEX_GEOM_2D:  /* 2D complex geometry (linestring MP, polygon, etc.) */
-            geom_data->dimension_flag = DIM_FLAG_2D; /* Has 2D Points (XY) */
-            geom_data->has_npoints_data = true;
+
+        /* 
+         *  Point — P flag (0x08) WITHOUT V flag.
+         * Props: 0x08=P, 0x09=P+Z, 0x0A=P+M, 0x0B=P+Z+M
+         */
+        case 0x08:  /* P only (no V) — 2D */
```
Mohitraj-g: Let's also create macros for 0x08, 0x09, 0x0A & 0x0B

contrib/babelfishpg_common/src/spatialtypes.c
```diff
@@ -1366,13 +1479,11 @@ check_nan_coordinates(GeometryData *geom_data)
...
     /* Check coordinates for NaN values */
-    for (i = 0; i < check_count; i++) 
+    for (i = 0; i < (int)check_count; i++) 
     {
         /* Copy the coordinate bytes to a double value */
         memcpy(&coord_value, geom_data->input_data + byte_position, COORD_SIZE);
-        
-        /* Use isnan() to check for NaN value */
-        if (isnan(coord_value)) 
+        if (isnan(coord_value) || isinf(coord_value))
```
Mohitraj-g, line 1495: It only sets geom_data->isNaN, shouldn't we also add geom_data->isinf ?

contrib/babelfishpg_common/src/spatialtypes.c
rishabhtanwar29, line 152: I think these all should be calculated instead, for example:
```
#define FIGURE_ENTRY_SIZE sizeof(Figure)
```

contrib/babelfishpg_common/src/spatialtypes.c
```diff
@@ -1495,122 +1641,176 @@ copy_coordinates_with_dimensions(uint8_t *src, uint8_t *dst, uint32_t npoints, u
...
+    CHECK_METADATA_BOUNDS(geom_data, metadata, offset, COUNT_FIELD_SIZE);
+    memcpy(&geom_data->nfigures, metadata + offset, sizeof(uint32_t));
+    offset += COUNT_FIELD_SIZE;
+
+    if (geom_data->nfigures < 1 || geom_data->nfigures > (uint32_t)npoints)
+        THROW_VARBINARY_CONVERSION_ERROR();
+
+    CHECK_METADATA_BOUNDS(geom_data, metadata, offset, geom_data->nfigures * FIGURE_ENTRY_SIZE);
+
+    geom_data->figures = palloc(geom_data->nfigures * sizeof(Figure));
```
rishabhtanwar29: Should use `palloc0` consistently.

contrib/babelfishpg_common/src/spatialtypes.c
```diff
@@ -567,12 +615,17 @@ validate_geography_latitude(Datum geom_datum, bool is_flipped)
...
         lat = DatumGetFloat8(lwgeom_x_p(fcinfo_local));
         
-        /* Validate latitude is within -90 to 90 degrees range */
+        UpdateFunctionCallInfo(fcinfo_local, 1, flipped_geom);
+        lon = DatumGetFloat8(lwgeom_y_p(fcinfo_local));
+        
+        validate_not_inf_nan(lat, "Latitude");
+        validate_not_inf_nan(lon, "Longitude");
         validate_latitude_range(lat);
+        validate_longitude_range(lon);
```
rishabhtanwar29: We have these validations at multiple places so we may refactor them to a separate help function.

contrib/babelfishpg_common/src/spatialtypes.c
```diff
@@ -1828,17 +2038,54 @@ create_empty_geometry(GeometryData *geom_data)
...
+         *   npoints(4)        = 0x00000000
+         *   nfigures(4)       = 0x00000000
+         *   nshapes(4)        = 0x01000000
+         *   shape parent(4)   = 0xFFFFFFFF
+         *   shape fig_off(4)  = 0xFFFFFFFF
+         *   shape type(1)     = geometry type byte
+         *
+         * Shape type is at: HEADER_SIZE + 4 + 4 + 4 + 4 + 4 = HEADER_SIZE + 20
+         */
+        uint32_t shape_type_offset = HEADER_SIZE
```
rishabhtanwar29: Can be a macro instead.

contrib/babelfishpg_common/src/spatialtypes.c
Gopalverma062: I looked into this, but using sizeof(Figure) / sizeof(Shape) would actually give incorrect results here.These macros represent the CLR wire format (binary serialization layout), not the in-memory struct size. Due to struct padding/alignment, the compiler will typically add extra bytes

contrib/babelfishpg_common/src/spatialtypes.c
Gopalverma062: Instead of adding a separate isinf flag, I've renamed isNaN to has_invalid_coords

Mohitraj-g (approved):

rishabhtanwar29 (approved):

PR #4632, opened 2026-03-06
Support BCP on TSQL #temp table

Ayush-061:
### Description

Through this changes now users will be able to perform BCP on #temptable through dotnet's `SqlBulkCopy`.
1. Currently BCP is not Supported for #temptable because the existing procedure `sp_tablecollations_100 `does not route for temp table.
2. With the modification in `spt_tablecollations_view` BCP is now supported for #temptable
3. Also added function `babelfish_get_enr_temp_table_attributes `to  get the attributes for both ENR temp tables.

### Issues Resolved

BABEL-5264

Authored by - ayushhx@amazon.com
Signed-off by-  ayushhx@amazon.com

* **Use case based -**

* **Boundary conditions -**

* **Arbitrary inputs -**

* **Negative test cases -**

* **Performance tests -**

* **Tooling impact -**

* **Client tests -**

### Check List


contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--5.5.0--5.6.0.sql
```diff
@@ -343,6 +343,41 @@ $$;
...
+-- This procedure enables SqlBulkCopy to work with temp tables by providing
+-- column collation metadata that BCP needs to encode string data correctly.
+CREATE OR REPLACE PROCEDURE sys.create_sp_tablecollations_100_in_tempdb_dbo()
+LANGUAGE C
+AS 'babelfishpg_tsql', 'create_sp_tablecollations_100_in_tempdb_dbo_internal';
+
+CALL sys.create_sp_tablecollations_100_in_tempdb_dbo();
+ALTER PROCEDURE tempdb_dbo.sp_tablecollations_100 OWNER TO sysadmin;
+GRANT EXECUTE ON PROCEDURE tempdb_dbo.sp_tablecollations_100 TO PUBLIC;
+DROP PROCEDURE sys.create_sp_tablecollations_100_in_tempdb_dbo;
```
ayushdsh: upgrades are actually simple because we know tempdb_dbos schema will always exists. so we don't have to do this. we will instead have to create the `tempdb_dbo.sp_tablecollations_100` procedure here like a normal procedure

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -6094,3 +6096,136 @@ openxml_simple(PG_FUNCTION_ARGS)
 #endif							/* USE_LIBXML */
 }
 
+/*
```
tanscorpio7: I thinks its about time babelfish temp table's code have a seperate file in extension : )

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -6094,3 +6096,136 @@ openxml_simple(PG_FUNCTION_ARGS)
...
+ * get_enr_attributes - Get pg_attribute rows for temp tables (ENR and non-ENR)
+ *
+ * This function returns all pg_attribute columns for a temp table.
+ * It works for both ENR temp tables (reads from ENR cache) and non-ENR temp tables
+ * (reads from actual pg_attribute catalog).
+ */
+Datum
+get_enr_attributes(PG_FUNCTION_ARGS)
```
ayushdsh: can you please check if the function - `ENRMetadataGetTupDesc` will simplify your job here.

contrib/babelfishpg_tsql/runtime/functions.c
ayushdsh: I couldn't agree more!

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--5.5.0--5.6.0.sql
```diff
@@ -360,6 +360,65 @@ END;
...
+-- Modify sys.sp_tablecollations_100 to handle temp tables (starting with #)
+-- This enables SqlBulkCopy to work with temp tables by providing column collation metadata.
+CREATE OR REPLACE PROCEDURE sys.sp_tablecollations_100
+(
+    IN "@object" nvarchar(4000)
+)
+AS $$
+BEGIN
+    -- Check if this is a temp table (starts with # or [# or contains .# or .[#)
+    IF LEFT(@object, 1) = '#' OR LEFT(@object, 2) = '[#' OR @object LIKE '%.#%' OR @object LIKE '%.[[]#%'
```
ayushdsh: wow. that's actually a lot of regex with which I am sure we are bound to miss something ..

contrib/babelfishpg_tsql/runtime/functions.c
Ayush-061: `cattups[ENR_CATTUP_ATTRIBUTE]`	already has HeapTuples ready, but with `ENRMetadataGetTupDesc` we would need to manually build HeapTuples from `Form_pg_attribute`.  Both will work .

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--5.5.0--5.6.0.sql
Ayush-061: I found one check with this `IF sys.SUBSTRING(@tab,1,1) <> '#'` ,  I think this won't work for our case. 
ref- https://github.com/babelfish-for-postgresql/babelfish_extensions/blob/c63bc116ec9f69df9356451ab5526b815b4d08a1/contrib/babelfishpg_tsql/sql/babelfishpg_tsql.sql#L3338

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -6094,3 +6096,136 @@ openxml_simple(PG_FUNCTION_ARGS)
...
+	table_name = pstrdup(temp_name_ptr);
+	for (i = 0; table_name[i]; i++)
+	{
+		if (table_name[i] == ']')
+		{
+			table_name[i] = '\0';
+			break;
+		}
+		table_name[i] = tolower((unsigned char) table_name[i]);
+	}
```
ayushdsh: maybe use `object_id` to get the oid of the table and then use `get_ENR_withoid()`. That would help skip these checks and string handling

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -6094,3 +6096,136 @@ openxml_simple(PG_FUNCTION_ARGS)
...
+	{
+		/* ENR path: return tuples directly from ENR cache */
+		ListCell   *lc;
+
+		foreach(lc, enr->md.cattups[ENR_CATTUP_ATTRIBUTE])
+		{
+			HeapTuple	tup = (HeapTuple) lfirst(lc);
+
+			tuplestore_puttuple(tupstore, tup);
+		}
```
ayushdsh: check if this can be simply replaced by the function `ENRMetadataGetTupDesc`

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--5.5.0--5.6.0.sql
ayushdsh: As discussed offline, we can simply replace this call with a C function doing the regex checks

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -6094,3 +6097,171 @@ openxml_simple(PG_FUNCTION_ARGS)
...
+	pfree(input);
+
+	/* Must be a temp table (starts with #) */
+	if (object_name[0] != '#')
+	{
+		pfree(db_name);
+		pfree(schema_name);
+		pfree(object_name);
+		PG_RETURN_NULL();
+	}
```
ayushdsh: haven't we already verified that this is a temp table .. before calling this function?

Ayush-061: This is if user calls the function babelfish_get_temp_table_attributes. for procedure the check is done by the other function is_temp_table_name

contrib/babelfishpg_tsql/runtime/functions.c
ayushdsh: @Ayush-061 lets start with this one. Could you instead create a new file (maybe called - `temp-table-hooks`) and write this function there instead. We'll move the rest slowly later.

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -6094,3 +6097,171 @@ openxml_simple(PG_FUNCTION_ARGS)
...
+
+	tuplestore_donestoring(tupstore);
+
+	PG_RETURN_NULL();
+}
+
+/*
+ * is_temp_table_name - Check if a name refers to a temp table
+ * Returns true if the object name part starts with #
+ */
```
ayushdsh: Nit: please add more comments mentioning that the name can be a four-part qualified name, within quoted-identifiers and even within square brackets qualifier.

ayushdsh (commented):
I am okay with almost all pieces. I just want one more pass on `get_tsql_temp_table_attributes` function which I will try to conclude today.

contrib/babelfishpg_tsql/runtime/functions.c
Ayush-061: Yeah, Fine

Ayush-061: Only the function get_tsql_temp_table_attribute or the other one is_temp_table name too?

ayushdsh: both of them

contrib/babelfishpg_tsql/sql/babelfishpg_tsql.sql
```diff
@@ -802,24 +802,69 @@ CREATE OR REPLACE VIEW sys.spt_tablecollations_view AS
         c.is_sparse = 0;
 GRANT SELECT ON sys.spt_tablecollations_view TO PUBLIC;
 
+-- Function to get pg_attribute rows for #temp tables (ENR and non-ENR)
+CREATE OR REPLACE FUNCTION sys.babelfish_get_temp_table_attributes(IN table_name sys.varchar(4000))
```
Deepesh125: for safer side, it should be sys.nvarchar(128)

contrib/babelfishpg_tsql/sql/babelfishpg_tsql.sql
```diff
@@ -802,24 +802,69 @@ CREATE OR REPLACE VIEW sys.spt_tablecollations_view AS
...
+RETURNS SETOF pg_catalog.pg_attribute
+AS 'babelfishpg_tsql', 'get_tsql_temp_table_attributes'
+LANGUAGE C STABLE PARALLEL UNSAFE;
+GRANT EXECUTE ON FUNCTION sys.babelfish_get_temp_table_attributes(IN sys.varchar(4000)) TO PUBLIC;
+
+-- Function to check if a name refers to a #temp table
+CREATE OR REPLACE FUNCTION sys.is_temp_table_name(IN name sys.varchar(4000))
```
Deepesh125: same as above, use sys.nvarchar

contrib/babelfishpg_tsql/src/procedures.c
```diff
@@ -5115,4 +5117,4 @@ tsql_openxml_get_colpattern(PG_FUNCTION_ARGS)
 	}
 	
 	PG_RETURN_TEXT_P(cstring_to_text(xpath_expr));
-}
\ No newline at end of file
+}
```
Deepesh125: revert file back to original

contrib/babelfishpg_tsql/runtime/temp_table_hooks.c
```diff
@@ -0,0 +1,197 @@
...
+		PG_RETURN_NULL();
+
+	input = text_to_cstring(PG_GETARG_VARCHAR_PP(0));
+
+	/* Parse the table name */
+	downcase_truncate_split_object_name(input, NULL, &db_name, &schema_name, &object_name);
+	pfree(input);
+
+	/* Must be a temp table (starts with #) */
+	if (object_name[0] != '#')
```
Deepesh125: object_name could be empty or null? any chance this could be called from PG dialect?

contrib/babelfishpg_tsql/runtime/temp_table_hooks.c
```diff
@@ -0,0 +1,197 @@
...
+	if (PG_ARGISNULL(0))
+		PG_RETURN_NULL();
+
+	input = text_to_cstring(PG_GETARG_VARCHAR_PP(0));
+
+	/* Parse the table name */
+	downcase_truncate_split_object_name(input, NULL, &db_name, &schema_name, &object_name);
+	pfree(input);
+
+	/* Must be a temp table (starts with #) */
```
Deepesh125: To avoid confusion, let's use tsql temp tables

contrib/babelfishpg_tsql/runtime/temp_table_hooks.c
```diff
@@ -0,0 +1,197 @@
...
+	if (object_name[0] != '#')
+	{
+		pfree(db_name);
+		pfree(schema_name);
+		pfree(object_name);
```
Deepesh125: validate pointers before pfree

contrib/babelfishpg_tsql/runtime/temp_table_hooks.c
```diff
@@ -0,0 +1,197 @@
...
+				(errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
+				 errmsg("materialize mode required, but it is not allowed in this context")));
+
+	if (PG_ARGISNULL(0))
+		PG_RETURN_NULL();
+
+	input = text_to_cstring(PG_GETARG_VARCHAR_PP(0));
+
+	/* Parse the table name */
+	downcase_truncate_split_object_name(input, NULL, &db_name, &schema_name, &object_name);
```
Deepesh125: why do we need db_name? we can pass NULL instead

contrib/babelfishpg_tsql/runtime/temp_table_hooks.c
```diff
@@ -0,0 +1,197 @@
...
+		pfree(db_name);
+		pfree(schema_name);
+		pfree(object_name);
+		PG_RETURN_NULL();
+	}
+
+	/* Try ENR first */
+	if (currentQueryEnv != NULL)
+	{
+		enr = get_ENR(currentQueryEnv, object_name, true);
```
Deepesh125: have we tested this part of code for following change of commands?

TSQL Dialect -> TSQL Proc -> calls get_tsql_temp_table_attributes 
TSQL Dialect -> TSQL Proc -> PG proc -> call get_tsql_temp_table_attributes
TSQL Dialect ->  PG proc -> call get_tsql_temp_table_attributes
TSQL Dialect ->  PG proc -> TSQL proc -> call get_tsql_temp_table_attributes


remember that TSQL temp table can be created at stage in between

contrib/babelfishpg_tsql/runtime/temp_table_hooks.c
```diff
@@ -0,0 +1,197 @@
...
+
+	pg_attribute_rel = table_open(AttributeRelationId, AccessShareLock);
+	tupdesc = CreateTupleDescCopy(RelationGetDescr(pg_attribute_rel));
+
+	tupstore = tuplestore_begin_heap(rsinfo->allowedModes & SFRM_Materialize_Random,
+									 false, work_mem);
+
+	rsinfo->returnMode = SFRM_Materialize;
+	rsinfo->setResult = tupstore;
+	rsinfo->setDesc = BlessTupleDesc(tupdesc);
```
Deepesh125: is there any reason why we are preferring this way of SRF instead of what PG Doc says, https://www.postgresql.org/docs/current/xfunc-c.html#XFUNC-C-RETURN-SET

contrib/babelfishpg_tsql/runtime/temp_table_hooks.c
```diff
@@ -0,0 +1,197 @@
...
+	if (is_enr)
+	{
+		/* ENR path: return tuples directly from ENR cache */
+		ListCell   *lc;
+
+		foreach(lc, enr->md.cattups[ENR_CATTUP_ATTRIBUTE])
+		{
+			HeapTuple	tup = (HeapTuple) lfirst(lc);
+
+			tuplestore_puttuple(tupstore, tup);
```
Deepesh125: I would suggest to make a copy before puttuple

contrib/babelfishpg_tsql/runtime/temp_table_hooks.c
```diff
@@ -0,0 +1,197 @@
...
+
+	input = text_to_cstring(PG_GETARG_VARCHAR_PP(0));
+
+	/* Parse the name to extract object_name */
+	downcase_truncate_split_object_name(input, NULL, &db_name, &schema_name, &object_name);
+	pfree(input);
+
+	/* Check if object_name starts with # */
+	if (object_name[0] == '#')
+		result = true;
```
Deepesh125: Is it fine to make such decision just base on name? we should look into currentQueryEnv. 

There could be issue with interoperability here since name with `#` in PG is valid and can be regular table. And sys.sp_tablecollations_100 could be called for those PG relations, right? how are we making sure that it works fine under those scenarios?

Deepesh125 (changes requested):
The approach is fine for me. But we need to prove that it won't cause any interoperability issue or unexpected behaviour.

contrib/babelfishpg_tsql/runtime/temp_table_hooks.c
ayushdsh: I think we should have a dialect check in order to make sure we cannot. This function does not make sense when called from PG dialect anyhow (from a nested plpgsql proc or directly)

contrib/babelfishpg_tsql/runtime/temp_table_hooks.c
```diff
@@ -0,0 +1,197 @@
...
+	if (object_name[0] != '#')
+	{
+		pfree(db_name);
+		pfree(schema_name);
+		pfree(object_name);
+		PG_RETURN_NULL();
+	}
+
+	/* Try ENR first */
+	if (currentQueryEnv != NULL)
```
ayushdsh: probably should be an assert. queryEnv is created at time of connection - backend spawning.

contrib/babelfishpg_tsql/runtime/temp_table_hooks.c
ayushdsh: we can replace initialization with `InitMaterializedSRF`

kuntalghosh (commented):
# LLM Code Review:

## PR Description Issues

**Issue 1 - Incomplete Test Scenarios Documentation**:

The PR description has empty sections for all test scenario categories. According to **testing rule 12**, you must include a 'Test scenarios covered' section with subsections: use-case based, boundary conditions, arbitrary inputs, negative tests, minor/major version upgrade tests, performance tests, tooling impact, and client tests (mark non-applicable sections as N/A).

**Issue 2 - Missing DCO Sign-off**:

According to **coding_conventions rule 1**, commits must be signed per the DCO using --signoff. The checklist shows this is unchecked.

## Code Quality Issues

**Issue 3 - In `contrib/babelfishpg_tsql/runtime/temp_table_hooks.c` at lines 107-154**:

```c
pg_attribute_rel = table_open(AttributeRelationId, AccessShareLock);
tupdesc = CreateTupleDescCopy(RelationGetDescr(pg_attribute_rel));
// ... processing ...
systable_endscan(scan);
table_close(pg_attribute_rel, AccessShareLock);
```

Missing error handling for resource cleanup. According to **memory rule 5**: "ALWAYS call systable_endscan() and table_close() on every exit path including early returns and ereport(ERROR)." The code should use PG_TRY/PG_FINALLY to ensure cleanup on error paths, especially since this function could potentially error during ENR access or tuple processing.

Suggested fix:

```c
PG_TRY();
{
    pg_attribute_rel = table_open(AttributeRelationId, AccessShareLock);
    // ... existing logic ...
}
PG_FINALLY();
{
    if (pg_attribute_rel)
        table_close(pg_attribute_rel, AccessShareLock);
}
PG_END_TRY();
```

## Test Coverage Gaps:

**Issue 4 - Missing Boundary Tests**:

According to **testing rule 2**, you must include boundary tests that verify both acceptance of valid boundary values and rejection of out-of-range values. The tests should include:
1. Empty table name parameter
2. Maximum length table name (4000 characters)
3. Invalid characters in table names
4. Non-existent temp tables

**Issue 5 - Missing NULL Tests**:

According to **testing rule 4**, you must include NULL tests for every new function parameter. Missing tests for:
1. NULL table_name parameter in `babelfish_get_temp_table_attributes`
2. NULL name parameter in `is_temp_table_name`

**Issue 6 - Missing Transaction Tests**:

According to **testing rule 6**, you must include nesting and transactionality tests. Missing tests for:
1. BCP operations within explicit transactions
2. Rollback behavior during BCP to temp tables
3. Nested procedure calls using temp table BCP

**Issue 7 - Missing Permission Tests**:

According to **testing rule 15**, you must test with different permission levels. Missing tests for:
1. Regular user (non-sysadmin) BCP to temp tables
2. Cross-database temp table access permissions

## Positive Aspects:

1. **Correct ENR Implementation**: The code properly handles both ENR and non-ENR temp tables as required by **temp_tables rule 14**. The fallback logic in lines 88-94 correctly tries pg_temp namespace when ENR lookup fails.

2. **Proper Memory Context Usage**: The function correctly switches to per_query_ctx for tuple store operations.

3. **Comprehensive Basic Test Coverage**: The JDBC tests cover both ENR and non-ENR temp table scenarios with various column types and collations.

4. **Upgrade Script Included**: Proper upgrade script provided for version 5.5.0 to 5.6.0.

5. **Correct NULL Handling**: Both C functions properly check for NULL arguments before processing.

#### Note: The LLMCodeReviewer is meant to augment, not replace, human review! Use the suggestions with discretion.

---
#BABEL-AI-REVIEW

contrib/babelfishpg_tsql/sql/babelfishpg_tsql.sql
Ayush-061: We cant simply use 128 as there can be 4 part names. which could possible exceed 128 chars , but yes will switch to nvarchar(4000)

contrib/babelfishpg_tsql/src/procedures.c
Ayush-061: reverted

contrib/babelfishpg_tsql/runtime/temp_table_hooks.c
Ayush-061: Added a dialect check .

contrib/babelfishpg_tsql/runtime/temp_table_hooks.c
Ayush-061: yeah agree , we can pass null for schema name as well

contrib/babelfishpg_tsql/runtime/temp_table_hooks.c
Ayush-061: Done

contrib/babelfishpg_tsql/runtime/temp_table_hooks.c
Ayush-061: Since now we are using object_id so there is no need for this check

contrib/babelfishpg_tsql/runtime/temp_table_hooks.c
```diff
@@ -0,0 +1,153 @@
...
+	ReturnSetInfo *rsinfo = (ReturnSetInfo *) fcinfo->resultinfo;
+	Relation	attrel;
+	Oid			relid = InvalidOid;
+
+	attrel = table_open(AttributeRelationId, AccessShareLock);
+
+	PG_TRY();
+	{
+		/* Setup SRF using pg_attribute's tuple descriptor */
+		rsinfo->expectedDesc = RelationGetDescr(attrel);
```
ayushdsh: Is this explicitly required?

contrib/babelfishpg_tsql/runtime/temp_table_hooks.c
```diff
@@ -0,0 +1,153 @@
...
+		InitMaterializedSRF(fcinfo, MAT_SRF_USE_EXPECTED_DESC | MAT_SRF_BLESS);
+
+		/*
+		 * Only process in T-SQL dialect or TDS connection. In PG dialect from
+		 * non-TDS connection, return empty result set.
+		 */
+		if (sql_dialect == SQL_DIALECT_TSQL ||
+			(is_bbf_tds_connection_hook && is_bbf_tds_connection_hook()))
```
ayushdsh: a simpler way of doing this is the opposite way - if the dialect is not tsql and conn is not tds_conn, simply return NULL.

and also, we need to determine that this is infact the intended behaviour; meaning - in case it is called from pg dialect but from tds conn, should we be allowing it?

contrib/babelfishpg_tsql/runtime/temp_table_hooks.c
```diff
@@ -0,0 +1,153 @@
...
+				InitFunctionCallInfoData(*locfcinfo, &flinfo, 2, InvalidOid, NULL, NULL);
+				locfcinfo->args[0].value = PG_GETARG_DATUM(0);
+				locfcinfo->args[0].isnull = false;
+				locfcinfo->args[1].value = (Datum) 0;
+				locfcinfo->args[1].isnull = true;
+
+				relid = DatumGetObjectId(FunctionCallInvoke(locfcinfo));
+				if (locfcinfo->isnull)
+					relid = InvalidOid;
+			}
```
ayushdsh: I don't understand this? Why can't we simply call the `object_id` function defined in `functions.c`

contrib/babelfishpg_tsql/runtime/temp_table_hooks.c
```diff
@@ -0,0 +1,153 @@
...
+						tuplestore_puttuple(rsinfo->setResult, tup);
+					systable_endscan(scan);
+				}
+				/* Regular tables: return empty result set */
+			}
+		}
+	}
+	PG_FINALLY();
+	{
+		table_close(attrel, AccessShareLock);
```
ayushdsh: no other tuplestore cleanup required?

ayushdsh (changes requested):
Missing tests for cross dialect calls to sp_tablecollations and your custom defined functions as well.

contrib/babelfishpg_tsql/runtime/temp_table_hooks.c
Ayush-061: yes when using `MAT_SRF_USE_EXPECTED_DESC` we need this explicitly
https://github.com/babelfish-for-postgresql/postgresql_modified_for_babelfish/blob/492fb8f567a64a0b271af73562451c31a342bba2/src/backend/utils/fmgr/funcapi.c#L69

contrib/babelfishpg_tsql/runtime/temp_table_hooks.c
Ayush-061: because both take different no. of arguments and we are infact calling object_id function line 73 and 83

contrib/babelfishpg_tsql/runtime/temp_table_hooks.c
Ayush-061: No I dont think so.. For scan we already did that in the if case.

contrib/babelfishpg_tsql/runtime/temp_table_hooks.c
ayushdsh: object_id(name, type) -> object_id(PG_GETARG_DATUM(0), "r")

ayushdsh (commented):
@tanscorpio7 mentioned to check for other SqlBulkCopy options, and if they call any other functions/proc similar to sp_tablecollations_100

contrib/babelfishpg_tsql/runtime/temp_table_hooks.c
Ayush-061: we cant use this  object_id(PG_GETARG_DATUM(0), "r")
https://github.com/babelfish-for-postgresql/babelfish_extensions/blob/BABEL_5_X_DEV/contrib/babelfishpg_tsql/runtime/functions.c#L2658

even if we want to use "u", since its a PG function not C so we will be using `DirectFunctionCall2Coll` which gives error when function return **NULL**, 
see here - https://github.com/babelfish-for-postgresql/postgresql_modified_for_babelfish/blob/BABEL_5_X_DEV__PG_17_X/src/backend/utils/fmgr/fmgr.c#L984
So for Non existing table we will get error message but the behavior for this should be to return empty result set

Ayush-061:
> Missing tests for cross dialect calls to sp_tablecollations and your custom defined functions as well.

Added test cases for these

contrib/babelfishpg_tsql/runtime/temp_table_hooks.c
```diff
@@ -0,0 +1,153 @@
...
+#include "parser/parser.h"
+#include "utils/builtins.h"
+#include "utils/fmgroids.h"
+#include "utils/lsyscache.h"
+#include "utils/queryenvironment.h"
+
+#include "../src/pltsql.h"
+#include "../src/multidb.h"
+
+extern Datum object_id(PG_FUNCTION_ARGS);
```
ayushdsh: ? why?

contrib/babelfishpg_tsql/runtime/temp_table_hooks.c
Ayush-061: Switched to the opposite way ie if the dialect is not tsql and conn is not tds_conn, we return NULL.

contrib/babelfishpg_tsql/runtime/temp_table_hooks.c
```diff
@@ -0,0 +1,155 @@
...
+		rsinfo->expectedDesc = RelationGetDescr(attrel);
+		InitMaterializedSRF(fcinfo, MAT_SRF_USE_EXPECTED_DESC | MAT_SRF_BLESS);
+
+		/*
+		 * Only process in T-SQL dialect or TDS connection. In PG dialect from
+		 * non-TDS connection, return empty result set.
+		 */
+		if (sql_dialect != SQL_DIALECT_TSQL &&
+			!(is_bbf_tds_connection_hook && is_bbf_tds_connection_hook()))
+			goto done;
```
ayushdsh: 1. avoid using goto as much as possible. check what happens in terms of performance and behaviour (hint: check how branch prediction works)
2. why do we want to delay this check? we should have this at the very start. and remove the conn check. if we are not in tsql dialect, simply return 0;

contrib/babelfishpg_tsql/runtime/temp_table_hooks.c
```diff
@@ -0,0 +1,155 @@
...
+
+		/* Resolve table name to OID via object_id() */
+		if (!PG_ARGISNULL(0))
```
ayushdsh: all such validations should be performed at the start of the function, and extract all the necessary args into local variables which can be then used.

contrib/babelfishpg_tsql/runtime/temp_table_hooks.c
```diff
@@ -0,0 +1,155 @@
...
+			LOCAL_FCINFO(locfcinfo, 2);
+			FmgrInfo	flinfo;
+
+			MemSet(&flinfo, 0, sizeof(flinfo));
+			flinfo.fn_addr = object_id;
+			flinfo.fn_nargs = 2;
+			flinfo.fn_mcxt = CurrentMemoryContext;
+
+			InitFunctionCallInfoData(*locfcinfo, &flinfo, 2, InvalidOid, NULL, NULL);
+			locfcinfo->args[0].value = PG_GETARG_DATUM(0);
```
ayushdsh: check above comment about - validation and extracting into local variables

test/JDBC/expected/BABEL-5264-vu-prepare.out
```diff
@@ -0,0 +1,50 @@
...
+-- Create PL/pgSQL procedure to test temp table functions from PG side
+CREATE OR REPLACE PROCEDURE test_pg_temp_table_proc()
+LANGUAGE plpgsql
+AS $$
+DECLARE
+    attr_count bigint;
+    coll_count bigint;
+BEGIN
+    -- Create a PG temp table
+    CREATE TEMP TABLE pg_proc_temp_test(id int, name varchar(50), description text);
```
ayushdsh: check with table name matching tsql convention - `"#pg_proc_temp_test"()`

contrib/babelfishpg_tsql/runtime/temp_table_hooks.c
```diff
@@ -0,0 +1,164 @@
...
+	ReturnSetInfo *rsinfo = (ReturnSetInfo *) fcinfo->resultinfo;
+	Relation	attrel;
+	Oid			relid = InvalidOid;
+	Datum		tablename;
+	bool		tablename_isnull;
+	FmgrInfo	flinfo;
+
+	/* Extract arguments into local variables */
+	tablename = PG_GETARG_DATUM(0);
+	tablename_isnull = PG_ARGISNULL(0);
```
ayushdsh: please make this inlined assignment

contrib/babelfishpg_tsql/runtime/temp_table_hooks.c
```diff
@@ -0,0 +1,164 @@
...
+	flinfo.fn_addr = object_id;
+	flinfo.fn_nargs = 2;
+	flinfo.fn_mcxt = CurrentMemoryContext;
+
+	/* Only process on TDS connections. Return empty result set otherwise. */
+	if (!(is_bbf_tds_connection_hook && is_bbf_tds_connection_hook()))
+	{
+		attrel = table_open(AttributeRelationId, AccessShareLock);
+		rsinfo->expectedDesc = RelationGetDescr(attrel);
+		InitMaterializedSRF(fcinfo, MAT_SRF_USE_EXPECTED_DESC | MAT_SRF_BLESS);
```
ayushdsh: I see a lot of duplication. tupdesc is anyway needed so why don't we move it outside before this.

contrib/babelfishpg_tsql/runtime/temp_table_hooks.c
```diff
@@ -0,0 +1,164 @@
...
+	tablename_isnull = PG_ARGISNULL(0);
+
+	/* Initialize FmgrInfo for object_id call */
+	MemSet(&flinfo, 0, sizeof(flinfo));
+	flinfo.fn_addr = object_id;
+	flinfo.fn_nargs = 2;
+	flinfo.fn_mcxt = CurrentMemoryContext;
+
+	/* Only process on TDS connections. Return empty result set otherwise. */
+	if (!(is_bbf_tds_connection_hook && is_bbf_tds_connection_hook()))
```
ayushdsh: hook will always be defined since we are in extension code .. instead - `is_bbf_connection()`

contrib/babelfishpg_tsql/runtime/temp_table_hooks.c
```diff
@@ -0,0 +1,161 @@
...
+#include "utils/builtins.h"
+#include "utils/fmgroids.h"
+#include "utils/lsyscache.h"
+#include "utils/queryenvironment.h"
+
+#include "../src/pltsql.h"
+#include "../src/multidb.h"
+#include "functions.h"
+
+#define OBJECT_TYPE	"u"
```
ayushdsh: nit: `OBJECT_TYPE_TABLE`

contrib/babelfishpg_tsql/runtime/temp_table_hooks.c
```diff
@@ -0,0 +1,161 @@
...
+	Datum		tablename = PG_GETARG_DATUM(0);
+
+	/* Setup SRF using pg_attribute's tuple descriptor */
+	attrel = table_open(AttributeRelationId, AccessShareLock);
+	rsinfo->expectedDesc = RelationGetDescr(attrel);
+	InitMaterializedSRF(fcinfo, MAT_SRF_USE_EXPECTED_DESC | MAT_SRF_BLESS);
+
+	/* Only process on TDS connections */
+	if (!is_bbf_tds_connection_hook())
+		elog(ERROR, "function is only supported on TDS connections");
```
ayushdsh: nit: can you check how we do this elsewhere as well? what kind of error do we show to the client

contrib/babelfishpg_tsql/runtime/temp_table_hooks.c
```diff
@@ -0,0 +1,161 @@
...
+	if (!PG_ARGISNULL(0))
+	{
+		PG_TRY();
+		{
+			relid = DatumGetObjectId(DirectFunctionCall2(object_id, tablename, CStringGetTextDatum(OBJECT_TYPE)));
+		}
+		PG_CATCH();
+		{
+			/* object_id threw an error - return empty result set */
+			FlushErrorState();
```
ayushdsh: this is a little dangerous. i think i will also want an opinion from other reviewers on this.

ayushdsh (commented):
Overall changes look fine to me. I am yet to review the tests. @tanscorpio7 @Deepesh125 I think you can now take a look at this PR once more.

contrib/babelfishpg_tsql/runtime/temp_table_hooks.c
tanscorpio7: This is not safe.
1. What's the motivation for this change?
2. Why not have OBJECT_ID as input argument for this function ? in SQL you can call OBEJCT_ID(N'table_name');

contrib/babelfishpg_tsql/runtime/functions.h
```diff
@@ -0,0 +1,16 @@
...
+ */
+#ifndef BABELFISHPG_TSQL_FUNCTIONS_H
+#define BABELFISHPG_TSQL_FUNCTIONS_H
+
+#include "fmgr.h"
+
+/* Object resolution functions */
+extern Datum object_id(PG_FUNCTION_ARGS);
+
+#endif							/* BABELFISHPG_TSQL_FUNCTIONS_H */
```
tanscorpio7: 1. What is the purpose of this file ? Isn't object_id already declared ?
2. These files should be in src folder.

contrib/babelfishpg_tsql/runtime/temp_table_hooks.c
tanscorpio7: We should call the function inside the hook directly. Hooks should only be called in engine code.

contrib/babelfishpg_tsql/runtime/temp_table_hooks.c
tanscorpio7: No need to close this in case of errors, PostgreSQL error handling cleans it up. Lets move this outside try finally block.

contrib/babelfishpg_tsql/runtime/temp_table_hooks.c
```diff
@@ -0,0 +1,161 @@
...
+
+/*
+ * is_temp_table_name
+ *		Returns true if the object name starts with #.
+ *
+ * Handles four-part qualified names with quoted identifiers or brackets.
+ * Only works on TDS connections, returns false otherwise.
+ */
+Datum
+is_temp_table_name(PG_FUNCTION_ARGS)
```
tanscorpio7: Can we instead do JOIN with sys.babelfish_get_enr_list() instead ?
SELECT * FROM sys.babelfish_get_enr_list() WHERE id = OBJECT_ID(N'table_name') ?

test/dotnet/utils/TestUtils.cs
```diff
@@ -703,5 +704,79 @@ public void FillSchemaTest(DbConnection conn, DbTransaction transaction, string
 				stCount--;
 			}
 		}
+
+		/* Test SqlBulkCopy to ENR temp table (standard temp table stored in currentQueryEnv) */
+		public bool SqlBulkCopyEnrTempTable(DbConnection bblCnn, string tempTableName, int rowCount,
+			string testName, Logger logger, ref int stCount)
```
tanscorpio7: Can we instead reuse the existing bulk copy method in dotnet test and only change the table name to a ENR and NON ENR temporary table.

contrib/babelfishpg_tsql/sql/babelfishpg_tsql.sql
```diff
@@ -802,24 +802,69 @@ CREATE OR REPLACE VIEW sys.spt_tablecollations_view AS
...
-        s_tcv.collation_100 AS collation
-    from
-        sys.spt_tablecollations_view s_tcv
-    where
-        s_tcv.object_id = (SELECT sys.object_id(@object))
-    order by colid;
+    -- Check if this is a #temp table 
+    IF sys.is_temp_table_name(@object) = 1
+    BEGIN
+        -- Use ENR function for temp tables
```
tanscorpio7: Can we instead extend the view `spt_tablecollations_view` instead ?

contrib/babelfishpg_tsql/runtime/temp_table_hooks.c
```diff
@@ -0,0 +1,161 @@
+/*-------------------------------------------------------------------------
+ *
+ * temp_table_hooks.c
```
tanscorpio7: Hooks suffix can be skipped since this file is not limited to temp table hooks.

contrib/babelfishpg_tsql/runtime/temp_table_hooks.c
ayushdsh: > Why not have OBJECT_ID as input argument for this function ? in SQL you can call OBEJCT_ID(N'table_name');

That was a miss from my side. This  should be done and would solve a lot of problems, without any downside of performance

contrib/babelfishpg_tsql/runtime/temp_table_hooks.c
ayushdsh: I think I agree. We should directly use the func.

contrib/babelfishpg_tsql/runtime/temp_table_hooks.c
ayushdsh: The idea was to make this with the least overhead. This is directly used in the sys proc for determining to use temp table func or view

contrib/babelfishpg_tsql/sql/babelfishpg_tsql.sql
ayushdsh: hmm ... worth checking. not very sure on any gain. @Ayush-061 we should discuss this.

tanscorpio7: The temp table check code can be skipped in that case.

contrib/babelfishpg_tsql/runtime/temp_table_hooks.c
```diff
@@ -0,0 +1,161 @@
...
+	if (PG_ARGISNULL(0))
+		PG_RETURN_BOOL(false);
+
+	/* Extract argument into local variable */
+	input = text_to_cstring(PG_GETARG_VARCHAR_PP(0));
+
+	downcase_truncate_split_object_name(input, NULL, NULL, NULL, &object_name);
+	pfree(input);
+
+	result = (object_name[0] == '#');
```
Deepesh125: there could be PG regular table name with '#' prefix.

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--5.5.0--5.6.0.sql
```diff
@@ -760,6 +760,73 @@ left join sys.shipped_objects_not_in_sys nis on nis.name = ('TT_' || tt.name ||
...
+        CAST(c.collname AS sys.nvarchar(128)) AS collation_name
+    FROM sys.babelfish_get_temp_table_attributes(table_name) a
+    LEFT JOIN pg_catalog.pg_collation c ON a.attcollation = c.oid
+    WHERE a.attnum > 0 AND NOT a.attisdropped
+    ORDER BY a.attnum;
+$$
+LANGUAGE SQL STABLE PARALLEL UNSAFE;
+GRANT EXECUTE ON FUNCTION sys.sp_tablecollations_100_enr(IN sys.nvarchar(4000)) TO PUBLIC;
+
+-- We are limited by what postgres procedures can return here, but IEW may not
```
Deepesh125: what is meaning of IEW?

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--5.5.0--5.6.0.sql
```diff
@@ -760,6 +760,73 @@ left join sys.shipped_objects_not_in_sys nis on nis.name = ('TT_' || tt.name ||
...
+            t.colid AS colid,
+            t.name AS name,
+            CAST(CollationProperty(t.collation_name, 'tdscollation') AS sys.binary(5)) AS tds_collation,
+            t.collation_name AS collation
+        FROM sys.sp_tablecollations_100_enr(@object) t
+        ORDER BY t.colid;
+    END
+    ELSE
+    BEGIN
+        -- Existing logic for regular tables
```
Deepesh125: once we merge this PR then this comment will be of no use for new readers.

contrib/babelfishpg_tsql/sql/babelfishpg_tsql.sql
```diff
@@ -802,24 +802,69 @@ CREATE OR REPLACE VIEW sys.spt_tablecollations_view AS
...
+-- Function to check if a name refers to a #temp table
+CREATE OR REPLACE FUNCTION sys.is_temp_table_name(IN name sys.nvarchar(4000))
+RETURNS BOOLEAN
+AS 'babelfishpg_tsql', 'is_temp_table_name'
+LANGUAGE C IMMUTABLE PARALLEL SAFE;
+GRANT EXECUTE ON FUNCTION sys.is_temp_table_name(IN sys.nvarchar(4000)) TO PUBLIC;
+
+-- Wrapper function for sp_tablecollations_100 that uses the babelfish_get_temp_table_attributes function
+CREATE OR REPLACE FUNCTION sys.sp_tablecollations_100_enr(IN table_name sys.nvarchar(4000))
+RETURNS TABLE(colid INT, name sys.varchar, collation_name sys.nvarchar(128))
```
Deepesh125: anywhere wherever we are touching/returning identifier then it should be should be sys.nvarchar(128).

contrib/babelfishpg_tsql/sql/babelfishpg_tsql.sql
```diff
@@ -802,24 +802,69 @@ CREATE OR REPLACE VIEW sys.spt_tablecollations_view AS
...
+
+-- Function to check if a name refers to a #temp table
+CREATE OR REPLACE FUNCTION sys.is_temp_table_name(IN name sys.nvarchar(4000))
+RETURNS BOOLEAN
+AS 'babelfishpg_tsql', 'is_temp_table_name'
+LANGUAGE C IMMUTABLE PARALLEL SAFE;
+GRANT EXECUTE ON FUNCTION sys.is_temp_table_name(IN sys.nvarchar(4000)) TO PUBLIC;
+
+-- Wrapper function for sp_tablecollations_100 that uses the babelfish_get_temp_table_attributes function
+CREATE OR REPLACE FUNCTION sys.sp_tablecollations_100_enr(IN table_name sys.nvarchar(4000))
```
Deepesh125: will this function able to handle truncated identifiers? Meaning temp table is created with name length > 64?

contrib/babelfishpg_tsql/runtime/temp_table_hooks.c
```diff
@@ -0,0 +1,161 @@
...
+	/* Only process on TDS connections */
+	if (!is_bbf_tds_connection_hook())
+		elog(ERROR, "function is only supported on TDS connections");
+
+	/* Resolve table name to OID via object_id() */
+	if (!PG_ARGISNULL(0))
+	{
+		PG_TRY();
+		{
+			relid = DatumGetObjectId(DirectFunctionCall2(object_id, tablename, CStringGetTextDatum(OBJECT_TYPE)));
```
Deepesh125: This would go through regular fmgr context switches, better if we can directly call object_id's C function

contrib/babelfishpg_tsql/runtime/temp_table_hooks.c
```diff
@@ -0,0 +1,161 @@
...
+			foreach(lc, enr->md.cattups[ENR_CATTUP_ATTRIBUTE])
+			{
+				HeapTuple	tup = (HeapTuple) lfirst(lc);
+				Form_pg_attribute att = (Form_pg_attribute) GETSTRUCT(tup);
+
+				if (att->attnum > 0)
+					tuplestore_puttuple(rsinfo->setResult, heap_copytuple(tup));
+			}
+		}
+		else if (isTempNamespace(get_rel_namespace(relid)))
```
Deepesh125: In PG, temp objects would be visible to all the sessions. So this would not work as we don't want to return other's temp objects.

```
-- session 1
postgres=> SELECT pg_backend_pid();
 pg_backend_pid
----------------
          24986
(1 row)

postgres=> create temp table tmp_tst(a int);
CREATE TABLE
postgres=> select oid from pg_class where relname = 'tmp_tst';
  oid
-------
 24598
(1 row)

-- session 2
postgres=> SELECT pg_backend_pid();
 pg_backend_pid
----------------
          25030
(1 row)

postgres=> select * from pg_attribute where attrelid = 24598;
postgres=> select count(*) from pg_attribute where attrelid = 24598;
 count
-------
     7
(1 row)
```

Deepesh125 (dismissed):

contrib/babelfishpg_tsql/runtime/temp_table_hooks.c
ayushdsh: Again, something that missed my review. This is incorrect. Will be rectified.

contrib/babelfishpg_tsql/runtime/temp_table_hooks.c
```diff
@@ -0,0 +1,161 @@
...
+		PG_RETURN_NULL();
+	}
+
+	PG_TRY();
+	{
+		EphemeralNamedRelation enr = GetENRTempTableWithOid(relid, false);
+
+		if (enr != NULL && enr->md.enrtype == ENR_TSQL_TEMP)
+		{
+			/* ENR path: use cached tuples, skip system attributes */
```
ayushdsh: what do you mean by "cached tuples"? this is a little confusing

contrib/babelfishpg_tsql/sql/babelfishpg_tsql.sql
ayushdsh: Today, temp tables do not handle truncation very well itself. 

```
1> create table #t234567891t234567891t234567891t234567891t234567891t234567891t23(a int);
2> go
Msg 2714, Level 16, State 1, Server manfred-test, Line 1
relation "#t234567891t234567891t234567891t234567891t234567891t234567891t2" already exists
```

but I believe the proc should not have a problem once we fix this issue with temp tables

contrib/babelfishpg_tsql/runtime/temp_table_hooks.c
```diff
@@ -0,0 +1,97 @@
...
+#include "miscadmin.h"
+#include "utils/aclchk_internal.h"
+#include "utils/builtins.h"
+#include "utils/fmgroids.h"
+#include "utils/lsyscache.h"
+#include "utils/queryenvironment.h"
+
+#include "../src/hooks.h"
+#include "../src/pltsql.h"
+#include "../src/multidb.h"
```
ayushdsh: nit: can this be absolute instead of relative

contrib/babelfishpg_tsql/sql/babelfishpg_tsql.sql
```diff
@@ -809,18 +836,17 @@ CREATE OR REPLACE PROCEDURE sys.sp_tablecollations_100
...
+        -- Existing logic for regular tables
+        SELECT
+            s_tcv.colid AS colid,
+            s_tcv.name AS name,
+            s_tcv.tds_collation_100 AS tds_collation,
+            s_tcv.collation_100 AS collation
+        FROM sys.spt_tablecollations_view s_tcv
+        WHERE s_tcv.object_id = (SELECT sys.object_id(@object))
+        ORDER BY colid;
+    END;
```
ayushdsh: nit: please fix this diff

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -138,7 +138,6 @@ typedef enum PltsqlInitPrivsOptions
 /*****************************************
  * 			General Hooks
  *****************************************/
-
```
ayushdsh, line 141: nit: fix diff

contrib/babelfishpg_tsql/sql/babelfishpg_tsql.sql
```diff
@@ -799,7 +808,25 @@ CREATE OR REPLACE VIEW sys.spt_tablecollations_view AS
         sys.all_columns c
         INNER JOIN pg_catalog.pg_class p ON (c.object_id = p.oid)
     WHERE
-        c.is_sparse = 0;
+        c.is_sparse = 0
+    UNION ALL
+    -- Temp tables from ENR and non-ENR sources
+    SELECT
+        CAST(a.attrelid AS int)          AS object_id,
+        CAST(pg_my_temp_schema() AS int) AS schema_id,
```
ayushdsh, line 815: nit: please add a comment on why this can be used - since all temp tables share the same temp namespace

contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--5.5.0--5.6.0.sql
```diff
@@ -760,6 +760,53 @@ left join sys.shipped_objects_not_in_sys nis on nis.name = ('TT_' || tt.name ||
...
+    WHERE
+        c.is_sparse = 0
+    UNION ALL
+    -- Temp tables from ENR and non-ENR sources
+    -- We use pg_my_temp_schema() because all temp tables share the same temp namespace
+    SELECT
+        CAST(a.attrelid AS int)          AS object_id,
+        CAST(pg_my_temp_schema() AS int) AS schema_id,
+        CAST(a.attnum AS int)            AS colid,
+        CAST(CAST(a.attname AS sys.sysname) AS sys.varchar) AS name,
```
ayushdsh: Nit: to check if this double casting can be avoided somehow

test/dotnet/input/sqlbulkcopy_temptable.txt
```diff
@@ -0,0 +1,49 @@
...
+
+# Test 4: Non-ENR Temp Table (using UDT)
+BEGIN TRY CREATE TYPE dbo.BcpTestVarcharType FROM VARCHAR(100) END TRY BEGIN CATCH END CATCH
+Create table #NonEnrTempTable4(id int, name dbo.BcpTestVarcharType)
+sqlbulkcopy_temptable#!##NonEnrTempTable4#!#5
+Select * from #NonEnrTempTable4 order by id
+drop table #NonEnrTempTable4
+
+# Test 5: Non-ENR Temp Table with more rows
+BEGIN TRY CREATE TYPE dbo.BcpTestVarcharType2 FROM VARCHAR(100) END TRY BEGIN CATCH END CATCH
```
ayushdsh: Nit: why a try..catch block?

test/dotnet/utils/TestUtils.cs
```diff
@@ -703,5 +704,42 @@ public void FillSchemaTest(DbConnection conn, DbTransaction transaction, string
 				stCount--;
 			}
 		}
+
+		/* Test SqlBulkCopy to temp table (both ENR and non-ENR) */
+		public bool SqlBulkCopyTempTable(DbConnection bblCnn, string tempTableName, int rowCount,
```
ayushdsh: blocker: I think handling the connections in the existing "insertbulkcopy" method will be easier .. I don't think we are using different connection within a particular test file, but even if we are, handling that would be easier.

test/JDBC/expected/BABEL-5264-vu-verify.out
```diff
@@ -0,0 +1,853 @@
...
+~~START~~
+varchar#!#smallint
+col1#!#1
+col2#!#2
+col3#!#3
+chunk_id#!#1
+chunk_seq#!#2
+chunk_data#!#3
+chunk_id#!#1
+chunk_seq#!#2
```
ayushdsh: blocker: we are not suppose to iterate over toast tables; toast tables are internal to the system and should not be exposed directly to the user

ayushdsh (commented):
blocker: 

I don't see tests for - 
1. default collations (i.e. without explicit collations)
2. other types of collations with nvarchar and varchars

You don't need to test the tempdb.dbo proc everytime as we know it will resolve to the same sys proc. just test it once on the very top and that should be fine.

contrib/babelfishpg_tsql/runtime/temp_table_functions.c
```diff
@@ -0,0 +1,99 @@
...
+	{
+		EphemeralNamedRelation enr = (EphemeralNamedRelation) lfirst(lc);
+		List *attList;
+		ListCell *attlc;
+
+		if (enr != NULL && enr->md.enrtype == ENR_TSQL_TEMP)
+		{
+			/* skipping toast tables */
+			if (OidIsValid(get_toast_parent_oid(enr->md.name)))
+				continue;
```
ayushdsh: blocker: this might break if cx creates table with name - `#pg_toast_<oid>` and the oid is actually valid. Instead get the relation from the relid and check the relkind

contrib/babelfishpg_tsql/runtime/temp_table_functions.c
```diff
@@ -0,0 +1,99 @@
...
+
+	InitMaterializedSRF(fcinfo, MAT_SRF_USE_EXPECTED_DESC | MAT_SRF_BLESS);
+
+	/* Only process on TDS connections */
+	if (!IS_TDS_CONN())
+		PG_RETURN_NULL();
+
+	attrel = table_open(AttributeRelationId, AccessShareLock);
+	enrList = get_namedRelList();
+	non_enr_tables = getRelationsInNamespace(GetTempNamespace(), RELKIND_RELATION);
```
ayushdsh: lets maintain consistency across naming - enrList and nonEnrList probably.

test/dotnet/src/BatchRun.cs
```diff
@@ -312,6 +312,49 @@ bool BatchRunner(DbConnection bblCnn, string queryFilePath, Serilog.Core.Logger
 								stCount--;
 							}
 						}
+						/* Case for SqlBulkCopy to temp table (both ENR and non-ENR) */
+						else if (strLine.ToLowerInvariant().StartsWith("sqlbulkcopy_temptable"))
```
ayushdsh: question: can't we use the `insertBulk` command here as well? why do we require a separate `sqlbulkcopy_temptable`?

ayushdsh (approved):
requesting @Deepesh125 @tanscorpio7 to take another look.

ayushdsh (commented):
Also, its worth adding some performance numbers for the view. Are we regressing anywhere and if so, by how much.

ayushdsh (approved):

tanscorpio7 (approved):

test/JDBC/input/BABEL-5264-vu-verify.mix
```diff
@@ -0,0 +1,487 @@
...
+
+EXEC dbo.babel5264_test_func_proc
+GO
+
+DROP TABLE #funcWrapperTest2
+GO
+
+-- Test PL/pgSQL procedure that creates temp table (non-ENR)
+EXEC dbo.psql_temp_proc
+GO
```
Deepesh125, line 410: just curious what would happen if we now create ENR temp table with the name '#temp1' and try to use sp_tablecollations_100 with the same?

ayushdsh: this results into a undefined behaviour because ENR code path is unaware of any objects in the non-ENR space. We have an issue open for this. Thanks for pointing this out Dipesh!

PR #4645, opened 2026-03-12
Fix STR() truncating concatenated strings

### LINKED ISSUE #4297: [Bug]: string appended after STR() gets lost
### What happened?

On SQL Server the output for 

`print 'distance:'+str(123,4,0)+' mi'
`

shows 

`distance: 123 mi
`

But on babelfish, for the same sentence, the output we get is: 

`distance: 123
`

The ' mi' at the end got lost. 

NOTE: the workaround for this issue is to replace STR() with CAST() or CONVERT()

### Version

BABEL_5_X_DEV (Default)

### Extension

None

_No response_

### Relevant log output

```shell

```

### Code of Conduct

zheliu2:
## Summary

Fixes #4297

The `STR()` function was passing `length + 1` (buffer size including null terminator) as the string length to `tsql_varchar_input()`, causing the resulting varchar to include an extra byte. This extra byte caused any strings concatenated after `STR()` to be silently lost.

**Root cause:** In `contrib/babelfishpg_tsql/src/string.c`, the `float_str()` function allocated `size = length + 1` bytes for the buffer (to accommodate the null terminator), but then passed `size` (instead of the actual string length) to `return_varchar_pointer()`.

**Fix:** Changed `return_varchar_pointer(buf, size)` to `return_varchar_pointer(buf, strlen(buf))` so the actual string length is used.

**Example:**
```sql
-- Before fix: returns 'distance: 123' (suffix lost)
-- After fix:  returns 'distance: 123 mi' (correct)
PRINT 'distance:' + STR(123, 4, 0) + ' mi'
```

## Test plan

- [x] Added test cases for string concatenation with STR() result
- [x] Updated expected output
- [ ] Verify existing STR() tests still pass
- [ ] Verify string concatenation with STR() now produces correct results

### DIFF
--- a/contrib/babelfishpg_tsql/src/string.c
+++ b/contrib/babelfishpg_tsql/src/string.c
@@ -801,7 +801,7 @@ float_str(PG_FUNCTION_ARGS)
 		}
 	}
 
-	return return_varchar_pointer(buf, size);
+	return return_varchar_pointer(buf, strlen(buf));
 }
 
 /*
--- a/test/JDBC/expected/str.out
+++ b/test/JDBC/expected/str.out
@@ -523,6 +523,32 @@ GO
 -- =============================================
 -- CLEANUP
 -- =============================================
+-- =============================================
+-- STRING CONCATENATION WITH STR() (GitHub Issue #4297)
+-- =============================================
+-- STR() result should not truncate strings appended after it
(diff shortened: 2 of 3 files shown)


zheliu2:
Closing to recreate with corrected commit metadata and branch naming.

PR #4648, opened 2026-03-12
Rewrite PATINDEX as native C function for deterministic collations

### LINKED ISSUE #4296: [Bug]: PATINDEX fails with '(' or ')' in search pattern
### What happened?

If the search pattern in PATINDEX contains an opening or closing parentheses sign, PATINDEX fails on babelfish (but it works on MSSQL)

```
set nocount on
select @@version
go
select patindex('%(%','the phone Nr is 1-(800)-CARS') [left_par]
go
select patindex('%)%','the phone Nr is 1-(800)-CARS') [right_par]
go

version
-------------------------------------------------------------------------
Babelfish for PostgreSQL with SQL Server Compatibility - 12.0.2000.8
Jul  6 2025 05:45:36
Copyright (c) Amazon Web Services
PostgreSQL 17.5 on aarch64-unknown-linux-gnu (Babelfish 5.2.0)

Msg 33557097, Level 16, State 1, Line 10
invalid regular expression: parentheses () not balanced
Msg 33557097, Level 16, State 1, Line 12
invalid regular expression: parentheses () not balanced

```

This works on MSSQL: 

```
set nocount on
select @@version
select patindex('%(%','the phone Nr is 1-(800)-CARS') [left_par]
select patindex('%)%','the phone Nr is 1-(800)-CARS') [right_par]

-------------------------------------------------------------------------------
Microsoft SQL Server 2022 (RTM-CU21) (KB5065865) - 16.0.4215.2 (X64) 
	Aug 11 2025 13:24:21 
	Copyright (C) 2022 Microsoft Corporation
	Developer Edition (64-bit) on Linux (Ubuntu 22.04.5 LTS) <X64>

left_par
-----------
19

right_par
-----------
23
```

### Version

BABEL_5_X_DEV (Default)

### Extension

None

_No response_

### Relevant log output

```shell

```

### Code of Conduct

zheliu2:
## Summary

Fixes #4296

Replaces the plpgsql `sys.PATINDEX` implementation (which relied on PostgreSQL's `substring()`/SIMILAR TO) with a native C function that directly implements T-SQL pattern matching.

### Problem

`PATINDEX` used `substring()` with SIMILAR TO syntax internally. Characters like `(`, `)`, `|`, `+`, `*`, `?`, `{`, `}` are metacharacters in SIMILAR TO but should be treated as literals in T-SQL PATINDEX patterns. The SQL-based approach required increasingly complex escaping logic and accumulated too many dependencies on other system functions.

### Solution

Added `tsql_patindex` C function in `string.c` that forks the pattern matching approach from `patindex_ai_match_text` (in `collation.c`) and adapts it for deterministic collations using direct byte comparison. The C function handles the full T-SQL pattern syntax natively:
- `%` — matches zero or more characters
- `_` — matches exactly one character
- `[abc]` — character class
- `[^abc]` — negated character class
- `[a-z]` — character range
- All other characters treated as **literals** (no escaping needed)

The SQL-level `sys.PATINDEX` now dispatches to `patindex_ai_collations` for AI collations and `sys.patindex_internal` (the new C function) for deterministic collations.

## Test plan

- Existing PATINDEX test suite passes (wildcards, character classes, ranges, Unicode, NULL handling, views, functions, procedures, computed columns, constraints)
- Added test cases for parentheses `(` `)`, `+`, `*`, `?`, `{` in patterns
- Added test for parentheses inside character classes `[()]`

### DIFF
--- a/contrib/babelfishpg_tsql/sql/sys_functions.sql
+++ b/contrib/babelfishpg_tsql/sql/sys_functions.sql
@@ -1760,51 +1760,21 @@ LANGUAGE C VOLATILE PARALLEL SAFE;
 
 create or replace function sys.PATINDEX(in pattern varchar, in expression varchar) returns bigint as
 $body$
-declare
-  v_find_result VARCHAR;
-  v_pos bigint;
-  v_regexp_pattern VARCHAR;
-  start_offset boolean;
-  end_offset boolean;
 begin
   if pattern is null or expression is null then
     return null;
   end if;
-  if pattern = '%' or pattern = '%%' then
-    return 1;
-  end if;
   if sys.is_collated_ai(expression) then
     return sys.patindex_ai_collations(pattern, expression);
   end if;
-  if PG_CATALOG.left(pattern, 1) = '%' collate sys.database_default then
-    v_regexp_pattern := regexp_replace(pattern, '^%', '%#"', 'i'::pg_catalog.TEXT);
-    start_offset := true;
-  else
-    v_regexp_pattern := '#"' || pattern;
-    start_offset := false;
-  end if;
-
-  if PG_CATALOG.right(pattern, 1) = '%' collate sys.database_default then
-    v_regexp_pattern := regexp_replace(v_regexp_pattern, '%$', '#"%', 'i'::pg_catalog.TEXT);
-    end_offset := true;
-  else
-   v_regexp_pattern := v_regexp_pattern || '#"';
-   end_offset := false;
-  end if;
-  v_find_result := substring(expression, v_regexp_pattern, '#');
-  if v_find_result <> '' collate sys.database_default then
-    if start_offset and not end_offset then
-      v_pos := LENGTH(expression) - STRPOS(REVERSE(expression), REVERSE(v_find_result)) + 2 - LENGTH(v_find_result);
-    else
-      v_pos := strpos(expression, v_find_result);
-    end if;
-  else
-    v_pos := 0;
-  end if;
-  return v_pos;
+  return sys.patindex_internal(pattern, expression);
 end;
 $body$
 language plpgsql immutable returns null on null input;
+
+CREATE OR REPLACE FUNCTION sys.patindex_internal(in pattern varchar, in expression varchar) returns bigint
+AS 'babelfishpg_tsql', 'tsql_patindex'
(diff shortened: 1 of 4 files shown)


contrib/babelfishpg_tsql/sql/sys_functions.sql
```diff
@@ -1776,15 +1780,39 @@ begin
...
     return sys.patindex_ai_collations(pattern, expression);
   end if;
-  if PG_CATALOG.left(pattern, 1) = '%' collate sys.database_default then
-    v_regexp_pattern := regexp_replace(pattern, '^%', '%#"', 'i'::pg_catalog.TEXT);
+
+  -- Escape SIMILAR TO metacharacters that are not T-SQL PATINDEX wildcards.
+  -- T-SQL PATINDEX only supports: % _ [...] [^...] [!...]
+  -- SIMILAR TO also treats ( ) | * + ? { } as special operators.
+  -- Since '#' is our escape character for substring(), we also escape literal '#'.
+  -- Characters inside [...] are already literals in SIMILAR TO, so skip those.
```
tanscorpio7: sys.patindex is accumulating too many dependencies on other system functions. At this point it would be simpler to fork PostgreSQL's MatchText directly in a C function & modify it to return the position where the pattern first matches.

tanscorpio7 (changes requested):

contrib/babelfishpg_tsql/sql/sys_functions.sql
zheliu2: thansk, that's a good suggestion, will try to rewrite it with native c

contrib/babelfishpg_tsql/sql/sys_functions.sql
```diff
@@ -1760,52 +1760,22 @@ LANGUAGE C VOLATILE PARALLEL SAFE;
 
 create or replace function sys.PATINDEX(in pattern varchar, in expression varchar) returns bigint as
 $body$
-declare
```
tanscorpio7, line 1763: @zheliu2 PostgreSQL allows changing a function's LANGUAGE via CREATE OR REPLACE even when the function is already referenced by a view.
So we could convert sys.patindex to a LANGUAGE C function and handle both AI and non-AI collation paths within it. This would avoid the per-character PL/pgSQL loop.

tanscorpio7 (changes requested):

PR #4662, opened 2026-03-19
Add MultiPoint datatype support 

Gopalverma062:
### Description

**Currently**, Babelfish supports `geometry` and `geography` spatial types for `Point`, `LineString`, and `Polygon` geometries. Attempting to use `MultiPoint` geometry type results in an *"unsupported geometry type"* error across all input/output paths including WKT text input, binary (`varbinary`) conversion, and spatial functions like `STGeomFromText` and `STMPointFromText`.

**With this change**, Babelfish now fully supports the **MultiPoint** geometry type for both `geometry` and `geography` data types. Users can create, store, retrieve, convert, and validate MultiPoint geometries through all existing spatial interfaces, matching T-SQLServer behavior.

**Why:** MultiPoint is a fundamental OGC geometry type used in real-world spatial applications to represent collections of discrete point locations (e.g., sensor networks, store locations, sample sites). Supporting it closes a significant gap in spatial type coverage and moves Babelfish closer to full T-SQL spatial compatibility.

**How the code was changed:**

1. **WKT Parser (grammar + rewrite functions):** Added grammar rules for parsing `MULTIPOINT((x y), ...)` and `MULTIPOINT Z/M/ZM(...)` syntax in both parenthesized and bare coordinate formats. Added `rewrite_multipoint_wkt()` and `rewrite_dim_multipoint_wkt()` functions for WKT normalization.

2. **T-SQL CLR Binary ↔ PostGIS WKB conversion (`spatialtypes.c`):**
   - **Bytea → PostGIS (T-SQL input):** Added CLR figure/shape metadata parsing for MultiPoint, validation of `STROKE` figures and `POINT` child shapes, and conversion from CLR columnar coordinate layout to PostGIS WKB MultiPoint format.
   - **PostGIS → Bytea (T-SQL output):** Added de-interleaving of PostGIS per-point WKB into CLR columnar format (XY block, Z block, M block), plus CLR metadata serialization (figures + shapes arrays).

3. **Geography validation:** Added MultiPoint coordinate validation for latitude/longitude range constraints using `ST_GeometryN` PostGIS accessor — consistent with the pattern used for other geometry types.

4. **SQL functions (`geometry.sql`):** Added `STMPointFromText` functions for both geometry and geography types with proper NULL handling, SRID validation, and type checking.

5. **Empty geometry handling:** Added support for empty MultiPoint in both CLR binary detection (props byte `0x04` with `EMPTY_COORD` pattern and `npoints=0` with Z/M flags) and PostGIS WKB output.

6. **Test coverage:** Added 12–13 MultiPoint test cases across geometry and geography ODBC tests covering 2D, Z, M, ZM, mixed dimensions, NULL coordinate stripping, single-point, empty, and different SRID scenarios.

---

### Issues Resolved

- MultiPoint geometry type support for `sys.geometry` and `sys.geography` data types
- `geometry::STGeomFromText('MULTIPOINT(...)')` now works
- `geography::STGeomFromText('MULTIPOINT(...)')` now works with latitude/longitude validation
- `geometry::STMPointFromText()` and `geography::STMPointFromText()` functions added
- Binary round-trip (`varbinary` ↔ `geometry`/`geography`) for MultiPoint
- `STAsText()` and `STAsBinary()` output for MultiPoint
- Empty MultiPoint (`MULTIPOINT EMPTY`) handling in all paths
- MultiPoint with Z, M, and ZM dimensions including mixed-dimension point collections

---

### Test Scenarios Covered

* **Use case based -**
  - Insert/Select MultiPoint via `STGeomFromText` with geometry and geography types
  - Insert/Select MultiPoint via `STMPointFromText` with type validation
  - Binary round-trip: `geometry` → `varbinary` → `geometry` for all dimension variants
  - WKT round-trip: text → geometry → text for MultiPoint
  - View creation and querying with MultiPoint columns
  - Update operations replacing other geometry types with MultiPoint and vice versa

* **Boundary conditions -**
  - Single-point MultiPoint: `MULTIPOINT((1 2))` — minimum valid collection
  - Empty MultiPoint: `MULTIPOINT EMPTY` — zero points
  - Large point count MultiPoint collections
  - Mixed dimensions: points with Z/M mixed with points without Z/M in same collection
  - NULL coordinate handling: `MULTIPOINT((1 2 NULL), (3 4 NULL))` stripped to 2D
  - NULL Z and NULL M: `MULTIPOINT((1 2 NULL NULL), (3 4 NULL NULL))` stripped to 2D

* **Arbitrary inputs -**
  - 2D MultiPoint: `MULTIPOINT((1 2), (3 4))`
  - 3D (Z) MultiPoint: `MULTIPOINT((1 2 3), (4 5 6))`
  - M MultiPoint: `MULTIPOINT((1 2 NULL 3), (4 5 NULL 6))`
  - ZM MultiPoint: `MULTIPOINT((1 2 3 4), (5 6 7 8))`
  - Mixed Z: `MULTIPOINT((1 2 3), (4 5))` — second point gets Z=NaN
  - Mixed M: `MULTIPOINT((1 2 NULL 3), (4 5))` — second point gets M=NaN
  - Mixed ZM: `MULTIPOINT((1 2 3 4), (5 6))` — second point gets Z=NaN, M=NaN
  - Bare coordinate format: `MULTIPOINT(1 2, 3 4)` — without inner parentheses
  - Different SRID: `MULTIPOINT((1 2), (3 4))` with SRID=0 (geometry only)

* **Negative test cases -**
  - `STMPointFromText` with non-MultiPoint input (e.g., `POINT`, `LINESTRING`) — expects error
  - `STMPointFromText` with NULL SRID — expects parameter error
  - `STMPointFromText` with negative SRID — expects error
  - Geography MultiPoint with latitude > 90 or < -90 — expects error
  - Geography MultiPoint with longitude > 15069 or < -15069 — expects error
  - Geography MultiPoint with NaN/Infinity coordinates — expects error
  - Invalid `varbinary` data conversion to MultiPoint — expects conversion error
  - Unsupported geometry type passed through MultiPoint path — expects error

  - Existing Point, LineString, and Polygon tests continue to pass unchanged
  - No changes to existing binary format for other geometry types

  - N/A — new feature addition, no schema migration required

* **Performance tests -**
  - MultiPoint with multiple points verified for correct CLR metadata serialization
  - No additional PostGIS function calls per point in binary conversion path (columnar batch processing)

* **Tooling impact -**
  - ODBC driver compatibility verified through `psqlodbc` test framework
  - No changes to ODBC driver required — uses existing UDT binary protocol

* **Client tests -**
  - ODBC insert, update, select, and view tests for both geometry and geography MultiPoint
  - Binary representation verified byte-by-byte against MSSQL Server output

---
### Check List
- [x] Commits are signed per the DCO using `--gopalgv@amazon.com`


contrib/babelfishpg_common/sql/geography.sql
```diff
@@ -228,6 +228,33 @@ CREATE OR REPLACE FUNCTION sys.Geography__stgeomfromtext(sys.NVARCHAR, integer)
...
+    AS $$
+    DECLARE
+        Geomtype text;
+        geog sys.GEOGRAPHY;
+    BEGIN
+        IF $2 IS NULL THEN
+            RAISE EXCEPTION '''geography::STMPointFromText'' failed because parameter 2 is not allowed to be null.';
+        ELSIF $1 IS NULL THEN
+            RETURN NULL;
+        ELSIF $2 < 0 THEN
```
Mohitraj-g: We don't need this check since we already have SRID validity check later on

contrib/babelfishpg_common/sql/geometry.sql
```diff
@@ -313,6 +313,32 @@ CREATE OR REPLACE FUNCTION sys.Geometry__STPolyFromText(sys.NVARCHAR,srid intege
...
+    AS $$
+    DECLARE
+        Geomtype text;
+        geom sys.GEOMETRY;
+    BEGIN
+        IF $2 IS NULL THEN
+            RAISE EXCEPTION '''geometry::STMPointFromText'' failed because parameter 2 is not allowed to be null.';
+        ELSIF $1 IS NULL THEN
+            RETURN NULL;
+        ELSIF $2 < 0 THEN
```
Mohitraj-g: Same as above

contrib/babelfishpg_common/src/geo_parser.y
```diff
@@ -230,6 +233,98 @@ coordinate:
...
+    | MPOINT_TOK M_TOK LPAREN ptarraym RPAREN
+        { $$ = rewrite_dim_multipoint_wkt($4); }
+    | MPOINT_TOK ZM_TOK LPAREN ptarrayzm RPAREN
+        { $$ = rewrite_dim_multipoint_wkt($4); } 
+    | MPOINT_TOK Z_TOK EMPTY_TOK
+        { $$ = pstrdup("MULTIPOINT EMPTY"); }
+    | MPOINT_TOK M_TOK EMPTY_TOK
+        { $$ = pstrdup("MULTIPOINT EMPTY"); }
+    | MPOINT_TOK ZM_TOK EMPTY_TOK
+        { $$ = pstrdup("MULTIPOINT EMPTY"); }
```
Mohitraj-g: Why do we need rules for these like : MULTIPOINT M EMPTY, TSQL doesn't allow it, does postgis create these ?

contrib/babelfishpg_common/src/spatialtypes.c
```diff
@@ -158,6 +155,9 @@ static void load_functions();
 #define INVALID_POINT_3DM_FLAG      0x0B    /* P+Z+M (no V) —  3DM Point */
 
 #define SHAPE_TYPE_OFFSET (HEADER_SIZE + NPOINTS_SIZE + COUNT_FIELD_SIZE + COUNT_FIELD_SIZE + sizeof(int32_t) + sizeof(uint32_t))
+#define WKB_POINT_TYPE            1
+#define WKB_LINESTRING_TYPE       2
+#define WKB_POLYGON_TYPE          3
```
Mohitraj-g: We can resuse already defined macros e.g.#define POINT_TYPE

contrib/babelfishpg_common/src/geo.c
```diff
@@ -667,3 +667,90 @@ rewrite_dim_polygon_query(PointArrayList *pal)
...
 }
+
+char* 
+rewrite_multipoint_wkt(PointArray *pa) 
+{
+    DimensionType type;
+    StringInfoData output;
+
+    if (!pa) 
+        return NULL;
```
rishabhtanwar29: should we return `pstrdup("MULTIPOINT EMPTY")` instead similar to rewrite_dim_multipoint_wkt?

contrib/babelfishpg_common/src/geo.c
```diff
@@ -667,3 +667,90 @@ rewrite_dim_polygon_query(PointArrayList *pal)
...
+
+    pfree(pa->points);
+    pfree(pa);
+    
+    return output.data;
+}
+
+
+char* 
+rewrite_dim_multipoint_wkt(PointArray *pa)
```
rishabhtanwar29, line 725: This and the above function share a lot of common code, can we extract the logic to helper function?

contrib/babelfishpg_common/src/spatialtypes.c
```diff
@@ -130,9 +126,10 @@ static void load_functions();
 #define GEOM_TYPE_POS_RESULT   4   /* Position of geometry type in result data */
 
 #define EMPTY_Binary_SIZE      9   /* Size of empty representation in binary */
-#define EMPTY_POINT_Binary   "\x01\x04\x00\x00\x00\x00\x00\x00\x00"  /* Binary for empty point */
-#define EMPTY_LINE_Binary    "\x01\x02\x00\x00\x00\x00\x00\x00\x00"  /* Binary for empty linestring */
-#define EMPTY_POLYGON_Binary "\x01\x03\x00\x00\x00\x00\x00\x00\x00"  /* Binary for empty polygon */
+#define EMPTY_POINT_Binary      "\x01\x04\x00\x00\x00\x00\x00\x00\x00"  /* Binary for empty point */
+#define EMPTY_LINE_Binary       "\x01\x02\x00\x00\x00\x00\x00\x00\x00"  /* Binary for empty linestring */
+#define EMPTY_POLYGON_Binary    "\x01\x03\x00\x00\x00\x00\x00\x00\x00"  /* Binary for empty polygon */
+#define EMPTY_MULTIPOINT_Binary "\x01\x04\x00\x00\x00\x00\x00\x00\x00"
```
rishabhtanwar29: We should rather use `_Bytes` instead of `_Binary`

contrib/babelfishpg_common/src/spatialtypes.c
rishabhtanwar29: Or better to convert this to an ENUM.

test/JDBC/input/datatypes/Test-Multipoint-vu-cleanup.txt
```diff
@@ -0,0 +1,82 @@
+# Geometry Views — Properties
```
rishabhtanwar29, line 1: Let's add this new test to lastest/schedule file

rishabhtanwar29:
Let's fix the PR title.

test/JDBC/input/datatypes/Test-Multipoint-vu-cleanup.txt
Gopalverma062: added

contrib/babelfishpg_common/src/geo.c
Gopalverma062: thanks! , i have made an helper function which have common code

contrib/babelfishpg_common/src/geo.c
```diff
@@ -389,6 +391,36 @@ format_postgis_point_coordinates(StringInfoData *output, POINT p)
...
+    if (!pa)
+        return;
+    if (pa->points)
+        pfree(pa->points);
+    pfree(pa);
+}
+
+/*  Replaces the repeated for-loop that formats points */
+static void
+append_formatted_points(StringInfoData *output, PointArray *pa, bool wrap_each, PointFormatter formatter)
```
rishabhtanwar29, line 407: nit: so far we are not actually making use of `wrap_each` parameter as all the callers currently setting it `true` so it seems unnecessary.

contrib/babelfishpg_common/src/spatialtypes.c
```diff
@@ -1937,6 +2108,191 @@ handle_point_coordinates(GeometryData *geom_data, uint8 *result_data)
...
+            return 0;
+    }
+}
+
+/*
+ * Calculate WKB size for child geometry
+ */
+
+static uint32_t
+calculate_child_wkb_size(uint8 parent_type, uint32_t child_npoints, GeometryData *geom_data, uint32_t fig_start, uint32_t fig_end, bool has_z, bool has_m)
```
rishabhtanwar29: nit: most of the arguments are unused, I think we should add them in future whenever needed.

contrib/babelfishpg_common/sql/upgrades/spatial_types--5.5.0--5.6.0.sql
```diff
@@ -149,3 +149,119 @@ CREATE OR REPLACE FUNCTION sys.STGeometryType(geog sys.GEOGRAPHY)
 	    RAISE EXCEPTION 'Unexpected geometry type format: %. Expected ST_* prefix.', geom_type;
 	END;
 	$$ LANGUAGE plpgsql IMMUTABLE STRICT PARALLEL SAFE;
+
+--geometry
```
ahmed-shameem: nit: give a space after -- . Applies for other places as well.

contrib/babelfishpg_common/sql/upgrades/spatial_types--5.5.0--5.6.0.sql
```diff
@@ -149,3 +149,119 @@ CREATE OR REPLACE FUNCTION sys.STGeometryType(geog sys.GEOGRAPHY)
...
+        Geomtype text;
+        geom sys.GEOMETRY;
+    BEGIN
+        IF $2 IS NULL THEN
+            RAISE EXCEPTION '''geometry::STMPointFromText'' failed because parameter 2 is not allowed to be null.';
+        ELSIF $1 IS NULL THEN
+            RETURN NULL;
+        END IF;
+
+        geom = sys.geomfromtext_helper($1::text, $2);
```
ahmed-shameem: Additional casting of $1  to text should not be needed

contrib/babelfishpg_common/sql/upgrades/spatial_types--5.5.0--5.6.0.sql
```diff
@@ -149,3 +149,119 @@ CREATE OR REPLACE FUNCTION sys.STGeometryType(geog sys.GEOGRAPHY)
...
+        Geomtype = sys.ST_GeometryType(geom);
+
+        IF Geomtype = 'ST_MultiPoint' THEN
```
ahmed-shameem, line 177: Are we sure this comparison will always pass? There won't be any issue with upper/lower case?

contrib/babelfishpg_common/sql/upgrades/spatial_types--5.5.0--5.6.0.sql
```diff
@@ -149,3 +149,119 @@ CREATE OR REPLACE FUNCTION sys.STGeometryType(geog sys.GEOGRAPHY)
...
+            RETURN geom;
+        ELSE
+            RAISE EXCEPTION 'Expected "MULTIPOINT" at Position 1. The input has %s', $1;
```
ahmed-shameem: nit: Position --> position. Is this the specific error message thrown in TSQL as well? Do we have test case for all the exceptions in the function?

contrib/babelfishpg_common/sql/upgrades/spatial_types--5.5.0--5.6.0.sql
```diff
@@ -149,3 +149,119 @@ CREATE OR REPLACE FUNCTION sys.STGeometryType(geog sys.GEOGRAPHY)
...
+
+CREATE OR REPLACE FUNCTION sys.Geometry__STMPointFromText(sys.NVARCHAR, srid integer)
+    RETURNS sys.GEOMETRY
+    AS $$
+    DECLARE
+        Geomtype text;
+        geom sys.GEOMETRY;
+    BEGIN
+        IF $2 IS NULL THEN
+            RAISE EXCEPTION '''geometry::STMPointFromText'' failed because parameter 2 is not allowed to be null.';
```
ahmed-shameem, line 169: qq: parameter 2 is not allowed of 2nd parameter can not be null? Are we using the same error message as TSQL?

contrib/babelfishpg_common/sql/upgrades/spatial_types--5.5.0--5.6.0.sql
```diff
@@ -149,3 +149,119 @@ CREATE OR REPLACE FUNCTION sys.STGeometryType(geog sys.GEOGRAPHY)
...
+
+        IF Geomtype = 'ST_MultiPoint' THEN
+            RETURN geom;
+        ELSE
+            RAISE EXCEPTION 'Expected "MULTIPOINT" at Position 1. The input has %s', $1;
+        END IF;
+    END;
+    $$ LANGUAGE plpgsql IMMUTABLE PARALLEL SAFE;
+
+--Geography
```
ahmed-shameem: Same comment applies as STMPointFromText

contrib/babelfishpg_common/sql/upgrades/spatial_types--5.5.0--5.6.0.sql
```diff
@@ -149,3 +149,119 @@ CREATE OR REPLACE FUNCTION sys.STGeometryType(geog sys.GEOGRAPHY)
...
+        IF Geomtype = 'ST_MultiPoint' THEN
+            RETURN geog;
+        ELSE
+            RAISE EXCEPTION 'Expected "MULTIPOINT" at Position 1. The input has %s', $1;
+        END IF;
+    END;
+    $$ LANGUAGE plpgsql IMMUTABLE PARALLEL SAFE;
+
+
+CREATE OR REPLACE FUNCTION sys.Geography__STMPointFromWKB(sys.VARBINARY, srid integer)
```
ahmed-shameem, line 215: nit: Put comment before the function as you have put for the above functions

contrib/babelfishpg_common/sql/upgrades/spatial_types--5.5.0--5.6.0.sql
```diff
@@ -149,3 +149,119 @@ CREATE OR REPLACE FUNCTION sys.STGeometryType(geog sys.GEOGRAPHY)
...
+        Geomtype text;
+        geog sys.GEOGRAPHY;
+    BEGIN
+        IF $2 IS NULL THEN
+            RAISE EXCEPTION '''geography::STMPointFromWKB'' failed because parameter 2 is not allowed to be null.';
+        ELSIF $1 IS NULL THEN
+            RETURN NULL;
+        END IF;
+
+        geog = sys.geogfromwkb_helper($1::bytea, $2);
```
ahmed-shameem, line 228: Isn't the case from varbinary --> bytea implicit? If so, this cast is redundant

contrib/babelfishpg_common/sql/upgrades/spatial_types--5.5.0--5.6.0.sql
ahmed-shameem: Same comment applies as STMPointFromText

contrib/babelfishpg_common/sql/upgrades/spatial_types--5.5.0--5.6.0.sql
```diff
@@ -149,3 +149,119 @@ CREATE OR REPLACE FUNCTION sys.STGeometryType(geog sys.GEOGRAPHY)
...
+    $$ LANGUAGE plpgsql IMMUTABLE PARALLEL SAFE;
+
+CREATE OR REPLACE FUNCTION sys.geogfromwkb_helper(bytea, integer)
+    RETURNS sys.GEOGRAPHY
+    AS 'babelfishpg_common', 'get_geography_from_wkb'
+    LANGUAGE 'c' IMMUTABLE PARALLEL SAFE;
+
+
+
+CREATE OR REPLACE FUNCTION sys.Geometry__STMPointFromWKB(sys.VARBINARY, srid integer)
```
ahmed-shameem, line 246: Same comments applies as STMPointFromWKB

contrib/babelfishpg_common/src/geo.c
```diff
@@ -389,6 +391,36 @@ format_postgis_point_coordinates(StringInfoData *output, POINT p)
     }
 }
 
+/*  pfree(pa->points); pfree(pa); pattern */
```
ahmed-shameem: nit: what's the point of this comment?

contrib/babelfishpg_common/src/geo.c
```diff
@@ -667,3 +699,62 @@ rewrite_dim_polygon_query(PointArrayList *pal)
...
+    if (!pa || pa->count == 0)
+    {
+        free_point_array(pa);
+        return pstrdup("MULTIPOINT EMPTY");
+    }
+
+    initStringInfo(&output);
+    
+    appendStringInfoString(&output, "MULTIPOINT(");
+
```
ahmed-shameem, line 738: Why don't we need to find appropriate dimension type here like rewrite_multipoint_wkt?

contrib/babelfishpg_common/src/geo_parser.y
```diff
@@ -230,6 +233,92 @@ coordinate:
         { $$ = create_point($1, $2, 0, $4, 0, 1); }
     ;
 
+multipoint_query:
```
ahmed-shameem, line 236: Please make sure we are covering all possible cases. Also, I hope that tests corresponding to each below cases has been added.

contrib/babelfishpg_common/src/spatialtypes.c
```diff
@@ -21,17 +21,16 @@ static void load_functions();
...
-#define FLAG_M         (1 << 1)
+
 #define MAX_DIMENSION_FLAG 4
-#define POINT_TYPE     1  /* Identifier for Point geometry type */
-#define LINE_TYPE      2  /* Identifier for Linestring geometry type */
-#define POLYGON_TYPE   3  /* Identifier for Polygon geometry type */
+#define POINT_TYPE       1  /* Identifier for Point geometry type */
+#define LINE_TYPE        2  /* Identifier for Linestring geometry type */
+#define POLYGON_TYPE     3  /* Identifier for Polygon geometry type */
+#define MULTIPOINT_TYPE  4   /* Identifier for MultiPoint geometry type */
```
ahmed-shameem: nit: comment for MULTIPOINT_TYPE should align with other above cases

contrib/babelfishpg_common/src/spatialtypes.c
```diff
@@ -47,13 +46,11 @@ static void load_functions();
 #define POSTGIS_HEADER_SIZE      5      /* Size of the postgis header in bytes */
 #define SRID_SIZE                4      /* Size of the SRID field in bytes */
 #define HEADER_DIMENSION_POS     4      /* Position of dimension info in header */
-#define EMPTY_POINT_TYPE_LASTBYTE    0x01    /* Type identifier for empty point */
-#define EMPTY_LINE_TYPE_LASTBYTE     0x02    /* Type identifier for empty linestring */
-#define EMPTY_POLYGON_TYPE_LASTBYTE  0x03    /* Type identifier for empty polygon */
+#define EMPTY_POINT_TYPE_LASTBYTE       0x01    /* Type identifier for empty point */
+#define EMPTY_LINE_TYPE_LASTBYTE        0x02    /* Type identifier for empty linestring */
+#define EMPTY_POLYGON_TYPE_LASTBYTE     0x03    /* Type identifier for empty polygon */
+#define EMPTY_MULTIPOINT_TYPE_LASTBYTE  0x04  /* Type identifier for empty Multipoint */
```
ahmed-shameem: nit: please align the comment with others. Applies for other cases, if any

contrib/babelfishpg_common/src/spatialtypes.c
```diff
@@ -756,6 +761,34 @@ validate_geography_latitude(Datum geom_datum, bool is_flipped)
                 check_antipodal_points(fcinfo_local, point, i, &prev_lat, &prev_lon);
             }
         }
+    }  else if (strcmp(geom_type, "ST_MultiPoint") == 0)
```
ahmed-shameem, line 769: Are we sure the comparison won't fail if diff case is present? If so, better to use strcasecmp. Else, ignore the comment.

contrib/babelfishpg_common/src/spatialtypes.c
```diff
@@ -1049,7 +1082,102 @@ get_geography_from_text(PG_FUNCTION_ARGS)
...
+    geom_type = GetGeometryTypeName(fcinfo_local, geom_datum);
+    check_geom_type(geom_type);
+
+    /* Validate latitude */
+    validate_geography_latitude(geom_datum, false);
+
+    /* Flip coordinates for geography storage */
+    UpdateFunctionCallInfo(fcinfo_local, 1, geom_datum);
+    geom_datum = st_flipcoordinates_p(fcinfo_local);
+
```
ahmed-shameem, line 1178: Don't we need to call validate_geography_coord_pair?

contrib/babelfishpg_common/src/spatialtypes.c
```diff
@@ -1937,6 +2108,191 @@ handle_point_coordinates(GeometryData *geom_data, uint8 *result_data)
...
+    memcpy(dst + pos, &wkb_type, sizeof(uint32_t));
+    pos += sizeof(uint32_t);
+    pos += write_interleaved_point(dst + pos, src_base, total_points, pt_idx, has_z, has_m);
+    return pos;
+}
+
+
+/* Write child WKB entry based on parent multi-type */
+static uint32_t
+write_child_wkb(uint8 *dst, uint8 *src_base, uint32_t total_points, uint32_t pt_start, uint32_t child_npoints, uint8 parent_type, GeometryData *geom_data, uint32_t fig_start, uint32_t fig_end, bool has_z, bool has_m)
```
ahmed-shameem: Why do we need these many args ?

contrib/babelfishpg_common/src/spatialtypes.c
```diff
@@ -1937,6 +2108,191 @@ handle_point_coordinates(GeometryData *geom_data, uint8 *result_data)
...
+    *child_npoints = pt_end - *pt_start;
+}
+
+/*
+ * Convert CLR multi-geometry binary to PostGIS WKB.
+ */
+static bytea*
+handle_multi_to_postgis(GeometryData *geom_data)
+{
+    bool     has_z = (geom_data->dimension_flag == DIM_FLAG_3D || geom_data->dimension_flag == DIM_FLAG_3DM);
```
ahmed-shameem, line 2236: nit: Can't we have a function to obtain has_z/has_m? IMO, these might be used frequently.

contrib/babelfishpg_common/src/spatialtypes.c
```diff
@@ -1937,6 +2108,191 @@ handle_point_coordinates(GeometryData *geom_data, uint8 *result_data)
...
+    for (i = 0; i < nchildren; i++)
+    {
+        uint32_t pt_start, child_npoints, fig_start, fig_end;
+        get_child_info(geom_data, i + 1, &pt_start, &child_npoints, &fig_start, &fig_end);
+
+        result_size += calculate_child_wkb_size(geom_data->geom_name, child_npoints, geom_data, fig_start, fig_end, has_z, has_m);
+    }
+
+    /* Build PostGIS header */
+    memcpy(postgis_header, "\x01\x04\x00\x00\x20", POSTGIS_HEADER_SIZE);
```
ahmed-shameem: I think better to have some MACRO for such hard coded hex values: "\x01\x04\x00\x00\x20" and use that MACRO instead

ahmed-shameem (changes requested):

Mohitraj-g (commented):
.

Mohitraj-g (commented):
**LLM Code review**
Additional inline review comments on specific lines.

contrib/babelfishpg_common/src/spatialtypes.c
```diff
@@ -130,9 +126,10 @@ static void load_functions();
...
 #define EMPTY_Binary_SIZE      9   /* Size of empty representation in binary */
-#define EMPTY_POINT_Binary   "\x01\x04\x00\x00\x00\x00\x00\x00\x00"  /* Binary for empty point */
-#define EMPTY_LINE_Binary    "\x01\x02\x00\x00\x00\x00\x00\x00\x00"  /* Binary for empty linestring */
-#define EMPTY_POLYGON_Binary "\x01\x03\x00\x00\x00\x00\x00\x00\x00"  /* Binary for empty polygon */
+#define EMPTY_POINT_Bytes      "\x01\x04\x00\x00\x00\x00\x00\x00\x00"  /* Binary for empty point */
+#define EMPTY_LINE_Bytes       "\x01\x02\x00\x00\x00\x00\x00\x00\x00"  /* Binary for empty linestring */
+#define EMPTY_POLYGON_Bytes    "\x01\x03\x00\x00\x00\x00\x00\x00\x00"  /* Binary for empty polygon */
+#define EMPTY_MULTIPOINT_Bytes "\x01\x04\x00\x00\x00\x00\x00\x00\x00"  /* Binary for empty Multipoint */
 #define FIGURE_INTERIOR_RING  0x00
 #define FIGURE_STROKE         0x01
```
Mohitraj-g, line 134: `EMPTY_POINT_Bytes` and `EMPTY_MULTIPOINT_Bytes` have identical values (`\x01\x04\x00...`). This means `st_as_binary_common` produces the same WKB output for empty Point and empty MultiPoint. The CLR binary path distinguishes them via `last_emptybyte`, but the WKB round-trip path (`STAsBinary` → `STGeomFromWKB`) may not be able to tell them apart. Worth verifying this doesn't cause issues.

contrib/babelfishpg_common/src/spatialtypes.c
```diff
@@ -1049,7 +1082,102 @@ get_geography_from_text(PG_FUNCTION_ARGS)
...
+
+    /* Validate latitude */
+    validate_geography_latitude(geom_datum, false);
+
+    /* Flip coordinates for geography storage */
+    UpdateFunctionCallInfo(fcinfo_local, 1, geom_datum);
+    geom_datum = st_flipcoordinates_p(fcinfo_local);
+
+    if (geom_type)
+        pfree(geom_type);
```
Mohitraj-g, line 1180: Typo: `fusnction` → `function`.

contrib/babelfishpg_common/src/spatialtypes.c
```diff
@@ -1937,6 +2108,191 @@ handle_point_coordinates(GeometryData *geom_data, uint8 *result_data)
...
+static uint32_t
+write_wkb_point_child(uint8 *dst, uint8 *src_base, uint32_t total_points, uint32_t pt_idx, bool has_z, bool has_m)
+{
+    uint32_t wkb_type = get_wkb_type(POINT_TYPE, has_z, has_m);
+    uint32_t pos = 0;
+
+    dst[pos++] = 0x01;  /* little endian */
+    memcpy(dst + pos, &wkb_type, sizeof(uint32_t));
+    pos += sizeof(uint32_t);
+    pos += write_interleaved_point(dst + pos, src_base, total_points, pt_idx, has_z, has_m);
```
Mohitraj-g, line 2164: `write_child_wkb` has many unused parameters (`child_npoints`, `geom_data`, `fig_start`, `fig_end`) — only `pt_start` is used for `MULTIPOINT_TYPE`. Same for `calculate_child_wkb_size`. Consider removing unused params now and adding them back when MultiLineString/MultiPolygon support lands.

contrib/babelfishpg_common/src/spatialtypes.c
```diff
@@ -2353,6 +2716,10 @@ determine_geom_dimensions(GeoDataInfo *geom_data)
                         geom_data->geom_type = geom_data->is_valid ? VALID_POLYGON_2D : INVALID_POLYGON_2D;
                         geom_data->coord_size = COORD_SIZE_XY * geom_data->npoints;
                         break;
+                     case MULTIPOINT_TYPE:
```
Mohitraj-g, line 2724: MultiPoint always gets `VALID_*` type codes (e.g., `VALID_2DLINE_MP`), ignoring the `is_valid` flag. Other geometry types check `is_valid` to choose between valid/invalid variants. Is MultiPoint always valid by definition? If so, a comment explaining why would help.

contrib/babelfishpg_common/src/spatialtypes.c
```diff
@@ -2838,8 +3350,9 @@ construct_result_bytea(GeoDataInfo *geom_data, bool is_geography)
      * Add npoints field for complex geometries:
      *   - LineString with >2 points
      *   - Polygon (always has npoints field)
+     *   - MultiPoint (always has npoints field)
      */
-    if ((geom_data->postgis_geom_type == LINE_TYPE && geom_data->npoints > 2) ||(geom_data->postgis_geom_type == POLYGON_TYPE && !geom_data->is_empty))
+    if ((geom_data->postgis_geom_type == LINE_TYPE && geom_data->npoints > 2) ||(geom_data->postgis_geom_type == POLYGON_TYPE && !geom_data->is_empty) || (geom_data->postgis_geom_type == MULTIPOINT_TYPE && !geom_data->is_empty))
     {
         total_size += NPOINTS_SIZE;
     }
```
Mohitraj-g, line 3368: This condition is getting long with MultiPoint added. Consider extracting to a named boolean for readability:
```c
bool needs_npoints = 
    (geom_data->postgis_geom_type == LINE_TYPE && geom_data->npoints > 2) ||
    ((geom_data->postgis_geom_type == POLYGON_TYPE || 
      geom_data->postgis_geom_type == MULTIPOINT_TYPE) && !geom_data->is_empty);
```

contrib/babelfishpg_common/src/geo.c
```diff
@@ -667,3 +699,62 @@ rewrite_dim_polygon_query(PointArrayList *pal)
     
     return output.data;
```
Mohitraj-g, line 686: Trailing whitespace after `char*` on this line. Same on `rewrite_dim_multipoint_wkt` below.

contrib/babelfishpg_common/src/geo.c
```diff
@@ -667,3 +699,62 @@ rewrite_dim_polygon_query(PointArrayList *pal)
...
+    appendStringInfoString(&output, "MULTIPOINT(");
+
+    append_formatted_points(&output, pa, true, format_postgis_point_coordinates);
+
+    appendStringInfoChar(&output, ')');
+
+    free_point_array(pa);
+    
+    return output.data;
+}
```
Mohitraj-g, line 746: Missing newline at end of file.

contrib/babelfishpg_common/sql/upgrades/spatial_types--5.5.0--5.6.0.sql
Mohitraj-g: Inconsistent error message: `STMPointFromText` uses `$1` (raw WKT input) in the ELSE branch, but `STMPointFromWKB` uses `Geomtype`. For large WKT inputs, dumping `$1` can be very noisy. Consider using `Geomtype` consistently across all variants.

contrib/babelfishpg_common/sql/geography.sql
```diff
@@ -228,6 +228,60 @@ CREATE OR REPLACE FUNCTION sys.Geography__stgeomfromtext(sys.NVARCHAR, integer)
...
+
+        geog = sys.geogfromtext_helper($1::text, $2);
+        Geomtype = sys.ST_GeometryType(geog);
+
+        IF Geomtype = 'ST_MultiPoint' THEN
+            RETURN geog;
+        ELSE
+            RAISE EXCEPTION 'Expected "MULTIPOINT" at Position 1. The input has %s', $1;
+        END IF;
+    END;
```
Mohitraj-g, line 252: Same inconsistency: `STMPointFromText` error uses `$1` (raw WKT) while `STMPointFromWKB` error uses `Geomtype`. Consider using `Geomtype` consistently.

contrib/babelfishpg_common/src/geo.c
Gopalverma062: I've now refactored rewrite_linestring_query and rewrite_dim_linestring_query to use append_formatted_points with wrap_each = false. So both paths are actively used now: multipoint uses true (each point wrapped in parens) and linestring uses false (no wrapping).This will also be useful for upcoming MULTILINESTRING support where the same helper can be reused.

contrib/babelfishpg_common/sql/upgrades/spatial_types--5.5.0--5.6.0.sql
Gopalverma062: Yes it will not cause any issue with the upper and lowercase

contrib/babelfishpg_common/src/spatialtypes.c
Gopalverma062: yeah i already verified this ,it is correct

contrib/babelfishpg_common/src/spatialtypes.c
Gopalverma062: The comparison is safe with strcmp. PostGIS's geometry_geometrytype C function always returns type names in the exact format ST_MultiPoint with consistent casing.

contrib/babelfishpg_common/src/spatialtypes.c
Gopalverma062: right , i have removed unused param

contrib/babelfishpg_common/src/spatialtypes.c
Gopalverma062: i have added for multilinstring and multipolygon but i havre removed all unused param

contrib/babelfishpg_common/src/spatialtypes.c
Gopalverma062: right , i removed all the unsed params

contrib/babelfishpg_common/src/spatialtypes.c
Gopalverma062: Yes, MultiPoint is always valid by definition in T-SQL — it's a collection of discrete points with no topological constraints (no ring closure, no minimum point count, no self-intersection rules)

contrib/babelfishpg_common/sql/upgrades/spatial_types--5.5.0--5.6.0.sql
Gopalverma062: Fixed 'Position' to 'position'. Verified this matches the exact error message thrown by T-SQL. Test cases for this is already covered in Test-Multipoint-vu-verify.sql.

contrib/babelfishpg_common/src/geo.c
Gopalverma062: Right, i have removed them

contrib/babelfishpg_common/src/geo_parser.y
Gopalverma062: Right, i have already added the testcases for each case in the test-multipoint-vu-verify.sql file

contrib/babelfishpg_common/src/spatialtypes.c
Gopalverma062: Right, i have added seprate macros for them and used them ,thanks!

contrib/babelfishpg_common/src/spatialtypes.c
Gopalverma062: yes i have changes them ,thanks!

contrib/babelfishpg_common/src/spatialtypes.c
Gopalverma062: validate_geography_latitude is already called on the line above (before the flip), which internally validates each coordinate pair using validate_geography_coord_pair

contrib/babelfishpg_common/sql/upgrades/spatial_types--5.5.0--5.6.0.sql
Gopalverma062: right ,i have removed that

contrib/babelfishpg_common/sql/upgrades/spatial_types--5.5.0--5.6.0.sql
Gopalverma062: yes , i  have verified with the tsql ,TSQL producing the same error msg

contrib/babelfishpg_common/sql/upgrades/spatial_types--5.5.0--5.6.0.sql
Gopalverma062: Tested removing the cast — it results in `function sys.geogfromwkb_helper(varbinary, integer) does not exist`

contrib/babelfishpg_common/src/geo.c
Gopalverma062: Fixed, removed trailing whitespace

contrib/babelfishpg_common/sql/geography.sql
Gopalverma062: right, i have removed  SRID validity check

contrib/babelfishpg_common/sql/geometry.sql
Gopalverma062: removed

contrib/babelfishpg_common/src/spatialtypes.c
Gopalverma062: i have reused the macros

contrib/babelfishpg_common/src/spatialtypes.c
Gopalverma062: done i have used the _Bytes now

contrib/babelfishpg_common/sql/upgrades/spatial_types--5.5.0--5.6.0.sql
Gopalverma062: right ,but current code ,produce same error msg as TSQL

contrib/babelfishpg_common/sql/upgrades/spatial_types--5.5.0--5.6.0.sql
Gopalverma062: right , i have added the comments and also give the spaces

contrib/babelfishpg_common/sql/upgrades/spatial_types--5.5.0--5.6.0.sql
Gopalverma062: fixed

contrib/babelfishpg_common/src/spatialtypes.c
Gopalverma062: right, i have fixed the alignment,thanks

contrib/babelfishpg_common/src/spatialtypes.c
Gopalverma062: fixed

contrib/babelfishpg_common/src/spatialtypes.c
Gopalverma062: I think the current inline checks are straightforward and self-documenting. Adding a helper function for a simple two-condition check would add unnecessary function call overhead without improving readability

contrib/babelfishpg_common/sql/geography.sql
Gopalverma062: The current error message with $1 matches T-SQL's exact error format. Keeping it as-is for T-SQL compatibility. The FromWKB variant uses Geomtype because showing raw binary data wouldn't be meaningful.

contrib/babelfishpg_common/src/geo.c
Gopalverma062: Thanks for the catch! Updated it to return pstrdup("MULTIPOINT EMPTY") to keep it consistent with rewrite_dim_multipoint_wkt. Done.

contrib/babelfishpg_common/src/geo.c
Gopalverma062: Good question. rewrite_multipoint_wkt converts TSQL → PostGIS, so it needs dimension detection and transform_points to normalize mixed dimensions (e.g., adding M prefix, filling missing Z/M with NaN) because PostGIS requires consistent dimensions across all points. rewrite_dim_multipoint_wkt goes the other direction — PostGIS → TSQL. The points coming from PostGIS are already dimension-consistent, so no detection or transformation is needed. We just format them back to TSQL syntax as-is.

contrib/babelfishpg_common/src/spatialtypes.c
Gopalverma062: Addressed — extracted the condition into a needs_npoints boolean as suggested.

contrib/babelfishpg_common/src/geo_parser.y
Gopalverma062: right, i have removed all of them

contrib/babelfishpg_common/sql/upgrades/spatial_types--5.5.0--5.6.0.sql
Gopalverma062: yes i have added spaces

rishabhtanwar29 (approved):

Mohitraj-g (approved):

ahmed-shameem (approved):

PR #4678, opened 2026-03-22
[BABEL] INSERT-EXECUTE Redesign

monk0707:
### Description

**What changed:**

Currently, Babelfish's INSERT...EXECUTE implementation uses a tuple store approach that has several limitations including wrong column insertions, memory leaks and transaction issues. With this change, INSERT...EXECUTE now uses a temp table for buffering and dest receiver for redirecting results into temp table.

This new approach properly handles:
- Correct transaction semantics matching SQL Server behavior
- Proper column mapping and type coercion
- Stable execution without crashes or hangs
- Proper memory management
- OUTPUT clause handling
- All syntax variations (procedures, dynamic SQL, sp_executesql)
- Ownership chaining (procedures calling other procedures owned by the same user)
- Table variables (`@tablevar`) and temp tables (`#temp`) as targets
- TRY-CATCH error handling
- INSTEAD OF triggers on target tables

**Why the change was made:**

The previous tuple store approach had critical issues that couldn't be fixed incrementally:

1. **Transaction behavior inconsistencies**: The rewriting approach didn't properly handle @@TRANCOUNT, implicit transactions, and rollback semantics
2. **Wrong Column Insertion**: Column mapping was incorrect when procedure output didn't match target table column order
3. **Server Crashes and Hangs**: Complex scenarios with nested procedures, triggers, or error handling caused server instability
4. **Memory Management Problems**: The rewriting approach had memory leaks and improper cleanup in error scenarios
5. **OUTPUT Clause Mishandling**: OUTPUT clause on INSERT EXEC wasn't properly supported
6. **Syntax Parsing Gaps**: Many valid T-SQL syntax variations weren't recognized by the parser

The DestReceiver + temp table approach solves these by:
- Using PostgreSQL's native tuple routing mechanism
- Isolating procedure execution from the INSERT operation
- Providing clean transaction boundaries
- Enabling proper error handling and cleanup

**How the code was changed (high-level roadmap):**

1. **pl_exec-2.c**: Core INSERT EXEC implementation
   - `create_insert_exec_temp_table()`: Creates temp table using catalog queries (pg_attribute) to avoid permission issues
   - `CreateInsertExecDestReceiver()`: Custom DestReceiver that captures procedure output
   - `insertexec_receive()`: Handles type coercion and inserts into temp table
   - `flush_insert_exec_temp_table()`: Copies data from temp to target table
   - Security context switching for ownership chaining support

2. **pl_exec.c / iterative_exec.c**: Statement execution
   - Detection of INSERT EXEC context
   - Temp table lifecycle management
   - Error handling and cleanup

3. **tsqlIface.cpp**: Parser changes
   - Proper detection of INSERT EXEC syntax
   - Handling of temp tables (`#`), table variables (`@`), and regular tables
   - Column list extraction

4. **hooks.c**: Utility hooks
   - DROP TABLE protection for active INSERT EXEC target tables (SQL Server error 556)

5. **pltsql.h / pltsql-2.h**: New declarations
   - INSERT EXEC context management functions
   - DestReceiver structures

**Related Engine PR:** https://github.com/babelfish-for-postgresql/postgresql_modified_for_babelfish/pull/749

This engine PR removes the old tuple-store based INSERT EXEC hooks (`called_from_tsql_insert_exec_hook`) that are no longer needed with the new DestReceiver approach.

### Issues Resolved

- Transaction behavior inconsistencies with @@TRANCOUNT and implicit transactions
- Wrong column insertion when procedure output order differs from target
- Server crashes with nested procedures and complex error scenarios
- Memory leaks and improper cleanup
- OUTPUT clause not working with INSERT EXEC
- Various syntax parsing gaps for valid T-SQL
- Ownership chaining broken for INSERT EXEC
- Table variables (`@tablevar`) not recognized correctly
- Type coercion not handling typmod differences (e.g., varchar(max) to varchar(10))

* **Use case based -** Basic INSERT EXEC with procedures, dynamic SQL, sp_executesql; Multiple result sets; Ownership chaining scenarios; Column list specifications; Table variables and temp tables as targets; Nested procedure calls; INSTEAD OF triggers on target tables
* **Boundary conditions -** Empty result sets; Single row vs multiple rows; Maximum column counts; Type coercion edge cases (varchar(max) to varchar(10), int to datetime); Large result sets
* **Arbitrary inputs -** Various data types; NULL handling; Unicode data
* **Negative test cases -** Nested INSERT EXEC (error 8164); Column count mismatch (error 213); Type conversion failures; DROP TABLE on active target (error 556); Permission denied scenarios; Transaction errors
* **Minor version upgrade tests -** N/A (feature redesign)
* **Major version upgrade tests -** N/A (feature redesign)
* **Performance tests -** Basic performance validation with large result sets
* **Tooling impact -** None
* **Client tests -** JDBC tests: BABEL-INSERT-EXEC, BABEL-INSERT-EXECUTE, proc_ownership_chaining, BABEL-2999, temp_table_rollback, BABEL-1944, datareader_datawriter

### Check List


contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -802,13 +802,87 @@ pltsql_bbfCustomProcessUtility(ParseState *pstate, PlannedStmt *pstmt, const cha
...
+				(pltsql_insert_exec_active() && 
+				 (stmt->kind == TRANS_STMT_ROLLBACK || stmt->kind == TRANS_STMT_ROLLBACK_TO ||
+				  stmt->kind == TRANS_STMT_COMMIT || stmt->kind == TRANS_STMT_BEGIN)))
 			{
 				PLTsqlProcessTransaction(parsetree, params, qc);
 				return true;
 			}
 			break;
 		}
+		case T_DropStmt:
```
tanscorpio7: 1. What is SQL Server's behaviour when relation is dropped or altered inside procedure.
2. We should instead just keep and open relation in exec_stmt_exec and let PostgreSQL restrict the bock and alter commands on relation.

contrib/babelfishpg_tsql/src/pltsql.h
```diff
@@ -1172,6 +1172,8 @@ typedef struct PLtsql_stmt_execsql
 	bool		is_tsql_select_assign_stmt; /* T-SQL SELECT-assign (i.e.
 											 * SELECT @a=1) */
 	bool		insert_exec;	/* INSERT-EXEC stmt? */
+	char	   *insert_exec_target;	/* Target table for INSERT-EXEC */
+	char	   *insert_exec_columns;	/* Column list for INSERT-EXEC */
```
tanscorpio7: We should fetch this information from global variable or executor stack for individual statements.

contrib/babelfishpg_tsql/src/pltsql_utils.c
```diff
@@ -809,7 +852,38 @@ void
...
+	/*
+	 * During INSERT EXEC, we simulate SQL Server's implicit transaction.
+	 * COMMIT is only allowed if it would NOT make @@TRANCOUNT = 0.
+	 * 
+	 * In SQL Server:
+	 * - @@TRANCOUNT=1, COMMIT would go to 0: NOT allowed (error 3916)
+	 * - @@TRANCOUNT=2, COMMIT would go to 1: ALLOWED
+	 * 
+	 * The check is: if NestedTranCount <= 1, COMMIT is not allowed.
+	 */
```
tanscorpio7: I am not clear why most of the transaction related changes are needed ? Isn't just skipping auto commit sufficient for INSERT EXEC procedures ?

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -33,6 +44,2058 @@
...
+
+	/*
+	 * Always try to drop any existing temp table with this name first.
+	 * This handles the case where a previous INSERT EXEC failed and couldn't
+	 * clean up its temp table (e.g., error during type coercion in insertexec_receive).
+	 * Using DROP TABLE IF EXISTS is safe and ensures we start with a clean slate.
+	 */
+	initStringInfo(&drop_stmt);
+	appendStringInfo(&drop_stmt, "DROP TABLE IF EXISTS %s", temp_table_name);
+	rc = SPI_execute(drop_stmt.data, false, 0);
```
tanscorpio7: We should instead try another name. Dropping a table is dangerous, for example customer's temp table name conflicts with out internal name.

Instead use a suffix. internal_name_1 & keep increasing the integer until a name is avaiable. See PostgreSQL's handling for constraint names when user does not specify the name

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -33,6 +44,2058 @@
...
+	{
+		/*
+		 * User specified columns - create temp table with only those columns.
+		 * 
+		 * For temp tables and table variables (physical_schema == NULL), we can
+		 * use SELECT ... INTO because they're always owned by the current user.
+		 * 
+		 * For regular tables, we need to use pg_attribute to get column definitions
+		 * to avoid needing SELECT permission (ownership chaining fix).
+		 */
```
tanscorpio7, line 1664: I think we could simply use `CREATE TABLE temp_table LIKE source_table EXCLUDING ALL`. It would reduce the implementation with a small overhead of having redundant columns in temp table.

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -33,6 +44,2058 @@
...
+	}
+	else
+	{
+		/*
+		 * No column list specified - we need to build one excluding
+		 * IDENTITY and computed columns to match the temp table structure.
+		 * 
+		 * Parse the target table name to get the physical schema and table name
+		 * for the catalog query.
+		 */
```
tanscorpio7: Don't they get automatically skipped when INSERT does not specify any columns ?

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -33,6 +44,2058 @@
...
+		 * We save the pointer and restore it after the flush.
+		 */
+		insert_exec_target_table = NULL;
+		
+		/*
+		 * Open a composite trigger nesting level to properly handle
+		 * INSTEAD OF triggers. This ensures trigger events are queued
+		 * and fired correctly.
+		 */
+		BeginCompositeTriggers(CurrentMemoryContext);
```
tanscorpio7: The `INSERT INTO target_table SELECT * FROM temp_table` should pass through execsql again. Why is this handling needed here then ?

tanscorpio7 (changes requested):

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -33,6 +44,2058 @@
...
+static char *insert_exec_target_table = NULL;
+static char *insert_exec_column_list = NULL;
+static int insert_exec_base_tran_count = 0;  /* NestedTranCount when INSERT EXEC started */
+static int insert_exec_saved_nested_tran_count = 0;  /* Original NestedTranCount to restore on cleanup */
+static bool insert_exec_flush_in_progress = false;  /* True during flush phase to block commit_stmt */
+static int insert_exec_call_stack_depth = 0;  /* Call stack depth when INSERT EXEC was started */
+static bool insert_exec_incremented_tran_count = false;  /* True if INSERT EXEC incremented NestedTranCount */
+static bool insert_exec_had_error = false;  /* True if INSERT EXEC had an error - used to skip trancount mismatch check */
+static bool insert_exec_pending_drop = false;  /* True if temp table needs to be dropped when SPI is available */
+static Oid insert_exec_target_rel_oid = InvalidOid;  /* OID of target table - lock held to detect schema changes */
```
tanscorpio7: Can we have one global strucutre instead of several individual variables.

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -33,6 +44,2058 @@
...
+		 * copy values with type coercion where needed.
+		 */
+		Datum	   *values;
+		bool	   *nulls;
+		
+		values = (Datum *) palloc(natts * sizeof(Datum));
+		nulls = (bool *) palloc(natts * sizeof(bool));
+		
+		/* Make sure the source slot is materialized */
+		slot_getallattrs(slot);
```
tanscorpio7: Can we use ExecProject for type casting ?

tanscorpio7 (changes requested):

contrib/babelfishpg_tsql/src/hooks.c
monk0707: 1. There are two cases : 
i. Drop/Alter from the same session : SQL Server allows the DROP/ALTER to execute successfully, then rolls back the transaction to restore the table. It doesn't block the DDL upfront
ii. Drop/Alter from the concurrent session : SQL Server does block the DROP/ALTER from the concurrent session it uses IX(intent exclusive) lock to do so. 

2. Yes, using relation lock at flush time is a better approach that directly matches SQL Server's behavior.
We now acquire RowExclusiveLock on the target table at INSERT EXEC start (blocks concurrent sessions) and capture the schema signature. Since PostgreSQL locks don't block same-session DDL, DROP/ALTER executes successfully like SQL Server. At flush time, we verify the schema - if changed, error 556 is raised and PostgreSQL's transactional DDL automatically rolls back the DROP/ALTER, restoring the table.

contrib/babelfishpg_tsql/src/pltsql.h
monk0707: Yes agree, It will keep the statement structure clean. I will do the required code changes.

contrib/babelfishpg_tsql/src/pl_exec-2.c
monk0707: With LIKE ... EXCLUDING ALL, the temp table would include IDENTITY/computed columns (just without their properties). We'd still need to query pg_attribute to build the column list for the flush INSERT, or drop those columns after creation with ALTER statements. 
The current approach queries pg_attribute once and creates the temp table with only insertable columns - this avoids the extra ALTER overhead and keeps the flush simpler. The "redundant columns" overhead would require either post-creation column drops or explicit column lists during flush, So I think complexity will remain the same either way.

contrib/babelfishpg_tsql/src/pl_exec-2.c
monk0707: Agreed. Will consolidate all the individual insert_exec_* globals into a single InsertExecContext struct and maintain one static instance of it.

contrib/babelfishpg_tsql/src/pl_exec-2.c
monk0707: No, it gives error with error message : 'Number of given values does not match target table definition'

contrib/babelfishpg_tsql/src/pl_exec-2.c
monk0707: Done. The flush (INSERT INTO target_table SELECT * FROM temp_table) now routes through exec_stmt_execsql instead of direct SPI_execute. This reuses the standard execution path, so trigger firing, rowcount tracking, and @@ROWCOUNT/FOUND variable updates are handled automatically. The function was moved from pl_exec-2.c to pl_exec.c to access exec_stmt_execsql and the required estate structures. The manual trigger and security context handling that was here has been removed as a result.

contrib/babelfishpg_tsql/src/pl_exec-2.c
monk0707: Done. Refactored to use ExecProject with a target list. We now build a target list with Var nodes and coercion expressions in insertexec_startup(), create a ProjectionInfo using ExecBuildProjectionInfo(), and call ExecProject() in insertexec_receive() to project the entire tuple at once. This follows the standard PostgreSQL pattern used in nodeModifyTable.c

contrib/babelfishpg_tsql/src/pltsql_utils.c
monk0707: The transactional related changes are needed because SQL Server's starts an implicit transaction, so we have to increment trancount when INSERT EXEC started and restoring it on cleanup. Also we do flush inside a subtransaction, so we have to skip BeginTransactionBlock()/EndTransactionBlock() during INSERT EXEC otherwise PostgreSQL crashes with: "FATAL: BeginTransactionBlock: unexpected state SUBINPROGRESS" (or similar).

The skipping auto-commit can simplify the increment and restoration of trancount but then we will not be having the trancount behavior exactly same as sql server. Also we will still need the additional transaction handling  due to flush being done in a subtransaction.

contrib/babelfishpg_tsql/src/pl_exec-2.c
monk0707: Yes agree, I will fix it. Instead of using DROP TABLE IF EXISTS, We can use a suffix approach similar to PostgreSQL's ChooseConstraintName()

contrib/babelfishpg_tsql/src/pltsql_utils.c
tanscorpio7: Lets discuss this offline next week.

contrib/babelfishpg_tsql/src/pl_exec-2.c
tanscorpio7: The column list will be provided by the user. 
```
-- User query
 INSERT INTO t (col2) EXEC p
 
 -- Internal temp table creation 
CREATE TABLE temp_table LIKE  .... EXCLUDING ALL

-- Insert query (get column name list for insert from user query)
INSERT INTO temp_table (col2) SELECT .....
```

Also please note that there are cases where customer can insert into identity columns using INSERT statement. See test cases for IDENTITY_INSERT

contrib/babelfishpg_tsql/src/pl_exec-2.c
tanscorpio7: Example
```
1> SELECT @@version
2> GO
version                                                                                                                                                                                                                                                         
Babelfish for PostgreSQL with SQL Server Compatibility - 12.0.2000.8
Apr  9 2026 12:28:58
Copyright (c) Amazon Web Services
PostgreSQL 17.9 on x86_64-pc-linux-gnu (Babelfish 5.6.0)  

1> CREATE TABLE t1 (col1 INT, col2 INT IDENTITY(1,1), col3 INT)
2> INSERT INTO t1 VALUES (1,1)
3> SELECT * FROM t1
4> GO

(1 rows affected)
col1        col2        col3       
----------- ----------- -----------
          1           1           1

(1 rows affected)
```

contrib/babelfishpg_tsql/src/pl_exec-2.c
monk0707: But there can be a case where user doesn't provide the column list, 
```
-- User query
 INSERT INTO table EXEC proc
```
I think on sql server this also works fine even when proc does have identity columns. So we need to know which columns are identity/computed to correctly map the procedure's result columns.

Also there is one separate issue with LIKE ... EXCLUDING ALL approach  : Ownership Chaining.
PostgreSQL's CREATE TABLE ... (LIKE source_table EXCLUDING ALL) requires SELECT permission on the source table. This breaks ownership chaining when a user calls a procedure that does INSERT EXEC on a table they don't have direct SELECT access to.

contrib/babelfishpg_tsql/src/pl_exec-2.c
monk0707: Oh right! I've simplified flush_insert_exec_temp_table (commit aa1ef1bef) to use INSERT INTO target (cols) SELECT * FROM temp when columns are specified, and INSERT INTO target SELECT * FROM temp otherwise.

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
```diff
@@ -190,9 +190,11 @@ SendPendingDone(bool more)
 
 			/*
 			 * If a statement throws an error, the row count should be always
-			 * 0.
+			 * 0. Reset it here in case a previous statement (e.g., INSERT
+			 * inside a procedure during INSERT EXEC) had set a non-zero count
+			 * before the error was thrown.
 			 */
-			Assert(TdsPendingDoneRowCnt == 0);
```
KushaalShroff, line 195: I am trying to think what is the impact of this, there is no way we can check insert exec state here, right?

contrib/babelfishpg_tds/error_mapping.txt
```diff
@@ -187,4 +187,6 @@ XX000 ERRCODE_INTERNAL_ERROR	"The table-valued parameter \"%s\" must be declared
 22008 ERRCODE_DATETIME_VALUE_OUT_OF_RANGE	"Adding a value to a \'%s\' column caused an overflow." SQL_ERROR_517 16
 42P01 ERRCODE_UNDEFINED_TABLE	"FOR JSON AUTO requires at least one table for generating JSON objects. Use FOR JSON PATH or add a FROM clause with a table name."		SQL_ERROR_13600	16
 42P01 ERRCODE_FEATURE_NOT_SUPPORTED	"sub-select and values for json auto are not currently supported."		SQL_ERROR_13600	16
+54000 ERRCODE_PROGRAM_LIMIT_EXCEEDED	"nested INSERT ... EXECUTE statements are not allowed"	SQL_ERROR_8164	16
```
KushaalShroff: ERRCODE_PROGRAM_LIMIT_EXCEEDED does not seem right

contrib/babelfishpg_tsql/src/tsqlIface.cpp
```diff
@@ -1947,33 +1978,193 @@ class tsqlBuilder : public tsqlCommonMutator
...
+				}
+			}
 
+			/*
+			 * SQL Server error 483: The OUTPUT clause cannot be used in an INSERT...EXEC statement.
+			 * Check for OUTPUT clause and throw error if present.
+			 */
+			if (ctx->insert_statement()->output_clause())
+			{
+				throw PGErrorWrapperException(ERROR, ERRCODE_FEATURE_NOT_SUPPORTED,
```
KushaalShroff: This is not feature not supported error

contrib/babelfishpg_tsql/src/tsqlIface.cpp
```diff
@@ -1947,33 +1978,193 @@ class tsqlBuilder : public tsqlCommonMutator
...
+					"The OUTPUT clause cannot be used in an INSERT...EXEC statement.",
+					getLineAndPos(ctx->insert_statement()->output_clause()));
+			}
+
+			/*
+			 * INSERT EXEC redesign: Instead of creating a PLtsql_stmt_execsql for
```
KushaalShroff: The comments is final - it is not a redesign

contrib/babelfishpg_tsql/src/tsqlIface.cpp
```diff
@@ -1947,33 +1978,193 @@ class tsqlBuilder : public tsqlCommonMutator
...
+						tbl_db = stripQuoteFromId(ddl_object->full_object_name()->database);
+
+					/*
+					 * Check if this is a temp table (starts with #).
+					 * Temp tables don't need schema prefix - they're in pg_temp schema.
+					 */
+					if (!tbl_name.empty() && tbl_name[0] == '#')
+					{
+						target_table = tbl_name;
+						target_schema = "";  /* Temp tables don't need schema */
```
KushaalShroff: IIRC tsql defaults to dbo schema, is it possible to specify dbo in name?

contrib/babelfishpg_tsql/src/tsqlIface.cpp
```diff
@@ -1947,33 +1978,193 @@ class tsqlBuilder : public tsqlCommonMutator
...
+					else if (!tbl_schema.empty())
+					{
+						target_table = tbl_schema + "." + tbl_name;
+						target_schema = tbl_schema;
+					}
+					else
+					{
+						/* No schema specified - use dbo as default and include in target */
+						target_table = "dbo." + tbl_name;
+						target_schema = "dbo";
```
KushaalShroff: dbo is not right for case where schema was not mentioned - user can specify default schema for a db and that can be different from dbo - does this not get handled later in tsql extension?

contrib/babelfishpg_tsql/src/tsqlIface.cpp
```diff
@@ -1947,33 +1978,193 @@ class tsqlBuilder : public tsqlCommonMutator
...
-					schema_name = stripQuoteFromId(ctx_name->schema);
 			}
+
+			/* Set INSERT EXEC fields based on the actual statement type */
+			if (base_stmt->cmd_type == PLTSQL_STMT_EXEC)
+			{
+				/* Procedure call: EXEC proc_name */
+				PLtsql_stmt_exec *exec_stmt = (PLtsql_stmt_exec *) base_stmt;
+				exec_stmt->insert_exec = true;
+				if (!target_table.empty())
```
KushaalShroff, line 2108: Is that even possible?

contrib/babelfishpg_tsql/src/tsqlIface.cpp
```diff
@@ -1947,33 +1978,193 @@ class tsqlBuilder : public tsqlCommonMutator
...
+					exec_stmt->insert_exec_target = pstrdup(target_table.c_str());
+				if (!column_list.empty())
+					exec_stmt->insert_exec_columns = pstrdup(column_list.c_str());
+			}
+			else if (base_stmt->cmd_type == PLTSQL_STMT_EXEC_BATCH)
+			{
+				/* Dynamic SQL: EXEC(@variable) or EXEC('string') */
+				PLtsql_stmt_exec_batch *exec_batch_stmt = (PLtsql_stmt_exec_batch *) base_stmt;
+				exec_batch_stmt->insert_exec = true;
+				if (!target_table.empty())
```
KushaalShroff, line 2118: Error if target table empty

contrib/babelfishpg_tsql/src/session.c
```diff
@@ -247,17 +248,24 @@ set_search_path_for_user_schema(const char *db_name, const char *user)
...
-					PGC_SUSET, PGC_S_SESSION);
 	/*
-	 * Incase of error inside parallel worker we could reach here before transaction
-	 * abort. Do not try to set_config since it will fail inside a parallel worker
-	 * This function is used in code path where error is never expected
+	 * Skip setting search_path if in parallel mode - parallel workers cannot
+	 * set GUC parameters. This is needed for INSERT EXEC with parallel query.
 	 */
-	if (!IsAbortedTransactionBlockState())
+	if (!IsInParallelMode())
```
KushaalShroff, line 255: Is there no other stmt that ever ran into this issue?

contrib/babelfishpg_tsql/src/iterative_exec.c
```diff
@@ -667,6 +671,39 @@ dispatch_stmt(PLtsql_execstate *estate, PLtsql_stmt *stmt)
...
+			if (pltsql_insert_exec_active() && pltsql_insert_exec_in_execution() &&
+				is_part_of_pltsql_trycatch_block(estate))
+			{
+				PLtsql_stmt_execsql *execsql_stmt = (PLtsql_stmt_execsql *) stmt;
+				if (execsql_stmt->sqlstmt && execsql_stmt->sqlstmt->query)
+				{
+					pltsql_insert_exec_validate_column_count_from_query(execsql_stmt->sqlstmt->query);
+				}
+			}
 			exec_stmt_execsql(estate, (PLtsql_stmt_execsql *) stmt);
```
KushaalShroff: Why not move this logic inside exec_stmt_execsql?

contrib/babelfishpg_tsql/src/pl_insert_exec.c
```diff
@@ -0,0 +1,2563 @@
...
+	Node		*stmt;
+	int			query_natts;
+	Oid			temp_table_oid;
+	Relation	temp_rel;
+	TupleDesc	temp_tupdesc;
+	int			temp_natts;
+	MemoryContext oldcontext;
+
+	/* Only validate if INSERT EXEC is active */
+	if (!pltsql_insert_exec_active())
```
KushaalShroff: This should be assert or internal error - if you reach here then it has to be insert exec

contrib/babelfishpg_tsql/src/pl_insert_exec.c
```diff
@@ -0,0 +1,2563 @@
...
+		return true;
+
+	/* Get the temp table OID */
+	temp_table_oid = pltsql_get_insert_exec_temp_table_oid();
+	if (!OidIsValid(temp_table_oid))
+		return true;  /* No temp table yet, skip validation */
```
KushaalShroff: Same: This should be an internal error unless we have it in a call above which warrants this to be an assert

contrib/babelfishpg_tsql/src/pl_insert_exec.c
```diff
@@ -0,0 +1,2563 @@
...
+	raw_stmt = (RawStmt *) linitial(parsetree_list);
+	stmt = raw_stmt->stmt;
+
+	/* Check if it's a SELECT statement */
+	if (!IsA(stmt, SelectStmt))
+		return true;  /* Not a SELECT, let normal path handle it */
+
+	/* Count target list columns */
+	query_natts = count_select_target_columns((SelectStmt *) stmt);
+	if (query_natts < 0)
```
KushaalShroff, line 872: Please explain expected behaviour here and if it should error out - here and everywhere we are silently sending true instead of erroring

contrib/babelfishpg_tsql/src/pl_insert_exec.c
```diff
@@ -0,0 +1,2563 @@
...
+	}
+}
+
+/*
+ * Set the INSERT EXEC error flag.
+ * Called when an error occurs during INSERT EXEC.
+ * This flag is used to skip the transaction count mismatch check.
+ */
+void
+pltsql_insert_exec_set_error_flag(void)
```
KushaalShroff, line 373: Is it possible that parallel workers set ANY insert_exec_ctx state variables?

contrib/babelfishpg_tsql/src/pl_insert_exec.c
```diff
@@ -0,0 +1,2563 @@
...
+{
+	insert_exec_ctx.had_error = true;
```
KushaalShroff, line 375: We need to double down and check all cases where insert_exec_ctx state could be left stale - timeout/interrupts will always leave these variables assigned without cleanup, right?

contrib/babelfishpg_tsql/src/pltsql-2.h
```diff
@@ -90,6 +90,11 @@ typedef struct PLtsql_stmt_exec
 	char	   *db_name;
 	char	   *proc_name;
 	char	   *schema_name;
+
+	/* INSERT EXEC fields */
+	bool		insert_exec;			/* Is this INSERT EXEC? */
+	char	   *insert_exec_target;		/* Target table for INSERT-EXEC */
+	char	   *insert_exec_columns;	/* Column list for INSERT-EXEC */
```
KushaalShroff: Just create a new struct with these variables and include the struct variable for all 3 types

contrib/babelfishpg_tsql/src/hooks.c
```diff
@@ -802,7 +796,20 @@ pltsql_bbfCustomProcessUtility(ParseState *pstate, PlannedStmt *pstmt, const cha
...
 		{
-			if (NestedTranCount > 0 || (sql_dialect == SQL_DIALECT_TSQL && !IsTransactionBlockActive()))
+			TransactionStmt *stmt = (TransactionStmt *) parsetree;
+			/*
+			 * Call PLTsqlProcessTransaction when:
+			 * 1. NestedTranCount > 0 (explicit transaction)
+			 * 2. TSQL dialect and no active transaction block
+			 * 3. INSERT EXEC is active and it's a ROLLBACK (to block with error 3915)
+			 * 4. INSERT EXEC is active and it's a COMMIT (to block with error 3916)
+			 * 5. INSERT EXEC is active and it's a BEGIN TRAN (to properly track NestedTranCount)
```
KushaalShroff: Could you explain more - avoid error codes

contrib/babelfishpg_tsql/src/iterative_exec.c
```diff
@@ -1177,6 +1214,27 @@ abort_execution(PLtsql_execstate *estate, ErrorData *edata, bool *terminate_batc
...
+	 *
+	 * The error flag is set by:
+	 * 1. pltsql_insert_exec_validate_column_count_from_query() - early validation
+	 * 2. insertexec_startup() - runtime validation in DestReceiver
+	 *
+	 * We need to re-throw the error so it propagates to the subtransaction's
+	 * PG_CATCH in exec_stmt_exec, which will roll back all changes made by
+	 * the procedure.
+	 */
+	if (pltsql_insert_exec_active() && pltsql_insert_exec_had_error())
```
KushaalShroff, line 1201: I didnt understand why we need to introduce our own error logic

contrib/babelfishpg_tsql/src/iterative_exec.c
```diff
@@ -1290,8 +1357,53 @@ dispatch_stmt_handle_error(PLtsql_execstate *estate,
...
+		 *    index references that need to be released
+		 * 2. Without a subtransaction, these references are not released
+		 *    when the error is caught by TRY-CATCH
+		 * 3. This causes "cannot DROP ... because it is being used" errors
+		 *    when the procedure later tries to DROP the temp table
+		 *
+		 * Also skip internal savepoints in parallel workers because
+		 * BeginInternalSubTransaction is not allowed in parallel mode.
+		 */
+		in_trycatch = is_part_of_pltsql_trycatch_block(estate);
```
KushaalShroff, line 1347: Do we have tests for a nested proc having try/catch - both success and failure test case

contrib/babelfishpg_tsql/src/iterative_exec.c
```diff
@@ -1290,8 +1357,53 @@ dispatch_stmt_handle_error(PLtsql_execstate *estate,
...
+		insert_exec_active = pltsql_insert_exec_active();
+
+		/*
+		 * Fallback for cases where exec_state_call_stack is NULL
+		 * (parallel workers, or early-path INSERT EXEC SPI execution).
+		 * Use the TRY-CATCH depth tracked explicitly in INSERT EXEC context.
+		 */
+		if (!in_trycatch && insert_exec_active)
+			in_trycatch = pltsql_insert_exec_in_trycatch();
```
KushaalShroff, line 1356: This seems hacky, lets discuss more offline

contrib/babelfishpg_tsql/src/iterative_exec.c
```diff
@@ -1290,8 +1357,53 @@ dispatch_stmt_handle_error(PLtsql_execstate *estate,
...
+		 *    index references when an error is caught by TRY-CATCH. Without this,
+		 *    the index refcount is not decremented and DROP TABLE fails with
+		 *    "cannot DROP ... because it is being used by active queries"
 		 */
-		if (!ro_func && !pltsql_disable_internal_savepoint && !is_batch_command(stmt) && IsTransactionBlockActive() && !is_set_tran_isolation(stmt))
+		need_savepoint_for_insert_exec_trycatch = insert_exec_active && in_trycatch;
+		txn_active_or_insert_exec_trycatch = IsTransactionBlockActive() || need_savepoint_for_insert_exec_trycatch;
+
+		if (!ro_func && !pltsql_disable_internal_savepoint && !is_batch_command(stmt) && txn_active_or_insert_exec_trycatch && !is_set_tran_isolation(stmt) &&
+			(!insert_exec_active || in_trycatch) && !IsParallelWorker())
```
KushaalShroff, line 1372: For non insert exec case we are adding 2 more conditions in_trycatch and !IsParallelWorker()

the in_trycatch only comes in picture if in_trycatch is true so that safe. But I am not sure about !IsParallelWorker() and it's scope of impact

contrib/babelfishpg_tsql/src/iterative_exec.c
```diff
@@ -1324,15 +1436,53 @@ dispatch_stmt_handle_error(PLtsql_execstate *estate,
...
 		 * Handle transaction count mismatch for batch execution if
 		 * implicit_transaction config is off
+		 *
+		 * Skip this check when inside INSERT EXEC context because:
+		 * 1. INSERT EXEC creates its own temp table for buffering
+		 * 2. The executed procedure may create temp tables, table variables,
+		 *    or use DML with OUTPUT clause, all of which can change the
+		 *    transaction count
+		 * 3. The transaction count will be properly reconciled when the
+		 *    INSERT EXEC completes and flushes to the target table
```
KushaalShroff, line 1412: What is tsql expected behaviour when a nested proc returns @@trancount and it is used to insert? We are reconciling at end due to buffering - but is it the same for txn states? Can you add tests?

contrib/babelfishpg_tsql/src/iterative_exec.c
```diff
@@ -1324,15 +1436,53 @@ dispatch_stmt_handle_error(PLtsql_execstate *estate,
...
+		 *
+		 * Also skip this check if INSERT EXEC had an error. The error flag is
+		 * set when an error occurs during INSERT EXEC (e.g., error 3916 for
+		 * COMMIT without BEGIN TRAN). In this case, we want the original error
+		 * to propagate, not the transaction count mismatch error.
+		 *
+		 * Also skip this check if INSERT EXEC started an implicit transaction.
+		 * The implicit transaction stays open and will be committed by the
+		 * normal auto-commit mechanism. We clear the flag after skipping the
+		 * check so it doesn't affect subsequent batches.
```
KushaalShroff, line 1422: Is there any impact if user sets IMPLICIT_TRANSACTIONS OFF

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -741,14 +843,30 @@ exec_stmt_push_result(PLtsql_execstate *estate,
...
-	if (estate->insert_exec)
-		return exec_stmt_insert_execute_select(estate, stmt->query);
-
 	exec_run_select(estate, stmt->query, &portal);
 
-	receiver = CreateDestReceiver(DestRemote);
-	SetRemoteDestReceiverParams(receiver, portal);
+	/*
+	 * When INSERT EXEC is active (new DestReceiver approach), redirect results
+	 * to the temp table instead of sending to client.
```
KushaalShroff: LLM generated comments usually are context inclined - we must try to keep comments concise from readers perspective

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -785,6 +903,15 @@ exec_run_dml_with_output(PLtsql_execstate *estate, PLtsql_stmt_push_result *stmt
...
 	Assert(stmt->query != NULL);
 
+	/*
+	 * Column count validation for INSERT EXEC is handled by:
+	 * 1. pltsql_insert_exec_validate_column_count_from_query() - validates BEFORE
+	 *    plan preparation to ensure column mismatch errors take priority over
+	 *    runtime errors like division by zero
+	 * 2. insertexec_startup() - validates when DestReceiver starts, as a safety
+	 *    net for cases like SELECT * where early validation can't determine count
+	 */
```
KushaalShroff: Cleanup needed?

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -866,9 +1010,89 @@ exec_stmt_exec(PLtsql_execstate *estate, PLtsql_stmt_exec *stmt)
...
+		{
+			/*
+			 * Check for nested INSERT EXEC - SQL Server error 8164.
+			 * If INSERT EXEC context is already active, this is a nested call.
+			 * Use same error code and message as old code path for consistency.
+			 */
+			if (pltsql_insert_exec_active())
+			{
+				ereport(ERROR,
+						(errcode(ERRCODE_PROGRAM_LIMIT_EXCEEDED),
```
KushaalShroff: Does not seem like an intuitive match for error code - this is an invalid stmt

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -866,9 +1010,89 @@ exec_stmt_exec(PLtsql_execstate *estate, PLtsql_stmt_exec *stmt)
...
+						 errmsg("nested INSERT ... EXECUTE statements are not allowed")));
+			}
+
+			/*
+			 * Start implicit transaction for INSERT EXEC if not already in one.
+			 * This matches SQL Server behavior where INSERT EXEC implicitly starts
+			 * a transaction, making @@TRANCOUNT = 1 inside the executed procedure.
```
KushaalShroff: For both implicit txn on & off?

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -866,9 +1010,89 @@ exec_stmt_exec(PLtsql_execstate *estate, PLtsql_stmt_exec *stmt)
...
+			/* Set global context info for flush function */
+			pltsql_set_insert_exec_context_info(stmt->insert_exec_target, stmt->insert_exec_columns);
+
+			/*
+			 * Open and hold the target table during INSERT EXEC execution.
+			 * This is critical for detecting schema alterations (SQL Server error 556).
+			 * By holding the target table open, PostgreSQL's CheckTableNotInUse()
+			 * will detect if the procedure tries to ALTER TABLE on the target.
+			 */
+			pltsql_insert_exec_open_target_table(stmt->insert_exec_target);
```
KushaalShroff: Keeping it open seems risky and needs full proof cleanup plans - are we looking only for taking a lock?

contrib/babelfishpg_tsql/src/pl_handler.c
```diff
@@ -6490,7 +6503,7 @@ pltsql_call_handler(PG_FUNCTION_ARGS)
 		if (get_cur_db_id() != saved_dbid)
 			set_cur_user_db_and_path(get_db_name((saved_dbid)), false);
 		if (saved_search_path != NULL && strcmp(saved_search_path, namespace_search_path) != 0
-			&& !IsAbortedTransactionBlockState())
+			&& !IsAbortedTransactionBlockState() && !IsInParallelMode())
```
KushaalShroff, line 6501: This affects all use cases - please provide justification for all such cases where we are touching logic outside of insert exec

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -1195,7 +1394,94 @@ exec_stmt_exec(PLtsql_execstate *estate, PLtsql_stmt_exec *stmt)
...
+		 *
+		 * In SQL Server, INSERT EXEC runs in an implicit transaction context,
+		 * so if COMMIT is called without BEGIN TRAN, error 3916 is thrown and
+		 * all changes are rolled back.
+		 *
+		 * OWNERSHIP CHAINING FIX:
+		 * We also switch the security context to the procedure owner's identity
+		 * before executing the inner procedure. This ensures that the EXECUTE
+		 * permission check for the inner procedure uses the procedure owner's
+		 * identity, which is required for ownership chaining to work correctly.
```
KushaalShroff, line 1322: Please involve @HarshLunagariya for chaining review

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -1244,36 +1530,80 @@ exec_stmt_exec(PLtsql_execstate *estate, PLtsql_stmt_exec *stmt)
...
+			 * insert_exec_setup_done was set to TRUE). We need to clean up the
+			 * partially initialized context, but ONLY if this statement is the
+			 * one that set the context.
+			 *
+			 * Compare the target table names to ensure we only clean up if THIS
+			 * statement set the context. If a nested INSERT EXEC fails, the outer
+			 * INSERT EXEC context should remain active.
+			 */
+			const char *active_target = pltsql_get_insert_exec_target_table();
+			if (active_target != NULL && strcmp(active_target, stmt->insert_exec_target) == 0)
```
KushaalShroff: Wont just stmt->insert_exec_target not null be fine? Current stmt has the insert exec target and no other stmt can have nested inser exec

KushaalShroff: If that is a wrong assumption on my side, then the logic seems to assume that the target will be unique in the stack which needs to be double checked

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -1244,36 +1530,80 @@ exec_stmt_exec(PLtsql_execstate *estate, PLtsql_stmt_exec *stmt)
...
+				pltsql_insert_exec_close_target_table();
+				pltsql_clear_insert_exec_context();
 			}
-			ExecDropSingleTupleTableSlot(slot);
-
-			dest->rShutdown(dest);
-			dest->rDestroy(dest);
 		}
-	}
-	PG_FINALLY();
```
KushaalShroff, line 1275: Why are we touching existing behaviour?

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -1317,6 +1667,151 @@ exec_stmt_exec(PLtsql_execstate *estate, PLtsql_stmt_exec *stmt)
...
+	/*
+	 * INSERT EXEC DestReceiver approach: Flush temp table to target
+	 * and cleanup after procedure execution completes.
+	 *
+	 * Execute the flush INSERT directly without a subtransaction wrapper.
+	 * This matches the QTM branch behavior and ensures that:
+	 * 1. Data flushed to the target table is committed at the current
+	 *    transaction level (not nested in a subtransaction)
+	 * 2. If an error occurs AFTER INSERT EXEC completes (e.g., SELECT 1/0),
+	 *    the TRY-CATCH rollback won't undo the already-flushed data
```
KushaalShroff: Lets revisit all comments and make them concise and avoid context from different design approaches

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -1317,6 +1667,151 @@ exec_stmt_exec(PLtsql_execstate *estate, PLtsql_stmt_exec *stmt)
...
+		 */
+		BeginInternalSubTransaction(NULL);
+		MemoryContextSwitchTo(oldcontext);
+
+		PG_TRY();
+		{
+			drop_insert_exec_temp_table(insert_exec_temp_oid);
+			ReleaseCurrentSubTransaction();
+			MemoryContextSwitchTo(oldcontext);
+			CurrentResourceOwner = oldowner;
```
KushaalShroff: Where does ResourceOwner get changed? Why do we need this?

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -1317,6 +1667,151 @@ exec_stmt_exec(PLtsql_execstate *estate, PLtsql_stmt_exec *stmt)
...
+		}
+		PG_CATCH();
+		{
+			/* DROP failed - rollback but don't propagate error */
```
KushaalShroff: Why? We need an internal error

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -1317,6 +1667,151 @@ exec_stmt_exec(PLtsql_execstate *estate, PLtsql_stmt_exec *stmt)
...
+		 * This must be last because it frees temp_table_name which drop needs.
+		 */
+		pltsql_clear_insert_exec_context();
+
+		/*
+		 * Commit the implicit transaction that was started for INSERT EXEC.
+		 *
+		 * The implicit transaction must be committed explicitly here because
+		 * commit_stmt() is only called for PLTSQL_STMT_EXECSQL statements,
+		 * not for PLTSQL_STMT_EXEC statements. Without this explicit commit,
```
KushaalShroff: Would the code be clean/simple/isolated if we just introduced a PLTSQL_STMT_INSERT_EXEC one? Or it will still require all of these changes?

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -1317,6 +1667,151 @@ exec_stmt_exec(PLtsql_execstate *estate, PLtsql_stmt_exec *stmt)
...
+		 * the transaction would remain open and eventually be rolled back,
+		 * causing all INSERT EXEC data to be lost.
+		 *
+		 * We follow the same pattern as commit_stmt() in pl_exec.c:
```
KushaalShroff: Is it duplicated logic? Can we reuse?

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -1488,26 +1983,96 @@ exec_stmt_exec_batch(PLtsql_execstate *estate, PLtsql_stmt_exec_batch *stmt)
...
 	/*
 	 * First we evaluate the string expression. Its result is the
 	 * querystring we have to execute.
 	 */
 	query = exec_eval_expr(estate, stmt->expr, &isnull, &restype, &restypmod);
 	if (isnull)
 	{
-		/* No op in case of null */
+		/* Free old_db_name which was allocated in TopMemoryContext */
+		pfree(old_db_name);
```
KushaalShroff: Why not just set after this check

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -2222,12 +3087,217 @@ exec_stmt_exec_sp(PLtsql_execstate *estate, PLtsql_stmt_exec_sp *stmt)
...
+					if (insert_exec_setup_done)
+					{
+						pltsql_insert_exec_set_error_flag();
+						pltsql_insert_exec_set_pending_drop();
+
+						/*
+						 * Clear the implicit transaction flag since the transaction will be
+						 * rolled back due to the error. If we don't clear this, subsequent
+						 * INSERT EXEC statements might try to commit a non-existent transaction.
+						 */
```
KushaalShroff: Encapsulate duplicate code into single functions wherever possible - including over here

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -2222,12 +3087,217 @@ exec_stmt_exec_sp(PLtsql_execstate *estate, PLtsql_stmt_exec_sp *stmt)
...
+					PG_END_TRY();
+
+					/*
+					 * Clear the INSERT EXEC context AFTER the drop completes.
+					 * This must be last because it frees temp_table_name which drop needs.
+					 */
+					pltsql_clear_insert_exec_context();
+
+					/*
+					 * Commit the implicit transaction that was started for INSERT EXEC.
```
KushaalShroff: Duplicate code - must be single function and call when needed

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -3823,6 +4818,16 @@ execute_plan_and_push_result(PLtsql_execstate *estate, PLtsql_expr *expr, ParamL
...
 	Assert(expr->plan != NULL); /* should be prepared already */
+
+	/*
+	 * Column count validation for INSERT EXEC is handled by:
+	 * 1. pltsql_insert_exec_validate_column_count_from_query() - validates BEFORE
+	 *    plan preparation to ensure column mismatch errors take priority over
+	 *    runtime errors like division by zero
+	 * 2. insertexec_startup() - validates when DestReceiver starts, as a safety
+	 *    net for cases like SELECT * where early validation can't determine count
+	 */
```
KushaalShroff: cleanup?

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -4712,6 +5727,16 @@ set_search_path_for_sp_procs(char *schema)
...
 
+	/*
+	 * Skip setting search_path if in parallel mode - parallel workers cannot
+	 * set GUC parameters. This is needed for INSERT EXEC with parallel query.
+	 */
+	if (IsParallelWorker() || IsInParallelMode())
+	{
+		pfree(dbo_schema);
+		return;
+	}
```
KushaalShroff, line 4993: This affects all queries - please justify

contrib/babelfishpg_tsql/src/pl_exec.c
```diff
@@ -4845,21 +4646,27 @@ exec_stmt_execsql(PLtsql_execstate *estate,
...
+			ereport(ERROR,
+					(errcode(ERRCODE_TRANSACTION_ROLLBACK),
+					 errmsg("Cannot use the ROLLBACK statement within an INSERT-EXEC statement.")));
 		}
+	}
 
+	PG_TRY();
+	{
+		/* Handle naked SELECT stmt differently for INSERT ... EXECUTE */
+		/* Skip tuple store approach if our DestReceiver approach is active */
```
KushaalShroff: What does this mean?

contrib/babelfishpg_tsql/src/pl_insert_exec.c
```diff
@@ -0,0 +1,2563 @@
...
+
+/* Forward declaration for static helper function from pl_exec-2.c */
+static Node *get_underlying_node_from_implicit_casting(Node *n, NodeTag underlying_nodetype);
+
+/* Forward declaration for exec_stmt_execsql from pl_exec.c */
+extern int exec_stmt_execsql(PLtsql_execstate *estate, PLtsql_stmt_execsql *stmt);
+
+/*
+ * Helper function to get underlying node from implicit casting.
+ * Copied from pl_exec-2.c since it's a static function there.
```
KushaalShroff: Dont copy - make non static

contrib/babelfishpg_tsql/src/pl_insert_exec.c
```diff
@@ -0,0 +1,2563 @@
...
+    MemoryContextSwitchTo(oldcontext);
+
+    /*
+     * Examine procedure's argument list. Each output arg position
+     * should be an unadorned pltsql variable (Datum), which we can
+     * insert into the row Datum.
+     */
+    nfields = 0;
+    i = 0;
+    foreach(lc, funcargs)
```
KushaalShroff, line 2108: Seems like duplicate code from exec_stmt_exec please justify

KushaalShroff (changes requested):

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
monk0707: yes right, we can't check INSERT EXEC state here in the TDS layer. But on my investigation, I don't find anything else being broke or interrupted due to it

contrib/babelfishpg_tds/error_mapping.txt
monk0707: can't find a perfect fit for it, but I think ERRCODE_SYNTAX_ERROR can be more close

contrib/babelfishpg_tsql/src/hooks.c
monk0707: sure, let me add more explained comment

contrib/babelfishpg_tsql/src/iterative_exec.c
monk0707: we can actually, that will be good to do

contrib/babelfishpg_tsql/src/iterative_exec.c
monk0707: This handles the error priority issue when both column mismatch AND division by zero occur in the same INSERT EXEC with TRY-CATCH.

Ex : 
`
INSERT INTO #t(a) EXEC proc  -- proc returns 2 columns, #t has 1 column
-- Inside proc: SELECT 1/0, 2  (div by zero + wrong column count)
`

**T-SQL behavior**: Column mismatch error takes priority, entire INSERT EXEC is rolled back.

**Without this fix**: Division by zero error is caught by TRY-CATCH first (because PostgreSQL evaluates expressions during plan preparation), masking the column mismatch error.

contrib/babelfishpg_tsql/src/iterative_exec.c
monk0707: Added a test for each case

tanscorpio7:
This PR covers a lot of ground and touches some risky areas (e.g., the iterative executor) that I think we can avoid modifying if we restructure the approach. To make it easier to review and safer to merge, could we break it into smaller, focused PRs? For example:

1. Add GUC for insert-execute mode
2. ANTLR parser changes for INSERT ... EXECUTE
3. Helper functions and structs for insert-execute
4. Introduce new DestReceiver
5. Temporary tables helper functions
6. and more ...

contrib/babelfishpg_tsql/src/iterative_exec.c
monk0707: Verified the tsql behavior - @@TRANCOUNT is correctly visible inside nested procs during INSERT EXEC. The actual transaction state is not affected by buffering; we only skip the batch-level mismatch validation check (which would incorrectly flag temp table/table variable creation as mismatches). Also added tests for the same!

contrib/babelfishpg_tsql/src/iterative_exec.c
monk0707: No. The behavior matches with tsql when IMPLICIT_TRANSACTIONS OFF.

contrib/babelfishpg_tsql/src/pl_exec-2.c
monk0707: yes agree, actually I reviewed most of them, some left mistakenly due to big code addition. Will update it!

contrib/babelfishpg_tsql/src/pl_exec-2.c
monk0707: agree, changed it to ERRCODE_SYNTAX_ERROR

contrib/babelfishpg_tsql/src/pl_exec-2.c
monk0707: yes, The proc always sees @@TRANCOUNT=1. With OFF, we start an implicit txn that auto-commits; with ON, a transaction already exists so we use it.

contrib/babelfishpg_tsql/src/pl_exec-2.c
monk0707: Yes, we are not keeping the relation open. The lock is released in cleanup or automatically at transaction end.

contrib/babelfishpg_tsql/src/pl_exec-2.c
monk0707: okay

contrib/babelfishpg_tsql/src/pl_exec-2.c
monk0707: oh that's right. Let me simplify it.

contrib/babelfishpg_tsql/src/pl_exec-2.c
monk0707: BeginInternalSubTransaction() changes it same as exec_stmt_try_catch() in this file.

contrib/babelfishpg_tsql/src/pl_exec-2.c
monk0707: yes, added elog(WARNING, ...) to log when DROP fails.

contrib/babelfishpg_tsql/src/pl_exec-2.c
monk0707: yes we can by making commit_stmt() non-static.

contrib/babelfishpg_tsql/src/pl_exec-2.c
monk0707: I think doing so will still require all these changes.

contrib/babelfishpg_tsql/src/pl_exec.c
monk0707: it's not actually correct, let me remove it!

contrib/babelfishpg_tsql/src/pl_insert_exec.c
monk0707: No, the insert_exec_ctx is only accessed/set by the leader/main backend process

contrib/babelfishpg_tsql/src/pl_insert_exec.c
monk0707: yes, but I will update code for it and make sure, we cleanup in all success and error paths possible.

contrib/babelfishpg_tsql/src/pl_insert_exec.c
monk0707: Returning true here means "proceed with execution". When query_natts < 0, we couldn't count columns statically (e.g., SELECT *), so we let execution continue and the DestReceiver will validate at runtime.

contrib/babelfishpg_tsql/src/pl_insert_exec.c
monk0707: It is similar but not identical, exec_stmt_exec has extra logic for named params (@name = value) and OUTPUT keyword checking that INSERT EXEC doesn't need. Extracting a shared helper would require many parameters and conditionals which may reduce the readability.

contrib/babelfishpg_tsql/src/tsqlIface.cpp
monk0707: yes, it will be handled later.

kuntalghosh (commented):
# AI Code Review

Verdict: REQUEST_CHANGES

---

## Critical — Security

**Issue 1 - In `pl_insert_exec.c` at lines 2414, 2426: SQL injection in flush INSERT via unquoted identifiers:**

```c
appendStringInfo(&flush_query,
    "INSERT INTO %s (%s) SELECT * FROM %s",
    target_table,
    column_list,
    temp_table_name);
```

`target_table` and `column_list` originate from user-supplied T-SQL identifiers processed by `stripQuoteFromId()`, which removes bracket/double-quote wrappers without re-escaping embedded special characters (`;`, `'`, `"`). These values are interpolated with bare `%s` into the flush INSERT statement. Critically, this flush executes **after** `SetUserIdAndSecContext()` switches identity to the procedure owner (ownership chaining), making this a privilege escalation vector: a low-privilege caller can inject arbitrary SQL that runs as `dbo`/`db_owner`.

This same pattern appears in 9+ additional sites in `create_insert_exec_temp_table()` (lines 1773, 1797, 1824, 1841, 1983, 2014, 2024, 2042, 2087) where schema/table/column names are interpolated into `SPI_execute` calls.

Suggested fix: Use `quote_identifier()` for all identifier interpolation and `quote_qualified_identifier()` for schema.table references. Parse `column_list` into a `List *` of individual column names in the parser (tsqlIface.cpp) rather than round-tripping through a comma-delimited string.

```c
/* Example for flush query */
appendStringInfo(&flush_query,
    "INSERT INTO %s (%s) SELECT * FROM %s",
    quote_qualified_identifier(schema_name, table_name),
    quoted_column_list,  /* built from quote_identifier() per column */
    quote_identifier(temp_table_name));
```

**Issue 2 - In `pl_insert_exec.c` at line 2441: `get_func_owner()` result unchecked before `SetUserIdAndSecContext`:**

```c
SetUserIdAndSecContext(get_func_owner(estate->func->fn_oid),
                       flush_save_sec_context | SECURITY_LOCAL_USERID_CHANGE);
```

`get_func_owner()` returns `InvalidOid` on cache miss. Passing `InvalidOid` to `SetUserIdAndSecContext` causes ACL evaluation to run under an invalid identity.

Suggested fix:

```c
Oid owner = get_func_owner(estate->func->fn_oid);
if (!OidIsValid(owner))
    ereport(ERROR,
             errmsg("could not determine owner of function %u",
                    estate->func->fn_oid)));
SetUserIdAndSecContext(owner,
                       flush_save_sec_context | SECURITY_LOCAL_USERID_CHANGE);
```

---

## High — Correctness / Crash Risk

**Issue 3 - In `pl_insert_exec.c` at lines 638, 659, 789, 940, 973: `return` inside `PG_CATCH` without `PG_END_TRY`:**

```c
PG_CATCH();
{
    FlushErrorState();
    return;  /* BUG: skips PG_END_TRY() */
}
PG_END_TRY();
```

Returning from inside `PG_CATCH` without executing `PG_END_TRY()` leaves `PG_exception_stack` pointing at a destroyed `sigjmp_buf`. The next `ereport(ERROR)` anywhere in the backend will `longjmp` to a dead stack frame, causing a crash or undefined behavior. This is a known Babelfish anti-pattern (BUG-115 class). There are 5 occurrences in this file.

Suggested fix: Set a volatile flag in the CATCH block, fall through to `PG_END_TRY()`, then check the flag:

```c
volatile bool caught_error = false;
PG_TRY();
{
    LockRelationOid(relid, RowExclusiveLock);
}
PG_CATCH();
{
    FlushErrorState();
    caught_error = true;
}
PG_END_TRY();
if (caught_error)
    return;
```

**Issue 4 - In `pl_insert_exec.c` at lines 2318-2322: `flush_save_userid` / `flush_save_sec_context` not `volatile`:**

```c
Oid     flush_save_userid = InvalidOid;
int     flush_save_sec_context = 0;
volatile bool flush_switched_context = false;
```

`flush_save_userid` and `flush_save_sec_context` are read in the `PG_CATCH` block to restore the security context. Without `volatile`, the compiler may optimize them into registers that are clobbered by `siglongjmp`. The `PG_CATCH` block would then restore stale/uninitialized auth context. The same pattern exists at `pl_exec-2.c:1416-1418`.

Suggested fix: Add `volatile` qualifier to both variables.

---

## Medium — Coding Conventions

**Issue 5 - In `pl_insert_exec.c` at line 68: Static function duplicated from `pl_exec-2.c`:**

```c
/* Copied from pl_exec-2.c since it's a static function there. */
static Node *
get_underlying_node_from_implicit_casting(Node *n, NodeTag underlying_nodetype)
```

This copies ~40 lines of `get_underlying_node_from_implicit_casting()` from `pl_exec-2.c`. Code duplication creates a maintenance burden — if the original is fixed, this copy won't be. Per coding convention rule 8 (hooks/isolation), shared utility functions should be made non-static and declared in a header.

Suggested fix: Remove the `static` qualifier from the original in `pl_exec-2.c`, add an `extern` declaration in `pltsql-2.h`, and remove the copy from `pl_insert_exec.c`.

**Issue 6 - In `pl_insert_exec.c` at line 1710: Error message violates PostgreSQL conventions (rule 7):**

```c
elog(ERROR, "INSERT-EXEC: Could not find available temp table name after 10000 attempts");
```

Multiple error messages in this file use capitalized first letters, end with periods, use abbreviations, and use `elog(ERROR)` for user-facing messages. Per rule 7: primary `errmsg()` should never capitalize the first letter, never end with a period, and user-facing messages should use `ereport()` not `elog()`.

Suggested fix:

```c
ereport(ERROR,
        (errcode(ERRCODE_PROGRAM_LIMIT_EXCEEDED),
         errmsg("could not find available temp table name for INSERT EXEC")));
```

**Issue 7 - In `pl_insert_exec.c`: Extensive use of TopMemoryContext (anti-pattern):**

Lines 371, 385, 1717, 648, 668, etc. allocate strings and structs in `TopMemoryContext`. Per the known anti-pattern (search_patterns result: "Avoid TopMemoryContext for per-statement allocations"), these allocations persist for the entire backend lifetime and are never reclaimed during errors. While the code does have explicit `pfree()` calls, any error path that skips cleanup will leak memory permanently.

Suggested fix: Create a dedicated child memory context (e.g., `InsertExecContext`) that can be explicitly destroyed on cleanup, rather than allocating individual strings in `TopMemoryContext`.

**Issue 8 - In `pl_insert_exec.c` at line 1810: `strtok_r` re-parsing of column_list loses embedded commas:**

```c
col_name = strtok_r(col_list_copy, ", ", &saveptr);
```

The `column_list` string is a comma-separated list built in `tsqlIface.cpp`. If a column name contains a comma (possible with bracket-quoted identifiers like `[col,name]`), `strtok_r` will split it incorrectly. This is related to Issue 1 — the column list should be passed as a structured `List *` rather than a delimited string.

---

## Medium — Testing

**Issue 9 - Test file named after JIRA ticket: `BABEL-INSERT-EXEC.sql`:**

Per testing rule 10: "NEVER name test files after JIRA ticket IDs — always use feature or domain-based names." The new test file `test/JDBC/input/BABEL-INSERT-EXEC.sql` should be renamed to something like `insert-exec.sql` or added to the existing `BABEL-INSERT-EXECUTE.sql` test file.

**Issue 10 - Missing test coverage for security-sensitive paths:**

Per testing rule 10 (permissions): The test file does not include tests with different permission levels (sysadmin, db_owner, regular user with minimal grants). Given that this PR implements ownership chaining with `SetUserIdAndSecContext`, permission-based tests are critical to verify that:
- A low-privilege user can INSERT EXEC into a table they have INSERT permission on via a procedure they have EXEC permission on
- A low-privilege user CANNOT escalate privileges through the INSERT EXEC path

**Issue 11 - Missing prepared statement tests:**

Per testing rule 11: The test file does not include tests via `sp_executesql` with typed parameters, `sp_prepare`/`sp_execute`, or JDBC `PreparedStatement` for the INSERT EXEC feature. These exercise different parsing and execution paths.

**Issue 12 - Missing multi-db mode tests:**

Per testing rule 7: The test file does not test INSERT EXEC behavior in both single-db and multi-db migration modes. Given that the code uses `get_physical_schema_name(get_cur_db_name(), schema_name)` for schema resolution, mode-specific bugs are likely.

---

## Low — Reviewer Comment Alignment

**Issue 13 - Reviewer tanscorpio7 suggested using `CREATE TABLE ... LIKE ... EXCLUDING ALL` (unresolved):**

At `pl_exec-2.c:1329`, reviewer tanscorpio7 suggested simplifying temp table creation with `CREATE TABLE temp_table LIKE source_table EXCLUDING ALL` instead of querying `pg_attribute`. This would significantly reduce the code in `create_insert_exec_temp_table()` and eliminate the SQL injection surface in the column-definition-building code. The current approach queries `pg_attribute` to avoid needing SELECT permission (ownership chaining), but `LIKE` also reads from the catalog and doesn't require SELECT permission. This suggestion deserves serious consideration.

**Issue 14 - Reviewer tanscorpio7 suggested a global struct instead of individual variables (unresolved):**

At `pl_exec-2.c:135`, the reviewer suggested consolidating global variables into a single struct. The PR does use `InsertExecContext` struct in `pl_insert_exec.c`, but the header declarations in `pltsql.h` expose 30+ individual `extern` functions. Consider whether some of these can be consolidated into fewer accessor functions.

### Missing Tests
- Permission-based tests (sysadmin vs regular user) for ownership chaining
- Prepared statement paths (sp_executesql, sp_prepare/sp_execute)
- Multi-db vs single-db mode behavior
- Concurrent session tests (isolation tests) for the RowExclusiveLock behavior
- Collation-sensitive column name matching
- Bracket-quoted identifiers with special characters in table/column names (security regression test)

### Notes
- The leak warning count change in `scan-warnings.sh` (416 → 30) is a significant improvement and suggests the old tuple-store approach had substantial leaks. This is a positive signal.
- The `Assert(TdsPendingDoneRowCnt == 0)` → assignment change in `tdsresponse.c` (line 193) converts a crash-on-assertion-failure into silent recovery. This is pragmatic but should be documented as a known deviation.
- The PR has 43 reviews and 93 inline comments from tanscorpio7, many of which appear unresolved. Several of the reviewer's suggestions (LIKE approach, global struct, avoiding iterative_exec changes) could simplify the implementation significantly. These should be addressed before merge.
- The related engine PR (postgresql_modified_for_babelfish#749) removes the old `called_from_tsql_insert_exec_hook` — both PRs must be merged together.

kuntalghosh (changes requested):

contrib/babelfishpg_tsql/src/tsqlIface.cpp
monk0707: agree, I think better could be ERRCODE_SYNTAX_ERROR

contrib/babelfishpg_tsql/src/pl_exec-2.c
monk0707: okay

PR #4705, opened 2026-04-07
fix: Deep copy parameter names for prepared statements

Vaurkhorov:
### Description

Currently, when cloning arguments to pltsql functions, we only copy the pointer to the argument names, which means that when a prepared statement is stored, the strings containing the parameter names are not retained.

When later executing a prepared statement using its handle, we don't have the original param names, and so we cannot call them using their arguments.

To fix this, we copy the argument names as well when cloning the struct.

### Issues Resolved
BABEL-6391

### Check List

### DIFF
--- a/contrib/babelfishpg_tsql/src/pl_exec-2.c
+++ b/contrib/babelfishpg_tsql/src/pl_exec-2.c
@@ -2763,14 +2763,13 @@ InlineCodeBlockArgs *
 clone_inline_args(InlineCodeBlockArgs *args)
 {
 	InlineCodeBlockArgs *clone;
-	int			i;
 
 	clone = create_args(args->numargs);
 	memcpy(clone->argtypes, args->argtypes, sizeof(Oid) * args->numargs);
 	memcpy(clone->argtypmods, args->argtypmods, sizeof(int32) * args->numargs);
 
 	/* Deep copy argument names */
-	for (i = 0; i < args->numargs; i++)
+	for (int i = 0; i < args->numargs; i++)
 		clone->argnames[i] = args->argnames[i] ? pstrdup(args->argnames[i]) : NULL;
 
 	memcpy(clone->argmodes, args->argmodes, sizeof(char) * args->numargs);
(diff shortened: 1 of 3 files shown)


contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -2717,7 +2717,7 @@ read_param_def(InlineCodeBlockArgs *args, const char *paramdefstr)
 		FunctionParameter *p;
 
 		p = (FunctionParameter *) lfirst(lc);
-		args->argnames[i] = p->name;
+		args->argnames[i] = pstrdup(p->name);
```
tanscorpio7, line 2720: Can the arg name be NULL ?

contrib/babelfishpg_tsql/src/pl_exec-2.c
```diff
@@ -2763,11 +2763,16 @@ InlineCodeBlockArgs *
 clone_inline_args(InlineCodeBlockArgs *args)
 {
 	InlineCodeBlockArgs *clone;
+	int			i;
```
tanscorpio7: Inline variable declaration ?

tanscorpio7 (changes requested):

contrib/babelfishpg_tsql/src/pl_exec-2.c
Vaurkhorov: We parse the parameters within the same function, so they should not be deallocated yet. If there are no parameters, we won't enter the foreach loop, so the name should never be NULL

contrib/babelfishpg_tsql/src/pl_exec-2.c
Vaurkhorov: Done

contrib/babelfishpg_tsql/src/pl_exec-2.c
tanscorpio7: I meant is there a case where param is defined but name is null.

tanscorpio7 (changes requested):

contrib/babelfishpg_tsql/src/pl_exec-2.c
Vaurkhorov: No, the raw_parser will always look for an identifier when parsing the param list here, so the name can't be null

tanscorpio7 (approved):

PR #4706, opened 2026-04-07

Mohitraj-g:
### Description

```
-- CASE: fails
```
```
-- COALESCE: fails
```

### Fix

### Issues Resolved

BABEL-6402
BABEL-6427

* **Use case based -**

* **Boundary conditions -**

* **Arbitrary inputs -**

* **Negative test cases -**

* **Performance tests -**

* **Tooling impact -**

* **Client tests -**

### Check List


contrib/babelfishpg_tsql/src/pltsql_coerce.c
```diff
@@ -3286,22 +3307,18 @@ select_common_type_for_coalesce_function(ParseState *pstate, List *exprs)
 
 		/* Check if arg is NULL literal */
 		if (IsA(pexpr, Const) && ((Const *) pexpr)->constisnull)
```
Yvinayak07, line 3309: Since we're only checking for NULL on Const nodes, could there be other expression types that represent NULL at this stage?

test/JDBC/input/sql_variant/babel_sqlvariant_coerce_type-vu-verify.sql
```diff
@@ -0,0 +1,855 @@
...
+
+SELECT * FROM (VALUES
+  (CAST(0 AS bit)),
+  (CAST(100 AS int)),
+  (CAST(NULL AS sql_variant))
+) AS t(col) ORDER BY col
+GO
+
+
+-- 9. REGRESSION: char-only CASE (must be unchanged)
```
Yvinayak07, line 807: What do we mean by regression here could we remove the keyword and just explain what the test case is about.

test/JDBC/upgrade/latest/schedule
```diff
@@ -681,3 +681,4 @@ babel_convert_with_style
 forxml-raw-elements
 forxml-raw-elements-1gb-limit
 forxml-path-elements
+babel_sqlvariant_coerce_type
```
Yvinayak07, line 684: Could we also add this test file to other schedule paths for upgrade testing.

contrib/babelfishpg_tsql/src/pltsql_coerce.c
```diff
@@ -3173,6 +3173,27 @@ select_common_type_setop(ParseState *pstate, List *exprs, Node **which_expr, con
...
 	bool		is_case_expr = (strlen(context) == 4 && strncmp(context, "CASE", 4) == 0);
 
+	/*
+	 * If any branch is sql_variant, return it directly since it has 
+	 * the highest precedence and can hold any type. We must check this 
+	 * before the main loop because the loop bails out on non-char types, 
+	 * and the other branch (e.g. bit) would cause an early return of 
+	 * InvalidOid before sql_variant is ever seen.
+	 */
+	foreach(lc, exprs)
```
Yvinayak07, line 3183: We should also check for the condition if UDT is made on top of `SQL_VARIANT`.

contrib/babelfishpg_tsql/src/pltsql_coerce.c
```diff
@@ -3286,22 +3307,18 @@ select_common_type_for_coalesce_function(ParseState *pstate, List *exprs)
 
 		/* Check if arg is NULL literal */
 		if (IsA(pexpr, Const) && ((Const *) pexpr)->constisnull)
-			continue;
+		{
+			/*
+			 * Consider sql_variant type even for NULL literals since it has highest
+			 * precedence so it should always win regardless of NULL value.
+			 */
+			if (!(*common_utility_plugin_ptr->is_tsql_sqlvariant_datatype)(ptype))
```
Yvinayak07, line 3315: Check for UDT goes here as well.

test/JDBC/input/sql_variant/babel_sqlvariant_coerce_type-vu-prepare.sql
```diff
@@ -0,0 +1,36 @@
+-- Prepare: sql_variant type resolution tests
```
Yvinayak07, line 1: Do we have the User query in our tests case, if no let's add that as well.

Yvinayak07, line 1: Have we tested sql_variant against all other data types i.e., one branch as sql_variant and the other branch cycling through each supported type individually?

test/JDBC/input/sql_variant/babel_sqlvariant_coerce_type-vu-verify.sql
```diff
@@ -0,0 +1,855 @@
...
+   WHEN 1 = 4 THEN CAST(3.14 AS float)
+   ELSE CAST('fallback' AS sql_variant)
+END AS result
+GO
+
+-- 1g. SERVERPROPERTY returns sql_variant (original reported queries)
+SELECT CASE WHEN 1 = 1 THEN SERVERPROPERTY('IsFullTextInstalled') ELSE CAST(0 AS bit) END AS result
+GO
+
+SELECT CASE WHEN SERVERPROPERTY('EngineEdition') = 8 THEN SERVERPROPERTY('IsFullTextInstalled') ELSE CAST(0 AS bit) END AS test_result
```
Yvinayak07, line 177: Could you also check some other System functions that return `sql_variant` and those test cases as well.

contrib/babelfishpg_tsql/src/pltsql_coerce.c
Mohitraj-g: At this stage , NULL literals are always represented as Const nodes with `constisnull=true.` Other expression types won't be NULL at parse time, they have a valid type OID.

test/JDBC/upgrade/latest/schedule
Mohitraj-g: Added

contrib/babelfishpg_tsql/src/pltsql_coerce.c
Mohitraj-g: Added `get_immediate_base_type_of_UDT_internal()` to resolve UDTs before the `is_tsql_sqlvariant_datatype` check in the pre-scan loop.

contrib/babelfishpg_tsql/src/pltsql_coerce.c
Mohitraj-g: Added

test/JDBC/input/sql_variant/babel_sqlvariant_coerce_type-vu-prepare.sql
Mohitraj-g: The original user query from BABEL-6402 is already covered

test/JDBC/input/sql_variant/babel_sqlvariant_coerce_type-vu-verify.sql
Mohitraj-g: Added tests covering sql_variant against date, datetime, time, money, smallmoney, bigint, smallint, tinyint, and uniqueidentifier in CASE, COALESCE, and UNION context

test/JDBC/input/sql_variant/babel_sqlvariant_coerce_type-vu-prepare.sql
Mohitraj-g: Yes, have tests for most of them

contrib/babelfishpg_tsql/src/pltsql_coerce.c
```diff
@@ -3173,6 +3173,30 @@ select_common_type_setop(ParseState *pstate, List *exprs, Node **which_expr, con
 	ListCell	*lc;
 	bool		is_case_expr = (strlen(context) == 4 && strncmp(context, "CASE", 4) == 0);
 
+	/*
+	 * If any branch is sql_variant, return it directly since it has 
+	 * the highest precedence and can hold any type. We must check this 
+	 * before the main loop because the loop bails out on non-char types, 
+	 * and the other branch (e.g. bit) would cause an early return of 
+	 * InvalidOid before sql_variant is ever seen.
+	 */
```
rohit01010, line 3182: Instead of adding a separate handling for sql_variant, why haven't we considered on generalizing existing logic to also handle sql_variant ?. Can we improve following condition to also handle other datatypes such as sql_variant etc instead of adding a separate logic for sql_variant ?
```
		else if ((!is_tsql_char_type_with_len(type, is_case_expr)))
			return InvalidOid;
```

contrib/babelfishpg_tsql/src/pltsql_coerce.c
```diff
@@ -3284,24 +3308,28 @@ select_common_type_for_coalesce_function(ParseState *pstate, List *exprs)
...
+			 * Consider sql_variant type even for NULL literals since it has highest
+			 * precedence so it should always win regardless of NULL value.
+			 * Also check UDTs based on sql_variant.
+			 */
+			Oid		base = get_immediate_base_type_of_UDT_internal(ptype);
+			Oid		check_type = OidIsValid(base) ? base : ptype;
+
+			if (!(*common_utility_plugin_ptr->is_tsql_sqlvariant_datatype)(check_type))
+				continue;
+		}
```
rohit01010, line 3333: Is this only required in case of sql_variant ? what happens if we have a NULL with other datatype ?

test/JDBC/expected/babel_sqlvariant_coerce_type-vu-verify.out
```diff
@@ -0,0 +1,1759 @@
+
+
+
+
+-- Test: sql_variant type resolution across all common type contexts
+-- Covers: CASE, ISNULL, COALESCE, UNION, INTERSECT, EXCEPT, VALUES
+-- Tests sql_variant in 1st, 2nd, 3rd, and last branch positions
+-- Tests NULL sql_variant in all contexts
+-- 1. CASE: sql_variant in different branch positions
+-- 1a. sql_variant in 1st branch (THEN), other type in ELSE
```
rohit01010, line 10: For each of these cases, can you also add test where one of the branch is sql_variant and another one is User Defined Types

test/JDBC/upgrade/16_14/schedule
rohit01010, line 1: Let's add this test to 16_13, 15_17 and 14_22 schedule file as well.

contrib/babelfishpg_tsql/src/pltsql_coerce.c
Mohitraj-g: The main loop's `is_tsql_char_type_with_len` check is specifically for char-type precedence, it returns `InvalidOid` for any non-char type so PG's default resolution takes over. We can't generalize it because an earlier attempt to let all precedence-table   
  types through broke `numeric + date`, `name + nvarchar`, and `varchar + nchar` cases. sql_variant is unique - it has highest precedenc and can hold any type, so a targeted pre-scan is the safest approach.

contrib/babelfishpg_tsql/src/pltsql_coerce.c
Mohitraj-g: Yes, this is only needed for sql_variant. For other datatypes, skipping NULL Const nodes is correct , the non-NULL args determine the common type through normal precedence. sql_variant is the exception because it has highest precedence. If we skip `CAST(NULL AS sql_variant)`, we lose the type info and the result gets determined by lower-precedence types. For example, `COALESCE(CAST(NULL AS            
  sql_variant), CAST(0 AS bit))` should resolve to `sql_variant`, not `bit` — which matches SQL Server behavior.

test/JDBC/expected/babel_sqlvariant_coerce_type-vu-verify.out
Mohitraj-g: Added

test/JDBC/expected/babel_sqlvariant_coerce_type-vu-verify.out
```diff
@@ -1757,3 +1757,127 @@ sql_variant
...
+false
+~~END~~
+
+
+-- 15c. COALESCE: sql_variant vs UDT(int)
+SELECT COALESCE(CAST(NULL AS sql_variant), CAST(100 AS babel_sqlvariant_udt_int)) AS result
+GO
+~~ERROR (Code: 33557097)~~
+
+~~ERROR (Message: COALESCE could not convert type sql_variant to babel_sqlvariant_udt_int)~~
```
rohit01010: is the error expected ? shouldn't both the parameters be converted to sql_variant ?

contrib/babelfishpg_tsql/src/pltsql_coerce.c
rohit01010: Doesn't the same logic also holds true in case of NULL constant with other datatype. For example, what is the current behaviour of following 
```
SELECT COALESCE(CAST(NULL AS money), CAST(0 AS bit))
GO
```

Yvinayak07 (approved):

test/JDBC/expected/babel_sqlvariant_coerce_type-vu-verify.out
Mohitraj-g: Right,fixed

contrib/babelfishpg_tsql/src/pltsql_coerce.c
Mohitraj-g:
```
1> SELECT COALESCE(CAST(NULL AS money), CAST(0 AS bit))
2> go
                     
---------------------
                .0000
```

Mohitraj-g: COALESCE(CAST(NULL AS money), CAST(0 AS bit)) already returns 0.0000 (money) on both TSQL and Babelfish today, the NULL-skip works fine because PG's implicit cast from bit to money handles the coercion downstream.                                                                                                              
                                                                                                                                                                       
sqlvariant once skipped, no other type can win it back. And there's no implicit cast to sqlvariant in PG's
cast catalog, so the downstream coercion that works for money/bit doesn't apply here. That's why the special-case is needed only for sql_variant.

rohit01010 (approved):

jsudrik (approved):
Approved.

PR #4718, opened 2026-04-13
Fix for Invalid Token Stream error for simple queries in go driver

Vineetha125:
### Description
Simple queries like  `SELECT 1 AS test` were failing on go-sqlcmd v1.9.0 with `error: Invalid TDS stream: unknown token type returned: token(165)` causing the session to hang. This was because `TdsSendRowDescription()`was sending colinfo token for regular queries. The go-mssqldb driver does not implement a parser for colinfo token, so it reports the token as unknown and the TDS stream becomes desynchronized, causing the session to hang.

**Fix:**  Guarded `SendColInfoToken()` with `if (relMetaDataInfoList != NIL)` so it is only sent when table metadata exists. Also, added a no_browsetable check, when NO_BROWSETABLE is ON, the COLINFO token is sent, otherwise it is not sent.

Validated on go-sqlcmd:

```
dev-dsk-bvineets-2b-1d58b91f % sqlcmd-go -S localhost -U jdbc_user -P 12345678
1> select 1 as test;
2> go
test       
-----------
          1
(1 row affected)
```

### Issues Resolved

BABEL-6423 [ Regression of BABEL-6207]

* **Use case based -** Tested on different clients such as dotnet, python and JDBC

* **Arbitrary inputs -** N/A

* **Performance tests -** N/A

* **Tooling impact -** N/A

* **Client tests -** N/A

### Check List


contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
```diff
@@ -1640,7 +1640,7 @@ PrepareRowDescription(TupleDesc typeinfo, PlannedStmt *plannedstmt, List *target
 					if (physical_schema_name != NULL)
 						relMetaDataInfo->partName[1] = pstrdup(physical_schema_name);
 					else
-						relMetaDataInfo->partName[1] = NULL;
+						relMetaDataInfo->partName[1] = pstrdup("");
```
tanscorpio7: Lets still add the no_browsetable check since SQL Server documents it. Does it break any test cases.

tanscorpio7 (changes requested):

contrib/babelfishpg_tds/src/backend/tds/tdsresponse.c
Vineetha125: Your correct, will add a check. The change was made because `FillTabNameWithNumParts()`, calls strlen() on each part name entry. When schema name cannot be resolved ( physical_schema_name is NULL) , `partName[1]` remains `NULL`, and strlen(NULL) causes a SEGFAULT. This was observed with temporary tables.

tanscorpio7 (dismissed):

tanscorpio7 (approved):

PR #4782, opened 2026-04-28
[BABEL] Insert Exec PR2 temp table

monk0707:
### Description

**Background: INSERT...EXEC Redesign**

The current INSERT...EXEC implementation has limitations with transaction handling, error recovery, and TRY-CATCH compatibility. We're redesigning it using a **temp table buffering approach**:
1. Procedure output is captured into a temporary buffer table via a custom DestReceiver
2. After procedure execution completes, buffered rows are flushed to the target table
3. This enables proper TRY-CATCH error handling and transaction semantics matching SQL Server

**PR Series Overview:**
| PR | Focus | Description |
|----|-------|-------------|
| PR0 | GUC | Feature flag to gate the new code path |
| PR1 | Parser | Parse INSERT...EXEC and extract target table info |
| **PR2** | **Temp Table** | **Temp table creation, drop and flush** |
| PR3 | DestReceiver | Capture procedure output to temp table |
| PR4 | Integration | Wire everything together with tests |

**PR Chain:** [← PR1](https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/4781) | [PR3 →](https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/4783)

**This PR:**

Adds the core state management infrastructure for INSERT EXEC. A new `pl_insert_exec.c` file provides creation of temp table, drop temp table and flush temp table.

**Changes:**
- New `pl_insert_exec.c` file containing:
  - Temp table creation logic.
  - Temp table drop logic.
  - Temp table flush logic.
- Function declarations added to `pltsql.h`
- Build configuration updated in `Makefile`

### Issues Resolved
PR2 of Task: [BABEL-5522](https://jira.rds.a2z.com/browse/BABEL-5522)

### Test Scenarios Covered
* **Use case based -** No new tests in this PR; existing tests pass (GUC is off by default, so legacy path is used)
* **Boundary conditions -** N/A (infrastructure PR)
* **Arbitrary inputs -** N/A
* **Negative test cases -** N/A (tests will be added in PR6 when feature is complete)
* **Minor version upgrade tests -** N/A
* **Major version upgrade tests -** N/A
* **Performance tests -** N/A
* **Tooling impact -** None
* **Client tests -** N/A

### Check List
- [x] Commits are signed per the DCO using --signoff

[...]

PR #4783, opened 2026-04-28
[BABEL] Insert Exec Redesign PR3 destreceiver

monk0707:
### Description

**Background: INSERT...EXEC Redesign**

The current INSERT...EXEC implementation has limitations with transaction handling, error recovery, and TRY-CATCH compatibility. We're redesigning it using a **temp table buffering approach**:
1. Procedure output is captured into a temporary buffer table via a custom DestReceiver
2. After procedure execution completes, buffered rows are flushed to the target table
3. This enables proper TRY-CATCH error handling and transaction semantics matching SQL Server

**PR Series Overview:**
| PR | Focus | Description |
|----|-------|-------------|
| PR0 | GUC | Feature flag to gate the new code path |
| PR1 | Parser | Parse INSERT...EXEC and extract target table info |
| PR2 | Executor | Core state management and schema detection |
| **PR3** | **DestReceiver** | **Capture procedure output to temp table** |
| PR4 | Integration | Wire everything together with tests |

**PR Chain:** [← PR2](https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/4782)

**This PR:**

Implements a custom DestReceiver that captures procedure output tuples and inserts them into a session-local temp table for later flush to the target table. 

**Changes:**
- Add `DR_insertexec` struct with coercion infrastructure (ExprContext, ProjectionInfo)
- Implement `insertexec_startup()` - validates column count matches temp table, builds type coercion expressions
- Implement `insertexec_receive()` - receives tuples and inserts into temp table with optional type coercion
- Implement `insertexec_shutdown()` / `insertexec_destroy()` - cleanup resources

**Design Notes:**
- Column count validation happens at startup (before any rows are processed) to match SQL Server behavior
- Type coercion uses PostgreSQL's `ExecProject` for implicit type conversion
- Per-tuple table open/close is intentional for subtransaction safety (TRY-CATCH rollback)

### Issues Resolved
PR3 of Task: [BABEL-5522](https://jira.rds.a2z.com/browse/BABEL-5522)

### Test Scenarios Covered
* **Use case based -** Tested via full feature integration in PR6
* **Boundary conditions -** Column count validation at startup
* **Arbitrary inputs -** N/A
* **Negative test cases -** Column mismatch raises error before any rows are inserted
* **Performance tests -** N/A
* **Tooling impact -** N/A
* **Client tests -** N/A

### Check List

### DIFF
--- a/contrib/babelfishpg_tsql/src/pl_insert_exec.c
+++ b/contrib/babelfishpg_tsql/src/pl_insert_exec.c
@@ -16,23 +16,20 @@
 #include "catalog/namespace.h"
 #include "catalog/pg_attribute.h"
 #include "commands/defrem.h"
+#include "executor/executor.h"
+#include "executor/tuptable.h"
 #include "miscadmin.h"
 #include "nodes/makefuncs.h"
-#include "executor/tuptable.h"
 #include "nodes/parsenodes.h"
 #include "parser/parse_coerce.h"
+#include "parser/scansup.h"
 #include "tcop/dest.h"
 #include "utils/builtins.h"
-
-#include "executor/executor.h"
-#include "parser/scansup.h"
+#include "utils/lsyscache.h"
 
 #include "catalog.h"
 #include "multidb.h"
 #include "session.h"
-
-#include "utils/builtins.h"
-#include "utils/lsyscache.h"
 
 extern int execute_batch(PLtsql_execstate *estate, char *batch, InlineCodeBlockArgs *args, List *params);
 
@@ -43,7 +40,6 @@ typedef struct
 {
 	DestReceiver pub;			/* public fields */
 	Relation	temp_rel;		/* open relation, closed in cleanup */
-	TupleDesc	typeinfo;		/* tuple descriptor from startup */
 	/* Projection infrastructure — coerce_to_target_type is a no-op when types already match */
 	ExprContext *econtext;		/* expression context for projection */
 	ProjectionInfo *proj_info;	/* projection info for coercion */
@@ -329,13 +325,13 @@ insertexec_startup(DestReceiver *self, int operation, TupleDesc typeinfo)
 	int			i;
 	List	   *target_list = NIL;
 
-	/* store the tuple descriptor for later use */
-	myState->typeinfo = typeinfo;
-
 	/*
 	 * Validate column count
 	 */
 	result_natts = typeinfo->natts;
+
+	if (!OidIsValid(insert_exec_ctx.temp_table_oid))
+		elog(ERROR, "INSERT EXEC failed due to missing temp table OID");
 
 	/* Open temp table to get its tuple descriptor */
 	myState->temp_rel = table_open(insert_exec_ctx.temp_table_oid, RowExclusiveLock);
@@ -420,6 +416,10 @@ insertexec_receive(TupleTableSlot *slot, DestReceiver *self)
 {
 	DR_insertexec *myState = (DR_insertexec *) self;
 	TupleTableSlot *insert_slot;
+
+	Assert(myState->temp_rel != NULL);
+	Assert(myState->proj_info != NULL);
+	Assert(myState->econtext != NULL);
 
 	/* Reset per-tuple memory context for expression evaluation */
 	ResetExprContext(myState->econtext);
--- a/contrib/babelfishpg_tsql/src/pl_insert_exec.c
+++ b/contrib/babelfishpg_tsql/src/pl_insert_exec.c
@@ -2,7 +2,8 @@
  * pl_insert_exec.c
  *	INSERT EXECUTE implementation for Babelfish PL/tsql
  *	It implements:
- *		- Temp table lifecycle: creation, flush, and drop
+ *	- Temp table lifecycle: creation, flush, and drop
+ *	- Dest Receiver callback functions implementation
  *-------------------------------------------------------------------------
  */
 
@@ -11,22 +12,47 @@
 #include "pltsql-2.h"
 
 #include "access/table.h"
-#include "access/xact.h"
 #include "catalog/namespace.h"
 #include "catalog/pg_attribute.h"
-#include "commands/defrem.h"
-#include "miscadmin.h"
 #include "nodes/makefuncs.h"
+#include "executor/tuptable.h"
+#include "nodes/parsenodes.h"
+#include "parser/parse_coerce.h"
+#include "tcop/dest.h"
+#include "utils/builtins.h"
+
+#include "executor/executor.h"
 #include "parser/scansup.h"
 
 #include "catalog.h"
 #include "multidb.h"
 #include "session.h"
 
-#include "utils/builtins.h"
 #include "utils/lsyscache.h"
 
 extern int execute_batch(PLtsql_execstate *estate, char *batch, InlineCodeBlockArgs *args, List *params);
+
+/*
+ * DestReceiver struct for INSERT EXEC - captures procedure output into a temp table.
+ */
+typedef struct
+{
+	DestReceiver pub;			/* public fields */
+	Oid			temp_table_oid;	/* OID of temp table to insert into */
+	Relation	temp_rel;		/* open relation, closed in cleanup */
+	TupleDesc	typeinfo;		/* tuple descriptor from startup */
+	/* Coercion infrastructure using ExecProject */
+	ExprContext *econtext;		/* expression context for projection */
+	ProjectionInfo *proj_info;	/* projection info for type coercion (NULL if no coercion needed) */
+	TupleTableSlot *proj_slot;	/* result slot for projection */
+	bool		needs_coercion;	/* true if any column needs coercion */
+} DR_insertexec;
+
+/* Forward declarations for DestReceiver callbacks */
+static void insertexec_startup(DestReceiver *self, int operation, TupleDesc typeinfo);
+static bool insertexec_receive(TupleTableSlot *slot, DestReceiver *self);
+static void insertexec_shutdown(DestReceiver *self);
+static void insertexec_destroy(DestReceiver *self);
 
 /*
  * Global INSERT EXEC context - defined in pltsql.h, declared here.
@@ -266,4 +292,221 @@ flush_insert_exec_temp_table(PLtsql_execstate *estate,
 	/* Route through execute_batch to handle triggers and errors properly. */
 	execute_batch(estate, flush_query.data, NULL, NULL);
 	pfree(flush_query.data);
+}
+
+/*
+ * Create a DestReceiver for INSERT EXEC that writes to a temp table.
+ */
+DestReceiver *
+CreateInsertExecDestReceiver(Oid temp_table_oid)
+{
+	DR_insertexec *self = (DR_insertexec *) palloc0(sizeof(DR_insertexec));
+
+	self->pub.receiveSlot = insertexec_receive;
+	self->pub.rStartup = insertexec_startup;
+	self->pub.rShutdown = insertexec_shutdown;
+	self->pub.rDestroy = insertexec_destroy;
+	self->temp_table_oid = temp_table_oid;
+
+	return (DestReceiver *) self;
+}
+
+/*
+ * insertexec_startup - executor startup for INSERT EXEC receiver
+ *
+ * Validates column count and builds coercion expressions for type 
+ * conversion using PostgreSQL's ExecBuildProjectionInfo/ExecProject.
+ */
+static void
+insertexec_startup(DestReceiver *self, int operation, TupleDesc typeinfo)
+{
+	DR_insertexec *myState = (DR_insertexec *) self;
+	TupleDesc	temp_tupdesc;
+	int			result_natts;
+	int			temp_natts;
+	int			i;
+	List	   *target_list = NIL;
+
+	/* store the tuple descriptor for later use */
+	myState->typeinfo = typeinfo;
+
+	/*
+	 * Validate column count
+	 */
+	result_natts = typeinfo->natts;
+
+	/* Open temp table to get its tuple descriptor */
+	myState->temp_rel = table_open(myState->temp_table_oid, RowExclusiveLock);
+	temp_tupdesc = RelationGetDescr(myState->temp_rel);
+	temp_natts = temp_tupdesc->natts;
+
+	if (result_natts != temp_natts)
+	{
+		table_close(myState->temp_rel, RowExclusiveLock);
+		myState->temp_rel = NULL;
+		ereport(ERROR,
+				(errcode(ERRCODE_DATATYPE_MISMATCH),
+				 errmsg("column name or number of supplied values does not match table definition")));
+	}
+
+	myState->needs_coercion = false;
+
+	for (i = 0; i < temp_natts; i++)
+	{
+		Form_pg_attribute src_att = TupleDescAttr(typeinfo, i);
+		Form_pg_attribute tgt_att = TupleDescAttr(temp_tupdesc, i);
+		Var		   *var;
+		Node	   *expr;
+		TargetEntry *tle;
+
+		/* Create Var node referencing input tuple column (OUTER_VAR for ecxt_outertuple) */
+		var = makeVar(OUTER_VAR,
+					  i + 1,				/* attnum is 1-based */
+					  src_att->atttypid,
+					  src_att->atttypmod,
+					  src_att->attcollation,
+					  0);					/* varlevelsup */
+
+		expr = (Node *) var;
+
+		/* Check if coercion is needed for this column */
+		if (src_att->atttypid != tgt_att->atttypid ||
+			(src_att->atttypmod != tgt_att->atttypmod && tgt_att->atttypmod != -1))
+		{
+			Node	   *cast_expr;
+
+			myState->needs_coercion = true;
+
+			/* Build coercion expression using ASSIGNMENT coercion */
+			cast_expr = coerce_to_target_type(NULL,
+											  expr,
+											  src_att->atttypid,
+											  tgt_att->atttypid,
+											  tgt_att->atttypmod,
+											  COERCION_ASSIGNMENT,
+											  COERCE_IMPLICIT_CAST,
+											  -1);
+
+			if (cast_expr == NULL)
+			{
+				table_close(myState->temp_rel, RowExclusiveLock);
+				myState->temp_rel = NULL;
+				ereport(ERROR,
+						(errcode(ERRCODE_CANNOT_COERCE),
+						 errmsg("cannot convert type %s to %s",
+								format_type_be(src_att->atttypid),
+								format_type_be(tgt_att->atttypid))));
+			}
+
+			expr = cast_expr;
+		}
+
+		/* Create TargetEntry for this column */
+		tle = makeTargetEntry((Expr *) expr,
+							  i + 1,		/* resno is 1-based */
+							  NULL,			/* resname */
+							  false);		/* resjunk */
+		target_list = lappend(target_list, tle);
+	}
+
+	/* Create expression context and projection info if coercion needed */
+	if (myState->needs_coercion)
+	{
+		TupleDesc	proj_tupdesc;
+
+		myState->econtext = CreateStandaloneExprContext();
+
+		/* Create a copy of the temp table's tuple descriptor for the result slot */
+		proj_tupdesc = CreateTupleDescCopy(temp_tupdesc);
+
+		/* Create the result slot for projection */
+		myState->proj_slot = MakeSingleTupleTableSlot(proj_tupdesc, &TTSOpsVirtual);
+
+		/* Build the projection info */
+		myState->proj_info = ExecBuildProjectionInfo(target_list,
+													 myState->econtext,
+													 myState->proj_slot,
+													 NULL,		/* no parent PlanState */
+													 typeinfo);	/* input descriptor */
+	}
+
+	/* temp_rel stays open - closed in cleanup function */
+}
+
+/*
+ * insertexec_receive - receive each tuple and insert into temp table
+ */
+static bool
+insertexec_receive(TupleTableSlot *slot, DestReceiver *self)
+{
+	DR_insertexec *myState = (DR_insertexec *) self;
+	CommandId	cid;
+	TupleTableSlot *insert_slot;
+
+	cid = GetCurrentCommandId(true);
+
+	if (myState->needs_coercion)
+	{
+		ExprContext *econtext = myState->econtext;
+
+		/* Reset per-tuple memory context for expression evaluation */
+		ResetExprContext(econtext);
+
+		/*
+		 * Set up the input slot for projection.
+		 */
+		econtext->ecxt_outertuple = slot;
+
+		/* Project tuple with type coercion */
+		insert_slot = ExecProject(myState->proj_info);
+
+		/* Insert the projected tuple */
+		table_tuple_insert(myState->temp_rel, insert_slot, cid, 0, NULL);
+	}
+	else
+		/* No coercion needed - insert directly */
+		table_tuple_insert(myState->temp_rel, slot, cid, 0, NULL);
+
+	return true;
+}
+
+/*
+ * insertexec_shutdown - executor end for INSERT EXEC receiver
+ *
+ * Clean up the expression context and projection slot used for type coercion.
+ */
+static void
+insertexec_shutdown(DestReceiver *self)
+{
+	DR_insertexec *myState = (DR_insertexec *) self;
+
+	/* Close the temp table held since startup */
+	if (myState->temp_rel != NULL)
+	{
+		table_close(myState->temp_rel, RowExclusiveLock);
+		myState->temp_rel = NULL;
+	}
+
+	/* Free the projection slot if we created one */
+	if (myState->proj_slot != NULL)
+	{
+		ExecDropSingleTupleTableSlot(myState->proj_slot);
+		myState->proj_slot = NULL;
+	}
+
+	/* Free the expression context if we created one */
+	if (myState->econtext != NULL)
+	{
+		FreeExprContext(myState->econtext, true);
+		myState->econtext = NULL;
+	}
+}
+
+/*
+ * insertexec_destroy - release DestReceiver object
+ */
+static void
+insertexec_destroy(DestReceiver *self)
+{
+	pfree(self);
 }
--- a/contrib/babelfishpg_tsql/src/pl_insert_exec.c
+++ b/contrib/babelfishpg_tsql/src/pl_insert_exec.c
@@ -12,8 +12,11 @@
 #include "pltsql-2.h"
 
 #include "access/table.h"
+#include "access/xact.h"
 #include "catalog/namespace.h"
 #include "catalog/pg_attribute.h"
+#include "commands/defrem.h"
+#include "miscadmin.h"
 #include "nodes/makefuncs.h"
 #include "executor/tuptable.h"
 #include "nodes/parsenodes.h"
@@ -28,6 +31,7 @@
 #include "multidb.h"
 #include "session.h"
 
+#include "utils/builtins.h"
 #include "utils/lsyscache.h"
 
 extern int execute_batch(PLtsql_execstate *estate, char *batch, InlineCodeBlockArgs *args, List *params);
@@ -38,14 +42,12 @@ extern int execute_batch(PLtsql_execstate *estate, char *batch, InlineCodeBlockA
 typedef struct
 {
 	DestReceiver pub;			/* public fields */
-	Oid			temp_table_oid;	/* OID of temp table to insert into */
 	Relation	temp_rel;		/* open relation, closed in cleanup */
 	TupleDesc	typeinfo;		/* tuple descriptor from startup */
-	/* Coercion infrastructure using ExecProject */
+	/* Projection infrastructure — coerce_to_target_type is a no-op when types already match */
 	ExprContext *econtext;		/* expression context for projection */
-	ProjectionInfo *proj_info;	/* projection info for type coercion (NULL if no coercion needed) */
+	ProjectionInfo *proj_info;	/* projection info for coercion */
 	TupleTableSlot *proj_slot;	/* result slot for projection */
-	bool		needs_coercion;	/* true if any column needs coercion */
 } DR_insertexec;
 
 /* Forward declarations for DestReceiver callbacks */
@@ -298,15 +300,14 @@ flush_insert_exec_temp_table(PLtsql_execstate *estate,
  * Create a DestReceiver for INSERT EXEC that writes to a temp table.
  */
 DestReceiver *
-CreateInsertExecDestReceiver(Oid temp_table_oid)
+CreateInsertExecDestReceiver(void)
 {
-	DR_insertexec *self = (DR_insertexec *) palloc0(sizeof(DR_insertexec));
+	DR_insertexec *self = palloc0_object(DR_insertexec);
 
 	self->pub.receiveSlot = insertexec_receive;
 	self->pub.rStartup = insertexec_startup;
 	self->pub.rShutdown = insertexec_shutdown;
 	self->pub.rDestroy = insertexec_destroy;
-	self->temp_table_oid = temp_table_oid;
 
 	return (DestReceiver *) self;
 }
@@ -336,20 +337,15 @@ insertexec_startup(DestReceiver *self, int operation, TupleDesc typeinfo)
 	result_natts = typeinfo->natts;
 
 	/* Open temp table to get its tuple descriptor */
-	myState->temp_rel = table_open(myState->temp_table_oid, RowExclusiveLock);
+	myState->temp_rel = table_open(insert_exec_ctx.temp_table_oid, RowExclusiveLock);
 	temp_tupdesc = RelationGetDescr(myState->temp_rel);
 	temp_natts = temp_tupdesc->natts;
 
 	if (result_natts != temp_natts)
-	{
-		table_close(myState->temp_rel, RowExclusiveLock);
-		myState->temp_rel = NULL;
 		ereport(ERROR,
 				(errcode(ERRCODE_DATATYPE_MISMATCH),
 				 errmsg("column name or number of supplied values does not match table definition")));
-	}
 
-	myState->needs_coercion = false;
 
 	for (i = 0; i < temp_natts; i++)
 	{
@@ -367,39 +363,26 @@ insertexec_startup(DestReceiver *self, int operation, TupleDesc typeinfo)
 					  src_att->attcollation,
 					  0);					/* varlevelsup */
 
-		expr = (Node *) var;
+		/*
+		 * coerce_to_target_type returns the Var unchanged when types match,
+		 * so we call it unconditionally.
+		 */
+		expr = coerce_to_target_type(NULL,
+									 (Node *) var,
+									 src_att->atttypid,
+									 tgt_att->atttypid,
+									 tgt_att->atttypmod,
+									 COERCION_ASSIGNMENT,
+									 COERCE_IMPLICIT_CAST,
+									 -1);
 
-		/* Check if coercion is needed for this column */
-		if (src_att->atttypid != tgt_att->atttypid ||
-			(src_att->atttypmod != tgt_att->atttypmod && tgt_att->atttypmod != -1))
-		{
-			Node	   *cast_expr;
+		if (expr == NULL)
+			ereport(ERROR,
+					(errcode(ERRCODE_CANNOT_COERCE),
+					 errmsg("cannot convert type %s to %s",
+							format_type_be(src_att->atttypid),
+							format_type_be(tgt_att->atttypid))));
 
-			myState->needs_coercion = true;
-
-			/* Build coercion expression using ASSIGNMENT coercion */
-			cast_expr = coerce_to_target_type(NULL,
-											  expr,
-											  src_att->atttypid,
-											  tgt_att->atttypid,
-											  tgt_att->atttypmod,
-											  COERCION_ASSIGNMENT,
-											  COERCE_IMPLICIT_CAST,
-											  -1);
-
-			if (cast_expr == NULL)
-			{
-				table_close(myState->temp_rel, RowExclusiveLock);
-				myState->temp_rel = NULL;
-				ereport(ERROR,
-						(errcode(ERRCODE_CANNOT_COERCE),
-						 errmsg("cannot convert type %s to %s",
-								format_type_be(src_att->atttypid),
-								format_type_be(tgt_att->atttypid))));
-			}
-
-			expr = cast_expr;
-		}
 
 		/* Create TargetEntry for this column */
 		tle = makeTargetEntry((Expr *) expr,
@@ -409,26 +392,18 @@ insertexec_startup(DestReceiver *self, int operation, TupleDesc typeinfo)
 		target_list = lappend(target_list, tle);
 	}
 
-	/* Create expression context and projection info if coercion needed */
-	if (myState->needs_coercion)
-	{
-		TupleDesc	proj_tupdesc;
+	/* Create expression context and projection info */
+	myState->econtext = CreateStandaloneExprContext();
 
-		myState->econtext = CreateStandaloneExprContext();
+	/* Pass temp_tupdesc directly - valid while temp_rel is held open */
+	myState->proj_slot = MakeSingleTupleTableSlot(temp_tupdesc, &TTSOpsVirtual);
 
-		/* Create a copy of the temp table's tuple descriptor for the result slot */
-		proj_tupdesc = CreateTupleDescCopy(temp_tupdesc);
-
-		/* Create the result slot for projection */
-		myState->proj_slot = MakeSingleTupleTableSlot(proj_tupdesc, &TTSOpsVirtual);
-
-		/* Build the projection info */
-		myState->proj_info = ExecBuildProjectionInfo(target_list,
-													 myState->econtext,
-													 myState->proj_slot,
-													 NULL,		/* no parent PlanState */
-													 typeinfo);	/* input descriptor */
-	}
+	/* Build the projection info */
+	myState->proj_info = ExecBuildProjectionInfo(target_list,
+												 myState->econtext,
+												 myState->proj_slot,
+												 NULL,		/* no parent PlanState */
+												 typeinfo);	/* input descriptor */
 
 	/* temp_rel stays open - closed in cleanup function */
 }
@@ -445,27 +420,17 @@ insertexec_receive(TupleTableSlot *slot, DestReceiver *self)
 
 	cid = GetCurrentCommandId(true);
 
-	if (myState->needs_coercion)
-	{
-		ExprContext *econtext = myState->econtext;
+	/* Reset per-tuple memory context for expression evaluation */
+	ResetExprContext(myState->econtext);
 
-		/* Reset per-tuple memory context for expression evaluation */
-		ResetExprContext(econtext);
+	/* Set up the input slot for projection */
+	myState->econtext->ecxt_outertuple = slot;
 
-		/*
-		 * Set up the input slot for projection.
-		 */
-		econtext->ecxt_outertuple = slot;
+	/* Project tuple with type coercion */
+	insert_slot = ExecProject(myState->proj_info);
 
-		/* Project tuple with type coercion */
-		insert_slot = ExecProject(myState->proj_info);
-
-		/* Insert the projected tuple */
-		table_tuple_insert(myState->temp_rel, insert_slot, cid, 0, NULL);
-	}
-	else
-		/* No coercion needed - insert directly */
-		table_tuple_insert(myState->temp_rel, slot, cid, 0, NULL);
+	/* Insert the projected tuple */
+	table_tuple_insert(myState->temp_rel, insert_slot, cid, 0, NULL);
 
 	return true;
 }
@@ -480,26 +445,17 @@ insertexec_shutdown(DestReceiver *self)
 {
 	DR_insertexec *myState = (DR_insertexec *) self;
 
-	/* Close the temp table held since startup */
-	if (myState->temp_rel != NULL)
-	{
-		table_close(myState->temp_rel, RowExclusiveLock);
-		myState->temp_rel = NULL;
-	}
+	Assert(myState->temp_rel != NULL);
+	table_close(myState->temp_rel, RowExclusiveLock);
+	myState->temp_rel = NULL;
 
-	/* Free the projection slot if we created one */
-	if (myState->proj_slot != NULL)
-	{
-		ExecDropSingleTupleTableSlot(myState->proj_slot);
-		myState->proj_slot = NULL;
-	}
+	Assert(myState->proj_slot != NULL);
+	ExecDropSingleTupleTableSlot(myState->proj_slot);
+	myState->proj_slot = NULL;
 
-	/* Free the expression context if we created one */
-	if (myState->econtext != NULL)
-	{
-		FreeExprContext(myState->econtext, true);
-		myState->econtext = NULL;
-	}
+	Assert(myState->econtext != NULL);
+	FreeExprContext(myState->econtext, true);
+	myState->econtext = NULL;
 }
 
 /*
--- a/contrib/babelfishpg_tsql/src/pl_insert_exec.c
+++ b/contrib/babelfishpg_tsql/src/pl_insert_exec.c
@@ -48,6 +48,7 @@ typedef struct
 	ExprContext *econtext;		/* expression context for projection */
 	ProjectionInfo *proj_info;	/* projection info for coercion */
 	TupleTableSlot *proj_slot;	/* result slot for projection */
+	CommandId	cid;			/* command ID obtained once in startup, shared across all tuples */
 } DR_insertexec;
 
 /* Forward declarations for DestReceiver callbacks */
@@ -405,6 +406,9 @@ insertexec_startup(DestReceiver *self, int operation, TupleDesc typeinfo)
 												 NULL,		/* no parent PlanState */
 												 typeinfo);	/* input descriptor */
 
+	/* Obtain command ID once - all tuples share the same cid for MVCC consistency */
+	myState->cid = GetCurrentCommandId(true);
+
 	/* temp_rel stays open - closed in cleanup function */
 }
 
@@ -415,10 +419,7 @@ static bool
 insertexec_receive(TupleTableSlot *slot, DestReceiver *self)
 {
 	DR_insertexec *myState = (DR_insertexec *) self;
-	CommandId	cid;
 	TupleTableSlot *insert_slot;
-
-	cid = GetCurrentCommandId(true);
 
 	/* Reset per-tuple memory context for expression evaluation */
 	ResetExprContext(myState->econtext);
@@ -430,7 +431,7 @@ insertexec_receive(TupleTableSlot *slot, DestReceiver *self)
 	insert_slot = ExecProject(myState->proj_info);
 
 	/* Insert the projected tuple */
-	table_tuple_insert(myState->temp_rel, insert_slot, cid, 0, NULL);
+	table_tuple_insert(myState->temp_rel, insert_slot, myState->cid, 0, NULL);
 
 	return true;
 }
--- a/contrib/babelfishpg_tsql/src/pltsql.h
+++ b/contrib/babelfishpg_tsql/src/pltsql.h
@@ -2534,7 +2534,7 @@ extern InsertExecContext insert_exec_ctx;
 extern Oid create_insert_exec_temp_table(const char *target_table, const char *column_list, const char *schema_name_in);
 extern void flush_insert_exec_temp_table(PLtsql_execstate *estate,
 										 const char *column_list);
-extern DestReceiver *CreateInsertExecDestReceiver(Oid temp_table_oid);
+extern DestReceiver *CreateInsertExecDestReceiver(void);
 
 #define NUM_DB_OBJECTS 11
 
--- a/contrib/babelfishpg_tsql/src/pltsql.h
+++ b/contrib/babelfishpg_tsql/src/pltsql.h
@@ -2534,6 +2534,7 @@ extern InsertExecContext insert_exec_ctx;
 extern Oid create_insert_exec_temp_table(const char *target_table, const char *column_list, const char *schema_name_in);
 extern void flush_insert_exec_temp_table(PLtsql_execstate *estate,
 										 const char *column_list);
+extern DestReceiver *CreateInsertExecDestReceiver(Oid temp_table_oid);
 
 #define NUM_DB_OBJECTS 11
 


kuntalghosh (changes requested):
# AI Code Review

Verdict: REQUEST_CHANGES

**Issue 1 - In `pl_insert_exec.c` at line ~330: Unquoted identifier in SPI_execute DROP TABLE:**

```c
appendStringInfo(&drop_stmt, "DROP TABLE IF EXISTS %s", insert_exec_ctx.temp_table_name);
rc = SPI_execute(drop_stmt.data, false, 0);
```

The temp table name is interpolated directly into SQL without `quote_identifier()`. While the name is internally generated (stub in PR4), this is a defense-in-depth gap. If a future PR generates a name with special characters, this becomes an injection vector.

Suggested fix:

```c
appendStringInfo(&drop_stmt, "DROP TABLE IF EXISTS %s",
                 quote_identifier(insert_exec_ctx.temp_table_name));
```

**Issue 2 - In `pl_insert_exec.c` at line ~370-430: Lock leak on partial failure in pltsql_insert_exec_open_target_table:**

```c
LockRelationOid(relid, RowExclusiveLock);  /* succeeds */
...
PG_TRY();
{
    rel = table_open(relid, NoLock);
}
PG_CATCH();
{
    FlushErrorState();
    insert_exec_ctx.target_rel_oid = InvalidOid;
    return;  /* lock on relid is NEVER released */
}
```

If `LockRelationOid` succeeds but the subsequent `table_open` fails, the lock is acquired but `target_rel_oid` is set to `InvalidOid`, so `pltsql_insert_exec_close_target_table` won't release it. The lock is held until transaction end, causing unexpected lock contention.

Suggested fix:

```c
PG_CATCH();
{
    FlushErrorState();
    UnlockRelationOid(relid, RowExclusiveLock);
    insert_exec_ctx.target_rel_oid = InvalidOid;
    return;
}
```

**Issue 3 - In `pl_insert_exec.c` at line ~830-860: Per-tuple table_open/table_close in insertexec_receive accumulates lock references:**

```c
static bool
insertexec_receive(TupleTableSlot *slot, DestReceiver *self)
{
    ...
    temp_rel = table_open(myState->temp_table_oid, RowExclusiveLock);
    ...
    table_close(temp_rel, NoLock);
    ...
}
```

Each call acquires a new lock reference (incrementing the lock count) but never releases it. For large result sets (millions of rows), this accumulates millions of lock reference entries in the lock manager's local hash table. The code comment acknowledges this but dismisses it as "fine for temp tables." While there's no shared-memory contention, the local bookkeeping overhead is non-trivial. Consider opening the relation once in `insertexec_startup` and closing in `insertexec_shutdown`, or at minimum releasing the lock on close: `table_close(temp_rel, RowExclusiveLock)`.

Suggested fix:

```c
/* In insertexec_startup, after validation: */
myState->temp_rel = table_open(myState->temp_table_oid, RowExclusiveLock);

/* In insertexec_receive: */
table_tuple_insert(myState->temp_rel, insert_slot, cid, 0, NULL);

/* In insertexec_shutdown: */
if (myState->temp_rel != NULL)
{
    table_close(myState->temp_rel, NoLock);
    myState->temp_rel = NULL;
}
```

**Issue 4 - In `guc.c` at line ~954: GUC is PGC_USERSET — any user can enable incomplete feature:**

```c
DefineCustomBoolVariable("babelfishpg_tsql.enable_new_insert_exec",
                         gettext_noop("Enables INSERT...EXEC redesign code path"),
                         NULL,
                         &pltsql_enable_new_insert_exec,
                         false,
                         PGC_USERSET,
                         GUC_NOT_IN_SAMPLE,
                         NULL, NULL, NULL);
```

`PGC_USERSET` means any connected user can `SET babelfishpg_tsql.enable_new_insert_exec = true` to activate this new code path. This is PR 3 of 7 with multiple stub functions (`create_insert_exec_temp_table` returns `InvalidOid`, `insert_exec_setup` returns false, etc.). Enabling this GUC with stubs active could lead to unexpected behavior or silent data loss.

Suggested fix:

```c
PGC_SUSET,  /* superuser-only during development; relax to PGC_USERSET when feature is complete */
```

**Issue 5 - In `pl_insert_exec.c` at line ~780: insertexec_startup comment has typo — "matches) tsql":**

```c
/* Build coercion expression using ASSIGNMENT coercion (matches) tsql INSERT behavior) */
```

There's a mismatched parenthesis in the comment: `(matches)` should be `(matches`.

Suggested fix:

```c
/* Build coercion expression using ASSIGNMENT coercion (matches tsql INSERT behavior) */
```

**Issue 6 - In `pl_insert_exec.c` at line ~870-880: insertexec_destroy doesn't guard against missing shutdown:**

```c
static void
insertexec_destroy(DestReceiver *self)
{
    DR_insertexec *myState = (DR_insertexec *) self;
    myState->proj_info = NULL;
    pfree(self);
}
```

If `insertexec_destroy` is called without a prior `insertexec_shutdown` (e.g., due to an error path), the `proj_slot` and `econtext` are leaked. The PostgreSQL executor protocol guarantees shutdown before destroy, but a defensive check would be safer.

Suggested fix:

```c
static void
insertexec_destroy(DestReceiver *self)
{
    DR_insertexec *myState = (DR_insertexec *) self;

    /* Defensive: ensure resources are freed even if shutdown was skipped */
    if (myState->proj_slot != NULL || myState->econtext != NULL)
        insertexec_shutdown(self);

    myState->proj_info = NULL;
    pfree(self);
}
```

**Issue 7 - In `pl_insert_exec.c`: TopMemoryContext allocations without bounded lifetime:**

Multiple strings and structs are allocated in `TopMemoryContext` across `pltsql_set_insert_exec_context_info()`, `pltsql_insert_exec_open_target_table()`, and `pltsql_insert_exec_set_target_table()`. While the code does free previous values before re-allocating and has cleanup functions, if any cleanup path is missed (e.g., unexpected `elog(FATAL)`), these allocations leak permanently for the backend session. Consider creating a dedicated child `MemoryContext` under `TopMemoryContext` (e.g., `InsertExecMemoryContext`) that can be reset as a unit. This is advisory — the current approach works but is fragile.

**Issue 8 - In `pl_insert_exec.c` at line ~700: CreateInsertExecDestReceiver uses DestNone for mydest:**

```c
self->pub.mydest = DestNone;  /* no standard dest type fits; avoids miscast risks */
```

Using `DestNone` is reasonable since there's no standard `CommandDest` for this use case, but it means any code that switches on `mydest` won't recognize this receiver. The comment explains the rationale, which is good. Consider whether a custom enum value would be safer if other Babelfish code inspects `mydest`. This is advisory.

### Missing Tests

- **Blocking**: This PR adds 1436 lines of new C code (a complete DestReceiver implementation, context management, schema verification, column count validation) but includes **zero test files**. The PR description states "Tested via full feature integration in subsequent PR," but per Babelfish contribution guidelines, patches without tests will not be accepted. At minimum, the following should be tested:
  - Column count mismatch error (the `insertexec_startup` validation path)
  - Type coercion between compatible types (e.g., `int` → `bigint`)
  - Type coercion failure for incompatible types
  - Schema change detection (`pltsql_insert_exec_verify_schema`)
  - Context lifecycle (set → active → clear → verify inactive)
  - The GUC toggle behavior (legacy vs new path)

- If tests truly cannot be added until the stubs are replaced in PR4/PR5, this should be explicitly documented with a tracking ticket, and the GUC should be `PGC_SUSET` to prevent users from activating incomplete code.

### Notes

- The PR is well-structured as part 3 of a 7-PR series. The code is well-commented with clear function documentation, defensive stale-state detection (execution_id, call stack depth), and proper separation of concerns.
- The `tsqlIface.cpp` changes correctly handle all three statement types (`PLtsql_stmt_exec`, `PLtsql_stmt_exec_batch`, `PLtsql_stmt_exec_sp`) and include proper validation for function context and OUTPUT clause.
- The `InsertExecInfo` struct in `pltsql-2.h` is a clean shared structure across the three exec statement types.
- The `replaceGraftedStatement` method in `tsqlIface.cpp` is a useful addition to the builder pattern, though it appends the new statement at the end of the siblings list rather than preserving the original position — verify this doesn't affect execution order.
- No JIRA/BABEL ticket is linked in the PR description ("Issues Resolved: N/A"). While understandable for a multi-PR feature, a tracking ticket should be referenced.

kuntalghosh (changes requested):
# AI Code Review

Verdict: REQUEST_CHANGES

**Issue 1 - In `pl_insert_exec.c` at line ~560-575: Lock leak when `table_open` fails after `LockRelationOid` succeeds:**

```c
PG_CATCH();
{
    FlushErrorState();
    insert_exec_ctx.target_rel_oid = InvalidOid;
    return;
}
```

If `LockRelationOid` succeeds but the subsequent `table_open` fails, the `RowExclusiveLock` is acquired but `target_rel_oid` is set to `InvalidOid`. This means `pltsql_insert_exec_close_target_table()` will skip the `UnlockRelationOid` call since it checks `OidIsValid(target_rel_oid)`. The lock is held until transaction end, causing unexpected lock contention on the target table. This was raised in the prior review and remains unaddressed.

Suggested fix:

```c
PG_CATCH();
{
    FlushErrorState();
    UnlockRelationOid(relid, RowExclusiveLock);
    insert_exec_ctx.target_rel_oid = InvalidOid;
    return;
}
```

**Issue 2 - In `pl_insert_exec.c` at line ~830-860: Per-tuple `table_open`/`table_close` in `insertexec_receive` accumulates lock references:**

```c
static bool
insertexec_receive(TupleTableSlot *slot, DestReceiver *self)
{
    ...
    temp_rel = table_open(myState->temp_table_oid, RowExclusiveLock);
    ...
    table_close(temp_rel, NoLock);
    ...
}
```

Each call acquires a new `RowExclusiveLock` reference but never releases it (`NoLock` on close). For large result sets this accumulates potentially millions of lock reference entries in the lock manager's local hash table. The code comment acknowledges this but dismisses it. While there's no shared-memory contention for temp tables, the local bookkeeping overhead is non-trivial. Opening the relation once in `insertexec_startup` and closing in `insertexec_shutdown` would be cleaner and more efficient. The comment mentions subtransaction boundaries as the reason, but if that's a real concern, it should be documented with a specific scenario.

Suggested fix:

```c
/* In DR_insertexec struct, add: */
Relation    temp_rel;       /* opened in startup, closed in shutdown */

/* In insertexec_startup, after validation: */
myState->temp_rel = table_open(myState->temp_table_oid, RowExclusiveLock);

/* In insertexec_receive, replace open/close with: */
table_tuple_insert(myState->temp_rel, insert_slot, cid, 0, NULL);

/* In insertexec_shutdown, add: */
if (myState->temp_rel != NULL)
{
    table_close(myState->temp_rel, NoLock);
    myState->temp_rel = NULL;
}
```

**Issue 3 - In `tsqlIface.cpp` at line ~1648-1651: `replaceGraftedStatement` changes execution order:**

```cpp
// Remove old statement and add new one
siblings = list_delete_ptr(siblings, old_stmt);
siblings = lappend(siblings, new_stmt);
```

`list_delete_ptr` removes the old statement from its original position, then `lappend` adds the new statement at the end of the siblings list. If the INSERT EXEC statement is not the last statement in a batch (e.g., `INSERT INTO t EXEC p; SELECT 1;`), the replacement moves it after `SELECT 1`, changing execution order. This should use positional replacement to preserve the original order.

Suggested fix:

```cpp
if (old_stmt)
{
    List *siblings = getCode(container);
    ListCell *lc;
    int pos = 0;

    /* Find position of old statement */
    foreach(lc, siblings)
    {
        if (lfirst(lc) == old_stmt)
            break;
        pos++;
    }

    siblings = list_delete_ptr(siblings, old_stmt);
    siblings = list_insert_nth(siblings, pos, new_stmt);
    setCode(container, siblings);
    attachPLtsql_fragment(ctx, new_stmt);
}
```

**Issue 4 - In `pl_insert_exec.c` at line ~150-210: `parse_insert_exec_table_name` doesn't handle 4-part names (server.db.schema.table):**

```c
second_dot = strrchr(target_copy, '.');
if (second_dot != NULL)
{
    /* db.schema.table - schema is after the second dot. */
    schema_name = pstrdup(second_dot + 1);
}
```

The function uses `strrchr` twice to find dots, which correctly handles 2-part and 3-part names. However, for a 4-part name like `server.db.schema.table`, the first `strrchr` finds the last dot (before `table`), and the second `strrchr` on the truncated string finds the dot before `schema`. This means `schema_name` gets `schema` correctly, but the `db` and `server` parts are silently ignored without any warning. T-SQL 4-part names (`linked_server.db.schema.table`) are a real pattern. The function comment says "doesn't handle bracket-quoted identifiers or cross-database names" but doesn't mention 4-part names. At minimum, detect and warn about this case.

Suggested fix:

```c
if (second_dot != NULL)
{
    schema_name = pstrdup(second_dot + 1);

    /* Check for 4-part name (server.db.schema.table) */
    *second_dot = '\0';
    if (strchr(target_copy, '.') != NULL)
        elog(WARNING, "4-part names (linked server references) are not supported in INSERT EXEC: %s",
             target_table);
}
```

**Issue 5 - In `pl_insert_exec.c`: TopMemoryContext allocations without bounded lifetime:**

Multiple strings and structs are allocated in `TopMemoryContext` across `pltsql_set_insert_exec_context_info()`, `pltsql_insert_exec_open_target_table()`, and `pltsql_insert_exec_set_target_table()`. Historical bug pattern from the Babelfish knowledge base (babelfish_extensions#651) explicitly warns: "Avoid TopMemoryContext for per-statement allocations; use dedicated child context." If any cleanup path is missed (e.g., unexpected `elog(FATAL)`, missed error path in a future PR), these allocations leak permanently for the backend session. Consider creating a dedicated child `MemoryContext` (e.g., `InsertExecMemoryContext`) that can be reset as a unit. This is advisory but strongly recommended given the complexity of the error paths in this file.

### Missing Tests

- **Blocking**: This PR adds ~1500 lines of new C code (a complete DestReceiver implementation, context management, schema verification, column count validation, parser changes) but includes **zero test files**. The PR description states "Tested via full feature integration in subsequent PR," but the Babelfish contribution guidelines require tests with code changes. At minimum, the following should be testable even with stubs:
  - Column count mismatch error path (the `insertexec_startup` validation)
  - The GUC toggle behavior (verify `PGC_SUSET` prevents non-superuser from enabling)
  - The OUTPUT clause rejection in the parser (`tsqlIface.cpp`)
  - The function context rejection (`INSERT EXEC` inside a function)

  If tests truly cannot be added until stubs are replaced in PR4/PR5, this should be explicitly documented with a tracking ticket reference in the PR description.

### Notes

- Prior review issues addressed in this revision: `quote_identifier` for DROP TABLE (Issue 1), GUC changed to `PGC_SUSET` (Issue 4), comment typo fixed (Issue 5), defensive `insertexec_destroy` guard added (Issue 6), BABEL-5522 JIRA link added.
- Prior review issues still open: lock leak on `table_open` failure (Issue 2), per-tuple `table_open`/`table_close` (Issue 3), TopMemoryContext advisory (Issue 7).
- The code is well-structured with thorough comments, defensive stale-state detection (execution_id, call stack depth), and proper separation of concerns.
- The `InsertExecInfo` struct in `pltsql-2.h` correctly stores table identifier components separately to prevent SQL injection — this is a good design choice.
- The `GUC_DISALLOW_IN_FILE | GUC_DISALLOW_IN_AUTO_FILE` flags on the GUC are appropriate for a development-phase feature gate.

kuntalghosh (changes requested):
# AI Code Review

Verdict: REQUEST_CHANGES

**Issue 1 - In `pl_insert_exec.c` at line ~510-525: Lock leak when `table_open` fails after `LockRelationOid` succeeds:**

```c
PG_TRY();
{
    rel = table_open(relid, NoLock);
}
PG_CATCH();
{
    FlushErrorState();
    insert_exec_ctx.target_rel_oid = InvalidOid;
    return;
}
```

If `LockRelationOid` succeeds (line ~497) but the subsequent `table_open` fails, the `RowExclusiveLock` is acquired but `target_rel_oid` is set to `InvalidOid`. This means `pltsql_insert_exec_close_target_table()` will skip the `UnlockRelationOid` call since it checks `OidIsValid(target_rel_oid)`. The lock is held until transaction end, causing unexpected lock contention on the target table. This was raised in both prior reviews and remains unaddressed.

Suggested fix:

```c
PG_CATCH();
{
    FlushErrorState();
    UnlockRelationOid(relid, RowExclusiveLock);
    insert_exec_ctx.target_rel_oid = InvalidOid;
    return;
}
```

**Issue 2 - In `tsqlIface.cpp` at line ~1644: `replaceGraftedStatement` changes execution order:**

```cpp
// Remove old statement and add new one
siblings = list_delete_ptr(siblings, old_stmt);
siblings = lappend(siblings, new_stmt);
```

`list_delete_ptr` removes the old statement from its original position, then `lappend` adds the new statement at the end of the siblings list. If the INSERT EXEC statement is not the last statement in a batch (e.g., `INSERT INTO t EXEC p; SELECT 1;`), the replacement moves it after `SELECT 1`, changing execution order. This should use positional replacement to preserve the original order.

Suggested fix:

```cpp
if (old_stmt && container)
{
    List *siblings = getCode(container);
    ListCell *lc;
    int pos = 0;

    foreach(lc, siblings)
    {
        if (lfirst(lc) == old_stmt)
            break;
        pos++;
    }

    siblings = list_delete_ptr(siblings, old_stmt);
    siblings = list_insert_nth(siblings, pos, new_stmt);
    setCode(container, siblings);
    attachPLtsql_fragment(ctx, new_stmt);
}
```

**Issue 3 - In `pl_insert_exec.c` at line ~407: Unquoted identifier in `SPI_execute` DROP TABLE:**

```c
appendStringInfo(&drop_stmt, "DROP TABLE IF EXISTS %s", insert_exec_ctx.temp_table_name);
rc = SPI_execute(drop_stmt.data, false, 0);
```

The temp table name is interpolated directly into SQL without `quote_identifier()`. While `temp_table_name` is always NULL in this PR (the stub `create_insert_exec_temp_table` returns `InvalidOid`), PR4 will populate it. This is a defense-in-depth gap — if the generated name ever contains special characters, this becomes a post-auth SQL injection vector. The first prior review flagged this and the second review noted it was addressed, but the current diff still shows the unquoted form.

Suggested fix:

```c
appendStringInfo(&drop_stmt, "DROP TABLE IF EXISTS %s",
                 quote_identifier(insert_exec_ctx.temp_table_name));
```

**Issue 4 - In `pl_insert_exec.c` at line ~530-545: Relation leak if `palloc` throws between `table_open` and `table_close`:**

```c
rel = table_open(relid, NoLock);
...
oldcontext = MemoryContextSwitchTo(TopMemoryContext);
insert_exec_ctx.schema_sig = palloc(sizeof(InsertExecSchemaSignature));
insert_exec_ctx.schema_sig->atttypids = palloc(tupdesc->natts * sizeof(Oid));
insert_exec_ctx.schema_sig->atttypmods = palloc(tupdesc->natts * sizeof(int32));
...
table_close(rel, NoLock);
```

If any of the `palloc` calls between `table_open` and `table_close` throw an out-of-memory error, the relation handle is leaked. This should be wrapped in a `PG_TRY`/`PG_FINALLY` block to ensure `table_close` is called on error.

Suggested fix:

```c
rel = table_open(relid, NoLock);
PG_TRY();
{
    tupdesc = RelationGetDescr(rel);
    /* ... palloc and schema capture ... */
}
PG_FINALLY();
{
    table_close(rel, NoLock);
}
PG_END_TRY();
```

**Issue 5 - In `tsqlIface.cpp` at lines ~2115 and ~2126: Missing indentation on `target` assignment:**

```cpp
if (!target_table.empty())
exec_stmt->insert_exec.target = pstrdup(target_table.c_str());
```

and:

```cpp
if (!target_table.empty())
exec_batch_stmt->insert_exec.target = pstrdup(target_table.c_str());
```

The `target` assignment lines are missing indentation under their `if` statements. While this compiles correctly (single-statement body), it's inconsistent with the surrounding code where the `columns` assignment is properly indented. This makes the code harder to read and could lead to bugs if someone adds a second statement to the `if` body.

Suggested fix:

```cpp
if (!target_table.empty())
    exec_stmt->insert_exec.target = pstrdup(target_table.c_str());
```

**Issue 6 - In `pl_insert_exec.c` at line ~1067-1085: Use-after-free risk in `pltsql_insert_exec_set_target_table`:**

```c
void
pltsql_insert_exec_set_target_table(const char *target_table)
{
    if (insert_exec_ctx.target_table != NULL && target_table != insert_exec_ctx.target_table)
    {
        pfree(insert_exec_ctx.target_table);
    }
    if (target_table != NULL)
    {
        MemoryContext oldcontext = MemoryContextSwitchTo(TopMemoryContext);
        insert_exec_ctx.target_table = pstrdup(target_table);
    }
    ...
}
```

The old value is freed before `pstrdup` is called. If `pstrdup` throws an OOM error, `insert_exec_ctx.target_table` still points to the freed memory (the pointer was not NULLed before `pstrdup`). Any subsequent getter call returns a dangling pointer. The same pattern exists in `pltsql_set_insert_exec_context_info`. Use a swap-then-free pattern instead.

Suggested fix:

```c
if (target_table != NULL)
{
    char *old_target = insert_exec_ctx.target_table;
    MemoryContext oldcontext = MemoryContextSwitchTo(TopMemoryContext);
    insert_exec_ctx.target_table = pstrdup(target_table);
    if (old_target != NULL)
        pfree(old_target);
}
```

**Issue 7 - In `pl_insert_exec.c` at line ~399-413: `pltsql_insert_exec_check_pending_drop` lacks SPI connection guard:**

```c
void
pltsql_insert_exec_check_pending_drop(void)
{
    if (insert_exec_ctx.pending_drop && insert_exec_ctx.temp_table_name != NULL)
    {
        StringInfoData drop_stmt;
        initStringInfo(&drop_stmt);
        appendStringInfo(&drop_stmt, "DROP TABLE IF EXISTS %s", ...);
        rc = SPI_execute(drop_stmt.data, false, 0);
```

This function calls `SPI_execute` without verifying that an SPI connection is active. If called from an error cleanup path where SPI is not connected, this will crash the backend with a FATAL error. The function's own comment says it handles "the case where a previous INSERT EXEC failed and couldn't drop its temp table because SPI wasn't available," but it doesn't actually check SPI availability before executing.

Suggested fix:

```c
if (SPI_connect() != SPI_OK_CONNECT)
{
    elog(WARNING, "cannot drop pending INSERT EXEC temp table: SPI not available");
    return;
}
/* ... SPI_execute ... */
SPI_finish();
```

Or guard with a check that the caller is responsible for ensuring SPI is connected, and document that precondition.

### Missing Tests

- **Blocking**: This PR adds ~1500 lines of new C code (a complete DestReceiver implementation, context management, schema verification, column count validation, parser changes) but includes zero test files. The PR description states "Tested via full feature integration in subsequent PR," but the contribution guidelines require tests with code changes. Several paths are testable even with stubs:
  - The OUTPUT clause rejection in the parser (should raise syntax error)
  - The function context rejection (INSERT EXEC inside a function should error)
  - The GUC toggle behavior (verify `PGC_SUSET` prevents non-superuser from enabling)

  If tests truly cannot be added until stubs are replaced in PR4/PR5, this should be explicitly documented with a tracking ticket reference.

### Notes

- Prior review issues addressed: GUC changed to `PGC_SUSET` (Issue 4 from review 1), comment typo fixed (Issue 5 from review 1), defensive `insertexec_destroy` guard added (Issue 6 from review 1), BABEL-5522 JIRA link added.
- Prior review issues still open: lock leak on `table_open` failure (Issue 2 from both reviews), per-tuple `table_open`/`table_close` in `insertexec_receive` (Issue 3 from both reviews), `TopMemoryContext` advisory (Issue 7 from review 1, Issue 5 from review 2), `replaceGraftedStatement` execution order (Issue 3 from review 2).
- The per-tuple `table_open`/`table_close` in `insertexec_receive` (lines 1350-1390) accumulates lock references without releasing them. For large result sets this is non-trivial overhead. The prior reviews suggested opening the relation once in startup and closing in shutdown. This remains advisory but is strongly recommended.
- The `TopMemoryContext` allocation pattern across multiple functions is fragile. A dedicated child `MemoryContext` that can be reset as a unit would be safer. This remains advisory.
- The code is well-structured with thorough comments, defensive stale-state detection (execution_id, call stack depth), and proper separation of concerns. The `InsertExecInfo` struct in `pltsql-2.h` is a clean shared structure across the three exec statement types.
- The `parse_insert_exec_table_name` function uses `strrchr` on `.` without bracket/quote awareness. T-SQL identifiers like `[db.with.dots].[schema].[table]` would be mis-split. This is low risk since `RangeVarGetRelid` does permission checks, but worth noting for correctness.

contrib/babelfishpg_tsql/src/pl_insert_exec.c
```diff
@@ -0,0 +1,474 @@
...
+
+	/*
+	 * Create a TupleConversionMap for type coercion if needed.
+	 *
+	 * convert_tuples_by_position() returns NULL if no conversion is needed
+	 *
+	 * Set called_from_tsql_insert_execute = true so that:
+	 *   - build_attrmap_by_position() skips the type mismatch error
+	 *   - execute_attr_map_tuple() will call exec_tsql_cast_value_hook for coercion
+	 */
```
tanscorpio7: 1. We shoudln't be skipping the type mismatch errors instead add an implicit type casting on top.
2. In SQL Server are the column matched based on column positions or names ?

tanscorpio7 (changes requested):

contrib/babelfishpg_tsql/src/pl_insert_exec.c
monk0707: 1. Using ExecBuildProjectionInfo in startup and use execproject in reciever for implicit type casting
2. Columns are matched based on column positions.

contrib/babelfishpg_tsql/src/pl_insert_exec.c
```diff
@@ -0,0 +1,513 @@
...
+	initStringInfo(&flush_query);
+
+	appendStringInfo(&flush_query,
+		"INSERT INTO %s%s SELECT * FROM %s",
+		quote_qualified_identifier(nspname, relname),
+		column_list_str ? column_list_str : "",
+		quote_identifier(temp_table_name));
+
+	/* Route through execute_batch to handle triggers and errors properly. */
+	execute_batch(estate, flush_query.data, NULL, NULL);
```
pranavJ23, line 267: Let's free flush_query.data

contrib/babelfishpg_tsql/src/pl_insert_exec.c
```diff
@@ -0,0 +1,513 @@
...
+	else
+		/* No coercion needed - insert directly */
+		table_tuple_insert(temp_rel,
+						   slot,
+						   cid,
+						   0,
+						   NULL);
+
+	/* Close the relation */
+	table_close(temp_rel, NoLock);
```
pranavJ23: Is this intentional, not dropping the RowExclusiveLock, we held while opening the relation ?

contrib/babelfishpg_tsql/src/pl_insert_exec.c
```diff
@@ -0,0 +1,513 @@
...
+
+	myState->rows_inserted++;
```
pranavJ23: Where exactly we are using this count? I see we haven't used it in this PR.

pranavJ23 (changes requested):

contrib/babelfishpg_tsql/src/pl_insert_exec.c
monk0707: yes, it is to remove the overhead of acquiring and releasing lock for each tuple, in case if no of tuples are very large. But I think a cleaner approach would be to open table once in startup and closing it in shutdown function.

contrib/babelfishpg_tsql/src/pl_insert_exec.c
monk0707: it will be used to update rowcount, but not in this PR, removing it

contrib/babelfishpg_tsql/src/pl_insert_exec.c
```diff
@@ -0,0 +1,506 @@
...
+/* Forward declarations for DestReceiver callbacks */
+static void insertexec_startup(DestReceiver *self, int operation, TupleDesc typeinfo);
+static bool insertexec_receive(TupleTableSlot *slot, DestReceiver *self);
+static void insertexec_shutdown(DestReceiver *self);
+static void insertexec_destroy(DestReceiver *self);
+
+/*
+ * Global INSERT EXEC context - defined in pltsql.h, declared here.
+ */
+InsertExecContext insert_exec_ctx;
```
tanscorpio7, line 59: 1. Lets rebase this PR
2. Is this information `temp_table_oid` duplicate with global context ?

contrib/babelfishpg_tsql/src/pl_insert_exec.c
```diff
@@ -0,0 +1,506 @@
...
+					  src_att->atttypid,
+					  src_att->atttypmod,
+					  src_att->attcollation,
+					  0);					/* varlevelsup */
+
+		expr = (Node *) var;
+
+		/* Check if coercion is needed for this column */
+		if (src_att->atttypid != tgt_att->atttypid ||
+			(src_att->atttypmod != tgt_att->atttypmod && tgt_att->atttypmod != -1))
```
tanscorpio7: Lets do ExecProject unconditionally, coerce_to_target_type automatically returns the same expression if type cast is not required. We can also remove `needs_coercion`

contrib/babelfishpg_tsql/src/pl_insert_exec.c
```diff
@@ -0,0 +1,506 @@
...
+											  src_att->atttypid,
+											  tgt_att->atttypid,
+											  tgt_att->atttypmod,
+											  COERCION_ASSIGNMENT,
+											  COERCE_IMPLICIT_CAST,
+											  -1);
+
+			if (cast_expr == NULL)
+			{
+				table_close(myState->temp_rel, RowExclusiveLock);
```
tanscorpio7: No need to close table before error. Error handling takes care of it. One or two other places that need this fix.

contrib/babelfishpg_tsql/src/pl_insert_exec.c
```diff
@@ -250,20 +271,236 @@ flush_insert_exec_temp_table(PLtsql_execstate *estate,
...
+
+	/* Create expression context and projection info if coercion needed */
+	if (myState->needs_coercion)
+	{
+		TupleDesc	proj_tupdesc;
+
+		myState->econtext = CreateStandaloneExprContext();
+
+		/* Create a copy of the temp table's tuple descriptor for the result slot */
+		proj_tupdesc = CreateTupleDescCopy(temp_tupdesc);
```
tanscorpio7: Why is this copy required ?

contrib/babelfishpg_tsql/src/pl_insert_exec.c
```diff
@@ -250,20 +271,236 @@ flush_insert_exec_temp_table(PLtsql_execstate *estate,
...
+ *
+ * Clean up the expression context and projection slot used for type coercion.
+ */
+static void
+insertexec_shutdown(DestReceiver *self)
+{
+	DR_insertexec *myState = (DR_insertexec *) self;
+
+	/* Close the temp table held since startup */
+	if (myState->temp_rel != NULL)
```
tanscorpio7: Change all these checks to assertions. Once we unconditionally do projection, all these resources should be avaiable.

contrib/babelfishpg_tsql/src/pl_insert_exec.c
```diff
@@ -250,20 +271,236 @@ flush_insert_exec_temp_table(PLtsql_execstate *estate,
...
 	pfree(flush_query.data);
 }
+
+/*
+ * Create a DestReceiver for INSERT EXEC that writes to a temp table.
+ */
+DestReceiver *
+CreateInsertExecDestReceiver(Oid temp_table_oid)
+{
+	DR_insertexec *self = (DR_insertexec *) palloc0(sizeof(DR_insertexec));
```
tanscorpio7: `palloc0_object` instead of `(DR_insertexec *) palloc0(sizeof` ?

tanscorpio7 (changes requested):

contrib/babelfishpg_tsql/src/pl_insert_exec.c
monk0707: not needed now, leftover from previous commit, Previously, temp_rel was opened at the start of insertexec_startup and closed at the end of it, before insertexec_receive ever runs, but now we changed to open in startup and closing in shutdown to remove unnecessary locking-unlocking overhead on each tuple.

contrib/babelfishpg_tsql/src/pl_insert_exec.c
```diff
@@ -267,3 +295,174 @@ flush_insert_exec_temp_table(PLtsql_execstate *estate,
...
+
+	/* Open temp table to get its tuple descriptor */
+	myState->temp_rel = table_open(insert_exec_ctx.temp_table_oid, RowExclusiveLock);
+	temp_tupdesc = RelationGetDescr(myState->temp_rel);
+	temp_natts = temp_tupdesc->natts;
+
+	if (result_natts != temp_natts)
+		ereport(ERROR,
+				(errcode(ERRCODE_DATATYPE_MISMATCH),
+				 errmsg("column name or number of supplied values does not match table definition")));
```
KushaalShroff, line 344: Do we need to provide any cleanup incase of early return/error

contrib/babelfishpg_tsql/src/pl_insert_exec.c
```diff
@@ -267,3 +295,174 @@ flush_insert_exec_temp_table(PLtsql_execstate *estate,
...
+ * insertexec_receive - receive each tuple and insert into temp table
+ */
+static bool
+insertexec_receive(TupleTableSlot *slot, DestReceiver *self)
+{
+	DR_insertexec *myState = (DR_insertexec *) self;
+	CommandId	cid;
+	TupleTableSlot *insert_slot;
+
+	cid = GetCurrentCommandId(true);
```
KushaalShroff: Should this be stored in our structure DR_insertexec since cid needs to be constant throughout the process?

contrib/babelfishpg_tsql/src/pl_insert_exec.c
```diff
@@ -267,3 +295,174 @@ flush_insert_exec_temp_table(PLtsql_execstate *estate,
...
+	myState->econtext = NULL;
+}
+
+/*
+ * insertexec_destroy - release DestReceiver object
+ */
+static void
+insertexec_destroy(DestReceiver *self)
+{
+	pfree(self);
```
KushaalShroff, line 468: this does not deep clean inside struct ptrs

contrib/babelfishpg_tsql/src/pl_insert_exec.c
monk0707: we don't need to explicitly cleanup or close table, PG Error handling takes care of it

contrib/babelfishpg_tsql/src/pl_insert_exec.c
monk0707: yes, that would be better

contrib/babelfishpg_tsql/src/pl_insert_exec.c
monk0707: no, it just frees the struct shell memory, the inner resources cleanup is being done in shutdown, which is called just before it.

kuntalghosh (changes requested):
# AI Code Review

Verdict: REQUEST_CHANGES

**Issue 1 - In `pl_insert_exec.c` at line 341: No validation of `temp_table_oid` before `table_open`:**

```c
/* Open temp table to get its tuple descriptor */
myState->temp_rel = table_open(insert_exec_ctx.temp_table_oid, RowExclusiveLock);
```

`insertexec_startup` reads `insert_exec_ctx.temp_table_oid` directly without an `OidIsValid()` guard. Since the planned `reset_insert_exec_context()` is deferred to a future wiring PR, stale state from a prior INSERT EXEC can survive into the next operation in the same session. Passing `InvalidOid` to `table_open` will crash the backend. Additionally, there is no check that the OID refers to a session-local temp table (`RELPERSISTENCE_TEMP`) or that the relkind is `RELKIND_RELATION`. Since `table_tuple_insert` bypasses ACL checks, this is a defense-in-depth gap.

Suggested fix:

```c
Oid oid = insert_exec_ctx.temp_table_oid;

if (!OidIsValid(oid))
    ereport(ERROR,
             errmsg("INSERT EXEC: temp table OID not initialized")));

myState->temp_rel = table_open(oid, RowExclusiveLock);

/* Verify it is a session-local temp table */
if (myState->temp_rel->rd_rel->relpersistence != RELPERSISTENCE_TEMP ||
    myState->temp_rel->rd_rel->relkind != RELKIND_RELATION)
{
    table_close(myState->temp_rel, RowExclusiveLock);
    ereport(ERROR,
             errmsg("INSERT EXEC: OID does not refer to a temp table")));
}
```

**Issue 2 - In `pl_insert_exec.c` at lines 20-34: Duplicate `#include "utils/builtins.h"`:**

```c
#include "utils/builtins.h"       /* line 25 */

#include "executor/executor.h"
#include "parser/scansup.h"

#include "catalog.h"
#include "multidb.h"
#include "session.h"

#include "utils/builtins.h"       /* line 34 - duplicate */
```

The header `utils/builtins.h` is included twice. The second inclusion at line 34 is redundant.

Suggested fix:

Remove the duplicate at line 34.

**Issue 3 - In `pl_insert_exec.c` at lines 20-27: Include ordering breaks alphabetical grouping:**

```c
#include "nodes/makefuncs.h"
#include "executor/tuptable.h"
#include "nodes/parsenodes.h"
#include "parser/parse_coerce.h"
#include "tcop/dest.h"
#include "utils/builtins.h"

#include "executor/executor.h"
```

The new includes are inserted in a way that breaks the alphabetical ordering within the PostgreSQL system headers group. `executor/tuptable.h` and `executor/executor.h` should be grouped together, and `nodes/parsenodes.h` should be adjacent to `nodes/makefuncs.h`. Additionally, `executor/executor.h` is isolated in its own group between the system headers and the Babelfish headers.

Suggested fix:

Group all PostgreSQL system includes together in alphabetical order:

```c
#include "commands/defrem.h"
#include "executor/executor.h"
#include "executor/tuptable.h"
#include "miscadmin.h"
#include "nodes/makefuncs.h"
#include "nodes/parsenodes.h"
#include "parser/parse_coerce.h"
#include "parser/scansup.h"
#include "tcop/dest.h"
#include "utils/builtins.h"
#include "utils/lsyscache.h"
```

**Issue 4 - In `pl_insert_exec.c` at line 412-430: `insertexec_receive` lacks defensive null checks:**

```c
static bool
insertexec_receive(TupleTableSlot *slot, DestReceiver *self)
{
    DR_insertexec *myState = (DR_insertexec *) self;
    TupleTableSlot *insert_slot;

    /* Reset per-tuple memory context for expression evaluation */
    ResetExprContext(myState->econtext);
```

If `insertexec_receive` is ever called without a prior `insertexec_startup` (e.g., due to a programming error in the future wiring code), `myState->econtext`, `myState->proj_info`, and `myState->temp_rel` are all NULL. `ResetExprContext(NULL)` will SIGSEGV. An `Assert` at the top would catch this in debug builds and document the precondition.

Suggested fix:

```c
static bool
insertexec_receive(TupleTableSlot *slot, DestReceiver *self)
{
    DR_insertexec *myState = (DR_insertexec *) self;
    TupleTableSlot *insert_slot;

    Assert(myState->temp_rel != NULL);
    Assert(myState->proj_info != NULL);
    Assert(myState->econtext != NULL);

    ResetExprContext(myState->econtext);
```

**Issue 5 - In `pl_insert_exec.c` at line 339-405: Resource leak on error between `table_open` and end of startup:**

```c
myState->temp_rel = table_open(insert_exec_ctx.temp_table_oid, RowExclusiveLock);
temp_tupdesc = RelationGetDescr(myState->temp_rel);
temp_natts = temp_tupdesc->natts;

if (result_natts != temp_natts)
    ereport(ERROR, ...);  /* temp_rel left open on error */
```

Between `table_open` at line 341 and the end of `insertexec_startup`, several operations can throw: the column-count mismatch `ereport(ERROR)`, `coerce_to_target_type` failures, and `ExecBuildProjectionInfo`. On a throw, PostgreSQL's transaction abort will release the lock and relcache reference, so this is not a leak in the current call graph. However, if the wiring layer ever performs partial cleanup outside transaction abort, the half-populated `DR_insertexec` state (with `temp_rel` set but `proj_slot`/`econtext` NULL) could cause issues in `insertexec_shutdown` which uses `Assert` on those fields.

Suggested fix:

Consider wrapping the post-`table_open` body in `PG_TRY`/`PG_CATCH` to close `temp_rel` on error and `PG_RE_THROW`, or document that startup errors are only expected within a transaction abort context. At minimum, change the `Assert` calls in `insertexec_shutdown` to conditional checks so that a partial-startup scenario doesn't crash in release builds:

```c
static void
insertexec_shutdown(DestReceiver *self)
{
    DR_insertexec *myState = (DR_insertexec *) self;

    if (myState->temp_rel != NULL)
    {
        table_close(myState->temp_rel, RowExclusiveLock);
        myState->temp_rel = NULL;
    }

    if (myState->proj_slot != NULL)
    {
        ExecDropSingleTupleTableSlot(myState->proj_slot);
        myState->proj_slot = NULL;
    }

    if (myState->econtext != NULL)
    {
        FreeExprContext(myState->econtext, true);
        myState->econtext = NULL;
    }
}
```

**Issue 6 - In `pl_insert_exec.c` at line 460-465: `insertexec_destroy` does not guard against missing shutdown:**

```c
static void
insertexec_destroy(DestReceiver *self)
{
    pfree(self);
}
```

If `insertexec_destroy` is called without a prior `insertexec_shutdown` (e.g., due to an error path that skips shutdown), the `proj_slot`, `econtext`, and `temp_rel` are leaked. The PostgreSQL executor protocol guarantees shutdown before destroy, but a defensive check would be safer given this is a new code path.

Suggested fix:

```c
static void
insertexec_destroy(DestReceiver *self)
{
    DR_insertexec *myState = (DR_insertexec *) self;

    /* Defensive: ensure resources are freed even if shutdown was skipped */
    if (myState->temp_rel != NULL || myState->proj_slot != NULL ||
        myState->econtext != NULL)
        insertexec_shutdown(self);

    pfree(self);
}
```

### Missing Tests

- **Blocking**: This PR adds ~175 lines of new C code (a complete DestReceiver implementation with type coercion, column validation, and tuple insertion) but includes zero test files. The PR description states "Tested via full feature integration in PR6," but the contribution guidelines require tests with code changes. The column count mismatch error path and type coercion failure path are independently testable once the temp table creation stub is wired. If tests truly cannot be added until the integration PR, this should be explicitly documented with a tracking ticket reference.

### Notes

- Prior review issues addressed in this revision: per-tuple `table_open`/`table_close` replaced with open-in-startup/close-in-shutdown pattern (Issue 2/3 from prior reviews), `needs_coercion` flag removed in favor of unconditional `ExecProject` (reviewer tanscorpio7's feedback), `palloc0_object` used instead of manual cast (reviewer tanscorpio7's feedback), `cid` stored in struct (reviewer KushaalShroff's feedback), unused row count removed (reviewer pranavJ23's feedback).
- Prior review issues still open: `insertexec_destroy` lacks defensive shutdown guard (Issue 6 from review 1), no `OidIsValid` check before `table_open` (new in this revision since the open moved to startup).
- The code is well-structured with clear function documentation and proper separation of concerns. The design of opening the relation once in startup and closing in shutdown is correct and addresses the prior per-tuple lock accumulation concern.
- The `typeinfo` field in `DR_insertexec` is stored but never read after startup. Consider removing it if it's not needed by a future PR, or add a comment explaining its intended use.
- Reviewer tanscorpio7 asked about removing `table_close` before error — the current code correctly relies on transaction abort for cleanup on the error path, which is consistent with PostgreSQL conventions.

monk0707:
AI code review : 
Issue1 : This is never practically possible if the code is written correctly, but adding a defensive null check is also fine.
Issue2 : Fixed, removed duplicate includes.
Issue3 : Fixed, Merged the split group of includes.
Issue4 : Added defensive Asserts in receive function.
Issue5 : Invalid, if rstartup throws an error, rshutdown can never be called.
Issue6 : Invalid, In dest receiver, rShutdown is always called before rDestroy.

kuntalghosh (changes requested):
# AI Code Review

Verdict: REQUEST_CHANGES

**Issue 1 - In `pl_insert_exec.c` at line 434-448: `insertexec_shutdown` uses `Assert` for resource checks — crashes in release builds on partial startup failure:**

```c
static void
insertexec_shutdown(DestReceiver *self)
{
	DR_insertexec *myState = (DR_insertexec *) self;

	Assert(myState->temp_rel != NULL);
	table_close(myState->temp_rel, RowExclusiveLock);
	myState->temp_rel = NULL;

	Assert(myState->proj_slot != NULL);
	ExecDropSingleTupleTableSlot(myState->proj_slot);
```

If `insertexec_startup` throws after `table_open` but before `proj_slot` or `econtext` are initialized (e.g., `coerce_to_target_type` returns NULL, `ExecBuildProjectionInfo` throws), and the wiring layer calls `rShutdown` on the partially-initialized receiver, the `Assert` calls are no-ops in release builds. The subsequent `ExecDropSingleTupleTableSlot(NULL)` or `FreeExprContext(NULL, true)` will crash the backend. Use conditional checks instead of `Assert` to make shutdown idempotent.

Suggested fix:

```c
static void
insertexec_shutdown(DestReceiver *self)
{
	DR_insertexec *myState = (DR_insertexec *) self;

	if (myState->proj_slot != NULL)
	{
		ExecDropSingleTupleTableSlot(myState->proj_slot);
		myState->proj_slot = NULL;
	}

	if (myState->econtext != NULL)
	{
		FreeExprContext(myState->econtext, true);
		myState->econtext = NULL;
	}

	if (myState->temp_rel != NULL)
	{
		table_close(myState->temp_rel, RowExclusiveLock);
		myState->temp_rel = NULL;
	}
}
```

Note: the ordering is also changed — drop the slot and free the expression context *before* closing the relation, because `proj_slot` holds a reference to the relation's tuple descriptor (obtained via `RelationGetDescr` which does not pin the descriptor). Closing the relation first creates a latent use-after-free risk if the slot drop touches the descriptor.

**Issue 2 - In `pl_insert_exec.c` at line 455-459: `insertexec_destroy` does not guard against missing shutdown:**

```c
static void
insertexec_destroy(DestReceiver *self)
{
	pfree(self);
}
```

If `insertexec_destroy` is called without a prior `insertexec_shutdown` (e.g., error during startup means shutdown was never invoked), the `temp_rel`, `proj_slot`, and `econtext` are leaked. The PostgreSQL executor protocol guarantees shutdown before destroy on the success path, but error paths can skip shutdown. This was raised in prior reviews and remains unaddressed.

Suggested fix:

```c
static void
insertexec_destroy(DestReceiver *self)
{
	DR_insertexec *myState = (DR_insertexec *) self;

	/* Defensive: ensure resources are freed even if shutdown was skipped */
	if (myState->temp_rel != NULL || myState->proj_slot != NULL ||
		myState->econtext != NULL)
		insertexec_shutdown(self);

	pfree(self);
}
```

**Issue 3 - In `pl_insert_exec.c` at line 333: No validation that `temp_table_oid` refers to a session-local temp table:**

```c
if (!OidIsValid(insert_exec_ctx.temp_table_oid))
	elog(ERROR, "INSERT EXEC failed due to missing temp table OID");

/* Open temp table to get its tuple descriptor */
myState->temp_rel = table_open(insert_exec_ctx.temp_table_oid, RowExclusiveLock);
```

The `OidIsValid` check is good, but there is no verification that the OID actually refers to a session-local temp table (`RELPERSISTENCE_TEMP`). Since `table_tuple_insert` bypasses ACL checks, a stale or reused OID could redirect procedure output to an unintended relation. The file has a TODO about missing `reset_insert_exec_context()` — until that is implemented, stale OIDs from a prior INSERT EXEC in the same session are a real risk.

Suggested fix:

```c
myState->temp_rel = table_open(insert_exec_ctx.temp_table_oid, RowExclusiveLock);

/* Verify it is a session-local temp table */
if (myState->temp_rel->rd_rel->relpersistence != RELPERSISTENCE_TEMP)
{
	table_close(myState->temp_rel, RowExclusiveLock);
	myState->temp_rel = NULL;
	elog(ERROR, "INSERT EXEC: OID does not refer to a temp table");
}
```

**Issue 4 - In `pl_insert_exec.c` at line 316-401: `insertexec_startup` is not exception-safe after `table_open`:**

```c
myState->temp_rel = table_open(insert_exec_ctx.temp_table_oid, RowExclusiveLock);
temp_tupdesc = RelationGetDescr(myState->temp_rel);
temp_natts = temp_tupdesc->natts;

if (result_natts != temp_natts)
	ereport(ERROR, ...);  /* temp_rel left open on error */
```

After `table_open` succeeds at line 333, multiple operations can throw: the column-count mismatch `ereport(ERROR)` at line 338, `coerce_to_target_type` failures, and `ExecBuildProjectionInfo`. On error, PostgreSQL's transaction abort will eventually release the lock and relcache reference. However, if the wiring layer (future PR) uses subtransactions or partial cleanup, the half-populated `DR_insertexec` state (with `temp_rel` set but `proj_slot`/`econtext` NULL) combined with `Assert`-only checks in shutdown will cause problems. Making shutdown idempotent (Issue 1) addresses the downstream impact, but documenting the precondition here would be valuable.

Suggested fix:

Add a comment documenting the error-handling contract:

```c
/*
 * Note: if any operation below throws, transaction abort handles cleanup
 * of temp_rel. The shutdown function must be idempotent to handle the case
 * where startup partially completed before an error.
 */
myState->temp_rel = table_open(insert_exec_ctx.temp_table_oid, RowExclusiveLock);
```

### Missing Tests

- **Blocking**: This PR adds ~175 lines of new C code (a complete DestReceiver implementation with type coercion, column validation, and tuple insertion) but includes zero test files. The PR description states "Tested via full feature integration in PR6," but the contribution guidelines require tests with code changes. The column count mismatch error path and type coercion failure path are independently testable once the temp table creation is wired. If tests truly cannot be added until the integration PR, this should be explicitly documented with a tracking ticket reference.

### Notes

- Prior review issues addressed in this revision: per-tuple `table_open`/`table_close` replaced with open-in-startup/close-in-shutdown pattern, `needs_coercion` flag removed in favor of unconditional `ExecProject`, `palloc0_object` used instead of manual cast, `cid` stored in struct, unused row count removed, defensive `Assert`s added in `insertexec_receive`, `OidIsValid` check added before `table_open`, duplicate `#include` removed, include ordering fixed.
- Prior review issues still open: `insertexec_destroy` lacks defensive shutdown guard (raised in reviews 1 and 4, still unaddressed), `insertexec_shutdown` uses `Assert` instead of conditional checks (raised in review 4, still unaddressed).
- The code is well-structured with clear function documentation and proper separation of concerns. The design of opening the relation once in startup and closing in shutdown is correct and addresses the prior per-tuple lock accumulation concern.
- The `cid` captured once at startup and reused for all tuples is correct for MVCC consistency within a single INSERT EXEC operation.
- The unconditional `ExecProject` approach (with `coerce_to_target_type` returning the Var unchanged when types match) is clean and avoids branching in the hot path.

kuntalghosh (dismissed):
# AI Code Review

Verdict: REQUEST_CHANGES

**Issue 1 - In `pl_insert_exec.c` at line 434-459: `insertexec_shutdown` uses `Assert` for resource checks — unsafe in release builds on partial startup failure:**

```c
static void
insertexec_shutdown(DestReceiver *self)
{
	DR_insertexec *myState = (DR_insertexec *) self;

	Assert(myState->temp_rel != NULL);
	table_close(myState->temp_rel, RowExclusiveLock);
```

If `insertexec_startup` throws after `table_open` but before `proj_slot` or `econtext` are initialized (e.g., `coerce_to_target_type` returns NULL triggering `ereport`, or `ExecBuildProjectionInfo` throws OOM), and the wiring layer calls `rShutdown` on the partially-initialized receiver, the `Assert` calls are no-ops in release builds. The subsequent `ExecDropSingleTupleTableSlot(NULL)` or `FreeExprContext(NULL, true)` will crash the backend. Use conditional checks instead of `Assert` to make shutdown idempotent. Additionally, the ordering should be changed — drop the slot and free the expression context *before* closing the relation, because `proj_slot` holds a reference to the relation's tuple descriptor (obtained via `RelationGetDescr` which does not copy the descriptor). Closing the relation first creates a latent use-after-free risk if the slot drop touches the descriptor.

Suggested fix:

```c
static void
insertexec_shutdown(DestReceiver *self)
{
	DR_insertexec *myState = (DR_insertexec *) self;

	if (myState->proj_slot != NULL)
	{
		ExecDropSingleTupleTableSlot(myState->proj_slot);
		myState->proj_slot = NULL;
	}

	if (myState->econtext != NULL)
	{
		FreeExprContext(myState->econtext, true);
		myState->econtext = NULL;
	}

	if (myState->temp_rel != NULL)
	{
		table_close(myState->temp_rel, RowExclusiveLock);
		myState->temp_rel = NULL;
	}
}
```

**Issue 2 - In `pl_insert_exec.c` at line 462-465: `insertexec_destroy` does not guard against missing shutdown:**

```c
static void
insertexec_destroy(DestReceiver *self)
{
	pfree(self);
}
```

If `insertexec_destroy` is called without a prior `insertexec_shutdown` (e.g., error during startup means shutdown was never invoked), the `temp_rel`, `proj_slot`, and `econtext` are leaked until ResourceOwner cleanup at transaction abort. While the PostgreSQL executor protocol guarantees shutdown before destroy on the success path, error paths can skip shutdown. This was raised in multiple prior reviews and remains unaddressed. Adding a defensive guard makes the code robust against future wiring changes.

Suggested fix:

```c
static void
insertexec_destroy(DestReceiver *self)
{
	DR_insertexec *myState = (DR_insertexec *) self;

	/* Defensive: ensure resources are freed even if shutdown was skipped */
	if (myState->temp_rel != NULL || myState->proj_slot != NULL ||
		myState->econtext != NULL)
		insertexec_shutdown(self);

	pfree(self);
}
```

**Issue 3 - In `pl_insert_exec.c` at line 333: No validation that `temp_table_oid` refers to a session-local temp table:**

```c
if (!OidIsValid(insert_exec_ctx.temp_table_oid))
	elog(ERROR, "INSERT EXEC failed due to missing temp table OID");

/* Open temp table to get its tuple descriptor */
myState->temp_rel = table_open(insert_exec_ctx.temp_table_oid, RowExclusiveLock);
```

The `OidIsValid` check is good, but there is no verification that the OID actually refers to a session-local temp table (`RELPERSISTENCE_TEMP`). Since `table_tuple_insert` bypasses triggers, CHECK constraints, FK validation, RLS, and policy enforcement, a stale or reused OID could redirect procedure output to an unintended relation. The file has a TODO about missing `reset_insert_exec_context()` — until that is implemented, stale OIDs from a prior INSERT EXEC in the same session are a real risk. This is a defense-in-depth check that turns a potential silent constraint-bypass into a hard error.

Suggested fix:

```c
myState->temp_rel = table_open(insert_exec_ctx.temp_table_oid, RowExclusiveLock);

/* Verify it is a session-local temp table */
if (myState->temp_rel->rd_rel->relpersistence != RELPERSISTENCE_TEMP)
{
	table_close(myState->temp_rel, RowExclusiveLock);
	myState->temp_rel = NULL;
	elog(ERROR, "INSERT EXEC: OID %u does not refer to a temp table",
		 insert_exec_ctx.temp_table_oid);
}
```

**Issue 4 - In `pl_insert_exec.c` at line 316-394: `insertexec_startup` is not exception-safe after `table_open`:**

```c
myState->temp_rel = table_open(insert_exec_ctx.temp_table_oid, RowExclusiveLock);
temp_tupdesc = RelationGetDescr(myState->temp_rel);
temp_natts = temp_tupdesc->natts;

if (result_natts != temp_natts)
	ereport(ERROR, ...);  /* temp_rel left open on error */
```

After `table_open` succeeds, multiple operations can throw: the column-count mismatch `ereport(ERROR)`, `coerce_to_target_type` failures, `ExecBuildProjectionInfo`, and OOM in `palloc`. On error, PostgreSQL's transaction abort will eventually release the lock and relcache reference via the ResourceOwner. However, if the wiring layer (future PR) uses subtransactions or T-SQL TRY/CATCH semantics that absorb the error without a full transaction abort, the half-populated `DR_insertexec` state (with `temp_rel` set but `proj_slot`/`econtext` NULL) combined with `Assert`-only checks in shutdown will cause problems. At minimum, add a comment documenting the error-handling contract. Ideally, making shutdown idempotent (Issue 1) addresses the downstream impact.

Suggested fix:

Add a comment documenting the precondition:

```c
/*
 * Note: if any operation below throws, transaction abort handles cleanup
 * of temp_rel via ResourceOwner. The shutdown function must be idempotent
 * to handle the case where startup partially completed before an error.
 */
myState->temp_rel = table_open(insert_exec_ctx.temp_table_oid, RowExclusiveLock);
```

### Missing Tests

- **Blocking**: This PR adds ~175 lines of new C code (a complete DestReceiver implementation with type coercion, column validation, and tuple insertion) but includes zero test files. The PR description states "Tested via full feature integration in PR6," but the contribution guidelines require tests with code changes. The column count mismatch error path and type coercion failure path are independently testable once the temp table creation is wired. If tests truly cannot be added until the integration PR, this should be explicitly documented with a tracking ticket reference (e.g., a sub-task of BABEL-5522 specifically for test coverage).

### Notes

- Prior review issues addressed in this revision: per-tuple `table_open`/`table_close` replaced with open-in-startup/close-in-shutdown pattern, `needs_coercion` flag removed in favor of unconditional `ExecProject`, `palloc0_object` used instead of manual cast, `cid` stored in struct, unused row count removed, defensive `Assert`s added in `insertexec_receive`, `OidIsValid` check added before `table_open`, duplicate `#include` removed, include ordering fixed.
- Prior review issues still open: `insertexec_shutdown` uses `Assert` instead of conditional checks (Issue 1 — raised in review 4, still unaddressed), `insertexec_destroy` lacks defensive shutdown guard (Issue 2 — raised in reviews 1, 2, 3, and 4, still unaddressed), no temp-table persistence check (Issue 3 — raised in reviews 3 and 4, still unaddressed).
- The code is well-structured with clear function documentation and proper separation of concerns. The design of opening the relation once in startup and closing in shutdown is correct and addresses the prior per-tuple lock accumulation concern.
- The unconditional `ExecProject` approach (with `coerce_to_target_type` returning the Var unchanged when types match) is clean and avoids branching in the hot path.
- The `cid` captured once at startup and reused for all tuples is correct for MVCC consistency within a single INSERT EXEC operation.

tanscorpio7 (approved):

KushaalShroff (approved):

PR #4786, opened 2026-04-28
Support .QUERY() XML datatype method

JapleenKaur3:
### Description

Add support for the `.query()` XML data type method in Babelfish.

Currently, Babelfish supports `.exist()` and `.value()` XML methods but `.query()` was unsupported and threw an error. With this change, `.query(xpath)` is now supported and returns XML results matching T-SQL behavior for XPath 1.0 compatible queries.

The implementation follows the same 3-stage architecture used by `.exist()` and `.value()`:
1. ANTLR rewriting: `@x.query('/path')` is rewritten to `sys.bbf_xmlquery('/path', @x)`
2. Grammar: `BBF_XMLQUERY` is mapped in `gram-tsql-rule.y`
3. PL/pgSQL function `sys.bbf_xmlquery()` validates the input type, checks QUOTED_IDENTIFIER, calls PostgreSQL's `xpath()`, and returns the result as XML

Authored-by: Japleen Kaur [amjj@amazon.com](mailto:amjj@amazon.com)
Signed-off-by: Japleen Kaur [amjj@amazon.com](mailto:amjj@amazon.com)

### Issues Resolved

BABEL-5224

### Test Scenarios Covered

* **Use case based -** .query() on variables and table columns, multiple elements with positional access, nested XML, attributes and predicates, special characters, Unicode, empty/whitespace XML, mixed content, variable assignment, NULL handling, chaining with .exist()/.value(), .query() on FOR XML results, subqueries, CAST/CONVERT, stored procedures, functions, views, computed columns, XPath expressions, large XML, CASE/JOIN/UNION/ORDER BY/TOP

* **Boundary conditions -** Empty XML, self-closing elements, whitespace-only content, out-of-range positional access, no-match XPath returning empty XML

* **Arbitrary inputs -** Large XML with 100 replicated items, multiline formatting, random spaces in dot notation syntax

* **Negative test cases -** Type validation errors for VARCHAR, INT, NVARCHAR, BIGINT, FLOAT, BIT, DECIMAL, TEXT columns; NULL xpath argument; invalid XPath syntax; XML fragments; DOCTYPE with internal entities; variable in XPath expression; 4-part name reference; spatial function chaining;  argument count validation (0 args, 2 args); QUOTED_IDENTIFIER OFF

* **Minor version upgrade tests -** Added to upgrade/latest schedule

* **Major version upgrade tests -** No, the change adds a new PL/pgSQL function and does not modify existing aggregate signatures or internal functions

* **Performance tests -** No, the implementation delegates to PostgreSQL's built-in xpath() function which is already optimized

* **Tooling impact -** No

* **Client tests -** No

### Check List


kuntalghosh (commented):
# AI Code Review

Verdict: COMMENT

**Issue 1 - In `contrib/babelfishpg_tsql/src/tsqlIface.cpp` at line 9451: Missing argument count validation for QUERY in `validateXMLFunctionArgs`:**

```cpp
/* XML EXIST function requires only 1 argument */
if (xml_func->EXIST() && (expr_list == NULL || expr_list->expression().size() != 1))
    throw PGErrorWrapperException(ERROR, ERRCODE_UNDEFINED_FUNCTION, "The exist function requires 1 argument(s).", getLineAndPos(xml_func));

/* XML VALUE function requires only 2 argument */
if (xml_func->VALUE() && (expr_list == NULL || expr_list->expression().size() != 2))
    throw PGErrorWrapperException(ERROR, ERRCODE_UNDEFINED_FUNCTION, "The value function requires 2 argument(s).", getLineAndPos(xml_func));
```

The `validateXMLFunctionArgs` function validates argument counts for EXIST (1 arg) and VALUE (2 args), but has no validation for QUERY. T-SQL `.query()` requires exactly 1 argument (the XPath expression). Without this check, calling `.query()` with 0 or 2+ arguments would pass ANTLR validation and reach the PL/pgSQL function, where it would either fail with a confusing PostgreSQL error or silently ignore extra arguments. This matches the SOP pattern from PR#3771 (fix_pattern: "XML method argument validation should be centralized in validateAndRewriteXMLFunctionArgs()").

Suggested fix:

```cpp
/* XML QUERY function requires only 1 argument */
if (xml_func->QUERY() && (expr_list == NULL || expr_list->expression().size() != 1))
    throw PGErrorWrapperException(ERROR, ERRCODE_UNDEFINED_FUNCTION, "The query function requires 1 argument(s).", getLineAndPos(xml_func));
```

**Issue 2 - In `contrib/babelfishpg_tsql/src/tsqlUnsupportedFeatureHandler.cpp` at line 1524: Misleading instrumentation constant name after removing QUERY from unsupported:**

```cpp
if (ctx->MODIFY())
    handle(INSTR_UNSUPPORTED_TSQL_XML_QUERY, "XML MODIFY", getLineAndPos(ctx));
```

After removing `.query()` from the unsupported handler, the constant `INSTR_UNSUPPORTED_TSQL_XML_QUERY` is now only used for XML MODIFY. The constant name is misleading — it says "XML_QUERY" but guards "XML MODIFY". Consider renaming to `INSTR_UNSUPPORTED_TSQL_XML_MODIFY` or a more generic `INSTR_UNSUPPORTED_TSQL_XML_METHOD` for clarity. This is advisory since the constant is just an enum value for telemetry, but it will confuse future readers.

Suggested fix: Rename `INSTR_UNSUPPORTED_TSQL_XML_QUERY` to `INSTR_UNSUPPORTED_TSQL_XML_MODIFY` in `pltsql_instr.h` and `tsqlUnsupportedFeatureHandler.cpp`.

**Issue 3 - In `contrib/babelfishpg_tsql/sql/sys_functions.sql` at line 195: Empty string return for no-match differs from SQL Server NULL behavior:**

```sql
result_set := xpath(xpath_pattern, xml_element);
IF (cardinality(result_set) = 0) THEN
    RETURN ''::xml;
```

When the XPath matches no nodes, the function returns `''::xml` (empty XML). In SQL Server, `.query()` with no matching nodes returns an empty XML fragment (effectively empty string), so this behavior appears correct for the common case. However, verify that `''::xml` produces the same wire-format result as SQL Server's empty XML fragment across all client drivers (JDBC, .NET, ODBC). The expected output shows empty lines for no-match cases, which suggests this is working, but it would be good to confirm with a `.value()` chain on the empty result to ensure type compatibility.

**Issue 4 - In `test/JDBC/input/xml/xml_query-vu-verify.sql`: Missing test for `.query()` with wrong argument count:**

The test file has 52 sections with excellent coverage of XPath expressions, type validation, NULL handling, QUOTED_IDENTIFIER, computed columns, and more. However, there are no tests for calling `.query()` with the wrong number of arguments:
- `.query()` with 0 arguments
- `.query('/path', 'extra')` with 2 arguments

These tests would validate the argument count check in `validateXMLFunctionArgs` (see Issue 1). The XML method test coverage checklist from PR#3771 explicitly requires testing argument count validation.

Suggested fix: Add test cases:

```sql
-- .query() with no arguments (should error)
DECLARE @x XML = '<root><name>John</name></root>';
SELECT @x.query();
GO

-- .query() with too many arguments (should error)
DECLARE @x XML = '<root><name>John</name></root>';
SELECT @x.query('/root/name', 'extra');
GO
```

**Issue 5 - In `test/JDBC/input/xml/xml_query-vu-verify.sql`: Missing prepared statement / sp_executesql tests:**

Per the testing rules (PreparedStatementTesting), features should be tested via prepared statements (JDBC PreparedStatement, sp_executesql, sp_prepare/sp_execute) since the RPC code path differs from SQL batch. The current tests only use direct SQL batch execution. While `.query()` takes a string literal argument (not a parameter), testing via `sp_executesql` with the XML variable as a parameter would exercise the RPC path.

**Issue 6 - In `test/JDBC/input/xml/xml_query-vu-verify.sql`: Missing transaction/rollback tests:**

Per the testing rules (TransactionAndNestingTesting), features should be tested inside explicit transactions with ROLLBACK and TRY/CATCH. No tests verify `.query()` behavior inside a transaction that rolls back, or inside a TRY/CATCH block.

**Issue 7 - In `test/JDBC/input/xml/xml_query-vu-verify.sql`: Missing table alias usage in test queries:**

Per the testing rules (TableAliasInTests), test queries involving relations should use table aliases. Many test queries use bare column references without aliases:

```sql
SELECT id, data.query('/root/name') FROM xml_query_t1 WHERE id = 1;
```

While some queries in Section 32 (JOIN) and Section 39 use aliases, the majority do not. Adding aliases to a representative subset would improve coverage of alias-related parser paths.

### Missing Tests
- Argument count validation: `.query()` with 0 args and 2+ args
- Prepared statement paths: sp_executesql, sp_prepare/sp_execute with XML parameter
- Transaction/rollback: `.query()` inside BEGIN/ROLLBACK and TRY/CATCH
- Table alias coverage in more test queries
- Major version upgrade test (PR description says "No" for major version upgrade tests, but the upgrade script adds a new function — this should have a major version upgrade test per UpgradeScriptTesting rule)

### Notes
- **PR Description**: Well-structured, includes BABEL-5224 JIRA link, test scenarios documented, DCO signed. Target branch BABEL_6_X_DEV is correct for new feature development.
- **Architecture**: The implementation correctly follows the established 3-stage pattern (ANTLR rewrite → grammar mapping → PL/pgSQL function) used by `.exist()` and `.value()`. The `bbf_xmlquery` function is structurally consistent with `bbf_xmlexist` and `bbf_xmlvalue`.
- **Upgrade script**: The function body in `babelfishpg_tsql--6.1.0--6.2.0.sql` is identical to `sys_functions.sql` — verified correct.
- **Security**: Security review passed with 0 critical/high/medium findings. Two low-severity pre-existing patterns noted (PUBLIC EXECUTE default, `current_setting` without `missing_ok`) that are consistent with the existing `bbf_xmlexist` and `bbf_xmlvalue` functions.
- **Test coverage**: Comprehensive — 52 test sections covering basic usage, table columns, predicates, attributes, namespaces, special characters, Unicode, NULL handling, UDT validation, type errors, QUOTED_IDENTIFIER, computed columns, FOR XML integration, subqueries, CAST/CONVERT, stored procedures, functions, views, JOINs, UNIONs, ORDER BY, TOP, CASE, GROUP BY, CDATA, XML fragments, and more. The `forxml-path-elements` expected output update correctly changes from an error to a successful result.
- **Minor version upgrade test**: Present in `test/JDBC/upgrade/latest/schedule` — good.
- **SQL validation framework**: `expected_create.out` correctly updated to include `sys.bbf_xmlquery` in the "could not find tests" list (matching the pattern of `bbf_xmlexist` and `bbf_xmlvalue`).

JapleenKaur3:
## AI Review Feedback

### Addressed

**Issue 1 - Missing argument count validation for QUERY**
Added argument count check in `validateXMLFunctionArgs` for `.query()` (requires exactly 1 argument), matching the pattern used by `.exist()` and `.value()`.

**Issue 2 - Misleading instrumentation constant**
Fixed `INSTR_UNSUPPORTED_TSQL_XML_QUERY` to `INSTR_UNSUPPORTED_TSQL_XML_MODIFY` in `visitXml_func_arg` since the correct constant already existed.

**Issue 4 - Missing test for wrong argument count**
Added test section 53 with `.query()` called with 0 args and 2 args, validating the error from Issue 1.

### Not Applicable

**Issue 3 - Empty string return for no-match**
Already verified against T-SQL. Both return empty XML for no-match. Test sections 27 (empty XML vs no match) and 12 (chaining with .exist()/.value()) cover this.

**Issue 5 - Missing prepared statement tests**
`.exist()` and `.value()` don't have prepared statement tests either. `.query()` goes through the same planner path regardless of execution method.

**Issue 6 - Missing transaction/rollback tests**
`.query()` is a read-only SELECT operation, not transactional. Same as `.exist()` and `.value()`.

**Issue 7 - Missing table alias usage**
Already covered in test section 28 (syntax variations) with `table.column.query()` and `xml_query_t1.data.query()` patterns.

kuntalghosh (changes requested):
# AI Code Review

Verdict: REQUEST_CHANGES

**Issue 1 - In `test/JDBC/input/xml/xml_query-vu-verify.sql` and `test/JDBC/expected/xml_query-vu-verify.out`: Test input and expected output files are out of sync:**

The `.sql` input file ends with:
- Section 50: Invalid XML Path value
- Section 51: .query() chained with spatial functions (error cases)
- Section 52: GROUP BY with .query()

The `.out` expected output file ends with:
- Section 50: Invalid XML Path value
- Section 51: GROUP BY with .query()
- Section 52: .query() argument count validation

The input file has a Section 51 (spatial function chaining with `.STArea()`) that is absent from the expected output. The expected output has a Section 52 (argument count validation with 0 and 2 arguments) that is absent from the input file. This means the test files will fail when run — the expected output does not correspond to the actual SQL being executed. The argument count validation tests are important (they exercise the `validateXMLFunctionArgs` check added in `tsqlIface.cpp`), so they should be added to the `.sql` file and the `.out` file should be regenerated to include both the spatial chaining and argument count sections.

Suggested fix: Add the missing argument count validation tests to the `.sql` input file and regenerate the `.out` expected output file so both files are consistent. The final `.sql` file should contain all three sections (50, 51, 52, 53) and the `.out` file should match.

**Issue 2 - In `contrib/babelfishpg_tsql/src/pltsql_instr.h` at line 48: Dead instrumentation constant `INSTR_UNSUPPORTED_TSQL_XML_QUERY`:**

```cpp
INSTR_UNSUPPORTED_TSQL_XML_QUERY,
INSTR_UNSUPPORTED_TSQL_XML_MODIFY,
```

The previous review correctly identified that `INSTR_UNSUPPORTED_TSQL_XML_QUERY` was being used for XML MODIFY (misleading). The fix correctly changed the handler to use `INSTR_UNSUPPORTED_TSQL_XML_MODIFY`, but `INSTR_UNSUPPORTED_TSQL_XML_QUERY` is now defined in the header and never referenced anywhere in the codebase. This is dead code. Since `.query()` is now a supported feature, this constant should be removed to avoid confusion.

Suggested fix: Remove the `INSTR_UNSUPPORTED_TSQL_XML_QUERY` enum value from `pltsql_instr.h`. Note that removing an enum value from the middle of an enum can change the integer values of subsequent members — verify that these enum values are not persisted or used as wire-protocol identifiers before removing.

### Missing Tests
- The `.sql` input file is missing the argument count validation tests (`.query()` with 0 args and 2 args) that appear in the `.out` file. These must be added to the input file.

### Notes
- Prior review issues addressed: The previous AI review raised 7 issues. The author addressed Issues 1 (argument count validation), 2 (instrumentation constant), and 4 (argument count tests) in the latest revision. Issues 3 (empty string return), 5 (prepared statements), 6 (transactions), and 7 (table aliases) were reasonably dismissed by the author — `.query()` follows the same patterns as `.exist()` and `.value()` which also lack those test categories, and the empty-string return for no-match was verified against T-SQL behavior.
- Architecture: The implementation correctly follows the established 3-stage pattern (ANTLR rewrite → grammar mapping → PL/pgSQL function) used by `.exist()` and `.value()`. The `bbf_xmlquery` function is structurally consistent with `bbf_xmlexist` and `bbf_xmlvalue`.
- Upgrade script: The function body in `babelfishpg_tsql--6.1.0--6.2.0.sql` is identical to `sys_functions.sql` — verified correct.
- Test coverage: Comprehensive — 52 test sections covering basic usage, table columns, predicates, attributes, namespaces, special characters, Unicode, NULL handling, UDT validation, type errors, QUOTED_IDENTIFIER, computed columns, FOR XML integration, subqueries, CAST/CONVERT, stored procedures, functions, views, JOINs, UNIONs, ORDER BY, TOP, CASE, GROUP BY, CDATA, XML fragments, and more.
- Minor version upgrade test: Present in `test/JDBC/upgrade/latest/schedule`.
- SQL validation framework: `expected_create.out` correctly updated to include `sys.bbf_xmlquery`.
- The `forxml-path-elements` and `xml_exist` expected output updates correctly change from unsupported errors to successful results or updated error messages now that `.query()` is supported.

JapleenKaur3:
## AI Review Feedback

**Issue 1 (test files out of sync):** Fixed. The .sql input file and .out expected output file are now in sync. 

**Issue 2 (dead INSTR_UNSUPPORTED_TSQL_XML_QUERY):** Intentionally kept. Removing an enum value from the middle of `pltsql_instr.h` shifts the integer values of all subsequent members. These values are used for telemetry instrumentation, so changing them could break telemetry data consistency across versions. The constant is harmless as dead code and safer to leave in place.

kuntalghosh (dismissed):
# AI Code Review

Verdict: APPROVE

**Issue 1 - In `contrib/babelfishpg_tsql/src/pltsql_instr.h` at line 48: Dead `INSTR_UNSUPPORTED_TSQL_XML_QUERY` enum constant:**

```c
INSTR_UNSUPPORTED_TSQL_XML_NODES,
INSTR_UNSUPPORTED_TSQL_XML_QUERY,
INSTR_UNSUPPORTED_TSQL_XML_MODIFY,
```

Now that `.query()` is a supported feature, `INSTR_UNSUPPORTED_TSQL_XML_QUERY` is defined in the header but never referenced anywhere in the codebase. The author's rationale for keeping it (removing a mid-enum value shifts subsequent integer values, which could break telemetry data consistency) is reasonable. Consider adding a comment next to the constant to explain why it is retained, e.g., `/* retained for telemetry enum stability — .query() now supported */`. This prevents future contributors from wondering whether it is an oversight.

### Notes

- Prior review issues resolved: The previous two AI reviews raised a total of 9 issues across two revisions. All blocking issues have been addressed:
  - Issue 1 (argument count validation for QUERY in `validateXMLFunctionArgs`) — fixed, now validates exactly 1 argument.
  - Issue 2 (misleading `INSTR_UNSUPPORTED_TSQL_XML_QUERY` used for XML MODIFY) — fixed, now correctly uses `INSTR_UNSUPPORTED_TSQL_XML_MODIFY`.
  - Issue 4 (missing argument count tests) — fixed, Section 52 added with 0-arg and 2-arg error cases.
  - Issue 1 from second review (test files out of sync) — fixed, `.sql` input and `.out` expected output now contain identical sections (1–52) and are consistent.
  - Issue 2 from second review (dead enum constant) — intentionally retained with valid justification (telemetry enum stability). Demoted to advisory.
- Architecture: The implementation correctly follows the established 3-stage pattern used by `.exist()` and `.value()`: (1) ANTLR rewriting in `exitXml_func_arg` prepends `bbf_xml` to the method name, (2) argument validation in `validateXMLFunctionArgs`, and (3) PL/pgSQL function `sys.bbf_xmlquery()` handles type checking, QUOTED_IDENTIFIER validation, and delegates to PostgreSQL's `xpath()`. The function structure is consistent with `bbf_xmlexist` and `bbf_xmlvalue`.
- Upgrade script: The function body in `babelfishpg_tsql--6.1.0--6.2.0.sql` is byte-identical to `sys_functions.sql` — verified correct.
- Unsupported feature handler: The `visitXml_func_arg` handler correctly removes the `.query()` guard and retains only `.modify()` with the correct `INSTR_UNSUPPORTED_TSQL_XML_MODIFY` constant.
- Test coverage: Comprehensive — 52 test sections covering: basic variable and column usage, positional access, nested XML, attributes, predicates, special characters, Unicode, empty/whitespace XML, mixed content, variable assignment, NULL handling (variable, column, expression), WHERE clause usage, chaining with `.exist()` and `.value()`, FOR XML integration, subqueries, derived tables, CAST/CONVERT, stored procedures, scalar and inline table-valued functions, views, UDT type validation (XML UDT works, VARCHAR/IMAGE UDTs error), type validation errors for 8 non-XML types, QUOTED_IDENTIFIER OFF behavior, computed columns, XPath axes (self, parent, descendant, wildcard), large XML (100 items), XML comments, whitespace variations, multiple calls on same variable, CASE expressions, JOINs, UNIONs, ORDER BY, TOP, GROUP BY, spaces in method call syntax, expression types (subquery, CAST, function call), table.column.query() syntax, NULL xpath argument, XPath functions (true/false), case insensitivity, CDATA, XML fragments, XML declaration, DOCTYPE, variable in XPath, digital signatures, invalid XPath, and argument count validation.
- Minor version upgrade test: Present in `test/JDBC/upgrade/latest/schedule`.
- SQL validation framework: `expected_create.out` correctly updated to include `sys.bbf_xmlquery`.
- The `forxml-path-elements`, `xml_exist`, and `xml_value` expected output updates correctly change from "XML QUERY is not currently supported" errors to successful results or updated error messages now that `.query()` is supported.
- Cleanup script: Correctly drops objects in dependency order (computed column table → functions → views → procedures → base tables → UDTs).

kuntalghosh (approved):
# AI Code Review

Verdict: APPROVE

**Issue 1 - In `contrib/babelfishpg_tsql/src/pltsql_instr.h` at line 48: Dead `INSTR_UNSUPPORTED_TSQL_XML_QUERY` enum constant:**

```c
INSTR_UNSUPPORTED_TSQL_XML_NODES,
INSTR_UNSUPPORTED_TSQL_XML_QUERY,
INSTR_UNSUPPORTED_TSQL_XML_MODIFY,
```

Now that `.query()` is a supported feature, `INSTR_UNSUPPORTED_TSQL_XML_QUERY` is defined in the header but never referenced anywhere in the codebase. The author's rationale for keeping it (removing a mid-enum value shifts subsequent integer values, which could break telemetry data consistency) is reasonable. Consider adding a comment next to the constant to explain why it is retained, e.g., `/* retained for telemetry enum stability — .query() now supported */`. This prevents future contributors from wondering whether it is an oversight.

### Notes

- Prior review issues resolved: The previous two AI reviews raised a total of 9 issues across two revisions. All blocking issues have been addressed:
  - Issue 1 (argument count validation for QUERY in `validateXMLFunctionArgs`) — fixed, now validates exactly 1 argument.
  - Issue 2 (misleading `INSTR_UNSUPPORTED_TSQL_XML_QUERY` used for XML MODIFY) — fixed, now correctly uses `INSTR_UNSUPPORTED_TSQL_XML_MODIFY`.
  - Issue 4 (missing argument count tests) — fixed, Section 52 added with 0-arg and 2-arg error cases.
  - Issue 1 from second review (test files out of sync) — fixed, `.sql` input and `.out` expected output now contain identical sections (1–52) and are consistent.
  - Issue 2 from second review (dead enum constant) — intentionally retained with valid justification (telemetry enum stability). Demoted to advisory.
- Architecture: The implementation correctly follows the established 3-stage pattern used by `.exist()` and `.value()`: (1) ANTLR rewriting in `exitXml_func_arg` prepends `bbf_xml` to the method name, (2) argument validation in `validateXMLFunctionArgs`, and (3) PL/pgSQL function `sys.bbf_xmlquery()` handles type checking, QUOTED_IDENTIFIER validation, and delegates to PostgreSQL's `xpath()`. The function structure is consistent with `bbf_xmlexist` and `bbf_xmlvalue`.
- Upgrade script: The function body in `babelfishpg_tsql--6.1.0--6.2.0.sql` is byte-identical to `sys_functions.sql` — verified correct.
- Unsupported feature handler: The `visitXml_func_arg` handler correctly removes the `.query()` guard and retains only `.modify()` with the correct `INSTR_UNSUPPORTED_TSQL_XML_MODIFY` constant.
- Test coverage: Comprehensive — 52 test sections covering: basic variable and column usage, positional access, nested XML, attributes, predicates, special characters, Unicode, empty/whitespace XML, mixed content, variable assignment, NULL handling (variable, column, expression), WHERE clause usage, chaining with `.exist()` and `.value()`, FOR XML integration, subqueries, derived tables, CAST/CONVERT, stored procedures, scalar and inline table-valued functions, views, UDT type validation (XML UDT works, VARCHAR/IMAGE UDTs error), type validation errors for 8 non-XML types, QUOTED_IDENTIFIER OFF behavior, computed columns, XPath axes (self, parent, descendant, wildcard), large XML (100 items), XML comments, whitespace variations, multiple calls on same variable, CASE expressions, JOINs, UNIONs, ORDER BY, TOP, GROUP BY, spaces in method call syntax, expression types (subquery, CAST, function call), table.column.query() syntax, NULL xpath argument, XPath functions (true/false), case insensitivity, CDATA, XML fragments, XML declaration, DOCTYPE, variable in XPath, digital signatures, invalid XPath, and argument count validation.
- Minor version upgrade test: Present in `test/JDBC/upgrade/latest/schedule`.
- SQL validation framework: `expected_create.out` correctly updated to include `sys.bbf_xmlquery`.
- The `forxml-path-elements`, `xml_exist`, and `xml_value` expected output updates correctly change from "XML QUERY is not currently supported" errors to successful results or updated error messages now that `.query()` is supported.
- Cleanup script: Correctly drops objects in dependency order (computed column table → functions → views → procedures → base tables → UDTs).

kuntalghosh (changes requested):
# AI Code Review

Verdict: REQUEST_CHANGES

**Issue 1 - In `contrib/babelfishpg_tsql/src/tsql_for/forxml.c` at line 727: `bbf_xml_normalize` invokes libxml2 without `pg_xml_init()` or `PG_TRY` cleanup:**

```c
xmlInitParser();

ctxt = xmlNewParserCtxt();
if (ctxt == NULL)
    PG_RETURN_XML_P(xml_input);

xml_str = text_to_cstring((text *) xml_input);
xml_len = strlen(xml_str);

doc = xmlCtxtReadMemory(ctxt, xml_str, xml_len,
                        NULL, "UTF-8",
                        XML_PARSE_NOERROR | XML_PARSE_NOWARNING);
```

Every other libxml2 parser invocation in this codebase (e.g., `xml_parse` in upstream PG, `sp_xml_openquery` in `runtime/functions.c`) wraps the work in `pg_xml_init()` + `PG_TRY/PG_CATCH` + `pg_xml_done()`. This pattern installs the safe entity loader that blocks external DTD/entity resolution and guarantees cleanup of libxml2 resources on `ereport`. The new function skips both.

`XML_PARSE_NOERROR | XML_PARSE_NOWARNING` only suppress stderr output — they do not disable external entity loading or network access. Without `pg_xml_init()`, the entity loader depends on thread-local libxml2 globals that other backend code paths may have set. Additionally, `text_to_cstring()` can `ereport(ERROR)` on OOM between `xmlNewParserCtxt()` and `xmlFreeParserCtxt()`, leaking the parser context.

Since `bbf_xml_normalize` is SQL-callable by any authenticated user on attacker-controlled XML, this is a meaningful gap.

Suggested fix:

1. Call `pg_xml_init(PG_XML_STRICTNESS_WELLFORMED)` before allocating the parser context.
2. Wrap the parse → walk → serialize block in `PG_TRY/PG_CATCH` that frees `doc`, `ctxt`, `result_buf`, and calls `pg_xml_done(xmlerrcxt, true)` before `PG_RE_THROW()`.
3. Call `pg_xml_done(xmlerrcxt, false)` on the success path.
4. Add `XML_PARSE_NONET` to the parser options as defense in depth.

**Issue 2 - In `contrib/babelfishpg_tsql/src/tsql_for/forxml.c` at lines 735, 748, 756: Silent fallback to raw input on parse/serialization failure:**

```c
if (ctxt == NULL)
    PG_RETURN_XML_P(xml_input);
...
if (doc == NULL || xmlDocGetRootElement(doc) == NULL)
{
    if (doc != NULL)
        xmlFreeDoc(doc);
    PG_RETURN_XML_P(xml_input);
}
...
if (result_buf == NULL)
{
    PG_RETURN_XML_P(xml_input);
}
```

When parsing or serialization fails, the function silently returns the original input without normalization. The callers (`bbf_xmlquery`, `bbf_xmlexist`, `bbf_xmlvalue`) then pass this un-normalized XML to `xpath()` / `xmlexists()`, which may evaluate differently than T-SQL would — CDATA sections remain unconverted, whitespace nodes remain present. This creates semantic drift between Babelfish and T-SQL behavior on malformed-but-recoverable XML.

Suggested fix: Replace the three `PG_RETURN_XML_P(xml_input)` fallback paths with `ereport(ERROR, ERRCODE_INVALID_XML_DOCUMENT, ...)`. If backward compatibility with XML fragments is needed, handle that case explicitly (similar to PG's `XMLOPTION_DOCUMENT` vs `XMLOPTION_CONTENT` distinction).

**Issue 3 - In `contrib/babelfishpg_tsql/src/tsql_for/forxml.c` at line 701: `xmlNewText` return value not checked before `xmlReplaceNode`:**

```c
xmlNodePtr text_node = xmlNewText(child->content);
xmlReplaceNode(child, text_node);
xmlFreeNode(child);
```

`xmlNewText` returns NULL on allocation failure. Passing NULL to `xmlReplaceNode` causes the CDATA node to be silently unlinked (deleted) rather than converted to a text node, producing incorrect output.

Suggested fix:

```c
xmlNodePtr text_node = xmlNewText(child->content);
if (text_node == NULL)
    ereport(ERROR, (errcode(ERRCODE_OUT_OF_MEMORY),
                    errmsg("out of memory in XML normalization")));
xmlReplaceNode(child, text_node);
xmlFreeNode(child);
```

**Issue 4 - In `contrib/babelfishpg_tsql/src/tsql_for/forxml.c` at line 766: XML declaration stripping uses unbounded `strstr`:**

```c
if (len > 5 && strncmp(buf, "<?xml", 5) == 0)
{
    char *end = strstr(buf, "?>");
    if (end != NULL)
    {
        end += 2;
        if (*end == '\n')
            end++;
        len -= (end - buf);
        buf = end;
    }
}
```

`strstr` searches until NUL without a length bound. If `end - buf` ever exceeds `len` (e.g., due to a malformed serialization buffer), the subtraction `len -= (end - buf)` produces a negative `int`, which `cstring_to_text_with_len` interprets as a huge unsigned length. Use a length-bounded search instead.

Suggested fix: Replace `strstr(buf, "?>")` with a bounded search:

```c
char *end = NULL;
for (int i = 0; i + 1 < len; i++) {
    if (buf[i] == '?' && buf[i+1] == '>') { end = buf + i + 2; break; }
}
if (end != NULL) {
    if (end < buf + len && *end == '\n') end++;
    Assert(end - buf <= len);
    len -= (int)(end - buf);
    buf = end;
}
```

**Issue 5 - In `contrib/babelfishpg_tsql/sql/sys_functions.sql` at line 163 and upgrade script: `bbf_xmlvalue` calls `bbf_xml_normalize` twice per invocation:**

```sql
result_set := xpath(xpath_pattern, sys.bbf_xml_normalize(xml_element::xml));
...
result := (xpath('string(' + xpath_pattern + ')', sys.bbf_xml_normalize(xml_element::xml)))[1];
```

The `.value()` function calls `sys.bbf_xml_normalize()` twice — once for the initial `xpath()` call and again for the `string()` extraction. Each call parses the XML into a DOM, walks the tree, and serializes back. This doubles the XML parsing overhead for every `.value()` invocation. Consider normalizing once into a local variable and reusing it.

Suggested fix:

```sql
DECLARE normalized_xml xml;
...
normalized_xml := sys.bbf_xml_normalize(xml_element::xml);
result_set := xpath(xpath_pattern, normalized_xml);
...
result := (xpath('string(' + xpath_pattern + ')', normalized_xml))[1];
```

Apply the same fix in the upgrade script.

**Issue 6 - In `test/JDBC/expected/xml_exist-vu-verify.out` at line 828 and `xml_value-vu-verify.out` at line 1360: Chaining `.query()` result with `.exist()` or `.value()` produces `could not parse XML document`:**

```
SELECT @xml.query('/artists/artist').exist('/artist/@name')
...
~~ERROR (Message: could not parse XML document)~~
```

When `.query()` returns multiple elements (e.g., `<artist .../><artist .../><artist .../>`), the result is an XML fragment without a root element. Chaining `.exist()` or `.value()` on this fragment fails because `bbf_xml_normalize` cannot parse it as a well-formed document. In T-SQL, `.query()` returns an `xml` value that can be chained with other XML methods. This is a functional gap — the error message is also confusing since the XML was valid before `.query()` was called.

This is not necessarily blocking for this PR since it's a known limitation of the fragment handling, but it should be documented as a known limitation or tracked as a follow-up.

### Notes
- Prior review issues resolved: Three previous AI reviews raised a total of 9 issues. All blocking issues from those reviews have been addressed: argument count validation added, instrumentation constant fixed, test files synced. The dead `INSTR_UNSUPPORTED_TSQL_XML_QUERY` enum constant is intentionally retained for telemetry stability — reasonable.
- Architecture: The implementation correctly follows the established 3-stage pattern (ANTLR rewrite → grammar mapping → PL/pgSQL function) used by `.exist()` and `.value()`. The `bbf_xmlquery` function is structurally consistent with `bbf_xmlexist` and `bbf_xmlvalue`.
- The `bbf_xml_normalize` C function is a good addition that improves T-SQL compatibility for whitespace and CDATA handling across all three XML methods. The approach of normalizing at the C level before passing to `xpath()`/`xmlexists()` is sound.
- Upgrade script: The function bodies in `babelfishpg_tsql--6.1.0--6.2.0.sql` match `sys_functions.sql` — verified correct. The upgrade script also correctly re-creates `bbf_xmlexist` and `bbf_xmlvalue` with the new `bbf_xml_normalize` calls.
- Test coverage: Comprehensive — 52 test sections covering basic usage, table columns, predicates, attributes, namespaces, special characters, Unicode, NULL handling, UDT validation, type errors, QUOTED_IDENTIFIER, computed columns, FOR XML integration, subqueries, CAST/CONVERT, stored procedures, functions, views, JOINs, UNIONs, ORDER BY, TOP, CASE, GROUP BY, CDATA, XML fragments, argument count validation, and more.
- PR description: Well-structured with BABEL-5224 JIRA link, detailed test scenarios, DCO sign-off. Target branch BABEL_6_X_DEV is correct.
- Minor version upgrade test present in `test/JDBC/upgrade/latest/schedule`.

contrib/babelfishpg_tsql/sql/sys_functions.sql
```diff
@@ -109,7 +116,7 @@ BEGIN
         RAISE EXCEPTION 'SELECT failed because the following SET options have incorrect settings: ''QUOTED_IDENTIFIER''. Verify that SET options are correct for XML data type methods.';
     END IF;
 
-    RETURN xmlexists($1 passing by value $2);
+    RETURN xmlexists($1 passing by value sys.bbf_xml_normalize($2::xml));
```
rohit01010: can you point me to any test case to understand why we need to normalise the XML input in .EXISTS() and .VALUE() methods

contrib/babelfishpg_tsql/sql/sys_functions.sql
```diff
@@ -161,6 +168,46 @@ END
...
+        RAISE EXCEPTION 'Cannot call methods on %.', temp_datatype;
+    END IF;
+
+    pltsql_quoted_identifier := current_setting('babelfishpg_tsql.quoted_identifier');
+
+    IF (pltsql_quoted_identifier = 'off') THEN
+        RAISE EXCEPTION 'SELECT failed because the following SET options have incorrect settings: ''QUOTED_IDENTIFIER''. Verify that SET options are correct for XML data type methods.';
+    END IF;
+
+    result_set := xpath(xpath_pattern, sys.bbf_xml_normalize(xml_element::xml));
```
rohit01010: We are parsing the xml_element twice once in bbf_xml_normalize and another in xpath. Is it possible to avoid such multiple parsing of same xml element.

test/JDBC/upgrade/latest/schedule
```diff
@@ -682,4 +682,5 @@ babel_convert_with_style
 forxml-raw-elements
 forxml-raw-elements-1gb-limit
 forxml-path-elements
+xml_query
```
rohit01010, line 685: Let's add this to other versions schedule files as well.

contrib/babelfishpg_tsql/sql/sys_functions.sql
JapleenKaur3: Whitespace-only text nodes. Example:

```sql
DECLARE @x XML = '<root>   </root>';
SELECT @x.exist('/root[text()]');
-- T-SQL: 0 (whitespace stripped at parse)
-- Without normalization: 1 (whitespace text node preserved by PG)

SELECT DATALENGTH(@x.value('(/root)[1]', 'VARCHAR(50)'));
-- T-SQL: 0
-- Without normalization: 3
```

test/JDBC/upgrade/latest/schedule
JapleenKaur3: added

contrib/babelfishpg_tsql/sql/sys_functions.sql
JapleenKaur3: Added a fast-path scan in `bbf_xml_normalize`: if the input has no CDATA and no whitespace-only text nodes (common case), it returns unchanged without parsing. So `xpath` is the only parse for most documents.

contrib/babelfishpg_tsql/sql/sys_functions.sql
```diff
@@ -156,7 +156,47 @@ BEGIN
...
+    result_set := xpath(xpath_pattern, xml_element);
+    IF (cardinality(result_set) = 0) THEN
+        RETURN ''::xml;
+    ELSE
+        result := array_to_string(result_set, '')::xml;
+        RETURN result;
+    END IF;
 END
 $BODY$
 LANGUAGE plpgsql STABLE STRICT PARALLEL SAFE;
```
Deepesh125: Why not C function? Why parallel safe?

contrib/babelfishpg_tsql/src/xml_methods.c
rohit01010, line 1: Creating a file just for .QUERY() xml method does not seem right to me. Can we create this functions in `babelfish_extensions/contrib/babelfishpg_tsql/runtime/functions.c` file instead ?

contrib/babelfishpg_tsql/src/xml_methods.c
```diff
@@ -0,0 +1,216 @@
...
+	Datum	   *elems;
+	bool	   *nulls;
+	int			nitems;
+	const char *quoted_identifier;
+	StringInfoData buf;
+	int			i;
+
+	/* Strict: return NULL if any argument is NULL */
+	if (PG_ARGISNULL(0) || PG_ARGISNULL(1))
+		PG_RETURN_NULL();
```
rohit01010: function is already marked strict, why do we need to add this extra check ?

contrib/babelfishpg_tsql/src/xml_methods.c
```diff
@@ -0,0 +1,216 @@
...
+
+	/* Try direct translation first */
+	if (common_utility_plugin_ptr)
+	{
+		Datum result;
+
+		result = (*common_utility_plugin_ptr->translate_pg_type_to_tsql)(tmp_fcinfo);
+		if (result)
+			return text_to_cstring(DatumGetTextPP(result));
+	}
```
rohit01010: let's use function resolve_pg_type_to_tsql() to get tsql type name from type_oid

contrib/babelfishpg_tsql/src/xml_methods.c
```diff
@@ -0,0 +1,216 @@
...
+
+		/* Get T-SQL type name for error message */
+		typname = get_tsql_type_name(arg_type);
+		if (typname == NULL)
+			typname = format_type_be(arg_type);
+
+		ereport(ERROR,
+				(errcode(ERRCODE_DATATYPE_MISMATCH),
+				 errmsg("Cannot call methods on %s.", typname)));
+	}
```
rohit01010:
```
    /* Lookup the datatype of the supplied argument */
	arg_type = get_fn_expr_argtype(fcinfo->flinfo, 1);

	/* UDT Handling. */
	immediate_base_type = get_immediate_base_type_of_UDT_internal(arg_type);
    if (OidIsValid(immediate_base_type))
	{
		arg_type = immediate_base_type;
	}

	if (arg_type != XMLOID)
	{
		const char *typname = NULL;

		/* Get T-SQL type name for error message */
        if (common_utility_plugin_ptr && common_utility_plugin_ptr->resolve_pg_type_to_tsql)
        {
       		typname = (*common_utility_plugin_ptr->resolve_pg_type_to_tsql)(arg_type);
        }
		if (typname == NULL)
			typname = format_type_be(arg_type);

		ereport(ERROR,
				(errcode(ERRCODE_DATATYPE_MISMATCH),
				 errmsg("Cannot call methods on %s.", typname)));
	}
```
I think with above code refactoring, we can avoid creation of get_tsql_type_name function. Let me know if there are any currently handled case which is not handled by above code snippet.

contrib/babelfishpg_tsql/src/xml_methods.c
```diff
@@ -0,0 +1,216 @@
...
+
+	/* Check QUOTED_IDENTIFIER setting (required for XML methods in T-SQL) */
+	quoted_identifier = GetConfigOption("babelfishpg_tsql.quoted_identifier",
+										true, false);
+	if (quoted_identifier != NULL && strcmp(quoted_identifier, "off") == 0)
```
rohit01010: let's use `strlen(quoted_identifier) == 3 && strncmp(quoted_identifier, "off", 3)` instead of strcmp()

contrib/babelfishpg_tsql/src/xml_methods.c
```diff
@@ -0,0 +1,216 @@
...
+
+		if (strstr(xml_str, "<!DOCTYPE") != NULL)
+		{
+			pfree(xml_str);
+			ereport(ERROR,
+					(errcode(ERRCODE_INVALID_XML_CONTENT),
+					 errmsg("%s", "invalid XML content")));
+		}
+		pfree(xml_str);
+	}
```
rohit01010: Isn't this should be detected in XML parsing and not in .QUERY() method ?

contrib/babelfishpg_tsql/src/xml_methods.c
```diff
@@ -0,0 +1,216 @@
...
+	/*
+	 * Call the built-in xpath(text, xml, text[][]) directly with an empty
+	 * namespace array. Returns xml[] (array of XML fragments).
+	 */
+	namespaces = construct_empty_array(TEXTOID);
+	xpath_result = DirectFunctionCall3Coll(xpath,
+										   InvalidOid,
+										   PointerGetDatum(xpath_expr),
+										   xml_datum,
+										   PointerGetDatum(namespaces));
```
rohit01010: can we directly call `xpath_internal` function ?

test/JDBC/expected/xml_exist-vu-verify.out
```diff
@@ -825,9 +825,11 @@ GO
 DECLARE @xml XML = '<artists> <artist name="John Doe"/> <artist name="Edward Poe"/> <artist name="Mark The Great"/> </artists>'
 SELECT @xml.query('/artists/artist').exist('/artist/@name')
 GO
+~~START~~
+bit
 ~~ERROR (Code: 33557097)~~
 
-~~ERROR (Message: 'XML QUERY' is not currently supported in Babelfish)~~
+~~ERROR (Message: could not parse XML document)~~
```
rohit01010: can you explain why this is not a valid xml document? also can we update this with some valid XML Document.

contrib/babelfishpg_tsql/src/xml_methods.c
JapleenKaur3: done

contrib/babelfishpg_tsql/src/xml_methods.c
JapleenKaur3: Right, removed the redundant check.

contrib/babelfishpg_tsql/src/xml_methods.c
JapleenKaur3: done

contrib/babelfishpg_tsql/src/xml_methods.c
JapleenKaur3: done

contrib/babelfishpg_tsql/src/xml_methods.c
JapleenKaur3: `xpath_internal` is `static`. And `xpath()` itself is just a tiny wrapper that builds the `ArrayBuildState`, calls `xpath_internal`, and finalizes, so calling the internal directly would just relocate that boilerplate into our code without saving anything meaningful.

test/JDBC/expected/xml_exist-vu-verify.out
JapleenKaur3: The chained `.query('/artists/artist')` returns three sibling `<artist>` elements with no single root wrapping them — that's a valid XML fragment but not a valid XML document (XML spec requires exactly one root element). Updated the test to `@xml.query('/artists').exist('/artists/artist/@name')` so the chained `.query()` returns a single-rooted document.

contrib/babelfishpg_tsql/src/xml_methods.c
JapleenKaur3: done

contrib/babelfishpg_tsql/src/xml_methods.c
JapleenKaur3: removed.

contrib/babelfishpg_tsql/sql/sys_functions.sql
JapleenKaur3: changed to C function now

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -6095,3 +6095,116 @@ openxml_simple(PG_FUNCTION_ARGS)
...
+				 errmsg("SELECT failed because the following SET options have "
+						"incorrect settings: 'QUOTED_IDENTIFIER'. Verify that "
+						"SET options are correct for XML data type methods.")));
+
+	/*
+	 * Call the built-in xpath(text, xml, text[][]) directly with an empty
+	 * namespace array. Returns xml[] (array of XML fragments).
+	 */
+	namespaces = construct_empty_array(TEXTOID);
+	xpath_result = DirectFunctionCall3Coll(xpath,
```
rohit01010: let's use `DirectFunctionCall3` instead.

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -6095,3 +6095,116 @@ openxml_simple(PG_FUNCTION_ARGS)
...
+				(errcode(ERRCODE_INVALID_PARAMETER_VALUE),
+				 errmsg("SELECT failed because the following SET options have "
+						"incorrect settings: 'QUOTED_IDENTIFIER'. Verify that "
+						"SET options are correct for XML data type methods.")));
+
+	/*
+	 * Call the built-in xpath(text, xml, text[][]) directly with an empty
+	 * namespace array. Returns xml[] (array of XML fragments).
+	 */
+	namespaces = construct_empty_array(TEXTOID);
```
rohit01010, line 6171: when WITH XMLNAMESPACES is supported we need to fill this namespaces with list of namespaces in WITH XMLNAMESPACES, right ? if yes then let's add a TODO comment that "whenever WITH XMLNAMESPACES is supported update the following with given namespaces"

contrib/babelfishpg_tsql/runtime/functions.c
JapleenKaur3: updated

contrib/babelfishpg_tsql/runtime/functions.c
JapleenKaur3: added comment

kuntalghosh (commented):
# AI Code Review

Verdict: COMMENT

**Issue 1 - In `contrib/babelfishpg_tsql/runtime/functions.c` at line 6172: Missing TODO comment for WITH XMLNAMESPACES support:**

```c
namespaces = construct_empty_array(TEXTOID);
xpath_result = DirectFunctionCall3Coll(xpath,
                                       InvalidOid,
                                       PointerGetDatum(xpath_expr),
                                       xml_datum,
                                       PointerGetDatum(namespaces));
```

The empty namespace array is a placeholder. When `WITH XMLNAMESPACES` is eventually supported, this array will need to be populated with the declared namespace mappings. A TODO comment here would help future developers understand the intent and find this code when implementing namespace support.

Suggested fix:

```c
/*
 * TODO: When WITH XMLNAMESPACES is supported, populate this array
 * with the namespace declarations from the WITH clause.
 */
namespaces = construct_empty_array(TEXTOID);
```

**Issue 2 - In `contrib/babelfishpg_tsql/runtime/functions.c` at line 6173: Use `DirectFunctionCall3` instead of `DirectFunctionCall3Coll`:**

```c
xpath_result = DirectFunctionCall3Coll(xpath,
                                       InvalidOid,
                                       PointerGetDatum(xpath_expr),
                                       xml_datum,
                                       PointerGetDatum(namespaces));
```

The `xpath()` function does not use collation for its operation (it processes XML, not collation-sensitive text). Passing `InvalidOid` as the collation is functionally correct but `DirectFunctionCall3` is simpler and communicates the intent more clearly — that collation is irrelevant here.

Suggested fix:

```c
xpath_result = DirectFunctionCall3(xpath,
                                   PointerGetDatum(xpath_expr),
                                   xml_datum,
                                   PointerGetDatum(namespaces));
```

**Issue 3 - In `contrib/babelfishpg_tsql/runtime/functions.c` at line 6202: Fragment concatenation uses strlen-based operations:**

```c
char *fragment = TextDatumGetCString(elems[i]);
appendStringInfoString(&buf, fragment);
pfree(fragment);
```

Using `TextDatumGetCString` + `appendStringInfoString` is strlen-based. While this is safe in practice (PostgreSQL's XML type rejects NUL bytes), using length-aware operations is more robust and avoids an unnecessary copy through CString conversion.

Suggested fix:

```c
text *t = DatumGetTextPP(elems[i]);
appendBinaryStringInfo(&buf, VARDATA_ANY(t), VARSIZE_ANY_EXHDR(t));
```

And the final return:
```c
PG_RETURN_XML_P((xmltype *) cstring_to_text_with_len(buf.data, buf.len));
```

**Issue 4 - In `contrib/babelfishpg_tsql/src/pltsql_instr.h` at line 48: Dead enum value `INSTR_UNSUPPORTED_TSQL_XML_QUERY`:**

Now that `.query()` is supported, the enum value `INSTR_UNSUPPORTED_TSQL_XML_QUERY` is no longer referenced anywhere in the codebase. It remains defined in the header as dead code.

Suggested fix:

Consider removing the unused enum value or adding a comment noting it is retained for backward compatibility with telemetry data. If the instrumentation enum values are serialized to persistent storage, removing it could shift subsequent values — check whether the enum is used as a stable ABI.

**Issue 5 - In `contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--6.1.0--6.2.0.sql` at line 78: Missing blank line before next section:**

```sql
LANGUAGE C STABLE STRICT PARALLEL SAFE;
-- BABELFISH_FUNCTION_EXT (antlr_parse_cache)
```

The upgrade SQL inserts the new function definition immediately before the next section comment without a separating blank line. This is a minor formatting issue but other sections in the file use blank lines as separators.

Suggested fix:

```sql
LANGUAGE C STABLE STRICT PARALLEL SAFE;

-- BABELFISH_FUNCTION_EXT (antlr_parse_cache)
```

### Notes

- **Prior review issues addressed**: The author has addressed feedback from previous reviews — the function was moved from a separate `xml_methods.c` file to `runtime/functions.c`, redundant NULL checks were removed (function is STRICT), `resolve_pg_type_to_tsql()` is now used for type name resolution, and `strlen+strncmp` replaced `strcmp` for the QUOTED_IDENTIFIER check.
- **Open reviewer comment (rohit01010)**: The reviewer asked about using `DirectFunctionCall3` instead of `DirectFunctionCall3Coll` — this is echoed in Issue 2 above.
- **Architecture**: The implementation correctly follows the established 3-stage pattern (ANTLR rewrite → grammar mapping → C function) used by `.exist()` and `.value()`. The C implementation is a good choice over PL/pgSQL for performance.
- **Test coverage**: Comprehensive — 52 test sections covering basic usage, table columns, nested XML, attributes, predicates, special characters, Unicode, NULL handling, stored procedures, functions, views, UDTs, type validation errors, QUOTED_IDENTIFIER enforcement, computed columns, XPath expressions, large XML, CASE/JOIN/UNION/ORDER BY/TOP, and argument count validation. The vu-prepare/vu-verify/vu-cleanup pattern is correctly used for upgrade testing.
- **Upgrade testing**: Properly added to `latest/schedule` and older version schedules with the `before-17_12-or-18_5` variant that expects errors for objects not yet created.
- **Bug fix in unsupported feature handler**: The old code incorrectly used `INSTR_UNSUPPORTED_TSQL_XML_QUERY` for both QUERY and MODIFY — this PR correctly fixes it to use `INSTR_UNSUPPORTED_TSQL_XML_MODIFY` for the MODIFY case.
- **xml_exist test change**: The chaining test was correctly updated from `.query('/artists/artist').exist(...)` to `.query('/artists').exist(...)` because the former returns a multi-element fragment (not a valid XML document), while the latter returns a single root element that can be further queried.

kuntalghosh (dismissed):
# AI Code Review

Verdict: APPROVE

**Issue 1 - In `contrib/babelfishpg_tsql/runtime/functions.c` at line 6152: Error message capitalization deviates from PostgreSQL convention:**

```c
ereport(ERROR,
        (errcode(ERRCODE_DATATYPE_MISMATCH),
         errmsg("Cannot call methods on %s.", typname)));
```

Per PostgreSQL coding standards, `errmsg()` text should not capitalize the first letter and should not end with a period. However, this matches the existing pattern used by `bbf_xmlexist` and `bbf_xmlvalue` (which replicate T-SQL error messages verbatim for compatibility). No change needed — flagging for awareness only.

**Issue 2 - In `contrib/babelfishpg_tsql/runtime/functions.c` at line 6160: QUOTED_IDENTIFIER check silently passes when GUC is not loaded:**

```c
quoted_identifier = GetConfigOption("babelfishpg_tsql.quoted_identifier",
                                    true, false);
if (quoted_identifier != NULL && strlen(quoted_identifier) == 3 &&
    strncmp(quoted_identifier, "off", 3) == 0)
```

The `missing_ok=true` parameter means `GetConfigOption` returns NULL when the GUC is not registered. The current logic treats NULL as "ON" (no error). Under normal operation the GUC is always defined, so this is not a practical concern — but it means the function fails open rather than closed. This matches the existing `bbf_xmlexist`/`bbf_xmlvalue` pattern (which use `current_setting` in PL/pgSQL with similar semantics), so no change required for consistency. Flagging as advisory.

**Issue 3 - In `contrib/babelfishpg_tsql/src/tsqlUnsupportedFeatureHandler.cpp` at line 1523: Good bug fix — MODIFY was using wrong enum:**

```cpp
if (ctx->MODIFY())
    handle(INSTR_UNSUPPORTED_TSQL_XML_MODIFY, "XML MODIFY", getLineAndPos(ctx));
```

The original code incorrectly used `INSTR_UNSUPPORTED_TSQL_XML_QUERY` for the MODIFY handler. This fix correctly uses `INSTR_UNSUPPORTED_TSQL_XML_MODIFY`. Good catch.

### Notes

Prior review issues addressed:
- TODO comment for WITH XMLNAMESPACES added (Issue 1 from prior review)
- `DirectFunctionCall3` used instead of `DirectFunctionCall3Coll` (Issue 2 from prior review)
- Length-aware `DatumGetTextPP` + `appendBinaryStringInfo` used instead of strlen-based operations (Issue 3 from prior review)
- Dead `INSTR_UNSUPPORTED_TSQL_XML_QUERY` enum removed (Issue 4 from prior review)
- Reviewer feedback from rohit01010 addressed: moved to `functions.c`, removed redundant NULL check (STRICT handles it), used `resolve_pg_type_to_tsql`, added UDT handling, used `strlen`+`strncmp` instead of `strcmp`

Implementation quality:
- The 3-stage architecture (ANTLR rewrite → grammar mapping → C function) is consistent with `.exist()` and `.value()`
- Test coverage is comprehensive: 52 sections covering basic usage, table columns, nested XML, attributes, predicates, special characters, Unicode, NULL handling, UDT validation, type errors, QUOTED_IDENTIFIER, stored procedures, functions, views, computed columns, various XPath expressions, large XML, CASE/JOIN/UNION/ORDER BY/TOP, and argument count validation
- Upgrade path is properly handled with `before-17_12-or-18_5` variant for older versions
- The `xml_exist` test fix (changing `.query('/artists/artist').exist(...)` to `.query('/artists').exist(...)`) correctly ensures the chained call operates on a valid XML document rather than a fragment

Security review (C source): No critical or high-severity issues found. The function delegates to PostgreSQL's built-in `xpath()` via typed `DirectFunctionCall3` (no string interpolation), uses `StringInfoData` for safe buffer management, and properly validates input types before processing.

test/JDBC/upgrade/17_11/schedule
```diff
@@ -679,4 +679,5 @@ babel_convert_with_style
 forxml-raw-elements
 forxml-raw-elements-1gb-limit
 forxml-path-elements
+xml_query-before-17_12-or-18_5
```
rohit01010: Latest versions are 17.11 and 18.5, if we are planning to merge this in 5_X_DEV branch we need to rename this test file to `xml_query-before-17_11-or-18_5` also in 17_11 schedule file you need to put `xml_query` latest test file and not before one. Also you need to first merge this branch in lower branch i.e.BABEL_5_X_DEV and then merge in BABEL_6_X_DEV.

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -6095,3 +6095,119 @@ openxml_simple(PG_FUNCTION_ARGS)
...
+
+		/* Get T-SQL type name for error message */
+		if (common_utility_plugin_ptr)
+			typname = (*common_utility_plugin_ptr->resolve_pg_type_to_tsql)(arg_type);
+		if (typname == NULL)
+			typname = format_type_be(arg_type);
+
+		ereport(ERROR,
+				(errcode(ERRCODE_DATATYPE_MISMATCH),
+				 errmsg("Cannot call methods on %s.", typname)));
```
Deepesh125, line 6153: be specific which method

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -6095,3 +6095,119 @@ openxml_simple(PG_FUNCTION_ARGS)
...
+	}
+
+	/* Check QUOTED_IDENTIFIER setting (required for XML methods in T-SQL) */
+	quoted_identifier = GetConfigOption("babelfishpg_tsql.quoted_identifier",
```
Deepesh125: there's variable `pltsql_quoted_identifier` available, we should use it directly

contrib/babelfishpg_tsql/runtime/functions.c
```diff
@@ -6095,3 +6095,119 @@ openxml_simple(PG_FUNCTION_ARGS)
...
+	 * Call the built-in xpath(text, xml, text[][]) directly with an empty
+	 * namespace array. Returns xml[] (array of XML fragments).
+	 *
+	 * TODO: when WITH XMLNAMESPACES is supported, populate this array with
+	 * the declared (prefix, uri) pairs from the active namespace context.
+	 */
+	namespaces = construct_empty_array(TEXTOID);
+	xpath_result = DirectFunctionCall3(xpath,
+									   PointerGetDatum(xpath_expr),
+									   xml_datum,
```
Deepesh125, line 6174: do we need any validation on this? or `xpath` will do it internally?

Deepesh125 (changes requested):

contrib/babelfishpg_tsql/runtime/functions.c
JapleenKaur3: T-SQL itself doesn't name the method here - it just says `Cannot call methods on <type>.` so kept the wording verbatim.

contrib/babelfishpg_tsql/runtime/functions.c
JapleenKaur3: nothing extra needed - `xpath()` validates the xpath text (errors on empty), the namespace array shape, and the XML itself via libxml2 strict parse. We only do the T-SQL-side checks (xml type, QUOTED_IDENTIFIER) before calling it

contrib/babelfishpg_tsql/runtime/functions.c
JapleenKaur3: done

test/JDBC/upgrade/17_11/schedule
JapleenKaur3: done

contrib/babelfishpg_tsql/runtime/functions.c
Deepesh125: Can you please point me the code where is it happening?

JapleenKaur3: Datum
xml_in(PG_FUNCTION_ARGS)
{
    char       *s = PG_GETARG_CSTRING(0);
    xmltype    *vardata;
    xmlDocPtr   doc;

    vardata = (xmltype *) cstring_to_text(s);

    /* Parse the data to check if it is well-formed XML data. */
    doc = xml_parse(vardata, xmloption, true, GetDatabaseEncoding(),
                    NULL, NULL, fcinfo->context);
    ...
}

Deepesh125 (approved):

rohit01010 (approved):

kuntalghosh (approved):
# AI Code Review

Verdict: APPROVE

**Issue 1 - In `contrib/babelfishpg_tsql/runtime/functions.c` at line 6194: `initStringInfo` memory not freed on multi-result path:**

```c
initStringInfo(&buf);
for (i = 0; i < nitems; i++)
{
    if (!nulls[i])
    {
        text *fragment = DatumGetTextPP(elems[i]);
        appendBinaryStringInfo(&buf, VARDATA_ANY(fragment), VARSIZE_ANY_EXHDR(fragment));
    }
}
PG_RETURN_XML_P((xmltype *) cstring_to_text_with_len(buf.data, buf.len));
```

The `StringInfoData buf` allocates memory via `initStringInfo`, then `cstring_to_text_with_len` copies it into a palloc'd text datum. The original `buf.data` is not freed before returning. In a function called from SQL, this is acceptable since the memory context will be cleaned up after the query, but `pfree(buf.data)` before the return would be good practice to reduce peak memory usage for large XML results. This is advisory only, matching the pattern used in other functions in this file.

**Issue 2 - In `contrib/babelfishpg_tsql/runtime/functions.c` at line 6185: Edge case when all results are NULL:**

```c
if (nitems == 0)
    PG_RETURN_XML_P((xmltype *) cstring_to_text(""));

if (nitems == 1 && !nulls[0])
    PG_RETURN_DATUM(elems[0]);

initStringInfo(&buf);
for (i = 0; i < nitems; i++)
{
    if (!nulls[i])
    ...
}
PG_RETURN_XML_P((xmltype *) cstring_to_text_with_len(buf.data, buf.len));
```

If `nitems == 1` and `nulls[0]` is true, or if all elements in the array are NULL, the function falls through to the concatenation loop where nothing is appended to `buf`, and it returns an empty XML string (empty `buf.data`). This is functionally equivalent to the `nitems == 0` case and produces correct behavior (empty XML). However, it's worth adding a comment noting that the all-NULLs case intentionally falls through to produce empty XML, for future readers.

**Issue 3 - In `contrib/babelfishpg_tsql/sql/upgrades/babelfishpg_tsql--6.1.0--6.2.0.sql`: Aggregate signature change is a breaking upgrade concern (advisory):**

The upgrade correctly follows the deprecate-then-drop pattern (rename to `_deprecated_in_6_2_0`, drop `_deprecated_in_5_7_0`). The addition of the `auto_metadata text` parameter to `tsql_select_for_xml_agg` and `tsql_select_for_xml_text_agg` creates new 8-arg aggregates alongside the old 7-arg ones. This is well-structured for safe upgrades.

However, note that the `babelfish_drop_deprecated_object` calls reference `deprecated_in_5_7_0` — this correctly drops objects from a *previous* deprecation cycle, not the ones being deprecated in this upgrade. This is the standard two-version deprecation pattern.

### Notes

- Prior AI review issues have been addressed: `DirectFunctionCall3` is used, `pltsql_quoted_identifier` variable is used directly, TODO comment for XMLNAMESPACES is present, and the function was moved to `runtime/functions.c` as reviewers requested.
- The implementation follows the established 3-stage architecture (ANTLR rewriting → grammar mapping → C function) consistent with `.exist()` and `.value()`.
- Test coverage is comprehensive: 800+ lines of verify SQL covering basic queries, table columns, multiple elements, nested XML, attributes, special characters, Unicode, NULL handling, stored procedures, functions, views, computed columns, CASE/JOIN/UNION, error cases, and QUOTED_IDENTIFIER validation.
- Minor version upgrade tests are properly added to `upgrade/latest/schedule`.
- The `before-17_11-or-18_5` test variant correctly expects errors for procedures/functions that don't exist in pre-upgrade schemas.
- The `xml_exist` and `xml_value` test changes are correct — previously they errored with "XML QUERY is not currently supported" and now they either succeed (`.query('/artists').exist(...)`) or produce a proper downstream error (`could not parse XML document` for fragments that aren't valid documents).
- The removal of `INSTR_UNSUPPORTED_TSQL_XML_QUERY` from the enum and updating `tsqlUnsupportedFeatureHandler.cpp` to use `INSTR_UNSUPPORTED_TSQL_XML_MODIFY` for the `.modify()` case (instead of reusing the `_XML_QUERY` enum) is a clean fix.
- The FOR XML output changes (`xmlns:xsi` now appearing on the root element) are a side-effect of the `auto_metadata` parameter addition and appear correct for T-SQL compatibility.

PR #4784, opened 2026-04-28
[BABEL] Insert Exec PR4 temp table buffer

monk0707:
### Description

**Background: INSERT...EXEC Redesign**

The current INSERT...EXEC implementation has limitations with transaction handling, error recovery, and TRY-CATCH compatibility. We're redesigning it using a **temp table buffering approach**:
1. Procedure output is captured into a temporary buffer table via a custom DestReceiver
2. After procedure execution completes, buffered rows are flushed to the target table
3. This enables proper TRY-CATCH error handling and transaction semantics matching SQL Server

**PR Series Overview:**
| PR | Focus | Description |
|----|-------|-------------|
| PR0 | GUC | Feature flag to gate the new code path |
| PR1 | Parser | Parse INSERT...EXEC and extract target table info |
| PR2 | Executor | Core state management and schema detection |
| PR3 | DestReceiver | Capture procedure output to temp table |
| **PR4** | **Temp Table** | **Create/drop temp buffer table with correct schema** |
| PR5 | Flush | Transfer buffered data to target table |
| PR6 | Integration | Wire everything together with tests |

**PR Chain:** [← PR3](https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/4783) | [PR5 →](https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/4788)

**This PR:**

Implements temp table lifecycle management for INSERT EXEC buffering. The temp table buffers procedure output before flushing to the target table, enabling proper transaction semantics and TRY-CATCH support. This replaces the stub functions that previously returned InvalidOid/did nothing.

**Changes:**
- `create_insert_exec_temp_table()` - creates a PostgreSQL temp table with schema matching the target table
  - Uses unique naming pattern `__insert_exec_buf_PID_N` to avoid conflicts
  - For regular tables: queries `pg_attribute` to get column definitions (avoids SELECT permission requirement for ownership chaining)
  - For temp tables: uses `SELECT INTO` for simplicity
  - Excludes IDENTITY columns (auto-generated on insert)
- `drop_insert_exec_temp_table()` - drops the temp table by name from context

**Security:**
- Uses `quote_identifier()` and `quote_literal_cstr()` to prevent SQL injection

### Issues Resolved
PR4 of Task: [BABEL-5522](https://jira.rds.a2z.com/browse/BABEL-5522)

### Test Scenarios Covered
* **Use case based -** Tested via full feature integration in PR6
* **Boundary conditions -** Unique name generation with suffix increment
[...]

PR #4788, opened 2026-04-28
[BABEL] Insert Exec PR5 Flush

monk0707:
### Description

**Background: INSERT...EXEC Redesign**

The current INSERT...EXEC implementation has limitations with transaction handling, error recovery, and TRY-CATCH compatibility. We're redesigning it using a **temp table buffering approach**:
1. Procedure output is captured into a temporary buffer table via a custom DestReceiver
2. After procedure execution completes, buffered rows are flushed to the target table
3. This enables proper TRY-CATCH error handling and transaction semantics matching SQL Server

**PR Series Overview:**
| PR | Focus | Description |
|----|-------|-------------|
| PR0 | GUC | Feature flag to gate the new code path |
| PR1 | Parser | Parse INSERT...EXEC and extract target table info |
| PR2 | Executor | Core state management and schema detection |
| PR3 | DestReceiver | Capture procedure output to temp table |
| PR4 | Temp Table | Create/drop temp buffer table with correct schema |
| **PR5** | **Flush** | **Transfer buffered data to target table** |
| PR6 | Integration | Wire everything together with tests |

**PR Chain:** [← PR4](https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/4784) | [PR6 →](https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/4794)

**This PR:**

Adds the flush logic that transfers buffered data from the temp table to the target table, along with setup and cleanup helper functions.

**Changes:**
- `flush_insert_exec_temp_table()` - Flushes buffered rows using `INSERT INTO target SELECT * FROM temp_table`
  - Uses subtransaction for atomicity
  - Implements ownership chaining by temporarily clearing INSERT EXEC context during flush
  - Allows INSTEAD OF triggers to fire correctly
- `insert_exec_setup()` - Sets up execution context, creates temp table, returns DestReceiver
- `insert_exec_error_cleanup()` - Cleans up on error, drops temp table, releases target table lock
- `insert_exec_success_cleanup()` - Cleans up on success after flush completes
- Exports `commit_stmt()` and `exec_stmt_execsql()` from `pl_exec.c` for use by flush logic

### Issues Resolved
PR5 of Task: [BABEL-5522](https://jira.rds.a2z.com/browse/BABEL-5522)

### Test Scenarios Covered
* **Use case based -** Tests included in PR6 (Executor Integration + Tests)
* **Boundary conditions -** Covered in PR6
* **Arbitrary inputs -** Covered in PR6
* **Negative test cases -** Covered in PR6
* **Minor version upgrade tests -** N/A (new feature, GUC-gated)
* **Major version upgrade tests -** N/A (new feature, GUC-gated)
* **Performance tests -** N/A
* **Tooling impact -** None
* **Client tests -** Covered in PR6 via JDBC tests

### Check List

### DIFF
--- a/contrib/babelfishpg_tsql/Makefile
+++ b/contrib/babelfishpg_tsql/Makefile
@@ -80,6 +80,7 @@ OBJS += src/fts_parser.o
 OBJS += src/pltsql_partition.o
 OBJS += src/bbf_parallel_query.o
 OBJS += src/temp_table.o
+OBJS += src/pl_insert_exec.o
 
 export ANTLR4_JAVA_BIN=java
 export ANTLR4_RUNTIME_LIB=-lantlr4-runtime
--- a/contrib/babelfishpg_tsql/src/guc.c
+++ b/contrib/babelfishpg_tsql/src/guc.c
@@ -956,8 +956,8 @@ define_custom_variables(void)
 							 NULL,
 							 &pltsql_enable_new_insert_exec,
 							 false,
-							 PGC_USERSET,
-							 GUC_NOT_IN_SAMPLE,
+							 PGC_SUSET,
+							 GUC_NOT_IN_SAMPLE | GUC_DISALLOW_IN_FILE | GUC_DISALLOW_IN_AUTO_FILE,
 							 NULL, NULL, NULL);
 
 	DefineCustomBoolVariable("babelfishpg_tsql.noexec",
--- a/contrib/babelfishpg_tsql/src/guc.c
+++ b/contrib/babelfishpg_tsql/src/guc.c
@@ -59,6 +59,7 @@ bool		pltsql_disable_batch_auto_commit = false;
 bool		pltsql_disable_internal_savepoint = false;
 bool		pltsql_disable_txn_in_triggers = false;
 bool		pltsql_recursive_triggers = false;
+bool		pltsql_enable_new_insert_exec = false;
 bool		pltsql_noexec = false;
 bool		pltsql_showplan_all = false;
 bool		pltsql_showplan_text = false;
@@ -948,6 +949,15 @@ define_custom_variables(void)
 							 false,
 							 PGC_USERSET,
 							 GUC_NOT_IN_SAMPLE | GUC_DISALLOW_IN_FILE | GUC_DISALLOW_IN_AUTO_FILE,
+							 NULL, NULL, NULL);
+
+	DefineCustomBoolVariable("babelfishpg_tsql.enable_new_insert_exec",
+							 gettext_noop("Enables INSERT...EXEC redesign code path"),
+							 NULL,
+							 &pltsql_enable_new_insert_exec,
+							 false,
+							 PGC_USERSET,
+							 GUC_NOT_IN_SAMPLE,
 							 NULL, NULL, NULL);
 
 	DefineCustomBoolVariable("babelfishpg_tsql.noexec",
--- a/contrib/babelfishpg_tsql/src/guc.h
+++ b/contrib/babelfishpg_tsql/src/guc.h
@@ -24,6 +24,7 @@ extern bool pltsql_allow_windows_login;
 extern bool pltsql_allow_fulltext_parser;
 extern bool pltsql_weak_view_binding;
 extern bool pltsql_no_browsetable;
+extern bool pltsql_enable_new_insert_exec;
 extern char *pltsql_psql_logical_babelfish_db_name;
 extern int  pltsql_isolation_level_repeatable_read;
 extern int  pltsql_isolation_level_serializable;
--- a/contrib/babelfishpg_tsql/src/pl_exec.c
+++ b/contrib/babelfishpg_tsql/src/pl_exec.c
@@ -322,7 +322,7 @@ static int	exec_stmt_raise(PLtsql_execstate *estate,
 							PLtsql_stmt_raise *stmt);
 static int	exec_stmt_assert(PLtsql_execstate *estate,
 							 PLtsql_stmt_assert *stmt);
-static int	exec_stmt_execsql(PLtsql_execstate *estate,
+int	exec_stmt_execsql(PLtsql_execstate *estate,
 							  PLtsql_stmt_execsql *stmt);
 static void updateColumnUpdatedList(Query *query);
 static int	exec_stmt_dynexecute(PLtsql_execstate *estate,
@@ -4474,7 +4474,7 @@ execute_txn_command(PLtsql_execstate *estate, PLtsql_stmt_execsql *stmt)
  * is recreated when needed for cases like commit/
  * rollbck/rollback to savepoint
  */
-static void
+void
 commit_stmt(PLtsql_execstate *estate, bool txnStarted)
 {
 	SimpleEcontextStackEntry *topEntry = simple_econtext_stack;
@@ -4805,7 +4805,7 @@ setup_procedure_output_target_for_insert_exec(PLtsql_execstate *estate, PLtsql_s
  * needs to use that, fix those callers to push/pop stmt_mcontext.
  * ----------
  */
-static int
+int
 exec_stmt_execsql(PLtsql_execstate *estate,
 				  PLtsql_stmt_execsql *stmt)
 {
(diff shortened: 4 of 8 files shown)


kuntalghosh (changes requested):
# AI Code Review

Verdict: REQUEST_CHANGES

**Issue 1 (Blocking) - In `pl_insert_exec.c` at line 1891: Flush query uses unquoted identifiers**

```c
appendStringInfo(&flush_query,
    "INSERT INTO %s (%s) SELECT * FROM %s",
    target_table, column_list, temp_table_name);
```

`target_table` and `column_list` are interpolated with bare `%s` into the flush SQL, which is then executed under the procedure owner's identity via `SetUserIdAndSecContext`. If a crafted table or column name reaches this path, it becomes a privilege escalation vector. The `temp_table_name` is internally generated and safe, but `target_table` and `column_list` originate from user input. Even though the ANTLR parser stores structured components in `InsertExecInfo`, the executor flattens them into strings before calling `pltsql_set_insert_exec_context_info`. The flush function should defensively quote all identifiers.

Suggested fix:

```c
/* For target_table: use the structured InsertExecInfo components
 * with quote_qualified_identifier(), or at minimum: */
appendStringInfo(&flush_query,
    "INSERT INTO %s (%s) SELECT * FROM %s",
    quote_identifier(target_table), column_list, quote_identifier(temp_table_name));
```

For `column_list`, each column name should be individually quoted. Since the parser already stores columns as a `List` of strings in `InsertExecInfo.columns`, consider passing the structured list through to the flush function and building the column clause with `quote_identifier()` per column, rather than passing a pre-concatenated string.

**Issue 2 (Blocking) - In `pl_insert_exec.c` at line 1558: Raw column_list in SELECT INTO for temp table targets**

```c
appendStringInfo(&create_stmt,
    "SELECT %s INTO %s FROM %s WHERE 1=0",
    column_list,
    quote_identifier(temp_table_name),
    quote_identifier(target_table));
```

`column_list` is interpolated with bare `%s` into the SELECT clause. The non-temp-table branch correctly queries `pg_attribute` to avoid this, but the temp-table branch uses raw string interpolation. This is the same class of issue as Issue 1.

Suggested fix:

Pass the structured column list (from `InsertExecInfo.columns`) and build the SELECT clause with `quote_identifier()` per column name, matching the safe approach used in the regular-table branch.

**Issue 3 (Blocking) - In `pl_insert_exec.c` at line 1915: `get_func_owner()` return value not checked before `SetUserIdAndSecContext`**

```c
SetUserIdAndSecContext(get_func_owner(estate->func->fn_oid),
                       flush_save_sec_context | SECURITY_LOCAL_USERID_CHANGE);
```

`get_func_owner()` returns `InvalidOid` on syscache miss (e.g., if the function was dropped concurrently). Passing `InvalidOid` to `SetUserIdAndSecContext` could break ACL enforcement or cause undefined behavior during the flush INSERT.

Suggested fix:

```c
Oid func_owner = get_func_owner(estate->func->fn_oid);
if (!OidIsValid(func_owner))
    elog(ERROR, "INSERT-EXEC: could not determine owner of function %u",
         estate->func->fn_oid);
GetUserIdAndSecContext(&flush_save_userid, &flush_save_sec_context);
SetUserIdAndSecContext(func_owner,
                       flush_save_sec_context | SECURITY_LOCAL_USERID_CHANGE);
```

**Issue 4 (Blocking) - In `pl_insert_exec.c` at line 1855: `flush_save_userid` and `flush_save_sec_context` should be `volatile`**

```c
Oid         flush_save_userid = InvalidOid;
int         flush_save_sec_context = 0;
```

These variables are set inside `PG_TRY` and read inside `PG_CATCH` to restore the security context. Without `volatile`, the compiler may optimize them into registers that are clobbered by `siglongjmp`, causing stale values to be restored. Note that `subtxn_started` and `flush_switched_context` are already correctly declared `volatile`.

Suggested fix:

```c
volatile Oid    flush_save_userid = InvalidOid;
volatile int    flush_save_sec_context = 0;
```

**Issue 5 (Blocking) - In `pl_insert_exec.c` at line 426: Dead-code prefix check in `pltsql_insert_exec_check_pending_drop`**

```c
if (strncmp(insert_exec_ctx.temp_table_name, "#__insert_exec_buf_", 19) != 0)
```

The temp table name is generated at line 1497 as `"__insert_exec_buf_%d_%d"` — without a `#` prefix. This means the defensive check always triggers, logging a spurious warning and skipping the drop. Pending temp tables are never cleaned up.

Suggested fix:

```c
if (strncmp(insert_exec_ctx.temp_table_name, "__insert_exec_buf_", 18) != 0)
```

**Issue 6 (Advisory) - In `pl_insert_exec.c` at lines 549, 574, 709, 901: `return` inside `PG_CATCH` blocks**

```c
PG_CATCH();
{
    FlushErrorState();
    return;
}
PG_END_TRY();
```

While PostgreSQL's `PG_CATCH` macro does restore `PG_exception_stack` before entering the catch block (making `return` technically safe), this pattern is fragile and deviates from PostgreSQL coding conventions. It makes it easy to accidentally skip cleanup code added after `PG_END_TRY()` in future maintenance. Consider using a flag variable and falling through to `PG_END_TRY()` instead.

**Issue 7 (Advisory) - In `pl_insert_exec.c`: TopMemoryContext allocations without a dedicated child context**

Multiple functions allocate in `TopMemoryContext` (lines 237, 251, 590, 610, etc.) with manual `pfree()` cleanup. If any error path skips cleanup, these allocations leak permanently for the backend's lifetime. A safer pattern is to create a dedicated child context (e.g., `InsertExecMemoryContext`) that can be destroyed in bulk during cleanup, similar to how PostgreSQL's bgwriter uses a child context.

**Issue 8 (Advisory) - In `pl_insert_exec.c` at line 1594: `strtok_r` column-list parsing breaks quoted identifiers**

```c
col_name = strtok_r(col_list_copy, ", ", &saveptr);
```

`strtok_r` with `", "` as delimiters splits on both comma and space individually. Column names containing spaces (e.g., bracket-quoted `[My Column]`) or commas would be incorrectly split. The comment at line 1590 acknowledges this limitation. Since the parser already provides a structured `List` of column names in `InsertExecInfo.columns`, consider passing that list through to `create_insert_exec_temp_table` instead of re-parsing a flat string.

**Issue 9 (Advisory) - In `tsqlIface.cpp` at line 1649: `replaceGraftedStatement` appends to end of sibling list**

```c
siblings = list_delete_ptr(siblings, old_stmt);
siblings = lappend(siblings, new_stmt);
```

When replacing the old `PLtsql_stmt_execsql` with the new `PLtsql_stmt_exec`, the new statement is appended to the end of the sibling list rather than inserted at the original position. If the INSERT EXEC is not the last statement in the batch, this changes execution order. Consider using `list_insert_nth` or tracking the position before deletion.

Suggested fix:

```cpp
int pos = list_cell_number(siblings, list_cell_of(siblings, old_stmt));
siblings = list_delete_ptr(siblings, old_stmt);
siblings = list_insert_nth(siblings, pos, new_stmt);
```

**Issue 10 (Advisory) - In `pl_insert_exec.c` at line 1217: DestReceiver uses `mydest = DestNone`**

```c
self->pub.mydest = DestNone;
```

`DestNone` signals "discard all output" in PostgreSQL's executor. While the custom `receiveSlot` callback overrides this behavior, some executor code paths check `mydest` to decide whether to call the receiver at all. This could cause silent row drops in edge cases (e.g., parallel query workers). Consider using `DestRemote` or another appropriate destination type, or document why `DestNone` is safe here.

### Missing Tests
- Tests are deferred to PR6 (Executor Integration + Tests). This is acceptable for a stacked PR series, but the flush logic, schema verification, and error cleanup paths should have dedicated test coverage when PR6 lands. Specifically: column mismatch errors, schema change detection, ownership chaining, and the pending-drop cleanup path.

### Notes
- The PR description is well-structured with clear context about the PR series and what this PR adds. BABEL-5522 JIRA reference is present.
- Target branch `BABEL_6_X_DEV` is correct for new feature development.
- The GUC gating (`babelfishpg_tsql.enable_new_insert_exec`, default false, superuser-only) is a good safety measure for incremental rollout.
- The `InsertExecInfo` struct in `pltsql-2.h` correctly stores table components separately for SQL safety, but the C executor code in `pl_insert_exec.c` still receives flattened strings. The structured fields should be threaded through to the SQL-building functions to fully realize the injection prevention.
- The `commit_stmt` and `exec_stmt_execsql` exports from `pl_exec.c` (changing `static` to extern) are clean and minimal.

PR #4791, opened 2026-04-30

Mohitraj-g:
### Description

```
-- CASE: fails
```
```
-- COALESCE: fails
```

### Fix

 `select_common_type_for_coalesce_function` also includes a minor cleanup: the `UNKNOWNOID` (string literal) handling was simplified by replacing `ptype` in-place and falling through to the shared precedence comparison, instead of duplicating the comparison inline with a separate `continue`. This is semantically equivalent and notstrictly required for the sql_variant fix.

### Issues Resolved

BABEL-6402
BABEL-6427

cherry-picked from https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/4706

* **Use case based -**
CASE (simple, searched, multi-branch with sql_variant in 1st/2nd/3rd/last position), ISNULL, COALESCE (2-4 args), UNION ALL, UNION, INTERSECT,  EXCEPT, VALUES — all with sql_variant mixed with bit, int, varchar, float, numeric, date, datetime, time, money, smallmoney, bigint, smallint, tinyint, uniqueidentifier

* **Boundary conditions -**
NULL sql_variant (CAST(NULL AS sql_variant)) in every position for all contexts; all-NULL sql_variant args; NULL sql_variant mixed with  NULL other types 

* **Arbitrary inputs -**
SERVERPROPERTY (system functions returning sql_variant); string literals mixed with sql_variant; table column references

* **Negative test cases -**
ISNULL with sql_variant as 2nd arg (errors preserved); incompatible types (date vs numeric) still error correctly 

`babel_sqlvariant_coerce_type_before_16_14-or-17_10-or-18_4` expected output covers pre-fix behavior (view creation fails, ISNULL errors) 

`babel_sqlvariant_coerce_type` in latest upgrade schedule with post-fix expected output

* **Performance tests -**

* **Tooling impact -**
UDT based on sql_variant and UDT based on int tested against sql_variant in CASE, COALESCE, UNION 

* **Client tests -**
Covered via JDBC test framework  

### Check List


kuntalghosh (approved):
# AI Code Review

Verdict: APPROVE

**Advisory 1 — In `contrib/babelfishpg_tsql/src/pltsql_coerce.c`, `select_common_type_setop` pre-scan: returning the UDT OID instead of the base sql_variant OID:**

```c
Oid		base_type = get_immediate_base_type_of_UDT_internal(type);

if (OidIsValid(base_type))
    type = base_type;

if ((*common_utility_plugin_ptr->is_tsql_sqlvariant_datatype)(type))
{
    if (which_expr)
        *which_expr = expr;
    return type;
}
```

When the expression is a UDT based on sql_variant (e.g., `babel_sqlvariant_udt_sv`), `type` is reassigned to the base sql_variant OID before the check. The function then returns the base OID, not the original UDT OID. This is consistent with how the main loop already resolves UDTs for CASE expressions, and the test output confirms the result type is `sql_variant` (not the UDT name), so this is correct behavior. Just noting it for clarity — the caller sees the resolved base type, which is the right thing to do for type resolution.

**Advisory 2 — In `contrib/babelfishpg_tsql/src/pltsql_coerce.c`, `select_common_type_for_coalesce_function`: UNKNOWNOID check after NULL sql_variant fall-through:**

```c
{
    if (!(*common_utility_plugin_ptr->is_tsql_sqlvariant_datatype)(ptype))
        continue;
}

/* If the arg is non-null string literal */
if (ptype == UNKNOWNOID)
    ptype = get_sys_varcharoid();
```

When a NULL sql_variant falls through (skips the `continue`), it proceeds to the `UNKNOWNOID` check. Since `CAST(NULL AS sql_variant)` has the sql_variant OID (not `UNKNOWNOID`), this path is safe — the `UNKNOWNOID` branch won't trigger. The comment "non-null string literal" is slightly misleading now that NULL sql_variant also falls through, but the logic is correct.

**Advisory 3 — COALESCE refactoring of UNKNOWNOID handling is a clean simplification:**

The old code duplicated the precedence comparison for `UNKNOWNOID` (string literal) arguments inline with a separate `continue`. The new code replaces `ptype` in-place with `get_sys_varcharoid()` and falls through to the shared precedence comparison. This is semantically equivalent and reduces code duplication. Good cleanup.

**Advisory 4 — In `test/python/expected/upgrade_validation/expected_dependency.out`: removal of `Function sys.bit_sqlvariant`:**

The removal indicates the new upgrade test now exercises the `bit_sqlvariant` cast path (via `CAST(x AS bit)` mixed with sql_variant in CASE/COALESCE/UNION). Similarly, `serverproperty` is removed from `expected_create.out` because the new test uses `SERVERPROPERTY()`. Both removals are justified by the test coverage.

### Missing Tests

No missing tests identified. The test coverage is thorough:

- All six type-resolution contexts are covered: CASE (simple and searched), ISNULL, COALESCE, UNION/UNION ALL, INTERSECT, EXCEPT, and VALUES.
- sql_variant is tested in every positional slot (1st, 2nd, 3rd, last) for multi-branch expressions.
- NULL sql_variant (`CAST(NULL AS sql_variant)`) is tested in all positions for all contexts.
- UDT-based-on-sql_variant and UDT-based-on-int are tested against sql_variant.
- Regression tests for the original reported queries (`SERVERPROPERTY` in CASE) are included.
- Char-type-only expressions are verified to be unaffected (sections 9, 11).
- Incompatible type errors are verified to still fire correctly (section 12).
- Additional data types (date, datetime, time, money, smallmoney, bigint, smallint, tinyint, uniqueidentifier) are covered.
- Minor version upgrade tests (`before_16_14-or-17_10-or-18_4`) with pre-fix expected output are provided.
- Major version upgrade test (`babel_sqlvariant_coerce_type` in `latest` and `16_14` schedules) is included.
- The `before` variant correctly shows pre-fix behavior (view creation failure for `babel_case_sqlvariant_v1`, `relation does not exist` error when querying it).

### Notes

The fix approach is sound. The `select_common_type_setop` function previously had a structural problem: its main loop returns `InvalidOid` as soon as it encounters any non-char type, which means if `bit` (or any non-char type) appears before sql_variant in the expression list, the function bails out before ever seeing sql_variant. The pre-scan loop correctly addresses this by checking for sql_variant upfront, which is justified because sql_variant has the highest T-SQL type precedence (0) and can hold any type.

The COALESCE fix is more targeted: the existing NULL-skipping logic was discarding the type information of NULL sql_variant arguments. Since sql_variant has the highest precedence, a `CAST(NULL AS sql_variant)` should still influence the common type even though its value is NULL. The fix correctly allows NULL sql_variant to participate in the precedence comparison while still skipping other NULL types.

The UDT resolution in both functions (`get_immediate_base_type_of_UDT_internal`) ensures that UDTs based on sql_variant are correctly recognized, which is consistent with how the existing code handles UDTs for CASE expressions in the main loop.

Upgrade schedule entries are correctly placed: pre-16.14/17.10/18.4 versions use the `before` variant, and 16.14+ and `latest` use the main test.

Yvinayak07 (approved):

shardgupta (approved):

PR #4793, opened 2026-05-01
Implement four-part-name remote procedure execution for linked servers

fizzlr:
__What:__ Currently, Babelfish supports linked servers for ad-hoc queries via `OPENQUERY()` but does not support executing remote stored procedures using four-part names (`EXEC server.database.schema.procedure`). With this change, Babelfish now supports remote procedure execution through linked servers using the T-SQL four-part naming convention. Users can call stored procedures on remote server instances with full parameter binding, result set streaming, OUTPUT parameter retrieval, and return status capture — matching the T-SQL `EXEC linked_server.db.schema.proc` behavior.

__Why:__ Remote procedure execution via four-part names is a core linked server capability that customers rely on for cross-server workflows. Without it, applications that call remote stored procedures through linked servers cannot migrate to Babelfish. This bridges a significant compatibility gap in the linked server feature set.

__How (high-level roadmap):__

Parser (tsqlIface.cpp, tsqlUnsupportedFeatureHandler.cpp):
- In EXEC context: extract server_name from four-part names and store in PLtsql_stmt_exec; 
  route to remote execution path at runtime
- Remove blanket unsupported-feature block for 4-part EXEC names from
  tsqlUnsupportedFeatureHandler (keep block for function-call context)
- Improve error messages: distinguish XML method calls (schema.table.column.method)
  from remote procedure references in both EXEC and function-call contexts

Runtime (pl_exec-2.c):
- New execute_remote_procedure_rpc() function: opens TDS connection, binds parameters
  via FreeTDS dbrpcparam, executes RPC, streams first result set via Portal/tuplestore,
  retrieves OUTPUT parameters and return status
- Runtime guards: check linked servers enabled, enable_remote_proc_exec GUC,
  rpc_out server option, and transaction state (NestedTranCount)
- Note: Only single result set is currently supported per RPC call

Infrastructure (linked_servers.c/h):
- RPC error capture mode: msg_handler saves first remote error, err_handler formats it
  as SQL Server error 7215 using ERRCODE_PLTSQL_RAISERROR for TRY/CATCH compatibility
- Type mapping (get_tds_type_from_pg_oid) and datum-to-TDS-bytes conversion
  (convert_datum_to_tds_bytes): native binary for int/float/bit/binary;
  string-based fallback for numeric, money, datetime variants, uniqueidentifier
- TDS type/RPC/OUTPUT abstraction macros wrapping FreeTDS API
- Set client charset to UTF-8 for proper multi-byte character handling
  (affects all linked server connections including OPENQUERY)
- Make linked_server_establish_connection() externally visible

Catalog (catalog.c/h, procedures.c):
- Add rpc_out boolean column to babelfish_server_options table
- Add get_rpc_out_option() catalog lookup function
- Support 'rpc out' in sp_serveroption (true/false/yes/no)
- Update sys.servers view: is_rpc_out_enabled reads from rpc_out column

GUCs (guc.c/h):
- babelfishpg_tsql.enable_remote_proc_exec (bool, default true)
- escape_hatch_remote_proc_transaction (strict/ignore) to control behavior
  when RPC is called inside an explicit transaction

Error mapping (error_mapping.txt):
- Map errors 7215, 7412, 7202 for TDS protocol compatibility

Upgrade (babelfishpg_tsql--5.4.0--5.5.0.sql):
- ALTER TABLE to add rpc_out column; recreate sys.servers view

Known limitations:
- Only first result set is streamed (no multi-result-set support)
- Distributed transactions not supported (no MS DTC)

### Issues Resolved

N/A (new feature)

### Test Scenarios Covered

- __Use case based —__ Remote procedure calls with various parameter types (int, varchar, nvarchar, float, datetime, numeric, money, bit, binary, uniqueidentifier); procedures returning result sets; procedures with OUTPUT parameters; procedures with RETURN status codes; procedures with no parameters; procedures that raise errors

- __Boundary conditions —__ NULL parameter values; empty string parameters; max-length varchar parameters; procedures returning zero rows; procedures returning multiple columns; mixed INPUT/OUTPUT parameters

- __Arbitrary inputs —__ Multi-byte UTF-8 characters (CJK, special characters) in nvarchar parameters; various numeric precision/scale combinations

- __Negative test cases —__ Calling non-existent procedures; RPC on server with `rpc out` disabled (error 7412); execution within explicit transactions (blocked by default with `escape_hatch_remote_proc_transaction = strict`); calling with linked servers disabled; calling with `enable_remote_proc_exec` disabled; remote procedures that RAISERROR (error 7215 with TRY/CATCH compatibility)

- __Minor version upgrade tests —__ `remote-proc-exec_upgrd_before_17_6` upgrade test verifies the `rpc_out` catalog column addition and view recreation survive minor version upgrade

- __Major version upgrade tests —__ Not applicable for this change

- __Performance tests —__ Not applicable (RPC execution is network-bound; no local performance-sensitive paths added)

- __Tooling impact —__ None

- __Client tests —__ Loopback tests (`remote-proc-exec-loopback`) exercise the full end-to-end path through JDBC client → Babelfish → FreeTDS → remote TDS server (loopback to self)


kuntalghosh (approved):
NA

kuntalghosh (dismissed):
# AI Code Review

Verdict: COMMENT

**Issue 1 - In `contrib/babelfishpg_tsql/src/linked_servers.c`: TopMemoryContext allocations in RPC error capture without guaranteed cleanup in all paths:**

```c
if (rpc_error_capture_mode && !rpc_error_captured)
{
    rpc_error_captured = true;
    rpc_error_code = error_code;
    ...
    rpc_error_msg = error_msg ? MemoryContextStrdup(TopMemoryContext, error_msg) : NULL;
    rpc_error_svr_name = svr_name ? MemoryContextStrdup(TopMemoryContext, svr_name) : NULL;
    rpc_error_proc_name = proc_name ? MemoryContextStrdup(TopMemoryContext, proc_name) : NULL;
```

The RPC error capture strings (`rpc_error_msg`, `rpc_error_svr_name`, `rpc_error_proc_name`, `rpc_error_linked_server_name`) are allocated in `TopMemoryContext`. While `linked_server_clear_rpc_error()` explicitly frees them, and `PG_FINALLY` calls it, there's a known anti-pattern risk: if the backend encounters a FATAL error or crashes between allocation and cleanup, these leak for the backend lifetime. More importantly, if `linked_server_msg_handler` is invoked multiple times in capture mode (the `!rpc_error_captured` guard prevents _overwriting_ but does not prevent the _first_ allocation from leaking if the error path never reaches `linked_server_clear_rpc_error` due to an unrelated FATAL). Consider using a dedicated child memory context that can be destroyed atomically, or document the rationale for TopMemoryContext explicitly (similar to how BulkCopyStmt justifies it).

**Issue 2 - In `contrib/babelfishpg_tsql/src/pl_exec-2.c`: Indentation inconsistency in the `rpc_out` check:**

```c
		/* Check if RPC out option is enabled for this server */
	if (!get_rpc_out_option(stmt->server_name))
		ereport(ERROR,
				(errcode(ERRCODE_FDW_ERROR),
				 errmsg("Server '%s' is not configured for RPC OUT.", stmt->server_name),
					 errhint("Execute: EXEC sp_serveroption '%s', 'rpc out', 'true'", 
							 stmt->server_name)));
```

This block uses inconsistent indentation — the `if` statement is indented with one tab while the surrounding code uses two tabs. This will fail pgindent or at minimum causes visual confusion. The `errhint` is also indented inconsistently relative to the `errmsg`.

Suggested fix:
```c
		/* Check if RPC out option is enabled for this server */
		if (!get_rpc_out_option(stmt->server_name))
			ereport(ERROR,
					(errcode(ERRCODE_FDW_ERROR),
					 errmsg("Server '%s' is not configured for RPC OUT.", stmt->server_name),
					 errhint("Execute: EXEC sp_serveroption '%s', 'rpc out', 'true'",
							 stmt->server_name)));
```

**Issue 3 - In `contrib/babelfishpg_tsql/src/linked_servers.c`: Excessive `/* LOG: ... */` comments throughout `linked_server_establish_connection()`:**

```c
		/* LOG: Foreign server retrieved successfully */
		...
		/* LOG: User mapping retrieved successfully */
		...
		/* LOG: TDS library initialization successful */
		...
		/* LOG: Connection parameters configured */
		...
		/* LOG: Initiating TDS connection */
		...
		/* LOG: Connection successful */
		...
		/* LOG: Connection ready */
```

These `/* LOG: ... */` comments appear to be development trace markers left in the final submission. They add noise without providing meaningful documentation value — the adjacent `LINKED_SERVER_DEBUG()` calls already document what's happening. Some are orphaned comments with no associated code (just marking a spot between existing statements). Consider removing the bare `/* LOG: ... */` markers that don't precede any code, and keeping only the meaningful `LINKED_SERVER_DEBUG()` calls which already serve as self-documenting trace points.

**Issue 4 - In `contrib/babelfishpg_tsql/src/pl_exec-2.c`: `MAX_COLS_SELECT` used without visible bounds check:**

```c
void *val[MAX_COLS_SELECT];
...
for (i = 0; i < colcount; i++)
{
    ...
    val[i] = LINKED_SERVER_DATA(lsproc, i + 1);
```

If the remote procedure returns more than `MAX_COLS_SELECT` columns, this is a stack buffer overflow. The existing OPENQUERY code has the same pattern, so this isn't new to this PR, but since this is new code being added, it would be good to add a bounds check: `if (colcount > MAX_COLS_SELECT) ereport(ERROR, ...)` before the loop, similar to what defensive code should do.

Suggested fix (add before the for loop):
```c
if (colcount > MAX_COLS_SELECT)
    ereport(ERROR,
            (errcode(ERRCODE_PROGRAM_LIMIT_EXCEEDED),
             errmsg("remote procedure returned %d columns, maximum supported is %d",
                    colcount, MAX_COLS_SELECT)));
```

**Issue 5 - In `contrib/babelfishpg_tsql/src/linked_servers.c`: Connection not cleaned up when `linked_server_establish_connection` succeeds but RPC init fails:**

```c
linked_server_establish_connection(stmt->server_name, &lsproc, false);

/* Initialize RPC call on clean connection */
if (LINKED_SERVER_RPC_INIT(lsproc, full_proc_name) != SUCCEED)
    ereport(ERROR,
            (errcode(ERRCODE_FDW_UNABLE_TO_CREATE_EXECUTION),
             errmsg("Failed to initialize RPC call for procedure %s", full_proc_name)));
```

This is actually fine because the `PG_FINALLY` block will clean up `lsproc` via `LINKED_SERVER_CANCEL` + `LINKED_SERVER_CLOSE`. The error from `ereport(ERROR)` will jump to `PG_FINALLY`. No issue here — the cleanup design is correct.

**Issue 6 - In `contrib/babelfishpg_tsql/src/tsqlIface.cpp`: Duplicate XML method detection logic:**

```cpp
bool is_xml_method = (pg_strcasecmp(proc_name.c_str(), "exist") == 0) ||
                     (pg_strcasecmp(proc_name.c_str(), "value") == 0) ||
                     (pg_strcasecmp(proc_name.c_str(), "query") == 0) ||
                     (pg_strcasecmp(proc_name.c_str(), "nodes") == 0) ||
                     (pg_strcasecmp(proc_name.c_str(), "modify") == 0);
```

This XML method name check is duplicated in two locations (`tsqlCommonMutator` and `handleGeospatialFunctionsInFunctionCall`). Consider extracting to a helper function like `is_xml_method_name(const char *name)` to avoid drift between the two copies if new methods are added in the future. This is a minor maintainability suggestion.

**Issue 7 - PR Description: Missing BABEL JIRA reference:**

The PR description says "Issues Resolved: N/A (new feature)" without a BABEL JIRA ticket reference. Per project conventions, every PR should link a BABEL-XXXX ticket unless tagged `[OSS-ONLY]`. This PR is a significant new feature and should have a tracking ticket.

### Notes

- The overall design is well-structured: proper separation between parser (tsqlIface.cpp), runtime (pl_exec-2.c), and infrastructure (linked_servers.c/h). The use of parameterized RPC binding via `dbrpcparam()` correctly avoids SQL injection. The PG_TRY/PG_FINALLY cleanup pattern is correct.

- The test coverage is comprehensive: loopback tests exercise parameters, OUTPUT, return codes, multi-result-sets, error propagation (errno preservation), transaction safety, and the GUC/escape-hatch controls. The upgrade test covers the catalog schema change.

- The UTF-8 charset setting (`LINKED_SERVER_SET_CHARSET(login, "UTF-8")`) affects all linked server connections including OPENQUERY — this is documented in the PR description and is the correct approach for multi-byte character support.

- The connection lifecycle (CANCEL + CLOSE in PG_FINALLY) correctly addresses the known anti-pattern of leaking idle connections from linked server operations (historically tracked in BABEL-5376).

- The `linked_server_establish_connection()` function was made externally visible (`static` removed). This is appropriate since the RPC path in `pl_exec-2.c` needs to call it directly.

fizzlr:
All 7 findings have been addressed — summary below:

  Issue 1 — TopMemoryContext allocations in RPC error capture: Addressed. Verified the cleanup is sound: linked_server_clear_rpc_error() frees all four strings and is called unconditionally
  in PG_FINALLY, so a normal ERROR always cleans up. A FATAL exit reclaims the entire backend address space (including TopMemoryContext), so there's no lasting leak. The !rpc_error_captured
  guard also prevents re-allocation. No change required.

  Issue 2 — Indentation in the rpc out check: Addressed. Re-indented the block to match the sibling check (if at 2 tabs, aligned errmsg/errhint) and removed the stray trailing whitespace.

  Issue 3 — Bare /* LOG: ... */ markers: Addressed. Removed the orphaned trace-marker comments that didn't precede any statement; kept the ones that act as section headers above a
  LINKED_SERVER_DEBUG() call.

  Issue 4 — MAX_COLS_SELECT bounds check: Fixed in 3c167b433. Added an explicit colcount > MAX_COLS_SELECT guard before the fixed-size stack array in execute_remote_procedure_rpc(), raising
  an error rather than relying on the remote server to honor the 4096-column limit. CI is green.

  Issue 5 — Connection cleanup on RPC-init failure: Addressed/confirmed. The establish_connection + RPC_INIT calls live inside the same PG_TRY whose PG_FINALLY runs CANCEL + CLOSE on the
  handle, so the ereport(ERROR) path cleans up correctly — no leak. No change required.

  Issue 6 — Duplicate XML-method detection: Addressed. Consolidated the exist/value/query/nodes/modify check into a single helper used by both call sites to avoid future drift.

  Issue 7 — Missing tracking reference: Addressed. Tracking ticket reference added.

  All items are resolved. Thanks again for the careful pass.

contrib/babelfishpg_tsql/src/linked_servers.c
```diff
@@ -815,6 +927,7 @@ linked_server_establish_connection(char *servername, LinkedServerProcess * lspro
 
 		LINKED_SERVER_SET_APP(login);
 		LINKED_SERVER_SET_VERSION(login);
+		LINKED_SERVER_SET_CHARSET(login, "UTF-8");
```
forestkeeper, line 942: Will this line be harmful to 3 parts remote procedure call or other linked server usage as well ?

forestkeeper: linked_server_establish_connection() is the shared connection setup used by both RPC and the pre-existing OPENQUERY() path (contrib/babelfishpg_tsql/src/linked_servers.c). There's no per-operation distinction — the login object is built once and reused. So the moment you set UTF-8 there, every existing OPENQUERY connection now also declares UTF-8 as its client charset, where before it inherited whatever the prior effective default was (FreeTDS compiled-in default, or client charset from freetds.conf, or the connection default).

forestkeeper: What a de-risking regression test should assert
A focused OPENQUERY test (ideally one whose expected output predates the PR, or a new one with known-correct expected bytes) covering:

Both type classes: a non-Unicode varchar literal and an nvarchar (N'…') literal in the same OPENQUERY, e.g. SELECT * FROM OPENQUERY(srv, 'SELECT ''café'' AS c, N''日本語'' AS d')

contrib/babelfishpg_tsql/src/linked_servers.c
```diff
@@ -79,9 +105,31 @@ linked_server_msg_handler(LinkedServerProcess lsproc, int error_code, int state,
...
+		 * errors from the same execution are ignored).
+		 * OPENQUERY uses the default path (rpc_error_capture_mode = false).
+		 */
+		if (rpc_error_capture_mode && !rpc_error_captured)
+		{
+			rpc_error_captured = true;
+			rpc_error_code = error_code;
+			rpc_error_state = state;
+			rpc_error_severity = severity;
+			rpc_error_msg = error_msg ? MemoryContextStrdup(TopMemoryContext, error_msg) : NULL;
```
forestkeeper, line 133: Why it used TopMemoryContext in here ?

contrib/babelfishpg_tsql/src/tsqlIface.cpp
```diff
@@ -1000,6 +1001,23 @@ class tsqlCommonMutator : public TSqlParserBaseListener
...
+				bool is_xml_method = (pg_strcasecmp(proc_name.c_str(), "exist") == 0) ||
+				                     (pg_strcasecmp(proc_name.c_str(), "value") == 0) ||
+				                     (pg_strcasecmp(proc_name.c_str(), "query") == 0) ||
+				                     (pg_strcasecmp(proc_name.c_str(), "nodes") == 0) ||
+				                     (pg_strcasecmp(proc_name.c_str(), "modify") == 0);
+
+				if (is_xml_method)
+					throw PGErrorWrapperException(ERROR, ERRCODE_FEATURE_NOT_SUPPORTED, "XML method calls with schema-qualified column references (schema.table.column.method) are not currently supported in Babelfish", getLineAndPos(ctx));
+				else
+					throw PGErrorWrapperException(ERROR, ERRCODE_FEATURE_NOT_SUPPORTED, "Remote procedure/function reference with 4-part object name is not currently supported in Babelfish", getLineAndPos(ctx));
```
forestkeeper: There's currently no test case for this error msg

contrib/babelfishpg_tsql/src/tsqlIface.cpp
```diff
@@ -10030,7 +10061,33 @@ handleGeospatialFunctionsInFunctionCall(TSqlParser::Function_callContext *ctx)
...
+			bool is_xml_method = (pg_strcasecmp(method_name.c_str(), "exist") == 0) ||
+			                     (pg_strcasecmp(method_name.c_str(), "value") == 0) ||
+			                     (pg_strcasecmp(method_name.c_str(), "query") == 0) ||
+			                     (pg_strcasecmp(method_name.c_str(), "nodes") == 0) ||
+			                     (pg_strcasecmp(method_name.c_str(), "modify") == 0);
+
+			if (is_xml_method)
+				throw PGErrorWrapperException(ERROR, ERRCODE_FEATURE_NOT_SUPPORTED, "XML method calls with schema-qualified column references (schema.table.column.method) are not currently supported in Babelfish", getLineAndPos(ctx));
+			else
+				throw PGErrorWrapperException(ERROR, ERRCODE_FEATURE_NOT_SUPPORTED, "Remote procedure/function reference with 4-part object name is not currently supported in Babelfish", getLineAndPos(ctx));
```
forestkeeper: There's currently no test case for this error msg 
also this msg looks confusing, as we already supported 4-part obj name.

contrib/babelfishpg_tsql/src/linked_servers.c
fizzlr: Yes that this is shared — you're right that linked_server_establish_connection() is the single setup path for both OPENQUERY and the new RPC path, the login object is built once,
  to that client charset before handing data to us; VARCHAR/CHAR go through the same conversion. Our backend encoding is UTF-8, so UTF-8 is the correct client charset. FreeTDS's prior
  effective default was nl_langinfo(CODESET), falling back to ISO-8859-1 under a C/POSIX locale. So:
  - UTF-8 locale (our CI / typical deployments): already UTF-8, so this line is a no-op.
  - C/POSIX locale: prior default was ISO-8859-1, under which multibyte NVARCHAR from OPENQUERY would already be corrupted against the UTF-8 backend, so forcing UTF-8 is a fix, not a
  regression.

  On the de-risking test: the pre-PR baseline you asked for already exists — openquery-vu-verify round-trips VARCHAR/NVARCHAR columns seeded with multibyte emoji (VARCHAR_dt1..4), with
  expected output committed before this PR, and the PR doesn't touch that file. I also added the explicit literal case you described: SELECT * FROM OPENQUERY(bbf_server, 'SELECT ''café'' AS
  c, N''日本語'' AS d') in openquery-vu-verify.mix — it round-trips correctly (header varchar#!#nvarchar, data café#!#日本語).

contrib/babelfishpg_tsql/src/linked_servers.c
fizzlr: Because a shorter-lived context would risk a use-after-free, not just to be safe. The captured strings are stashed in file-static globals inside the FreeTDS message-handler callback, but
  they aren't consumed until later, in linked_server_throw_rpc_error() — after the RPC result-draining loop has run SPI / Portal / tuplestore work that can switch or reset
  CurrentMemoryContext. A short-lived allocation could be freed before it's read, and the matching pfree() in linked_server_clear_rpc_error() would then fault. TopMemoryContext (backend
  lifetime) avoids that, and there's no leak: execute_remote_procedure_rpc()'s PG_FINALLY unconditionally clears them on both success and error paths, and the clear is idempotent.

  Added a comment at the allocation site spelling this out. I kept TopMemoryContext rather than a dedicated child context — a child would need a backend-lifetime parent anyway to avoid
  reintroducing the same dangling window, so it'd be extra machinery for three short strings with no real benefit.

contrib/babelfishpg_tsql/src/tsqlIface.cpp
fizzlr: This fires for a 4-part name used as a function call in expression position (e.g. SELECT srv.db.schema.func(1)), which is distinct from the now-supported EXEC server.db.schema.proc. The
  literal message was actually already exercised by BABEL-2455 (select yourserver.master.dbo.f1(1)), but the feature's own suite had no case, so I added a negative test to
  four-part-names-vu-verify.sql:

    SELECT bbf_fpn_server.master.dbo.fpn_remote_func(1)   -- rejected at parse time

  I also sharpened the message to point at the supported path: "Remote function calls using a 4-part object name (server.database.schema.function) are not currently supported in Babelfish;
  use EXEC to invoke a remote procedure".

contrib/babelfishpg_tsql/src/tsqlIface.cpp
fizzlr: This throw is in handleGeospatialFunctionsInFunctionCall, so it fires on a method invoked on a multi-part qualified column like dbo.tbl.col.STIntersects()
  — a spatial/CLR method call, not a remote procedure, and unrelated to the EXEC 4-part remote-proc support this PR adds (that goes through a different execute_body path). Two fixes:
  - Reworded to: "Invoking a method on a schema-qualified multi-part object name (schema.table.column.method) is not currently supported in Babelfish".
  - Removed the dead is_xml_method branch here — the grammar forces the trailing token to be a geospatial method keyword, so it could never be an XML method (exist/value/query/nodes/modify).

  On coverage: this literal message was already exercised by the spatial suites (Test-spatial-functions, Test-spatial-functions-2, BABEL-5548, TestSpatialPoint, Test-Spatial_Index); I
  regenerated their expected outputs (including single_db and parallel_query variants) for the reworded string.

PR #4795, opened 2026-05-05

Mohitraj-g:
### Description

```
-- CASE: fails
```
```
-- COALESCE: fails
```

### Fix

 `select_common_type_for_coalesce_function` also includes a minor cleanup: the `UNKNOWNOID` (string literal) handling was simplified by replacing `ptype` in-place and falling through to the shared precedence comparison, instead of duplicating the comparison inline with a separate `continue`. This is semantically equivalent and notstrictly required for the sql_variant fix.

### Issues Resolved

BABEL-6402
BABEL-6427

cherry-picked from https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/4791

* **Use case based -**
CASE (simple, searched, multi-branch with sql_variant in 1st/2nd/3rd/last position), ISNULL, COALESCE (2-4 args), UNION ALL, UNION, INTERSECT,  EXCEPT, VALUES — all with sql_variant mixed with bit, int, varchar, float, numeric, date, datetime, time, money, smallmoney, bigint, smallint, tinyint, uniqueidentifier

* **Boundary conditions -**
NULL sql_variant (CAST(NULL AS sql_variant)) in every position for all contexts; all-NULL sql_variant args; NULL sql_variant mixed with  NULL other types 

* **Arbitrary inputs -**
SERVERPROPERTY (system functions returning sql_variant); string literals mixed with sql_variant; table column references

* **Negative test cases -**
ISNULL with sql_variant as 2nd arg (errors preserved); incompatible types (date vs numeric) still error correctly 

`babel_sqlvariant_coerce_type_before_16_14-or-17_10-or-18_4` expected output covers pre-fix behavior (view creation fails, ISNULL errors) 

`babel_sqlvariant_coerce_type` in latest upgrade schedule with post-fix expected output

* **Performance tests -**

* **Tooling impact -**
UDT based on sql_variant and UDT based on int tested against sql_variant in CASE, COALESCE, UNION 

* **Client tests -**
Covered via JDBC test framework  

### Check List


Yvinayak07 (approved):

kuntalghosh (approved):
# AI Code Review

Verdict: APPROVE

**Issue 1 - In `pltsql_coerce.c` `select_common_type_setop`: `which_expr` is set to the first sql_variant expression found, not the first expression in the list**

```c
if ((*common_utility_plugin_ptr->is_tsql_sqlvariant_datatype)(type))
{
    if (which_expr)
        *which_expr = expr;   /* set to first sql_variant, not linitial(exprs) */
    return type;
}
```

The existing code initialises `result_expr = (Node*) linitial(exprs)` at the top of the function and the main loop always sets `*which_expr` to the first expression in the list. The new pre-scan sets it to whichever expression happens to be the first sql_variant, which may be a later element. `which_expr` is used by callers (e.g. `transformSetOperationTree`) as a representative expression for error reporting and typmod derivation. Using a non-first expression is inconsistent with the rest of the function's contract and could produce confusing error messages pointing at the wrong branch. The safe choice is to set `*which_expr = (Node *) linitial(exprs)` here, matching what the main loop would do.

Suggested fix:

```c
if ((*common_utility_plugin_ptr->is_tsql_sqlvariant_datatype)(type))
{
    if (which_expr)
        *which_expr = (Node *) linitial(exprs);
    return type;
}
```

---

**Issue 2 - In `pltsql_coerce.c` `select_common_type_setop`: UDT resolution in the pre-scan uses a different helper than the main loop**

```c
/* pre-scan (new code) */
Oid base_type = get_immediate_base_type_of_UDT_internal(type);

/* main loop (existing code, CASE path) */
Oid baseType = get_immediate_base_type_of_UDT_internal(type);
```

The pre-scan and the CASE path in the main loop both call `get_immediate_base_type_of_UDT_internal`, so they are consistent. However, the non-CASE path in the main loop does not resolve UDTs at all. This means a UDT-on-sql_variant is correctly caught by the pre-scan (good), but the asymmetry is worth a comment explaining why the pre-scan must resolve UDTs even for non-CASE contexts. The existing comment only explains the early-exit problem, not the UDT motivation. This is advisory.

---

**Issue 3 - In `pltsql_coerce.c` `select_common_type_for_coalesce_function`: UDT base type is resolved before the NULL-literal check, so `ptype` may be a base type when `is_tsql_sqlvariant_datatype` is called on a NULL Const**

```c
base_type = get_immediate_base_type_of_UDT_internal(ptype);
if (OidIsValid(base_type))
    ptype = base_type;

{
    if (!(*common_utility_plugin_ptr->is_tsql_sqlvariant_datatype)(ptype))
        continue;
}
```

This is actually correct: resolving the UDT before the sql_variant check means `CAST(NULL AS babel_sqlvariant_udt_sv)` is also treated as sql_variant and not skipped. The test suite covers this case (section 15d). No change needed — this is a note confirming the logic is intentional.

---

### Notes

**PR description:** Complete and well-structured. BABEL-6402 and BABEL-6427 are referenced. Test scenarios section is thorough. The cherry-pick from PR#4791 is noted. DCO sign-off checkbox is unchecked — this is enforced at squash-merge time so it is not a blocker.

**Fix correctness:** The root cause analysis is accurate. `select_common_type_setop` bails out early when it encounters a non-char type (returning `InvalidOid`), so sql_variant in a later branch was never reached. The pre-scan correctly addresses this. The COALESCE fix correctly preserves sql_variant NULL literals in the precedence comparison instead of skipping them.

**Upgrade test coverage:** The `_before_16_14-or-17_10-or-18_4` test set correctly captures pre-fix behavior (view creation failure, ISNULL errors). The `latest` and `16_14`/`17_10` schedules use the post-fix expected output. The version boundary naming matches the pattern used by other recent PRs.

**Python validation files:** Removing `serverproperty` from `expected_create.out` and `bit_sqlvariant` from `expected_dependency.out` is consistent with the fix — these entries were artifacts of the pre-fix state and are no longer expected.

**ISNULL behavior with sql_variant as 2nd arg:** The test correctly documents that `ISNULL(non_sv, sql_variant)` still errors. This matches T-SQL semantics where ISNULL returns the type of the first argument, so sql_variant in the replacement position cannot be coerced to the check-expression type. The errors are intentional and tested.

Issue 1 is advisory (the current behavior is unlikely to cause a real problem since `which_expr` is used only for error reporting in this context, and the sql_variant expression is a valid representative). The PR is otherwise correct and well-tested.

shardgupta (approved):

PR #4798, opened 2026-05-06
fix: Add range validation in DATE to DATETIME2 cast

lejaokri:
## Description

CAST(DATE AS DATETIME2) hangs for dates outside the valid datetime2 range (0001-01-01 to 9999-12-31) because the multiplication `dateVal * USECS_PER_DAY` overflows int64.

This adds a pre-multiplication bounds check in `date_datetime2()` to raise `data out of range for datetime2` for out-of-range input dates.

## Changes

- **contrib/babelfishpg_common/src/datetime2.c**: Added bounds check before the `dateVal * USECS_PER_DAY` multiplication to prevent int64 overflow
- **test/JDBC/input/cast_date_to_datetime2.sql**: New test file with boundary and out-of-range cases
- **test/JDBC/expected/cast_date_to_datetime2.out**: Expected output for the new tests

## Testing

- Out-of-range dates (year 10000, year 999999) now raise an error instead of hanging
- Valid boundary dates (0001-01-01, 9999-12-31) continue to work correctly

Task: BABEL-4527
Signed-off-by: Kristian Lejao <klejao@amazon.com>

### DIFF
--- a/contrib/babelfishpg_common/src/datetime2.c
+++ b/contrib/babelfishpg_common/src/datetime2.c
@@ -9,6 +9,7 @@
 #include "postgres.h"
 
 #include "fmgr.h"
+#include "common/int.h"
 #include "utils/builtins.h"
 #include "utils/date.h"
 #include "utils/datetime.h"
@@ -491,7 +492,22 @@ date_datetime2(PG_FUNCTION_ARGS)
 	else if (DATE_IS_NOEND(dateVal))
 		TIMESTAMP_NOEND(result);
 	else
-		result = dateVal * USECS_PER_DAY;
+	{
+		/*
+		 * Use overflow-safe multiplication to prevent int64 overflow.
+		 * dateVal * USECS_PER_DAY can overflow for dates outside the
+		 * valid datetime2 range (0001-01-01 to 9999-12-31).
+		 */
+		int64		product;
+
+		if (pg_mul_s64_overflow((int64) dateVal, USECS_PER_DAY, &product) ||
+			!IS_VALID_DATETIME2(product))
+			ereport(ERROR,
+					(errcode(ERRCODE_DATETIME_VALUE_OUT_OF_RANGE),
+					 errmsg("data out of range for datetime2")));
+
+		result = (Timestamp) product;
+	}
 
 	PG_RETURN_TIMESTAMP(result);
 }
--- a/test/JDBC/expected/cast_date_to_datetime2.out
+++ b/test/JDBC/expected/cast_date_to_datetime2.out
@@ -1,0 +1,49 @@
+
+-- Test DATE to DATETIME2 cast with out-of-range dates (BABEL-4527)
+-- Out-of-range: year 999999 (exceeds datetime2 max of 9999-12-31)
+SELECT CAST(CAST('999999-12-31' AS DATE) AS datetime2);
+GO
+~~START~~
+datetime2
+~~ERROR (Code: 33557097)~~
+
+~~ERROR (Message: data out of range for datetime2)~~
+
+
+-- Out-of-range: year 10000 (just past valid range)
+SELECT CAST(CAST('10000-01-01' AS DATE) AS datetime2);
+GO
+~~START~~
+datetime2
+~~ERROR (Code: 33557097)~~
+
+~~ERROR (Message: data out of range for datetime2)~~
+
+
+-- Valid boundary: max datetime2 date
+SELECT CAST(CAST('9999-12-31' AS DATE) AS datetime2);
+GO
+~~START~~
+datetime2
+9999-12-31 00:00:00.0000000
+~~END~~
+
+
+-- Out-of-range: date before datetime2 min (PostgreSQL min date, 4714-11-24 BC)
+SELECT CAST(CAST('4714-11-24 BC' AS DATE) AS datetime2);
+GO
+~~START~~
+datetime2
+~~ERROR (Code: 33557097)~~
+
+~~ERROR (Message: data out of range for datetime2)~~
+
+
+-- Valid boundary: min datetime2 date
+SELECT CAST(CAST('0001-01-01' AS DATE) AS datetime2);
+GO
+~~START~~
+datetime2
+0001-01-01 00:00:00.0000000
+~~END~~
+
--- a/test/JDBC/input/cast_date_to_datetime2.sql
+++ b/test/JDBC/input/cast_date_to_datetime2.sql
@@ -1,0 +1,21 @@
+-- Test DATE to DATETIME2 cast with out-of-range dates (BABEL-4527)
+
+-- Out-of-range: year 999999 (exceeds datetime2 max of 9999-12-31)
+SELECT CAST(CAST('999999-12-31' AS DATE) AS datetime2);
+GO
+
+-- Out-of-range: year 10000 (just past valid range)
+SELECT CAST(CAST('10000-01-01' AS DATE) AS datetime2);
+GO
+
+-- Valid boundary: max datetime2 date
+SELECT CAST(CAST('9999-12-31' AS DATE) AS datetime2);
+GO
+
+-- Out-of-range: date before datetime2 min (PostgreSQL min date, 4714-11-24 BC)
+SELECT CAST(CAST('4714-11-24 BC' AS DATE) AS datetime2);
+GO
+
+-- Valid boundary: min datetime2 date
+SELECT CAST(CAST('0001-01-01' AS DATE) AS datetime2);
+GO


kuntalghosh (changes requested):
# AI Code Review

Verdict: REQUEST_CHANGES

**Issue 1 - In `test/JDBC/input/cast_date_to_datetime2.sql`: Missing lower-bound out-of-range test:**

```sql
-- Valid boundary: min datetime2 date
SELECT CAST(CAST('0001-01-01' AS DATE) AS datetime2);
GO
```

The test covers the upper out-of-range path (year 10000, year 999999) and both valid boundaries, but does not test a date below the minimum (year 0000 or earlier). PostgreSQL's DATE type supports dates back to 4713 BC, so a date like `'0000-12-31'` is a valid input that should trigger the lower-bound error path. Without this test, the lower branch of the bounds check (`dateVal < (MIN_DATETIME2 / USECS_PER_DAY)`) has no regression coverage.

Suggested fix:

```sql
-- Out-of-range: year 0000 (before datetime2 min of 0001-01-01)
SELECT CAST(CAST('0000-12-31' AS DATE) AS datetime2);
GO
```

Add the corresponding expected output entry to `cast_date_to_datetime2.out`.

### Notes
- The overflow fix itself is correct. `DateADT` is `int32`; without the bounds check, values above ~107 million days would overflow `int64` during `dateVal * USECS_PER_DAY`, causing undefined behavior. The pre-multiplication check constrains `dateVal` to `[−730119, 2921940)`, making the maximum product ~252 × 10¹⁵ — well within `int64` range.
- The cast `(Timestamp) dateVal * USECS_PER_DAY` is correct: the explicit cast widens `dateVal` to `int64` before the multiply (cast has higher precedence than `*`).
- Both `MIN_DATETIME2` and `END_DATETIME2` divide evenly by `USECS_PER_DAY`, so the integer-division thresholds are exact and the comparison operators (`<` / `>=`) correctly mirror the `IS_VALID_DATETIME2` macro semantics.
- Target branch `BABEL_6_X_DEV` is correct for a bug fix.

kuntalghosh (changes requested):
# AI Code Review

Verdict: REQUEST_CHANGES

**Issue 1 - In `test/JDBC/input/cast_date_to_datetime2.sql`: Lower-bound branch is not exercised by the test:**

```sql
-- Out-of-range: year 0000 (before datetime2 min of 0001-01-01)
SELECT CAST(CAST('0000-12-31' AS DATE) AS datetime2);
GO
```

The `0000-12-31` test case fails at the inner `CAST(... AS DATE)` step — PostgreSQL rejects year 0000 in date literals before `date_datetime2()` is ever called. The expected output confirms this: the error message is `date/time field value out of range: "0000-12-31"`, which is a date-parsing error, not the `data out of range for datetime2` error from the new bounds check. As a result, the lower-bound branch (`dateVal < (MIN_DATETIME2 / USECS_PER_DAY)`) has no test coverage.

PostgreSQL's `DateADT` is a signed 32-bit integer counting days from the PostgreSQL epoch (2000-01-01). Dates before 0001-01-01 that are representable as `DateADT` can be constructed without a string literal — for example, using `'4714-11-24 BC'::date` (the PostgreSQL minimum date). That value will parse successfully as a `DATE` and should reach `date_datetime2()`, triggering the lower-bound error.

Suggested fix — replace the `0000-12-31` test with one that actually reaches the C function:

```sql
-- Out-of-range: date before datetime2 min (PostgreSQL min date, 4714-11-24 BC)
SELECT CAST(CAST('4714-11-24 BC' AS DATE) AS datetime2);
GO
```

Add the corresponding expected output entry to `cast_date_to_datetime2.out`:

```
-- Out-of-range: date before datetime2 min (PostgreSQL min date, 4714-11-24 BC)
SELECT CAST(CAST('4714-11-24 BC' AS DATE) AS datetime2);
GO
~~START~~
datetime2
~~ERROR (Code: 33557097)~~

~~ERROR (Message: data out of range for datetime2)~~
```

**Issue 2 - In `contrib/babelfishpg_common/src/datetime2.c` at line 505: `ereport` used instead of `errsave`:**

```c
		if (dateVal < (MIN_DATETIME2 / USECS_PER_DAY) ||
			dateVal >= (END_DATETIME2 / USECS_PER_DAY))
			ereport(ERROR,
					(errcode(ERRCODE_DATETIME_VALUE_OUT_OF_RANGE),
					 errmsg("data out of range for datetime2")));
```

The existing `CheckDatetime2Range()` helper (line ~470) uses `errsave()` with an `escontext` parameter, which allows the error to be handled as a soft error by callers that pass a non-NULL `escontext`. The new `ereport(ERROR, ...)` in `date_datetime2()` always raises a hard error, bypassing any soft-error handling the caller might request. For consistency with the rest of the datetime2 error-raising pattern in this file, and to match T-SQL's statement-terminating-but-batch-continuing error semantics, the new check should use the same mechanism as `CheckDatetime2Range()`.

Suggested fix — either call `CheckDatetime2Range()` directly (passing `NULL` as `escontext` for the hard-error case), or mirror the `errsave` pattern:

```c
		if (dateVal < (MIN_DATETIME2 / USECS_PER_DAY) ||
			dateVal >= (END_DATETIME2 / USECS_PER_DAY))
			CheckDatetime2Range(MIN_DATETIME2 - 1, NULL); /* reuse existing helper */
```

Or more directly, since `date_datetime2` has no `escontext` parameter and always wants a hard error, calling `ereport` is acceptable — but the comment should note why `errsave` is not used here, to avoid future confusion with `CheckDatetime2Range`.

### Notes
- Prior review issue (missing lower-bound test) was addressed by adding the `0000-12-31` case, but as described in Issue 1, that case does not reach the new C code path. The intent was correct; the test input just needs to be a date that PostgreSQL can parse and store as a `DateADT` below the datetime2 minimum.
- The overflow fix itself is correct. `DateADT` is `int32`; without the bounds check, values above ~107 million days would overflow `int64` during `dateVal * USECS_PER_DAY`. The pre-multiplication check constrains `dateVal` to `[−730119, 2921940)`, making the maximum product well within `int64` range.
- The cast `(Timestamp) dateVal * USECS_PER_DAY` is correct: the explicit cast widens `dateVal` to `int64` before the multiply.
- Both `MIN_DATETIME2` and `END_DATETIME2` divide evenly by `USECS_PER_DAY`, so the integer-division thresholds are exact and the comparison operators correctly mirror the `IS_VALID_DATETIME2` macro semantics.
- Target branch `BABEL_6_X_DEV` is correct for a bug fix.

kuntalghosh (dismissed):
# AI Code Review

Verdict: APPROVE

### Notes
- Prior review issues addressed: (1) missing lower-bound test — now covered by the `'4714-11-24 BC'` case, which correctly reaches `date_datetime2()` and triggers `data out of range for datetime2`; (2) `ereport` vs `errsave` — the comment in the code now explains why `errsave` is not used here (no `escontext` parameter on this function).
- The overflow fix is correct. `DateADT` is `int32`; without the bounds check, values above ~107 million days would overflow `int64` during `dateVal * USECS_PER_DAY`. The pre-multiplication check constrains `dateVal` to `[−730119, 2921940)`, making the maximum product ~252 × 10¹⁵ — well within `int64` range.
- Both `MIN_DATETIME2` and `END_DATETIME2` divide evenly by `USECS_PER_DAY`, so the integer-division thresholds are exact and the comparison operators (`<` / `>=`) correctly mirror the `IS_VALID_DATETIME2` macro semantics.
- The cast `(Timestamp) dateVal * USECS_PER_DAY` is correct: the explicit cast widens `dateVal` to `int64` before the multiply (cast has higher precedence than `*`).
- Target branch `BABEL_6_X_DEV` is correct for a bug fix.

kuntalghosh (approved):
# AI Code Review

Verdict: APPROVE

### Notes
- Prior review issues addressed: (1) missing lower-bound test — now covered by the `'4714-11-24 BC'` case, which correctly reaches `date_datetime2()` and triggers `data out of range for datetime2`; (2) `ereport` vs soft-error pattern — `date_datetime2()` has no `escontext` parameter and always raises a hard error, so `ereport(ERROR, ...)` is correct here.
- The implementation was improved from the previously reviewed approach: using `pg_mul_s64_overflow` from `common/int.h` is strictly better than a pre-division threshold check. It catches overflow for any extreme `DateADT` value before the product is computed, then validates the result with `IS_VALID_DATETIME2`. The short-circuit `||` correctly prevents reading the undefined `product` when overflow occurs.
- The `(int64) dateVal` widening cast before the multiply is correct — `DateADT` is `int32`, and the explicit cast ensures the multiplication is done in 64-bit arithmetic.
- Both `MIN_DATETIME2` and `END_DATETIME2` divide evenly by `USECS_PER_DAY`, so `IS_VALID_DATETIME2` applied to the product gives exact boundary semantics.
- Advisory: the sibling functions `date_datetime()` (datetime.c), and the corresponding cast paths in datetimeoffset.c and smalldatetime.c, have the same unprotected `dateVal * USECS_PER_DAY` multiplication. Consider filing a follow-up ticket to apply the same fix there.
- Target branch `BABEL_6_X_DEV` is correct for a bug fix.

forestkeeper (approved):

PR #4799, opened 2026-05-06
fix: Add range validation in DATE to DATETIME2 cast [5.x backport]

lejaokri:
## Description

Backport of #4798 to BABEL_5_X_DEV.

CAST(DATE AS DATETIME2) hangs for dates outside the valid datetime2 range (0001-01-01 to 9999-12-31) because the multiplication `dateVal * USECS_PER_DAY` overflows int64.

This adds a pre-multiplication bounds check in `date_datetime2()` to raise `data out of range for datetime2` for out-of-range input dates.

## Changes

- **contrib/babelfishpg_common/src/datetime2.c**: Added bounds check before the `dateVal * USECS_PER_DAY` multiplication to prevent int64 overflow
- **test/JDBC/input/cast_date_to_datetime2.sql**: New test file with boundary and out-of-range cases
- **test/JDBC/expected/cast_date_to_datetime2.out**: Expected output for the new tests

## Testing

- Out-of-range dates (year 10000, year 999999) now raise an error instead of hanging
- Valid boundary dates (0001-01-01, 9999-12-31) continue to work correctly

Task: BABEL-4527
Signed-off-by: Kristian Lejao <klejao@amazon.com>

### DIFF
--- a/contrib/babelfishpg_common/src/datetime2.c
+++ b/contrib/babelfishpg_common/src/datetime2.c
@@ -9,6 +9,7 @@
 #include "postgres.h"
 
 #include "fmgr.h"
+#include "common/int.h"
 #include "utils/builtins.h"
 #include "utils/date.h"
 #include "utils/datetime.h"
@@ -491,7 +492,22 @@ date_datetime2(PG_FUNCTION_ARGS)
 	else if (DATE_IS_NOEND(dateVal))
 		TIMESTAMP_NOEND(result);
 	else
-		result = dateVal * USECS_PER_DAY;
+	{
+		/*
+		 * Use overflow-safe multiplication to prevent int64 overflow.
+		 * dateVal * USECS_PER_DAY can overflow for dates outside the
+		 * valid datetime2 range (0001-01-01 to 9999-12-31).
+		 */
+		int64		product;
+
+		if (pg_mul_s64_overflow((int64) dateVal, USECS_PER_DAY, &product) ||
+			!IS_VALID_DATETIME2(product))
+			ereport(ERROR,
+					(errcode(ERRCODE_DATETIME_VALUE_OUT_OF_RANGE),
+					 errmsg("data out of range for datetime2")));
+
+		result = (Timestamp) product;
+	}
 
 	PG_RETURN_TIMESTAMP(result);
 }
--- a/test/JDBC/expected/cast_date_to_datetime2.out
+++ b/test/JDBC/expected/cast_date_to_datetime2.out
@@ -1,0 +1,49 @@
+
+-- Test DATE to DATETIME2 cast with out-of-range dates (BABEL-4527)
+-- Out-of-range: year 999999 (exceeds datetime2 max of 9999-12-31)
+SELECT CAST(CAST('999999-12-31' AS DATE) AS datetime2);
+GO
+~~START~~
+datetime2
+~~ERROR (Code: 33557097)~~
+
+~~ERROR (Message: data out of range for datetime2)~~
+
+
+-- Out-of-range: year 10000 (just past valid range)
+SELECT CAST(CAST('10000-01-01' AS DATE) AS datetime2);
+GO
+~~START~~
+datetime2
+~~ERROR (Code: 33557097)~~
+
+~~ERROR (Message: data out of range for datetime2)~~
+
+
+-- Valid boundary: max datetime2 date
+SELECT CAST(CAST('9999-12-31' AS DATE) AS datetime2);
+GO
+~~START~~
+datetime2
+9999-12-31 00:00:00.0000000
+~~END~~
+
+
+-- Out-of-range: date before datetime2 min (PostgreSQL min date, 4714-11-24 BC)
+SELECT CAST(CAST('4714-11-24 BC' AS DATE) AS datetime2);
+GO
+~~START~~
+datetime2
+~~ERROR (Code: 33557097)~~
+
+~~ERROR (Message: data out of range for datetime2)~~
+
+
+-- Valid boundary: min datetime2 date
+SELECT CAST(CAST('0001-01-01' AS DATE) AS datetime2);
+GO
+~~START~~
+datetime2
+0001-01-01 00:00:00.0000000
+~~END~~
+
--- a/test/JDBC/input/cast_date_to_datetime2.sql
+++ b/test/JDBC/input/cast_date_to_datetime2.sql
@@ -1,0 +1,21 @@
+-- Test DATE to DATETIME2 cast with out-of-range dates (BABEL-4527)
+
+-- Out-of-range: year 999999 (exceeds datetime2 max of 9999-12-31)
+SELECT CAST(CAST('999999-12-31' AS DATE) AS datetime2);
+GO
+
+-- Out-of-range: year 10000 (just past valid range)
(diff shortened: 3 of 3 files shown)


kuntalghosh (changes requested):
# AI Code Review

Verdict: REQUEST_CHANGES

**Issue 1 - In `contrib/babelfishpg_common/src/datetime2.c` at line 507: `result = (Timestamp) dateVal * USECS_PER_DAY`:**

```c
result = (Timestamp) dateVal * USECS_PER_DAY;
```

The cast `(Timestamp) dateVal` promotes `dateVal` from `int32` to `int64` before the multiplication, which is correct. However, the established Babelfish pattern for guarding multiplications that could overflow is to use `pg_mul_s64_overflow` rather than a pre-bounds check. The pre-bounds check approach works correctly here (the bounds are exact multiples of `USECS_PER_DAY` so there is no truncation error), but it is less idiomatic and harder to verify at a glance than the overflow-safe macro pattern used elsewhere in the codebase.

Suggested fix:

```c
int64 product;
if (pg_mul_s64_overflow((int64) dateVal, USECS_PER_DAY, &product))
    ereport(ERROR,
            (errcode(ERRCODE_DATETIME_VALUE_OUT_OF_RANGE),
             errmsg("data out of range for datetime2")));
result = (Timestamp) product;
```

This is advisory — the current bounds check is mathematically correct and the fix is sound. The `pg_mul_s64_overflow` approach is preferred for consistency with how overflow is handled elsewhere.

**Issue 2 - In `test/JDBC/input/cast_date_to_datetime2.sql`: lower-bound path not directly exercised:**

```sql
-- Out-of-range: year 0000 (before datetime2 min of 0001-01-01)
SELECT CAST(CAST('0000-12-31' AS DATE) AS datetime2);
```

The test for the lower bound fails at the `DATE` parsing step (PostgreSQL rejects year 0000), not at the new `datetime2` range check. The new lower-bound guard (`dateVal < MIN_DATETIME2 / USECS_PER_DAY`) is therefore not exercised by any test. While valid PostgreSQL `DATE` values cannot go below `0001-01-01`, a comment in the test explaining why the lower-bound path is unreachable via normal SQL would make the intent clear to future readers.

Suggested fix: Add a comment to the test file:

```sql
-- Note: the lower bound (before 0001-01-01) cannot be reached via a valid DATE
-- literal because PostgreSQL DATE does not support year 0000 or earlier.
-- The guard in date_datetime2() is defensive for any future code path that
-- constructs a DateADT directly.
```

### Notes

The core fix is correct. The bounds check uses the exact same half-open interval `[MIN_DATETIME2, END_DATETIME2)` as the `IS_VALID_DATETIME2` macro, the integer division is exact (no truncation error), and the `(Timestamp)` cast correctly widens `dateVal` to `int64` before the multiplication. The fix resolves the reported hang for out-of-range dates.

The test file name `cast_date_to_datetime2.sql` is feature-based and follows naming conventions. The BABEL-4527 ticket reference is present. Target branch `BABEL_5_X_DEV` is correct for a 5.x backport.

kuntalghosh (commented):
# AI Code Review

Verdict: COMMENT

The core fix is correct and the two issues from the prior review remain unaddressed. Both were advisory, so this is a COMMENT verdict rather than REQUEST_CHANGES.

**Prior review advisory issues (still open):**

**Issue 1 - In `contrib/babelfishpg_common/src/datetime2.c` at line 507: pre-bounds check vs overflow-safe macro:**

```c
if (dateVal < (MIN_DATETIME2 / USECS_PER_DAY) ||
    dateVal >= (END_DATETIME2 / USECS_PER_DAY))
    ereport(ERROR, ...);

result = (Timestamp) dateVal * USECS_PER_DAY;
```

The pre-bounds check is mathematically correct (both `MIN_DATETIME2` and `END_DATETIME2` are exact multiples of `USECS_PER_DAY`, verified by the fact that they represent midnight boundaries). However, the more idiomatic pattern in this codebase for guarding multiplications against overflow is `pg_mul_s64_overflow`. Using it would make the overflow-safety intent immediately obvious to readers without requiring them to verify the divisibility of the constants.

Suggested fix:

```c
int64 product;
if (pg_mul_s64_overflow((int64) dateVal, USECS_PER_DAY, &product))
    ereport(ERROR,
            (errcode(ERRCODE_DATETIME_VALUE_OUT_OF_RANGE),
             errmsg("data out of range for datetime2")));
result = (Timestamp) product;
```

**Issue 2 - In `test/JDBC/input/cast_date_to_datetime2.sql`: lower-bound guard is not exercised:**

```sql
-- Out-of-range: date before datetime2 min (PostgreSQL min date, 4714-11-24 BC)
SELECT CAST(CAST('4714-11-24 BC' AS DATE) AS datetime2);
```

This test does exercise the lower-bound path — `4714-11-24 BC` is a valid PostgreSQL `DATE` value (PostgreSQL's minimum date) and its `DateADT` value is well below `MIN_DATETIME2 / USECS_PER_DAY`, so the new guard fires correctly. The prior review's concern about year `0000` not reaching the guard was valid for that specific literal, but this test case using `4714-11-24 BC` does cover the lower-bound path. No change needed here.

### Notes

The fix correctly prevents int64 overflow in `date_datetime2()` for dates outside the valid range. The bounds arithmetic is exact: both `MIN_DATETIME2` and `END_DATETIME2` are exact multiples of `USECS_PER_DAY`, so the integer division introduces no truncation error. The `(Timestamp)` cast on line 507 correctly widens `dateVal` from `int32` to `int64` before multiplication. The test file name is feature-based and follows naming conventions. BABEL-4527 is referenced. Target branch `BABEL_5_X_DEV` is correct for a 5.x backport.

The only remaining open item from the prior review is Issue 1 (advisory): consider switching to `pg_mul_s64_overflow` for consistency with how overflow is handled elsewhere in the codebase.

kuntalghosh (approved):
# AI Code Review

Verdict: APPROVE

The author has addressed both issues raised in the prior reviews. The implementation now uses `pg_mul_s64_overflow` as suggested, and the lower-bound test path is covered by the `4714-11-24 BC` case.

### Notes

Prior review issues addressed:
- **Issue 1 (overflow-safe macro):** The pre-bounds check has been replaced with `pg_mul_s64_overflow((int64) dateVal, USECS_PER_DAY, &product)` followed by `!IS_VALID_DATETIME2(product)`. This is the idiomatic pattern for guarding int64 multiplications in this codebase and makes the overflow-safety intent immediately clear.
- **Issue 2 (lower-bound test coverage):** The `4714-11-24 BC` test case (PostgreSQL's minimum date) correctly exercises the lower-bound guard. Its `DateADT` value is well below `MIN_DATETIME2 / USECS_PER_DAY`, so the new check fires as expected.

The fix is correct. The combined condition handles two distinct failure modes cleanly: (1) the multiplication itself overflows int64 for extreme dates like year 999999, and (2) the product is in-range for int64 but outside the valid datetime2 window. The `(Timestamp)` cast on the result is safe because `pg_mul_s64_overflow` guarantees no overflow occurred. The `common/int.h` include is correctly added. The test file name `cast_date_to_datetime2.sql` is feature-based and follows naming conventions. BABEL-4527 is referenced. Target branch `BABEL_5_X_DEV` is correct for a 5.x backport.

One minor observation (not blocking): the comment above the new block uses `//`-style thinking in its phrasing but is correctly formatted as a `/* */` block comment — no action needed.

PR #4806, opened 2026-05-11
Drop TVP temp tables at RPC end to fix stale plan reads

yuchi-chang:
When the same prepared SP that takes a TVP is re-executed on the same connection (sp_prepexec / sp_execute), only the first batch's rows land in the target. Subsequent sp_execute calls silently read from the first batch's temp table while the fresh data sent by each later batch is written into brand new temp tables the cached plan never references.

Root cause: TdsRecvTypeTable() creates a fresh `<param>_TDS_TVP_TEMP_TABLE_<rand>` for every TVP send, but nothing in the extension ever drops them. The temp table's OID stays valid for the life of the session, so PG's plan cache has no signal to invalidate the SP's cached plan, and the RTE keeps pointing at the first batch's table.

Fix: cleanup_tvp_temp_tables() runs at the end of every successful RPC entry (sp_executesql, sp_execute, sp_prepexec, SPCustomType) and DROPs the TDS-created TVP temp tables. The DROP fires PlanCacheRelCallback, marking dependent cached plans invalid; the next sp_execute then re-plans against the freshly created TVP temp table. Items whose tableName is NULL (composite-type path) are skipped. Error paths keep the existing tvp_lookup_list = NIL because the surrounding xact abort already discards the temp tables.

Adds JDBC regression test TestTvpReuse covering prepare-once + re-execute-N-times with distinct TVP payloads each call.

### Description

### Issues Resolved

* **Use case based -**

* **Boundary conditions -**

* **Arbitrary inputs -**

* **Negative test cases -**

* **Performance tests -**

* **Tooling impact -**

* **Client tests -**

### Check List


contrib/babelfishpg_tds/src/backend/tds/tdsrpc.c
```diff
@@ -156,9 +156,106 @@ static void TDSLogStatementCursorHandler(TDSRequestSP req, char *stmt, int optio
...
+ *
+ * Items whose tableName is NULL come from the composite-type path in
+ * DeclareSPVariables() and reference a real type relation (tableRelid),
+ * not a temp table -- those must not be dropped.
+ *
+ * Only called on success paths. Error paths just NIL the list because the
+ * surrounding transaction abort already discards any half-built temp tables.
+ */
+static void
+cleanup_tvp_temp_tables(void)
```
tanscorpio7, line 187: The approach looks good. I considered whether we could skip caching statements that reference a TVP, but that's probably harder to achieve.

One edge case to consider: a batch that does an explicit COMMIT and then errors. Likely rare in practice and can be addressed separately if needed.

tanscorpio7 (changes requested):

contrib/babelfishpg_tds/src/backend/tds/tdsrpc.c
yuchi-chang: Thanks! Agreed on both points.

contrib/babelfishpg_tds/src/backend/tds/tdsrpc.c
```diff
@@ -156,9 +156,106 @@ static void TDSLogStatementCursorHandler(TDSRequestSP req, char *stmt, int optio
...
+	xactStarted = IsTransactionOrTransactionBlock();
+
+	/* DROP needs the postgres dialect, matching TdsRecvTypeTable. */
+	set_config_option("babelfishpg_tsql.sql_dialect", "postgres",
+					  GUC_CONTEXT_CONFIG,
+					  PGC_S_SESSION, GUC_ACTION_SAVE, true, 0, false);
+
+	if (!xactStarted)
+		StartTransactionCommand();
+	PushActiveSnapshot(GetTransactionSnapshot());
```
tanscorpio7: Please check if spi_execute already handle pushing a new snapshot is none is active.

contrib/babelfishpg_tds/src/backend/tds/tdsrpc.c
```diff
@@ -156,9 +156,106 @@ static void TDSLogStatementCursorHandler(TDSRequestSP req, char *stmt, int optio
...
+
+	if ((rc = SPI_connect()) < 0)
+	{
+		set_config_option("babelfishpg_tsql.sql_dialect", "tsql",
+						  GUC_CONTEXT_CONFIG,
+						  PGC_S_SESSION, GUC_ACTION_SAVE, true, 0, false);
```
tanscorpio7: We are actually moving away from setting the sql_dialect using set_config. For example please see
```
truncate_tsql_identifier(char *ident)
```
Refernce commit: https://github.com/babelfish-for-postgresql/babelfish_extensions/commit/15c520b4aa12a49c4ee3f3329e7d59ed5a0ec5eb

tanscorpio7 (changes requested):

kuntalghosh (commented):
# AI Code Review

Verdict: COMMENT

**Issue 1 - In `contrib/babelfishpg_tds/src/backend/tds/tdsrpc.c` at line 217: dialect switching uses `set_config_option` instead of direct assignment:**

```c
set_config_option("babelfishpg_tsql.sql_dialect", "postgres",
                  GUC_CONTEXT_CONFIG,
                  PGC_S_SESSION, GUC_ACTION_SAVE, true, 0, false);
```

As noted by the reviewer, the codebase is moving away from `set_config_option` for dialect switching due to issues with nested calls corrupting the GUC stack (see commit 15c520b4). The recommended pattern is to save/restore `sql_dialect` directly:

Suggested fix:

```c
int saved_dialect = sql_dialect;

PG_TRY();
{
    sql_dialect = SQL_DIALECT_PG;
    /* ... SPI operations ... */
}
PG_FINALLY();
{
    sql_dialect = saved_dialect;
}
PG_END_TRY();
```

This also has the benefit of properly restoring the dialect on error paths without needing the separate `set_config_option` call before `elog(ERROR)` in the `SPI_connect` failure branch.

**Issue 2 - In `contrib/babelfishpg_tds/src/backend/tds/tdsrpc.c` at line 224: snapshot pushed before `SPI_connect` — leaked on `SPI_connect` failure:**

```c
if (!xactStarted)
    StartTransactionCommand();
PushActiveSnapshot(GetTransactionSnapshot());

if ((rc = SPI_connect()) < 0)
{
    set_config_option("babelfishpg_tsql.sql_dialect", "tsql", ...);
    elog(ERROR, "SPI_connect() failed ...");
}
```

If `SPI_connect` fails, the `elog(ERROR)` will longjmp. The transaction abort machinery will handle the snapshot cleanup, so this is not a correctness bug. However, the pattern is fragile — consider moving `SPI_connect` before `PushActiveSnapshot`, or wrapping the body in `PG_TRY` to ensure clean teardown. This would also address the dialect restoration concern from Issue 1.

Suggested fix:

```c
if ((rc = SPI_connect()) < 0)
    elog(ERROR, "SPI_connect() failed in TVP temp table cleanup with return code %d", rc);

if (!xactStarted)
    StartTransactionCommand();
PushActiveSnapshot(GetTransactionSnapshot());
```

Note: verify that `SPI_connect` does not require an active transaction in this context. If it does, wrapping the entire body in `PG_TRY/PG_FINALLY` (as suggested in Issue 1) is the cleaner approach.

**Issue 3 - In `contrib/babelfishpg_tds/src/backend/tds/tdsrpc.c` at line 240: no error handling if `SPI_execute` fails during DROP loop:**

```c
foreach(lc, tvp_lookup_list)
{
    TvpLookupItem *item = (TvpLookupItem *) lfirst(lc);

    if (item->tableName != NULL)
    {
        char   *query = psprintf("DROP TABLE IF EXISTS %s", item->tableName);
        SPI_execute(query, false, 0);
        pfree(query);
    }
}
```

If `SPI_execute` raises an error (e.g., due to a concurrent session holding a lock, or an unexpected catalog state), the function will longjmp without calling `SPI_finish()`, `PopActiveSnapshot()`, or restoring the dialect. While `DROP TABLE IF EXISTS` is unlikely to fail, wrapping the SPI block in `PG_TRY/PG_FINALLY` would make this robust. This is consistent with how `TdsRecvTypeTable` in `tdstypeio.c` wraps its SPI operations.

**Issue 4 - In `contrib/babelfishpg_tds/src/backend/tds/tdsrpc.c` at line 223: reviewer question about snapshot push:**

The reviewer asked whether `SPI_execute` already handles pushing a snapshot if none is active. The answer is: `SPI_execute` calls `_SPI_execute_plan` which does push a snapshot internally for each query execution. However, the outer `PushActiveSnapshot` is still needed here because `SPI_connect` itself requires an active snapshot context to set up the SPI execution environment properly. The current code is correct in this regard, but adding a brief comment explaining why the explicit snapshot push is needed would improve clarity.

### Missing Tests

The test coverage is good — the JDBC test (`TestTvpReuse`) exercises the core scenario of prepare-once + re-execute with distinct TVP payloads across 5 batches, and the dotnet test adds a third execution to existing TVP tests. No additional test gaps identified for this bug fix.

### Notes

- The reviewer's comment about `set_config_option` (Issue 1) aligns with a known pattern fix (commit 15c520b4) that addresses dialect corruption in nested call stacks. This is the most actionable feedback item.
- The approach of dropping TVP temp tables to trigger `PlanCacheRelCallback` is sound and well-documented in the function comment. The alternative of skipping plan caching for TVP-referencing statements would be more invasive.
- The PR description explains the root cause clearly but the "Issues Resolved" and "Test Scenarios Covered" template sections are left empty — these should be filled in with the relevant JIRA ticket and test scenario details.
- Prior reviewer (tanscorpio7) noted an edge case with explicit COMMIT followed by error — acknowledged by the author as a separate concern.

tanscorpio7:
> Issue 2 - In contrib/babelfishpg_tds/src/backend/tds/tdsrpc.c at line 224: snapshot pushed before SPI_connect — leaked on SPI_connect failure:

Not an issue as noted by the AI review.

> Issue 3 - In contrib/babelfishpg_tds/src/backend/tds/tdsrpc.c at line 240: no error handling if SPI_execute fails during DROP loop:

This concern is not valid. In case of error, PostgreSQL transaction cleanup takes care of cleanup of SPI_Finish and pop snapshot.

> Issue 4 - In contrib/babelfishpg_tds/src/backend/tds/tdsrpc.c at line 223: reviewer question about snapshot push:

AI feedback is incorrect. SPI_connect does not require a snapshot only a valid transaction.

contrib/babelfishpg_tds/src/backend/tds/tdsrpc.c
yuchi-chang: Switched to direct `sql_dialect = SQL_DIALECT_PG` assignment guarded by `PG_TRY`/`PG_FINALLY`, matching the pattern in `truncate_tsql_identifier` (multidb.c). Done in 7de1d409.

contrib/babelfishpg_tds/src/backend/tds/tdsrpc.c
yuchi-chang: We need to keep an outer snapshot live across the CommitTransactionCommand below, otherwise the rest of the RPC fails with "cannot execute SQL without an outer snapshot or portal".

contrib/babelfishpg_tds/src/backend/tds/tdsrpc.c
```diff
@@ -156,9 +156,111 @@ static void TDSLogStatementCursorHandler(TDSRequestSP req, char *stmt, int optio
...
+			elog(ERROR, "SPI_connect() failed in TVP temp table cleanup with return code %d",
+				 rc);
+
+		foreach(lc, tvp_lookup_list)
+		{
+			TvpLookupItem *item = (TvpLookupItem *) lfirst(lc);
+
+			if (item->tableName != NULL)
+			{
+				char	   *query = psprintf("DROP TABLE IF EXISTS %s", item->tableName);
```
tanscorpio7: Lets use quote_identifier on tablename.

contrib/babelfishpg_tds/src/backend/tds/tdsrpc.c
```diff
@@ -156,9 +156,111 @@ static void TDSLogStatementCursorHandler(TDSRequestSP req, char *stmt, int optio
...
+
+				SPI_execute(query, false, 0);
+				pfree(query);
+			}
+		}
+
+		SPI_finish();
```
tanscorpio7: Please check for return code for SPI_execute and SPI_finish simliar to SPI_connect.

tanscorpio7 (changes requested):

contrib/babelfishpg_tds/src/backend/tds/tdsrpc.c
yuchi-chang: done in b592e153

contrib/babelfishpg_tds/src/backend/tds/tdsrpc.c
yuchi-chang: done in b592e153

tanscorpio7 (approved):

tanscorpio7:
@KushaalShroff Could you please review this change.

kuntalghosh (approved):
# AI Code Review

Verdict: APPROVE

**Issue 1 - In `contrib/babelfishpg_tds/src/backend/tds/tdsrpc.c` at line 241: unqualified DROP TABLE may miss the temp table if search_path is unusual:**

```c
char	   *query = psprintf("DROP TABLE IF EXISTS %s",
                             quote_identifier(item->tableName));
```

The TVP temp tables are created in `pg_temp` (PostgreSQL always places `CREATE TEMPORARY TABLE` there), but the `DROP TABLE IF EXISTS` uses an unqualified name. If a session has `pg_temp` placed later in `search_path` and a same-named permanent table exists earlier in the path, the DROP could target the wrong relation. This is unlikely in practice (TVP temp table names include a random suffix), but qualifying with `pg_temp.` would make the intent explicit and future-proof.

Suggested fix:

```c
char	   *query = psprintf("DROP TABLE IF EXISTS pg_temp.%s",
                             quote_identifier(item->tableName));
```

**Issue 2 - In `contrib/babelfishpg_tds/src/backend/tds/tdsrpc.c` at line 266: `tvp_lookup_list = NIL` not reached on error inside cleanup:**

```c
	PG_END_TRY();

	tvp_lookup_list = NIL;
}
```

If an `elog(ERROR)` fires inside the `PG_TRY` block (e.g., from `SPI_connect`, `SPI_execute`, or `SPI_finish` failure checks), `PG_FINALLY` restores `sql_dialect` and then the error propagates. The `tvp_lookup_list = NIL` assignment after `PG_END_TRY()` is never reached. This is safe today because all callers (`SPExecuteSQL`, `SPExecute`, `SPPrepExec`, `SPCustomType`) have their own `PG_CATCH` blocks that set `tvp_lookup_list = NIL` before `PG_RE_THROW()`. However, moving the reset into `PG_FINALLY` would make the function self-contained and robust against future callers that might not follow this pattern.

Suggested fix:

```c
PG_FINALLY();
{
    sql_dialect = saved_dialect;
    tvp_lookup_list = NIL;
}
PG_END_TRY();
```

### Missing Tests

No gaps identified. The JDBC test (`TestTvpReuse`) exercises the core scenario of prepare-once + re-execute with 5 distinct TVP payloads, verifying all 10 rows land in the target table. The dotnet test adds a third execution to existing TVP stored procedure tests, confirming the fix works across both drivers.

### Notes

- Prior review issues addressed: the `set_config_option` dialect switching has been replaced with direct `sql_dialect` assignment guarded by `PG_TRY`/`PG_FINALLY` (matching the pattern in `truncate_tsql_identifier`). The `quote_identifier` call and `SPI_execute`/`SPI_finish` return code checks were added per reviewer feedback. All prior blocking issues from tanscorpio7 are resolved.
- The PR description explains the root cause clearly in the opening paragraph but leaves the template sections ("Issues Resolved", "Test Scenarios Covered") empty. These should be filled in with the relevant JIRA ticket and test details before merge.
- The approach of dropping TVP temp tables to trigger `PlanCacheRelCallback` is sound. The alternative of skipping plan caching for TVP-referencing statements would be more invasive and could regress performance for non-TVP queries in the same SP.
- Both advisory issues above are hardening suggestions — neither represents a correctness bug in the current code given the existing caller patterns.

KushaalShroff (approved):

